PR #309, opened 2016-10-23
EventsSwitch is now an interface

kobigurk:


ebuchman:
see latest handshake branch, which has been rebased on latest develop to include this and more!

PR #311, opened 2016-10-31
Update Dockerfile

faddat:
Nothing at all major yet, just updated to use the latest go and the latest node.  Might have destroyed it all for all I know.


ebuchman:
let's save this for when we get docker image building and shipping into the CI pipeline. Also, please all pull requests against develop

Issue #315, opened 2016-11-16
do sign bytes with struct literals


ebuchman:
rather than manually forming the json ...

zballs:
opened pull request

PR #318, opened 2016-11-16
sign bytes with struct

zballs:


ebuchman:
Thanks for this. Problem is, the fields of structs are encoded in the order they are defined, whereas we'd like the canonical form to be alphabetical. So we need to either change all structs to have their fields alphabetical, or use struct literals with the right field ordering only in the WriteSignBytes functions. I'm into the latter. See, for example, https://github.com/tendermint/lil-voterin/blob/master/types/tx.go#L57

ebuchman:
pinging issue https://github.com/tendermint/tendermint/issues/315

types/proposal.go
```diff
@@ -41,11 +41,22 @@ func (p *Proposal) String() string {
...
-	p.BlockPartsHeader.WriteSignBytes(w, n, err)
-	wire.WriteTo([]byte(Fmt(`,"height":%v,"pol_block_id":`, p.Height)), w, n, err)
-	p.POLBlockID.WriteSignBytes(w, n, err)
-	wire.WriteTo([]byte(Fmt(`,"pol_round":%v`, p.POLRound)), w, n, err)
-	wire.WriteTo([]byte(Fmt(`,"round":%v}}`, p.Round)), w, n, err)
+
+	wire.WriteJSON(
+		struct {
+			ChainID          string        `json:"chain_id"`
+			BlockPartsHeader PartSetHeader `json:"block_parts_header"`
```
ebuchman: all this should be encapsulated in "proposal". Can we even do nested anonymous structs?

types/vote.go
```diff
@@ -57,10 +57,28 @@ type Vote struct {
...
 func (vote *Vote) WriteSignBytes(chainID string, w io.Writer, n *int, err *error) {
-	wire.WriteTo([]byte(Fmt(`{"chain_id":"%s"`, chainID)), w, n, err)
-	wire.WriteTo([]byte(`,"vote":{"block_id":`), w, n, err)
-	vote.BlockID.WriteSignBytes(w, n, err)
-	wire.WriteTo([]byte(Fmt(`,"height":%v,"round":%v,"type":%v}}`, vote.Height, vote.Round, vote.Type)), w, n, err)
+
+	wire.WriteJSON(
+		struct {
+			ChainID          string           `json:"chain_id"`
+			BlockID          BlockID          `json:"block_id"`
```
ebuchman: same as above, should be nested under a "vote"

ebuchman:
I rebased on latest develop, squashed your commits into one, and decided to go with explicit structs rather than literals. Thanks for your help. See here: https://github.com/tendermint/tendermint/pull/325

PR #326, opened 2016-12-03
Develop

charltonh:
Testcase


charltonh:
Changes pull request

state/execution_test.go
```diff
@@ -23,8 +28,124 @@ var (
...
+			txs = append(txs, txdata)
+			valtxns++
+		} else {
+			txdata = types.Tx([]byte{byte(0), byte(0)})
+			txs = append(txs, txdata)
+			invtxns++
+		}
+
+		// Create an event switch to listen for the newly inserted txn
+		eventSwitch := events.NewEventSwitch()
```
ebuchman: lets move this (and the start) outside the for loop

state/execution_test.go
```diff
@@ -23,8 +28,124 @@ var (
...
+		block, parts := types.MakeBlock(i, chainID, txs, lastCommit, prevBlockID, valHash, state.AppHash, testPartSize)
+		fmt.Println("i =", i)
+		fmt.Println("BlockID:", prevBlockID)
+
+		err2 := state.ApplyBlock(eventSwitch, proxyApp.Consensus(), block, block.MakePartSet(testPartSize).Header(), mempool)
+		if err2 != nil {
+			t.Fatal(i, err2)
+		}
+
+		voteSet := types.NewVoteSet(chainID, i, 0, types.VoteTypePrecommit, state.Validators)
```
ebuchman: this stuff is unecessary

ebuchman:
Can we get this rebased on latest develop and commits squashed to one commit, please?

ebuchman:
also would be nice to clean up / remove some of the `fmt.Print`, as well as to have blocks with more than one txs (eg. a block with 10 invalid txs, then one with 1 valid and 9 invalid, then 2 and 8, and so on to 10 and 0). at that point, best to use the same channel for all events instead of calling `subscribeToEvent` each time

ebuchman:
No need to close the PR - if you just push to the same branch (even force push to overwrite commits) it will update the PR automagically!

charltonh:
Squashed and merged commits

PR #336, opened 2016-12-18
More fixes

ebuchman:
- Use Signature instead of SignatureEd25519 everywhere
- Flush the autofile on each write to the cswal
- Automate building the cswal test files


ebuchman:
Replaced by https://github.com/tendermint/tendermint/pull/337

PR #339, opened 2016-12-21
p2p crash all nodes test

akalyaev:
Test: we crash all the nodes and bring them back up a few times

How do we check: get a height from any peer, and then wait till all peers reporting a height higher than that one

STATUS: **PASSING**


ebuchman:
rebased and a few commits squashed into here: https://github.com/tendermint/tendermint/pull/343

Thanks Anton!

PR #343, opened 2016-12-23
Crash/Restart tests

ebuchman:
- More tests:
  - Crash all nodes in net and restart them (https://github.com/tendermint/tendermint/pull/339)
  - Crash a single node at pre-defined break points and restart 
    - replace AppHashIsStale by saving an IntermediateState


consensus/state.go
```diff
@@ -362,6 +362,14 @@ func (cs *ConsensusState) OnStart() error {
 		// let's go for it anyways, maybe we're fine
 	}
 
+	// If the latest block was applied in the tmsp handshake,
```
ebuchman: yep, we want the cswal to actually handle all the cases: https://github.com/tendermint/tendermint/issues/352

PR #351, opened 2017-01-03
make progress on precommit votes optional optional

akalyaev:
See #348

### DIFF
--- a/config/tendermint_test/config.go
+++ b/config/tendermint_test/config.go
@@ -98,7 +98,7 @@ func ResetConfig(localPath string) cfg.Config {
 	mapConfig.SetDefault("timeout_precommit", 10)
 	mapConfig.SetDefault("timeout_precommit_delta", 1)
 	mapConfig.SetDefault("timeout_commit", 10)
-	mapConfig.SetDefault("skip_timeout_commit", false)
+	mapConfig.SetDefault("skip_timeout_commit", true)
 	mapConfig.SetDefault("mempool_recheck", true)
 	mapConfig.SetDefault("mempool_recheck_empty", true)
 	mapConfig.SetDefault("mempool_broadcast", true)
--- a/config/tendermint_test/config.go
+++ b/config/tendermint_test/config.go
@@ -98,6 +98,7 @@ func ResetConfig(localPath string) cfg.Config {
 	mapConfig.SetDefault("timeout_precommit", 10)
 	mapConfig.SetDefault("timeout_precommit_delta", 1)
 	mapConfig.SetDefault("timeout_commit", 10)
+	mapConfig.SetDefault("skip_timeout_commit", false)
 	mapConfig.SetDefault("mempool_recheck", true)
 	mapConfig.SetDefault("mempool_recheck_empty", true)
 	mapConfig.SetDefault("mempool_broadcast", true)
--- a/consensus/reactor_test.go
+++ b/consensus/reactor_test.go
@@ -132,6 +132,49 @@ func TestValidatorSetChanges(t *testing.T) {
 	// TODO: test more changes!
 }
 
+// Check we can make blocks with skip_timeout_commit=false
+func TestReactorWithTimeoutCommit(t *testing.T) {
+	N := 4
+	css := randConsensusNet(N, "consensus_reactor_with_timeout_commit_test", newMockTickerFunc(false))
+
+	// override default SkipTimeoutCommit == true for tests
+	for i := 0; i < N; i++ {
+		css[i].timeoutParams.SkipTimeoutCommit = false
+	}
+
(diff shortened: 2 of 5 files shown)


consensus/reactor_test.go
```diff
@@ -65,6 +65,12 @@ func TestValidatorSetChanges(t *testing.T) {
 	nPeers := 8
 	nVals := 4
 	css := randConsensusNetWithPeers(nVals, nPeers, "consensus_val_set_changes_test", newMockTickerFunc(true))
+
+	// otherwise, the test is failing with timeout error
```
akalyaev: I don't see how it is connected for now :(

akalyaev: Ok, I see now. It's either that or change onlyOnce to false: `newMockTickerFunc(false)` (otherwise, ticker only fires once and cannot make it to a new height >= 2).

akalyaev: But if I pass false to `newMockTickerFunc`, I get `nil` vote (how?):

```
INFO[01-04|10:45:00] validateBlock                            module=consensus vote=nil-Vote
[signal 0xb code=0x1 addr=0x0 pc=0x4bf58b]

goroutine 2693 [running]:
panic(0xb31920, 0xc82000e0d0)
        /usr/local/go/src/runtime/panic.go:464 +0x3e6
github.com/tendermint/tendermint/consensus.validateBlock(0xc82562f040, 0xc824b13860, 0x0, 0x0)
        /home/vagrant/go/src/github.com/tendermint/tendermint/consensus/reactor_test.go:168 +0x53b
github.com/tendermint/tendermint/consensus.waitForAndValidateBlock.func1(0xc8254574e0, 0x0)
        /home/vagrant/go/src/github.com/tendermint/tendermint/consensus/reactor_test.go:145 +0x3a3
created by github.com/tendermint/tendermint/consensus.timeoutWaitGroup
        /home/vagrant/go/src/github.com/tendermint/tendermint/consensus/reactor_test.go:179 +0x8b
exit status 2
```

akalyaev: can't understand how `nil` makes his way into `cs.LastCommit.Precommits` :(

ebuchman: the commits are nil unless explicitly received otherwise. i think for the tests its better to not use the timeout_commit, and to wait for all the votes, because otherwise we can end up with timing dependencies in the testing code which can lead to nondeterministic failures. that was part of the reason for this change originally. but for live chains we want to be able to use timeout_commit. so maybe the way to test it is to start a 4 validator testnet but with only 3 validators online and have them all send votes and then after timeout_commit theyll move to the next height

config/tendermint_test/config.go
```diff
@@ -98,6 +98,7 @@ func ResetConfig(localPath string) cfg.Config {
 	mapConfig.SetDefault("timeout_precommit", 10)
 	mapConfig.SetDefault("timeout_precommit_delta", 1)
 	mapConfig.SetDefault("timeout_commit", 10)
+	mapConfig.SetDefault("skip_timeout_commit", false)
```
ebuchman: lets make it true for tests

consensus/state.go
```diff
@@ -32,6 +33,7 @@ type TimeoutParams struct {
 	Precommit0     int
 	PrecommitDelta int
 	Commit0        int
+	SkipCommit0    bool
```
ebuchman: why not just `SkipTimeoutCommit` ?

akalyaev: to be consistent with the naming above? to be honest, I like `SkipTimeoutCommit` more

consensus/reactor_test.go
akalyaev: @ebuchman do you want me to write another test with 4 validators and `skip_timeout_commit` set to false?

akalyaev:
I think we're done here

consensus/reactor_test.go
```diff
@@ -131,6 +132,62 @@ func TestValidatorSetChanges(t *testing.T) {
 	// TODO: test more changes!
 }
 
+// Check we can make blocks with skip_timeout_commit=false
+func TestReactorWithTimeoutCommit(t *testing.T) {
```
ebuchman: I don't think this does what we want since it only creates 3 validators. We want 4 validator, where eg one of them is offline. So let's copy `TestReactor` with N=4 but use `N-1` in the for loop, the MakeConnectedSwitches, and the timeoutWaitGroup.

akalyaev: fixed in https://github.com/tendermint/tendermint/pull/351/commits/e46f14be896a739868028db2780b2ff0d07cff6d

ebuchman:
rebased here: https://github.com/tendermint/tendermint/pull/360

Issue #245, opened 2016-08-07
AppendTx -> DeliverTx


ebuchman:
The literature on consensus/atomic broadcast commonly refers to "broadcasting" and "delivering" a transaction. The former corresponds to our "broadcast_tx" rpc calls. The later to the AppendTx TMSP request. I propose we rename it DeliverTx to conform with tradition and emphasize the meaning.

Also, I just thought of TMSP -> ABCI, where ABCI is "Atomic BroadCast Interface", but it could also stand for "A BlockChain Interface" :D 

Related: https://github.com/tendermint/tendermint/issues/223

PR #365, opened 2017-01-12
Rename

ebuchman:


ebuchman:
https://github.com/tendermint/tendermint/issues/245 https://github.com/tendermint/tendermint/issues/223

Issue #335, opened 2016-12-14
PEX issues


ebuchman:
yeah, pex still has a problem with a former seed node not creating an adress book. therefor if you restart a seed node, it will not connect to the other peers.

you can manually work around by restarting the seed node with a list of seeds

akalyaev:
So strange to see `pex` code in master branch, but no code in develop 😕

ebuchman:
yes, I should have merged it to develop right after releasing 0.7.4 but did not for some reason. I did merge it to develop earlier today before the release-0.8.0 branch was pushed so should be there now

akalyaev:
> so should be there now

nope

Last commit:

```
commit c1952523cdb42939a71143e47c283db952366db2
Author: Ethan Buchman <ethan@coinculture.info>
Date:   Thu Jan 12 19:20:26 2017 -0500

    update glide
```

ebuchman:
Thanks, my bad - just pushed

ebuchman:
sorry, I just forced push right away afterwards because I pushed one commit too many, in case that trips you up

akalyaev:
Fixed the bug above by saving `seeds` to `addrBook`. But now I've encountered another bug: for some reason, when I restart the app with `--pex` flag (without `--seeds`), it keeps failing.

melekes:
The problem above: I wasn’t running other peers with `—pex`, so they were failing upon receiving a message for an unknown channel [FIXED]

faddat:
Is a peer exchange for all of Ur-Tendermint, voluntary society of aligned blockchains, or is it for a single blockchain?  

I ask because I'm still thinking about the DHT situation.  Imagine if genesis.json looked more like this instead:

chain=546cf15f724d19c4319cc17b179d7e035f89c1f4
billy=validatorpubkey
jimmy=validatorpubkey
sarah=validatorpubkey

Same technique as a magnet link, and when's the last time you had any trouble with one of those?  The nodes themselves should also have some sort of PEX that allows a single, connected node in the swarm to give all the needed information (IP addresses) to any new, "fresh" node that simply knows a list of keys & values and has the special key "chain" that is always used to get the addresses of the other peers.  

Operational simplicity is a feature and a virtue at the same time!

If you want to move chain bootstraps in this direction, I'll look into how to implement this.

melekes:
[...]

PR #370, opened 2017-01-16
PEX issues #335

melekes:
This is a 2 PR change fixing #335

2nd PR: https://github.com/tendermint/go-p2p/pull/12

In this PR:
- test p2p pex reactor
- save seeds when they are provided to avoid losing them.

Saving seeds while starting the node might look like a hack, but while I was exploring the options, this was the best possible solution I could think of. If you think different, please let me know.

### DIFF
--- a/blockchain/pool.go
+++ b/blockchain/pool.go
@@ -26,7 +26,7 @@ var peerTimeoutSeconds = time.Duration(15) // not const so we can override with 
 	in sequence from peers that reported higher heights than ours.
 	Every so often we ask peers what height they're on so we can keep going.
 
-	Requests are continuously made for blocks of heigher heights until
+	Requests are continuously made for blocks of higher heights until
 	the limits. If most of the requests have no available peers, and we
 	are not at peer limits, we can probably switch to consensus reactor
 */
--- a/node/node.go
+++ b/node/node.go
@@ -114,10 +114,8 @@ func NewNode(config cfg.Config, privValidator *types.PrivValidator, clientCreato
 	sw.AddReactor("CONSENSUS", consensusReactor)
 
 	// Optionally, start the pex reactor
-	// TODO: this is a dev feature, it needs some love
 	if config.GetBool("pex_reactor") {
 		addrBook := p2p.NewAddrBook(config.GetString("addrbook_file"), config.GetBool("addrbook_strict"))
-		addrBook.Start()
 		pexReactor := p2p.NewPEXReactor(addrBook)
 		sw.AddReactor("PEX", pexReactor)
 	}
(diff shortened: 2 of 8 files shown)


melekes:
Sample test output:

```
----------------------------------------------------------------------
Testing pex creates the addrbook and uses it if seeds are not provided
(assuming peers are started with pex enabled)
1. restart peer 1
local_testnet_1
local_testnet_1
starting tendermint peer ID=1
af1ed95319785f2a24896139bf4d750cc9782e3285d5114cf04e67b29c4bb05b
with the following addrbook:
{
        "Key": "9a8257adb4be92e57da5414b",
        "Addrs": [
                {
                        "Addr": {
                                "IP": "172.57.0.102",
                                "Port": 46656
                        },
                        "Src": {
                                "IP": "172.57.0.101",
                                "Port": 46656
                        },
                        "Attempts": 0,
                        "LastAttempt": "2017-01-16T14:50:42.949996907Z",
                        "LastSuccess": "0001-01-01T00:00:00Z",
                        "BucketType": 1,
                        "Buckets": [
                                194
                        ]
                },
                {
                        "Addr": {
                                "IP": "172.57.0.103",
                                "Port": 46656
                        },
                        "Src": {
                                "IP": "172.57.0.101",
                                "Port": 46656
                        },
                        "Attempts": 0,
                        "LastAttempt": "2017-01-16T14:50:42.951182011Z",
                        "LastSuccess": "0001-01-01T00:00:00Z",
                        "BucketType": 1,
                        "Buckets": [
                                240
                        ]
                },
                {
                        "Addr": {
                                "IP": "172.57.0.104",
                                "Port": 46656
                        },
                        "Src": {
                                "IP": "172.57.0.101",
                                "Port": 46656
                        },
                        "Attempts": 0,
                        "LastAttempt": "2017-01-16T14:50:42.951244649Z",
                        "LastSuccess": "0001-01-01T00:00:00Z",
                        "BucketType": 1,
                        "Buckets": [
                                95
                        ]
                }
        ]
}
starting test client container with CMD=test/p2p/pex/check_peer.sh 1 4
2. wait until peer 1 connects to other nodes using pex reactor
... peers count = , expected = 3
... peers count = 1, expected = 3
... peers count = 1, expected = 3
... peers count = 1, expected = 3
... peers count = 3, expected = 3
... successful

PASS
```

ebuchman:
Nice tests. I'm not sure if adding seeds right away is good incase they're bung seeds. If we fail to connect, is the address eventually removed?

Also, I might be missing something, but my read on the problem was that if you start node1 by itself, and then start node2 with `--seeds node1`, and then kill and restart node1 (still without using --seeds), it would fail to connect back to node1. Can you replicate that? Is that maybe solved by the other go-p2p PR?

melekes:
> If we fail to connect, is the address eventually removed?
yes
> Can you replicate that?
will try right away

melekes:
@ebuchman can't replicate. added a test just in case.

ebuchman:
Rebased and merged into https://github.com/tendermint/tendermint/tree/release-0.9.0 .

PR #373, opened 2017-01-18
Check `peer.Send` failures

melekes:
See #174 for the issue & discussion.

### DIFF
--- a/blockchain/reactor.go
+++ b/blockchain/reactor.go
@@ -3,11 +3,10 @@ package blockchain
 import (
 	"bytes"
 	"errors"
-	"fmt"
 	"reflect"
 	"time"
 
-	. "github.com/tendermint/go-common"
+	cmn "github.com/tendermint/go-common"
 	cfg "github.com/tendermint/go-config"
 	"github.com/tendermint/go-p2p"
 	"github.com/tendermint/go-wire"
@@ -17,7 +16,9 @@ import (
 )
 
 const (
-	BlockchainChannel      = byte(0x40)
+	// BlockchainChannel is a channel for blocks and status updates (`BlockStore` height)
+	BlockchainChannel = byte(0x40)
+
 	defaultChannelCapacity = 100
 	defaultSleepIntervalMS = 500
 	trySyncIntervalMS      = 100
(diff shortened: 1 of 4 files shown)


consensus/reactor.go
```diff
@@ -365,14 +364,20 @@ func makeRoundStepMessages(rs *RoundState) (nrsMsg *NewRoundStepMessage, csMsg *
...
-func (conR *ConsensusReactor) sendNewRoundStepMessage(peer *p2p.Peer) {
+func (conR *ConsensusReactor) sendNewRoundStepMessages(peer *p2p.Peer) {
+	log := log.New("peer", peer)
+
 	rs := conR.conS.GetRoundState()
 	nrsMsg, csMsg := makeRoundStepMessages(rs)
 	if nrsMsg != nil {
-		peer.Send(StateChannel, struct{ ConsensusMessage }{nrsMsg})
+		if !peer.Send(StateChannel, struct{ ConsensusMessage }{nrsMsg}) {
+			log.Warn("Failed to send NewRoundStepMessage to peer")
```
melekes: not sure whenever we should `log.Warn` every unsuccessful Send

blockchain/reactor.go
```diff
@@ -55,12 +57,13 @@ type BlockchainReactor struct {
...
 
+// NewBlockchainReactor returns new reactor instance.
 func NewBlockchainReactor(config cfg.Config, state *sm.State, proxyAppConn proxy.AppConnConsensus, store *BlockStore, fastSync bool) *BlockchainReactor {
 	if state.LastBlockHeight == store.Height()-1 {
-		store.height -= 1 // XXX HACK, make this better
+		store.height-- // XXX HACK, make this better
 	}
 	if state.LastBlockHeight != store.Height() {
-		PanicSanity(Fmt("state (%v) and store (%v) height mismatch", state.LastBlockHeight, store.Height()))
+		cmn.PanicSanity(fmt.Sprintf("state (%v) and store (%v) height mismatch", state.LastBlockHeight, store.Height()))
```
ebuchman: do you think we should stop using cmn.Fmt?

melekes: I've created this a long time ago)) and changed my mind since then. I do think we should rely on common as little as possible. But here, it is somewhat justified

melekes:
> Also, maybe we can test this on mintnet-k8 to see how chatty the warnings are?

good idea!

melekes:
https://gist.github.com/melekes/4aee5707a027ed5af18d1bcfc974ce54

a lot of `Send failed` messages when I block the incoming port on one pod/node.

melekes:
That said, we should either stick with `go-p2p/connection` logging or move to a higher level (ie `consensus/reactor` checking result of `peer.Send` and reacting to it). For now, removed duplicated logging from `consensus/reactor`.

ebuchman:
Ok, will merge PR and leave original issue open about higher level management of failed sends. Tho the checks in here should hopefully help (ie. calling HasX unless the Send succeeds)

PR #374, opened 2017-01-19
Tx indexing (WIP)

melekes:
See #237

### DIFF
--- a/state/execution.go
+++ b/state/execution.go
@@ -3,27 +3,27 @@ package state
 import (
 	"bytes"
 	"errors"
+	"fmt"
 
 	"github.com/ebuchman/fail-test"
 
+	abci "github.com/tendermint/abci/types"
 	. "github.com/tendermint/go-common"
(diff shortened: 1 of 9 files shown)


state/execution.go
```diff
@@ -224,26 +233,26 @@ func (s *State) validateBlock(block *types.Block) error {
...
 	}
 
 	// lock mempool, commit state, update mempoool
 	err = s.CommitStateUpdateMempool(proxyAppConn, block, mempool)
 	if err != nil {
-		return errors.New(Fmt("Commit failed for application: %v", err))
+		return fmt.Errorf("Commit failed for application: %v", err)
 	}
+
+	for _, r := range txResults {
```
melekes: Is this the best place to save transactions?

ebuchman:
Cool, great start. See comment in #237 about new structures. ExecBlock funcs should probably be returning BlockResults. Not sure where the best place to save txs is yet, or if the txindexer should be independent from the blockstore

PR #378, opened 2017-01-23
`make dist` for creating binaries and packaging them for distribution

melekes:
Preface: if we want to engage more people (not just hardcore devs), I believe we must provide statically compiled binaries for each release. Besides, this could be easily done with Golang. 

Story: as a dev, if I just want to try Tendermint (or I am planning to write my app in Java), I don't want to install Golang, I just want `brew install tendermint` or similar (btw, what do you think about adding Tendermint to `brew`).

This will fix many issues (https://tendermint.com/intro/getting-started/install):

1. user don't have to install Golang
2. user don't have to learn Golang tools (go get)
3. user won't be having any problems with updated dependencies

Usage
=====

```
$ make dist
```
will create binaries for almost all platforms and will package them for distribution. It will also create a tag (e.g. "v0.8.0").

If you've already created a tag, set `NOTAG` env variable to true:
```
$ NOTAG=1 make dist
```

RELEASE could be set explicitly via env variable. Otherwise, the script will try to fetch it from `version/version.go`

Broken builds
==========

Our code currently does not compile for freebsd/amd64:

```
--> freebsd/amd64 error: exit status 2
Stderr: # github.com/tendermint/tendermint/vendor/github.com/spf13/pflag
vendor/github.com/spf13/pflag/uint16.go:89: illegal NUL byte
vendor/github.com/spf13/pflag/uint16.go:89: syntax error: illegal character U+0000
```

and solaris/amd64:

```
--> solaris/amd64 error: exit status 1
Stderr: vendor/github.com/tendermint/log15/root.go:7:2: no buildable Go source filesin /go/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/log15/term
```

Example output
===========

```
==> Results:
total 49M
-rw-r--r-- 1 vagrant vagrant 5.4M Jan 23 11:12 tendermint_0.8.0_darwin_386.zip
-rw-r--r-- 1 vagrant vagrant 5.7M Jan 23 11:12 tendermint_0.8.0_darwin_amd64.zip
-rw-r--r-- 1 vagrant vagrant 6.1K Jan 23 10:51 tendermint_0.8.0_.DS_Store
-rw-r--r-- 1 vagrant vagrant 5.4M Jan 23 11:12 tendermint_0.8.0_freebsd_386.zip
-rw-r--r-- 1 vagrant vagrant 5.2M Jan 23 11:12 tendermint_0.8.0_freebsd_arm.zip
-rw-r--r-- 1 vagrant vagrant 5.4M Jan 23 11:12 tendermint_0.8.0_linux_386.zip
-rw-r--r-- 1 vagrant vagrant 5.6M Jan 23 11:12 tendermint_0.8.0_linux_amd64.zip
-rw-r--r-- 1 vagrant vagrant 5.2M Jan 23 11:12 tendermint_0.8.0_linux_arm.zip
-rw-r--r-- 1 vagrant vagrant 1002 Jan 23 11:12 tendermint_0.8.0_SHA256SUMS
-rw-r--r-- 1 vagrant vagrant 5.4M Jan 23 11:12 tendermint_0.8.0_windows_386.zip
-rw-r--r-- 1 vagrant vagrant 5.6M Jan 23 11:12 tendermint_0.8.0_windows_amd64.zip
```

What could be done
===============

1. sign tag with GPG signature **(easily done)**
2. sign tendermint_VERSION_SHA256SUMS with GPG signature **(easily done)**
3. figure out why our code does not compile for freebsd and solaris, and fix it

Inspiration: https://www.hashicorp.com/security.html

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,7 +1,7 @@
 package version
 
 const Maj = "0"
-const Min = "8" // validator set changes, tmsp->abci, app persistence/recovery, BFT-liveness fix
-const Fix = "0" //
+const Min = "8"
+const Fix = "0"
 
-const Version = Maj + "." + Min + "." + Fix
+const Version = "0.8.0"
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[Makefile]
+indent_style = tab
+
+[*.sh]
+indent_style = tab
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ remote_dump
 vendor
 .vagrant
 test/p2p/data/
+.glide
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,10 @@ build:
 build_race:
 	go build -race -o build/tendermint ./cmd/tendermint
 
+# dist builds binaries for all platforms and packages them for distribution
+dist:
+	@BUILD_TAGS='$(BUILD_TAGS)' sh -c "'$(CURDIR)/scripts/dist.sh'"
+
 test: build
 	@echo "--> Running go test"
 	@go test $(PACKAGES)
--- a/Makefile
+++ b/Makefile
@@ -1,59 +1,65 @@
-.PHONY: get_deps build all list_deps install
+GOTOOLS = \
+					github.com/mitchellh/gox \
+					github.com/Masterminds/glide
+PACKAGES=$(shell go list ./... | grep -v '/vendor/')
+BUILD_TAGS?=tendermint
+TMROOT = $${TMROOT:-$$HOME/.tendermint}
 
 all: get_deps install test
 
-TMROOT = $${TMROOT:-$$HOME/.tendermint}
-define NEWLINE
-
-
-endef
-NOVENDOR = go list github.com/tendermint/tendermint/... | grep -v /vendor/
-
 install: get_deps
-	go install github.com/tendermint/tendermint/cmd/tendermint
+	@go install ./cmd/tendermint
 
 build:
-	go build -o build/tendermint github.com/tendermint/tendermint/cmd/tendermint
+	go build -o build/tendermint ./cmd/tendermint
 
 build_race:
-	go build -race -o build/tendermint github.com/tendermint/tendermint/cmd/tendermint
+	go build -race -o build/tendermint ./cmd/tendermint
 
 test: build
-	go test `${NOVENDOR}`
+	@echo "--> Running go test"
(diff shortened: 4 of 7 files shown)


version/version.go
```diff
@@ -1,7 +1,7 @@
 package version
 
 const Maj = "0"
-const Min = "8" // validator set changes, tmsp->abci, app persistence/recovery, BFT-liveness fix
```
melekes: Is this the best place for such comments? maybe we should create `CHANGELOG.md` instead?

ebuchman: I've been keeping the changelog in the release notes on github for every release. Not opposed to also adding them all into a CHANGELOG.md. the comment here is just a rough reference if you're cruising by

ethanfrey:
I just saw this issue.  Yes, I love the idea!

Re outstanding points 1 and 2.

I think the gpg signing should happen outside of the `make dist`. It can just build stuff, if we want signing, this is a special key kept on a special machine, so we shouldn't make that part of dist, which I may want to do on my own without signing.  It would be great to have a second command to do that, maybe `make signed-dist`, which calls `make dist` and then gpg signs using the location of a key set in an environment variable???

`GPG_KEY=/home/foo/.gpg/tm.key make signed-dist` ?

The other question... can we check all of them behave properly to a basic test?  I know there are some issues with cgo when cross-compiling, maybe you dealt with that, maybe it is not an issue,but it would be nice to have an integration test.

Also, that is some crazy shell scripting skills.  Props!

melekes:
> I think the gpg signing should happen outside of the make dist. It can just build stuff, if we want signing, this is a special key kept on a special machine, so we shouldn't make that part of dist, which I may want to do on my own without signing.

or we can have a flag (`NOSIGN`)

> The other question... can we check all of them behave properly to a basic test? I know there are some issues with cgo when cross-compiling, maybe you dealt with that, maybe it is not an issue,but it would be nice to have an integration test.

interesting question! for now I just disabled cgo (do we have C libraries?) 👿

ethanfrey:
I think this command would be used more often without signing than with signing.  At least pre-1.0. If I want to deploy a develop version onto my testnet, I would like to `make dist` and then `scp`.  I could add something like that to the (advanced) tutorials as well.  But maybe longer term, the signing is the more common use case.

Either way is fine actually. Just aesthetics.

I would also like to add a similar `make dist` command to basecoin, so that can also be deployed and hosted easily, if it is going to be our ultimate prototype app. Maybe you can help me with that, or just do this, once this PR is finalized and merged?  Whichever is easier for you.

melekes:
> If I want to deploy a develop version onto my testnet, I would like to make dist and then scp.

hm.. we have `make build`, which will build a single binary. You can change the desired arch/os by using `GOOS` and `GOARCH` env variables: `GOOS=windows GOARCH=amd64 make build`

ethanfrey:
Okay, then `make dist` for official releases, and document how to use `make build` with the flags.  Sounds like a good solution.

PR #379, opened 2017-01-24
Output all tendermint commands

melekes:
I find it very confusing when I am going through the tutorial (https://tendermint.com/intro/getting-started/install) and **do not see `init` in the list of available commands**.

Also, now we just dump all the flags, even whose not relevant to the given command:

```
./build/tendermint replay wal --gee
unknown flag: --gee
Usage of main:
      --abci string         Specify abci transport (socket | grpc) (default "socket")
      --fast_sync           Fast blockchain syncing (default true)
      --grpc_laddr string   GRPC listen address (BroadcastTx only). Port required
      --help                Print this help message.
      --log_level string    Log level (default "notice")
      --moniker string      Node Name (default "anonymous")
      --node_laddr string   Node listen address. (0.0.0.0:0 means any interface, anyport) (default "tcp://0.0.0.0:46656")
      --pex                 Enable Peer-Exchange (dev feature)
      --proxy_app string    Proxy app address, or 'nilapp' or 'dummy' for local testing. (default "tcp://127.0.0.1:46658")
      --rpc_laddr string    RPC listen address. Port required (default "tcp://0.0.0.0:46657")
      --seeds string        Comma delimited host:port seed nodes
      --skip_upnp           Skip UPNP configuration
```

I believe most of these belong to `node`. We should probably move to something like https://github.com/spf13/cobra

### DIFF
--- a/cmd/tendermint/main.go
+++ b/cmd/tendermint/main.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 
-	. "github.com/tendermint/go-common"
 	cfg "github.com/tendermint/go-config"
 	"github.com/tendermint/go-logger"
 	tmcfg "github.com/tendermint/tendermint/config/tendermint"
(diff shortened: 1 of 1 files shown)


ebuchman:
Thanks anton. I've assigned moving to spf13/cobra or urfave/cli to @rigelrozanski

PR #383, opened 2017-01-30
Rename docker image to tendermint

melekes:
The main reason is people usually expect docker image to have the same name as the repo. Plus, `tendermint` is cleaner than `tmbase`.

`tmbase` would make sense if we had multiple docker images within `tendermint`, but we don't.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,18 +20,18 @@ build_race:
 dist:
 	@BUILD_TAGS='$(BUILD_TAGS)' sh -c "'$(CURDIR)/scripts/dist.sh'"
 
-test: build
+test:
 	@echo "--> Running go test"
 	@go test $(PACKAGES)
 
-test_race: build
+test_race:
 	@echo "--> Running go test --race"
 	@go test -race $(PACKAGES)
 
 test_integrations:
 	@bash ./test/test.sh
 
-test100: build
+test100:
 	@for i in {1..100}; do make test; done
 
 draw_deps:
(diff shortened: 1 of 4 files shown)


tomtruitt:
Tmbase is a specific container deployed by mintnet. One of 3. Does that container rely on this image?

melekes:
> Does that container rely on this image?

yes. if this PR is approved, I will make another PR for mintnet

tomtruitt:
`tmbase` would make sense if we had multiple docker images within `tendermint`, but we don't.

Right but we do in `mintnet`

melekes:
> Right but we do in mintnet

where? all I see is mintnet using the same image for an app, TM and data containers. Don't think we should keep doing that. Take a look at https://github.com/tendermint/mintnet-kubernetes/blob/master/examples/basecoin/basecoin.yaml#L126. Here both app and data containers are using `golang:latest` image.

Makefile
```diff
@@ -20,18 +20,18 @@ build_race:
 dist:
 	@BUILD_TAGS='$(BUILD_TAGS)' sh -c "'$(CURDIR)/scripts/dist.sh'"
 
-test: build
+test:
```
ebuchman: Why remove this? Is there anything `test` won't catch that `build` will?

melekes: If I just want to test, why do I have to build?

ebuchman: fair :)

ebuchman:
You're probably right @melekes . The reason it was `tmbase` was because we abuse docker in `mintnet` ... we don't build tendermint into the image, we copy in a run script that fetches it from github and builds at run time to allow quicker dev cycles and avoid pushing/pulling docker images.

That said, there's no reason we cant pick the latest version of tendermint to be just built into the image, and still allow devs to pull from github at run time if they like.

But also as we improve the non-docker deploy strategy, we can have quick dev cycles there, and let the docker strategy use docker properly. Great.

PR #389, opened 2017-02-02
Tiny docker image

faddat:
Golang lends itself really well to a setup where you have a little tiny golang executable in a container with no OS.  Here it is.

### DIFF
--- a/DOCKER/Dockerfile
+++ b/DOCKER/Dockerfile
@@ -1,54 +1,5 @@
-# Pull base image.
-FROM golang:1.6
-
-ENV USER tmuser
-ENV DATA_ROOT /data/tendermint
-
-# Set user right away for determinism
-RUN groupadd -r $USER \
-  && useradd -r -s /bin/false -g $USER $USER
-
-# Create home directory for USER
-# Needed for nodejs/nom
-RUN mkdir -p /home/$USER \
-  && chown -R $USER:$USER /home/$USER
-
-# Create directory for persistence and give our user ownership
-RUN mkdir -p $DATA_ROOT \
-  && chown -R $USER:$USER $DATA_ROOT
-
-# Set the env variables to non-interactive
-ENV DEBIAN_FRONTEND noninteractive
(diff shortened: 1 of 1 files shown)


melekes:
I was thinking of moving to something like alpine once we have binaries.

melekes:
Also, "Please follow standard github best practices: fork the repo, branch from the tip of develop, make some commits, and submit a pull request to develop."

https://tendermint.com/docs/guides/contributing

We should probably move this to `CONTRIBUTING.md`

faddat:
@merkeles what do you mean by "once we have binaries"-- something that would download the binary and then build an image from that?  

and in the end, why have an OS at all?  Am I missing something?

melekes:
https://github.com/hashicorp/docker-consul/blob/master/0.X/Dockerfile#L28

maybe I don't know something. Doesn't executable need a runtime (even a tiny core)?

melekes:
https://tendermint.com/intro/getting-started/download

tomtruitt:
Are those binaries?

faddat:
Those are binaries-- well worth changing the script I submitted to use.  

As for your question, here's my best explaination:  

docker runs everything on your host kernel.  So if you have a docker image with just an executable, that works just fine.  The difference being hypervisor (virtual machine) vs container (running from system kernel).

tomtruitt:
The point of containerized environments is not needing to worry about the host environment right?

faddat:
:confused: 

It is still entirely abstracted from the host environment.  It doesn't need OS bits, or even an environment because go binaries are executable.  So in the end they just run in their little box, happy as a clam.

melekes:
very interesting topic! haven't seen anybody creating docker images from scratch. Will research this on the weekends. what about addgroup/user functionality? doesn't we need OS to provide that?

melekes:
> regarding alpine
> Yes to alpine.
> No to what I was suggesting, ti introduces edge cases and their nastiness

Could you elaborate on that? Or throw some links?

Issue #387, opened 2017-01-31
CI: add docker logs to artifacts when a test fails


ebuchman:
eg. fluke failure here https://github.com/tendermint/tendermint/pull/386 that we can't investigate

ebuchman:
Also how to get outputs from `go test` failures ... https://circleci.com/gh/tendermint/tendermint/606

melekes:
Added `/var/log/upstart/docker.log` to artifacts. But this error:

```
Error deleting container: Error response from daemon: Driver btrfs failed to remove root filesystem 50cc4e21460626294e81de10f5b38541023d128eaac2320a092fe3ded7ca017c: Failed to destroy btrfs snapshot /var/lib/docker/btrfs/subvolumes for ce9458a9638f989ea73fdb4da6399c936a65598cc13013b2ff00332968c67aa2: operation not permitted
```

 is present in all builds and do not have any impact on tests. [Upgrading docker](https://github.com/docker/docker/issues/22547) seems to be the solution, but we can't do it because of the CircleCI.

Anyway, having docker logs is a good idea.

melekes:
> Also how to get outputs from go test failures ... https://circleci.com/gh/tendermint/tendermint/606

how did you guess that it was `go test` failure?

melekes:
Ok, I think I've addressed this issue https://github.com/tendermint/tendermint/issues/387#issuecomment-276567484 in #394

melekes:
As for the https://github.com/tendermint/tendermint/issues/387#issue-204394209, how do you feel about printing a response from `curl ../status` commands?

melekes:
Ahh, it won't tell us anything... we need TM logs

ebuchman:
> how did you guess that it was go test failure?

It timed out in the middle of the consensus package test, which I've spent inordinate hours debugging in the past. Thought I was done but alas, guess not.


> Error deleting container: Error response from daemon: Driver btrfs failed to remove root filesystem 50cc4e21460626294e81de10f5b38541023d128eaac2320a092fe3ded7ca017c: Failed to destroy btrfs snapshot /var/lib/docker/btrfs/subvolumes for ce9458a9638f989ea73fdb4da6399c936a65598cc13013b2ff00332968c67aa2: operation not permitted

Yeh this is one of those annoying CircleCI things but fortunately it doesn't really effect the tests (unless we try to remove docker containers in a shell script ...)

[...]

PR #394, opened 2017-02-03
Improve CircleCI setup (WIP)

melekes:
Refs #387

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,11 +22,11 @@ dist:
 
 test:
 	@echo "--> Running go test"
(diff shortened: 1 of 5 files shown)


ebuchman:
Cool. So does `/var/log/upstart/docker.log` contain the output from all the docker containers that ran? I would have though it's just for the daemon

melekes:
No :) it's just for the daemon.. see https://github.com/tendermint/tendermint/issues/387#issuecomment-278012372

PR #402, opened 2017-02-14
collect and add docker logs to CircleCI artifacts (Refs #387)

melekes:
How: 1) we start syslog docker container 2) all other containers use
syslog logging driver to ship their logs to that container

Refs #387


melekes:
https://674-19799491-gh.circle-artifacts.com/0/tmp/circle-artifacts.4ay6P0n/docker_logs.txt

PR #404, opened 2017-02-14
dockerfile for develop branch

melekes:


DOCKER/Dockerfile.develop
```diff
@@ -0,0 +1,35 @@
...
+    adduser -S -G tmuser tmuser
+
+RUN mkdir -p $DATA_ROOT && \
+    chown -R tmuser:tmuser $DATA_ROOT
+
+RUN apk add --no-cache bash curl jq
+
+ENV GOPATH /go
+ENV PATH "$PATH:/go/bin"
+RUN mkdir -p /go/src/github.com/tendermint/tendermint && \
```
ebuchman: what do you think about doing something like https://github.com/tendermint/tendermint/blob/master/test/docker/Dockerfile#L9 here also to take advantage of caching ?

melekes: It's a compromise between image size and caching. I've designed this so our k8s cluster starts as fast as possible (the smaller the image size, the better). On the other hand, is caching with faster build times, but resulting image size ~ 200Mb (versus 70Mb)

ebuchman: Ok great, thanks for explanation

PR #405, opened 2017-02-14
Rpc commit

ebuchman:
- BlockMeta now contains BlockID instead of the Hash and PartSetHeader
- expose new `/commit` that returns the header and commit for a block at height H

Question: should the commit also return the BlockID (ie. the block hash and partsetheader - the former it can get from the header, the latter it can only get from the next block, or by fetching the whole block and building the part set) ?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ dist:
 
 test:
 	@echo "--> Running go test"
-	@go test -v $(PACKAGES)
+	@go test $(PACKAGES)
 
 test_race:
 	@echo "--> Running go test --race"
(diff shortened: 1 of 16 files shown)


ethanfrey:
Cool.  Nice job with the commits from current block.  Maybe we could send this along with the header and a flag as I suggest in #407 

```Go
type BlockCommit struct {
	*Header    `json:"header"`
	*Commit `json:"commit"`
        Canonical bool `json:"canonical"`
}
```

What do you think?  It would make me happy ;)

ebuchman:
Isn't that what it already does? What am I missing?

ethanfrey:
I just looked at the diff for the "added /commit" commit.  I didn't see you added the Header in a later field.

yes, I like it.  and if it makes in by 0.9 even more awesome!

Issue #416, opened 2017-02-21
Better client interface (and mockable)


ethanfrey:
Based on the `rpc/tests/client_test.go`, I extracted a nice interface to make all this rpc calling much easier and type safe for the clients.  I am using this in the light-client repo, but as @ebuchman mentioned, it really belongs in tendermint.

I will extract this code, clean it up a bit, and prepare a mock version for testing as well. If it is merged in develop, I will remove the duplicate code from light-client (which originally came from my personal repo tenderize...)

@ebuchman let me know if this is something that can make it in 0.9?

ebuchman:
Potentially, but no need to rush. Definitely for 0.9.1

ethanfrey:
Okay, I was gonna do it, so I could pull out the code for light-client and make it more accessible.  After talking with piotr today, I realized this really could be useful for developers.

But cool, no rush, it doesn't change any APIs, and 0.9.1 is fine with me.

ebuchman:
Ok - you're right. Do it. It'll be good for 0.9 to encourage more testing.

Issue #285, opened 2016-10-01
Panic in tendermint core when subscribing to tx event over ws rpc


ethanfrey:
How to reproduce:

```
go get github.com/ethanfrey/tenderize
go test github.com/ethanfrey/tenderize/client/tests
```

Causes a nice panic.  If you want the tendermint logs (I turned them off, too much noise for me), check out `main_test.go`.

After digging through the source, I think I figured out the issue, not sure I should be bold enough to change core code (even on a fork...). The wrong type is returned when new transactions are appended.

`tendermint/tendermint/state/execution.go:78` is guaranteed to fire `*tmsp.Response_AppendTx` cast to `*tmsp.Response`

```
	proxyCb := func(req *tmsp.Request, res *tmsp.Response) {
		switch r := res.Value.(type) {
		case *tmsp.Response_AppendTx:
...
			eventCache.FireEvent(types.EventStringTx(req.GetAppendTx().Tx), res)
```

This behavior is expected and used in the `tendermint/tendermint/rpc/core/mempool.go:56-58` (`BroadcastTxCommit`)

```
	eventSwitch.AddListenerForEvent("rpc", types.EventStringTx(tx), func(data events.EventData) {
		appendTxResCh <- data.(*tmsp.Response)
	})
```

However, when we subscribe to this event, it expects a completely different type, namely `EventDataTx`, which we see here:
`tendermint/tendermint/rpc/core/events.go:17-18`

```
		tmResult := ctypes.TMResult(&ctypes.ResultEvent{event, types.TMEventData(msg)})
		wsCtx.TryWriteRPCResponse(rpctypes.NewRPCResponse(wsCtx.Request.ID+"#event", &tmResult, ""))
```

It all works good until it tried to serialize 😞 

I'm not sure how to transform the structs properly, or if that will break some other code...

ethanfrey:
In case you want the logs, the ws subscriber got this struct:

`&types.Response{Value:(*types.Response_AppendTx)(0xc8203d24c8)}`

And this is the panic that results: 

```
panic: Paniced on a Sanity Check: Unexpected pointer type *types.Response for registered interface TMEventData. Was it registered as a value receiver rather than as a pointer receiver?

goroutine 9 [running]:
panic(0x631de0, 0xc8204b0bb0)
	/usr/local/go/src/runtime/panic.go:481 +0x3e6
github.com/tendermint/go-common.PanicSanity(0x744460, 0xc8204b0b70)
[...]

PR #418, opened 2017-02-22
Add flexible rpc client to tendermint

ethanfrey:
This code to call the rpc was in light-client, but I abstracted it to make it more testable, and provided mocks, so we can actually can simulate the server. 

A bit of clean up of the existing code to test the rpc interface, and then a lot of code in `rpc/client/*`

I would like to get this in by 0.9.0 and then use it as the basis for light-client (so I can clean up that code as well).

Please let me know if I should change anything.  Mocks are not finished, especially as faking sigs is hard, but they are halfway done and already provide a boon for testing (`client/helpers_test.go`).


ethanfrey:
Forgot to mention, this is in response to issue #416

melekes:
Wow! So much code!

rpc/client/local/app_test.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"math/rand"
+
+	meapp "github.com/tendermint/merkleeyes/app"
+
+	wire "github.com/tendermint/go-wire"
+)
+
+// MakeTxKV returns a text transaction, allong with expected key, value pair
+func MakeTxKV() ([]byte, []byte, []byte) {
```
melekes: not a fan of go-common, but code duplication here hurts me a bit.

rpc/client/local/client.go
```diff
@@ -0,0 +1,139 @@
...
+* Communication between an ABCI app and tendermin core when they
+are compiled in process.
+
+For real clients, you probably want the "http" package.  For more
+powerful control during testing, you probably want the "mock" package.
+*/
+package local
+
+import (
+	nm "github.com/tendermint/tendermint/node"
```
melekes: nm - node module?

rpc/client/local/rpc_test.go
```diff
@@ -0,0 +1,183 @@
+package local_test
```
melekes: also, this testing suite is duplicated in 2 places. I understand we want to test both implementations. Not sure what is the best way to deal with this.

rpc/client/mock/abci.go
```diff
@@ -0,0 +1,186 @@
...
+)
+
+// ABCIApp will send all abci related request to the named app,
+// so you can test app behavior from a client without needing
+// an entire tendermint node
+type ABCIApp struct {
+	App abci.Application
+}
+
+func (a ABCIApp) _assertABCIClient() client.ABCIClient {
```
melekes: Why do we need such methods (`_assertIsClient`, `_assertABCI...`)?

rpc/client/mock/abci.go
```diff
@@ -0,0 +1,186 @@
...
+	if !c.IsOK() {
+		return &res, nil
+	}
+	d := a.App.DeliverTx(tx)
+	res.DeliverTx = &abci.ResponseDeliverTx{d.Code, d.Data, d.Log}
+	return &res, nil
+}
+
+func (a ABCIApp) BroadcastTxAsync(tx types.Tx) (*ctypes.ResultBroadcastTx, error) {
+	d := a.App.DeliverTx(tx)
```
melekes: I think it should be `CheckTx` here, no? and empty result in this case

```
// Returns right away, with no response
func BroadcastTxAsync(tx types.Tx) (*ctypes.ResultBroadcastTx, error) {
	err := mempool.CheckTx(tx, nil)
	if err != nil {
		return nil, fmt.Errorf("Error broadcasting transaction: %v", err)
	}
	return &ctypes.ResultBroadcastTx{}, nil
}
```

rpc/client/mock/abci.go
```diff
@@ -0,0 +1,186 @@
...
+	return &ctypes.ResultBroadcastTx{d.Code, d.Data, d.Log}, nil
+}
+
+func (a ABCIApp) BroadcastTxSync(tx types.Tx) (*ctypes.ResultBroadcastTx, error) {
+	d := a.App.DeliverTx(tx)
```
melekes: same here (`CheckTx`)

rpc/test/helpers.go
```diff
@@ -1,98 +1,135 @@
...
-	requestAddr = rpcAddr
-	websocketAddr = rpcAddr
-	websocketEndpoint = "/websocket"
+const tmLogLevel = "error"
 
-	clientURI = client.NewClientURI(requestAddr)
-	clientJSON = client.NewClientJSONRPC(requestAddr)
-	clientGRPC = core_grpc.StartGRPCClient(grpcAddr)
+// f**ing long, but unique for each test
+func makePathname() string {
```
melekes: https://golang.org/pkg/io/ioutil/#TempDir ?

melekes:
Nice job! Learned a lot even by reading the code!

rpc/client/local/app_test.go
ethanfrey: Didn't know this was in go-common, will take a look and fix

rpc/client/mock/abci.go
ethanfrey: Not needed by nice helpers. Talked with Bucky about them, he liked the idea too.

Basically, when you create Struct explicitly to fulfill an interface, you can assert so if the methods don't match, you get a compile error in the module itself, not when someone else tried to use it.

Matt had this problem with endblock changing SIG and then quietly being ignored in basecoin without any compiler warnings.

rpc/client/mock/abci.go
ethanfrey: Yes, will fix. Good eyes.

rpc/client/local/rpc_test.go
ethanfrey: Yes, this hurts me a bit too. If you have a clean solution so I can convert this to a table-driven test, let me know.

Also, for code coverage, it only really measures the coverage from tests in the same package.

I'll think of a way to do code reuse. Let me know if it is too ugly.

rpc/client/local/app_test.go
melekes: No, I meant that we should probably move this into one package and reuse rather than duplicating the code.

rpc/client/http/rpc_test.go
```diff
@@ -0,0 +1,211 @@
...
+func TestSubscriptions(t *testing.T) {
+	require := require.New(t)
+	c := GetClient()
+	err := c.StartWebsocket()
+	require.Nil(err)
+	defer c.StopWebsocket()
+
+	// subscribe to a transaction event
+	_, _, tx := MakeTxKV()
+	// this causes a panic in tendermint core!!!
```
ethanfrey: old comment... I think I pushed a fix to tendermint in October, before I got hired even.

this client code was born out of a personal project.  But yeah, due to some typos and crazy interface magic, subscribing to a tx event would cause tendermint to panic:
https://github.com/tendermint/tendermint/issues/285

Removing this comment now.

rpc/client/local/app_test.go
ethanfrey: that code belongs in dummy actually... will put it there.

ethanfrey: RandAsciiBytes is almost the same as common.RandStr, will use RandStr instead

rpc/client/local/client.go
ethanfrey: yup, i use node too much as a local variable and didn't want naming conflict

rpc/client/mock/abci.go
ethanfrey: We want this action to actually happen in the mock, but I change it to return the results of CheckTx, and trigger DeliverTx in a go-routine on success, so we can actually simulate this race in test cases if we want.

ethanfrey:
@melekes I pulled local and http subpackages into client, renamed all structs, rewrote the rpc_tests to be table driven and accept any number of client.Client implementations.

Oh, and rebased everything on develop (which is why all commits are "new")

Please take a look at 7fe9ff5141457e3cc05ebc87bfb7f561c28e52c8 and let me know what you think.

ethanfrey:
@ebuchman I did lots of cleanup from anton's comments, please take a look.  I'd like to pull out the code in light-client.

melekes:
LGTM

rpc/client/mock/abci.go
ebuchman: we don't need this anymore. It was only an issue because there was an embedded interface type in abci.Application (the BlockchainAware bit). Now that has been moved into the abci.Application, so as long as that interface is being passed around (it is, everywhere), then the compiler already type checks

ethanfrey: In this case, I want to make sure my mock structs properly implement the client interface. 

the abci app was just to explain to anton why the idea came up.

rpc/client/rpc_test.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	res := <-r
+	checkData(res, ctypes.ResultTypeSubscribe)
+
+	// read one event, must be success
+	select {
+	case res := <-r:
+		checkData(res, ctypes.ResultTypeEvent)
+		// this is good.. let's get the data... ugh...
```
ebuchman: hm

ethanfrey:
I'll clean this up tomorrow. Thanks for the review.

Yeah, all the event stuff issue stuff is not really supported well.  I added what I had to the client.HTTP only and the user has to do all the casting.  I want to figure out a good way of exposing events, getting feedback from people trying to use it, and then make a nice event system for 0.10 and add this to the Client interface.

Note I had trouble parsing the event that came out - some json blob... maybe we can overhaul the json format for 0.10 as well...

ethanfrey:
Addressed some changes. Docs missing and thinking about how/if to expose websockets?  Maybe just expose an EventSwitch on the client? Does it make sense to work on that today?  Or rather targeting 0.9.1?

rpc/client/httpclient.go
```diff
@@ -0,0 +1,349 @@
...
+	ws       *rpcclient.WSClient
+
+	// used for signaling the goroutine that feeds ws -> EventSwitch
+	quit chan bool
+	done chan bool
+
+	// used to maintain counts of actively listened events
+	// so we can properly subscribe/unsubscribe
+	// FIXME: thread-safety???
+	// FIXME: reuse code from go-events???
```
ebuchman: hmmm - why are we replicating all this code?

rpc/client/helpers.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// WaitForOneEvent subscribes to a websocket event for the given
+// event time and returns upon receiving it one time, or
+// when the timeout duration has expired.
+//
+// This handles subscribing and unsubscribing under the hood
+func WaitForOneEvent(evsw types.EventSwitch,
+	evtTyp string, timeout time.Duration) (types.TMEventData, error) {
+	listener := cmn.RandStr(12)
```
ethanfrey: Which part is a bad idea?  Waiting?  subscribing for the events?

rpc/client/helpers.go
```diff
@@ -0,0 +1,89 @@
...
+
+	evts, quit := make(chan events.EventData, 10), make(chan bool, 1)
+	// start timeout count-down
+	go func() {
+		time.Sleep(1 * time.Second)
```
ethanfrey: huh?  comment on the name of the channel?
I have a timeout as time.Duration, so I picked another name.  Don't mind changing names, if you have an idea.

rpc/client/helpers.go
ebuchman: sorry  I just meant using a random string for the listener. I'm just thinking of the day there might be many calls to WaitForOneEvent and no one will know which listener is which

rpc/client/helpers.go
```diff
@@ -0,0 +1,89 @@
...
+		quit <- true
+	}()
+
+	// register for the next event of this type
+	evsw.AddListenerForEvent(listener, evtTyp, func(data events.EventData) {
+		evts <- data
+	})
+	// make sure to unregister after the test is over
+	// TODO: why doesn't the other call work???
+	// defer evsw.RemoveListenerForEvent(listener, evtTyp)
```
ethanfrey: On client.Local, if I just call `RemoveListener`, it doesn't properly deregister.  And in the end the listener fills up the channel and the node blocks and tests hang.

It seemed to deep to debug (Node.EventSwitch()), so I just put a TODO.
In any case, the fix is out of scope of this pr.

rpc/client/httpclient.go
ethanfrey: Because it is impossible to reuse code from go-event.  All private (package local).  I tried... then I did it.  As above, I am down to improve, but changes to go-event are out of scope.  Otherwise this PR will never get finished.

rpc/client/helpers.go
ebuchman: shouldn't `time.Sleep(1 * time.Second)` be `time.Sleep(timeout * time.Second)`

rpc/client/helpers.go
ethanfrey: the "listener" is just for that one call-back.  I can let the caller decide, but it will probably be hard-coded various places.  and duplicate names is worse, then second might not get the message.

PR #419, opened 2017-02-23
added use of Cobra CLI

### LINKED ISSUE #312: use a cli package
urfave/cli or spf13/cobra ?!

I'm leaning towards cobra.

rigelrozanski:
closes #312 

replaced internal CLI system with Cobra


cmd/tendermint/commands/replay.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+
+	"github.com/tendermint/tendermint/consensus"
+
+	"github.com/spf13/cobra"
+)
+
+var replayCmd = &cobra.Command{
+	Use:   "replay",
```
melekes: `replay [walfile]`

cmd/tendermint/commands/replay.go
```diff
@@ -0,0 +1,40 @@
...
+		if len(args) > 1 {
+			consensus.RunReplayFile(config, args[1], false)
+		} else {
+			fmt.Println("replay requires an argument (walfile)")
+		}
+	},
+}
+
+var replayConsoleCmd = &cobra.Command{
+	Use:   "replay_console",
```
melekes: `replay_console [walfile]`

cmd/tendermint/commands/root.go
```diff
@@ -0,0 +1,76 @@
...
+func init() {
+
+	// Get configuration
+	config = tmcfg.GetConfig("")
+
+	/////////////////////
+	// parse flags
+
+	// configuration options
+	RootCmd.PersistentFlags().StringVar(&moniker, "moniker", config.GetString("moniker"), "Node Name")
```
melekes: Correct me if I am wrong, but all these flags belong to `runNodeCmd` and should be moved there. Except maybe `log_level`. Can we do that?

cmd/tendermint/commands/run_node.go
```diff
@@ -0,0 +1,111 @@
...
+
+var runNodeCmd = &cobra.Command{
+	Use:   "node",
+	Short: "Run the tendermint node",
+	Run:   runNode,
+}
+
+//flags
+var (
+	printHelp     bool
```
melekes: unused var?

rigelrozanski: yes, thanks

rigelrozanski:
Add use of Viper to replace config as a part of this PR?

ethanfrey:
I think the go-config -> viper should be a separate pull request.

Cobra alone cleans up the cli.  Changing the config file, doesn't just involve changing the cli, it means changing every call to `config.GetString("foo")` scattered around the code.  And the test config setup, in tendermint and other packages.  And possibly the basecoin cli. I know I set the config directly in some of my test cases...

It's a big piece of work.  Maybe get cobra/viper in basecoin first, then go for viper in tendermint core.

ethanfrey:
P.S. It looks like you need to rebase on develop again...

rigelrozanski:
Thanks Frey, I will make a separate PR off of cli_cobra branch

glide.yaml
```diff
@@ -1,4 +1,4 @@
-package: github.com/tendermint/tendermint
+eackage: github.com/tendermint/tendermint
```
melekes: typo

PR #428, opened 2017-03-06
Provide means to configure a logfile

srmo:
*THIS REQUIRES THE PR https://github.com/tendermint/go-logger/pull/1 FOR go-logger*

- make ResetWithLogLevel public
- new config parameter "log_dir" in config.toml
- if not-set, logging will only happen to stdout
- if set, logging will be written to a file named "node.log", located in the given dir
- logging will _always_ happen on stdout, whether a log_dir is configured or not
- tendermint will not start if the given log_dir does not exist

### DIFF
--- a/cmd/tendermint/main.go
+++ b/cmd/tendermint/main.go
@@ -39,7 +39,7 @@ Commands:
 	parseFlags(config, args[1:]) // Command line overrides
 
 	// set the log level
-	logger.SetLogLevel(config.GetString("log_level"))
+	logger.ResetWithLogLevel(config.GetString("log_level"), config.GetString("log_dir"))
 
 	switch args[0] {
 	case "node":
--- a/config/tendermint/config.go
+++ b/config/tendermint/config.go
@@ -75,6 +75,7 @@ func GetConfig(rootDir string) cfg.Config {
 	mapConfig.SetDefault("cs_wal_file", rootDir+"/data/cs.wal/wal")
 	mapConfig.SetDefault("cs_wal_light", false)
 	mapConfig.SetDefault("filter_peers", false)
+	mapConfig.SetDefault("log_dir", "") // don't use logfile output by default
 
 	mapConfig.SetDefault("block_size", 10000)      // max number of txs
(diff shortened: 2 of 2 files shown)


melekes:
Thank you for this, but my personal opinion is that routing (plus rotating, compressing, ...) of the output stream should not be a concern of Tendermint. Check out [12-factor app ](https://12factor.net/logs) (PS. I'd like to see TM adopt as many principles of 12-factor app as possible). What if the user wants to send them to syslog on a remote host? What if the user wants to send them to ELK? Should all of this be configurable? Of course, no. I think you will agree with me here.

I know that in Java it may considered as a good practice for the app to have a broad range of responsibilities. But it does not necessary mean we should adopt this in Tendermint.

You can configure your process manager to log into a file: `>> logfile 2>&1` Example: https://github.com/tendermint/tools/blob/master/ansible-tendermint/templates/tendermint.conf.j2#L9

If you are running TM inside Docker you can use https://github.com/gliderlabs/logspout

ethanfrey:
I think the Go philosophy and tooling is for 12-factor apps and redirecting them.  It took me time to adjust for my last job, but it is not too hard, just changing the running scripts.

What I think we do need is a better logging controls (setting levels via cli, structured logging), and some example ops code showing how to save to a file, save to syslog, ship to ELK, greylog, whatever....

For the time being:

```sh
tendermint start >> $logfile 2>&1
```

would work as well as your PR.

melekes:
> What I think we do need is a better logging controls (setting levels via cli, structured logging)

yes, we have an issue for that #358 

> some example ops code showing how to save to a file, save to syslog, ship to ELK, greylog, whatever....

don't think we ought to do that unless these tools need some custom setup for Tendermint. I mean, there are a lot of tutorials on the Internet.

srmo:
Ok, yes. I'm used to have my app configure how logfiles are structure/placed and not rely on (maybe) platform dependent redirect pipes. Usually, I configure logging pipes like log io to look in a certain place for logfiles, the place I configured via the app. But I'm perfectly fine with your approach. It's a clear statement on what is to be considered during deployment.

So, feel free to close this one and the related PR

PR #432, opened 2017-03-13
go-rpc updates

melekes:
jsonrpc now uses the same interface as http (key-value pairs)

Refs https://github.com/tendermint/go-rpc/pull/9

rename rpc clients

See https://github.com/tendermint/go-rpc/commit/e6c083f589d93d5cada7c5e9fff1c3c05873a167

update deps

### DIFF
--- a/benchmarks/simu/counter.go
+++ b/benchmarks/simu/counter.go
@@ -2,15 +2,14 @@ package main
 
 import (
 	"encoding/binary"
-	"time"
-	//"encoding/hex"
 	"fmt"
+	"time"
 
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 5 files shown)


rigelrozanski:
p.s. this is a breaking change for basecoin too... maybe update basecoin and PR as well?

melekes:
Done https://github.com/tendermint/basecoin/pull/53

ebuchman:
implemented here https://github.com/tendermint/tendermint/commit/e4e17a2c956cfaebbe9e9d1565a94ff20a23003f

Sorry for letting this linger, resulting in dup.

PR #437, opened 2017-03-27
rename TMROOT to TMHOME

melekes:


ebuchman:
Thanks Anton. Can we make this backwards compatible for now? Ie. if TMROOT is set but TMHOME is not, use TMROOT, but log a deprecation warning. Then we can merge without breaking :)

melekes:
yeah, sure. didn't think about backward compatibility

Issue #436, opened 2017-03-27
Interrupt catch-up can never be resumed


srmo:
BUG REPORT

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):
0.9.0

**ABCI app**
Proprietary Java app

**OS**
Linux 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt25-2+deb8u3 (2016-07-02) x86_64 GNU/Linux

**Environment**
* 4 nodes on physically dislocated VMs (in different networks around the globe) (N1-N4)
* all 4 nodes are validators
* ABCI App persists height/hash after each commit for ABCI Handshake
* all nodes use the same node N1 as seed
* all nodes have `fastsync = true`

**What happened**:
One of the nodes stopped. Reason unknown. Last messages before stopping
```WARN[03-22|15:15:20] Need to set a buffer and log.Warn() here for sanity. module=consensus startTime=2017-03-22T15:15:21-0400 now=2017-03-22T15:15:20-0400
NOTE[03-22|15:15:20] enterNewRound(481148/1). Current: 481148/0/RoundStepNewHeight module=consensus
NOTE[03-22|15:15:20] enterPrecommit: No +2/3 prevotes during enterPrecommit. Precommitting nil. module=consensus
WARN[03-22|15:15:20] Attempt to finalize failed. We don't have the commit block. module=consensus height=481148 proposal-block= commit-block=2104096F72AB8DB6C252D6178F677B4026707144
WARN[03-22|15:15:20] Connection failed @ recvRoutine (reading byte) module=p2p conn=MConn{83.169.3.162:46656} error=EOF
NOTE[03-22|15:15:20] Stopping peer for error                  module=p2p peer="Peer{MConn{83.169.3.162:46656} 50D4E2821096 out}" error=EOF
NOTE[03-22|15:15:20] Stopping gossipVotesRoutine for Peer{MConn{83.169.3.162:46656} 50D4E2821096 out}. module=consensus peer="Peer{MConn{83.169.3.162:46656} 50D4E2821096 out}"
NOTE[03-22|15:15:20] Stopping gossipDataRoutine for Peer{MConn{83.169.3.162:46656} 50D4E2821096 out}. module=consensus peer="Peer{MConn{83.169.3.162:46656} 50D4E2821096 out}"
[...]

PR #438, opened 2017-03-27
consensus: timeout on replayLastBlock

ebuchman:
Fix https://github.com/tendermint/tendermint/issues/436

Handshake replay logic did not account for need to catch up by fast sync!

### DIFF
--- a/config/tendermint/config.go
+++ b/config/tendermint/config.go
@@ -79,6 +79,8 @@ func GetConfig(rootDir string) cfg.Config {
 	mapConfig.SetDefault("block_size", 10000)      // max number of txs
 	mapConfig.SetDefault("block_part_size", 65536) // part size 64K
 	mapConfig.SetDefault("disable_data_hash", false)
+
+	// all timeouts are in ms
 	mapConfig.SetDefault("timeout_handshake", 10000)
 	mapConfig.SetDefault("timeout_propose", 3000)
 	mapConfig.SetDefault("timeout_propose_delta", 500)
@@ -87,6 +89,7 @@ func GetConfig(rootDir string) cfg.Config {
 	mapConfig.SetDefault("timeout_precommit", 1000)
 	mapConfig.SetDefault("timeout_precommit_delta", 500)
 	mapConfig.SetDefault("timeout_commit", 1000)
+
 	// make progress asap (no `timeout_commit`) on full precommit votes
 	mapConfig.SetDefault("skip_timeout_commit", false)
 	mapConfig.SetDefault("mempool_recheck", true)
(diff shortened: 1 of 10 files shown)


config/tendermint/config.go
```diff
@@ -79,6 +79,7 @@ func GetConfig(rootDir string) cfg.Config {
 	mapConfig.SetDefault("block_size", 10000)      // max number of txs
 	mapConfig.SetDefault("block_part_size", 65536) // part size 64K
 	mapConfig.SetDefault("disable_data_hash", false)
+	mapConfig.SetDefault("timeout_handshake", 10000)
```
melekes: Does it make sense to specify units of time: either as a part of the name (`handshake_timeout_ms`) or comment `// ms` ?

consensus/replay.go
```diff
@@ -320,6 +326,7 @@ func (h *Handshaker) replayBlocks(proxyApp proxy.AppConns, appBlockHeight, store
 func (h *Handshaker) replayLastBlock(proxyApp proxy.AppConnConsensus) ([]byte, error) {
 	mempool := types.MockMempool{}
 	cs := NewConsensusState(h.config, h.state, proxyApp, h.store, mempool)
+	defer cs.Stop()
```
melekes: is there a particular reason to place `defer cs.Stop` before `cs.Start`? I thought the common pattern in Golang is `smth.Start(); defer smth.Stop()`

config/tendermint/config.go
ebuchman: Yes it makes sense. For now the standard is everything in `ms`. I'll add a comment to that effect. `_ms` might be nice if we ever want another denomination, but will be breaking.

consensus/replay.go
ebuchman: Indeed. Though I've seen it before as well. I think in this case, where `Start` may start a go-routine and then return an error, it makes sense to put the `defer Stop` before, otherwise we would need to have it twice - once in the error handling and once outside it.

ethanfrey: Smells fishy to me...

If I look at stdlib, for example `file, err := os.Open(path)`.  If `err != nil` then I can just do error handling.  If `err == nil`, then the file is created, I can do something with it, and must remember to close it (or `defer file.Close()` in the line after error checking).  In this case, if I call `file.Close()` when an error is returned, I may trigger a panic.

Why not have our services do the same?  If they return an error, they failed, and *leave no state that needs cleanup* if they return success, then the caller must be responsible for stopping them and doing cleanup.  Is there a reason such assumptions would not work?

ebuchman: Ok, you're right. I'll do the cleanup within the Start function if there's an error. Thanks.

Issue #431, opened 2017-03-12
Explicit --data flag instead of `~/.tendermint`


ebuchman:
jae:
```
really ~/.tendermint is just a bad idea
it used to make sense when we wanted tendermint to be a blockchain.
now it doesn’t.
```

ebuchman:
TMROOT -> TMHOME

melekes:
[patch.zip](https://github.com/tendermint/tendermint/files/872537/patch.zip)

Patch for #419 which adds `data` flag

ethanfrey:
@ebuchman I don't know the decision. Is this fixed on develop with the - - root and - - home and TMROOT and TMHOME variables? 

Or is this one to make them required?

adrianbrink:
We now have `--home`.

PR #439, opened 2017-03-29
explicit --data flag instead of `~/.tendermint` (Refs #431)

melekes:


cmd/tendermint/commands/run_node.go
```diff
@@ -35,27 +35,27 @@ var (
 func init() {
 
 	// configuration options
-	runNodeCmd.Flags().StringVar(&moniker, "moniker", config.GetString("moniker"),
```
rigelrozanski: What's the logic surrounding changing all of the default flag values from dependent on the config to hard-coded?

melekes: Given that path to config is now dynamic (`--data path`; `path/config.toml`), I don't see a way to read it beforehand and display default values cause we don't know the path to config, doh.

Maybe we could create an instance of default Config in memory and get defaults from it.  How does that sound? Or maybe there is a way to solve this puzzle using Viper?

rigelrozanski: Config is a global variable within `commands`, so long as it has been initialized we should be able to retrieve values from it without knowing what path it was derived from. We can make sure that it is set beforehand by doing all the data flag processing in the `cobra.command.prerun` I believe.

rigelrozanski: Oops. Yeah I get why this a problem now. The only option I see where the defaults can be initialized from config is if we define a persistent config file location (which could be modified with a tendermint command). This way the config file would be initialized before the flags and could be read in... maybe not as nice as just using the `--data` flag. As for initializing from a default config, this seems unwieldy - the current hard coded defaults seems reasonable for the time being

melekes: Ok, thanks

ebuchman:
Closing as this will be handled by new viper work.

PR #442, opened 2017-04-10
go-config -> viper, commands: Run -> RunE

rigelrozanski:
Replace Go-Config with Viper, In conjunction with go-p2p PR https://github.com/tendermint/go-p2p/pull/21/files

Also includes updates from the branch `cli_updates` which is used for the `cli_cobra` branch of basecoin, PR https://github.com/tendermint/basecoin/pull/64

### DIFF
--- a/cmd/tendermint/commands/reset_priv_validator.go
+++ b/cmd/tendermint/commands/reset_priv_validator.go
@@ -36,17 +36,18 @@ func resetAll(cmd *cobra.Command, args []string) {
 // XXX: this is totally unsafe.
 // it's only suitable for testnets.
 func resetPrivValidator(cmd *cobra.Command, args []string) {
-	ResetPrivValidator(config, log)
+	resetPrivValidatorLocal(config, log)
 }
 
 // Exported so other CLI tools can use  it
 func ResetAll(c *viper.Viper, l log15.Logger) {
-	ResetPrivValidator(c, l)
-	os.RemoveAll(c.GetString("db_dir"))
+	resetPrivValidatorLocal(c, l)
+	dataDir := c.GetString("db_dir")
+	os.RemoveAll(dataDir)
 	l.Notice("Removed all data", "dir", dataDir)
 }
 
-func ResetPrivValidator(c *viper.Viper, l log15.Logger) {
+func resetPrivValidatorLocal(c *viper.Viper, l log15.Logger) {
 
 	// Get PrivValidator
 	var privValidator *types.PrivValidator
--- a/cmd/tendermint/commands/reset_priv_validator.go
+++ b/cmd/tendermint/commands/reset_priv_validator.go
@@ -4,8 +4,8 @@ import (
 	"os"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 
-	cfg "github.com/tendermint/go-config"
 	"github.com/tendermint/log15"
 	"github.com/tendermint/tendermint/types"
 )
@@ -40,13 +40,14 @@ func resetPrivValidator(cmd *cobra.Command, args []string) {
 }
 
 // Exported so other CLI tools can use  it
-func ResetAll(c cfg.Config, l log15.Logger) {
+func ResetAll(c *viper.Viper, l log15.Logger) {
 	ResetPrivValidator(c, l)
 	os.RemoveAll(c.GetString("db_dir"))
 	l.Notice("Removed all data", "dir", dataDir)
 }
 
-func ResetPrivValidator(c cfg.Config, l log15.Logger) {
+func ResetPrivValidator(c *viper.Viper, l log15.Logger) {
+
 	// Get PrivValidator
 	var privValidator *types.PrivValidator
 	privValidatorFile := c.GetString("priv_validator_file")
--- a/cmd/tendermint/commands/reset_priv_validator.go
+++ b/cmd/tendermint/commands/reset_priv_validator.go
@@ -43,20 +43,21 @@ func resetPrivValidator(cmd *cobra.Command, args []string) {
 func ResetAll(c cfg.Config, l log15.Logger) {
 	ResetPrivValidator(c, l)
 	os.RemoveAll(c.GetString("db_dir"))
+	l.Notice("Removed all data", "dir", dataDir)
 }
 
 func ResetPrivValidator(c cfg.Config, l log15.Logger) {
 	// Get PrivValidator
 	var privValidator *types.PrivValidator
-	privValidatorFile := config.GetString("priv_validator_file")
+	privValidatorFile := c.GetString("priv_validator_file")
 	if _, err := os.Stat(privValidatorFile); err == nil {
 		privValidator = types.LoadPrivValidator(privValidatorFile)
 		privValidator.Reset()
-		log.Notice("Reset PrivValidator", "file", privValidatorFile)
+		l.Notice("Reset PrivValidator", "file", privValidatorFile)
 	} else {
 		privValidator = types.GenPrivValidator()
 		privValidator.SetFile(privValidatorFile)
 		privValidator.Save()
-		log.Notice("Generated PrivValidator", "file", privValidatorFile)
+		l.Notice("Generated PrivValidator", "file", privValidatorFile)
 	}
 }
--- a/glide.yaml
+++ b/glide.yaml
@@ -10,6 +10,7 @@ import:
 - package: github.com/gorilla/websocket
 - package: github.com/pkg/errors
 - package: github.com/spf13/cobra
+- package: github.com/spf13/viper
 - package: github.com/stretchr/testify
   subpackages:
   - require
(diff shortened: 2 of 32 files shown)


cmd/tendermint/commands/reset_priv_validator.go
```diff
@@ -28,25 +30,38 @@ func init() {
...
+	ResetPrivValidator(c, l)
+	dataDir := c.GetString("db_dir")
+	walFile := c.GetString("cs_wal_file")
+	os.RemoveAll(dataDir)
+	os.Remove(walFile)
+	l.Notice("Removed all data", "dir", dataDir)
+	l.Notice("Removed walfile", "file", walFile)
+}
+
+func ResetPrivValidator(c *viper.Viper, l log15.Logger) {
```
melekes: do we have to export this?

cmd/tendermint/commands/reset_priv_validator.go
```diff
@@ -28,25 +30,38 @@ func init() {
...
 		privValidator = types.LoadPrivValidator(privValidatorFile)
 		privValidator.Reset()
-		log.Notice("Reset PrivValidator", "file", privValidatorFile)
+		l.Notice("Reset PrivValidator", "file", privValidatorFile)
 	} else {
 		privValidator = types.GenPrivValidator()
 		privValidator.SetFile(privValidatorFile)
 		privValidator.Save()
-		log.Notice("Generated PrivValidator", "file", privValidatorFile)
+		l.Notice("Generated PrivValidator", "file", privValidatorFile)
```
melekes: not sure I like the new naming, but I'll leave it to your discretion

glide.yaml
```diff
@@ -19,7 +17,7 @@ import:
 - package: github.com/tendermint/go-merkle
   version: develop
 - package: github.com/tendermint/go-p2p
-  version: develop
+  version: viper
```
melekes: I suppose this was used for testing?

node/node.go
```diff
@@ -119,7 +121,11 @@ func NewNode(config cfg.Config, privValidator *types.PrivValidator, clientCreato
 	consensusReactor := consensus.NewConsensusReactor(consensusState, fastSync)
 
 	// Make p2p network switch
-	sw := p2p.NewSwitch(config.GetConfig("p2p"))
```
melekes: don't think anybody was using this. looks like the config for p2p has to be in `p2p/config.toml` in order to be read.

cmd/tendermint/commands/reset_priv_validator.go
rigelrozanski: hmm we need to export `ResetAll` for being able to import this code for the basecoin CLI (https://github.com/tendermint/basecoin/pull/64/files#diff-81718560b770949c7ec1743c3485f9df) but I don't believe we need to export ResetPrivValidator, I'll switch it. Thanks

cmd/tendermint/commands/reset_priv_validator.go
rigelrozanski: The naming was changed to avoid confusion with the global variables `config` and `log` -  these are now local variables in these functions

glide.yaml
rigelrozanski: ... yes I would actually appreciate some advice as to how to deal with managing multi-repo pull requests, obviously as soon as `go-p2p` has `viper` merged into `develop` then this should be changed back, but until then it makes sense to be left on `viper` while in PR - Is there a better way of doing this?

node/node.go
rigelrozanski: hmm okay, well I rebuilt this functionality (because it doesn't exist in viper) in the few added lines right below, if there is no functional use for this code of loading in p2p from a config instead of just initializing a new config for go-p2p, then I reckon it makes sense to remove or comment out this code.

rigelrozanski:
Also @ebuchman any thoughts on why circle tests are failing here?

ebuchman:
Looks like there's a race condition:

```
==================
WARNING: DATA RACE
Write at 0x00c4201856e0 by goroutine 36:
  runtime.mapassign1()
      /usr/local/go/src/runtime/hashmap.go:442 +0x0
  github.com/tendermint/tendermint/vendor/github.com/spf13/viper.(*Viper).Set()
      /go/src/github.com/tendermint/tendermint/vendor/github.com/spf13/viper/viper.go:1080 +0x257
  github.com/tendermint/tendermint/rpc/test.TestURIBroadcastTxSync()
      /go/src/github.com/tendermint/tendermint/rpc/test/client_test.go:70 +0x386
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:610 +0xc9

Previous read at 0x00c4201856e0 by goroutine 21:
  [failed to restore the stack]

Goroutine 36 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:646 +0x52f
  testing.RunTests.func1()
      /usr/local/go/src/testing/testing.go:793 +0xb9
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:610 +0xc9
  testing.RunTests()
      /usr/local/go/src/testing/testing.go:799 +0x4ba
  testing.(*M).Run()
      /usr/local/go/src/testing/testing.go:743 +0x12f
  github.com/tendermint/tendermint/rpc/test.TestMain()
      /go/src/github.com/tendermint/tendermint/rpc/test/main_test.go:30 +0x91
  main.main()
      github.com/tendermint/tendermint/rpc/test/_test/_testmain.go:126 +0x2f8

Goroutine 21 (running) created at:
  github.com/tendermint/tendermint/consensus.(*ConsensusState).OnStart()
      /go/src/github.com/tendermint/tendermint/consensus/state.go:372 +0x5d5
```

cmd/tendermint/commands/reset_priv_validator.go
melekes: Ok, cool

glide.yaml
melekes: > better way

you test it using `glide mirror`, but, right before you send a PR, you change the version to develop, so it could be merged into develop?

rigelrozanski: hmm yeah, I've had problems using `glide mirror` I'll give it another shot next time

ethanfrey:
@rigelrozanski  Things like this byte me way to often when I try to make some minor update to tendermint core.  You can reproduce some of these locally with `make test_race` (I re-discovered this today while banging my head on some other issue).

When you stare at these stack traces and jump through the offending code long enough, you should see what is going on.  It has been a bit tricky to wrap my head around them usually, but I think debugging this did give me a better understanding of concurrency in our system.

node/node.go
```diff
@@ -119,7 +121,11 @@ func NewNode(config cfg.Config, privValidator *types.PrivValidator, clientCreato
...
+	p2pConfig := viper.New()
+	if config.IsSet("p2p") { //TODO verify this necessary, where is this ever set?
```
ebuchman: the idea was that tendermint's config could have a `[p2p]` section as a sub config that could be passed in here

rpc/test/client_test.go
```diff
@@ -64,17 +65,18 @@ func TestJSONBroadcastTxSync(t *testing.T) {
 }
 
 func testBroadcastTxSync(t *testing.T, client rpc.HTTPClient) {
-	config.Set("block_size", 0)
-	defer config.Set("block_size", -1)
+	//	config.Set("block_size", 0)
+	//	defer config.Set("block_size", -1)
```
ebuchman: why are these commented out?

cmd/tendermint/commands/reset_priv_validator.go
```diff
@@ -28,25 +30,38 @@ func init() {
...
 // XXX: this is totally unsafe.
 // it's only suitable for testnets.
 func resetPrivValidator(cmd *cobra.Command, args []string) {
+	resetPrivValidatorLocal(config, log)
+}
+
+func ResetAll(c *viper.Viper, l log15.Logger) {
+	resetPrivValidatorLocal(c, l)
+	dataDir := c.GetString("db_dir")
+	walFile := c.GetString("cs_wal_file")
```
rigelrozanski: so anything that uses walFile should use dataDir instead?

node/node.go
rigelrozanski: yup, I get the intention. I was just wondering if this had become antiquated, can remove the comment if we're certain we want to leave. Not currently used anywhere that I can find

rpc/test/client_test.go
rigelrozanski: Commented out because this was involved in the old (unsafe) way of performing this test, which caused race errors to the viper config... aka I modified this test routine but left the comment in there in-case. will remove though

cmd/tendermint/commands/reset_priv_validator.go
ebuchman: no I mean the walFile is now a folder within the dataDir, so if you wipe the dataDir, the walFile will be wiped too

PR #435, opened 2017-03-22
Repair draw_deps in Makefile

adrianbrink:
Sorry for the merge commit. Next time I'll rebase :-)

PR #443, opened 2017-04-10
Draw deps

ebuchman:
Replaces https://github.com/tendermint/tendermint/pull/435

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,9 +35,9 @@ test100:
 	@for i in {1..100}; do make test; done
 
 draw_deps:
-	# requires brew install graphviz
-	go get github.com/hirokidaichi/goviz
-	goviz -i ./cmd/tendermint | dot -Tpng -o huge.png
+	# requires brew install graphviz or apt-get install graphviz
+	go get github.com/RobotsAndPencils/goviz
+	@goviz -i github.com/tendermint/tendermint/cmd/tendermint -d 3 | dot -Tpng -o dependency-graph.png
 
 list_deps:
 	@go list -f '{{join .Deps "\n"}}' ./... | \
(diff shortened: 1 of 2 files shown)


ebuchman:
Hmm I'm not getting the same result:

dependency-graph

ebuchman:
@adrianbrink ^

adrianbrink:
@ebuchman What do you mean by: "I'm not getting the same results."

ebuchman:
My png looks different than the one you originally uploaded!

adrianbrink:
Mh, the difference between the big graph and the small graph is whether you pass 2 or 1 to the -d option. Maybe there is also a difference between platforms.

adrianbrink:
Should I change anything else before we can merge this?

ebuchman:
Mmm I'm using `-d 2` on a mac. I'd like to be able to reproduce that same graph with all the actual deps before we merge

adrianbrink:
I just ran it off the most current ```develop``` branch and ran ```glide install``` and now I am getting this image.

dependency-graph

ebuchman:
yeh. so what gives ?

adrianbrink:
I generated the [original image](https://github.com/tendermint/tendermint/pull/435/files#diff-0ff7d988a264a22196dcb21b415934ad) with ```-d 1```. 

Aren't we now getting the same results, which means that this should be mergeable :-)

Even though there is one small difference between the two pictures. You have an extra dependency ```github.com/spf13/cobra```. Are you also building on the most recent version of ```develop```?

ebuchman:
dependency-graph

That's what I get with `-d 1` :(

ebuchman:
`-d 3` ftw

PR #447, opened 2017-04-11
Add Docker orchestration

adrianbrink:
This PR adds the ability to run any ABCI app against 4 validator nodes with a single ```docker-compose up``` command. 

In the future I would like to extend it so that an arbitrary number of nodes can be spawned as well as that it can be used for production environments.


adrianbrink:
@jaekwon @ebuchman

DockerOrchestration/TendermintCore/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM golang:1.8
+MAINTAINER adrian@brink-holdings.com
+
+ARG priv_validator
```
ebuchman: WOW docker added build variables!

DockerOrchestration/ABCIApp/Dockerfile.Dummy
```diff
@@ -0,0 +1,15 @@
+FROM golang:1.8
```
ebuchman: is there really an advantage to having two dockerfiles here? everything is basically the same except the final command

DockerOrchestration/TendermintCore/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN go get github.com/Masterminds/glide
+RUN go get -d github.com/tendermint/tendermint/cmd/tendermint
+WORKDIR $GOPATH/src/github.com/tendermint/tendermint
+RUN glide install
+RUN go install ./cmd/tendermint
+
+# VOLUME $DATA_ROOT
+
+# p2p
+#EXPOSE 46656
```
ebuchman: why commented?

melekes:
I think we should move this into `https://github.com/tendermint/tools/mintnet-swarm` (swarm because, if I am not mistaken, it is the same thing except a few variables https://docs.docker.com/compose/production/#modify-your-compose-file-for-production)

DockerOrchestration/ABCIApp/Dockerfile.Dummy
melekes: Do we need a dockerfile at all? when we could run `tendermint node --proxy_app=dummy`

melekes: same for `counter`

DockerOrchestration/TendermintCore/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM golang:1.8
```
melekes: why not to use `tendermint/tendermint` container?

DockerOrchestration/TendermintCore/config.toml
```diff
@@ -0,0 +1,11 @@
+# This is a TOML config file.
```
melekes: no need for config either if it replicates default config. TM will create one if it is missing

DockerOrchestration/ABCIApp/Dockerfile.Dummy
adrianbrink: @ebuchman  Those dockerfiles will be cleaned up, I just wanted an easy way to test multiple ABCIs in this setup.

@melekes The reason why I have 2 dockerfiles is that I generally think that it is a better idea to run the core and abci nodes separately. 
It allows us to more easily scale the ABCI parts, since we can orchestrate multiple ABCI apps in front of a single core node. 
Furthermore, it makes the separation of interfaces easier. 
Lastly, I think most people will run the ABCI apps and core nodes as separate processes since it seems easier than running them in-process and hence my idea was to emulate it with the dummy apps. :-)

DockerOrchestration/TendermintCore/config.toml
adrianbrink: Ah perfect, I wasn't sure about that :-)

DockerOrchestration/TendermintCore/Dockerfile
adrianbrink: I'll switch to that. Before easter I wasn't sure whether those were actively maintained :)

adrianbrink:
@melekes I think you are right. I'll move it to the tools repo. I'll add another paragraph in the repo then on how to run it using docker.

@ebuchman Moving this to the tools repo will also solve the problem of merging it with the other DOCKER folder.

DockerOrchestration/ABCIApp/Dockerfile.Dummy
ebuchman: I'm all for having them run in separate processes. 

> since we can orchestrate multiple ABCI apps in front of a single core node.

I don't think this makes sense yet. Tendermint isn't set up for it.

ebuchman:
Cool - other repo seems like a better home. Thanks!

PR #455, opened 2017-04-21
Unstable

ebuchman:


rpc/test/client_test.go
```diff
@@ -85,16 +85,17 @@ func testBroadcastTxSync(t *testing.T, client rpc.HTTPClient) {
...
 	v := randBytes(t)
-	return k, v, []byte(Fmt("%s=%s", k, v))
+	return k, v, types.Tx(Fmt("%s=%s", k, v))
 }
 
 func sendTx(t *testing.T, client rpc.HTTPClient) ([]byte, []byte) {
 	tmResult := new(ctypes.TMResult)
 	k, v, tx := testTxKV(t)
-	_, err := client.Call("broadcast_tx_commit", map[string]interface{}{"tx": tx}, tmResult)
+	txBytes := []byte(tx) // XXX
```
ebuchman: the types.Tx was not being encoded correctly in the http call

types/priv_validator.go
```diff
@@ -105,13 +97,14 @@ func LoadPrivValidator(filePath string) *PrivValidator {
 	if err != nil {
 		Exit(err.Error())
 	}
-	privVal := wire.ReadJSON(&PrivValidator{}, privValJSONBytes, &err).(*PrivValidator)
+	privVal := PrivValidator{}
```
ebuchman: why not just `privVal := new(PrivValidator)`, since we always need the pointer ?

ethanfrey: `&PrivValidator{}` and `new(PrivValidator)` are the same.  We could take the pointer here.

In all the go code I read before coming to tendermint (which was various repos, but not extensive), I saw the first idiom much more often and it is easier for me to parse visually.  I think you like the `new` format.  Is there a reason, or just style?

We should add this to coding guidelines.  But I wonder why you like new? (except for instantiating pointers to interfaces which no sane code should do)

rpc/test/client_test.go
ethanfrey: Can we trace why this is the case?  It seem quite odd and maybe a bug in another level we should fix rather than work around.

ethanfrey: Especially given the definition `type Tx []byte` very, very odd.  I thought for a second it was using data.Bytes or some such, but here...

types/priv_validator.go
ebuchman: I don't have a preference, but since the only use of `privVal` here is the pointer, I'd rather use `new` than have to include `&` everywhere later

ethanfrey: I understand not wanting `privVal := PrivValidator{}`.  But the two solutions are functionally the same:

* `privVal := &PrivValidator{}`
* `privVal := new(PrivValidator)`

Do you want to enforce one style?  Or ask other devs what they prefer?

On a related note... Just did: `bres := (*tmResult).(*ctypes.ResultBroadcastTxCommit)` and cringed again

ebuchman: yes right. either of those two is fine. I lean towards the latter one.

melekes: I prefer the former one

Issue #457, opened 2017-04-25
Tendermint's http endpoint response status codes


davebryson:
BUG REPORT
version: 0.9.0
issue with Tendermint's http rpc return status codes

Making a GET call to any undefined path returns status code 200.  For non-browser clients this makes it difficult to detect an error and determine whether or not to parse the response.  For example, calling:

GET http://localhost:46657/broadcast_tx_cc  returns status code 200 (OK)  'broadcast_tx_cc' is not actually a recognized command.  It should return a NON 200 code, perhaps a 501 code.

ebuchman:
Great point. We currently just dump an html list of all the endpoints, which is convenient in the browser, but we should change to only do that for `/`, and to return a real error for any undefined route.

melekes:
> We currently just dump an html list of all the endpoints, which is convenient in the browser, but we should change to only do that for /

this is exactly what we're doing right now https://github.com/tendermint/go-rpc/blob/develop/server/handlers.go#L32

> to return a real error for any undefined route.

we do return an error, but as @davebryson said, the status code in response is always 200 
https://github.com/tendermint/go-rpc/blob/develop/server/handlers.go#L120
https://github.com/tendermint/go-rpc/blob/develop/server/http_server.go#L56

ebuchman:
How about something like https://github.com/tendermint/tendermint/pull/459

ebuchman:
merged on develop

PR #459, opened 2017-04-26
rpc: use HTTP error codes

ebuchman:
https://github.com/tendermint/tendermint/issues/457

### DIFF
--- a/rpc/server/handlers.go
+++ b/rpc/server/handlers.go
@@ -14,8 +14,8 @@ import (
 
 	"github.com/gorilla/websocket"
 	"github.com/pkg/errors"
-	types "github.com/tendermint/tendermint/rpc/types"
 	wire "github.com/tendermint/go-wire"
+	types "github.com/tendermint/tendermint/rpc/types"
 	cmn "github.com/tendermint/tmlibs/common"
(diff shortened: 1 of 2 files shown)


mappum:
👍 Good choices of response codes

ebuchman:
Thanks. I wasn't sure about some - like 404 when the http endpoint is found but the jsonrpc method is not ...

mappum:
That use of 404 is common in REST APIs, it should be fine.

Soon we'll be able to make use of 402 (Payment Required) 😄

mappum:
Looks like it matches the JSONRPC spec closely enough: http://www.jsonrpc.org/historical/json-rpc-over-http.html#errors

ebuchman:
Cool. What's the deal with those codes, beside status? 

And should we be doing things like 202 for broadcast_tx_sync where its in the mempool but hasn't been committed?

mappum:
Those are the codes to be set in the response JSON. But honestly we don't need to adhere to the spec 100%, people will probably just check the HTTP status code over the JSONRPC code. I'd never heard of 202, and some clients just check `statusCode == 200` so probably not worth it.

PR #463, opened 2017-04-27
InitChain message gets send at genesis.

ebuchman:
Openning PR for review


ebuchman:
Interesting. I was thinking of putting this in the node/node.go but maybe this is better. I think we can get away with just one call tho - why don't we just add a single `if appHeight == 0` and send InitChain there, before we start the rest of the logic ?

adrianbrink:
I thought about node.go as well, but it felt more natural here.

I pulled it out of the conditional blocks and removed some old comments.

ebuchman:
Great. Also, let's move the `covertValidators` function to here: https://github.com/tendermint/tendermint/blob/master/types/protobuf.go

consensus/replay.go
```diff
@@ -257,6 +257,13 @@ func (h *Handshaker) ReplayBlocks(appHash []byte, appBlockHeight int, proxyApp p
 	stateBlockHeight := h.state.LastBlockHeight
 	log.Notice("ABCI Replay Blocks", "appHeight", appBlockHeight, "storeHeight", storeBlockHeight, "stateHeight", stateBlockHeight)
 
+	// If appBlockHeight == 0 it means that we are at genesis and hence should send InitChain
+	// TODO: Currently we have 2 separate types of Validators (abci and types) and hence we need the conversion.
```
ebuchman: no need for this TODO

ebuchman:
Ok. remove that TODO and squash down to one commit and we're good :)

consensus/replay.go
adrianbrink: Good catch

adrianbrink:
@ebuchman It fails, but it still seems cryptic to me. The TestHandshakeReplayAll test fails.

adrianbrink:
@ebuchman The tests are passing locally inside the vagrant box. I've pushed again to see if they pass CircleCI.

I think the problem for the inconsistent test failures was that the VM didn't have enough memory. How much memory is available in CircleCI?

melekes:
@adrianbrink note that CircleCI does not use our Vagrantfile, so setting higher memory for it won't help.

ebuchman:
I believe the issue was due to some hard coded test data that is changed by InitChain, since calling InitChain actually updates the app state, this changes the app root hash

adrianbrink:
@melekes Yes, but I think that's the reason why my test failed locally in the beginning. Plus the fact that some of the test data was incorrect.

@ebuchman That would be nice. You did update the tests in 495283e, correct?

ebuchman:
Yep

PR #465, opened 2017-04-27
RPC Serialization Overhaul

ebuchman:
- types: `[]byte -> data.Bytes`
- Introduce EventDataInner for serializing events
- remove all use of go-wire around user interfaces
- rpc responses no longer have type information :)

NOTE: go-wire JSON is still used for SignBytes and for the consensus WAL


ebuchman:
I suspect the codecov hit is related to copying in 200 lines from go-wire without tests. It would be better if we exposed that function in go-wire instead, but it needs to be modified to check if a byte array is data.Bytes, which is not possible because it causes import cycle. Should we merge `go-wire/data` into `go-wire` ? Or maybe move `data.Bytes` into `go-wire/types` ? That would at least resolve the import cycle but not sure if its the best idea. @ethanfrey @jaekwon

ebuchman:
:D

PR #473, opened 2017-05-01
Logging architecture proposal

melekes:


adrianbrink:
@melekes Should this be merged, since the implementation is already merged?

melekes:
I will update it once all the PRs are merged and impl is stable.

PR #475, opened 2017-05-01
Config structs

ebuchman:
- Isolate viper. Besides config/cmd packages, only used in one place per package to unmarshal test config
- Config structs. Main struct is in `tendermint/config/config.go`. Others are defined in `consensus`, `mempool`, `p2p`. The `config.toml` unmarshalls into `node.Config`, which contains all the configs


ebuchman:
Question: should config structs be all together in the config package, or should we put each one in its respect package - eg. `mempl.Config` rather than `cfg.MempoolConfig` or `cfg.Mempool` ?

adrianbrink:
In terms of readability I think the first and last one are the best. 

Also, I like the idea of grouping the config struct with the code that will use it, however I see the advantages of grouping them under ```cfg```, since it will allow for a super simple overview.

In the end however, I think I would prefer to put them each in their respective package.

cmd/tendermint/commands/run_node.go
```diff
@@ -65,16 +66,16 @@ func init() {
...
-	config.Set("abci", abciTransport)
-	config.Set("pex_reactor", pex)
+	config.Set("node.moniker", moniker)
+	config.Set("node.listen_addr", nodeLaddr)
+	config.Set("network.seeds", seeds)
+	config.Set("network.skip_upnp", skipUPNP)
+	config.Set("network.pex_reactor", pex)
+	config.Set("blockchain.fast_sync", fastSync)
+	config.Set("rpc.listen_addr", rpcLaddr)
+	config.Set("rpc.grpc_listen_addr", grpcLaddr)
```
melekes: `grpc.listen_addr` ^^^

config/tendermint/config.go
```diff
@@ -53,59 +53,54 @@ func GetConfig(rootDir string) *viper.Viper {
...
+	config.SetDefault("network.skip_upnp", false)
+	config.SetDefault("blockchain.fast_sync", true)
+	config.SetDefault("network.addrbook_file", rootDir+"/addrbook.json")
+	config.SetDefault("network.addrbook_strict", true) // disable to allow connections locally
+	config.SetDefault("network.pex_reactor", false)    // enable for peer exchange
+	config.SetDefault("node.priv_validator_file", rootDir+"/priv_validator.json")
+	config.SetDefault("db.backend", "leveldb")
+	config.SetDefault("db.dir", rootDir+"/data")
+	config.SetDefault("rpc.listen_addr", "tcp://0.0.0.0:46657")
+	config.SetDefault("rpc.grpc_listen_addr", "")
```
melekes: `grpc.listen_addr` (or change line above)

config/tendermint/types.go
```diff
@@ -0,0 +1,89 @@
...
+	Blockchain BlockchainConfig `mapstructure:"blockchain"`
+	Consensus  ConsensusConfig  `mapstructure:"consensus"`
+	Block      BlockConfig      `mapstructure:"block"`
+	Mempool    MempoolConfig    `mapstructure:"mempool"`
+	RPC        RPCConfig        `mapstructure:"rpc"`
+	DB         DBConfig         `mapstructure:"db"`
+}
+
+type NodeConfig struct {
+	Moniker           string `mapstructure:"moniker"`             // "anonymous"
```
melekes: don't think comments are necessary here

melekes:
> In the end however, I think I would prefer to put them each in their respective package.

agree

besides that, awesome job 💚 ❤️ 💛

consensus/common_test.go
```diff
@@ -37,6 +37,23 @@ func ensureDir(dir string, mode os.FileMode) {
...
 }
 
+type NodeConfig struct {
+	cfg.Config `mapstructure:",squash"`
+	P2P        *p2p.Config   `mapstructure:"p2p"`
+	Mempool    *mempl.Config `mapstructure:"mempool"`
+	Consensus  *Config       `mapstructure:"consensus"`
+}
+
+// TODO: This is the same as NodeConfig. Should we move the configs to types (?)
```
ebuchman: Note this is a duplicate of node.Config. Not sure if we can do better but it kind of bugs me. One solution is to put all the configs together where any package can import them, like tendermint/tendermint/config, but it does seem better to have a config in its own package. Maybe this is just the price and its fine?

melekes: and why we cannot use `node.Config` here?

ebuchman: import cycle!

melekes: oh.. then I guess types/config.go is the price we'll have to pay

ebuchman: You think putting all config in `types/config.go` is better than this duplication of the definition of the top level config ?

melekes: I don't like the duplication. What if somebody forgets to update the copy?

ebuchman: An alternative is to give up control of eg the p2p config details in the consensus tests from the test toml file. Eg. here we only unmarshal the piece we need: https://github.com/tendermint/tendermint/blob/config_structs/mempool/mempool_test.go#L14

We could do that also in consensus test and just use p2p.NewDefaultConfig where we need a p2p config ...

ethanfrey: Okay, maybe everyone will hate me now, but... (with a very brief review of the code)...

What about putting all the config files together in `types/config.go` and just import that.  They seem to be needed in multiple packages, and data that is passed from package to package all seems to go in `types`.

If nothing else, they are needed by command package to parse into, and the other packages to configure themselves.

ebuchman: Yes I think that's reasonable and was actually my initial intuition. But I was lured by the appeal of having config structs defined where they are used (something about being able to look for the consensus config in the consensus package just feels right). As right as that feels, it makes other things feel quite wrong, so probably best to put all the config together in types/config.go. So we'll have `Config`, `ConsensusConfig`, `MempoolConfig`, `P2PConfig`, and then `NodeConfig` will include all the above ?

I should have put this all down in an ADR. I'm doing that now too

ebuchman: Hmm. `p2p` doesn't currently import `types` and I don't think it should. and `types` shouldn't import `p2p`. So unless we remove p2p options from the config file, we're back to square one. 

Maybe we should be thinking about why some packages need to read the whole config file. This all arose because `consensus` ends up using the same full config as `node`, since it uses p2p and mempool, but has to duplicate the definition to avoid import cycle. What if instead `consensus` only loads its own config, and uses defaults for mempool and p2p. Then we can mostly keep things as they are, and not have the duplicated definition, but we would lose the ability to change p2p parameters from the config file in the consensus unit tests. That said, we'd still be doing this once in each package for tests:

```
func ResetConfig(name string) *Config {
        viperConfig := tendermint_test.ResetConfig(name)
        config := new(struct {
                cfg.Config `mapstructure:",squash"`
                Mempool    *Config `mapstructure:"mempool"`
        })  
        if err := viperConfig.Unmarshal(config); err != nil {
                panic(err)
        }   
        return config.Mempool

}
```

ebuchman:
Closed in favour of https://github.com/tendermint/tendermint/pull/484

Issue #480, opened 2017-05-03
Logging if node is validator/observer


mdyring:
It seems like a valuable addition to log if node is participating as validator or observer, as it is not clear from just looking at the configuration (validator set might change).

I suggest logging "Node is participating as observer" or "Node is participating as validator" when there is a state change.

@melekes pointed out on slack that app should also log this when it changes the validator set via ABCI, but there was agreement that since there is the validator set is bootstrapped via config, that it makes sense to log in tendermint core as well.

ebuchman:
This is cool. We can just check if the node's priv_validator key is in the validator set on startup and log the result.

mdyring:
Since validator set might change over time (via ABCI), ideally it should be logged when it changes.. 
But yes on startup would be good as well, just as long as it takes into account the most recent validator set (so after replay on startup I guess).

ebuchman:
Oh yes of course. So on startup after replay (once State has latest validators), and once removed or added to the val set.

adrianbrink:
PR: #488

haasted:
I'd like to suggest that this is feature is expanded to have all nodes in the network log any modifications they see in the validator set, e.g. that another node switches from observer to validator, or that a new validator is added to the P2P network. Would this be possible?

adrianbrink:
@haasted I've added the extended logging as another issue.

srmo:
What about a message to the abci app informing it about the current valset and votepower, and if the connected node is validator or not? Maybe like the current info handshake or as answer to an rpc call?

PR #488, opened 2017-05-06
Logging if node is validator/observer

adrianbrink:
This PR implements #480 .

@ebuchman What's our convention for log levels? Should it be Notice?

Also, for logging when the validator changes over time I thought about putting it in state.go into SetBlockAndValidators(), but I have no access to the current node and hence can't figure out whether the current node was removed. Is there a global way to get the node that is currently running?

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -775,15 +775,21 @@ func (cs *ConsensusState) enterPropose(height int, round int) {
 
 	// Nothing more to do if we're not a validator
 	if cs.privValidator == nil {
+		cs.Logger.Info("This node is not a validator")
 		return
 	}
 
 	if !bytes.Equal(cs.Validators.GetProposer().Address, cs.privValidator.GetAddress()) {
 		cs.Logger.Info("enterPropose: Not our turn to propose", "proposer", cs.Validators.GetProposer().Address, "privValidator", cs.privValidator)
+		if cs.Validators.HasAddress(cs.privValidator.GetAddress()) {
+			cs.Logger.Info("This node is a validator")
+		} else {
+			cs.Logger.Info("This node is not a validator")
+		}
 	} else {
 		cs.Logger.Info("enterPropose: Our turn to propose", "proposer", cs.Validators.GetProposer().Address, "privValidator", cs.privValidator)
+		cs.Logger.Info("This node is a validator")
 		cs.decideProposal(height, round)
-
 	}
 }
 
@@ -1141,6 +1147,7 @@ func (cs *ConsensusState) tryFinalizeCommit(height int) {
 		cs.Logger.Error("Attempt to finalize failed. We don't have the commit block.", "height", height, "proposal-block", cs.ProposalBlock.Hash(), "commit-block", blockID.Hash)
 		return
 	}
+
 	//	go
 	cs.finalizeCommit(height)
 }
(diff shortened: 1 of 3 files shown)


state/state.go
```diff
@@ -143,6 +143,11 @@ func (s *State) SetBlockAndValidators(header *types.Header, blockPartsHeader typ
 		log.Warn("Error changing validator set", "error", err)
 		// TODO: err or carry on?
 	}
+
+	// TODO: Log whether this node was removed from the validator set or added to it.
```
ebuchman: You should be able to do this within ConsensusState

node/node.go
```diff
@@ -115,6 +115,13 @@ func NewNode(config *cfg.Config, privValidator *types.PrivValidator, clientCreat
 		}
 	}
 
+	// Log whether this node is a validator or an observer
+	if state.Validators.HasAddress(privValidator.Address) {
+		log.Notice("This node is a validator")
+	} else {
+		log.Notice("This node is an observer")
```
ebuchman: Let's say "This node is not a validator"

state/state.go
adrianbrink: Done

ebuchman: Remove these comments.

consensus/state.go
```diff
@@ -521,6 +521,13 @@ func (cs *ConsensusState) updateToState(state *sm.State) {
 		lastPrecommits = cs.Votes.Precommits(cs.CommitRound)
 	}
 
+	// Log whether this node is a validator
+	if privVal := cs.privValidator; privVal != nil && validators.HasAddress(privVal.GetAddress()) {
```
ebuchman: Should we do this every time or only if it changes ?

adrianbrink: I thought about it and I think it is better if we do, because otherwise you would have to track the context of each block in the log.
For example, if you wanted to see whether I was a validator for a specific block, I would have to go through the previous log messages until I find one that shows that I was indeed the validator.

Also, once we are moving towards structured logging in JSON I think it will be helpful to be able to easily display graphical logs that show whether you are currently a validator or not.

Actually, do we log whether we are the proposer (if not we should probably add that)?

state/state.go
adrianbrink: Sorry, I thought I had removed those. Fixed now and pushed.

consensus/state.go
ebuchman: > Actually, do we log whether we are the proposer (if not we should probably add that)?

Great point! So let's just do the logging there: https://github.com/tendermint/tendermint/blob/master/consensus/state.go#L809

Either you're a proposer, or you're not a proposer, or you're not a validator. I'd think that, if you're a validator, it should log whether you're the proposer or not, otherwise nothing. But I guess we could have three cases and log for each of them: proposer, validator but not proposer, not a validator.

adrianbrink: done :-)

consensus/state.go
```diff
@@ -764,15 +764,17 @@ func (cs *ConsensusState) enterPropose(height int, round int) {
 
 	// Nothing more to do if we're not a validator
 	if cs.privValidator == nil {
+		log.Notice("This node is not a validator")
 		return
 	}
 
 	if !bytes.Equal(cs.Validators.GetProposer().Address, cs.privValidator.GetAddress()) {
 		log.Info("enterPropose: Not our turn to propose", "proposer", cs.Validators.GetProposer().Address, "privValidator", cs.privValidator)
+		log.Notice("This node is a validator")
```
adrianbrink: A node can have a privValidator without it being in the validator set. 

Do the same check as in node/node.go

ebuchman:
merged https://github.com/tendermint/tendermint/commit/eb9ca2325071035c33029e9e1ad883e08f1d7e3d

Issue #469, opened 2017-04-29
Use ld flags to set git hash instead of "revision_file"


ebuchman:
https://github.com/tendermint/ethermint/blob/master/Makefile#L12

adrianbrink:
PR: #489

PR #489, opened 2017-05-06
Use ld flags to set git hash instead of "revision_file"

adrianbrink:
Issue: #469

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,8 @@ install: get_vendor_deps
 	@go install ./cmd/tendermint
 
 build:
-	go build -o build/tendermint ./cmd/tendermint
+	go build --ldflags '-extldflags "-static"' \
+		--ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse HEAD`"  -o build/tendermint ./cmd/tendermint/
 
 build_race:
 	go build -race -o build/tendermint ./cmd/tendermint
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ TMHOME = $${TMHOME:-$$HOME/.tendermint}
 all: install test
 
 install: get_vendor_deps
-	@go install ./cmd/tendermint
+	@go install --ldflags '-extldflags "-static"' ./cmd/tendermint
 
 build:
 	go build --ldflags '-extldflags "-static"' \
--- a/Makefile
+++ b/Makefile
@@ -8,10 +8,11 @@ TMHOME = $${TMHOME:-$$HOME/.tendermint}
 all: install test
 
 install: get_vendor_deps
-	@go install --ldflags '-extldflags "-static"' ./cmd/tendermint
+	@go install --ldflags '-extldflags "-static"' \
+		--ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse HEAD`" ./cmd/tendermint
 
 build:
-	go build --ldflags '-extldflags "-static"' \
+	go build \
 		--ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse HEAD`"  -o build/tendermint ./cmd/tendermint/
 
 build_race:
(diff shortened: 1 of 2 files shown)


version/version.go
```diff
@@ -4,4 +4,4 @@ const Maj = "0"
 const Min = "10"
 const Fix = "0"
 
-const Version = "0.10.0"
+const Version = Maj + "." + Min + "." + Fix
```
melekes: https://github.com/tendermint/tendermint/blob/6312eb91bee05171d97cecd509be115b2c6b6477/scripts/dist.sh#L6

we should probably add a comment here

Makefile
```diff
@@ -11,7 +11,8 @@ install: get_vendor_deps
 	@go install ./cmd/tendermint
 
 build:
-	go build -o build/tendermint ./cmd/tendermint
+	go build --ldflags '-extldflags "-static"' \
```
melekes: this command is mostly used by devs. Do we need these flags for dev build?

this is where we build a release
https://github.com/tendermint/tendermint/blob/6312eb91bee05171d97cecd509be115b2c6b6477/scripts/dist_build.sh#L38

version/version.go
adrianbrink: Could you explain? 

Is the reason why we write it out as "0.10.0" that we grab the value from the dist.sh script?

melekes: > Is the reason why we write it out as "0.10.0" that we grab the value from the dist.sh script?

yes

ebuchman: Yes let's explicitly write it. Should add this to the coding standards too that repos should have a `version/version.go` structured like this.

Makefile
```diff
@@ -11,7 +11,8 @@ install: get_vendor_deps
...
+		--ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse HEAD`"  -o build/tendermint ./cmd/tendermint/
```
ebuchman: dont we need to define version.GitCommit ?

Makefile
ebuchman: I dont think so. But it might be nice to have in `make install` for reference when people are installing development version

version/version.go
adrianbrink: I've added it to the coding standards.

adrianbrink:
Okay, so I've added GitCommit and moved the static ldflags to ```make install```.

The version number is now being updated with the GitCommit hash that it was build from and our build script can still get the original pure version number.

This should enable us to have clearer error reports in the future.

ebuchman:
Would it make sense for the git commit to only be set for non-release builds? Or do we want `tendermint version` to always include a commit ?

Makefile
```diff
@@ -8,10 +8,12 @@ TMHOME = $${TMHOME:-$$HOME/.tendermint}
...
 
 install: get_vendor_deps
-	@go install ./cmd/tendermint
+	@go install --ldflags '-extldflags "-static"' \
+		--ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse HEAD`" ./cmd/tendermint
 
 build:
-	go build -o build/tendermint ./cmd/tendermint
+	go build \
+		--ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse HEAD`"  -o build/tendermint ./cmd/tendermint/
```
melekes: you can extract `git...` in a variable (example https://github.com/Masterminds/glide/blob/master/Makefile#L3)

melekes:
> Would it make sense for the git commit to only be set for non-release builds? Or do we want tendermint version to always include a commit ?

if the only point we set it is to be able to inspect/debug the binary someone gives us, then no.

adrianbrink:
I think it should always include the git commit, because then it enables us to have uniform reporting of errors.

Makefile
adrianbrink: I think it is easier to read when inlined like this.

PR #494, opened 2017-05-13
Clean up json output

ethanfrey:
Adjust naming of a few fields (lower case), and pretty print the rpc output for nice viewing

### DIFF
--- a/rpc/lib/server/handlers.go
+++ b/rpc/lib/server/handlers.go
@@ -539,7 +539,7 @@ func (wsc *wsConnection) writeRoutine() {
 				return
 			}
 		case msg := <-wsc.writeChan:
-			jsonBytes, err := json.Marshal(msg)
+			jsonBytes, err := json.MarshalIndent(msg, "", "  ")
 			if err != nil {
 				wsc.Logger.Error("Failed to marshal RPCResponse to JSON", "error", err)
 			} else {
--- a/types/validator_set.go
+++ b/types/validator_set.go
@@ -6,9 +6,9 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/tendermint/go-wire"
 	cmn "github.com/tendermint/tmlibs/common"
 	"github.com/tendermint/tmlibs/merkle"
-	"github.com/tendermint/go-wire"
 )
 
 // ValidatorSet represent a set of *Validator at a given height.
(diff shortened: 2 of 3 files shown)


rpc/lib/server/handlers.go
```diff
@@ -538,7 +538,7 @@ func (wsc *wsConnection) writeRoutine() {
 				return
 			}
 		case msg := <-wsc.writeChan:
-			jsonBytes, err := json.Marshal(msg)
+			jsonBytes, err := json.MarshalIndent(msg, "", "  ")
```
melekes: Shouldn't it be the responsibility of a client if it wants JSON to look pretty? `curl ... | json_pp` for example

ebuchman: Yes.

types/validator_set.go
```diff
@@ -23,8 +23,8 @@ import (
 // TODO: consider validator Accum overflow
 // TODO: move valset into an iavl tree where key is 'blockbonded|pubkey'
 type ValidatorSet struct {
-	Validators []*Validator // NOTE: persisted via reflect, must be exported.
-	Proposer   *Validator
+	Validators []*Validator `json:"validators"`
```
ebuchman: Let's keep the comment but move it above the struct def

rpc/lib/server/handlers.go
ethanfrey: Can't do this with websockets.

I find it very nice to work with api's like this.  Especially when you check them with the webpage.  Just more readable.  And it is not much more overhead.  If you use gzip compression (if you care about bandwidth, we can add a middleware), then it goes down to a few % overhead).

Maybe this is just my aethetics going overboard.  But I see no reason not to make it more readable.

types/validator_set.go
ethanfrey: Corrected in 926fb83

rpc/lib/server/handlers.go
ethanfrey: Here's a nice article about lots on api usability, where I got this idea.  Linking to one section, but it is nice to take a look at the whole thing...

http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#pretty-print-gzip

melekes: makes sense

ebuchman: > Can't do this with websockets.

Why not? You dont need to see the pretty print until it hits a terminal or a webpage. Install a json pretty printer plugin. I do and its way better than just using MarshalIndent (colors and close/open functionality and so on)

PR #501, opened 2017-05-15
Feature/493 per module log levels

melekes:
DEPENDS ON https://github.com/tendermint/tmlibs/pull/17

### DIFF
--- a/cmd/tendermint/commands/flags/log_level.go
+++ b/cmd/tendermint/commands/flags/log_level.go
@@ -1,0 +1,85 @@
+package flags
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+
+	cfg "github.com/tendermint/tendermint/config"
+	"github.com/tendermint/tmlibs/log"
+)
+
+const (
+	defaultLogLevelKey = "*"
+)
+
+// IsLogLevelSimple returns true if log level is a single word ("info", ...).
+func IsLogLevelSimple(l string) bool {
+	return !strings.Contains(l, ":")
+}
+
+// ParseComplexLogLevel parses complex log level - comma-separated
+// list of module:level pairs with an optional *:level pair (* means
+// all other modules).
+//
+// Example:
+//		ParseComplexLogLevel("consensus:debug,mempool:debug,*:error", log.NewTMLogger(os.Stdout))
+func ParseComplexLogLevel(l string, logger log.Logger) (log.Logger, error) {
+	if l == "" {
+		return nil, errors.New("Empty log level")
+	}
+
+	options := make([]log.Option, 0)
+
+	isDefaultLogLevelSet := false
+	var option log.Option
+	var err error
+
+	list := strings.Split(l, ",")
+	for _, item := range list {
+		moduleAndLevel := strings.Split(item, ":")
+
+		if len(moduleAndLevel) != 2 {
+			return nil, fmt.Errorf("Expected list in a form of \"module:level\" pairs, given pair %s, list %s", item, list)
+		}
+
+		module := moduleAndLevel[0]
+		level := moduleAndLevel[1]
+
+		if module == defaultLogLevelKey {
+			option, err = log.AllowLevel(level)
+			if err != nil {
+				return nil, errors.Wrap(err, fmt.Sprintf("Failed to parse default log level (pair %s, list %s)", item, l))
+			}
+			options = append(options, option)
+			isDefaultLogLevelSet = true
+		} else {
+			switch level {
+			case "debug":
+				option = log.AllowDebugWith("module", module)
+			case "info":
+				option = log.AllowInfoWith("module", module)
+			case "error":
+				option = log.AllowErrorWith("module", module)
+			case "none":
+				option = log.AllowNoneWith("module", module)
+			default:
+				return nil, fmt.Errorf("Expected either \"info\", \"debug\", \"error\" or \"none\" log level, given %s (pair %s, list %s)", level, item, list)
+			}
+			options = append(options, option)
+
+		}
+	}
+
+	// if "*" is not provided, set default global level
+	if !isDefaultLogLevelSet {
+		option, err = log.AllowLevel(cfg.DefaultBaseConfig().LogLevel)
+		if err != nil {
+			return nil, err
+		}
+		options = append(options, option)
+	}
+
+	return log.NewFilter(logger, options...), nil
+}
--- a/cmd/tendermint/commands/flags/log_level.go
+++ b/cmd/tendermint/commands/flags/log_level.go
@@ -14,20 +14,22 @@ const (
 	defaultLogLevelKey = "*"
 )
 
-// IsLogLevelSimple returns true if log level is a single word ("info", ...).
-func IsLogLevelSimple(l string) bool {
-	return !strings.Contains(l, ":")
-}
-
-// ParseComplexLogLevel parses complex log level - comma-separated
+// ParseLogLevel parses complex log level - comma-separated
 // list of module:level pairs with an optional *:level pair (* means
 // all other modules).
 //
 // Example:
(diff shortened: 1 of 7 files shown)


ebuchman:
Hmm. I like that the second approach doesnt change the NewNode signature. After reading the first one I thought of a `Loggers interface` that exposes `Consensus() log.Logger` etc. so we can just pass a `loggers` into the NewNode and call the respective method to get each module logger. Then we shouldn't nee to modify tmlibs, and we'd have a unified way to init all the loggers for a tendermint node, but we'd still be changing the NewNode sig (though the number of args wouldnt change, and we'd just be replacing a `logger` with a `loggers`). But your second approach is probably better, unless I'm missing something ...

cmd/tendermint/commands/flags/log_level.go
```diff
@@ -0,0 +1,66 @@
...
+// - comma-separated list of module:level pairs with an optional *:level pair (* means all other modules).
+func ParseLogLevel(l string) (moduleToLogLevelMap map[string]string, err error) {
+	if l == "" {
+		return nil, errors.New("Empty log level")
+	}
+
+	moduleToLogLevelMap = make(map[string]string)
+
+	if IsLogLevelSimple(l) {
+		if l != "info" && l != "debug" && l != "error" && l != "none" {
```
ebuchman: probably worth moving this check to tmlibs

cmd/tendermint/commands/run_node.go
```diff
@@ -79,8 +81,45 @@ func runNode(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	moduleToLogLevelMap, err := tmflags.ParseLogLevel(config.LogLevel)
```
ebuchman: shouldn't we just put all this logic in root? `tendermint node` is prob the only command that really needs it for now, but down the road we may forget to enable if there are other commands that could benefit from fine grained logger control ...

cmd/tendermint/commands/run_node.go
```diff
@@ -79,8 +81,45 @@ func runNode(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return err
+	}
+
+	options := make([]log.Option, 0)
+	for module, level := range moduleToLogLevelMap {
+		if module == "*" {
+			switch level {
+			case "debug":
+				options = append(options, log.AllowDebug())
```
ebuchman: can't we do these as `AllowXxxWith` with no args ? and then only need the switch statement once ?

node/node.go
```diff
@@ -37,9 +37,10 @@ type Node struct {
 	cmn.BaseService
 
 	// config
-	config        *cfg.Config
-	genesisDoc    *types.GenesisDoc    // initial validator set
-	privValidator *types.PrivValidator // local node's validator key
+	config              *cfg.Config
+	genesisDoc          *types.GenesisDoc    // initial validator set
+	privValidator       *types.PrivValidator // local node's validator key
+	moduleToLogLevelMap map[string]string    // logging levels per module
```
ebuchman: dont think this is needed now

node/node_test.go
```diff
@@ -12,7 +12,7 @@ func TestNodeStartStop(t *testing.T) {
 	config := cfg.ResetTestRoot("node_node_test")
 
 	// Create & start node
-	n := NewNodeDefault(config, log.TestingLogger())
+	n := NewNodeDefault(config, log.TestingLogger(), make(map[string]string))
```
ebuchman: nope

rpc/test/helpers.go
```diff
@@ -104,11 +104,10 @@ func NewTendermint(app abci.Application) *nm.Node {
 	// Create & start node
 	config := GetConfig()
 	logger := log.NewTMLogger(log.NewSyncWriter(os.Stdout))
-	logger = log.NewFilter(logger, log.AllowError())
 	privValidatorFile := config.PrivValidatorFile()
 	privValidator := types.LoadOrGenPrivValidator(privValidatorFile, logger)
 	papp := proxy.NewLocalClientCreator(app)
-	node := nm.NewNode(config, privValidator, papp, logger)
+	node := nm.NewNode(config, privValidator, papp, logger, map[string]string{"*": "error"})
```
ebuchman: also nope

cmd/tendermint/commands/run_node.go
melekes: can't we do these as AllowXxxWith with no args ?

probably not a good idea... from the user perspective `AllowInfoWith()` is like calling `SetLevel()` without passing an "info" as an arg

ethanfrey:
I love this work.  So nice and clean code....  I would like to move the commands/flags directory into tmlibs, so I can use this same logger parsing logic in the light-client, basecoin, etc....

I will make a few comments on the code to make it more reusable for my usecase.

cmd/tendermint/commands/flags/log_level.go
```diff
@@ -0,0 +1,85 @@
...
+	cfg "github.com/tendermint/tendermint/config"
+	"github.com/tendermint/tmlibs/log"
+)
+
+const (
+	defaultLogLevelKey = "*"
+)
+
+// IsLogLevelSimple returns true if log level is a single word ("info", ...).
+func IsLogLevelSimple(l string) bool {
```
ethanfrey: Can we just expose one method that Parses simple or complex log levels?

We can keep these two, just add one that does the check for simple, like you do in root.go

cmd/tendermint/commands/flags/log_level_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	complexFlags := []string{"mempool:error", "mempool:error,*:debug"}
+	for _, f := range complexFlags {
+		if tmflags.IsLogLevelSimple(f) {
+			t.Fatalf("%s is a complex flag", f)
+		}
+	}
+}
+
+func TestParseComplexLogLevel(t *testing.T) {
```
ethanfrey: Can you add a test that they work properly? Not just that they parse?

You have some nice examples of how these log levels are added... you could create a logger writing to a bytes.Buffer, then parse a log level, and send a message that should either be blocked or added, then check if it something was written to the buffer or not.

cmd/tendermint/commands/root.go
```diff
@@ -26,9 +28,21 @@ var RootCmd = &cobra.Command{
 		err := viper.Unmarshal(config)
 		config.SetRoot(config.RootDir)
 		cfg.EnsureRoot(config.RootDir)
-		logger, err = log.NewFilterByLevel(logger, config.LogLevel)
-		if err != nil {
-			return err
+		if tmflags.IsLogLevelSimple(config.LogLevel) {
```
ethanfrey: Can you move this logic into flags/log_level.go?

cmd/tendermint/commands/run_node.go
```diff
@@ -79,6 +80,14 @@ func runNode(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	if !tmflags.IsLogLevelSimple(config.LogLevel) {
```
ethanfrey: why is this in run_node.go rather than integrated in root.go with the other parse logic?

melekes:
@ethanfrey addressed your comments in https://github.com/tendermint/tendermint/pull/501/commits/fb0df75de097c2b92b4dfb5396db36e02e526bbe
code looks even better now, thanks!

node/node.go
```diff
@@ -128,7 +129,7 @@ func NewNode(config *cfg.Config, privValidator *types.PrivValidator, clientCreat
 	bcReactor.SetLogger(logger.With("module", "blockchain"))
 
 	// Make MempoolReactor
-	mempoolLogger := logger.With("module", "consensus")
+	mempoolLogger := logger.With("module", "mempool")
```
ebuchman: lol exactly what  I was talking about today

PR #507, opened 2017-05-16
Release v0.10.0 rc.1

ebuchman:
https://github.com/tendermint/tendermint/blob/release-v0.10.0-rc.1/CHANGELOG.md#0100-may-17-2017


types/block.go
```diff
@@ -70,18 +70,18 @@ func (b *Block) ValidateBasic(chainID string, lastBlockHeight int, lastBlockID B
...
 			return err
 		}
 	}
 	if !bytes.Equal(b.DataHash, b.Data.Hash()) {
-		return errors.New(Fmt("Wrong Block.Header.DataHash.  Expected %X, got %X", b.DataHash, b.Data.Hash()))
+		return errors.New(Fmt("Wrong Block.Header.DataHash.  Expected %v, got %v", b.DataHash, b.Data.Hash()))
 	}
 	if !bytes.Equal(b.AppHash, appHash) {
-		return errors.New(Fmt("Wrong Block.Header.AppHash.  Expected %X, got %X", appHash, b.AppHash))
+		return errors.New(Fmt("Wrong Block.Header.AppHash.  Expected %X, got %v", appHash, b.AppHash))
```
melekes: missed one %X

ebuchman: nah, `appHash` is a `[]byte` here, so needs `%X`. TBH not thrilled that I need to think about the type to decide to use %x or %v. Maybe we should just wrap the data.Bytes in []byte before printing ? or expose some `.Bytes()` method ? I much prefer to use %X where we want hex, rather than implicitly relying on the type to use hex in its string rep ...

ethanfrey: Ugh... Looking at https://golang.org/pkg/fmt/

"5. If an operand implements method String() string, that method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any)."

If I took off the `String` method from `Bytes`, then it would work with %X.

I think there was a request that I add it, I even wrote that and a test during the retreat. Don't think there is any way to deal with this better without removing the String() method.

PR #517, opened 2017-05-22
Updates to blockchain/

adrianbrink:


blockchain/pool.go
```diff
@@ -28,7 +28,7 @@ var peerTimeoutSeconds = time.Duration(15) // not const so we can override with
 	Every so often we ask peers what height they're on so we can keep going.
 
 	Requests are continuously made for blocks of higher heights until
-	the limits. If most of the requests have no available peers, and we
+	the we reach the limits. If most of the requests have no available peers, and we
```
ebuchman: the we ...

blockchain/pool.go
```diff
@@ -142,8 +140,9 @@ func (pool *BlockPool) IsCaughtUp() bool {
 		maxPeerHeight = MaxInt(maxPeerHeight, peer.height)
 	}
 
-	isCaughtUp := (height > 0 || time.Now().Sub(pool.startTime) > 5*time.Second) && (maxPeerHeight == 0 || height >= maxPeerHeight)
-	pool.Logger.Info(Fmt("IsCaughtUp: %v", isCaughtUp), "height", height, "maxPeerHeight", maxPeerHeight)
+	// ('receive at least one block' or 'wait at least 5 seconds') and ('all peers are also at 0' or 'my height is higher than the highest peer I know')
```
ebuchman: maybe better to explain the logic here rather than just putting it in boolean english like this.

Also why the pool.height? isnt it easier to read without it ?

blockchain/reactor.go
```diff
@@ -114,9 +114,8 @@ func (bcR *BlockchainReactor) GetChannels() []*p2p.ChannelDescriptor {
 
 // AddPeer implements Reactor by sending our state to peer.
 func (bcR *BlockchainReactor) AddPeer(peer *p2p.Peer) {
-	if !peer.Send(BlockchainChannel, struct{ BlockchainMessage }{&bcStatusResponseMessage{bcR.store.Height()}}) {
-		// doing nothing, will try later in `poolRoutine`
-	}
+	// doing nothing, will try later in `poolRoutine`
```
ebuchman: rather keep this in an `if` clause.  thats the context in which this comment is relevant.

blockchain/pool.go
```diff
@@ -169,17 +168,15 @@ func (pool *BlockPool) PopRequest() {
...
 	defer pool.mtx.Unlock()
 
-	if r := pool.requesters[pool.height]; r != nil {
-		/*  The block can disappear at any time, due to removePeer().
-		if r := pool.requesters[pool.height]; r == nil || r.block == nil {
-			PanicSanity("PopRequest() requires a valid block")
-		}
-		*/
+	r := pool.requesters[pool.height]
+	if r == nil {
```
ebuchman: is it really safe to uncomment this? need to review further

blockchain/pool.go
adrianbrink: Beforehand it was extracted like this ```height := pool.height``` and then height was used everywhere. Personally I think it is more readable to inline these short statements, as it means that you then don't have to refer back to the local variable to understand the meaning.

melekes:
```
receivedAtLeastOneBlockOrWaitedNSeconds := (pool.height > 0 || time.Now().Sub(pool.startTime) > 5*time.Second)
ourChainIsTheLongestOneAmoungPeers := maxPeerHeight == 0 || pool.height >= maxPeerHeight
isCaughtUp := receivedAtLeastOneBlockOrWaitedNSeconds && ourChainIsTheLongestOneAmoungPeers
```
?

blockchain/pool.go
```diff
@@ -194,7 +191,6 @@ func (pool *BlockPool) RedoRequest(height int) {
 		PanicSanity("Expected block to be non-nil")
 	}
 	// RemovePeer will redo all requesters associated with this peer.
-	// TODO: record this malfeasance
```
melekes: Have we done that?

ebuchman:
Included some of this in https://github.com/tendermint/tendermint/commit/77a3d03385c5512777b2749eaead3e243f6eab73

Issue #496, opened 2017-05-14
Deduplicate tests in rpc/tests and rpc/client


ebuchman:
Just use the ones in rpc/client, but make sure nothing is missing from rpc/tests

melekes:
https://github.com/tendermint/tendermint/pull/520

PR #520, opened 2017-05-26
deduplicate tests in rpc/test and rpc/client (Refs #496)

melekes:


rpc/test/helpers.go
```diff
@@ -66,30 +60,24 @@ func GetConfig() *cfg.Config {
...
-func GetJSONClient() *client.JSONRPCClient {
-	rpcAddr := GetConfig().RPCListenAddress
-	return client.NewJSONRPCClient(rpcAddr)
-}
+// func GetURIClient() *client.URIClient {
+// 	rpcAddr := GetConfig().RPCListenAddress
+// 	return client.NewURIClient(rpcAddr)
+// }
 
 func GetGRPCClient() core_grpc.BroadcastAPIClient {
```
melekes: What should we do with GRPC? leave it as it is?

adrianbrink:
@melekes I think this needs your attention again.

PR #521, opened 2017-05-26
updated json response to match spec by @davebryson

melekes:
Port of https://github.com/tendermint/go-rpc/pull/19

link to spec: http://www.jsonrpc.org/specification

### DIFF
--- a/rpc/lib/client/ws_client.go
+++ b/rpc/lib/client/ws_client.go
@@ -195,7 +195,7 @@ func (c *WSClient) Send(ctx context.Context, request types.RPCRequest) error {
 
 // Call the given method. See Send description.
 func (c *WSClient) Call(ctx context.Context, method string, params map[string]interface{}) error {
-	request, err := types.MapToRequest("", method, params)
+	request, err := types.MapToRequest("ws-client", method, params)
 	if err != nil {
 		return err
 	}
@@ -205,7 +205,7 @@ func (c *WSClient) Call(ctx context.Context, method string, params map[string]in
 // CallWithArrayParams the given method with params in a form of array. See
 // Send description.
 func (c *WSClient) CallWithArrayParams(ctx context.Context, method string, params []interface{}) error {
-	request, err := types.ArrayToRequest("", method, params)
+	request, err := types.ArrayToRequest("ws-client", method, params)
 	if err != nil {
 		return err
 	}
--- a/rpc/lib/client/ws_client.go
+++ b/rpc/lib/client/ws_client.go
@@ -422,8 +422,8 @@ func (c *WSClient) readRoutine() {
 			c.ErrorsCh <- err
 			continue
 		}
-		if response.Error != "" {
-			c.ErrorsCh <- errors.Errorf(response.Error)
+		if response.Error != nil {
+			c.ErrorsCh <- errors.New(response.Error.Message)
 			continue
 		}
 		c.Logger.Info("got response", "resp", response.Result)
--- a/rpc/lib/server/handlers.go
+++ b/rpc/lib/server/handlers.go
@@ -123,12 +123,8 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, logger log.Logger) http.Han
 			return
 		}
 		rpcFunc := funcMap[request.Method]
-		if rpcFunc == nil {
+		if rpcFunc == nil || rpcFunc.ws {
 			WriteRPCResponseHTTP(w, types.RPCMethodNotFoundError(request.ID))
-			return
-		}
-		if rpcFunc.ws {
-			WriteRPCResponseHTTP(w, types.RPCInternalError(request.ID, errors.New("Trying to use Websocket method in non-ws context")))
 			return
 		}
 		args, err := jsonParamsToArgsRPC(rpcFunc, request.Params)
@@ -234,7 +230,7 @@ func makeHTTPHandler(rpcFunc *RPCFunc, logger log.Logger) func(http.ResponseWrit
 	// Exception for websocket endpoints
 	if rpcFunc.ws {
 		return func(w http.ResponseWriter, r *http.Request) {
-			WriteRPCResponseHTTP(w, types.RPCInternalError("", errors.New("Trying to use Websocket method in non-ws context")))
+			WriteRPCResponseHTTP(w, types.RPCMethodNotFoundError(""))
 		}
 	}
 	// All other endpoints
--- a/rpc/lib/server/handlers.go
+++ b/rpc/lib/server/handlers.go
@@ -113,6 +113,11 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, logger log.Logger) http.Han
 			WriteRPCResponseHTTP(w, types.RPCParseError("", errors.Wrap(err, "Error unmarshalling request")))
 			return
 		}
+		// A Notification is a Request object without an "id" member.
+		// The Server MUST NOT reply to a Notification, including those that are within a batch request.
+		if request.ID == "" {
+			return
+		}
 		if len(r.URL.Path) > 1 {
 			WriteRPCResponseHTTP(w, types.RPCInvalidRequestError(request.ID, errors.Errorf("Path %s is invalid", r.URL.Path)))
 			return
@@ -510,6 +515,12 @@ func (wsc *wsConnection) readRoutine() {
 			err = json.Unmarshal(in, &request)
 			if err != nil {
 				wsc.WriteRPCResponse(types.RPCParseError("", errors.Wrap(err, "Error unmarshaling request")))
+				continue
+			}
+
+			// A Notification is a Request object without an "id" member.
+			// The Server MUST NOT reply to a Notification, including those that are within a batch request.
+			if request.ID == "" {
 				continue
 			}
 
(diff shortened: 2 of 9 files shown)


rpc/lib/server/handlers.go
```diff
@@ -109,35 +109,35 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, logger log.Logger) http.Han
 		var request types.RPCRequest
 		err := json.Unmarshal(b, &request)
 		if err != nil {
-			WriteRPCResponseHTTPError(w, http.StatusBadRequest, types.NewRPCResponse("", nil, fmt.Sprintf("Error unmarshalling request: %v", err.Error())))
+			WriteRPCResponseHTTP(w, types.RPCParseError(""))
```
melekes: I don't like we are loosing error messages. Should we log them instead or expose one more field (like `additional` or something)?

melekes: @davebryson what do you think?

davebryson: The Error response provides a base message corresponding to the error code.  In my first version of this code, I had included the "data" field in the error that could have been used to add additional info - but I removed it based on earlier feedback.    Logging will probably work just fine.

melekes: I have put data field back https://github.com/tendermint/tendermint/pull/521/commits/d75c1e44b89d85cc94fb0fa428f69383632af3b5

melekes:
Looks good now.

rpc/lib/client/ws_client.go
```diff
@@ -111,8 +111,8 @@ func (wsc *WSClient) receiveEventsRoutine() {
 				wsc.ErrorsCh <- err
 				continue
 			}
-			if response.Error != "" {
-				wsc.ErrorsCh <- errors.Errorf(response.Error)
+			if response.Error != nil && response.Error.Message != "" {
```
odeke-em: So what happens to all the other, response.Error's that have empty messages?
Will they all now just fall through and be treated as success?

Perhaps we can tweak this a little e.g
```go
if response.Error != nil {
   switch response.Error.Message {
   case "":
       wsc.ErrorsCh <- errors.Errorf(response.Error)
   default:
       wsc.ErrorsCh <- errors.Errorf(response.Error.Message)
   }
}
```
that way we can catch all errors.
PS: I obviously don't have much context about the code base since I noticed the change from a string to nullable type :)

rpc/lib/types/types.go
```diff
@@ -42,34 +47,69 @@ func ArrayToRequest(id string, method string, params []interface{}) (RPCRequest,
...
+}
 
 type RPCResponse struct {
 	JSONRPC string           `json:"jsonrpc"`
-	ID      string           `json:"id"`
-	Result  *json.RawMessage `json:"result"`
-	Error   string           `json:"error"`
+	ID      string           `json:"id,omitempty"`
+	Result  *json.RawMessage `json:"result,omitempty"`
+	Error   *RpcError        `json:"error,omitempty"`
```
odeke-em: Just a random question: does this change break compatibility for folks consuming types/types?

melekes: yes

rpc/lib/client/ws_client.go
melekes: Thanks. Actually, the error message should always be present. I removed the checks.

melekes:
@ebuchman can we merge this?

rpc/lib/types/types.go
```diff
@@ -42,34 +47,69 @@ func ArrayToRequest(id string, method string, params []interface{}) (RPCRequest,
...
+		Error:   &RPCError{Code: code, Message: msg, Data: data},
 	}
 }
 
+func RPCParseError(id string, err error) RPCResponse {
+	return NewRPCErrorResponse(id, -32700, "Parse error. Invalid JSON", err.Error())
+}
+
+func RPCInvalidRequestError(id string, err error) RPCResponse {
+	return NewRPCErrorResponse(id, -32600, "Invalid Request", err.Error())
```
odeke-em: Just a drive-by question, how come the negative status codes?

davebryson: They are from the JSON-RPC spec

odeke-em: Thank you @davebryson, after your comment I then referred to them here http://www.jsonrpc.org/specification#response_object.

melekes:
LGTM

rpc/lib/server/handlers.go
```diff
@@ -228,25 +228,25 @@ func makeHTTPHandler(rpcFunc *RPCFunc, logger log.Logger) func(http.ResponseWrit
 	// Exception for websocket endpoints
 	if rpcFunc.ws {
 		return func(w http.ResponseWriter, r *http.Request) {
-			WriteRPCResponseHTTPError(w, http.StatusMethodNotAllowed, types.NewRPCResponse("", nil, "This RPC method is only for websockets"))
+			WriteRPCResponseHTTP(w, types.RPCInternalError("", errors.New("Trying to use Websocket method in non-ws context")))
```
ebuchman: should we be doing JSONRPC error codes for the makeHTTPHandler? yes the response is JSON, but since the request wasn't done via JSONRPC, perhaps we should be returning http error codes ? ie. we only follow the JSONRPC spec when the request is via JSONRPC

rpc/lib/types/types.go
```diff
@@ -42,34 +47,69 @@ func ArrayToRequest(id string, method string, params []interface{}) (RPCRequest,
...
+	Message string `json:"message"`
+	Data    string `json:"data,omitempty"`
+}
 
 type RPCResponse struct {
 	JSONRPC string           `json:"jsonrpc"`
-	ID      string           `json:"id"`
-	Result  *json.RawMessage `json:"result"`
-	Error   string           `json:"error"`
+	ID      string           `json:"id,omitempty"`
```
ebuchman: from the spec:

```
This member is REQUIRED.
It MUST be the same as the value of the id member in the Request Object.
If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it MUST be Null.
```

I think this means we can't use `omitempty` and that we need to use a pointer for the ID so that it can be encoded properly as `null` rather than empty string ...

rpc/lib/server/handlers.go
```diff
@@ -109,35 +109,35 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, logger log.Logger) http.Han
 		var request types.RPCRequest
 		err := json.Unmarshal(b, &request)
 		if err != nil {
-			WriteRPCResponseHTTPError(w, http.StatusBadRequest, types.NewRPCResponse("", nil, fmt.Sprintf("Error unmarshalling request: %v", err.Error())))
+			WriteRPCResponseHTTP(w, types.RPCParseError("", errors.Wrap(err, "Error unmarshalling request")))
 			return
 		}
```
ebuchman: from spec:

```
A Notification is a Request object without an "id" member.
...
 The Server MUST NOT reply to a Notification, including those that are within a batch request.
```

So I think we need to check the request.ID here and not return anything if it's empty

ebuchman:
Oh, I'd also love for @davebryson to get the contributing credits on the first commit here which would otherwise be lost because original PR is in deprecated repo. Can we do some git magic-foo to make that happen ?

melekes:
@ebuchman of course, we can. see  https://github.com/jayphelps/git-blame-someone-else. we can even put Linus Torvalds email if we want to :)

adrianbrink:
And now we know why you should sign your commits. Even though if I remember correctly Linus had some strong opinions about that one too.

odeke-em:
How's it going @melekes? Gentle ping here, as this PR has some merge conflicts and perhaps if we could get it merged in, we'd have our JSON responses matching the spec.

melekes:
I am gonna work on this now.

rpc/lib/server/handlers.go
melekes: > but since the request wasn't done via JSONRPC

yes, but the format in use is JSONRPC (i.e. both request and response objects contain `jsonrpc` field). I think it is ok to leave the error codes

rpc/lib/server/handlers.go
melekes: fixed

rpc/lib/server/handlers.go
```diff
@@ -109,35 +109,35 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, logger log.Logger) http.Han
...
 		rpcFunc := funcMap[request.Method]
 		if rpcFunc == nil {
-			WriteRPCResponseHTTPError(w, http.StatusNotFound, types.NewRPCResponse(request.ID, nil, "RPC method unknown: "+request.Method))
+			WriteRPCResponseHTTP(w, types.RPCMethodNotFoundError(request.ID))
 			return
 		}
 		if rpcFunc.ws {
-			WriteRPCResponseHTTPError(w, http.StatusMethodNotAllowed, types.NewRPCResponse(request.ID, nil, "RPC method is only for websockets: "+request.Method))
+			WriteRPCResponseHTTP(w, types.RPCInternalError(request.ID, errors.New("Trying to use Websocket method in non-ws context")))
```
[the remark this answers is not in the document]
melekes: yes

rpc/lib/types/types.go
melekes: fixed

melekes:
> of course, we can. see https://github.com/jayphelps/git-blame-someone-else. we can even put Linus Torvalds email if we want to :)

this tool does not work properly :(

rpc/lib/server/handlers.go
ebuchman: Ok. Just to re-iterate, it means it will return 200, and the caller has to check the json for the error. I guess it's fine

melekes:
@ebuchman can we merge it now?

PR #522, opened 2017-05-30
core: apply megacheck vet tool (unused, gosimple, staticcheck)

zramsay:
- this PR applies the `megacheck` tool, which itself applies the `unused`, `gosimple`, and `staticcheck` golang code vetting/improvement tools. See https://github.com/dominikh/go-tools for more information.
- a few fixes caught by the tool were left as-is due to insufficient context for what the best fix ought to be. These will be addressed in a follow-up PR after discussion with @ebuchman

### DIFF
--- a/blockchain/reactor.go
+++ b/blockchain/reactor.go
@@ -19,7 +19,6 @@ const (
 	BlockchainChannel = byte(0x40)
 
 	defaultChannelCapacity = 100
-	defaultSleepIntervalMS = 500
 	trySyncIntervalMS      = 100
 	// stop syncing when last block's time is
 	// within this much of the system time.
@@ -49,7 +48,6 @@ type BlockchainReactor struct {
 	fastSync     bool
 	requestsCh   chan BlockRequest
 	timeoutsCh   chan string
-	lastBlock    *types.Block
 
 	evsw types.EventSwitch
 }
@@ -134,6 +132,8 @@ func (bcR *BlockchainReactor) Receive(chID byte, src *p2p.Peer, msgBytes []byte)
 
 	bcR.Logger.Debug("Receive", "src", src, "chID", chID, "msg", msg)
 
+	// [zr] note: the `megacheck` tool (go vet on steroids)
+	// really dislikes the logic in this switch
 	switch msg := msg.(type) {
 	case *bcBlockRequestMessage:
 		// Got a request for a block. Respond with block if we have it.
@@ -194,10 +194,10 @@ FOR_LOOP:
 			if peer != nil {
 				bcR.Switch.StopPeerForError(peer, errors.New("BlockchainReactor Timeout"))
 			}
-		case _ = <-statusUpdateTicker.C:
+		case <-statusUpdateTicker.C:
 			// ask for status updates
 			go bcR.BroadcastStatusRequest()
-		case _ = <-switchToConsensusTicker.C:
+		case <-switchToConsensusTicker.C:
 			height, numPending, _ := bcR.pool.GetStatus()
 			outbound, inbound, _ := bcR.Switch.NumPeers()
 			bcR.Logger.Info("Consensus ticker", "numPending", numPending, "total", len(bcR.pool.requesters),
@@ -211,7 +211,7 @@ FOR_LOOP:
 
 				break FOR_LOOP
 			}
-		case _ = <-trySyncTicker.C: // chan time
+		case <-trySyncTicker.C: // chan time
 			// This loop can be slow as long as it's doing syncing work.
 		SYNC_LOOP:
 			for i := 0; i < 10; i++ {
--- a/consensus/replay.go
+++ b/consensus/replay.go
@@ -104,6 +104,9 @@ func (cs *ConsensusState) catchupReplay(csHeight int) error {
 	// NOTE: This is just a sanity check. As far as we know things work fine without it,
 	// and Handshake could reuse ConsensusState if it weren't for this check (since we can crash after writing ENDHEIGHT).
 	gr, found, err := cs.wal.group.Search("#ENDHEIGHT: ", makeHeightSearchFunc(csHeight))
+	if err != nil {
+		return err
+	}
 	if gr != nil {
 		gr.Close()
 	}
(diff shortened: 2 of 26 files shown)


blockchain/reactor.go
```diff
@@ -134,6 +132,8 @@ func (bcR *BlockchainReactor) Receive(chID byte, src *p2p.Peer, msgBytes []byte)
 
 	bcR.Logger.Debug("Receive", "src", src, "chID", chID, "msg", msg)
 
+	// [zr] note: the `megacheck` tool (go vet on steroids)
+	// really dislikes the logic in this switch
```
ebuchman: note

consensus/replay.go
```diff
@@ -132,7 +135,7 @@ func (cs *ConsensusState) catchupReplay(csHeight int) error {
 	if !found {
 		// if we upgraded from 0.9 to 0.9.1, we may have #HEIGHT instead
 		// TODO (0.10.0): remove this
```
ebuchman: note

consensus/state.go
```diff
@@ -648,10 +648,10 @@ func (cs *ConsensusState) handleMsg(mi msgInfo, rs RoundState) {
 	case *VoteMessage:
 		// attempt to add the vote and dupeout the validator if its a duplicate signature
 		// if the vote gives us a 2/3-any or 2/3-one, we transition
-		err := cs.tryAddVote(msg.Vote, peerKey)
-		if err == ErrAddingVote {
-			// TODO: punish peer
-		}
+		_ = cs.tryAddVote(msg.Vote, peerKey)
```
ebuchman: my preference would probably be to leave this kind of dead code rather than comment but fine to just go with the tool

consensus/state_test.go
```diff
@@ -523,7 +523,10 @@ func TestLockPOLRelock(t *testing.T) {
 	<-voteCh // prevote
 
 	signAddVotes(cs1, types.VoteTypePrevote, cs1.ProposalBlock.Hash(), cs1.ProposalBlockParts.Header(), vs2, vs3, vs4)
-	_, _, _ = <-voteCh, <-voteCh, <-voteCh // prevotes
+	// prevotes
+	<-voteCh
```
ebuchman: oh common! :P

state/execution.go
```diff
@@ -155,18 +155,6 @@ func updateValidators(validators *types.ValidatorSet, changedValidators []*abci.
 	return nil
 }
 
-// return a bit array of validators that signed the last commit
```
ebuchman: we're gonna use this block so shouldnt delete it

types/block.go
```diff
@@ -282,10 +282,7 @@ func (commit *Commit) GetByIndex(index int) *Vote {
 }
 
 func (commit *Commit) IsCommit() bool {
-	if len(commit.Precommits) == 0 {
-		return false
-	}
-	return true
+	return len(commit.Precommits) == 0
```
ebuchman: !=

p2p/upnp/probe.go
```diff
@@ -18,26 +17,26 @@ type UPNPCapabilities struct {
 func makeUPNPListener(intPort int, extPort int, logger log.Logger) (NAT, net.Listener, net.IP, error) {
 	nat, err := Discover()
 	if err != nil {
-		return nil, nil, nil, errors.New(fmt.Sprintf("NAT upnp could not be discovered: %v", err))
+		return nil, nil, nil, fmt.Errorf("NAT upnp could not be discovered: %v", err)
```
melekes: should be `errors.Errorf`

p2p/upnp/probe.go
```diff
@@ -18,26 +17,26 @@ type UPNPCapabilities struct {
...
 	}
 	logger.Info(cmn.Fmt("ourIP: %v", nat.(*upnpNAT).ourIP))
 
 	ext, err := nat.GetExternalAddress()
 	if err != nil {
-		return nat, nil, nil, errors.New(fmt.Sprintf("External address error: %v", err))
+		return nat, nil, nil, fmt.Errorf("External address error: %v", err)
```
melekes: the same

consensus/state_test.go
zramsay: lol yeye I know ... was just going with what the tool recommended

types/block.go
zramsay: touche. that's probably why tests failed

p2p/upnp/probe.go
zramsay: `errors.Errorf` throws as undefined. The official documentation (of the `errors` package) recommends using `fmt.Errorf` for errors that need formatting

blockchain/reactor.go
zramsay: this is also the reason I wouldn't add `megacheck` to Makefile _yet_, because some (e.g., this one) code blocks were left as-is since I don't know enough about the codebase to refactor-to-not-fail-the-megacheck

p2p/upnp/probe.go
melekes: https://github.com/pkg/errors we are using this

consensus/state.go
zramsay: sort of agree, but as above, then `megacheck` would fail if added to Makefile

state/execution.go
zramsay: returned it but commented out the function

consensus/state.go
zramsay: would need to comment out all dead code in order to add `megacheck` to Makefile

ebuchman: dont add it to the normal `make install`. just a separate command that we can run at our leisure for now ...

p2p/upnp/probe.go
zramsay: got it. fixed

consensus/state.go
zramsay: done

zramsay:
megacheck added to Makefile & comments addresses. `TestWALCrashAfterWrite` fails though, so I'll have to dig deeper to figure out why

adrianbrink:
@zramsay Please reach out for help if you still have problems with the tests failing. I struggled with them for a while, especially the WAL.

PR #530, opened 2017-06-12
Enable VerifyCommitAny

adrianbrink:


blockchain/pool.go
```diff
@@ -452,9 +453,9 @@ func (bpr *bpRequester) getPeerID() string {
 
 func (bpr *bpRequester) reset() {
 	bpr.mtx.Lock()
+	defer bpr.mtx.Unlock()
```
odeke-em: Really minor comment that can be ignored: so at times, while defer can be more
convenient to read, it can be slower but idk if this matters

blockchain/pool.go
```diff
@@ -105,6 +105,7 @@ func (pool *BlockPool) removeTimedoutPeers() {
 		if !peer.didTimeout && peer.numPending > 0 {
 			curRate := peer.recvMonitor.Status().CurRate
 			// XXX remove curRate != 0
+			// curRate can be safely removed, since it will only be 0 if the peer is offline, but then the 2nd condition would also be true
```
odeke-em: Could we add line breaks for easier reading?
So:
```go
// curRate can be safely removed, since it will only be 0 if the
// peer is offline, but then the 2nd condition would also be true.
if curRate != 0 && curRate < minRecvRate {
	pool.sendTimeout(peer.id)
	pool.Logger.Error("SendTimeout", "peer", peer.id, "reason", "curRate too low")
	peer.didTimeout = true
}
```
Thanks.

types/validator_set.go
```diff
@@ -264,29 +264,69 @@ func (valSet *ValidatorSet) VerifyCommit(chainID string, blockID BlockID, height
...
+		}
+		curVotingPower += val.VotingPower
+
+		// check that the validator from the new set has signed the precommit
+		_, newVal := newVals.GetByIndex(idx)
+		if newVal.PubKey.Equals(val.PubKey) {
+			newVotingPower += newVal.VotingPower
+		}
+
+		if curVotingPower <= valSet.TotalVotingPower()*2/3 {
```
ebuchman: this should probably be outside the for loop

ebuchman:
Still needs tests

types/validator_set.go
```diff
@@ -263,30 +263,73 @@ func (valSet *ValidatorSet) VerifyCommit(chainID string, blockID BlockID, height
...
+			return fmt.Errorf("Invalid commit -- wrong height: %v vs %v", height, precommit.Height)
+		}
+		if precommit.Round != round {
+			return fmt.Errorf("Invalid commit -- wrong round: %v vs %v", round, precommit.Round)
+		}
+		if precommit.Type != VoteTypePrecommit {
+			return fmt.Errorf("Invalid commit -- not precommit @ index %v", idx)
+		}
+		if !blockID.Equals(precommit.BlockID) {
+			continue // Not an error, but doesn't count
```
odeke-em: Perhaps we should log.Info this event?

types/validator_set.go
```diff
@@ -263,30 +263,73 @@ func (valSet *ValidatorSet) VerifyCommit(chainID string, blockID BlockID, height
...
+
+		// check that the validator from the new set has signed the precommit
+		_, newVal := newVals.GetByIndex(idx)
+		if newVal.PubKey.Equals(val.PubKey) {
+			newVotingPower += newVal.VotingPower
+		}
+
+	}
+
+	if curVotingPower <= valSet.TotalVotingPower()*2/3 {
```
odeke-em:
```
var minValSetVotingPower := 1 + (valSet.TotalVotingPower() * 2/3)
var minNewVotingPower := 1 + (newVals.TotalVotingPower() * 2/3)

if curVotingPower < minValSetVotingPower {
   return fmt.Errorf("Invalid commit -- insufficient current voting power: got %v, needed %v", curVotingPower, minValSetVotingPower)
} else if newVotingPower < minNewVotingPower {
   return fmt.Errorf("Invalid commit -- insufficient new voting power: got %v, needed %v", newVotingPower, minNewVotingPower)
}
```

types/validator_set_test.go
```diff
@@ -4,29 +4,93 @@ import (
...
+		{keys, vals, 3, 0, len(keys) - 1, false},
+	}
+
+	for _, tc := range cases {
+		commit, header := tc.keys.genCommit(chainID, tc.height, nil, tc.vals,
+			[]byte("bar"), tc.first, tc.last)
+		blockID := BlockID{Hash: header.Hash()}
+		err := tc.vals.VerifyCommit(chainID, blockID, tc.height, commit)
+		if tc.proper {
+			assert.Nil(err, "%+v", err)
```
odeke-em: And perhaps the index ie
```go
for i, tc := range cases {
   ...
   if tc.proper {
      assert.Nil(err, "#%d: %v", i, err)
   } else {
      assert.NotNil(err, "#%d", i)
   }
}
```

blockchain/pool.go
```diff
@@ -32,6 +33,7 @@ var peerTimeoutSeconds = time.Duration(15) // not const so we can override with
 	are not at peer limits, we can probably switch to consensus reactor
 */
 
+// BlockPool ...
```
odeke-em: Could you please add proper docs, as you have done for the other new functions?

blockchain/pool.go
```diff
@@ -48,6 +50,7 @@ type BlockPool struct {
 	timeoutsCh chan<- string
 }
 
+// NewBlockPool ...
```
odeke-em: ditto about adding comments.

blockchain/pool.go
```diff
@@ -63,12 +66,14 @@ func NewBlockPool(start int, requestsCh chan<- BlockRequest, timeoutsCh chan<- s
 	return bp
 }
 
+// OnStart ...
```
odeke-em: ditto for proper comments.

blockchain/pool.go
```diff
@@ -63,12 +66,14 @@ func NewBlockPool(start int, requestsCh chan<- BlockRequest, timeoutsCh chan<- s
...
 func (pool *BlockPool) OnStart() error {
 	go pool.makeRequestersRoutine()
 	pool.startTime = time.Now()
 	return nil
 }
 
+// OnStop ...
```
odeke-em: ditto about proper comments.

blockchain/pool.go
```diff
@@ -117,13 +124,15 @@ func (pool *BlockPool) removeTimedoutPeers() {
 	}
 }
 
+// GetStatus gets the height, number of pending blocks and number of peers.
```
odeke-em: Awesome, thanks!

blockchain/pool.go
```diff
@@ -144,12 +153,13 @@ func (pool *BlockPool) IsCaughtUp() bool {
 	receivedBlockOrTimedOut := (pool.height > 0 || time.Since(pool.startTime) > 5*time.Second)
 	ourChainIsLongestAmongPeers := maxPeerHeight == 0 || pool.height >= maxPeerHeight
 	isCaughtUp := receivedBlockOrTimedOut && ourChainIsLongestAmongPeers
-	pool.Logger.Info(Fmt("IsCaughtUp: %v", isCaughtUp), "height", pool.height, "maxPeerHeight", maxPeerHeight)
+	pool.Logger.Info(Fmt("IsCaughtUp: %v", isCaughtUp), "height", pool.height,
```
odeke-em: Am assuming this is because of proper line breaks when reading on your editor ;)

types/validator_set.go
```diff
@@ -263,30 +263,73 @@ func (valSet *ValidatorSet) VerifyCommit(chainID string, blockID BlockID, height
...
-				}
-			}
-	*/
+// VerifyCommitAny verifies that +2/3 of this set had signed the given signBytes.
+// Unlike VerifyCommit(), this function can verify commits with different sets.
+// VerifyCommitAny verifies that more than 2/3 of the old validator set have
+// signed this block, as well as whether more than 2/3 of the new validator set
+// have signed this block.
+func (valSet *ValidatorSet) VerifyCommitAny(newVals *ValidatorSet, chainID string, blockID BlockID, height int, commit *Commit) error {
+	if newVals.Size() != len(commit.Precommits) {
```
odeke-em:
```go
if g, w := newVals.Size(), len(commits.Precommits); g != w {
   return fmt.Errorf("Invalid commit -- wrong set size: %v vs %v", g, w)
}
```

types/validator_set.go
```diff
@@ -263,30 +263,73 @@ func (valSet *ValidatorSet) VerifyCommit(chainID string, blockID BlockID, height
...
+		return fmt.Errorf("Invalid commit -- wrong set size: %v vs %v", newVals.Size(), len(commit.Precommits))
+	}
+	if height != commit.Height() {
```
odeke-em: Same drill here
```go
if g, w := height, commit.Height(); g != w {
   return fmt.Errorf("Invalid commit -- wrong height: %v vs %v", g, w)
}
```

types/validator_set.go
```diff
@@ -263,30 +263,73 @@ func (valSet *ValidatorSet) VerifyCommit(chainID string, blockID BlockID, height
...
+	newVotingPower := int64(0)
+	seen := map[int]bool{}
+	round := commit.Round()
+
+	for idx, precommit := range commit.Precommits {
+		// may be nil if validator skipped.
+		if precommit == nil {
+			continue
+		}
+		if precommit.Height != height {
```
odeke-em: Same paradigm about `if g, w := gotVal, wantVal; g != w { return fmt.Errorf("...%v vs %v", g, w)` as I mentioned above.

types/validator_set.go
```diff
@@ -263,30 +263,73 @@ func (valSet *ValidatorSet) VerifyCommit(chainID string, blockID BlockID, height
...
+			return fmt.Errorf("Invalid commit -- not precommit @ index %v", idx)
+		}
+		if !blockID.Equals(precommit.BlockID) {
+			continue // Not an error, but doesn't count
+		}
+
+		// check whether a current validator has signed this precommit
+		valIdx, val := valSet.GetByAddress(precommit.ValidatorAddress)
+		// if val is nill then the current validator hasn't signed the precommit
+		if val == nil || seen[valIdx] {
```
odeke-em: Same drill here, perhaps we should log.Debugf this event.

ethanfrey:
Why did you build this again?
All this is in the light client since many months. 
Had actually trimmed down the light client on develop to just what belongs in tendermint 

Look at certifiers/dynamic.go that does this also

blockchain/pool.go
adrianbrink: Yes, I like the 100 line limit :-) Allows me to have two emacs windows on a 13" screen.

ebuchman:
Indeed it is! https://github.com/tendermint/light-client/blob/develop/certifiers/dynamic.go#L86
Thanks @ethanfrey 

@adrianbrink , do you want to remove that part from this PR and just change it to some small things/comments?

I believe the plan is to bring lightclient in here eventually. Perhaps the tests in this PR can in the meantime be added there.

adrianbrink:
The purpose of this PR is to move the validation logic from the light client into Tendermint core.

ebuchman:
Ok - I believe the entire light client is coming in, so no need for this. We'll wait for the whole thing to move.

PR #538, opened 2017-06-19
rework docs

zramsay:
moves a couple files into the repo from https://github.com/tendermint/tendermint.github.io/pull/68


CONTRIBUTING.md
```diff
@@ -1,16 +1,90 @@
-# Contributing guidelines
+# Contributing
 
-**Thanks for considering making contributions to Tendermint!**
+Thanks for considering making contributions to Tendermint and related repositories (Basecoin, Merkleeyes, etc.)!
```
adrianbrink: *Thank you ...

ebuchman:
Looks good. Let's bring all the docs into this repo. We need to upgrade the website to read them directly from here.

I think ultimately we're going to have independent sites for Tendermint, ABCI, Basecoin, etc. While we don't really have the resources to move on that yet, we can certainly get the docs in order with that in mind.

So lets bring all the "Internals" docs in here tool, as well as the "Install" and "Using Tendermint" bit. The ABCI/App stuff can go in `github.com/tendermint/abci`.

PR #539, opened 2017-06-19
Add Code of Conduct

adrianbrink:


melekes:
Are we gonna put it in every repo?

adrianbrink:
I don't know, but I think it can't hurt. I've put it in ethermint and rust-abci.

ebuchman:
I think we should have a central source of truth for these things. Maybe put them all in the coding repo and the md in all the other repos can just point there, and contain any exceptions for that particular repo

adrianbrink:
The code of conduct is very fundamental and shouldn't have any exceptions. I think from a central point would be good as well, but it seemed easier to just add it to the repos. :-)

PR #549, opened 2017-06-21
RPC server ADR

melekes:


zramsay:
should be `docs/hidden` as of #538

Issue #551, opened 2017-06-22
addresses should default to tcp


ebuchman:

ethanfrey:
This is fixed by #552, right?  Or do we only close issues when they merge into master?

If so, we should have a flag "fixed on develop"

adrianbrink:
I think once it is merged into develop we should close the issues.

PR #552, opened 2017-06-22
httpDialer accepts no prefix or http:// as tcp://

ethanfrey:
Also, no longer panics but instead return an error.

### DIFF
--- a/rpc/lib/client/http_client.go
+++ b/rpc/lib/client/http_client.go
@@ -13,7 +13,6 @@ import (
 
 	"github.com/pkg/errors"
 	types "github.com/tendermint/tendermint/rpc/lib/types"
-	cmn "github.com/tendermint/tmlibs/common"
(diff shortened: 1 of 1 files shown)


ethanfrey:
This fixes issue #551

rpc/lib/client/http_client.go
```diff
@@ -23,13 +22,23 @@ type HTTPClient interface {
...
 		protocol, address = parts[0], parts[1]
+	} else {
+		// return a invalid message
+		msg := fmt.Sprintf("Invalid addr: %s", remoteAddr)
+		return msg, func(_ string, _ string) (net.Conn, error) {
+			return nil, errors.New(msg)
+		}
+	}
+	// accept http as an alias for tcp
+	if protocol == "http" {
```
melekes: 👍

PR #555, opened 2017-06-23
Import docs

keppel:
These are the docs from tendermint.com, which should live here now so that there's a single source of truth. The website doc build script downloads these now (from master), and once these make it to master I'll set up a webhook to sync the website whenever there's a change here.

I also moved the [Architecture Decision Records](https://product.reverb.com/documenting-architecture-decisions-the-reverb-way-a3563bb24bd0#.78xhdix6t) stuff to the `hidden` directory, which the website will ignore.


zramsay:
duplicate of #538 which is part of https://github.com/tendermint/tendermint.github.io/pull/68

ebuchman:
superseded by https://github.com/tendermint/tendermint/pull/538

PR #560, opened 2017-06-26
Docker build automation

melekes:
Refs https://github.com/tendermint/tools/issues/35

### DIFF
--- a/DOCKER/build.sh
+++ b/DOCKER/build.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+set -e
+
+# Get the tag from the version, or try to figure it out.
+if [ -z "$TAG" ]; then
+	TAG=$(awk -F\" '/Version =/ { print $2; exit }' < ../version/version.go)
+fi
+if [ -z "$TAG" ]; then
+		echo "Please specify a tag."
+		exit 1
+fi
+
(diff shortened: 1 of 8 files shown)


DOCKER/build.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+set -e
+
+# Get the tag from the version, or try to figure it out.
+if [ -z "$TAG" ]; then
+	TAG=$(awk -F\" '/Version =/ { print $2; exit }' < ../version/version.go)
```
ebuchman: Oy. Cant we just assume tendermint is built already and run `tendermint version` ?

melekes: We can safely assume the version is new, but the binary - not so sure.

adrianbrink: @melekes is right. You are able to be on 0.9.1, run `make install` which places the 0.9.1 binary in your $PATH. Afterwards you switch to 0.10.0 and run `make docker` and you now have the wrong version number.

The problem with using `tendermint version` is that it relies on information from outside the git repo (the binary in the $PATH).

I use this approach: https://github.com/tendermint/ethermint/blob/master/Makefile#L41
https://github.com/tendermint/ethermint/blob/master/scripts/docker/Dockerfile

PR #568, opened 2017-06-29
replacing events package with a new pubsub package

melekes:
DEPENDS ON:
https://github.com/tendermint/tmlibs/pull/32
https://github.com/tendermint/tmlibs/pull/31


melekes:
Two most important files:
https://github.com/tendermint/tendermint/blob/ec6341215e61d04892c263318a8d06313c3f6968/types/pubsub.go
https://github.com/tendermint/tendermint/blob/ec6341215e61d04892c263318a8d06313c3f6968/rpc/core/events.go

blockchain/reactor.go
```diff
@@ -49,7 +49,7 @@ type BlockchainReactor struct {
 	requestsCh   chan BlockRequest
 	timeoutsCh   chan string
 
-	evsw types.EventSwitch
+	pubsub types.EventsPublisher
```
melekes: maybe we should preserve the old naming?

consensus/byzantine_test.go
```diff
@@ -63,17 +64,15 @@ func TestByzantine(t *testing.T) {
...
+		pubsub := pubsub.NewServer(1)
+		pubsub.SetLogger(logger.With("module", "pubsub", "validator", i))
+		_, err := pubsub.Start()
+		assert.NoError(t, err)
+		eventChans[i] = pubsub.Subscribe(types.EventQueryNewBlock)
 
 		conR := NewConsensusReactor(css[i], true) // so we dont start the consensus states
 		conR.SetLogger(logger.With("validator", i))
-		conR.SetEventSwitch(eventSwitch)
+		conR.SetEventsPubsub(pubsub)
```
melekes: again, preserve old interface?

consensus/event_buffer.go
```diff
@@ -0,0 +1,45 @@
...
+
+const (
+	eventBufferCapacity = 1000
+)
+
+// Interface assertions
+var _ types.EventsPublisher = (*EventBuffer)(nil)
+
+// EventBuffer is a buffer of events.
+type EventBuffer struct {
```
melekes: should this eventBuffer be a part of `pubsub` package? or `events`?

node/node.go
```diff
@@ -29,10 +29,15 @@ import (
 	cmn "github.com/tendermint/tmlibs/common"
 	dbm "github.com/tendermint/tmlibs/db"
 	"github.com/tendermint/tmlibs/log"
+	"github.com/tendermint/tmlibs/pubsub"
 
 	_ "net/http/pprof"
 )
 
+const (
+	maxQueuedEventsPerSubscriber = 1000
```
melekes: benchmark

rpc/core/events.go
```diff
@@ -1,24 +1,52 @@
...
+func Unsubscribe(wsCtx rpctypes.WSRPCContext, queryStr string) (*ctypes.ResultUnsubscribe, error) {
+	logger.Info("Unsubscribe from query", "remote", wsCtx.GetRemoteAddr(), "query", queryStr)
+	ch := wsCtx.DeleteSubscription(queryStr)
+	if ch != nil {
+		eventsSub.Unsubscribe(ch)
+	}
+	return &ctypes.ResultUnsubscribe{}, nil
+}
+
+func UnsubscribeAll(wsCtx rpctypes.WSRPCContext) (*ctypes.ResultUnsubscribe, error) {
```
melekes: add a test

rpc/lib/server/handlers.go
```diff
@@ -356,17 +355,20 @@ type wsConnection struct {
 	pingTicker  *time.Ticker
 
 	funcMap map[string]*RPCFunc
-	evsw    events.EventSwitch
+
+	// pubsub
+	pubsubChannels map[string]chan interface{} // user subscriptions
```
melekes: should this be a part of the pubsub package?

```go
// pubsub/client.go
Client struct {
  subscriptions map[Query]chan interface{}
}

// pubsub/server.go
Server {
  queries[Client]Query
}
```

rpc/lib/types/types.go
```diff
@@ -74,10 +72,11 @@ func NewRPCResponse(id string, res interface{}, err string) RPCResponse {
 
 // *wsConnection implements this interface.
 type WSRPCConnection interface {
+  IsRunning() bool
```
melekes: reformat

types/pubsub.go
```diff
@@ -0,0 +1,124 @@
...
+	fireEvent(p, EventStringRelock(), TMEventData{rs})
+}
+
+func FireEventLock(p EventsPublisher, rs EventDataRoundState) {
+	fireEvent(p, EventStringLock(), TMEventData{rs})
+}
+
+// All events should be based on this FireEvent to ensure they are TMEventData
+func fireEvent(p EventsPublisher, event string, data TMEventData) {
+	if p != nil {
```
melekes: document why this is needed (`consensus/state may fire event before we set pubsub on it. see consensus/state.go`)

rpc/client/localclient.go
```diff
@@ -108,3 +113,40 @@ func (c Local) Validators() (*ctypes.ResultValidators, error) {
 func (c Local) Tx(hash []byte, prove bool) (*ctypes.ResultTx, error) {
 	return core.Tx(hash, prove)
 }
+
+func (c Local) Subscribe(query string, out chan<- types.TMEventData) error {
```
melekes: iterate on this interface. is this the best interface? maybe callback style is better here? https://godoc.org/cloud.google.com/go/pubsub#Subscription.Receive

Also, pass context!

rpc/client/localclient.go
```diff
@@ -1,11 +1,14 @@
...
 
 import (
+	"github.com/pkg/errors"
+
 	data "github.com/tendermint/go-wire/data"
 	nm "github.com/tendermint/tendermint/node"
 	"github.com/tendermint/tendermint/rpc/core"
 	ctypes "github.com/tendermint/tendermint/rpc/core/types"
 	"github.com/tendermint/tendermint/types"
+	pquery "github.com/tendermint/tmlibs/pubsub/query"
```
melekes: `tmquery`

rpc/client/httpclient.go
```diff
@@ -342,22 +305,17 @@ func (w *WSEvents) parseEvent(data []byte) (err error) {
...
 	}
+	return nil
 }
 
-func (w *WSEvents) unsubscribe(event string) {
-	err := w.ws.Unsubscribe(event)
-	if err != nil {
-		panic(err)
+func extractEventType(query string) string {
+	re := regexp.MustCompile(types.EventKey + "=" + "([^\\s]+)")
```
melekes: this is very bad

rpc/client/helpers.go
```diff
@@ -56,33 +54,21 @@ func WaitForHeight(c StatusClient, h int, waiter Waiter) error {
...
-		quit <- true
-	}()
+func WaitForOneEvent(c EventsClient, evtTyp string, timeout time.Duration) (types.TMEventData, error) {
+	evts := make(chan types.TMEventData, 1)
 
 	// register for the next event of this type
-	evsw.AddListenerForEvent(listener, evtTyp, func(data events.EventData) {
-		evts <- data
-	})
+	err := c.Subscribe(types.EventKey+"="+evtTyp, evts)
```
melekes: extract var

blockchain/reactor.go
ethanfrey: i never did like evsw.
pubsub is okay.  open to other ideas as well.

consensus/event_buffer.go
ethanfrey: I'd put the buffering into the pubsub package as well, as it is a standard utility all clients would want.

rpc/client/httpclient.go
ethanfrey: hehehe.... panic on an ugly event key

ethanfrey:
Just added notes to your comments. No time to do a deeper review today.  Will try tomorrow, but I think @ebuchman should really add more thoughts, as I only have a shallow understanding of the consensus engine itself.

consensus/event_buffer.go
melekes: Note: there is already some buffering since I am using buffered channels. But EventBuffer gives you control on when to fire/process events.

rpc/client/httpclient.go
```diff
@@ -185,123 +181,80 @@ func (c *HTTP) Validators() (*ctypes.ResultValidators, error) {
...
+		quit:          make(chan bool, 1),
+		done:          make(chan bool, 1),
 	}
 }
 
-func (w *WSEvents) _assertIsEventSwitch() types.EventSwitch {
-	return w
-}
-
 // Start is the only way I could think the extend OnStart from
```
ebuchman: update comment

rpc/client/httpclient.go
```diff
@@ -185,123 +181,80 @@ func (c *HTTP) Validators() (*ctypes.ResultValidators, error) {
...
-	// if we did start, then OnStart here...
-	if st && err == nil {
-		ws := rpcclient.NewWSClient(w.remote, w.endpoint)
-		_, err = ws.Start()
-		if err == nil {
-			w.ws = ws
-			go w.eventListener()
-		}
+	ws := rpcclient.NewWSClient(w.remote, w.endpoint)
+	_, err = ws.Start()
```
ebuchman: shouldnt we grab the started value here and return it ?

rpc/client/httpclient.go
```diff
@@ -185,123 +181,80 @@ func (c *HTTP) Validators() (*ctypes.ResultValidators, error) {
...
-	return stop
+	// send a message to quit to stop the eventListener
+	w.quit <- true
+	<-w.done
+	w.ws.Stop()
+	w.ws = nil
+	return true
 }
 
 /** TODO: more intelligent subscriptions! **/
```
ebuchman: maybe we can remove this TODO now ? :)

rpc/client/httpclient.go
```diff
@@ -342,22 +295,17 @@ func (w *WSEvents) parseEvent(data []byte) (err error) {
...
+		out <- result.Data
 	}
+	return nil
 }
 
-func (w *WSEvents) unsubscribe(event string) {
-	err := w.ws.Unsubscribe(event)
-	if err != nil {
-		panic(err)
+func extractEventType(query string) string {
```
ebuchman: can we expose something in the query package instead ?

rpc/client/localclient.go
ebuchman: how to decide when to pass a channel and when to pass a cb ?

rpc/client/localclient.go
```diff
@@ -108,3 +113,40 @@ func (c Local) Validators() (*ctypes.ResultValidators, error) {
...
+		return errors.New("already subscribed")
+	}
+
+	q, err := tmquery.New(query)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse query")
+	}
+	ch := make(chan interface{})
+	c.PubSub.Subscribe(q, ch)
+	go func() {
```
ebuchman: extra go-routine per subscription. doesnt feel great.

rpc/lib/server/handlers.go
ebuchman: not sure

blockchain/reactor.go
```diff
@@ -49,7 +49,7 @@ type BlockchainReactor struct {
 	requestsCh   chan BlockRequest
 	timeoutsCh   chan string
 
-	evsw types.EventSwitch
+	pubsub types.PubSub
```
melekes: I like mux

consensus/byzantine_test.go
```diff
@@ -63,17 +64,17 @@ func TestByzantine(t *testing.T) {
...
-		if err != nil {
-			t.Fatalf("Failed to start switch: %v", err)
-		}
-		eventChans[i] = subscribeToEvent(eventSwitch, "tester", types.EventStringNewBlock(), 1)
+		pubsub := tmpubsub.NewServer()
+		pubsub.SetLogger(logger.With("module", "pubsub", "validator", i))
+		_, err := pubsub.Start()
+		require.NoError(t, err)
+
+		eventChans[i] = make(chan interface{})
```
melekes: it works without a buffer

consensus/reactor.go
```diff
@@ -290,27 +290,41 @@ func (conR *ConsensusReactor) Receive(chID byte, src *p2p.Peer, msgBytes []byte)
...
-	types.AddListenerForEvent(conR.evsw, "conR", types.EventStringNewRoundStep(), func(data types.TMEventData) {
-		rs := data.Unwrap().(types.EventDataRoundState).RoundState.(*RoundState)
-		conR.broadcastNewRoundStep(rs)
-	})
-
-	types.AddListenerForEvent(conR.evsw, "conR", types.EventStringVote(), func(data types.TMEventData) {
-		edv := data.Unwrap().(types.EventDataVote)
-		conR.broadcastHasVoteMessage(edv.Vote)
-	})
+	stepsCh := make(chan interface{}, 100)
```
melekes: oh boy... I think I've brought more magic to the streets of Tendermint

blockchain/reactor.go
melekes: Bus http://doc.akka.io/docs/akka/2.4.4/scala/event-bus.html

consensus/reactor.go
```diff
@@ -290,27 +290,41 @@ func (conR *ConsensusReactor) Receive(chID byte, src *p2p.Peer, msgBytes []byte)
...
+	conR.eventBus.Subscribe("consensus-reactor", types.EventQueryNewRoundStep, stepsCh)
+	votesCh := make(chan interface{}, 100)
+	conR.eventBus.Subscribe("consensus-reactor", types.EventQueryVote, votesCh)
+	go func() {
+		for {
+			select {
+			case data := <-stepsCh:
+				if data != nil {
```
melekes: why so? write a comment or get rid of that

consensus/reactor.go
melekes: why 100 and not 1?

consensus/reactor_test.go
```diff
@@ -19,27 +19,27 @@ func init() {
...
 		reactors[i].SetLogger(logger.With("validator", i))
 
-		eventSwitch := events.NewEventSwitch()
-		eventSwitch.SetLogger(logger.With("module", "events", "validator", i))
-		_, err := eventSwitch.Start()
+		pubsub := tmpubsub.NewServer()
+		pubsub.SetLogger(logger.With("module", "pubsub", "validator", i))
+
+		eventBus := types.NewEventBus(pubsub)
+		_, err := eventBus.Start()
```
melekes: stop eventBus at `stopConsensusNet`

consensus/replay_file.go
```diff
@@ -251,14 +258,13 @@ func newConsensusStateForReplay(config cfg.BaseConfig, csConfig *cfg.ConsensusCo
 		cmn.Exit(cmn.Fmt("Error starting proxy app conns: %v", err))
 	}
 
-	// Make event switch
-	eventSwitch := types.NewEventSwitch()
-	if _, err := eventSwitch.Start(); err != nil {
-		cmn.Exit(cmn.Fmt("Failed to start event switch: %v", err))
+	eventBus := types.NewEventBus(tmpubsub.NewServer())
+	if _, err := eventBus.Start(); err != nil {
+		cmn.Exit(cmn.Fmt("Failed to start event server: %v", err))
```
melekes: event bus

consensus/replay_test.go
```diff
@@ -125,7 +125,7 @@ func writeWAL(walMsgs string) string {
 }
 
 func waitForBlock(newBlockCh chan interface{}, thisCase *testCase, i int) {
-	after := time.After(time.Second * 10)
+	after := time.After(time.Second * 20)
```
melekes: ?

consensus/replay_test.go
```diff
@@ -149,17 +149,10 @@ func runReplayTest(t *testing.T, cs *ConsensusState, walFile string, newBlockCh
 	// Assuming the consensus state is running, replay of any WAL, including the empty one,
 	// should eventually be followed by a new block, or else something is wrong
 	waitForBlock(newBlockCh, thisCase, i)
-	cs.evsw.Stop()
+	cs.eventBus.Unsubscribe(testClientID, types.EventQueryNewBlock)
```
melekes: can we safely remove this line?

consensus/replay_test.go
```diff
@@ -238,26 +232,25 @@ func TestWALCrashBeforeWritePropose(t *testing.T) {
...
-		testReplayCrashBeforeWriteVote(t, thisCase, thisCase.precommitLine, types.EventStringPolka())
+		testReplayCrashBeforeWriteVote(t, thisCase, thisCase.precommitLine)
 	}
 }
 
-func testReplayCrashBeforeWriteVote(t *testing.T, thisCase *testCase, lineNum int, eventString string) {
+func testReplayCrashBeforeWriteVote(t *testing.T, thisCase *testCase, lineNum int) {
 	// setup replay test where last message is a vote
 	cs, newBlockCh, voteMsg, walFile := setupReplayTest(t, thisCase, lineNum, false)
-	types.AddListenerForEvent(cs.evsw, "tester", eventString, func(data types.TMEventData) {
```
melekes: Is it ok to remove this?

consensus/state.go
```diff
@@ -97,11 +97,12 @@ type RoundState struct {
 }
 
 func (rs *RoundState) RoundStateEvent() types.EventDataRoundState {
+	rsCopy := *rs
```
melekes: explain why (`data race`)

consensus/state_test.go
```diff
@@ -55,8 +55,10 @@ func TestProposerSelection0(t *testing.T) {
 	cs1, vss := randConsensusState(4)
 	height, round := cs1.Height, cs1.Round
 
-	newRoundCh := subscribeToEvent(cs1.evsw, "tester", types.EventStringNewRound(), 1)
-	proposalCh := subscribeToEvent(cs1.evsw, "tester", types.EventStringCompleteProposal(), 1)
+	newRoundCh := make(chan interface{}, 1)
```
melekes: return `subscribeToEvent`

node/node.go
```diff
@@ -196,14 +189,17 @@ func NewNode(config *cfg.Config, privValidator *types.PrivValidator, clientCreat
 		})
 	}
 
-	// add the event switch to all services
-	// they should all satisfy events.Eventable
-	SetEventSwitch(eventSwitch, bcReactor, mempoolReactor, consensusReactor)
+	pubsub := tmpubsub.NewServer(tmpubsub.BufferCapacity(1000), tmpubsub.OverflowStrategyDrop())
```
melekes: finalize options and write a comment

rpc/client/helpers.go
```diff
@@ -56,33 +54,21 @@ func WaitForHeight(c StatusClient, h int, waiter Waiter) error {
...
-		quit <- true
-	}()
+func WaitForOneEvent(c EventsClient, evtTyp string, timeout time.Duration) (types.TMEventData, error) {
+	evts := make(chan interface{}, 1)
 
 	// register for the next event of this type
-	evsw.AddListenerForEvent(listener, evtTyp, func(data events.EventData) {
-		evts <- data
-	})
+	err := c.Subscribe(types.EventTypeKey+"="+evtTyp, evts)
```
melekes: let's add a function for this `types.EventQueryFor(evtTyp)`

rpc/client/httpclient.go
```diff
@@ -185,123 +182,85 @@ func (c *HTTP) Validators() (*ctypes.ResultValidators, error) {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to subscribe")
 	}
+	w.subscriptions[query] = struct{}{}
 
-	// now we can update the subscriptions
-	w.evtCount[event] -= 1
-	if w.evtCount[event] == 0 {
-		w.unsubscribe(event)
+	w.mtx.Lock()
```
melekes: find a way to get rid of this

rpc/client/httpclient.go
```diff
@@ -342,22 +301,10 @@ func (w *WSEvents) parseEvent(data []byte) (err error) {
...
-	if err != nil {
-		panic(err)
-	}
-}
-
-func (w *WSEvents) unsubscribe(event string) {
-	err := w.ws.Unsubscribe(event)
-	if err != nil {
-		panic(err)
+	if result.Name != "" {
```
melekes: when there could be empty (`nil`) events?

rpc/client/localclient.go
```diff
@@ -1,11 +1,18 @@
...
 	data "github.com/tendermint/go-wire/data"
 	nm "github.com/tendermint/tendermint/node"
 	"github.com/tendermint/tendermint/rpc/core"
 	ctypes "github.com/tendermint/tendermint/rpc/core/types"
 	"github.com/tendermint/tendermint/types"
+	tmquery "github.com/tendermint/tmlibs/pubsub/query"
+)
+
+const (
+	pubsubClientID = "rpc-localclient"
```
melekes: subscriber

rpc/client/localclient.go
```diff
@@ -108,3 +115,27 @@ func (c Local) Validators() (*ctypes.ResultValidators, error) {
...
+		return errors.Wrap(err, "failed to parse query")
+	}
+
+	c.EventBus.Subscribe(pubsubClientID, q, out)
+
+	return nil
+}
+
+func (c Local) Unsubscribe(query string) {
+	q, err := tmquery.New(query)
```
melekes: cache query?

rpc/core/events.go
```diff
@@ -1,24 +1,54 @@
...
 	return &ctypes.ResultSubscribe{}, nil
 }
 
-func Unsubscribe(wsCtx rpctypes.WSRPCContext, event string) (*ctypes.ResultUnsubscribe, error) {
-	logger.Info("Unsubscribe to event", "remote", wsCtx.GetRemoteAddr(), "event", event)
-	wsCtx.GetEventSwitch().RemoveListenerForEvent(event, wsCtx.GetRemoteAddr())
+func Unsubscribe(wsCtx rpctypes.WSRPCContext, query string) (*ctypes.ResultUnsubscribe, error) {
+	addr := wsCtx.GetRemoteAddr()
+	logger.Info("Unsubscribe from query", "remote", addr, "query", query)
+	q, err := tmquery.New(query)
```
melekes: cache query? possibly inside `wsCtx`

rpc/lib/types/types.go
```diff
@@ -74,8 +72,8 @@ func NewRPCResponse(id string, res interface{}, err string) RPCResponse {
 
 // *wsConnection implements this interface.
 type WSRPCConnection interface {
+	IsRunning() bool
```
melekes: remove?

consensus/common_test.go
```diff
@@ -248,10 +249,11 @@ func newConsensusStateWithConfig(thisConfig *cfg.Config, state *sm.State, pv *ty
 	cs.SetLogger(log.TestingLogger())
 	cs.SetPrivValidator(pv)
 
-	evsw := types.NewEventSwitch()
-	evsw.SetLogger(log.TestingLogger().With("module", "events"))
-	cs.SetEventSwitch(evsw)
-	evsw.Start()
+	pubsub := tmpubsub.NewServer()
+	pubsub.SetLogger(log.TestingLogger().With("module", "pubsub"))
+	pubsub.Start()
```
melekes: does not matter. important that we start cs (publisher) after we start pubsub

consensus/reactor_test.go
```diff
@@ -27,18 +27,24 @@ func startConsensusNet(t *testing.T, css []*ConsensusState, N int, subscribeEven
...
+
+		eventChans[i] = make(chan interface{})
+		pubsub.Subscribe(types.EventQueryNewBlock, eventChans[i])
 		if subscribeEventRespond {
-			eventChans[i] = subscribeToEventRespond(eventSwitch, "tester", types.EventStringNewBlock())
-		} else {
-			eventChans[i] = subscribeToEvent(eventSwitch, "tester", types.EventStringNewBlock(), 1)
+			go func(ch chan interface{}) {
+				for data := range ch {
+					ch <- data
```
melekes: got rid of that

consensus/replay_test.go
```diff
@@ -149,17 +150,10 @@ func runReplayTest(t *testing.T, cs *ConsensusState, walFile string, newBlockCh
 	// Assuming the consensus state is running, replay of any WAL, including the empty one,
 	// should eventually be followed by a new block, or else something is wrong
 	waitForBlock(newBlockCh, thisCase, i)
-	cs.evsw.Stop()
 	cs.Stop()
-LOOP:
```
melekes: because we don't need it anymore. `cs.eventBus.Stop()` will close the channel

consensus/state_test.go
```diff
@@ -55,8 +55,10 @@ func TestProposerSelection0(t *testing.T) {
 	cs1, vss := randConsensusState(4)
 	height, round := cs1.Height, cs1.Round
 
-	newRoundCh := subscribeToEvent(cs1.evsw, "tester", types.EventStringNewRound(), 1)
-	proposalCh := subscribeToEvent(cs1.evsw, "tester", types.EventStringCompleteProposal(), 1)
+	newRoundCh := make(chan interface{})
+	cs1.pubsub.Subscribe(types.EventQueryNewRound, newRoundCh)
+	proposalCh := make(chan interface{})
```
melekes: we do (in state_test at least). sorry about that

rpc/client/httpclient.go
melekes: yes, thanks

consensus/state_test.go
```diff
@@ -120,7 +123,8 @@ func TestEnterProposeNoPrivValidator(t *testing.T) {
 	height, round := cs.Height, cs.Round
 
 	// Listen for propose timeout event
-	timeoutCh := subscribeToEvent(cs.evsw, "tester", types.EventStringTimeoutPropose(), 1)
+	timeoutCh := make(chan interface{})
```
melekes: I didn't like that the only reason was this "one line" benefit... I've renamed it to `subscribe` and reduced the arguments list to just 2 items (`eventBus` and `query`)

rpc/client/localclient.go
melekes: I will work on this tomorrow

consensus/mempool_test.go
```diff
@@ -16,11 +16,11 @@ func init() {
 }
 
 func TestTxConcurrentWithCommit(t *testing.T) {
-
 	state, privVals := randGenesisState(1, false, 10)
 	cs := newConsensusState(state, privVals[0], NewCounterApplication())
 	height, round := cs.Height, cs.Round
-	newBlockCh := subscribeToEvent(cs.evsw, "tester", types.EventStringNewBlock(), 1)
+	newBlockCh := make(chan interface{})
```
melekes: working without it

consensus/common_test.go
```diff
@@ -207,11 +208,11 @@ func validatePrevoteAndPrecommit(t *testing.T, cs *ConsensusState, thisRound, lo
 
 // genesis
 func subscribeToVoter(cs *ConsensusState, addr []byte) chan interface{} {
-	voteCh0 := subscribeToEvent(cs.evsw, "tester", types.EventStringVote(), 1)
+	voteCh0 := make(chan interface{})
```
melekes: working without it

consensus/replay_file.go
```diff
@@ -41,7 +42,11 @@ func (cs *ConsensusState) ReplayFile(file string, console bool) error {
 	cs.startForReplay()
 
 	// ensure all new step events are regenerated as expected
-	newStepCh := subscribeToEvent(cs.evsw, "replay-test", types.EventStringNewRoundStep(), 1)
+	newStepCh := make(chan interface{})
```
melekes: if tests pass without it, does it mean it is not needed or it is just we do not cover this in tests?

melekes: ah, I see. we don't have tests for replay-file

consensus/replay_test.go
melekes: why was it added in the first place? i.e. why we needed to drain the channel before `cs.Wait()`?

rpc/client/helpers.go
```diff
@@ -56,33 +55,23 @@ func WaitForHeight(c StatusClient, h int, waiter Waiter) error {
...
-		evts <- data
-	})
+	err := c.Subscribe(ctx, types.EventTypeKey+"="+evtTyp, evts)
+	if err != nil {
+		return types.TMEventData{}, errors.Wrap(err, "failed to subscribe")
+	}
 	// make sure to unregister after the test is over
-	defer evsw.RemoveListenerForEvent(evtTyp, listener)
-	// defer evsw.RemoveListener(listener)  // this also works
+	defer c.UnsubscribeAll(ctx)
```
melekes: `c.Unsubscribe(ctx, types.EventTypeKey+"="+evtTyp)`

rpc/client/httpclient.go
```diff
@@ -185,123 +187,77 @@ func (c *HTTP) Validators() (*ctypes.ResultValidators, error) {
 /** websocket event stuff here... **/
 
 type WSEvents struct {
-	types.EventSwitch
+	cmn.BaseService
 	remote   string
 	endpoint string
 	ws       *rpcclient.WSClient
 
+	out chan<- interface{}
```
melekes: create a Github issue to get rid of that

consensus/replay_test.go
ebuchman: because everything is synchronous. so cs would block on firing on newBlockCh while cs.Wait is waiting for it to exit -> deadlock

consensus/replay_test.go
ebuchman: uh  I don't think so. These tests are overdue for some refactoring but the point was to make sure we're at a particular point in the consensus state machine before

consensus/replay_test.go
melekes: > because everything is synchronous.

well, not anymore

consensus/state.go
```diff
@@ -97,11 +97,13 @@ type RoundState struct {
 }
 
 func (rs *RoundState) RoundStateEvent() types.EventDataRoundState {
+	// copy to avoid data race
```
ebuchman: thats new

melekes:
What's left here:

[ ] consensus/state:99 check we do need to make a copy to avoid data races
[ ] fix TestValidatorSetChanges
[ ] fix testReplayCrashBeforeWriteVote

melekes:
```
I[07-17|09:24:25.595] Starting ConsensusState                      impl=ConsensusState
I[07-17|09:24:25.595] Starting WAL                                 wal=/tmp/walwIEZYXZnRPmu/wal impl=WAL
I[07-17|09:24:25.595] Starting TimeoutTicker                       impl=TimeoutTicker
D[07-17|09:24:25.596] Starting timeout routine
I[07-17|09:24:25.596] Catchup by replaying consensus messages      height=1
I[07-17|09:24:25.596] Replay: Timeout                              height=1 round=0 step=RoundStepNewHeight dur=972.946821ms
D[07-17|09:24:25.596] Received tock                                timeout=972.946821ms height=1 round=0 step=RoundStepNewHeight
I[07-17|09:24:25.596] Need to set a buffer and log message here for sanity. startTime=2017-07-17T09:24:25.605212694Z now=2017-07-17T09:24:25.596515063Z
I[07-17|09:24:25.596] enterNewRound(1/0). Current: 1/0/RoundStepNewHeight
I[07-17|09:24:25.596] enterPropose(1/0). Current: 1/0/RoundStepNewRound
D[07-17|09:24:25.597] Timer already stopped
D[07-17|09:24:25.597] Scheduled timeout                            dur=2s height=1 round=0 step=RoundStepPropose
I[07-17|09:24:25.596] enterPropose: Our turn to propose            proposer=D028C9981F7A87F3093672BF0D5B0E2A1B3ED456 privValidator="PrivValidator{D028C9981F7A87F3093672BF0D5B0E2A1B3ED456 LH:1, LR:0, LS:2}"
D[07-17|09:24:25.597] This node is a validator
I[07-17|09:24:25.597] Replay: New Step                             height=1 round=0 step=RoundStepPropose
I[07-17|09:24:25.597] Replay: Proposal                             height=1 round=0 header=1:ACED4A95DDEB pol=-1 peer=local
I[07-17|09:24:25.598] Replay: BlockPart                            height=1 round=0 peer=local
I[07-17|09:24:25.598] Received complete proposal block             height=1 hash=F3BBFBE7E4A5D619E2C498C3D1B912883786DD71
D[07-17|09:24:25.598] published event
I[07-17|09:24:25.598] enterPrevote(1/0). Current: 1/0/RoundStepPropose
I[07-17|09:24:25.598] enterPrevote: ProposalBlock is valid         height=1 round=0
D[07-17|09:24:25.599] signing vote                                 pv="PrivValidator{D028C9981F7A87F3093672BF0D5B0E2A1B3ED456 LH:1, LR:0, LS:2}"
E[07-17|09:24:25.599] Error signing vote                           height=1 round=0 vote="Vote{0:D028C9981F7A 1/00/1(Prevote) F3BBFBE7E4A5 %!v(PANIC=runtime error: invalid memory address or nil pointer dereference)}" err="Error signing vote: Step regression"
I[07-17|09:24:25.599] Replay: New Step                             height=1 round=0 step=RoundStepPrevote
D[07-17|09:24:25.599] got event                                    s=/35C937C78D06.../
I[07-17|09:24:25.599] Replay: New Step                             height=1 round=0 step=RoundStepPropose
I[07-17|09:24:25.600] Replay: New Step                             height=1 round=0 step=RoundStepPrevote
I[07-17|09:24:25.600] Replay: Done
I[07-17|09:24:27.607] Timed out                                    dur=2s height=1 round=0 step=RoundStepPropose
D[07-17|09:24:27.607] Received tock                                timeout=2s height=1 round=0 step=RoundStepPropose
D[07-17|09:24:27.607] Ignoring tock because we're ahead            height=1 round=0 step=RoundStepPrevote
--- FAIL: TestWALCrashBeforeWritePrevote (10.09s)
```

the problem with TestWALCrashBeforeWritePrevote is timing (by the time we "got event", it is already late to set the last signature because "signing" had occurred earlier)

melekes:
Managed to get around https://github.com/tendermint/tendermint/pull/568#issuecomment-315708867. Not sure you will like the solution I've come up with (I don't like it, that's for sure).

consensus/state.go
melekes: we don't need it before because callbacks were executed within the same context. now we spawn a thread for each event type https://github.com/tendermint/tendermint/blob/feature/548-indexing-tags/consensus/reactor.go#L317

anyway, I think transferring mutable state between 2 threads is a bad idea. events should be immutable, IMHO.

ebuchman: Why not have one routine with a select statement over the two pubsub channels and the quit channel ?

types/event_buffer.go
```diff
@@ -0,0 +1,41 @@
...
+var _ TxEventPublisher = (*TxEventBuffer)(nil)
+
+// TxEventBuffer is a buffer of events.
+type TxEventBuffer struct {
+	next   TxEventPublisher
+	events []EventDataTx
+}
+
+// NewTxEventBuffer returns a new buffer
+func NewTxEventBuffer(next *EventBus) *TxEventBuffer {
```
ebuchman: Maybe we can take the buffer size as an argument since we know how many txs are in the block

types/events.go
```diff
@@ -135,100 +137,39 @@ func (_ EventDataTx) AssertIsTMEventData()             {}
...
+const (
+	// EventTypeKey is a reserved key, used to specify event type in tags.
+	EventTypeKey = "tm.events.type"
+)
 
-func FireEventUnlock(fireable events.Fireable, rs EventDataRoundState) {
-	fireEvent(fireable, EventStringUnlock(), TMEventData{rs})
-}
+var (
+	EventQueryBond             = tmquery.MustParse(EventTypeKey + "=" + EventBond)
```
ebuchman: how about a helper function so it's just like `EventQueryBond = mustParseTMEventType(EventBond)`

types/event_buffer.go
```diff
@@ -0,0 +1,41 @@
...
+	return &TxEventBuffer{
+		next:   next,
+		events: make([]EventDataTx, 0, txEventBufferCapacity),
+	}
+}
+
+// PublishWithTags buffers an event to be fired upon finality.
```
ebuchman: fix comment. `PublishEventTx`

consensus/state.go
melekes: speed? maybe not worth it though. will return back select

melekes:
> Once the ABCI returns DeliverTx tags, presumably we will PublishWithTags with all those tags, correct ?

correct

node/node.go
```diff
@@ -330,7 +326,7 @@ func (n *Node) startRPC() ([]net.Listener, error) {
 	listeners := make([]net.Listener, len(listenAddrs))
 	for i, listenAddr := range listenAddrs {
 		mux := http.NewServeMux()
-		wm := rpcserver.NewWebsocketManager(rpccore.Routes, n.evsw)
+		wm := rpcserver.NewWebsocketManager(rpccore.Routes)
```
melekes: do you think it is not a good idea? I thought why event switch/hub should be a part of WebSocket manager (WebSocket conn)? why instead not just set event hub on RPC

types/event_buffer.go
melekes: good idea!

node/node.go
ebuchman: like here `rpccore.SetEventBus(n.eventBus)` ? and just let the websockets access it globally like everything else is global in that package? 

I guess that could be fine.

consensus/state_test.go
```diff
@@ -237,9 +241,11 @@ func TestFullRound1(t *testing.T) {
 	cs, vss := randConsensusState(1)
 	height, round := cs.Height, cs.Round
 
-	voteCh := subscribeToEvent(cs.evsw, "tester", types.EventStringVote(), 0)
-	propCh := subscribeToEvent(cs.evsw, "tester", types.EventStringCompleteProposal(), 1)
-	newRoundCh := subscribeToEvent(cs.evsw, "tester", types.EventStringNewRound(), 1)
+	voteCh := make(chan interface{}, 0)
+	err := cs.eventBus.Subscribe(context.Background(), testSubscriber, types.EventQueryVote, voteCh)
```
melekes: custom case (the only place where we need unbuffered channel)

node/node.go
```diff
@@ -279,6 +280,8 @@ func (n *Node) OnStop() {
 			n.Logger.Error("Error closing listener", "listener", l, "err", err)
 		}
 	}
+
+	n.eventBus.Stop()
```
melekes: so instead of this:
```
	n.BaseService.OnStop()

	n.Logger.Info("Stopping Node")
	// TODO: gracefully disconnect from peers.
	n.sw.Stop()

	for _, l := range n.rpcListeners {
		n.Logger.Info("Closing rpc listener", "listener", l)
		if err := l.Close(); err != nil {
			n.Logger.Error("Error closing listener", "listener", l, "err", err)
		}
	}

	n.eventBus.Stop()
```

you are suggesting 

```
          defer n.BaseService.OnStop()

	  defer n.Logger.Info("Stopping Node")
	// TODO: gracefully disconnect from peers.
	defer n.sw.Stop()

        defer func() {
	for _, l := range n.rpcListeners {
		n.Logger.Info("Closing rpc listener", "listener", l)
		if err := l.Close(); err != nil {
			n.Logger.Error("Error closing listener", "listener", l, "err", err)
		}
	}
        }()

	defer n.eventBus.Stop()
```

or just `eventBus`? why should `eventBus` be stopped with `defer` while others are not?

node/node.go
melekes: > like here rpccore.SetEventBus(n.eventBus) ?

yes

rpc/client/helpers.go
```diff
@@ -56,33 +55,23 @@ func WaitForHeight(c StatusClient, h int, waiter Waiter) error {
...
 	select {
-	case <-quit:
-		return types.TMEventData{}, errors.New("timed out waiting for event")
 	case evt := <-evts:
-		tmevt, ok := evt.(types.TMEventData)
-		if ok {
-			return tmevt, nil
-		}
-		return types.TMEventData{}, errors.Errorf("Got unexpected event type: %#v", evt)
+		return evt.(types.TMEventData), nil
```
melekes: if we are listening for events (i.e using a `tm.events.type=X` in the query), then yes. all TM events are published through https://github.com/tendermint/tendermint/blob/feature/548-indexing-tags/types/event_bus.go#L43

types/event_buffer.go
melekes: fixed

consensus/reactor_test.go
```diff
@@ -140,11 +146,21 @@ func TestVotingPowerChange(t *testing.T) {
 }
 
 func TestValidatorSetChanges(t *testing.T) {
+	// FIXME: need to block css[j] until we check all txs! otherwise validator
```
melekes: @ebuchman will need your vote on this or suggestions. 

note: previously we were blocking the consensus state by blocking the events loop (they were executing in the same thread). new events package runs in its own thread. this is where callback vs channels plays a big role. because you can't exchange msg within the same thread/goroutine.

consensus/replay_test.go
```diff
@@ -238,26 +236,39 @@ func TestWALCrashBeforeWritePropose(t *testing.T) {
...
 
 func TestWALCrashBeforeWritePrecommit(t *testing.T) {
 	for _, thisCase := range testCases {
-		testReplayCrashBeforeWriteVote(t, thisCase, thisCase.precommitLine, types.EventStringPolka())
+		testReplayCrashBeforeWriteVote(t, thisCase, thisCase.precommitLine, types.EventQueryPolka)
 	}
 }
 
-func testReplayCrashBeforeWriteVote(t *testing.T, thisCase *testCase, lineNum int, eventString string) {
+// FIXME: For every test (TestWALCrashBeforeWritePrevote,
```
melekes: same comment here

consensus/state_test.go
ebuchman: cool

consensus/reactor_test.go
ebuchman: Though it would still be a hack, we might be able to use TimeoutCommit here instead of the new param/sleep. TimeoutCommit  is the time a node waits after committing a block, before it goes into the next height. So it will finish everything from the last block, but then wait a bit. The idea is this gives it time to hear more votes from other validators, to strengthen the commit it includes in the next block. But it also gives it time to hear about new transactions.

Regardless, its not good that any of our tests depend on time, and we'll need to design something else.

melekes:
Now clients can subscribe for all txs.

melekes:
Don't forget to choose "Squash and merge" when merging

consensus/state.go
```diff
@@ -626,16 +627,22 @@ func (cs *ConsensusState) receiveRoutine(maxSteps int) {
 
 		select {
 		case mi = <-cs.peerMsgQueue:
+			cs.mtx.Lock()
```
melekes: hate these locks

consensus/replay_test.go
```diff
@@ -54,223 +59,120 @@ var data_dir = path.Join(cmn.GoPath, "src/github.com/tendermint/tendermint/conse
...
+		wal.cs = cs
+		wal.next = cs.wal
+		wal.mtx.Unlock()
+		cs.SetWAL(wal)
+
+		select {
+		case err := <-walPaniced:
+			t.Logf("WAL paniced: %v", err)
+
+			time.Sleep(100 * time.Millisecond)
```
melekes: huh?

melekes:
```
panic: Panicked on a Crisis: rename /root/.tendermint_test/consensus_replay_test/priv_validator.json.new /root/.tendermint_test/consensus_replay_test/priv_validator.json: no such file or directory

goroutine 47860 [running]:
github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common.PanicCrisis(0xcba800, 0xc42152d640)
	/go/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common/errors.go:33 +0x10f
github.com/tendermint/tendermint/types.(*PrivValidator).save(0xc42235f2c0)
	/go/src/github.com/tendermint/tendermint/types/priv_validator.go:165 +0x159
github.com/tendermint/tendermint/types.(*PrivValidator).signBytesHRS(0xc42235f2c0, 0x6, 0x0, 0xc424e88f03, 0xc429908580, 0xca, 0x155, 0x80, 0xc424e88f00, 0x7f4ecafc88d0, ...)
	/go/src/github.com/tendermint/tendermint/types/priv_validator.go:249 +0x2bb
github.com/tendermint/tendermint/types.(*PrivValidator).SignVote(0xc42235f2c0, 0xc4228c7460, 0xf, 0xc424e88f00, 0x0, 0x0)
	/go/src/github.com/tendermint/tendermint/types/priv_validator.go:186 +0x1a2
github.com/tendermint/tendermint/consensus.(*ConsensusState).signVote(0xc424efd520, 0xc400000002, 0xc422d5e3c0, 0x14, 0x20, 0x1, 0xc4247b6560, 0x14, 0x20, 0x0, ...)
	github.com/tendermint/tendermint/consensus/_test/_obj_test/state.go:1556 +0x35e
github.com/tendermint/tendermint/consensus.(*ConsensusState).signAddVote(0xc424efd520, 0x2, 0xc422d5e3c0, 0x14, 0x20, 0x1, 0xc4247b6560, 0x14, 0x20, 0xc42001b300)
	github.com/tendermint/tendermint/consensus/_test/_obj_test/state.go:1568 +0x200
github.com/tendermint/tendermint/consensus.(*ConsensusState).enterPrecommit(0xc424efd520, 0x6, 0x0)
	github.com/tendermint/tendermint/consensus/_test/_obj_test/state.go:1082 +0x13a4
github.com/tendermint/tendermint/consensus.(*ConsensusState).addVote(0xc424efd520, 0xc424e88780, 0x0, 0x0, 0x39, 0x1dc, 0x28c)
	github.com/tendermint/tendermint/consensus/_test/_obj_test/state.go:1477 +0x1be5
github.com/tendermint/tendermint/consensus.(*ConsensusState).tryAddVote(0xc424efd520, 0xc424e88780, 0x0, 0x0, 0xd7fb00, 0xc42152ce00)
	github.com/tendermint/tendermint/consensus/_test/_obj_test/state.go:1382 +0x93
github.com/tendermint/tendermint/consensus.(*ConsensusState).handleMsg(0xc424efd520, 0xcb58e0, 0xc42547feb8, 0x0, 0x0, 0x6, 0x0, 0x4, 0xed10ca07e, 0x3077bfea, ...)
	github.com/tendermint/tendermint/consensus/_test/_obj_test/state.go:660 +0x9fc
github.com/tendermint/tendermint/consensus.(*ConsensusState).receiveRoutine(0xc424efd520, 0x0)
	github.com/tendermint/tendermint/consensus/_test/_obj_test/state.go:615 +0x5f5
created by github.com/tendermint/tendermint/consensus.(*ConsensusState).OnStart
	github.com/tendermint/tendermint/consensus/_test/_obj_test/state.go:332 +0x4a7
exit status 2
FAIL	github.com/tendermint/tendermint/consensus	76.644s
make: *** [test_integrations] Error 1
```

melekes:
https://github.com/tendermint/tmlibs/pull/32

consensus/reactor_test.go
```diff
@@ -254,16 +256,37 @@ func waitForAndValidateBlock(t *testing.T, n int, activeVals map[string]struct{}
...
+func waitForBlockWithUpdatedValsAndValidateIt(t *testing.T, n int, activeVals map[string]struct{}, eventChans []chan interface{}, css []*ConsensusState, txs ...[]byte) {
+	timeoutWaitGroup(t, n, func(wg *sync.WaitGroup, j int) {
+	receive:
+		newBlockI := <-eventChans[j]
+		newBlock := newBlockI.(types.TMEventData).Unwrap().(types.EventDataNewBlock).Block
+		if newBlock.LastCommit.Size() == len(activeVals) {
+			t.Logf("Block with new validators height=%v validator=%v", newBlock.Height, j)
+		} else {
+			t.Logf("Block with no new validators height=%v validator=%v. Skipping...", newBlock.Height, j)
+			goto receive
```
ebuchman: why `goto` and not a for loop ?

consensus/state.go
```diff
@@ -30,6 +30,10 @@ var (
 	ErrInvalidProposalPOLRound  = errors.New("Error invalid proposal POL round")
 	ErrAddingVote               = errors.New("Error adding vote")
 	ErrVoteHeightMismatch       = errors.New("Error vote height mismatch")
+
+	// see testReplayCrashBeforeWriteVote in replay_test.go
+	afterPublishEventPolkaCallback            func(*ConsensusState)
```
ebuchman: can we remove these now ?

consensus/state.go
```diff
@@ -328,10 +335,16 @@ func (cs *ConsensusState) LoadCommit(height int) *types.Commit {
...
+	// we may set the WAL in testing before calling Start,
+	// so only OpenWAL if its still the nilWAL
+	if _, ok := cs.wal.(nilWAL); ok {
+		walFile := cs.config.WalFile()
+		wal, err := cs.OpenWAL(walFile)
+		if err != nil {
+			cs.Logger.Error("Error loading ConsensusState wal", "err", err.Error())
+			return err
+		}
+		cs.wal = wal // do we need a lock for this ?
```
ebuchman: think we can remove comment

consensus/state.go
```diff
@@ -921,7 +933,10 @@ func (cs *ConsensusState) enterPrevote(height int, round int) {
 
 	// fire event for how we got here
 	if cs.isProposalComplete() {
-		types.FireEventCompleteProposal(cs.evsw, cs.RoundStateEvent())
+		cs.eventBus.PublishEventCompleteProposal(cs.RoundStateEvent())
+		if afterPublishEventCompleteProposalCallback != nil {
```
ebuchman: think we can remove

consensus/state_test.go
```diff
@@ -237,9 +241,13 @@ func TestFullRound1(t *testing.T) {
 	cs, vss := randConsensusState(1)
 	height, round := cs.Height, cs.Round
 
-	voteCh := subscribeToEvent(cs.evsw, "tester", types.EventStringVote(), 0)
-	propCh := subscribeToEvent(cs.evsw, "tester", types.EventStringCompleteProposal(), 1)
-	newRoundCh := subscribeToEvent(cs.evsw, "tester", types.EventStringNewRound(), 1)
+	// NOTE: voteChan cap of 0 ensures we can complete this
+	// before consensus can move to the next height (and cause a race condition)
```
ebuchman: is this still true now that eventbus runs in its own routine ?

rpc/client/localclient.go
```diff
@@ -33,78 +40,104 @@ type Local struct {
...
+		return errors.Wrap(err, "failed to parse query")
+	}
+	c.subscriptions[query] = q
+	return c.EventBus.Subscribe(ctx, "rpclocalclient", q, out)
+}
+
+func (c *Local) Unsubscribe(ctx context.Context, query string) error {
+	q, ok := c.subscriptions[query]
+	if ok {
+		return c.EventBus.Unsubscribe(ctx, "rpclocalclient", q)
```
ebuchman: should we also `delete(subscriptions, query)` ?

consensus/reactor_test.go
melekes: you do not like goto?

consensus/state_test.go
melekes: no

odeke-em:
Hello @melekes, unfortunately the branch went still and will need a rebase and merge conflict fixes. Also the PRs on which it depended have been merged in so perhaps we also need to update the glide/dependency locks.

melekes:
Wow) Do we still want to merge this? If yes, I will try to rebase tomorrow

melekes:
Finishing rebasing it.

melekes:
Done rebasing

ebuchman:
Let's push this to the next release so we can review again

PR #570, opened 2017-07-05
p2p: sw.peers.List() is empty in sw.OnStart

cjqpker:
sw.peers.List() will always be a empty list here

### DIFF
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -166,10 +166,7 @@ func (sw *Switch) OnStart() error {
 			return err
 		}
 	}
-	// Start peers
-	for _, peer := range sw.peers.List() {
-		sw.startInitPeer(peer)
-	}
+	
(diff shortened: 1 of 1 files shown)


melekes:
> sw.peers.List() will always be a empty list here

looks like it

ebuchman:
How do you like that! 

Could you please update the commit msg, eg: `p2p: dont bother starting peers in sw.OnStart`

ebuchman:
Oh I see the rest of the commit message now! Still, could be better as `p2p: sw.peers.List() is empty in sw.OnStart`

cjqpker:
got it

ebuchman:
ah, I meant the commit message. no worries, merged: https://github.com/tendermint/tendermint/commit/086544e367f98bd1625e0aa79f55f4553bcbe374

PR #579, opened 2017-07-10
Add fast-sync status to Status() call

adrianbrink:


srmo:
Out of curiosity, why do you want to only report FastSync and not "regular sync" too?

adrianbrink:
Regular sync is when you are caught up with the network. For example on ethereum you would consider that fully caught up with the network. 

That why I only support fast-sync, because that shows when we are still behind the tip of the chain.[](url)

srmo:
That's new to me. I understood that the catch-up state knows "sync" and "fast-sync" as reflected by it's properties. So maybe it should rather be "catching-up" instead of "fast-sync", because we can be in two states catching-up. Or did you guys remove the slow state and now there's only fast-sync, or what I'm missing here?

ebuchman:
Weren't we going to include info about the highest height we know of ? 

@srmo if we're not in fast-sync we're in regular sync, which is just the normal consensus algorithm. Nothing really to report on there. Though if we're falling behind, we should go back to fast-syncing ( https://github.com/tendermint/tendermint/issues/129)

ebuchman:
Um, is it just me, or do we never set bcR.fastSync back to false ?! Lol

srmo:
@ebuchman @adrianbrink  Sorry for bothering, but I'm really lost now.
My world was: 
A) our node is in sync with the chain, everyhting ok
B) our node is not in sync with the chain, well, we need to catch-up
B1) do the catch up in a normal way
B2) do the catch up in a fast-sync way
C) now we also have the scenarios were the  ABCI  can be behind, which we call the replay?

So we have 2-3 states: synced, catching-up, replaying. I might have the semantics totally wrong but seriously, why do we suddenly have fast-sync only? Even #129 talks about "slow-syncing". I don't see how this is nothing to report, it's clear a state where we are not at the same height as the chain.

adrianbrink:
There are two possible states for syncing:
1. Fast-sync - we lag behind significantly (B2) and need to catch-up to the tip of the chain (A)
2. Normal-sync - we are close to the tip of the chain (A) and just need to process new transactions and blocks as they come in

Replaying (C) is a state that should be handled by the ABCI app and is not of concern to Tendermint core. However, while syncing Tendermint core will of course also replay any transactions to the ABCI app.

If our node lags behind the rest of the network it should always fast-sync (B2) to the tip of the chain (A). Th state B1 does not exist.

@srmo Does that clarify this?

ebuchman:
When we're in fast sync mode, we explicitly ask for the highest height our peers know about, so we can report on that. When we're in normal sync, but behind, we're not really asking for the highest height. We might hear about some higher height, but no reason to think its the highest, so it won't be the best reporting, so it doesn't really make sense to report on it. If we're not in fast sync, we're in normal sync. If we're in normal sync and behind, then we should probably go back to fast-sync. That needs to be implemented.

adrianbrink:
Okay, I'll also tackle going back to fast-sync later this week.

So, let's build it like this:
Sync status reporting:
When in fast sync we report true for syncing and the highest height we know about and when in normal-sync we report false.

And I'll ensure that if we lack behind we go back to fast-sync. The question here is: How do we know that we are too far behind? Shall we say that if we are 5 minutes worth of blocks behind (this value is based on the average block height we hear from our peers) that we go back into fast-sync?

srmo:
You see, the trouble I'm having with this semantic is that our use case considers even being 1 block behind to be too late, i.e. we consider this a catch up state.

adrianbrink:
@ebuchman We do set it to false when we call SwitchToConsensus()

ebuchman:
We do in the consensusReactor. But in this PR we're reading the fastSync variable from the blockchainReactor, which never does get set to false!

ramilexe:
@ebuchman, @adrianbrink please take a look at https://github.com/tendermint/tendermint/pull/587

consensus/reactor.go
```diff
@@ -290,6 +290,11 @@ func (conR *ConsensusReactor) SetEventSwitch(evsw types.EventSwitch) {
 	conR.conS.SetEventSwitch(evsw)
 }
 
+// FastSync returns whether the consensus reactor is currently fast syncing
```
ebuchman: actually the consensus reactor doesn't do the fast sync (the blockchain reactor does), so the comment should be updated

ebuchman: Also I think this is racey since there's no lock around this var

rpc/core/pipe.go
```diff
@@ -89,6 +90,10 @@ func SetTxIndexer(indexer txindex.TxIndexer) {
 	txIndexer = indexer
 }
 
+func SetConsensusReactor(conR *consensus.ConsensusReactor) {
+	consensusReactor = conR
```
odeke-em: Same thing here that @ebuchman had raised about racy behavior, perhaps we need a global mutex of sorts since this function mutates a global consensusReactor.

ebuchman: Currently we're structured to only call this once from node/node.go at setup time. But we could certainly use comments/docs reflecting that fact!

odeke-em: Gotcha! Thanks for the explanation. SGTM!

PR #586, opened 2017-07-14
remove getRoundState()

needkane:
I don't know why use getRoundState, it looks like a superfluous function

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -270,12 +270,7 @@ func (cs *ConsensusState) GetState() *sm.State {
 func (cs *ConsensusState) GetRoundState() *RoundState {
 	cs.mtx.Lock()
 	defer cs.mtx.Unlock()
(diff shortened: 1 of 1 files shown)


ebuchman:
Thanks for the pull request!

Unfortunately, we can't merge it, as `getRoundState` actually performs a useful function: it ensures that we return a *copy* of the RoundState. This is what the comment is referring to. It's important because `GetRoundState` is called by concurrent routines, who would otherwise be modifying the same spot in memory concurrently, which would be unsafe. You can confirm this for yourself by running the tests with Go's builtin race detector: 

```
go test --race ./consensus
```

You should see numerous race conditions. Of course, if you just run our integration test (`make test_integrations`), it will run this test for you (along with many others), and will report the failure. Hence why this patch failed the continuous integration test on CircleCI.

In the future, please make pull requests against the develop branch. Thanks!

PR #596, opened 2017-07-28
Deduplications

zramsay:
use one install.md and one contributing.md
- closes https://github.com/zramsay/tendermint/issues/7 and https://github.com/zramsay/tendermint/issues/2


ebuchman:
shouldn't contributing also be in that docs folder? do we not want it on the site for some reason? I think we want one contributing guidelines for all repos so makes sense to be on the site.

Also, looks like all links in these md files linking to md on the site are broken when clicked from github. Not sure what the right way to solve is ...

zramsay:
> shouldn't contributing also be in that docs folder?
I don't think so, especially because when it's in the repo as it currently is, GH renders it as a link for folks making PRs

> do we not want it on the site for some reason?
anyone who plans on contributing will visit the repo they want to contribute to, rather than site (so no, don't think it should be)

re: broken links ... yeah that's the price we pay for having docs in markdown in the repo that are then rendered by a website. Not ideal but it works, mostly

PR #597, opened 2017-07-29
fix dead tendermint types links in specs

zramsay:
- closes https://github.com/zramsay/tendermint/issues/1
- cleans up any other links required for longevity in the face of changing code.

### DIFF
--- a/docs/specs/block-structure.md
+++ b/docs/specs/block-structure.md
@@ -8,36 +8,36 @@ But what exactly is stored in these blocks?
 
 ### Block
 
-A [Block](/docs/specs/tendermint-types#Block) contains:
+A [Block](https://godoc.org/github.com/tendermint/tendermint/types#Block) contains:
 
 * a [Header](#header) contains merkle hashes for various chain states
-* the [Data]((/docs/specs/tendermint-types#Data) is all transactions which are to be processed
+* the [Data](https://godoc.org/github.com/tendermint/tendermint/types#Data) is all transactions which are to be processed
 * the [LastCommit](#commit) > 2/3 signatures for the last block
 
 The signatures returned along with block `H` are those validating block `H-1`.
 This can be a little confusing, but we must also consider that the
-[Header](/docs/specs/tendermint-types#Header) also contains the `LastCommitHash`.
+`Header` also contains the `LastCommitHash`.
 It would be impossible for a Header to include the commits that sign it, as it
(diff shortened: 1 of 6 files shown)


docs/specs/block-structure.md
```diff
@@ -89,13 +88,6 @@ all these stringent requirements. If the total number of voting power for a sing
 than 2/3 of all voting power, then we can finally trust the
 block header, the AppHash, and the proof we got from the ABCI application.
 
-To make this a bit more concrete, you can take a look at a
-[StaticCertifier](https://github.com/tendermint/light-client/blob/develop/rpc/certifier.go#L23)
```
ebuchman: why removing ?

docs/specs/byzantine-consensus-algorithm.md
```diff
@@ -187,9 +189,3 @@ there are no significant network partitions, to avoid situations where two
 conflicting reorg-proposals are signed.
 
 Assuming that the external coordination medium and protocol is robust, it follows that forks are less of a concern than [censorship attacks](#censorship-attacks).
-
-### Revisions
```
ebuchman: Why remove ?

docs/specs/genesis.md
```diff
@@ -16,7 +16,8 @@ NOTE: This does not (yet) specify the application state (e.g. initial distributi
 
 ### Sample genesis.json
 
-This example is from the Basecoin mintnet example ([link to file](https://github.com/tendermint/mintnet/blob/master/examples/basecoin/mach1/core/genesis.json)).
+This example is from the Basecoin mintnet example:
```
ebuchman: need to fix this since mintnet no longer a thing. should have a way to make testnet genesis in tendermint/tools I think

docs/specs/block-structure.md
zramsay: 404

docs/specs/byzantine-consensus-algorithm.md
zramsay: its a header with nothing in it ...

zramsay: this stuff should be in changelog

docs/specs/genesis.md
zramsay: link was dead, so i only removed that and left rest as-is

PR #599, opened 2017-08-01
docs fixes throughout

zramsay:
fix/clarify a bunch of things
closes:
- https://github.com/zramsay/tendermint/issues/3
- https://github.com/zramsay/tendermint/issues/5

### DIFF
--- a/docs/getting-started/#02-first-app.md
+++ b/docs/getting-started/#02-first-app.md
@@ -8,6 +8,8 @@ and just forwards transactions to the application when they need to be validated
 
 In this guide, we show you some examples of how to run an application using Tendermint.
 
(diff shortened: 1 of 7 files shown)


docs/getting-started/#02-first-app.md
```diff
@@ -8,6 +8,8 @@ and just forwards transactions to the application when they need to be validated
 
 In this guide, we show you some examples of how to run an application using Tendermint.
 
+**Note:** It is highly recommended to read the [Using Tendermint Guide](/docs/guides/using-tendermint) prior to working through this tutorial.
```
ebuchman: Hmph. Ideally one wouldn't have to read that first. Do you think its really necessary? How can we reduce this need ?

zramsay: i found it helpful ... it provides context for what one is doing although there's a lot of duplicate stuff between the getting started and guides

PR #604, opened 2017-08-03
Biff up RPC WSClient

melekes:
What's new:
- auto reconnect
- ping/pong
- colored tests
- latency timer embedded in WS client

### DIFF
--- a/rpc/lib/client/ws_client.go
+++ b/rpc/lib/client/ws_client.go
@@ -54,6 +54,8 @@ type WSClient struct {
 	reconnectAfter     chan error            // reconnect requests
 	receiveRoutineQuit chan struct{}         // a way for receiveRoutine to close writeRoutine
 
+	reconnecting bool
+
 	wg  sync.WaitGroup
 	mtx sync.RWMutex
 }
@@ -115,6 +117,16 @@ func (c *WSClient) Stop() bool {
 	return success
 }
 
+// IsReconnecting returns true if the client is reconnecting right now.
+func (c *WSClient) IsReconnecting() bool {
+	return c.reconnecting
+}
+
+// IsActive returns true if the client is running and not reconnecting.
+func (c *WSClient) IsActive() bool {
+	return c.IsRunning() && !c.IsReconnecting()
+}
+
 // Send asynchronously sends the given RPCRequest to the server. Results will
 // be available on ResultsCh, errors, if any, on ErrorsCh.
 func (c *WSClient) Send(ctx context.Context, request types.RPCRequest) error {
@@ -169,6 +181,11 @@ func (c *WSClient) dial() error {
 // backoff.
 func (c *WSClient) reconnect() error {
 	attempt := 0
+
+	c.reconnecting = true
+	defer func() {
+		c.reconnecting = false
+	}()
 
 	for {
 		c.Logger.Info("reconnecting", "attempt", attempt+1)
--- a/rpc/lib/client/ws_client.go
+++ b/rpc/lib/client/ws_client.go
@@ -324,7 +324,7 @@ func (c *WSClient) writeRoutine() {
 		case <-c.readRoutineQuit:
 			return
 		case <-c.Quit:
-			c.conn.WriteMessage(websocket.CloseMessage, []byte{})
+			c.conn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""))
 			return
 		}
 	}
@@ -352,7 +352,7 @@ func (c *WSClient) readRoutine() {
 	for {
 		_, data, err := c.conn.ReadMessage()
 		if err != nil {
-			if !websocket.IsUnexpectedCloseError(err, websocket.CloseGoingAway) {
+			if websocket.IsCloseError(err, websocket.CloseNormalClosure) {
 				return
 			}
 
--- a/rpc/lib/client/ws_client.go
+++ b/rpc/lib/client/ws_client.go
@@ -39,22 +39,23 @@ type WSClient struct {
 	Dialer   func(string, string) (net.Conn, error)
 
 	PingPongLatencyTimer metrics.Timer
-	sentLastPingAt       time.Time
 
 	// user facing channels, closed only when the client is being stopped.
 	ResultsCh chan json.RawMessage
 	ErrorsCh  chan error
 
 	// internal channels
-	send               chan types.RPCRequest // user requests
-	backlog            chan types.RPCRequest // stores a single user request received during a conn failure
-	reconnectAfter     chan error            // reconnect requests
-	receiveRoutineQuit chan struct{}         // a way for receiveRoutine to close writeRoutine
+	send            chan types.RPCRequest // user requests
+	backlog         chan types.RPCRequest // stores a single user request received during a conn failure
+	reconnectAfter  chan error            // reconnect requests
+	readRoutineQuit chan struct{}         // a way for readRoutine to close writeRoutine
 
 	reconnecting bool
 
-	wg  sync.WaitGroup
-	mtx sync.RWMutex
+	wg sync.WaitGroup
+
+	mtx            sync.RWMutex
+	sentLastPingAt time.Time
 
 	// Time allowed to read the next pong message from the server.
 	pongWait time.Duration
@@ -147,8 +148,9 @@ func (c *WSClient) IsActive() bool {
 	return c.IsRunning() && !c.IsReconnecting()
 }
 
-// Send asynchronously sends the given RPCRequest to the server. Results will
-// be available on ResultsCh, errors, if any, on ErrorsCh.
+// Send the given RPC request to the server. Results will be available on
+// ResultsCh, errors, if any, on ErrorsCh. Will block until send succeeds or
+// ctx.Done is closed.
 func (c *WSClient) Send(ctx context.Context, request types.RPCRequest) error {
 	select {
 	case c.send <- request:
@@ -159,8 +161,7 @@ func (c *WSClient) Send(ctx context.Context, request types.RPCRequest) error {
 	}
 }
 
-// Call asynchronously calls a given method by sending an RPCRequest to the
-// server. Results will be available on ResultsCh, errors, if any, on ErrorsCh.
+// Call the given method. See Send description.
 func (c *WSClient) Call(ctx context.Context, method string, params map[string]interface{}) error {
 	request, err := types.MapToRequest("", method, params)
 	if err != nil {
@@ -169,9 +170,8 @@ func (c *WSClient) Call(ctx context.Context, method string, params map[string]in
 	return c.Send(ctx, request)
 }
 
-// CallWithArrayParams asynchronously calls a given method by sending an
-// RPCRequest to the server. Results will be available on ResultsCh, errors, if
-// any, on ErrorsCh.
+// CallWithArrayParams the given method with params in a form of array. See
+// Send description.
 func (c *WSClient) CallWithArrayParams(ctx context.Context, method string, params []interface{}) error {
 	request, err := types.ArrayToRequest("", method, params)
 	if err != nil {
@@ -231,8 +231,8 @@ func (c *WSClient) reconnect() error {
 
 func (c *WSClient) startReadWriteRoutines() {
 	c.wg.Add(2)
-	c.receiveRoutineQuit = make(chan struct{})
-	go c.receiveRoutine()
+	c.readRoutineQuit = make(chan struct{})
+	go c.readRoutine()
 	go c.writeRoutine()
 }
 
@@ -240,7 +240,7 @@ func (c *WSClient) reconnectRoutine() {
 	for {
 		select {
 		case originalError := <-c.reconnectAfter:
-			// wait until writeRoutine and receiveRoutine finish
+			// wait until writeRoutine and readRoutine finish
 			c.wg.Wait()
 			err := c.reconnect()
 			if err != nil {
@@ -310,7 +310,7 @@ func (c *WSClient) writeRoutine() {
 			c.sentLastPingAt = time.Now()
 			c.mtx.Unlock()
 			c.Logger.Debug("sent ping")
-		case <-c.receiveRoutineQuit:
+		case <-c.readRoutineQuit:
 			return
 		case <-c.Quit:
 			c.conn.WriteMessage(websocket.CloseMessage, []byte{})
@@ -321,13 +321,14 @@ func (c *WSClient) writeRoutine() {
 
 // The client ensures that there is at most one reader to a connection by
 // executing all reads from this goroutine.
-func (c *WSClient) receiveRoutine() {
+func (c *WSClient) readRoutine() {
 	defer func() {
 		c.conn.Close()
 		c.wg.Done()
 	}()
 
 	c.conn.SetReadDeadline(time.Now().Add(c.pongWait))
+
 	c.conn.SetPongHandler(func(string) error {
 		c.conn.SetReadDeadline(time.Now().Add(c.pongWait))
 		c.mtx.RLock()
@@ -336,6 +337,7 @@ func (c *WSClient) receiveRoutine() {
 		c.Logger.Debug("got pong")
 		return nil
 	})
+
 	for {
 		_, data, err := c.conn.ReadMessage()
 		if err != nil {
@@ -344,7 +346,7 @@ func (c *WSClient) receiveRoutine() {
 			}
 
 			c.Logger.Error("failed to read response", "err", err)
-			close(c.receiveRoutineQuit)
+			close(c.readRoutineQuit)
 			c.reconnectAfter <- err
 			return
 		}
--- a/rpc/lib/client/ws_client.go
+++ b/rpc/lib/client/ws_client.go
@@ -22,14 +22,11 @@ const (
 	// Time allowed to write a message to the peer.
 	writeWait = 10 * time.Second
 
-	// Time allowed to read the next pong message from the server.
-	pongWait = 30 * time.Second
-
-	// Send pings to server with this period. Must be less than pongWait.
-	pingPeriod = (pongWait * 9) / 10
-
 	// Maximum reconnect attempts
 	maxReconnectAttempts = 25
+
+	defaultPongWait   = 30 * time.Second
+	defaultPingPeriod = (defaultPongWait * 9) / 10
 )
 
 type WSClient struct {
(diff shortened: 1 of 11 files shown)


rpc/lib/client/ws_client_test.go
```diff
@@ -0,0 +1,211 @@
...
+	c := NewWSClient(addr.String(), "/websocket")
+	_, err := c.Start()
+	require.Nil(t, err)
+	c.SetLogger(log.TestingLogger())
+	return c
+}
+
+func call(t *testing.T, method string, c *WSClient) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
melekes: does not make much sense

rpc/lib/rpc_test.go
```diff
@@ -326,6 +353,22 @@ func TestWSHandlesArrayParams(t *testing.T) {
...
+		t.Skip("skipping ping pong in short mode")
+	}
+
+	cl := client.NewWSClient(tcpAddr, websocketEndpoint)
+  cl.SetLogger(log.TestingLogger())
+	_, err := cl.Start()
+	require.Nil(t, err)
+	defer cl.Stop()
+
+	time.Sleep(35 * time.Second)
```
melekes: change ping/pong timeouts so the test does not take forever to finish

rpc/lib/test/integration_test.sh
```diff
@@ -9,6 +9,9 @@ DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 # Change into that dir because we expect that.
 pushd "$DIR"
 
+echo "==> Installing deps"
```
melekes: remove. does not help me really

melekes:
any other comments?

rpc/lib/client/ws_client.go
```diff
@@ -1,160 +1,390 @@
...
+
+// The client ensures that there is at most one reader to a connection by
+// executing all reads from this goroutine.
+func (c *WSClient) receiveRoutine() {
+	defer func() {
+		c.conn.Close()
+		c.wg.Done()
+	}()
+
+	c.conn.SetReadDeadline(time.Now().Add(c.pongWait))
```
ebuchman: is this a duplicate/unecessary line ?

rpc/lib/client/ws_client.go
```diff
@@ -1,160 +1,390 @@
...
+func (c *WSClient) OnStart() error {
+	err := c.dial()
 	if err != nil {
 		return err
 	}
-	wsc.ResultsCh = make(chan json.RawMessage, wsResultsChannelCapacity)
-	wsc.ErrorsCh = make(chan error, wsErrorsChannelCapacity)
-	go wsc.receiveEventsRoutine()
+
+	c.ResultsCh = make(chan json.RawMessage)
```
ebuchman: these are unbuffered. is there a good reason for this to always be the case? should we put this initialization in the constructor and let clients overwrite it with options if they want buffer?

rpc/lib/client/ws_client.go
```diff
@@ -1,160 +1,390 @@
...
+	c.conn.SetPongHandler(func(string) error {
+		c.conn.SetReadDeadline(time.Now().Add(c.pongWait))
+		c.mtx.RLock()
+		c.PingPongLatencyTimer.UpdateSince(c.sentLastPingAt)
+		c.mtx.RUnlock()
+		c.Logger.Debug("got pong")
+		return nil
+	})
+	for {
+		_, data, err := c.conn.ReadMessage()
```
ebuchman: Do we also need a SetReadDeadline before the ReadMessage?

rpc/lib/client/ws_client.go
```diff
@@ -1,160 +1,390 @@
...
+		if err != nil {
+			if !websocket.IsUnexpectedCloseError(err, websocket.CloseGoingAway) {
+				return
+			}
 
-	// Cleanup
-	close(wsc.ResultsCh)
-	close(wsc.ErrorsCh)
+			c.Logger.Error("failed to read response", "err", err)
+			close(c.receiveRoutineQuit)
```
ebuchman: why do we need this ?

rpc/lib/client/ws_client_test.go
```diff
@@ -0,0 +1,211 @@
...
+	// start server
+	h := &myHandler{}
+	s := httptest.NewServer(h)
+
+	c := startClient(t, s.Listener.Addr())
+	defer c.Stop()
+
+	wg.Add(2)
+	go func() {
+		for {
```
ebuchman: worth deduplicating  ?

melekes: probably not

rpc/lib/rpc_test.go
melekes: the problem here is that the server has a hard coded timeout:

```
	wsReadTimeoutSeconds  = 30 // connection times out if we haven't received *anything* in this long, not even pings.
```

and if we wanna test this we need to either wait >30sec or make it an option

rpc/lib/client/ws_client.go
melekes:
```
	readRoutineQuit chan struct{}         // a way for readRoutine to close writeRoutine
```

see WSClient struct. Should I duplicate this?

rpc/lib/client/ws_client.go
melekes: line 331. we do that already

rpc/lib/client/ws_client.go
melekes: -> https://github.com/tendermint/tendermint/pull/604#discussion_r131510952

no, we need this deadline

rpc/lib/client/ws_client.go
```diff
@@ -1,160 +1,390 @@
...
+		if attempt > maxReconnectAttempts {
+			return errors.Wrap(err, "reached maximum reconnect attempts")
+		}
+	}
+}
+
+func (c *WSClient) startReadWriteRoutines() {
+	c.wg.Add(2)
+	c.receiveRoutineQuit = make(chan struct{})
+	go c.receiveRoutine()
```
melekes: good point

rpc/lib/client/ws_client.go
```diff
@@ -1,160 +1,390 @@
...
 		Proxy:   http.ProxyFromEnvironment,
 	}
 	rHeader := http.Header{}
-	con, _, err := dialer.Dial("ws://"+wsc.Address+wsc.Endpoint, rHeader)
+	conn, _, err := dialer.Dial("ws://"+c.Address+c.Endpoint, rHeader)
 	if err != nil {
 		return err
 	}
-	// Set the ping/pong handlers
-	con.SetPingHandler(func(m string) error {
```
melekes: The default handler is enough, I think. By using `send` channel (unbuffered) we implement a simple backoff for any user. This should guaranty us write buffer won't be filled.

rpc/lib/client/ws_client.go
```diff
@@ -1,160 +1,390 @@
...
+func (c *WSClient) IsReconnecting() bool {
+	return c.reconnecting
+}
+
+// IsActive returns true if the client is running and not reconnecting.
+func (c *WSClient) IsActive() bool {
+	return c.IsRunning() && !c.IsReconnecting()
+}
+
+// Send asynchronously sends the given RPCRequest to the server. Results will
```
melekes: yes, it blocks

rpc/lib/client/ws_client.go
```diff
@@ -1,160 +1,390 @@
...
+	// only close user-facing channels when we can't write to them
+	c.wg.Wait()
+	close(c.ResultsCh)
+	close(c.ErrorsCh)
+	return success
+}
+
+// IsReconnecting returns true if the client is reconnecting right now.
+func (c *WSClient) IsReconnecting() bool {
+	return c.reconnecting
```
melekes: no. does it have to be thread-safe?

rpc/lib/client/ws_client.go
```diff
@@ -1,160 +1,390 @@
...
 
-// OnReset implements cmn.BaseService interface
-func (wsc *WSClient) OnReset() error {
-	return nil
+// OnStop implements cmn.Service.
+func (c *WSClient) OnStop() {}
+
+// Stop overrides cmn.Service#Stop. There is no other way to wait until Quit
+// channel is closed.
+func (c *WSClient) Stop() bool {
```
melekes: I want to close ResultsCh and ErrorsCh only when we can't write to them (to avoid panic). There is no other way to wait until Quit channel is closed other than overriding Stop.

rpc/lib/client/ws_client.go
melekes: good question. probably not. if I am a WS client, in 97% I want to get results and process them as fast as possible.

rpc/lib/client/ws_client.go
```diff
@@ -1,160 +1,390 @@
...
 	addr, dialer := makeHTTPDialer(remoteAddr)
-	wsClient := &WSClient{
-		Address:  addr,
-		Dialer:   dialer,
-		Endpoint: endpoint,
-		Conn:     nil,
+	c := &WSClient{
+		Address:              addr,
+		Dialer:               dialer,
+		Endpoint:             endpoint,
```
melekes: like for people who may use this library? we use `"/websocket"` everywhere. don't think we need a validation for that.

melekes: also, adding validation means breaking public API (returning `(WSClient, error)`)

rpc/lib/client/ws_client.go
melekes: it's still hard for me to choose between channels and callbacks. should we use callbacks while delivering results? maybe. google uses callback style in his pubsub https://godoc.org/cloud.google.com/go/pubsub#hdr-Receiving

rpc/lib/rpc_test.go
melekes: I will make `wsReadTimeoutSeconds` an option

rpc/lib/client/ws_client.go
cloudhead: I'd favor channels. In cases like these, callbacks give control to the message producer, channels give control to the message receiver. The later is thus more flexible, ex: with the google method you're forced to process messages out-of-order.

rpc/lib/client/ws_client.go
ebuchman: Ok I guess OnStart will return the error if its missing the `/`. I'll add a comment

rpc/lib/client/ws_client.go
ebuchman: Ah. In other services we use a `Wait()` method that waits on a channel that is closed after receiving on Quit, eg. https://github.com/tendermint/tendermint/blob/master/consensus/state.go#L364

Not sure what's better - I didn't even think of overwriting Stop.

That said, maybe we should consider closing the Quit chan in service before calling impl.OnStop() ? ie. right now its:

```
func (bs *BaseService) Stop() bool {
	if atomic.CompareAndSwapUint32(&bs.stopped, 0, 1) {
		bs.Logger.Info(Fmt("Stopping %v", bs.name), "impl", bs.impl)
		bs.impl.OnStop()
		close(bs.Quit)
		return true
	} else {
		bs.Logger.Debug(Fmt("Stopping %v (ignoring: already stopped)", bs.name), "impl", bs.impl)
		return false
	}
}
```

Is there some good reason why it should happen after ? @jaekwon

rpc/lib/client/ws_client.go
melekes: Thanks

rpc/lib/client/ws_client.go
ebuchman: Ok - what about read concurrency ?

rpc/lib/client/ws_client.go
ebuchman: I mean, why is it outside the for loop ?

rpc/lib/client/ws_client.go
ebuchman: We do it once, but this is in a loop, no ?

rpc/lib/client/ws_client.go
melekes: The original issue is about filled write buffer, reads should not be affected.

rpc/lib/client/ws_client.go
melekes: Do we want reads to be time-limited?

rpc/lib/client/ws_client.go
melekes: > That said, maybe we should consider closing the Quit chan in service before calling impl.OnStop() ?

I am in favor of doing that. Consider this example:

```
func (wsc *wsConnection) OnStop() {
	wsc.BaseService.OnStop()
	if wsc.evsw != nil {
		wsc.evsw.RemoveListener(wsc.remoteAddr)
	}
	wsc.readTimeout.Stop()
	wsc.pingTicker.Stop()
	// The write loop closes the websocket connection
	// when it exits its loop, and the read loop
	// closes the writeChan
}
```

we stop tickers before we send Quit, so in writeRoutine we may access stopped tickers. Now imagine something like channels (not tickers), which may trigger panic if you use them after closing.

rpc/lib/client/ws_client.go
ebuchman: nope you're right. thanks

rpc/lib/client/ws_client.go
ebuchman: nvm

rpc/lib/client/ws_client_test.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (h *myHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	conn, err := upgrader.Upgrade(w, r, nil)
+	if err != nil {
+		panic(err)
+	}
+	for {
+		messageType, _, err := conn.ReadMessage()
+		if err != nil {
```
odeke-em: We'll leak the connection here if we encounter an error, perhaps it doesn't matter in the tests but anyways, the Read mutex protected call below that checks  whether we should close the connection after reading would be good here too, if am not mistaken.

rpc/lib/server/handlers.go
```diff
@@ -349,80 +349,89 @@ const (
...
+		pingPeriod:        defaultWSPingPeriod,
+	}
+	for _, option := range options {
+		option(wsc)
 	}
 	wsc.BaseService = *cmn.NewBaseService(nil, "wsConnection", wsc)
 	return wsc
 }
 
+func WriteWait(writeWait time.Duration) func(*wsConnection) {
```
odeke-em: Just from a quick skim of the code, perhaps we need concurrent protection for these wrapping calls?

rpc/lib/server/handlers.go
```diff
@@ -349,80 +349,89 @@ const (
...
+	return func(wsc *wsConnection) {
+		wsc.writeWait = writeWait
```
odeke-em: ditto

rpc/lib/server/handlers.go
```diff
@@ -349,80 +349,89 @@ const (
...
+	}
+}
+
+func WriteChanCapacity(cap int) func(*wsConnection) {
+	return func(wsc *wsConnection) {
+		wsc.writeChanCapacity = cap
```
odeke-em: ditto and all down

rpc/lib/server/handlers.go
ebuchman: @odeke-em are you familiar with the API options paradigm? https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

That's what we're going for here. Don't think we need locks but we should add documentation to these public functions

rpc/lib/client/ws_client.go
```diff
@@ -60,10 +56,16 @@ type WSClient struct {
...
+	maxReconnectAttempts int
+
+	// Time allowed to write a message to the server. 0 means block until operation succeeds.
+	writeWait time.Duration
+
+	// Time allowed to read the next message from the server. 0 means block until operation succeeds.
+	readWait time.Duration
 
-	// Send pings to server with this period. Must be less than pongWait.
+	// Send pings to server with this period. Must be less than readWait. If 0, no pings will be sent.
```
ebuchman: do we enforce this anywhere? Or just up to user? If later we should document in a place where users will see.

rpc/lib/server/handlers.go
odeke-em: Yes in deed, I am familiar with them and I like them :) https://github.com/odeke-em/mapbox/blob/8bb614ae9e4db26d0a89d9fc1ea5bde442c72fca/options.go#L7-L21

However, my problem here is that these are exposed options ie exported options for potential concurrent misuse

rpc/lib/server/handlers.go
```diff
@@ -349,19 +349,23 @@ const (
...
+
+	// each write times out after this.
+	writeWait time.Duration
+
 	// Connection times out if we haven't received *anything* in this long, not even pings.
-	pongWait time.Duration
+	readWait time.Duration
 
-	// Send pings to server with this period. Must be less than pongWait.
+	// Send pings to server with this period. Must be less than readWait, but greater than zero.
```
ebuchman: enforced anywhere?

rpc/lib/server/handlers.go
odeke-em: Actually never mind, they are used only for initiatialization so no locks as you mentioned, that's the part that confused me in the review.

ebuchman:
I'm going to spend a minute adding comments to all the rpc funcs, will do a final review, and then merge unless I find something I dont understand

rpc/lib/client/ws_client.go
```diff
@@ -348,21 +379,25 @@ func (c *WSClient) readRoutine() {
...
+		c.PingPongLatencyTimer.UpdateSince(t)
+
 		c.Logger.Debug("got pong")
 		return nil
 	})
 
 	for {
+		// reset deadline for every message type (control or data)
+		if c.readWait > 0 {
+			c.conn.SetReadDeadline(time.Now().Add(c.readWait))
```
ebuchman: I thought we didn't want this ? Doesn't it mean we always timeout if a read doesnt come in this time, even if we're not expecting one ? and do the pongs come through here?

rpc/lib/server/handlers.go
odeke-em: Roger that, yap yap, only used in the constructor so doesn't make sense to add extraneous locks.

rpc/lib/server/handlers.go
```diff
@@ -558,15 +542,33 @@ func (wsc *wsConnection) readRoutine() {
...
+	pingTicker := time.NewTicker(wsc.pingPeriod)
+	defer func() {
+		pingTicker.Stop()
+		wsc.baseConn.Close()
+	}()
+
+	// https://github.com/gorilla/websocket/issues/97
+	pongs := make(chan string, 1)
+	wsc.baseConn.SetPingHandler(func(m string) error {
+		select {
```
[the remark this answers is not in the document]
melekes: > How does this actually cause a deadlock if we're just using WriteControl in the PingHandler?

I think a deadlock is a wrong word here because it can lock up to `write timeout`. So it is not a deadlock per se, just temporary lock.

PR #622, opened 2017-08-26
p2p: fully test PeerSet

odeke-em:
* Full test PeerSet and check its concurrent guarantees
* Improve the doc for PeerSet.Has and remove unnecessary
defer for a path that sets a variable, make it fast anyways.
* Parallelize PeerSet tests with t.Parallel()

### DIFF
--- a/p2p/peer_set.go
+++ b/p2p/peer_set.go
@@ -27,6 +27,7 @@ type peerSetItem struct {
 	index int
 }
 
+// NewPeerSet creates a new peerSet with a list of initial capacity of 256 items.
 func NewPeerSet() *PeerSet {
 	return &PeerSet{
 		lookup: make(map[string]*peerSetItem),
@@ -34,7 +35,8 @@ func NewPeerSet() *PeerSet {
 	}
 }
 
-// Returns false if peer with key (PubKeyEd25519) is already set
+// Add returns false if the peer's Key (PubKeyEd25519) is already memoized.
+// If the peer was already added, it returns ErrSwitchDuplicatePeer.
 func (ps *PeerSet) Add(peer *Peer) error {
 	ps.mtx.Lock()
 	defer ps.mtx.Unlock()
@@ -50,13 +52,16 @@ func (ps *PeerSet) Add(peer *Peer) error {
 	return nil
 }
 
+// Has returns true iff the peerset contains
+// the peer referred to by this peerKey.
 func (ps *PeerSet) Has(peerKey string) bool {
 	ps.mtx.Lock()
-	defer ps.mtx.Unlock()
 	_, ok := ps.lookup[peerKey]
+	ps.mtx.Unlock()
 	return ok
 }
 
+// Get looks up a peer by the provided peerKey.
 func (ps *PeerSet) Get(peerKey string) *Peer {
 	ps.mtx.Lock()
 	defer ps.mtx.Unlock()
@@ -68,6 +73,7 @@ func (ps *PeerSet) Get(peerKey string) *Peer {
 	}
 }
 
+// Remove discards peer by its Key, if the peer was previously memoized.
 func (ps *PeerSet) Remove(peer *Peer) {
 	ps.mtx.Lock()
 	defer ps.mtx.Unlock()
@@ -77,8 +83,8 @@ func (ps *PeerSet) Remove(peer *Peer) {
 	}
 
 	index := item.index
-	// Copy the list but without the last element.
-	// (we must copy because we're mutating the list)
+	// Create a new copy of the list but with one less item.
+	// (we must copy because we'll be mutating the list).
 	newList := make([]*Peer, len(ps.list)-1)
 	copy(newList, ps.list)
 	// If it's the last peer, that's an easy special case.
@@ -88,7 +94,7 @@ func (ps *PeerSet) Remove(peer *Peer) {
 		return
 	}
 
-	// Move the last item from ps.list to "index" in list.
+	// Replace the popped item with the last item in the old list.
 	lastPeer := ps.list[len(ps.list)-1]
 	lastPeerKey := lastPeer.Key
 	lastPeerItem := ps.lookup[lastPeerKey]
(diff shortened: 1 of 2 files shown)


p2p/peer_set_test.go
```diff
@@ -65,3 +70,50 @@ func TestAddRemoveMany(t *testing.T) {
...
+	}
+	wg.Wait()
+}
+
+func TestPeerSetGet(t *testing.T) {
+	t.Parallel()
+	peerSet := NewPeerSet()
+	peer := randPeer()
+	var nilPeer *Peer
+	assert.Equal(t, peerSet.Get(peer.Key), nilPeer, "expecting a nil lookup, before .Add")
```
melekes: shouldn't `PeerSet.Get` return `nil`?

odeke-em: It returns a typed nil so (*PeerSet)(nil) , which assert.Equal trips out over unfortunately.

odeke-em:
Oh and also parallelized the tests in `peer_set_test.go` with `t.Parallel()`

p2p/peer_set_test.go
melekes: what about `assert.Nil`?

odeke-em: Aha, good plan, thanks for letting me know. Let me try that.

odeke-em:
Thank you @melekes for the suggestion to use `assert.Nil`, it worked like a charm, PTAL!

p2p/peer_set.go
```diff
@@ -50,10 +50,12 @@ func (ps *PeerSet) Add(peer *Peer) error {
 	return nil
 }
 
+// Has returns true iff the peerset contains
+// the peer referred to by this peerKey.
 func (ps *PeerSet) Has(peerKey string) bool {
```
ebuchman: Would you mind adding proper comments to all the functions in this file, please ?

p2p/peer_set_test.go
```diff
@@ -30,16 +34,17 @@ func TestAddRemoveOne(t *testing.T) {
...
 
-	peerSet.Remove(peer)
-	if peerSet.Has(peer.Key) {
-		t.Errorf("Failed to remove peer")
-	}
-	if peerSet.Size() != 0 {
-		t.Errorf("Failed to remove peer and decrement size")
+	// Invoke Remove more than once to ensure that the stored
+	// value is purged anyways i.e take the different paths.
+	for i := 0; i < 3; i++ {
```
ebuchman: Why 3 times?  Wouldn't 2 be enough ? Also, should we test the different case of the peer being at the end of the list ?

p2p/peer_set_test.go
```diff
@@ -65,3 +70,49 @@ func TestAddRemoveMany(t *testing.T) {
...
+	peerSet := NewPeerSet()
+	peer := randPeer()
+	if err := peerSet.Add(peer); err != nil {
+		t.Fatalf("Failed to add new peer: %v", err)
+	}
+
+	var wg sync.WaitGroup
+	for i := 0; i < 10; i++ {
+		// Add them asynchronously to test the
+		// concurrent guarantees of our APIs.
```
ebuchman: Is this really testing anything useful in the "AddDuplicate" case ?

p2p/peer_set_test.go
```diff
@@ -65,3 +70,49 @@ func TestAddRemoveMany(t *testing.T) {
...
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
```
ebuchman: Same Q - what are we actually testing by running concurrent routines here? Perhaps we should Set in one routine and Get in another ?

p2p/peer_set.go
odeke-em: For sure, great suggestion, thanks!

p2p/peer_set_test.go
odeke-em: So 3 times was a value out of the air, nothing too particular to it.

p2p/peer_set_test.go
odeke-em: What I had in mind was ensuring that we having at any point multiple goroutines all trying to add duplicates: just 4 goroutines going on might not cut it, so I chose 10.

p2p/peer_set_test.go
odeke-em: Just one doing a set in one routine and get in another would be too deterministic and non-clashing IMO. However, I like the idea of some Set going on in different goroutines, then Get in the others.

p2p/peer_set_test.go
ebuchman: Maybe it'd be more effective if we ran all the Add in a go-routine, and counted the results, and assert they're all the DuplicatePeer error except one.

p2p/peer_set_test.go
ebuchman: Should we do that?

ebuchman: Should we do that?

p2p/peer_set.go
odeke-em: I added comments to all the functions in this file.

p2p/peer_set_test.go
odeke-em: So currently, that's what the test does. Counting the tests in separate goroutines requires synchronization for the counting which might make them unnecessarily complex(e.g a channel collecting results, then later counting each) IMO, and at the beginning I test that no error occurs the first time. However, if that eases reading I'll update it and after review we can decide which one we wanna pick :)

odeke-em: That was actually a great suggestion that I realized while writing the test, thank you @ebuchman!

odeke-em: By the above, I mean I updated the AddDuplicate case.

p2p/peer_set_test.go
odeke-em: @ebuchman now coming back to think of it, doing Gets and Sets in separate goroutines just serves as perhaps a test for race detector, but not actual actionable tests for us because we won't have reliable assertions to make e.g
```go
go func() {
  // Reads here
  // Can't guarantee the number of failed or successful reads
  // as that depends on at least one Set running before majority
  // of the reads here. What do we assert? A percentage of successful
  // reads?
}()

go func() {
   // Writes here
   // Can guarantee that we get a duplicate here
}()
```

odeke-em:
Thanks for the review and merge @ebuchman!

PR #623, opened 2017-08-28
cmd: don't load config for version command. closes #620

### LINKED ISSUE #620: version command should not check the config file
BUG REPORT:

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source): `0.10.2-'b467515` and `0.10.3-'8d76408`

**Environment**:
- **OS** (e.g. from /etc/os-release): MacOS Sierra
- **Install tools**: download zip from github.com/tendermint/tendermint
- **Others**:

**What happened**:
```
> ./tendermint version
ERROR: Failed to parse default log level (pair *:notice, list *:notice): Expected either "info", "debug", "error" or "none" level, given notice
```

but 

```
 > ./tendermint --home ./.ethermint/tendermint version
0.10.2-'b467515
```
IMHO, the `--home` option should not be necessary (including defaults, i.e. the version should show even if the home is not properly setup or not available where expected)

see above.

ebuchman:
https://github.com/tendermint/tendermint/issues/620

### DIFF
--- a/cmd/tendermint/commands/root.go
+++ b/cmd/tendermint/commands/root.go
@@ -38,6 +38,9 @@ var RootCmd = &cobra.Command{
 	Use:   "tendermint",
 	Short: "Tendermint Core (BFT Consensus) in Go",
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) (err error) {
+		if cmd.Name() == versionCmd.Name() {
+			return nil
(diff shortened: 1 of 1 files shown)


ethanfrey:
Looks fine for me.  No idea why it is important, but as long as version remains simple, it is fine.

Issue #628, opened 2017-08-30
config improvements


ebuchman:
- [x] we should return an error if a config element is unknown (rather than ignore it)
- [x] add some new elements to the p2p config (#636):
  - max_msg_packet_payload_size
  - send_rate
  - recv_rate
- [ ] we should add some function, eg. `config.Parse()` to check the validity of all arguments. Ie that ints are positive, that strings are from the allowed set, etc.
- [x] separate config files and data by creating 2 subfolders (config, data) (#556, #709)
- [x] config docs (#705, #733)

odeke-em:
Some questions:
a) By `we should return an error if a config element is unknown (rather than ignore it)` do you mean while unmarshaling/parsing config? While doing the `mapstructure` --> CLI args parsing? Or do you mean if a field is set and we can't find it?
b) 
```
 add some new elements to the p2p config:
```

- max_msg_packet_payload_size: int?
- send_rate: float?
- recv_rate: float?

ebuchman:
a) dont care too much, as long as we catch it before tendermint starts and provide a useful error to the user

b)

send_rate: https://github.com/tendermint/tendermint/blob/master/p2p/connection.go#L31
recv_rate: https://github.com/tendermint/tendermint/blob/master/p2p/connection.go#L34
payload_size: https://github.com/tendermint/tendermint/blob/master/p2p/connection.go#L678

odeke-em:
For the sake of clarity, I might be repeating myself but today on the Thursday company call, a new member had trouble using
```shell
tendermint --home ~/.tendermint node
```
where `~/.tendermint` was non-existent and their tendermint had never been initialized. We should have thrown an error and told them that `~/.tendermint` doesn't exist.

odeke-em:
Another requirement raised offline by @ebuchman. Let's make sure that:
- priv_validator.json
- genesis.json
files are all valid
after we've added the `--verify` flag

ebuchman:
punt to next release for more code review

ebuchman:
Closing this since most PRs are merged, and only one is left open that might need some cleanup

Issue #625, opened 2017-08-29
mempool: speed up broadcastTxRoutine


ebuchman:
`broadcastTxRoutine` does expensive interface handling every loop and only sends one tx per message.

We should allow it to batch txs per peer.Send.

odeke-em:
I can work on this, thanks for the suggestion @ebuchman.

melekes:
@odeke-em I think at least the last part was done in https://github.com/tendermint/tendermint/pull/632. not sure about the "expensive interface handling"

odeke-em:
Nice catch @melekes! The interface handling I think can't go away since we do
https://github.com/tendermint/tendermint/pull/632/files#diff-faa5920fe04b5bb0d81cec8ad247ccdcR152
<img width="1252" alt="screen shot 2017-09-22 at 9 32 48 pm" src="https://user-images.githubusercontent.com/4898263/30769798-9bc63eb2-9fdd-11e7-9de1-aa6561a4f68a.png">

ebuchman:
think this requires more analysis on whats causing the slowdown and if TxsMessage is really the right solution

ValarDragon:
Batching wasn't done. (I'm not sure why, reading into the conversation in the previous PR) Currently we are only sending 1 tx per message, and then waiting for the response. The reason why batching is such a huge performance win is because we wait for the recipient to send an ACK that they've received it (TCP).

ebuchman:
Yeh we should probably bring that back. I think it was closed under suggestion by @jaekwon that there might be some other slowness we're missing and we should find that first, but seems batching would be a clear benefit and based on my benchmarking and profiling seems it provides real value. Not a launch priority per-se though.

ValarDragon:
> `broadcastTxRoutine` does expensive interface handling every loop

Local benchmarks show that a single type assertion took `0.63ns` (though its possible that there was some caching going on), but regardless I don't think that was a significant cause of the slowness. (It was probably purely network delay)

PR #632, opened 2017-08-31
mempool: TxsMessage for sending many txs at once

ebuchman:
Allow mempool reactor to broadcast a batch of txs at a time

### DIFF
--- a/mempool/reactor.go
+++ b/mempool/reactor.go
@@ -9,6 +9,7 @@ import (
 	abci "github.com/tendermint/abci/types"
 	wire "github.com/tendermint/go-wire"
 	"github.com/tendermint/tmlibs/clist"
+	"github.com/tendermint/tmlibs/log"
 
 	cfg "github.com/tendermint/tendermint/config"
 	"github.com/tendermint/tendermint/p2p"
@@ -38,6 +39,12 @@ func NewMempoolReactor(config *cfg.MempoolConfig, mempool *Mempool) *MempoolReac
 	}
 	memR.BaseReactor = *p2p.NewBaseReactor("MempoolReactor", memR)
 	return memR
+}
+
+// SetLogger sets the Logger on the reactor and the underlying Mempool.
+func (memR *MempoolReactor) SetLogger(l log.Logger) {
+	memR.Logger = l
+	memR.Mempool.SetLogger(l)
 }
 
 // GetChannels implements Reactor.
@@ -129,7 +136,7 @@ func (memR *MempoolReactor) broadcastTxRoutine(peer Peer) {
 		batchSize = 1 // TODO: enforce >1 when we load config ?
 	}
 
-	// next tx in mempool and possible list of next txs
+	// Next tx in mempool and possible list of next txs
 	var next *clist.CElement
 	memTxs := make(types.Txs, batchSize)
 	for {
@@ -144,7 +151,7 @@ func (memR *MempoolReactor) broadcastTxRoutine(peer Peer) {
 		}
 		memTx := next.Value.(*mempoolTx)
 
-		// make sure the peer is up to date
+		// Make sure the peer is up to date
 		height := memTx.Height()
 		if peerState_i := peer.Get(types.PeerStateKey); peerState_i != nil {
 			peerState := peerState_i.(PeerState)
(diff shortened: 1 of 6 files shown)


mempool/reactor.go
```diff
@@ -113,7 +124,14 @@ func (memR *MempoolReactor) broadcastTxRoutine(peer Peer) {
 		return
 	}
 
+	batchSize := memR.config.TxBatchSize
+	if batchSize <= 0 {
```
melekes: this is usually being handled during config parsing.

mempool/reactor_test.go
```diff
@@ -0,0 +1,123 @@
...
+	case <-timer:
+		t.Fatal("Timed out waiting for txs")
+	case <-done:
+	}
+}
+
+// wait for all txs on a single mempool
+func _waitForTxs(t *testing.T, wg *sync.WaitGroup, txs types.Txs, mempool *Mempool) {
+
+	for mempool.Size() != len(txs) {
```
melekes: infinite loop?

melekes: ah, we're waiting for N txs

mempool/reactor.go
ebuchman: where/how? is there a nice way to enforce constraints on config values when loaded?

melekes: I was thinking about some function like `Config#Parse()`

ebuchman: added this to https://github.com/tendermint/tendermint/issues/628

ebuchman:
Here's a benchmark, with high logging on, to run through 100k txs:

```
batchSize 1 time 35.542584529s
batchSize 10 time 20.808908968s
batchSize 100 time 16.290339121s
batchSize 1000 time 17.166387044s
batchSize 10000 time 17.459444369s
batchSize 20000 time 17.52522394s
```

So it cuts the time by more than 1/2. Without logging, the gain isn't as significant - we save about 1/4:

```
batchSize 1 time 12.503367204s
batchSize 10 time 8.905486953s
batchSize 100 time 8.467260415s
batchSize 1000 time 8.397051062s
batchSize 10000 time 8.912131556s
batchSize 20000 time 9.385051452s
```

ebuchman:
With some tweaking of the p2p parameters (ie. higher send/recv rates, and larger msg packet size), we can do even better. With logging:

```
batchSize 1 time 32.661843574s
batchSize 10 time 21.963164531s
batchSize 100 time 20.697038597s
batchSize 1000 time 15.027744323s
batchSize 10000 time 16.214248785s
batchSize 20000 time 14.47589791s
```
and without:

```
batchSize 1 time 10.356906551s
batchSize 10 time 5.367920871s
batchSize 100 time 4.980086614s
batchSize 1000 time 5.091395907s
batchSize 10000 time 4.902664069s
batchSize 20000 time 4.619128274s
```

ebuchman:
And if we look at the profiler, we see a 10x reduction in the amount of time spent in the broadcastTxRoutine function.

BatchSize of 1:

```
     390ms      7.73s (flat, cum) 23.81% of Total
```

BatchSize of 100:

```
      30ms      350ms (flat, cum)  2.60% of Total
```

I think we've discussed before that it's cheaper to make one big send than many small ones, and this would be another reflection of that.

mempool/reactor.go
```diff
@@ -134,16 +160,40 @@ func (memR *MempoolReactor) broadcastTxRoutine(peer Peer) {
...
+		}
+
+		var msg MempoolMessage
+		if nTxs > 1 {
+			msg = &TxsMessage{Txs: memTxs[:nTxs]}
+		} else {
+			msg = &TxMessage{Tx: memTxs[0]}
+		}
+
 		success := peer.Send(MempoolChannel, struct{ MempoolMessage }{msg})
```
melekes: what about backword-compatibility? will old versions crash upon receiving MempoolMessage with `Txs`?

ebuchman: They shouldn't. See https://github.com/tendermint/tendermint/blob/master/mempool/reactor.go#L74.

We could probably do with a test for this. 

That said according to Jae's comments this might be abandoned if we can figure out the performance issue ...

melekes: I see. Thanks

odeke-em:
@ebuchman can we tag this PR as the fix for #625? @melekes just caught that in a comment https://github.com/tendermint/tendermint/issues/625#issuecomment-331607303 and I hadn't seen or noticed that this PR fixes that issue.

ebuchman:
Follow https://github.com/tendermint/tendermint/pull/632

zramsay:
@ebuchman did you mean another issue/PR? #632 is this one

ebuchman:
Woops - https://github.com/tendermint/tendermint/issues/625

ValarDragon:
I think we should definitely do this, and consider using it at launch. Its going to be a p2p breaking change otherwise. 

I'm significantly less concerned with the local nodes case than I am with the many nodes case. Then a tx send is over the internet. Batching multiple together seems like it would be a clear speedup due to network delay. In general, its good to send all the data you want to send at once, as the operating system will handle optimizing packet sizes for network conditions / error correction. Since we're in TCP, there are ACKS as well, so doing multiple sends here further slows it down, if more bytes would have been packed into a single packet. (Keep in mind that most txs are probably around 100 bytes.) 

I think we can reuse alot of the code here as well, though perhaps we can make the only gossip option "batched".

ebuchman:
Yes I tend to agree this would be valuable, though I'd love to see it done after we have protocol versioning so we can practice a p2p upgrade - let's take it up in #625

ValarDragon:
Oh I forgot that issue existed lol

Issue #204, opened 2016-04-11
gossip block parts rarest first


ebuchman:

adrianbrink:
What do you mean?

ebuchman:
Updating openning comment.

ebuchman:
An early attempt is here: https://github.com/tendermint/tendermint/pull/633.

We don't need it for launch, its a performance optimization, we can revisit at a later time and just leave this issue open :)

PR #633, opened 2017-08-31
WIP: Propagate block parts rarest first

ebuchman:
Still need to write tests

### DIFF
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -232,7 +232,10 @@ func (conR *ConsensusReactor) Receive(chID byte, src *p2p.Peer, msgBytes []byte)
 		case *ProposalPOLMessage:
 			ps.ApplyProposalPOLMessage(msg)
 		case *BlockPartMessage:
-			ps.SetHasProposalBlockPart(msg.Height, msg.Round, msg.Part.Index)
+			// NOTE: this may update abundance before checking part validity,
+			// enabling an adversary to mess with rarest first.
+			// can solve by tracking peer states block parts bitarrays in ConsensusState
+			conR.updateBlockPartAbundance(ps, msg.Height, msg.Round, msg.Part.Index)
 			conR.conS.peerMsgQueue <- msgInfo{msg, src.Key}
 		default:
 			conR.Logger.Error(cmn.Fmt("Unknown message type %v", reflect.TypeOf(msg)))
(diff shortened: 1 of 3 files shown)


melekes:
What was the reason to do this?

ebuchman:
Performance. I originally wrote this code while running thesis experiments last year. I couldn't really see a performance improvement at the time, but with recent testnet work thought I would resurrect the code and revisit. In theory it should help

types/part_set.go
```diff
@@ -274,3 +279,74 @@ func (ps *PartSet) StringShort() string {
...
+	defer pc.mtx.Unlock()
+	pc.counts[index] += 1
+}
+
+// Pick the first index from the bit array with the lowest count, increment the index,
+// and pass it to the setFunc.
+// Index must be incremented so that multiple concurrent calls don't pick the same value,
+// and passed to setFunc so the abundance doesnt get accidentally incremented again
+// Assumes possibleParts.Size() == len(pc.counts)
+func (pc *PartCounts) PickRarest(possibleParts *cmn.BitArray, setFunc func(int)) int {
```
melekes: can't we use here some tree (like binary heap) where the rarest element would always be on top?

consensus/reactor.go
```diff
@@ -425,14 +461,21 @@ OUTER_LOOP:
 
 		// Send proposal Block parts?
 		if rs.ProposalBlockParts.HasHeader(prs.ProposalBlockPartsHeader) {
-			if index, ok := rs.ProposalBlockParts.BitArray().Sub(prs.ProposalBlockParts.Copy()).PickRandom(); ok {
+			possibleParts := rs.ProposalBlockParts.BitArray().Sub(prs.ProposalBlockParts.Copy())
+			if !possibleParts.IsEmpty() {
+				index := rs.ProposalBlockParts.Abundance.PickRarest(possibleParts, func(setIndex int) {
```
melekes: why pick func need to accept `setFunc`? can't we call setFunc later if the `Send` succeeds?

odeke-em:
@melekes this is the issue that describes its utility https://github.com/tendermint/tendermint/issues/204.
@ebuchman perhaps we should also tag this in the body of the commit message as
`Fixes #204` and also in this issue's description?

odeke-em:
Also seems like we need to fix a merge conflict

consensus/reactor.go
odeke-em: Yeah, @ebuchman puts a TODO comment in the body of this code wondering what should we do if send fails.

odeke-em:
@ebuchman aside from the tests, we need to figure out a way to benchmark/measure our throughput and overall speed.

types/part_set.go
```diff
@@ -274,3 +279,74 @@ func (ps *PartSet) StringShort() string {
...
+	if check() {
+		pc.counts[index] += 1
+	}
+}
+
+// panics if index >= len(pc.counts)
+func (pc *PartCounts) IncrementIndex(index int) {
+	pc.mtx.Lock()
+	defer pc.mtx.Unlock()
+	pc.counts[index] += 1
```
cloudhead: Can we use `atomic.AddUInt32` here and get rid of the lock?

ebuchman: Believe so - this code can probably be improved a bunch

types/part_set.go
ebuchman: yep believe so!

ebuchman:
closing for issue https://github.com/tendermint/tendermint/issues/204

PR #634, opened 2017-08-31
p2p: put maxMsgPacketPayloadSize, recvRate, sendRate in config

ebuchman:


config/config.go
```diff
@@ -224,16 +224,28 @@ type P2PConfig struct {
...
-		AddrBookStrict:       true,
-		MaxNumPeers:          50,
-		FlushThrottleTimeout: 100,
+		ListenAddress:           "tcp://0.0.0.0:46656",
+		AddrBook:                "addrbook.json",
+		AddrBookStrict:          true,
+		MaxNumPeers:             50,
+		FlushThrottleTimeout:    100,
+		MaxMsgPacketPayloadSize: 1024,   // 1 kB
+		SendRate:                512000, // 500 kB/s
```
odeke-em: Minor nit, could we use the defaults in `p2p/connection.go`

ebuchman: We could, but we'd have circular imports.

We do import `types` to use a default value, but I really don't like that either. Ideally we clear the default values out of `p2p` soon and put them all here. Does that sound right?

odeke-em: Oh crap, my bad, for some reason it seemed like they were in the same package, thanks for letting me know. The only thing left is the commit message request then my LGTM is in.

PR #636, opened 2017-08-31
commands, config: tally and report unknown fields in config

odeke-em:
Updates #628

With new flag `--check-config`, we can now tally fields from
the config file mapped to Go struct by tag "mapstructure",
with those parsed out by viper and report an error if we
find flags/fields that aren't accepted in the config.

Note: We use the `--check-config` flag to maintain backward/forward
compatibility with different versions of the file, as raised
by @ebuchman at
https://github.com/tendermint/tendermint/pull/636#issuecomment-327011405

* Exhibit:
Given config in:
```shell
proxy_app = "tcp://127.0.0.1:46658"
moniker = "anonymous"
fast_sync = true
db_backend = "leveldb"
log_level = "state:info,*:error"

[rpc]
laddr = "tcp://0.0.0.0:46657"

[p2p]
laddr = "tcp://0.0.0.0:46656"
seeds = ""
trex = "hey"
picked = "bonjour"
```

Running:
```shell
$ tendermint --home ~/.tendermint node --config
ERROR: We've detected these unknown flags in your config file
p2p:
  picked
  trex

```

### DIFF
--- a/cmd/tendermint/commands/root.go
+++ b/cmd/tendermint/commands/root.go
@@ -19,14 +19,17 @@ var (
 
 func init() {
 	RootCmd.PersistentFlags().String("log_level", config.LogLevel, "Log level")
+	RootCmd.PersistentFlags().Bool(verifyFlag, false, "if set, verifies that all the fields in the config file match up with those in the main struct")
 }
 
 // ParseConfig retrieves the default environment configuration,
 // sets up the Tendermint root and ensures that the root exists
 func ParseConfig() (*cfg.Config, error) {
-	conf := cfg.DefaultConfig()
-	err := viper.Unmarshal(conf)
+	conf, err := handleConfigVerification(cfg.DefaultConfig())
 	if err != nil {
+		return nil, err
+	}
+	if err := viper.Unmarshal(conf); err != nil {
 		return nil, err
 	}
 	conf.SetRoot(conf.RootDir)
--- a/cmd/tendermint/commands/root_test.go
+++ b/cmd/tendermint/commands/root_test.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"strconv"
 	"testing"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -100,4 +101,56 @@ func TestRootConfig(t *testing.T) {
 		assert.Equal(tc.maxPeer, config.P2P.MaxNumPeers, i)
 	}
 
+}
+
+func TestParseInvalidConfig(t *testing.T) {
+	assert := assert.New(t)
+	cases := []struct {
+		cvals   map[string]string
+		verify  bool
+		wantErr string
+	}{
+		// Unwanted flags in the global "scope/namespace"
+		{verify: true, wantErr: "abc", cvals: map[string]string{"fast_sync": "false", "abc": "de"}},
+		{verify: false, wantErr: "", cvals: map[string]string{"fast_sync": "false", "la": "CA"}},
+		{verify: false, wantErr: "", cvals: map[string]string{"moniker": "foo"}},
+		{verify: false, wantErr: "", cvals: nil},
+
+		// No invalid fields present, despite the verify flag
+		{verify: true, wantErr: "", cvals: map[string]string{"moniker": "foo"}},
+		{verify: true, wantErr: "", cvals: nil},
+
+		// Unwanted flags in the sub commands
+		{
+			verify: true, wantErr: "rpc:\n\taseed\n\tbanx\n",
+			cvals: map[string]string{"rpc.aseed": "bonjour", "rpc.banx": "bar"},
+		},
+		{verify: false, wantErr: "", cvals: map[string]string{"rpc.aseed": "bonjour"}},
+	}
+
+	for i, tt := range cases {
+		viper.Reset()
+		if tt.verify {
+			viper.Set(verifyFlag, "true")
+		}
+		for k, v := range tt.cvals {
+			viper.Set(k, v)
+		}
+
+		// To avoid clutter in the current directory,
+		// use the designated defaultRoot at HOME
+		viper.Set("HOME", defaultRoot)
+
+		_, err := ParseConfig()
+		if tt.wantErr != "" {
+			if err == nil {
+				t.Errorf("#%d: want non-nil err", i)
+			} else if g, w := err.Error(), tt.wantErr; !strings.Contains(g, w) {
+				t.Errorf("#%d: got=%q wantSubstring=%q", i, g, w)
+			}
+			continue
+		}
+
+		assert.Nil(err, "#%d", i)
+	}
 }
--- a/cmd/tendermint/commands/util.go
+++ b/cmd/tendermint/commands/util.go
@@ -1,0 +1,221 @@
+package commands
+
+import (
+	"bytes"
+	"fmt"
+	"reflect"
+	"sort"
+	"strings"
+
+	"github.com/spf13/viper"
+
+	cfg "github.com/tendermint/tendermint/config"
+)
+
+// indexTaggedStructFields walks through a struct, looking up those
+// fields tagged by the requested tagName recursively flattening
+// anonymous struct fields by calling indexTaggedStructFields
+// on them, only if the ",squash" sub-tag is applied to the fields.
+// For example, given:
+//  type A struct {
+//    Foo int `json:"foo"`
+//    Bar string `mapstructure:"bar"`
+//    Baz string `mapstructure:"baz"`
+//    B   *A `mapstructure:"b,squash"`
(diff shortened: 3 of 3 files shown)


ebuchman:
Question: should we be worried about future/backwards compatibility with this? Ie if the config struct changes, an old/new file could fail with a new/old version of tendermint.

Or maybe we make it a command to just `tendermint init --verify` or something which checks all files are properly formatted and all config options match up? 

Also wondering if this is something that might be useful to PR into Viper itself ?

odeke-em:
```
Question: should we be worried about future/backwards compatibility with this? Ie if the config struct changes, an old/new file could fail with a new/old version of tendermint.
```
That's a great question. Thinking about it, in deed, that's gonna cause future compatibility problems. But as you've mentioned in your follow up suggestion to add the `--verify` flag, that cuts the deal. It makes sense to have a strict mode for compatibility violating behaviors.

We had talked offline about the PR into Viper, and the `--verify` behavior is the right direction towards getting that back into Viper. However, I think let's first roll with it within Tendermint and once we have some solid thoughts on how we can do it, make that PR to Viper as we had previously talked about.

odeke-em:
Requires https://github.com/tendermint/tmlibs/pull/47 to be merged in first.

ebuchman:
Let's not put this in the upstream cli. Can we do it as a global hidden option on the tendermint command here? Maybe `--check-config`

cmd/tendermint/commands/root_test.go
```diff
@@ -103,3 +104,55 @@ func TestRootConfig(t *testing.T) {
 	}
 
 }
+
+func TestParseInvalidConfig(t *testing.T) {
+	assert := assert.New(t)
+	cases := [...]struct {
```
melekes: `[]` should work too

cmd/tendermint/commands/root.go
```diff
@@ -21,19 +21,54 @@ func init() {
...
+// and if not, returns the passed in config unfiltered and unchecked
+// otherwise it tallies up the values in the config file with the fields
+// of the config struct itself.
+// See https://github.com/tendermint/tendermint/issues/628.
+func handleConfigVerification(conf *cfg.Config) (*cfg.Config, error) {
+	// TODO: Move this step into viper and push up stream.
+	if !viper.GetBool(cli.VerifyFlag) {
+		return conf, nil
+	}
+	tagName := "mapstructure"
```
melekes: why create a var? and not inline the string

cmd/tendermint/commands/util.go
```diff
@@ -0,0 +1,160 @@
...
+	"sort"
+	"strings"
+)
+
+// indexTaggedStructFields walks through a struct, looking up
+// those fields tagged by the requested tagName e.g:
+//    Foo int `json:"foo"`
+//    Bar string `mapstructure:"bar"`
+// recursively flattening anonymous struct fields by calling indexTaggedStructFields
+// on them.
```
melekes: so how the result will look like? `{tx.size: true}`?

cmd/tendermint/commands/util.go
```diff
@@ -0,0 +1,160 @@
...
+
+func extractTag(tagName string, f reflect.StructField) (theTagName string, omitEmpty, squash bool) {
+	tag := f.Tag.Get(tagName)
+	splits := strings.Split(tag, ",")
+	if len(splits) == 0 {
+		return "", false, false
+	}
+	if len(splits) > 1 {
+		switch splits[1] {
+		case "omitempty":
```
melekes: what if both of them are present? `omitempty,squash`

cmd/tendermint/commands/util.go
```diff
@@ -0,0 +1,160 @@
...
+	"strings"
+)
+
+// indexTaggedStructFields walks through a struct, looking up
+// those fields tagged by the requested tagName e.g:
+//    Foo int `json:"foo"`
+//    Bar string `mapstructure:"bar"`
+// recursively flattening anonymous struct fields by calling indexTaggedStructFields
+// on them.
+func indexTaggedStructFields(v interface{}, tagName string) map[string]interface{} {
```
melekes: have you tried to search for such func? maybe it exists in go or viper?

cmd/tendermint/commands/util.go
```diff
@@ -0,0 +1,160 @@
...
+// with those in gotFieldsMap and
+// reporting an error if there are fields in
+// gotFieldsMap that don't match up.
+//
+// A usecase is to report fields that were set in
+// the config file for a command but are unexpected.
+func filterOutUnknownFields(want interface{}, gotFieldsMap map[string]interface{}, tagName string, excusedFlags map[string]map[string]bool) error {
+	wantFieldsMap := indexTaggedStructFields(want, tagName)
+	unpermittedFields := make(map[string][]string)
+	globalScope := "global"
```
melekes: maybe this should be a const?

cmd/tendermint/commands/util.go
```diff
@@ -0,0 +1,160 @@
...
+// filterOutUnknownFields tallies fields
+// retrieved by parsing out fields in want
+// with those in gotFieldsMap and
+// reporting an error if there are fields in
+// gotFieldsMap that don't match up.
+//
+// A usecase is to report fields that were set in
+// the config file for a command but are unexpected.
+func filterOutUnknownFields(want interface{}, gotFieldsMap map[string]interface{}, tagName string, excusedFlags map[string]map[string]bool) error {
+	wantFieldsMap := indexTaggedStructFields(want, tagName)
```
melekes: don't you think it will be better to extract this call and pass `wantFieldsMap` to `filterOutUnknownFields`?

```
func filterOutUnknownFields(wantFieldsMap map[string]interface{}, gotFieldsMap map[string]interface{}, excusedFlags map[string]map[string]bool) error {
```

melekes:
The code looks a bit complicated to me. I've added a few comments. Maybe we should add more comments or split existing functions into smaller functions.

cmd/tendermint/commands/root_test.go
odeke-em: Thanks for the suggestion @melekes, however in this case, when I get a failed test I'd like to easily jump to the index that failed, hence the `[...]` and subsequent `2: {}`

cmd/tendermint/commands/root.go
odeke-em: It was just an oversight. Thanks @melekes, yap I can inline that.

cmd/tendermint/commands/util.go
odeke-em: So far I haven't seen an inclusion of both in "mapstructure" as "squash" is the equivalent "omitempty" and the idea is that they are mutually exclusive.

cmd/tendermint/commands/util.go
odeke-em: I agree.

cmd/tendermint/commands/util.go
odeke-em: Not in Go at least, nor in Viper from what I've checked in their godoc https://godoc.org/github.com/spf13/viper

cmd/tendermint/commands/util.go
odeke-em: I could, however we still need to pass in the tagName for the internal caller inside https://github.com/orijtech/tendermint/blob/11f3c811332779f438f45322b433b399711ca748/cmd/tendermint/commands/util.go#L93.
If that weren't the case, I agree, I'd actually be nicer to extract it out.

cmd/tendermint/commands/root_test.go
odeke-em: I'll update this as agreed upon on in the meeting.

ebuchman:
lets move this to next non breaking release so we can do more review

zramsay:
what are we doing with this?

cmd/tendermint/commands/util.go
ebuchman: We don't use both anywhere but I don't see why they would be mutually exclusive? 

`squash` brings a substruct up to the current level in json serialization. `omitempty` leaves it out all together. I could imagine using both.

cmd/tendermint/commands/util.go
ebuchman: Can't we simplify this by having `indexTaggedStructFields` grab all sub structs and put them under keys like `p2p.laddr` ? Then we don't need to call `indexTaggedStructFields` in `filterOutUnknownFields` and don't care about vars like `specializedCommand`. Then we can do what Anton suggested too.

zramsay:
do we want to renew this PR now that #792 is merged?

zramsay:
I think it's a lot of extra code for not a lot of functionality gain. And I realized that now with #792, the whole `config.toml` is written with default, so adding a bad field is much less likely (the problem this PR was trying to solve).

PR #638, opened 2017-08-31
Upgrade state_test.go

adrianbrink:
I've capped the line length because otherwise, it is annoying to have split screen vim/emacs on a 13" screen.

Also, I've extracted common setup in the test file.

### DIFF
--- a/state/state_test.go
+++ b/state/state_test.go
@@ -19,8 +19,8 @@ import (
 	"github.com/tendermint/tendermint/types"
 )
 
+// setupTestCase does setup common to all test cases
 func setupTestCase(t *testing.T) (func(t *testing.T), dbm.DB, *State) {
-
 	config := cfg.ResetTestRoot("state_")
 	stateDB := dbm.NewDB("state", config.DBBackend, config.DBDir())
 	state := GetState(stateDB, config.GenesisFile())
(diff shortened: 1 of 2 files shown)


ebuchman:
Ahhh. Thanks Adrian. I added many of these comments in another PR already: https://github.com/tendermint/tendermint/pull/618/files

We should probably make a more concerted effort to separate out comment commits so they can be merged right away, and not let them linger with other changes waiting review, to avoid duplicate effort like this.

Sorry about that.

adrianbrink:
I think this can still be a useful simplification for the test cases. Is that mergeable?

state/state_test.go
```diff
@@ -5,74 +5,87 @@ import (
...
+// setupTestCase does setup common to all test cases
+func setupTestCase(t *testing.T) (func(t *testing.T), dbm.DB, *State) {
 	config := cfg.ResetTestRoot("state_")
-
-	// Get State db
 	stateDB := dbm.NewDB("state", config.DBBackend, config.DBDir())
 	state := GetState(stateDB, config.GenesisFile())
 	state.SetLogger(log.TestingLogger())
 
+	tearDown := func(t *testing.T) {}
```
ebuchman: why bother with this?

state/state_test.go
```diff
@@ -5,74 +5,87 @@ import (
...
-		},
-	}}
+	abciResponses.DeliverTx[0] =
+		&abci.ResponseDeliverTx{Data: []byte("foo")}
+	abciResponses.DeliverTx[1] =
+		&abci.ResponseDeliverTx{Data: []byte("bar"), Log: "ok"}
+	abciResponses.EndBlock =
+		abci.ResponseEndBlock{Diffs: []*abci.Validator{{
+			PubKey: crypto.GenPrivKeyEd25519().PubKey().
+				Bytes(),
```
ebuchman: Does this really need to be on a new line ?

state/state_test.go
adrianbrink: In case we are going to add some teardown behaviour in the future. I wanted to make it obvious how to add it.

Issue #629, opened 2017-08-30
rpc: auto generate the slate docs


ebuchman:
Would be amazing to have a script that tested each function and populated the doc.

zramsay:
@melekes can you host them on heroku like mentioned and we can update the link here on RTD

melekes:
https://glacial-waters-46750.herokuapp.com/#block

melekes:
now I am thinking maybe we should move API docs to RTD also. What do you think?

zramsay:
yes. although maybe the `.rst` format is problematic

melekes:
Closing in favor of #691 ?

PR #640, opened 2017-08-31
Implement read the docs

zramsay:
- consolidates the documentation into one place: http://tendermint.readthedocs.io/en/read-the-docs/
- we no longer need to update the website when docs are updated - RTD includes a docs building CI
- converted everything from `docs/*.md`to `docs/*.rst`and organized the `docs` directory in-line with idiomatic RTD 
- ports `/intro` from website into docs


rigelrozanski:
Based on conversation with @zramsay 
Within the new `.rst` files there are links to other `.html` files. Within the readthedocs system this is fine because the new `.html` files are generated from the `.rst` files, So this is all fine and dandy it just means that if you happen to be reading the `.rst` files directly in the github repo, none of the links will work, I think this is OK so long as directly in the docs directly we include a `README.md` which states "these docs are intended to be used as a part of the readthedocs website" and then provides a link or somthing to that matter... so yeah let's include that in this PR plz

rigelrozanski:
Also what's the deal with the failing tests? Is this still non-determinism issues?

zramsay:
readme added. tests failing = non-determinism. (at time of writing latest commit has passed)

docs/rpc.rst
```diff
@@ -0,0 +1,188 @@
+RPC
+===
+
+Coming soon: RPC docs powered by `slate <https://github.com/lord/slate>`__. Until then, read on.
```
zramsay: @melekes what does it take to host the slate docs live?

melekes: I hosted the slate docs on Heroku just to provide an example. But it could be any server.

melekes:
Awesome work!

docs/rpc.rst
zramsay: #629

PR #650, opened 2017-09-11
Consensus params

ebuchman:
Move the MaxBlockSize and the BlockPartSize into the Genesis file.

Removes the BlockPartSize from the config. 

TODO: fix the replay_test data generator


blockchain/reactor.go
```diff
@@ -163,6 +162,12 @@ func (bcR *BlockchainReactor) Receive(chID byte, src *p2p.Peer, msgBytes []byte)
 	}
 }
 
+// maxMsgSize returns the maximum allowable size of a
+// message on the blockchain reactor.
+func (bcR *BlockchainReactor) maxMsgSize() int {
+	return bcR.state.Params().MaxBlockSizeBytes + 2
```
odeke-em: What's the magic number 2 for?

blockchain/reactor.go
```diff
@@ -290,11 +295,11 @@ var _ = wire.RegisterInterface(
 
 // DecodeMessage decodes BlockchainMessage.
 // TODO: ensure that bz is completely read.
-func DecodeMessage(bz []byte) (msgType byte, msg BlockchainMessage, err error) {
+func DecodeMessage(bz []byte, maxSize int) (msgType byte, msg BlockchainMessage, err error) {
```
odeke-em: What's the rationale for us changing this API?

types/genesis.go
```diff
@@ -54,13 +50,48 @@ func (genDoc *GenesisDoc) ValidatorHash() []byte {
...
+}
+
 //------------------------------------------------------------
 // Make genesis state from file
 
 // GenesisDocFromJSON unmarshalls JSON data into a GenesisDoc.
 func GenesisDocFromJSON(jsonBlob []byte) (*GenesisDoc, error) {
 	genDoc := GenesisDoc{}
 	err := json.Unmarshal(jsonBlob, &genDoc)
+	if err != nil {
```
odeke-em: Nice catch!

blockchain/reactor.go
ebuchman: The extra encoding that happens to turn the block into a BlockchainMessage. I think it might actually be `+1` though. We should add a test for this stuff.

odeke-em: Roger that, and also perhaps a variable ;)

blockchain/reactor.go
ebuchman: Because we need some way to get the size in there that doesn't make the size a global.  Another option is to make this a method on something - I might rather that, but at least this is clear. Regardless, we're discussing un-exporting these functions anyways

odeke-em: Oh yes, I recall the unexporting mention, thanks for the reminder. Gotcha.

odeke-em:
@ebuchman might need a rebase here.

types/genesis_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+func TestGenesis(t *testing.T) {
+
+	// test some bad ones from raw json
+	testCases := [][]byte{
+		[]byte{},                                // empty
+		[]byte{1, 1, 1, 1, 1},                   // junk
+		[]byte(`{}`),                            // empty
+		[]byte(`{"chain_id": "mychain"}`),       // missing validators
+		[]byte(`{"validators": [{"data":abcd}`), // missing validators
```
melekes: missing `]` after `abcd}`. or is this what you test here?

types/genesis_test.go
```diff
@@ -0,0 +1,88 @@
...
+	_, err := GenesisDocFromJSON(genDocBytes)
+	assert.Nil(t, err, "expected no error for good genDoc json")
+
+	// create a base gendoc from struct
+	baseGenDoc := &GenesisDoc{
+		ChainID:    "abc",
+		Validators: []GenesisValidator{{crypto.GenPrivKeyEd25519().PubKey(), 10, "myval"}},
+	}
+	genDocBytes, err = json.Marshal(baseGenDoc)
+	assert.Nil(t, err, "error marshalling genDoc")
```
melekes: `assert.NoError`

types/params.go
```diff
@@ -0,0 +1,76 @@
...
+// Validate validates the ConsensusParams to ensure all values
+// are within their allowed limits, and returns an error if they are not.
+func (params *ConsensusParams) Validate() error {
+	if params.BlockSizeParams.MaxBytes <= 0 {
+		return errors.Errorf("BlockSizeParams.MaxBytes must be greater than 0. Got %d", params.BlockSizeParams.MaxBytes)
+	}
+	if params.BlockGossipParams.BlockPartSizeBytes <= 0 {
+		return errors.Errorf("BlockGossipParams.BlockPartSizeBytes must be greater than 0. Got %d", params.BlockGossipParams.BlockPartSizeBytes)
+	}
+	return nil
```
melekes: what about the upper limits?

PR #654, opened 2017-09-13
consolidate docs from tools repo

zramsay:
- moves in the README of each tool from the tools repo into the docs, creating a "Tendermint Tools" section on read-the-docs
- closes https://github.com/tendermint/tools/issues/46
- links to https://github.com/tendermint/tools/pull/51
- from the fifth commit onwards, modifications were made to the content, to go from this:
screenshot from 2017-09-13 10-01-59

to this:
screenshot from 2017-09-13 16-51-21


ebuchman:
I was about to merge this but it dawned on me - can/should we not keep this in the `tools` repo ? Can we have a second readthedocs site for it? or some combine docs from multiple repos into one nice readthedocs site ?

zramsay:
we already have 3 rtd sites so not in favour of another. The readme in root of `tools` repo can link to the rtd tools section.

ebuchman:
I'm concerned about updates. I don't like that updating how a tool works in another repo means we have to merge to master here.  Is it possible to have the tendermint readthedocs read from two repos ?

Greg-Szabo:
If you're concerned about updates, then shouldn't we have a `docs` repository that contains all docs for all projects? In that case, docs would be updated whenever any of the projects are updated - but it would still not affect any project.

zramsay:
i went with something like https://github.com/tendermint/tendermint/pull/654/files#diff-85987f48f1258d9ee486e3191495582dR185 which does the trick. Note that the PR in tools is now also updated

zramsay:
unblocked by tendermint/tools#51 being merged

docs/abci-cli.rst
```diff
@@ -1,5 +1,5 @@
-Using the abci-cli
-==================
+Using ABCI
```
ebuchman: This really should be the ABCI-CLI ... using ABCI is something an app dev does. Here we're talking about using the CLI

zramsay:
fixed

Issue #595, opened 2017-07-28
Error loading ConsensusState wal module=consen sus err="open \.tendermint\data\cs.wal\wal: The system cannot find the path specified."


inovatecc:
**BUG REPORT (?)**:

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):
b467515719e686e4678e6da4e102f32a491b85a0

**ABCI app** (name for built-in, URL for self-written if it's publicly available):
installed using go get -u github.com/tendermint/abci/cmd/...

**Merkleeyes version** (use `git rev-parse --verify HEAD`, skip if you don't use it):
N/A

**Environment**:
- **OS** (e.g. from /etc/os-release): Windows 7 run inside virtualbox vm
- **Install tools**: go, nodejs, git per instruction
- **Others**:

**What happened**:
Just following along getting started guide starting with https://tendermint.com/downloads

**What you expected to happen**:
tried running first sample apps, dummy, counter, counter nodejs app
the sample apps run fine, but when running tendermint with 
tendermint init (later unsafe_reset_all)
tendermint node
get:
E[07-28|03:18:39.616] Error loading ConsensusState wal             module=consen
sus err="open \\.tendermint\\data\\cs.wal\\wal: The system cannot find the path
specified."

if run tendermint node without running the sample app then get:
E[07-28|03:21:13.788] abci.socketClient failed to connect to tcp://127.0.0.1:466
58.  Retrying... module=abci-client connection=query

**How to reproduce it** (as minimally and precisely as possible):
on windows 7 "inside virtualbox vm", just following the getting started guide.  used install tendermint from source option.  everything installed fine.  defined $GOPATH=c:\tendermint\go in user environment variable.  (installed go under tendermint)

am a starter with tendermint, with blockchain as a whole.  don't have lot of experience with go, git command line, nodejs, etc yet.

inovatecc:
Got it working... only a workaround though... 
Following is for Windows but may apply to other OS:

After typing tendermint init and tendermint node OR tendermint unsafe_reset_all and tendermint node... ^C out and then create a file .tendermint\data\cs.wal\wal (.tendermint is located under C: drive).  Then rerun tendermint node.  Not tendermint unsafe_reset_all b/c this will delete the file again.

[...]

PR #664, opened 2017-09-18
use filepath for windows compatibility

zramsay:
- background: #595 (closes issue)
- user reports that this changed fixed their problem on windows. Using `filepath` is (AFAIK) recommended for Linux and other OSs.


odeke-em:
@melekes would you like to do the merge honours or should I?

zramsay:
oops, misread the username. anyways...

Issue #573, opened 2017-07-07
handle corrupt wal file


ebuchman:
node should still be able to start if arbitrary bytes have been truncated from end of file.

 we should add a checksum and line length to each entry to help detect corruption

aphyr:
To elaborate somewhat: we've discussed three theoretical issues in Tendermint's write-ahead-log. 

1. First, WAL entries are not checksummed, which potentially allows bit-flips or truncation to go undetected in WAL records.
2. The second issue is that we're not sure whether WAL writes are fsynced before being acknowledged to the client, which might, in theory, allow committed operations to be rolled back.
3. The third problem is that depending on the filesystem, a write of a log entry may not be appended atomically. If the node crashes prior to an fsync, some sequence of bytes might be left in the WAL which are not recoverable.

Exploratory testing with process crashes has failed to produce WAL safety issues, but I haven't yet tested WAL truncation, node power failure or filesystem-level snapshots. It might be worth testing with those failure modes to see what happens.

One simple approach to writing recoverable WALs is to write a fixed-size checksum, fixed-size length header, then variable-length record. Errors can be detected whenever a record fails its checksum, or not enough bytes are present in the file to complete a read. For more advanced WAL formats, you might consider taking a look at the [SQLite](https://www.sqlite.org/fileformat.html#wal_file_format) WAL format, which incorporates checkpoints.

aphyr:
To follow up: the `truncate-tendermint` nemesis now lets us test the effects of arbitrary truncation of the tendermint WAL file; Tendermint silently ignores corrupt entries at the end of the file.

However, we *are* now relatively certain that Tendermint can lose acknowledged writes due to the lack of any fsync call on the write-ahead log. Saving an entry to the WAL calls writeLine to append data to the autofile group backing the WAL, then calls [group.Flush()](https://github.com/tendermint/tendermint/blob/01e008cedd7f6cdee31ae5f1463a9f6bbb233fd3/consensus/wal.go#L92) to flush that data to persistent storage. group.Flush() in turn [proxies](https://github.com/tendermint/tmlibs/blob/7ce4da1eee6004d627e780c8fe91e96d9b99e459/autofile/group.go#L156) to headBuf.flush(), flushing the active [bufio.BufferedWriter](https://golang.org/pkg/bufio/#Writer.Flush). However, BufferedWriter wraps a Writer, not another BufferedWriter, and the way it flushes is by calling io.Writer.Write() to clear the BufferedWriter's buffer.

[...]

PR #672, opened 2017-09-21
Add checksum and length to CS WAL record

melekes:
See #573


consensus/test_data/small_block2.cswal
```diff
@@ -1,15 +1,11 @@
...
...tes":"363331011A3631363236333634333633323344363436333632363133363332011A3631363236333634333633333344363436333632363133363333011A3631363236333634333633343344363436333632363133363334011A3631363236333634333633353344363436333632363133363335011A3631363236333634333633363344363436333632363133363336011A3631363236333634333633373344363436333632363133363337011A3631363236333634333633383344363436333632363133363338011A3631363236333634333633393344363436333632363133363339011A3631363236333634333733303344363436333632363133373330011A3631363236333634333733313344363436333632363133373331011A3631363236333634333733323344363436333632363133373332011A3631363236333634333733333344363436333632363133373333011A3631363236333634333733343344363436333632363133373334011A3631363236333634333733353344363436333632363133373335011A3631363236333634333733363344363436333632363133373336011A3631363236333634333733373344363436333632363133373337011A3631363236333634333733383344363436333632363133373338011A3631363236333634333733393344363436333632363133373339011A363136","proof":{"aunts":["56097661A1B2707588100586B3B1C2C8A51057D1","6DE889147DF528EEB5F7422E95DC45900CAFB619","247C721D5CEB90BB1FE389BA74C43DF0955E1647"]}}}],"peer_key":""}]}
```
melekes: Still, don't understand why TXs are bigger in old version cause I did not change random.sh params!

melekes: I've tried running the build.sh script both locally on my Mac and inside VirtualBox (Linux)

consensus/replay.go
```diff
@@ -46,6 +47,18 @@ func (cs *ConsensusState) readReplayMessage(msgBytes []byte, newStepCh chan inte
 		fmt.Println("MsgBytes:", msgBytes, string(msgBytes))
 		return fmt.Errorf("Error reading json data: %v", err)
 	}
+	// check checksum
+	innerMsgBytes := wire.JSONBytes(msg.Msg)
+	crc32c := crc32.MakeTable(crc32.Castagnoli)
```
odeke-em: Minor nit: if this function is in a hot path, we can declare as a global `var crc32c = crc32.MakeTable(crc32.Castagnoli)`

consensus/test_data/build.sh
```diff
@@ -16,73 +23,72 @@ DIR_TO_COPY=$HOME/.tendermint_test/consensus_state_test
...
 function small_block2(){
-echo "" >> ~/.tendermint/config.toml
-echo "block_part_size = 512" >> ~/.tendermint/config.toml
-bash scripts/txs/random.sh 1000 36657 &> /dev/null &
-PID=$!
-tendermint node --proxy_app=persistent_dummy &> /dev/null &
-sleep 5
-killall tendermint
-kill -9 $PID
+	{ echo ""; echo "[consensus]"; echo "block_part_size = 64"; } >> ~/.tendermint/config.toml
```
odeke-em:
```shell
echo -e "\n[consensus]\nblock_part_size = 64" >> ~/.tendermint/config.toml
```

consensus/wal.go
```diff
@@ -83,7 +86,11 @@ func (wal *WAL) Save(wmsg WALMessage) {
 		}
 	}
 	// Write the wal message
-	var wmsgBytes = wire.JSONBytes(TimedWALMessage{time.Now(), wmsg})
+	innerMsgBytes := wire.JSONBytes(wmsg)
```
odeke-em: I see similar code duplicated from `consensus/replay.go`, perhaps we can extract it to a common helper
```go
var crc32CastagnoliTable = crc32.MakeTable(crc32.Castagnoli)
func crc32Castagnoli(msg interface{}) (msgBytes []byte, crc uint32) {
   msgBytes = wire.JSONBytes(msg)
   crc = crc32.Checksum(msgBytes, crc32CastagnoliTable)
   return msgBytes, crc
}
```

consensus/test_data/small_block2.cswal
odeke-em: Well technically I think you did, at https://github.com/tendermint/tendermint/pull/672/files/e874c021dc21d536302e23f3201aa71bf1584fff#diff-d7715a8a80c6b17d47e8a27ed32d059bL72 see
screen shot 2017-09-25 at 5 47 08 pm

And also it makes sense that you have an 8X reduction in size
```go
package main

import (
	"log"

	"github.com/tendermint/tendermint/consensus"
)

var data = map[string]string{
	"old": `{"Height":1,"Round":0,"Part":{"index":0,"bytes":"0101010F74656E6465726D696E745F74657374010114B96164A30A118001620000000001141F6753D22BACA2180B1EADD722434EB28444D91D0114354594CBFC1A7BCA1AD0050ED6AA010023EADA3900010162011A3631363236333634333133303344363436333632363133313330011A3631363236333634333133313344363436333632363133313331011A3631363236333634333133323344363436333632363133313332011A3631363236333634333133333344363436333632363133313333011A3631363236333634333133343344363436333632363133313334011A3631363236333634333133353344363436333632363133313335011A3631363236333634333133363344363436333632363133313336011A3631363236333634333133373344363436333632363133313337011A3631363236333634333133383344363436333632363133313338011A3631363236333634333133393344363436333632363133313339011A3631363236333634333233303344363436333632363133323330011A3631363236333634333233313344363436333632363133323331011A3631363236333634333233323344363436333632363133323332011A3631363236333634333233333344363436333632363133323333011A3631363236333634333233343344363436333632363133323334011A36313632363336","proof":{"aunts":["49F4B71E3D7C457415069E2EA916DB12F67AA8D0","D35A72BEDAAAAC17045D7BFAAFA94C2EC0B0A4C2","705BC647374F3495EE73C3F44C21E9BDB4731738"]}}}`,
	"new": `{"Height":1,"Round":0,"Part":{"index":0,"bytes":"0101010F74656E6465726D696E745F74657374010114E67C5AEDA6288000000000000001147297262C6CD96190E46846C9A0DE1227E76077CF00010001000000","proof":{"aunts":["C81B94933420221A7AC004A90242D8B1D3E5070D"]}}}`,
}

func main() {
	for name, datum := range data {
		var err error
		part := wire.ReadJSON(&consensus.BlockPartMessage{}, []byte(datum), &err).(*consensus.BlockPartMessage)
		if err != nil {
			log.Printf("%s: decoding: %v", name, err)
			continue
		}
		log.Printf("%s: len(part.Bytes): %d\n", name, len(part.Part.Bytes))
	}
}
```

giving
```shell
 go run main.go 
2017/09/25 17:53:41 old: len(part.Bytes): 512
2017/09/25 17:53:41 new: len(part.Bytes): 64
```

melekes: This setting only changes part_size for small_block2 case. I was talking more about all cases, not only this one.

consensus/replay.go
melekes: nice idea

consensus/wal.go
melekes: for 2 lines. don't think it's worth doing

consensus/wal.go
```diff
@@ -83,7 +86,10 @@ func (wal *WAL) Save(wmsg WALMessage) {
...
```
ebuchman: I think we probably shouldn't use JSON here. It's kind of convenient for glancing at the wal, but for the sake of speed we should use a binary format. We can have a tool to map the binary lines to json for human readability.

Also, in that setting, I think we want the message size immediately after the checksum, ie. as Kyle put it "One simple approach to writing recoverable WALs is to write a fixed-size checksum, fixed-size length header, then variable-length record."

consensus/test_data/build.sh
```diff
@@ -16,73 +23,72 @@ DIR_TO_COPY=$HOME/.tendermint_test/consensus_state_test
...
+	tendermint node --proxy_app=persistent_dummy &> /dev/null &
+	sleep 5
+	killall tendermint
 
-# /q would print up to and including the match, then quit.
-# /Q doesn't include the match.
-# http://unix.stackexchange.com/questions/11305/grep-show-all-the-file-up-to-the-match
-sed '/ENDHEIGHT: 1/Q' ~/.tendermint/data/cs.wal/wal  > consensus/test_data/empty_block.cswal
+	# /q would print up to and including the match, then quit.
+	# /q doesn't include the match.
```
ebuchman: eh ?

consensus/test_data/build.sh
```diff
@@ -16,73 +23,72 @@ DIR_TO_COPY=$HOME/.tendermint_test/consensus_state_test
...
+	# http://unix.stackexchange.com/questions/11305/grep-show-all-the-file-up-to-the-match
+	sed -e "/ENDHEIGHT: 1/Q" ~/.tendermint/data/cs.wal/wal > consensus/test_data/empty_block.cswal
```
ebuchman: `-e` ?

consensus/test_data/build.sh
```diff
@@ -16,73 +23,72 @@ DIR_TO_COPY=$HOME/.tendermint_test/consensus_state_test
...
 function small_block2(){
-echo "" >> ~/.tendermint/config.toml
-echo "block_part_size = 512" >> ~/.tendermint/config.toml
-bash scripts/txs/random.sh 1000 36657 &> /dev/null &
-PID=$!
-tendermint node --proxy_app=persistent_dummy &> /dev/null &
-sleep 5
-killall tendermint
-kill -9 $PID
+	echo -e "\n[consensus]\nblock_part_size = 64" >> ~/.tendermint/config.toml
```
ebuchman: block_part_size is now in the genesis file, not the config, so we have to edit that!

consensus/test_data/build.sh
melekes: it does not work on MacOS without `-e`

consensus/wal.go
melekes: By here you mean the whole WAL? or just this intermediate encoding?

ebuchman: yeh both. certainly dont really see any utility to the outer encoding being json if the inner encoding is binary, so might as well do it all binary. Do you have a different opinion?

melekes: No, it is just that we will lose readability. I guess if we will provide a wal2json-viewer script, which outputs WAL as json, it should be fine.

melekes:
I am gonna work on this today, thanks

consensus/wal.go
melekes: doh, all `tmlibs/autofile/group` code is about strings (separated with "\n") including searching for "#ENDHEIGHT". I am rewriting it right now

melekes: "You can open a Group to keep restrictions on an AutoFile, like
the maximum size of each chunk, and/or the total amount of bytes
stored in the group."

interesting. so now it is possible to have one part of an msg in file X and the rest of it in file Y.

ebuchman: That doesn't sound good. Are you saying AutoFile won't work with binary format?

melekes: No, I will work, but autofile/group won't. There is no `Write([]byte)` method to start with. And search functionality will be broken once we start using binary format.

melekes:
TODO:

- [ ] write docs for cutWALUntil and wal2json

consensus/replay_test.go
```diff
@@ -176,11 +173,13 @@ func setupReplayTest(t *testing.T, thisCase *testCase, nLines int, crashAfter bo
 		lineStep -= 1
 	}
 
-	split := strings.Split(thisCase.log, "\n")
+	split := bytes.Split(thisCase.log, walSeparator)
 	lastMsg := split[nLines]
 
 	// we write those lines up to (not including) one with the signature
-	walFile := writeWAL(strings.Join(split[:nLines], "\n") + "\n")
+	bytes := bytes.Join(split[:nLines], walSeparator)
```
ebuchman: lets try not to reuse import names

consensus/test_data/build.sh
```diff
@@ -1,14 +1,26 @@
 #!/usr/bin/env bash
 
 # XXX: removes tendermint dir
+# TODO: does not work on OSX
```
ebuchman: i think I figured it out - you need to install `gnu-sed`. eg. `brew install gnu-sed --with-default-names` ... that should make your `sed` point to `usr/local/bin/sed` which is actually `gnu-sed`. 

We should document this better, but eg. https://www.topbug.net/blog/2013/04/14/install-and-use-gnu-command-line-tools-in-mac-os-x/

consensus/test_data/build.sh
```diff
@@ -20,71 +32,71 @@ TMHOME="$HOME/.tendermint"
...
+	./scripts/cutWALUntil/cutWALUntil ~/.tendermint/data/cs.wal/wal 1 consensus/test_data/new_small_block1.cswal
+	mv consensus/test_data/new_small_block1.cswal consensus/test_data/small_block1.cswal
 
-reset
+	reset
 }
 
 
-# small block 2 (part size = 512)
+# small block 2 (part size = 64)
```
ebuchman: 512

consensus/wal.go
```diff
@@ -12,18 +17,29 @@ import (
 //--------------------------------------------------------
 // types and functions for savings consensus messages
 
+var (
+	walSeparator = []byte{55, 127, 6, 130} // 0x377f0682 - magic number
```
ebuchman: what happens when this sequence shows up in a block ?

consensus/wal.go
```diff
@@ -70,35 +89,195 @@ func (wal *WAL) OnStop() {
...
+
+	msg := make([]byte, totalLength)
+	binary.BigEndian.PutUint32(msg[0:4], crc)
+	binary.BigEndian.PutUint32(msg[4:8], length)
+	copy(msg[8:], data)
+
+	_, err := enc.wr.Write(msg)
+
+	if err == nil {
+		// TODO [Anton Kaliaev 23 Oct 2017]: remove separator
```
ebuchman: does this mean we're hoping to do all this without a separator ?

consensus/test_data/build.sh
melekes: Cool! Thanks

melekes: Actually, the script does not contain sed anymore, so we can remove this line safely. But good to know anyway!

melekes:
> Looking good. A little unsure about the separator but not immediately sure how else to do things.

We only need the separator for tests, where we use it to split WAL into pieces (from these pieces we are building temporary WAL). But nobody prevents us from just reading N pieces from WAL directly without splitting it.

> Also, is scripts/wal2json/wal2json accidentally checked in ?

yes, sorry.

consensus/wal.go
melekes: see https://github.com/tendermint/tendermint/pull/672#issuecomment-338910947

consensus/wal.go
melekes: If we're using our old group#Search method or `gr.ReadLine`, we would definitely get bad results. Imagine for some reason we failed to write a separator "\n" or it's lost due to FS errors, then if we use `gr.ReadLine` we'll get 2 messages, but only parse one.

```
{"time":"2017-10-24T08:13:04.564Z","msg":{"height":6,"round":0,"step":"RoundStepNewHeight"}}{"time":"2017-10-24T08:13:04.564Z","msg":{"height":7,"round":0,"step":"RoundStepNewHeight"}}\n
```

new WAL#SearchForHeight method sequentially reads the WAL file and panics if no separator is found.

> what happens when this sequence shows up in a block ?

answering your original question, nothing bad will happen because as I explained above we're no longer splitting the log (using `strings.Split` or `bytes.Split`; `bytes.Split` is only used in tests), but rather read it byte by byte (4bytes CRC -> 4 bytes length -> (length) data -> separator -> 4bytes CRC -> 4 bytes length ...)

ebuchman: Yep thanks

consensus/wal.go
```diff
@@ -70,35 +89,195 @@ func (wal *WAL) OnStop() {
...
+	}
+	length := binary.BigEndian.Uint32(b)
+
+	data := make([]byte, length)
+	n, err = dec.rd.Read(data)
+	if err == io.EOF {
+		return nil, err
+	}
+	if err != nil {
+		return nil, fmt.Errorf("not enough bytes for data: %v (want: %d, read: %v)", err, length, n)
```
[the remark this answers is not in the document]
melekes: I had an impression that `Read` always returns err if n < length, but that seems not to be the case `When Read encounters an error or end-of-file condition after successfully reading n > 0 bytes, it returns the number of bytes read. It may return the (non-nil) error from the same call or return the error (and n == 0) from a subsequent call.`. We probably need to write more tests and revisit implementation of groupReader#Read.

consensus/replay.go
```diff
@@ -113,12 +113,11 @@ func (cs *ConsensusState) catchupReplay(csHeight int) error {
 		cs.Logger.Error("Replay: wal.group.Search returned EOF", "#ENDHEIGHT", csHeight-1)
 	} else if err != nil {
 		return err
-	} else {
-		defer gr.Close()
 	}
 	if !found {
 		return errors.New(cmn.Fmt("Cannot replay height %d. WAL does not contain #ENDHEIGHT for %d.", csHeight, csHeight-1))
 	}
+	defer gr.Close()
```
melekes: I think it should be above `if !found`.

ebuchman: It seems in all cases where `!found`, we also have `gr == nil`. Otherwise, we panic when we try to close `gr.Close()`.

consensus/wal.go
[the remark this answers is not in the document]
ebuchman: https://github.com/tendermint/tmlibs/issues/69

PR #677, opened 2017-09-25
blockchain: add tests for BlockStore

odeke-em:
Add tests to test store, to the fullest reasonable extent for
paths that can taken by input arguments altering internal behavior.


blockchain/store_test.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	for i, panicCauser := range panicCausers {
+		// Expecting a panic here on trying to parse an invalid blockStore
+		_, _, err := doFn(func() (interface{}, error) {
+			db.Set(blockStoreKey, panicCauser)
+			_ = blockchain.NewBlockStore(db)
+			return nil, nil
+		})
+		assert.NotNil(t, err, "#%d panicCauser: %q expected a panic", i, panicCauser)
```
melekes: there is an https://godoc.org/github.com/stretchr/testify/assert#Panics

blockchain/store_test.go
```diff
@@ -0,0 +1,263 @@
...
+		// SaveBlock
+		res, err, panicErr := doFn(func() (interface{}, error) {
+			bs.SaveBlock(tuple.block, tuple.parts, tuple.seenCommit)
+			if tuple.block == nil {
+				return nil, nil
+			}
+			bBlock := bs.LoadBlock(tuple.block.Height)
+			bBlockMeta := bs.LoadBlockMeta(tuple.block.Height)
+			bSeenCommit := bs.LoadSeenCommit(tuple.block.Height)
+			if tuple.eraseDBForCommit {
```
melekes: replayed in the associated PR. I think we should not erase db by calling `Close()`. feels wrong

adrianbrink: Can you reattach to the in-mem db after closing it?

odeke-em: @melekes thank you for the advocacy, I've reflected on your comments in the PR https://github.com/tendermint/tmlibs/pull/56 and I concede with your thoughts :)

@adrianbrink yes we can :) However, I'll instead invoke an iterator and delete all the content in the DB for this test instead of invoking .Close()

blockchain/store_test.go
odeke-em: Nice, thank you for the reference, I hadn't seen that function. I'll skip using it though for this test particularly because for some of these we want panics, for others we don't and constructing the code with assert.Panics might be a little hairy, I haven't seen an easy way of doing it :)

blockchain/store_test.go
odeke-em: I've updated the code, no longer need to invoke db.Close(), I now have a helper that literally erases the DB, PTAL @melekes and @adrianbrink, thanks.

blockchain/store_test.go
```diff
@@ -0,0 +1,268 @@
...
+	// End of setup, test data
+
+	tuples := []struct {
+		block            *types.Block
+		parts            *types.PartSet
+		seenCommit       *types.Commit
+		wantErr          bool
+		wantPanic        bool
+		eraseDBForCommit bool
+		skipReason       string
```
melekes: can't see where this is used

odeke-em: Ahh crap, I forgot to remove it after the last commit, I'll update that, thanks.

blockchain/store_test.go
```diff
@@ -0,0 +1,267 @@
...
+	noBlockHeights := []int{0, -1, 100, 1000, 2}
+	for i, height := range noBlockHeights {
+		if g := bs.LoadBlock(height); g != nil {
+			t.Errorf("#%d: height(%d) got a block; want nil", i, height)
+		}
+	}
+
+	// Setup, test data
+	part2 := &types.Part{Index: 1, Bytes: data.Bytes([]byte{
+		0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x01, 0x01, 0x01, 0x00,
```
melekes: how did you generate these? or it's just random bytes?

odeke-em: So I generated them by
```go
package main

import (
	"bytes"
	"log"

)

func main() {
	buf := new(bytes.Buffer)
	var n int
	var err error

	block := &types.Block{
		Header: &types.Header{
			Height:  1,
			NumTxs:  100,
			ChainID: "block_test",
		},
		LastCommit: &types.Commit{Precommits: []*types.Vote{{Height: 10}}},
	}
	wire.WriteBinary(block, buf, &n, &err)
	log.Printf("% x", buf.Bytes())
}
```
and then split the bytes over two parts.

However, I didn't include the generation code in the test, lest it caused confusion.

melekes: But if something changes, how a next person supposed to change this stuff? Maybe we should at least add a comment?

adrianbrink: Let's include the generation code as well and ideally generate them on each test run.

odeke-em: Roger that, I'll add that comment in in a second. @adrianbrink am not too sure about regenerating it on each test run as that makes the test harder to grok, however a note on how to generate the parts bytes.

odeke-em:
Alright, I've updated the test to also include how to generate the parts' data, PTAL.

blockchain/store_test.go
```diff
@@ -0,0 +1,269 @@
...
+		assert.Nil(t, err, "#%d: expecting a non-nil error", i)
+		qua, ok := res.(*quad)
+		if !ok || qua == nil {
+			t.Errorf("#%d: got nil quad back; gotType=%T", i, res)
+			continue
+		}
+	}
+}
+
+func doFn(fn func() (interface{}, error)) (res interface{}, err error, panicErr error) {
```
ebuchman: is this worth it? why not just let it panic in the test?

blockchain/store_test.go
```diff
@@ -0,0 +1,269 @@
...
+		}
+	}()
+
+	res, err = fn()
+	return res, err, panicErr
+}
+
+func eraseDB(db db.DB) {
+	it := db.Iterator()
+	for {
```
ebuchman: can't we just do:

```
for it.Next(){
 db.Delete(it.Key())
}
```

blockchain/store_test.go
ebuchman: arent all three of these test cases expecting a panic ?

blockchain/store_test.go
```diff
@@ -0,0 +1,269 @@
...
+	for i, height := range noBlockHeights {
+		if g := bs.LoadBlock(height); g != nil {
+			t.Errorf("#%d: height(%d) got a block; want nil", i, height)
+		}
+	}
+
+	// Setup, test data
+	// The parts' data can be generated by running
+	// the code at https://gist.github.com/odeke-em/9ffac2b5df44595fad7084ece4c9bd98
+	part1 := &types.Part{Index: 0, Bytes: data.Bytes([]byte{
```
ebuchman: i feel like we can simplify this test a bunch. do we really need these part sets? we can derrive part sets directly from blocks, and we have functions to make blocks so we shouldnt need much external madness. 

If we want incompelte/bad part sets, we can mutate them after derriving them from blocks.

ebuchman:
Can we also take this opportunity to fix up the comments in store.go.

Also might be nice to have some tests about the Height() - ie. we should always be able to fetch block bs.Height(), but bs.Height()+1 should be nil

blockchain/store_test.go
odeke-em: @ebuchman in this isolated test with these 3 test cases, yes all those 3 expect panics, but doFn is also used throughout the file in other cases that don't necessarily want to panic.

blockchain/store_test.go
```diff
@@ -0,0 +1,269 @@
...
+
+	bsj := &blockchain.BlockStoreStateJSON{Height: 1000}
+	bsj.Save(db)
+
+	retrBSJ := blockchain.LoadBlockStoreStateJSON(db)
+
+	assert.Equal(t, *bsj, retrBSJ, "expected the retrieved DBs to match")
+}
+
+var blockStoreKey = []byte("blockStore")
```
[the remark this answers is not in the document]
odeke-em: @ebuchman I was just mocking malicious behavior, but sure, I can move the file to be a part of package `blockchain` instead of `blockchain_test`

blockchain/store_test.go
odeke-em: I believe we need this function. The point of it is to capture the panic and relay it back to whatever caller needs to check if we expected a panic or not -- for example if the blockMeta content is messed up vs when a block is successfully read.
If we just let the test panic, we wouldn't be asserting expected behavior and the tests wouldn't matter. I hope this makes sense.

blockchain/store_test.go
odeke-em: For some reason that was initially buggy on my side and not erasing the last element sometime back, but that now looks functional, thank you for the update.

odeke-em:
@ebuchman by fix, what do you mean? Do you mean add comments? Or review the available comments?

blockchain/store_test.go
odeke-em: @ebuchman there are 3 different partsets that we are testing:
1) validAndCompletePartSet
2) incompletePartSet
3) UncontiguousPartSet

Mutating after deriving IMO makes it even more complex. Could you please help with an example of what you mean?

odeke-em: For reference https://github.com/tendermint/tendermint/pull/677/files/23cb7d7c12ab48d30f23b2590c6bc3d3e72c9e29#diff-2dcc7092e28fd6d98ff9d9ef3b6bdcd1R118
screen shot 2017-10-03 at 9 24 23 pm

ebuchman:
>  Do you mean add comments?

Yeh, just something simple for the exported values

odeke-em:
Thanks for the suggestions errbody. I've updated the code and added even more tests for BlockStore plus the requested docs for the exposed functions, PTAL.

blockchain/store.go
```diff
@@ -7,10 +7,10 @@ import (
 	"io"
 	"sync"
 
-	. "github.com/tendermint/tmlibs/common"
-	dbm "github.com/tendermint/tmlibs/db"
 	"github.com/tendermint/go-wire"
 	"github.com/tendermint/tendermint/types"
+	. "github.com/tendermint/tmlibs/common"
```
ebuchman: we must beat back these anonymous imports one PR at a time!

ebuchman:
Thanks @odeke-em . I'll make any final edits and merge

blockchain/store.go
zramsay: +++

ebuchman:
@odeke-em thanks for the extensive comments. That said, I feel they are a bit too verbose, with undue emphasis on implementation details. I have updated them in https://github.com/tendermint/tendermint/pull/677/commits/4b393bd0512bbf9819ea1edaf595c36a737a777c - PTAL for style

blockchain/store_test.go
```diff
@@ -0,0 +1,459 @@
...
+
+func TestNewBlockStore(t *testing.T) {
+	db := db.NewMemDB()
+	db.Set(blockStoreKey, []byte(`{"height": 10000}`))
+	bs := NewBlockStore(db)
+	assert.Equal(t, bs.Height(), 10000, "failed to properly parse blockstore")
+
+	panicCausers := []struct {
+		data    []byte
+		wantErr string
```
ebuchman: this feels way to fragile

ebuchman: can you point to an example of how other large Go projects test their panics ?

blockchain/store_test.go
```diff
@@ -0,0 +1,459 @@
...
+	db.Set([]byte("Foo"), []byte("Bar"))
+	db.Set([]byte("Foo1"), nil)
+
+	bs := NewBlockStore(db)
+
+	tests := [...]struct {
+		key  []byte
+		want []byte
+	}{
+		0: {key: []byte("Foo"), want: []byte("Bar")},
```
ebuchman: This doesn't feel like a very useful test. Maybe we can change it to test the Reader in the context of actually reading blocks, rather than simple strings? Ie. the sequence of: 

- marshal block
- set block in db
- GetReader
- ReadBinary
- compare result to original block

blockchain/store_test.go
```diff
@@ -0,0 +1,459 @@
...
+func freshBlockStore() (*BlockStore, db.DB) {
+	db := db.NewMemDB()
+	return NewBlockStore(db), db
+}
+
+var (
+	// Setup, test data
+	// If needed, the parts' data can be generated by running
+	// the code at https://gist.github.com/odeke-em/9ffac2b5df44595fad7084ece4c9bd98
+	part1 = &types.Part{Index: 0, Bytes: data.Bytes([]byte{
```
ebuchman: Feels too fragile. We should use MakeBlock and break it to parts instead of constructing from scratch like this.

blockchain/store_test.go
```diff
@@ -0,0 +1,459 @@
...
+
+		corruptBlockInDB      bool
+		corruptCommitInDB     bool
+		corruptSeenCommitInDB bool
+		eraseCommitInDB       bool
+		eraseSeenCommitInDB   bool
+	}{
+		{
+			block: &types.Block{
+				Header: &types.Header{
```
ebuchman: these headers are all the same, save maybe an odd change. lets make a var for each distinct one and use those.

```
header1 := types.Header{
		Height:  1,
		NumTxs:  100,
		ChainID: "block_test",
}
header2 := header1
header2.Height = 4
```
...

blockchain/store_test.go
```diff
@@ -0,0 +1,459 @@
...
+		0x73, 0x74, 0x01, 0x01, 0xa1, 0xb2, 0x03, 0xeb, 0x3d, 0x1f, 0x44, 0x40, 0x01, 0x64, 0x00,
+	})}
+	part2 = &types.Part{Index: 1, Bytes: data.Bytes([]byte{
+		0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x01, 0x01, 0x01, 0x00,
+		0x00, 0x01, 0x0a, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+	})}
+	seenCommit1 = &types.Commit{Precommits: []*types.Vote{{Height: 10}}}
+)
+
+func TestBlockStoreSaveLoadBlock(t *testing.T) {
```
ebuchman: Overall this test is super large and unweildly. Think we could break it up into Test..SaveLoadBlock, Test...SaveLoadBlockIncompletePartSet, Test...SaveLoadBlokcUncontiguousPartSet

ebuchman:
>  I'll make any final edits and merge

I meant this with respect to the doc comments. Leaving test cleanup to you please :)

blockchain/store_test.go
ebuchman: cant we use block.MakePartSet ?

blockchain/store_test.go
odeke-em: Great question. So I picked up this pattern from the Go standard library and here is an exhibit of the usage of `strings.Contains` for substring checks https://gist.github.com/odeke-em/32a71113ee50da57d8a1b861d926d328
Or you can run it in your terminal like this
```shell
$ cd $GOPATH/src && grep -Rn 'strings.Contains' * | grep '.*_test.go' | pbcopy 
```

blockchain/store_test.go
odeke-em: @ebuchman good point, however, those sequences/suggestions are covered in the end-to-end tests in other cases. The purpose of this test is to ensure:
a) If we have nil as the value, that we get back a nil Reader
b) A reader if non-nil will not error out if we try to slurp out all the data.

However, if we still can't come to consensus, we can remove this test altogether
since the behavior that it tests is arguably already covered by other tests.

blockchain/store.go
zramsay: #711

blockchain/store_test.go
odeke-em: I definitely hadn't thought of that, that's way simpler and cleaner. Thank you.

ebuchman:
Thanks @odeke-em. Your comments all sound good. Indeed, I think mentioning the panic situation once up front is sufficient. Please feel free to make the changes you've suggested and the test changes I suggested when you have time

odeke-em:
Alright squad, thanks for the feedback. I've made the tests less fragile, PTAL.

blockchain/store_test.go
odeke-em: @ebuchman PTAL at my comments above.

odeke-em:
Alright, I've rebased this code to fix a merge conflict that had crept up.

blockchain/store_test.go
```diff
@@ -0,0 +1,424 @@
...
+		wantPanic  string
+
+		corruptBlockInDB      bool
+		corruptCommitInDB     bool
+		corruptSeenCommitInDB bool
+		eraseCommitInDB       bool
+		eraseSeenCommitInDB   bool
+	}{
+		{
+			block: &types.Block{
```
ebuchman: there's lots of repetition in laying out these structs that could be compressed

odeke-em: Interesting thought that I hadn't considered. Thanks for raising it, I've updated that.

blockchain/store.go
```diff
@@ -23,7 +23,7 @@ There are three types of information stored:
 
 Currently the precommit signatures are duplicated in the Block parts as
 well as the Commit.  In the future this may change, perhaps by moving
-the Commit data outside the Block.
+the Commit data outside the Block. (TODO)
```
zramsay: link to an issue?

ebuchman: not priority enough

PR #689, opened 2017-09-28
rpc: subscribe on reconnection

alexppxela:
When connection is lost with server, client does not re-subscribe so client is not notified anymore.

### DIFF
--- a/rpc/lib/client/ws_client.go
+++ b/rpc/lib/client/ws_client.go
@@ -41,8 +41,9 @@ type WSClient struct {
 	PingPongLatencyTimer metrics.Timer
 
 	// user facing channels, closed only when the client is being stopped.
-	ResultsCh chan json.RawMessage
-	ErrorsCh  chan error
+	ResultsCh   chan json.RawMessage
+	ErrorsCh    chan error
+	ReconnectCh chan bool
 
 	// internal channels
 	send            chan types.RPCRequest // user requests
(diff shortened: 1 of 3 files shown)


melekes:
wow! thanks a lot

rpc/lib/client/ws_client.go
```diff
@@ -254,6 +256,7 @@ func (c *WSClient) reconnect() error {
 			c.Logger.Error("failed to redial", "err", err)
 		} else {
 			c.Logger.Info("reconnected")
+			c.ReconnectCh <- true
```
melekes: I'd prefer an optional callback here:

```
go c.OnReconnect()
```

odeke-em: @melekes does this apply though since I believe that ReconnectCh was exported so that we could use it in `rpc/client/httpclient.go` at https://github.com/tendermint/tendermint/pull/689/files#diff-a969276db2d519208a2bf7aa9447acf6R338
screen shot 2017-09-29 at 1 21 50 am
as this file is in `rpc/lib/client` so we need that visibility to be able to use it there.

alexppxela: You're true, for me it is the best way to delegate reconnection behavior to WSEvents.
I've just fixed the unit tests.

melekes: @odeke-em yes, it is. you probably did not notice, but `OnReconnect` is also exported. so in httpclient we should be able to do:

```
w.ws.OnReconnect = func() {
  w.redoSubscriptions()
}
```

odeke-em: Ooh I see, roger that @melekes.

melekes: I rewrote some code here https://github.com/tendermint/tendermint/commit/da5bd254b6d124a6b4fbb8721f4e001f7e370682. Feel free to leave your comments.

PR #692, opened 2017-10-02
add the unstable changes

zramsay:


DOCKER/Dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM alpine:3.6
 
 # This is the release of tendermint to pull in.
-ENV TM_VERSION 0.10.0
-ENV TM_SHA256SUM a29852b8d51c00db93c87c3d148fa419a047abd38f32b2507a905805131acc19
+ENV TM_VERSION 0.11.0
```
ebuchman: lol. one day, when we receive our Greg II from the heavens, we will automate this :P

README.md
```diff
@@ -56,6 +56,7 @@ All resources involving the use of, building application on, or developing for,
 
 * [Ethermint](http://github.com/tendermint/ethermint); Ethereum on Tendermint
 * [Cosmos SDK](http://github.com/cosmos/cosmos-sdk); a cryptocurrency application framework
+* [Passchain](http://github.com/trusch/passchain); a secret sharing system for teams
```
ebuchman: I feel like we should have a "community built" section here and list a bunch of them ...

rpc/grpc/types.proto
```diff
@@ -1,7 +1,7 @@
 syntax = "proto3";
 package core_grpc;
 
-import "github.com/tendermint/abci/types/types.proto";
+import "github.com/tendermint/abci/blob/master/types/types.proto";
```
ebuchman: really? was this broken ?

scripts/release.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
```
ebuchman: LOL I spoke to soon. you folks are heroes!

ebuchman:
I think this is GTM except we should fix the README to have the community section rather than just dropping Passchain there like that :)

ebuchman:
ok I changed the README to point to the ecosystem doc, and removed Passchain (its still in the ecosystem doc)  - @trusch I hope that's ok with you - I'd rather keep all community apps in a common place other than the README

ebuchman:
@zramsay how can I review this branch on readthedocs? Need to make sure my formatting is ok. 

Also we should think about how we can improve the ecosystem page - the layout of apps is kind of bland. Servers table looks good. Why not do something similar for the apps? Maybe @nylira has an idea?

rpc/grpc/types.proto
melekes: I did not change this

docs/ecosystem.rst
```diff
@@ -36,37 +36,44 @@ Ethermint
 
 The go-ethereum state machine run as a ABCI app, written in Go, `authored by Tendermint <https://github.com/tendermint/ethermint>`__.
 
+Lotion
+^^^^^^^^^^^^^
 
-Merkle AVL Tree
-^^^^^^^^^^^^^^^
+A Javascript microframework for building blockchain applications with Tendermint, written in Javascript, `authored by Judd Keppel of Tendermint <https://github.com/keppel/lotion>`__. See also `lotion-chat <https://github.com/keppel/lotion-chat>` and `lotion-coin <https://github.com/keppel/lotion-coin>` apps written using Lotion.
```
zramsay: links need a double underscore after the last backtick

zramsay:
either run `make html` from the docs dir and view locally, or we enable the specific branch to build on RTD. Because it displays publicly, it's tedious to turn on and off, although we could have a relatively stable `docs-staging` branch that lingers and we use as a public staging area

zramsay:
agree WRT to the apps display; it's been on my mind for a bit. What if we give each one a page and list each one in a standard format: 
- Install
- Setup/Initialize
- Run
- Clean up
- Other Info

and invite the authors to fill it in?

ebuchman:
I think the maintenance cost of that will be too high. They don't each need a page.

zramsay:
k, we'll brainstorm; this PR is good to go tho

rpc/grpc/types.proto
ebuchman: @zramsay can you explain this?

zramsay: https://github.com/tendermint/tendermint/pull/686

PR #696, opened 2017-10-03
WIP: 548 postgresql indexer

melekes:
See #548

### DIFF
--- a/state/txindex/postgresql/postgresql.go
+++ b/state/txindex/postgresql/postgresql.go
@@ -1,18 +1,22 @@
 package postgresql
 
 import (
-	"database/sql/driver"
+	"database/sql"
 
+	"github.com/lib/pq"
+	"github.com/pkg/errors"
+
+	abci "github.com/tendermint/abci/types"
 	"github.com/tendermint/tendermint/state/txindex"
 	"github.com/tendermint/tendermint/types"
 )
 
 type TxIndex struct {
-	db driver.Conn
+	db *sql.DB
 }
 
 // NewTxIndex returns new instance of TxIndex.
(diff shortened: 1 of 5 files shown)


state/txindex/postgresql/postgresql.go
```diff
@@ -0,0 +1,93 @@
...
+		return errors.Wrap(err, "failed to begin PG transaction")
+	}
+
+	stmt, err := txn.Prepare(pq.CopyIn("txs", "hash", "height", "index", "tx", "result_data", "result_code", "result_log"))
+	if err != nil {
+		return errors.Wrap(err, "failed to prepare PG statement")
+	}
+
+	for _, result := range b.Ops {
+		_, err = stmt.Exec(fmt.Sprintf("%X", result.Tx.Hash()), int64(result.Height), int64(result.Index), result.Tx, result.Result.Data, int64(result.Result.Code), result.Result.Log)
```
odeke-em: Looks unsafe to me, perhaps we should do `"? ? ? ? ?", fmt.Sprintf("%x"....`

ebuchman:
@melekes What should we do with this ? Hard to see the core changes amidst the pubsub stuff

ebuchman:
Oh maybe we just need to remove that first commit ?

melekes:
> What should we do with this ?

this thing probably needs more work

melekes:
- [ ] index Tags
- [ ] config for Postgresql connection?

ebuchman:
Closing in favour of focused issue: https://github.com/tendermint/tendermint/issues/1161. Let's keep the branch around so we can re-open when we need to. 

Don't think we need it for launch, but probably shortly after.

PR #698, opened 2017-10-03
WIP: Run tests on AppVeyor

adrianbrink:


appveyor.yml
```diff
@@ -0,0 +1,14 @@
...
+- x64
+- x86
+clone_folder: c:\go\path\src\github.com\tendermint\tendermint
+before_build:
+- cmd: set GOPATH=%GOROOT%\path
+- cmd: set PATH=%GOPATH%\bin;%PATH%
+- cmd: go get github.com/Masterminds/glide
+- cmd: glide install
+build_script:
+- cmd: go test ./...
```
zramsay: `make test_integrations` ?

ebuchman: we shouldnt have CI block on running the entire suite of tests twice.

ebuchman:
We should run the full integration tests, but I don't want PRs to block on waiting for that, unless it will run in parallel with CircleCI - and even still.

I think rather we should have this run in the background every time we merge to develop and alert us if it fails. 

We should then make it part of release policy to ensure the windows build passed before merging.

Sound reasonable ?

adrianbrink:
This would be run in parallel with CircleCi.

Let's start with running and reporting the normal tests and if it works and we like it we can add the integration tests too.

@ebuchman

appveyor.yml
```diff
@@ -0,0 +1,14 @@
+version: 1.0.{build}
```
melekes: what's that? and why it is 1.0.X?

adrianbrink:
* ensure that no vendor tests are run

ebuchman:
@adrianbrink are we waiting on something here ?

melekes:
https://github.com/tendermint/tendermint/issues/878#issuecomment-347175863

appveyor.yml
mdyring: https://www.appveyor.com/docs/build-configuration/#build-versioning

I believe it can be changed to whatever is desired, even to a tag name when building a release (using APPVEYOR_REPO_TAG_NAME).

My understanding is that release binaries are built outside of CI at present, so the version number above could just as well be ignored.

Issue #695, opened 2017-10-03
Document that app should return AppHash in response to Commit


melekes:
From Rocket chat:

```
And when I send a transaction, I'm getting the error:

2017-10-02T12:51:08.578728711Z panic: Panicked questionably: Failed to process committed block (118:37353841374132414642413934354234433837334530434432313837424532313742444432314644): Exec failed for application: Wrong Block.Header.AppHash. Expected 7B9835027FC7FDCB48E2F827410BFD10192DE28A, got 
2017-10-02T12:51:08.57877568Z 
2017-10-02T12:51:08.578787576Z goroutine 26 [running]:
2017-10-02T12:51:08.578797317Z panic(0xa5d160, 0xc4204b7200)
2017-10-02T12:51:08.578806051Z /usr/local/go/src/runtime/panic.go:500 +0x1a1
2017-10-02T12:51:08.578815911Z github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common.PanicQ(0xa5d160, 0xc4204b71e0)
2017-10-02T12:51:08.578825117Z /go/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common/errors.go:44 +0xe0
2017-10-02T12:51:08.578849808Z github.com/tendermint/tendermint/blockchain.(*BlockchainReactor).poolRoutine(0xc420094000)
2017-10-02T12:51:08.578860637Z /go/src/github.com/tendermint/tendermint/blockchain/reactor.go:249 +0x1119
2017-10-02T12:51:08.578870609Z created by github.com/tendermint/tendermint/blockchain.(*BlockchainReactor).OnStart
2017-10-02T12:51:08.578879108Z /go/src/github.com/tendermint/tendermint/blockchain/reactor.go:93 +0x6e

melekes Cosmos 5:05 PM
well, the height is increasing over time. it means TM is making blocks
as for tx processing
looks like the app did not send an apphash (it's empty)

gian 2:31 AM
But how can I generate an apphash ? Every example I've seen, the apphash is empty in genesis file

adribeiro 4:19 AM
hi @gian, you can update when send the answer in commit with the OK status code.
 gian 2:31 AM
But how can I generate an apphash ? Every example I've seen, the apphash is empty in genesis file
so you can you into info answers as well

julian.smith 8:47 AM   
@melekes - what’s best documentation around (or specific lines of sample code) having an ABCI app “sending the apphash” to TM so we can rectify ? 😀

[...]

PR #700, opened 2017-10-03
Improve app dev docs

melekes:
Refs #695 

Questions:

1) Can we embed code directly from Github? something like `https://github.com/tendermint/abci/blob/master/example/dummy/persistent_dummy.go#L10-20`
2) I think we need examples in another language(s). Java?
3) persistent-dummy may not be the best example. Other suggestions?

### DIFF
--- a/docs/app-development.rst
+++ b/docs/app-development.rst
@@ -142,9 +142,12 @@ It is unlikely that you will need to implement a client. For details of
 our client, see
 `here <https://github.com/tendermint/abci/tree/master/client>`__.
 
-All examples below are from `persistent-dummy application
-<https://github.com/tendermint/abci/blob/master/example/dummy/persistent_dummy.go>`__,
-which is a part of the abci repo.
+Most of the examples below are from `dummy application
+<https://github.com/tendermint/abci/blob/master/example/dummy/dummy.go>`__,
+which is a part of the abci repo. `persistent_dummy application
+<https://github.com/tendermint/abci/blob/master/example/dummy/persistent_dummy.go>`__
+application is used to show ``BeginBlock``, ``EndBlock`` and ``InitChain``
+example implementations.
 
 Blockchain Protocol
 -------------------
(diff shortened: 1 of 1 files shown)


zramsay:
1. maybe with some javascript but I wouldn't recommend because that file can easily change & keeping track (and updating) of all embedded code blocks would be a PITA. A repo could have self-generating docs (embedding the code at that time) that is then imported by RTD (see last lines of `conf.py`).
2. Ya agree. #701
3. Hmm, we do need to improve upon the dummy/counter/persistent-dummy examples we use. Its mostly a naming issue, no? Maybe we can have a random name for each binary instead?

docs/app-development.rst
```diff
@@ -215,6 +225,22 @@ The block header will be updated (TODO) to include some commitment to
 the results of DeliverTx, be it a bitarray of non-OK transactions, or a
 merkle root of the data returned by the DeliverTx requests, or both.
 
+::
+
+    // tx is either "key=value" or just arbitrary bytes
```
ebuchman: we need to fix this comment because tx can also be `val:pubkey/power`

melekes: fixed in abci:develop

docs/app-development.rst
```diff
@@ -142,6 +142,13 @@ It is unlikely that you will need to implement a client. For details of
 our client, see
 `here <https://github.com/tendermint/abci/tree/master/client>`__.
 
+Most of the examples below are from `dummy application
+<https://github.com/tendermint/abci/blob/master/example/dummy/dummy.go>`__,
+which is a part of the abci repo. `persistent_dummy application
+<https://github.com/tendermint/abci/blob/master/example/dummy/persistent_dummy.go>`__
+application is used to show ``BeginBlock``, ``EndBlock`` and ``InitChain``
```
ebuchman: application twice

docs/app-development.rst
```diff
@@ -228,7 +254,7 @@ Commit, or there will be deadlock. Note also that all remaining
 transactions in the mempool are replayed on the mempool connection
 (CheckTx) following a commit.
 
-The Commit response includes a byte array, which is the deterministic
+The app should respond with a byte array, which is the deterministic
```
ebuchman: "should respond to the Commit request with"

PR #702, opened 2017-10-03
initial commit of gofuzz files

caffix:
moving the gofuzz files into tendermint core

### DIFF
--- a/p2p/gofuzz.go
+++ b/p2p/gofuzz.go
@@ -93,17 +93,16 @@ func init() {
 
 // Each execution of the fuzzer calls this function with random data provided.
 func Fuzz(data []byte) int {
-	var i int
 	datalen := len(data)
 
 	if datalen < 12 {
-		// small amounts of data will not due
+		// small amounts of data will not do
 		return -1
 	}
 
 	// simulate random remote peer
 	peer := randomPeer(data)
-	i += 6
+	i := 6
 
 	// send a PEXRequestMessage to the reactor
 	request := wire.BinaryBytes(struct{ PexMessage }{&pexRequestMessage{}})
--- a/p2p/gofuzz.go
+++ b/p2p/gofuzz.go
@@ -1,0 +1,139 @@
+// Copyright 2017 Tendermint. All rights reserved.
+// Use of this source code is governed by Apache 2 LICENSE that can be found in the LICENSE file.
+
+package p2p
+
+import (
+	"encoding/binary"
+	"net"
+
+	wire "github.com/tendermint/go-wire"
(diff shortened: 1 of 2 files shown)


p2p/gofuzz.go
```diff
@@ -0,0 +1,139 @@
...
+	remoteConn, _ = net.Pipe()
+}
+
+// Each execution of the fuzzer calls this function with random data provided.
+func Fuzz(data []byte) int {
+	var i int
+	datalen := len(data)
+
+	if datalen < 12 {
+		// small amounts of data will not due
```
odeke-em: Did you mean to say `will not do`?

p2p/gofuzz.go
```diff
@@ -0,0 +1,139 @@
...
+		return -1
+	}
+
+	// simulate random remote peer
+	peer := randomPeer(data)
+	i += 6
```
odeke-em: This is the first instantiation of i, so perhaps could we make it:
```go
i := 6
```

p2p/gofuzz.go
caffix: Yes. That is an error.

p2p/gofuzz.go
caffix: Sounds good.

p2p/gofuzz.go
```diff
@@ -0,0 +1,138 @@
...
+// randomPeer uses random bytes provided by the fuzzer to create a random peer.
+// TODO: the peer can be made more complete in order to support testing of other message types.
+func randomPeer(b []byte) *Peer {
+	privKey := crypto.GenPrivKeyEd25519()
+	config := DefaultPeerConfig()
+	config.AuthEnc = false
+	netAddr := randomAddr(b)
+
+	p, err := newInboundPeer(remoteConn, make(map[byte]Reactor),
+		make([]*ChannelDescriptor, 0), func(p *Peer, r interface{}) {}, privKey, config)
```
ebuchman: can a peer with no channels do anything ?

ebuchman:
This is a good start to fuzzing the pexAddrsMessage, but I think we'll need to be more comprehensive.

Could we have the passed in `data` be well formed serialized messages themselves, and we build up a corpus of them? This Fuzz function feels a little too specific, and limitted in how much of the reactor.Receive it covers.

We should also look to fuzz the underlying connections recv routine too for more complete coverage of the peer package, independent of the pex_reactor

ebuchman:
Ah, there's also lots of compile errors because of the recent refactor in the peer package to make `Peer` an interface

ebuchman:
see https://github.com/tendermint/tendermint/pull/719

PR #643, opened 2017-09-05
add metalinter to Makefile & apply (mostly) 'errcheck' lints throughout

zramsay:
- helps #616, which will be tackled one linting tool at a time
- note: metalinter includes the megacheck lints
- see commit history for which linters were applied. not all errors were address (e.g., #645)
- additional PRs will continue to chip away at the linting errors

PR #703, opened 2017-10-04
add metalinter to CI and include fixes

zramsay:
- replaces #643


benchmarks/map_test.go
```diff
@@ -1,4 +1,4 @@
-package benchmarks
+package benchmarks // nolint: goimports
```
ebuchman: As far as I know, there's no case ever wherein we'd want to use nolint for goimports. Please correct me if I'm wrong

ebuchman:
Also, see https://github.com/tendermint/tendermint/issues/715

We should add that to the coding repo and reflect it here

benchmarks/map_test.go
zramsay: thought i'd removed them all

ebuchman:
Hey @zramsay can we get one more rebase here plx

circle.yml
```diff
@@ -24,6 +24,7 @@ dependencies:
 
 test:
   override:
+    - cd "$PROJECT_PATH" && make get_vendor_deps && make metalinter_test
```
ebuchman: can we do the metalinter test without vendor deps? we're going to download the vendor deps later ... no need to do it twice

ebuchman:
Think this is good to merge. Just waiting for CI :)

test/run_test.sh
```diff
@@ -6,6 +6,10 @@ pwd
 BRANCH=$(git rev-parse --abbrev-ref HEAD)
 echo "Current branch: $BRANCH"
 
+# run the linter
+# TODO: drop the `_test` once we're ballin' enough
```
zramsay: :1st_place_medal:

Issue #556, opened 2017-06-24
separate config files and data by creating 2 subfolders (config, data)


antoineherzog:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): FEATURE REQUEST

FEATURE REQUEST
We want to run tendermint on Amazon EC2 and S3.
Could you create a subfolder .tendermint/config where we have all configs files.

We would like to specify the location for each subfolder.

So we can do something like that
.tendermint/data ==> located on the EC2 instance
.tendermint/config ==> located on a S3 Bucket

BENEFITS

We could shut-down our EC2 instance but still keep alive all the configuration files. This way we could turn-on again later the validator without loosing the pub_key, genesis file etc ..

antoineherzog:
@adrianbrink tagged here

melekes:
We discussed it with Bucky and Greg before. Prehistory: it was done mainly for convenience (easy to operate on one dir than two). But we agree that state and configs must be separated.

antoineherzog:
@melekes any ETA on this? or at least could we pass the priv_key file path in the tendermint cli?

ethanfrey:
You can kind of do something already.  Take a good look at our config parser (file structure): https://github.com/tendermint/tendermint/blob/master/config/config.go

`TMHOME=/path/mounted/to/s3 tendermint node`

config.toml:

```
db_dir=/some/local/path

[mempool]
wal_dir=/some/other/local/path
```

If these directories don't start with `/`, they are relative to `TMHOME` (`--home`).  If they are absolute paths, they will ignore TMHOME.

I don't know if there is other data you don't want to write?  And also if it is possible to mount an s3 directory to the filesystem (since we just use `fopen` to open the config and data files).

Please let me know if this helps and what is missing.

zramsay:
closing as stale/resolved

melekes:
I still think there are things to do here. We could create 2 dirs: `~/.tendermint/config` and `~/.tendermint/data`, where the former consists of `config.toml` and `genesis.json` and the latter - WAL, DB files and `priv_validator.json`.

antoineherzog:
i agree. two folders make sense.

zramsay:
cool. I'll pick this up

antoineherzog:
cool!

zramsay:
see #709. Feel free to try that branch

antoineherzog:
sure

zramsay:
consolidated in #628

[...]

Issue #705, opened 2017-10-04
config docs


ebuchman:
We should improve the documentation in the config/config.go and try to make it self documenting so we dont have to copy those details into  docs/specification/configuration.rst

zramsay:
consolidated in #628

PR #709, opened 2017-10-04
config: split data and config directories

zramsay:
#556

### DIFF
--- a/cmd/tendermint/commands/testnet.go
+++ b/cmd/tendermint/commands/testnet.go
@@ -2,7 +2,7 @@ package commands
 
 import (
 	"fmt"
-	"path"
+	"path/filepath"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -45,7 +45,7 @@ func testnetFiles(cmd *cobra.Command, args []string) {
 			cmn.Exit(err.Error())
 		}
 		// Read priv_validator.json to populate vals
-		privValFile := path.Join(dataDir, mach, cfg.DefaultPrivValPath)
+		privValFile := filepath.Join(dataDir, mach, cfg.DefaultPrivValPath)
 		privVal := types.LoadPrivValidatorFS(privValFile)
 		genVals[i] = types.GenesisValidator{
 			PubKey: privVal.GetPubKey(),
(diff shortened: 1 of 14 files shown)


cmd/tendermint/commands/testnet.go
```diff
@@ -44,7 +45,7 @@ func testnetFiles(cmd *cobra.Command, args []string) {
 			cmn.Exit(err.Error())
 		}
 		// Read priv_validator.json to populate vals
-		privValFile := path.Join(dataDir, mach, "priv_validator.json")
+		privValFile := path.Join(dataDir, mach, cfg.DefaultPrivValPath)
```
melekes: `filepath`

cmd/tendermint/commands/testnet.go
```diff
@@ -63,7 +64,7 @@ func testnetFiles(cmd *cobra.Command, args []string) {
 	// Write genesis file.
 	for i := 0; i < nValidators; i++ {
 		mach := cmn.Fmt("mach%d", i)
-		genDoc.SaveAs(path.Join(dataDir, mach, "genesis.json"))
+		genDoc.SaveAs(path.Join(dataDir, mach, cfg.DefaultGenesisJSONPath))
```
melekes: same here

config/config.go
```diff
@@ -6,6 +6,20 @@ import (
 	"time"
 )
 
+var (
+	DefaultTendermintDir = ".tendermint"
```
melekes: where is it used?

config/config.go
```diff
@@ -6,6 +6,20 @@ import (
...
+	DefaultConfigDir     = "config"
+	DefaultDataDir       = "data"
+
+	DefaultConfigFileName  = "config.toml"
```
melekes: I like these, but we already have the config https://github.com/tendermint/tendermint/blob/develop/config/config.go#L69, so we probably use those

config/config.go
zramsay: added to `cmd/tendermint/main.go`

config/config.go
zramsay: i'll tackle that in #705 when diving deeper into the config

ebuchman: Mmm I think Anton is right. Best if we use the values in the config object where we need them rather than these package level globals - it should be part of this PR

ebuchman: Well - I like having all the strings at the top in one place, but I don't think they should be exported.

The only one that should be exported is `DefaultTendermintDir`

config/toml.go
```diff
@@ -13,9 +12,10 @@ import (
 
 func EnsureRoot(rootDir string) {
 	cmn.EnsureDir(rootDir, 0700)
```
ebuchman: do we even need this one if we Ensure subdirs?

zramsay:
addressed all the comments, PTAL

zramsay:
rebased on `config`, merging to that branch

PR #714, opened 2017-10-04
Feature/no block response

ebuchman:
replaces https://github.com/tendermint/tendermint/pull/648

### DIFF
--- a/blockchain/reactor_test.go
+++ b/blockchain/reactor_test.go
@@ -1,0 +1,218 @@
+package blockchain
+
+import (
+	"bytes"
+	"testing"
+
+	cfg "github.com/tendermint/tendermint/config"
+	"github.com/tendermint/tendermint/p2p"
+	sm "github.com/tendermint/tendermint/state"
+	"github.com/tendermint/tendermint/types"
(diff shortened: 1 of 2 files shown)


blockchain/reactor_test.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	for _, tt := range tests {
+		reqBlockMsg := &bcBlockRequestMessage{tt.height}
+		reqBlockBytes := wire.BinaryBytes(struct{ BlockchainMessage }{reqBlockMsg})
+		bcr.Receive(chID, peer, reqBlockBytes)
+		value := peer.lastValue()
+
+		if tt.existent {
+			if _, ok := value.(struct{ BlockchainMessage }).BlockchainMessage.(*bcBlockResponseMessage); !ok {
```
odeke-em: I don't think this test confirms that the block received is actually for that exact same height in tt.height, the code in BlockchainReactor can literally send any height that it wants and this test will still pass. That was the purpose of the height tallying.

ebuchman: Whoops! Thanks, forgot that check - fork pushed

Issue #710, opened 2017-10-04
clean out rpc/lib


zramsay:
has a bunch of left-overs from what was presumably a repo merge (e.g., a circle.yml, a readme, a Makefile). if these aren't used, we should remove

zramsay:
#711

PR #718, opened 2017-10-05
restore rpc/lib readme as doc.go (Refs #710) [ci skip]

melekes:
I don't want to lose any documentation. Correct me if I am wrong, but we
don't have this docs anywhere else.

### DIFF
--- a/rpc/lib/doc.go
+++ b/rpc/lib/doc.go
@@ -1,0 +1,102 @@
+/*
+HTTP RPC server supporting calls via uri params, jsonrpc, and jsonrpc over websockets
+
+# Client Requests
+
+Suppose we want to expose the rpc function `HelloWorld(name string, num int)`.
+
+## GET (URI)
+
+As a GET request, it would have URI encoded parameters, and look like:
+
+```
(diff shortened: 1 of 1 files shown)


zramsay:
```
Step 13 : RUN go install ./cmd/tendermint
 ---> Running in 20b40e702886
node/node.go:26:2: 
rpc/lib/doc.go:102:4: expected 'package', found 'EOF'
```

rpc/lib/doc.go
```diff
@@ -0,0 +1,102 @@
...
+
+Now see all available endpoints by sending a GET request to `0.0.0.0:8008`.
+Each route is available as a GET request, as a JSONRPCv2 POST request, and via JSONRPCv2 over websockets.
+
+
+# Examples
+
+* [Tendermint](https://github.com/tendermint/tendermint/blob/master/rpc/core/routes.go)
+* [tm-monitor](https://github.com/tendermint/tools/blob/master/tm-monitor/rpc.go)
+*/
```
odeke-em: As @zramsay reported from the CI failure, we need the last line after this to be the package e.g
```go
/*
...
*/
package rpc
```

PR #719, opened 2017-10-05
all: gofuzz

ebuchman:
replaces https://github.com/tendermint/tendermint/pull/702


ebuchman:
This is a good start to fuzzing the pexAddrsMessage, but I think we'll need to be more comprehensive.

Could we have the passed in data be well formed serialized messages themselves, and we build up a corpus of them? This Fuzz function feels a little too specific, and limitted in how much of the reactor.Receive it covers.

We should also look to fuzz the underlying connections recv routine too for more complete coverage of the peer package, independent of the pex_reactor

odeke-em:
Thank you @caffix and @ebuchman for working on this and getting it started.

@ebuchman in deed, we shall be more comprehensive with follow-up PRs.

In order for us to effectively fuzz tendermint and p2p, we'll probably want to fuzz it piecemeal i.e disjoint function by function but go-fuzz can only use one Fuzz function https://github.com/dvyukov/go-fuzz#usage so perhaps we might need to break out a directory with different sub-directories and different corpa.

What do y'all think about this?

caffix:
@odeke-em I agree with your proposal to break out the different fuzz tests into separate directories

odeke-em:
I've been fuzzing a bit more with:
- https://github.com/tendermint/tendermint/commit/de2429d2e31b22ad8352fa485f784430fff4dc3d
- https://github.com/tendermint/tendermint/commit/2ba16561d00b6675ee996bd92cad734fae29a318
- https://github.com/tendermint/tendermint/commit/8137fa579d212a49e8c03ee4a417ae60df160d01
- https://github.com/tendermint/tendermint/commit/46c7f2f19ef4ede161fa7b5ecdc103b8bfbb99db
and others that I'll push up soon.

ebuchman:
migrated to https://github.com/tendermint/fuzz

Issue #693, opened 2017-10-02
Ensure all funcs in types pkg are well guarded


ebuchman:
Eg. https://github.com/tendermint/tendermint/blob/master/types/vote_set.go#L137 is highly unsafe

ebuchman:
@odeke-em I was somewhat unnerved to discover this. Want to help level up this package? We need to make sure all functions are well guarded against bad input, or if not, have clear comments that it's the callers responsibility to pass in valid arguments, and if not a panic could result.

I'm mostly worried about aberrant votes/msgs coming in on the consensus reactor and hitting this data structures. For instance here if we received a vote from a different chain that had more validators than us we'd panic - obviously very bad behaviour. This particular one is fixed here: https://github.com/tendermint/tendermint/commit/cd8dff0c96ca3ea1ec44d637f93c10c418edac1a#diff-bd488e139d955daba658159f507010f7R105. But I'm now worried there's more lurking - we could probably do with much improved test suite to try and hunt for all these cases and fuzz these datastructures! 

Interested?

odeke-em:
Great stuff discovering and reporting this @ebuchman. Yes in deed, I am interested in working on this.

ebuchman:
As an indicator, we should probably be shooting for 80% coverage on all of these: https://codecov.io/gh/tendermint/tendermint/tree/master/types

xla:
@melekes After your recent improvements should we still keep this open?

melekes:
Yes. I think there may be still some bugs lurking.

jackzampolin:
Going to go ahead and close this issue.

PR #720, opened 2017-10-05
types/heartbeat: test all Heartbeat functions

odeke-em:
    
* Adjusted Heartbeat.Copy to return nil on
trying to copy a nil value instead of panicking.
* Also documented that WriteSignBytes panics
if the Heartbeat is nil.

### DIFF
--- a/types/heartbeat_test.go
+++ b/types/heartbeat_test.go
@@ -1,0 +1,56 @@
+package types
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/tendermint/go-crypto"
+)
+
+func TestHeartbeatCopy(t *testing.T) {
+	hb := &Heartbeat{ValidatorIndex: 1, Height: 10, Round: 1}
+	hbCopy := hb.Copy()
+	require.Equal(t, hbCopy, hb, "heartbeat copy should be the same")
+	hbCopy.Round = hb.Round + 10
+	require.NotEqual(t, hbCopy, hb, "heartbeat copy mutation should not change original")
+
+	var nilHb *Heartbeat
+	nilHbCopy := nilHb.Copy()
+	require.Nil(t, nilHbCopy, "copy of nil should also return nil")
+}
+
+func TestHeartbeatString(t *testing.T) {
+	var nilHb *Heartbeat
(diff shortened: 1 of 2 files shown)


types/heartbeat_test.go
```diff
@@ -0,0 +1,63 @@
...
+	buf := new(bytes.Buffer)
+
+	hb := &Heartbeat{ValidatorIndex: 1, Height: 10, Round: 1}
+	hb.WriteSignBytes("0xdeadbeef", buf, &n, &err)
+	require.Equal(t, string(buf.Bytes()), `{"chain_id":"0xdeadbeef","heartbeat":{"height":10,"round":1,"sequence":0,"validator_address":"","validator_index":1}}`)
+
+	buf.Reset()
+	plainHb := &Heartbeat{}
+	plainHb.WriteSignBytes("0xdeadbeef", buf, &n, &err)
+	// TODO: File an issue to add ",omitempty" to
```
ebuchman: We actually don't want omit empty because we want SignBytes to have a complete and canonical form

types/heartbeat_test.go
```diff
@@ -0,0 +1,63 @@
...
+	// some of these fields to prune the output
+	require.Equal(t, string(buf.Bytes()), `{"chain_id":"0xdeadbeef","heartbeat":{"height":0,"round":0,"sequence":0,"validator_address":"","validator_index":0}}`)
+
+	// TODO: Currently WriteSignBytes panics when
+	// passed in a nil value. Decide if we should
```
ebuchman: I think panic is fine for this - we should add comments on the WriteSignBytes method to that effect. These types should always be instantiated before WriteSignBytes is called on them - there is no case where we should be able to call it on a nil value, so it's correct to panic

types/heartbeat_test.go
odeke-em: Gotcha gotcha @ebuchman.

types/heartbeat_test.go
odeke-em: Aye aye, I'll document this. Thanks for the advocation.

odeke-em:
@ebuchman thank you for the first round of review and suggestions.
I've updated the code with a force-push, PTAL.

odeke-em:
Thanks @melekes too!

odeke-em:
@ebuchman feeling in the mood for a merge here?

Issue #708, opened 2017-10-04
When leaving out params-field in Websocket-RPC tm crashes


wolfposd:
How to repoduce:

send `{"jsonrpc":"2.0","method":"dump_consensus_state","id":"0"}`

```
17:17:01 sent:  {"jsonrpc":"2.0","method":"dump_consensus_state","id":"0"}
17:17:01 system: Connection closed.Close status: 1006 (Abnormal Closure)Session length: 168714ms
```

tendermint crashes with following log:

```
github.com/tendermint/tendermint/rpc/lib/server.jsonParamsToArgsRPC(0xc4204cc4d0, 0x0, 0xc4207ee480, 0x14, 0xc420077220, 0x0, 0x0)
	/Volumes/HDD/wolf/golang/src/github.com/tendermint/tendermint/rpc/lib/server/handlers.go:210 +0x26
github.com/tendermint/tendermint/rpc/lib/server.(*wsConnection).readRoutine(0xc4202381b0)
	/Volumes/HDD/wolf/golang/src/github.com/tendermint/tendermint/rpc/lib/server/handlers.go:506 +0x7eb
created by github.com/tendermint/tendermint/rpc/lib/server.(*wsConnection).OnStart
	/Volumes/HDD/wolf/golang/src/github.com/tendermint/tendermint/rpc/lib/server/handlers.go:385 +0xab
```

still on tendermint 0.10.3 btw

melekes:
```
{"jsonrpc":"2.0","method":"dump_consensus_state","id":"0","params":{}}
```
this works

odeke-em:
@melekes your sample gives it the zero value whereas it omitting it entirely sets it as nil as at
https://play.golang.org/p/0qLcHQiX2w

Perhaps we should guard against this as on the failing line we do a dereference yet params is nil i.e.
https://github.com/tendermint/tendermint/blob/7682ad9a60162dd17fd6f61aeed7049a8635ac78/rpc/lib/server/handlers.go#L211-L213

and also guard against it in
https://github.com/tendermint/tendermint/blob/7682ad9a60162dd17fd6f61aeed7049a8635ac78/rpc/lib/server/handlers.go#L216-L222

odeke-em:
This is actually on my list of things to fuzz soon :)

melekes:
> Perhaps we should guard against this as on the failing line we do a dereference yet params is nil i.e.

[...]

PR #721, opened 2017-10-06
Add app_options to GenesisDoc

melekes:
Fixes #564

PR #724, opened 2017-10-09
Leaving out params crashes tm rpc

melekes:
See #708

### DIFF
--- a/rpc/core/consensus.go
+++ b/rpc/core/consensus.go
@@ -1,13 +1,10 @@
 package core
 
 import (
-	"encoding/json"
-
 	cm "github.com/tendermint/tendermint/consensus"
+	cstypes "github.com/tendermint/tendermint/consensus/types"
 	ctypes "github.com/tendermint/tendermint/rpc/core/types"
 	"github.com/tendermint/tendermint/types"
-
-	"github.com/pkg/errors"
 )
 
 // Get the validator set at the given block height.
@@ -85,22 +82,12 @@ func Validators(heightPtr *int) (*ctypes.ResultValidators, error) {
 // }
 // ```
 func DumpConsensusState() (*ctypes.ResultDumpConsensusState, error) {
-	roundStateBytes, err := json.Marshal(consensusState.GetRoundState())
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to marshal round state")
-	}
-
-	peerRoundStates := make(map[string]json.RawMessage)
-	for i, peer := range p2pSwitch.Peers().List() {
+	peerRoundStates := make(map[string]*cstypes.PeerRoundState)
+	for _, peer := range p2pSwitch.Peers().List() {
 		// TODO: clean this up?
 		peerState := peer.Get(types.PeerStateKey).(*cm.PeerState)
 		peerRoundState := peerState.GetRoundState()
-		peerRoundStateBytes, err := json.Marshal(peerRoundState)
-		if err != nil {
-			return nil, errors.Wrapf(err, "failed to marshal peer#%d round state", i)
-		}
-		peerRoundStates[peer.Key()] = json.RawMessage(peerRoundStateBytes)
+		peerRoundStates[peer.Key()] = peerRoundState
 	}
-
-	return &ctypes.ResultDumpConsensusState{json.RawMessage(roundStateBytes), peerRoundStates}, nil
+	return &ctypes.ResultDumpConsensusState{consensusState.GetRoundState(), peerRoundStates}, nil
 }
--- a/rpc/core/consensus.go
+++ b/rpc/core/consensus.go
@@ -1,10 +1,13 @@
 package core
 
 import (
-	wire "github.com/tendermint/go-wire"
+	"encoding/json"
+
 	cm "github.com/tendermint/tendermint/consensus"
 	ctypes "github.com/tendermint/tendermint/rpc/core/types"
 	"github.com/tendermint/tendermint/types"
+
+	"github.com/pkg/errors"
 )
 
 // Get the validator set at the given block height.
@@ -82,14 +85,22 @@ func Validators(heightPtr *int) (*ctypes.ResultValidators, error) {
 // }
 // ```
 func DumpConsensusState() (*ctypes.ResultDumpConsensusState, error) {
-	roundState := consensusState.GetRoundState()
-	peerRoundStates := []string{}
-	for _, peer := range p2pSwitch.Peers().List() {
+	roundStateBytes, err := json.Marshal(consensusState.GetRoundState())
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to marshal round state")
+	}
+
+	peerRoundStates := make(map[string]json.RawMessage)
+	for i, peer := range p2pSwitch.Peers().List() {
 		// TODO: clean this up?
 		peerState := peer.Get(types.PeerStateKey).(*cm.PeerState)
 		peerRoundState := peerState.GetRoundState()
-		peerRoundStateStr := peer.Key() + ":" + string(wire.JSONBytes(peerRoundState))
-		peerRoundStates = append(peerRoundStates, peerRoundStateStr)
+		peerRoundStateBytes, err := json.Marshal(peerRoundState)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to marshal peer#%d round state", i)
+		}
+		peerRoundStates[peer.Key()] = json.RawMessage(peerRoundStateBytes)
 	}
-	return &ctypes.ResultDumpConsensusState{roundState.String(), peerRoundStates}, nil
+
+	return &ctypes.ResultDumpConsensusState{json.RawMessage(roundStateBytes), peerRoundStates}, nil
 }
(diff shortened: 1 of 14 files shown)


melekes:
> Could we also ensure that that's the only path to those functions?

what do you mean? try with and without "params: {}"? I already did that

odeke-em:
By path I meant, like code usage so perhaps a grep for usage of those functions might suffice?

rpc/core/types/responses.go
```diff
@@ -75,8 +76,8 @@ type ResultValidators struct {
 }
 
 type ResultDumpConsensusState struct {
-	RoundState      string   `json:"round_state"`
-	PeerRoundStates []string `json:"peer_round_states"`
+	RoundState      json.RawMessage   `json:"round_state"`
```
ebuchman: Oh - does this make it easier to parse the result?!

ebuchman:
CI failure:

```
Step 13 : RUN go install ./cmd/tendermint
 ---> Running in c30821f47119
node/node.go:26:2: 
rpc/lib/doc.go:102:4: expected 'package', found 'EOF'
```

rpc/core/types/responses.go
melekes: yeah.. before we're spitting a string (2 strings), now it's a JSON https://github.com/tendermint/tendermint/issues/708#issuecomment-335115500

melekes:
I fixed the tests in https://github.com/tendermint/tendermint/pull/721 https://github.com/tendermint/tendermint/pull/721/commits/6fca03c36af1f0ab20434f02a8a84b9879d98fe1

rpc/core/types/responses.go
ebuchman: So, it seems like we could use the actual structs here, but we don't want to have to import consensus to get them. So probably we should move some types from `consensus` into `consensus/types` so we can import here. 

Will these raw messages be identical to:

```
type ResultDumpConsensusState struct {
  RoundState cstypes.RoundState
  PeerRoundStates map[string]cstypes.PeerRoundState
}
```

ebuchman:
This also causes a panic in the non-ws RPC, but that server's handler is protected by a `recover` statement, so you only see the panic in the logs.

Is there a reason the RawMessage is a pointer in the first place?

ebuchman:
We should include this in 0.11.1. The format of dump_consensus was so poor previously it shouldn't have been used - a great bug fix

odeke-em:
Interestingly the fuzzer finds the other crash in 9s
given code
```go
package handler

import (
	"bytes"
	"encoding/json"
	"io/ioutil"
	"net/http"
	"net/http/httptest"

	rs "github.com/tendermint/tendermint/rpc/lib/server"
	types "github.com/tendermint/tendermint/rpc/lib/types"
)

var rpcFuncMap = map[string]*rs.RPCFunc{
	"c": rs.NewRPCFunc(func(s string, i int) (string, int) { return "foo", 200 }, "s,i"),
}

func Fuzz(data []byte) int {
	mux := http.NewServeMux()
	buf := new(bytes.Buffer)
	lgr := log.NewTMLogger(buf)
	rs.RegisterRPCFuncs(mux, rpcFuncMap, lgr)

	req, _ := http.NewRequest("POST", "http://localhost/", bytes.NewReader(data))
	rec := httptest.NewRecorder()
	mux.ServeHTTP(rec, req)
	res := rec.Result()
	blob, err := ioutil.ReadAll(res.Body)
	if err != nil {
		panic(err)
	}
	recv := new(types.RPCResponse)
	if err := json.Unmarshal(blob, recv); err != nil {
		panic(err)
	}
	return 1
}
```

We could also adapt the code in the fuzzer for a test.

odeke-em:
where fuzzer crash output is
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x133a066]

goroutine 1 [running]:
github.com/tendermint/tendermint/rpc/lib/server.jsonParamsToArgsRPC(0xc4201b32d0, 0x0, 0xc4201f2d58, 0x1, 0xc42007b198, 0x0, 0x0)
	/var/folders/v3/7z434qpx5v3bw0wh8h2myfpw0000gn/T/go-fuzz-build199365166/gopath/src/github.com/tendermint/tendermint/rpc/lib/server/handlers.go:213 +0x26
github.com/tendermint/tendermint/rpc/lib/server.makeJSONRPCHandler.func1(0x14253a0, 0xc4201e15c0, 0xc420205b00)
	/var/folders/v3/7z434qpx5v3bw0wh8h2myfpw0000gn/T/go-fuzz-build199365166/gopath/src/github.com/tendermint/tendermint/rpc/lib/server/handlers.go:130 +0x3b7
net/http.HandlerFunc.ServeHTTP(0xc42023a480, 0x14253a0, 0xc4201e15c0, 0xc420205b00)
	/var/folders/v3/7z434qpx5v3bw0wh8h2myfpw0000gn/T/go-fuzz-build199365166/goroot/src/net/http/server.go:1921 +0x44
net/http.(*ServeMux).ServeHTTP(0xc4202299e0, 0x14253a0, 0xc4201e15c0, 0xc420205b00)
	/var/folders/v3/7z434qpx5v3bw0wh8h2myfpw0000gn/T/go-fuzz-build199365166/goroot/src/net/http/server.go:2299 +0x130
github.com/tendermint/tendermint/fuzz/rpc/lib/server/handler.Fuzz(0x2230000, 0x17, 0x200000, 0x3)
	/var/folders/v3/7z434qpx5v3bw0wh8h2myfpw0000gn/T/go-fuzz-build199365166/gopath/src/github.com/tendermint/tendermint/fuzz/rpc/lib/server/handler/main.go:27 +0x1f4
go-fuzz-dep.Main(0x13fc5e8)
	/var/folders/v3/7z434qpx5v3bw0wh8h2myfpw0000gn/T/go-fuzz-build199365166/goroot/src/go-fuzz-dep/main.go:49 +0xad
main.main()
	/var/folders/v3/7z434qpx5v3bw0wh8h2myfpw0000gn/T/go-fuzz-build199365166/gopath/src/github.com/tendermint/tendermint/fuzz/rpc/lib/server/handler/go.fuzz.main/main.go:10 +0x2d

goroutine 5 [syscall]:
os/signal.signal_recv(0x0)
	/var/folders/v3/7z434qpx5v3bw0wh8h2myfpw0000gn/T/go-fuzz-build199365166/goroot/src/runtime/sigqueue.go:131 +0xa7
os/signal.loop()
	/var/folders/v3/7z434qpx5v3bw0wh8h2myfpw0000gn/T/go-fuzz-build199365166/goroot/src/os/signal/signal_unix.go:22 +0x52
created by os/signal.init.0
	/var/folders/v3/7z434qpx5v3bw0wh8h2myfpw0000gn/T/go-fuzz-build199365166/goroot/src/os/signal/signal_unix.go:28 +0x59
exit status 2
```
and the other unsafe spot is 
https://github.com/tendermint/tendermint/blob/bcca27ee2032299a7c571ae62a87b596eb96a899/rpc/lib/server/handlers.go#L130

and the input is
```JSON
{"method":"c","id":"0"}
```

odeke-em:
I actually think that just like @ebuchman might have alluded to in https://github.com/tendermint/tendermint/pull/724#issuecomment-335316484, if we don't make Params a pointer, and remove the extra indirection, we won't even need the current PR.
json.RawMessage is an alias for `[]byte` and we can always just check if it wasn't set with a
```go
if len(Params) == 0 {
   return
}
```

rpc/core/types/responses.go
melekes: Good ideas! Will do now

melekes: > clients are importing too much crap

that was the reason behind `json.RawMsg`

melekes: moving types will be a breaking change, though

melekes: see https://github.com/tendermint/tendermint/pull/724/commits/dc0e8de9b0671fc93d7479b20c2c9b5ea54c48c6. I did not move all of the types. just ones which are necessary

melekes:
@odeke-em @ebuchman addressed all your comments. PTAL

melekes:
@odeke-em thanks for finding another bug! fixed it too.

rpc/core/consensus.go
```diff
@@ -82,14 +82,12 @@ func Validators(heightPtr *int) (*ctypes.ResultValidators, error) {
 // }
 // ```
 func DumpConsensusState() (*ctypes.ResultDumpConsensusState, error) {
-	roundState := consensusState.GetRoundState()
-	peerRoundStates := []string{}
+	peerRoundStates := make(map[string]*cstypes.PeerRoundState)
 	for _, peer := range p2pSwitch.Peers().List() {
 		// TODO: clean this up?
```
ebuchman: think we can remove this comment now!

rpc/lib/server/handlers.go
```diff
@@ -127,10 +129,13 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, logger log.Logger) http.Han
 			WriteRPCResponseHTTP(w, types.RPCMethodNotFoundError(request.ID))
 			return
 		}
-		args, err := jsonParamsToArgsRPC(rpcFunc, request.Params)
-		if err != nil {
-			WriteRPCResponseHTTP(w, types.RPCInvalidParamsError(request.ID, errors.Wrap(err, "Error converting json params to arguments")))
-			return
+		var args []reflect.Value
+		if len(request.Params) > 0 {
```
ebuchman: as far as I can tell, now that params is not a pointer, we shouldn't need this guard, but it's fine to leave it

melekes: no, we need it because overwise this results in the following error: "params should be a map or array"

PR #725, opened 2017-10-09
types: test Vote* and implement Copy for many related structs

odeke-em:

* Tests for VoteSet
* More guards for GetByIndex
* Implemented a mostly deep-Copy for:
- Block
- BlockID
- Commit
- Data
- Header
- PartSetHeader
- VoteSet
Whose purposes are to ensure that
after a vote has been added/cast to a VoteSet
it cannot be mutated by the outside actor. Not too sure
though if the extra copy might instead become a DOS Vector
but I believe most importantly no outside actor should
be able to modify a vote after we've added it to the VoteSet.

Piecemeal PR for the mentioned issue, more work coming as we go.

### DIFF
--- a/types/validator_set.go
+++ b/types/validator_set.go
@@ -100,9 +100,9 @@ func (valSet *ValidatorSet) GetByAddress(address []byte) (index int, val *Valida
 }
 
 // GetByIndex returns the validator by index.
-// It returns nil values if index >= len(ValidatorSet.Validators)
+// It returns nil values if index < 0  or index >= len(ValidatorSet.Validators)
 func (valSet *ValidatorSet) GetByIndex(index int) (address []byte, val *Validator) {
-	if index >= len(valSet.Validators) {
+	if index < 0 || index >= len(valSet.Validators) {
 		return nil, nil
 	}
 	val = valSet.Validators[index]
--- a/types/vote_set.go
+++ b/types/vote_set.go
@@ -121,8 +121,8 @@ func (voteSet *VoteSet) Size() int {
 //		InvalidSignature | InvalidBlockHash | ConflictingVotes ]
 // Duplicate votes return added=false, err=nil.
 // Conflicting votes return added=*, err=ErrVoteConflictingVotes.
-// NOTE: vote should not be mutated after adding.
 // NOTE: VoteSet must not be nil
+// NOTE: Vote must not be nil
 func (voteSet *VoteSet) AddVote(vote *Vote) (added bool, err error) {
 	if voteSet == nil {
 		cmn.PanicSanity("AddVote() on nil VoteSet")
@@ -135,6 +135,9 @@ func (voteSet *VoteSet) AddVote(vote *Vote) (added bool, err error) {
 
 // NOTE: Validates as much as possible before attempting to verify the signature.
 func (voteSet *VoteSet) addVote(vote *Vote) (added bool, err error) {
+	if vote == nil {
+		return false, errNilVote
+	}
 	valIndex := vote.ValidatorIndex
 	valAddr := vote.ValidatorAddress
 	blockKey := vote.BlockID.Key()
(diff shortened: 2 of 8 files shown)


types/validator_set.go
```diff
@@ -100,9 +100,9 @@ func (valSet *ValidatorSet) GetByAddress(address []byte) (index int, val *Valida
 }
 
 // GetByIndex returns the validator by index.
-// It returns nil values if index >= len(ValidatorSet.Validators)
+// It returns nil values if index < 0  or index >= len(ValidatorSet.Validators)
 func (valSet *ValidatorSet) GetByIndex(index int) (address []byte, val *Validator) {
-	if index >= len(valSet.Validators) {
+	if index < 0 || index >= len(valSet.Validators) {
```
ebuchman: ++

types/vote_set.go
```diff
@@ -133,8 +134,13 @@ func (voteSet *VoteSet) AddVote(vote *Vote) (added bool, err error) {
 	return voteSet.addVote(vote)
 }
 
+var errNilVote = errors.New("nil/non-dereferenceable votes are not allowed")
```
ebuchman: maybe we can put this in vote.go with the other errors

types/vote_set.go
```diff
@@ -394,7 +403,7 @@ func (voteSet *VoteSet) HasTwoThirdsAny() bool {
 }
 
 func (voteSet *VoteSet) HasAll() bool {
-	return voteSet.sum == voteSet.valSet.TotalVotingPower()
+	return voteSet != nil && voteSet.sum == voteSet.valSet.TotalVotingPower()
```
ebuchman: if we're calling these functions on a nil voteSet, something is probably very wrong. Perhaps we should panic? Maybe not HasAll, but I'd think for MakeCommit. Hard to always know ...

types/vote_set_test.go
```diff
@@ -472,3 +474,153 @@ func TestMakeCommit(t *testing.T) {
...
+	require.Nil(t, vs.GetByIndex(1), "1: hasn't yet cast their vote")
+
+	// Invalid/nil/non-dereferenceable vote
+	added, err := vs.AddVote(nil)
+	require.False(t, added, "nil votes can't be dereferenced")
+	require.NotNil(t, err, "nil votes can't be dereferenced")
+
+	// Malicious vote
+	added, err = vs.AddVote(&Vote{
+		ValidatorIndex: -1,
```
ebuchman: lets try one with index too high too?

types/vote_set.go
odeke-em: Roger that, thanks.

types/vote_set_test.go
odeke-em: Yes we have a high one here https://github.com/tendermint/tendermint/pull/725/files/1b4593fc2945f1af0245c709a973f2a254751851#diff-b751ab127cf8b221ebc8084facc19a38R615

But I can def it clearer here as well.

types/vote_set.go
odeke-em: You are right, I was just bullet proofing it since this VoteSet code is used in `consensus`.

odeke-em:
Great point @ebuchman about the copying, that I had also mentioned above. So all the copying comes about because after adding a vote, the contract is that it shouldn't be mutated. In order to enforce that without trusting the outside caller, we need to copy the Vote itself which includes
BlockID.Copy() --> PartsHeader.Copy() --> Hash.Copy
which translates to Copy methods on at least 3 types.

types/validator_set.go
odeke-em: TODO for myself, re-PR this.

types/vote_set_test.go
```diff
@@ -472,3 +474,155 @@ func TestMakeCommit(t *testing.T) {
 	}
 
 }
+
+func TestVoteSetInvalid(t *testing.T) {
```
ebuchman: Let's keep this test

ebuchman:
@odeke-em I stole some of the changes in 5466720d

I'm going to close the PR. Please open a new one with VoteSet tests :)

PR #733, opened 2017-10-10
write config.toml with all defaults & improve config doc 

### LINKED ISSUE #705: config docs

zramsay:
closes #705 
links to:  #709, #636

Running `tendermint init` used to produce the following `config.toml`:
```
# This is a TOML config file.
# For more information, see https://github.com/toml-lang/toml

proxy_app = "tcp://127.0.0.1:46658"
moniker = "anonymous"
fast_sync = true
db_backend = "leveldb"
log_level = "state:info,*:error"

[rpc]
laddr = "tcp://0.0.0.0:46657"

[p2p]
laddr = "tcp://0.0.0.0:46656"
seeds = ""
```
which did not provide all that much information about the many config options. Now, the `config.toml` looks like:
```
# This is a TOML config file.
# For more information, see https://github.com/toml-lang/toml

##### main base config options #####

# TCP or UNIX socket address of the ABCI application,
# or the name of an ABCI application compiled in with the Tendermint binary
proxy_app = "tcp://127.0.0.1:46658"

# A custom human readable name for this node
moniker = "anonymous"

# If this node is many blocks behind the tip of the chain, FastSync
# allows them to catchup quickly by downloading blocks in parallel
# and verifying their commits
fast_sync = true

# Database backend: leveldb | memdb
db_backend = "leveldb"

# Database directory
db_path = "data"

# Output level for logging
log_level = "state:info,*:error"

##### additional base config options #####

# The ID of the chain to join (should be signed with every transaction and vote)
chain_id = ""

# Path to the JSON file containing the initial validator set and other meta data
genesis_file = "genesis.json"

# Path to the JSON file containing the private key to use as a validator in the consensus protocol
priv_validator_file = "priv_validator.json"

# Mechanism to connect to the ABCI application: socket | grpc
abci = "socket"

# TCP or UNIX socket address for the profiling server to listen on
prof_laddr = ""

# If true, query the ABCI app on connecting to a new peer
# so the app can decide if we should keep the connection or not
filter_peers = false

# What indexer to use for transactions
tx_index = "kv"


[rpc]

# TCP or UNIX socket address for the RPC server to listen on
laddr = "tcp://0.0.0.0:46657"

# TCP or UNIX socket address for the gRPC server to listen on
# NOTE: This server only supports /broadcast_tx_commit
grpc_laddr = ""

# Activate unsafe RPC commands like /dial_seeds and /unsafe_flush_mempool
unsafe = false


[p2p]

# Address to listen for incoming connections
laddr = "tcp://0.0.0.0:46656"

# Comma separated list of seed nodes to connect to
seeds = ""

# Path to address book
addr_book_file = "addrbook.json"

# Set true for strict address routability rules
addr_book_strict = true

# Time to wait before flushing messages out on the connection, in ms
flush_throttle_timeout = 100

# Maximum number of peers to connect to
max_num_peers = 50

# Maximum size of a message packet payload, in bytes
max_msg_packet_payload_size = 1024

# Rate at which packets can be sent, in bytes/second
send_rate = 512000

# Rate at which packets can be received, in bytes/second
recv_rate = 512000


[mempool]

recheck = true
recheck_empty = true
broadcast = true
wal_dir = "data/mempool.wal"


[consensus]

wal_file = "data/cs.wal/wal"
wal_light = false

# All timeouts are in milliseconds
timeout_propose = 3000
timeout_propose_delta = 500
timeout_prevote = 1000
timeout_prevote_delta = 500
timeout_precommit = 1000
timeout_precommit_delta = 500
timeout_commit = 1000

# Make progress as soon as we have all the precommits (as if TimeoutCommit = 0)
skip_timeout_commit = false

# BlockSize
max_block_size_txs = 10000
max_block_size_bytes = 1

# EmptyBlocks mode and possible interval between empty blocks in seconds
create_empty_blocks = true
create_empty_blocks_interval = 0

# Reactor sleep duration parameters are in milliseconds
peer_gossip_sleep_duration = 100
peer_query_maj23_sleep_duration = 2000
```
and the documentation for the `config.toml` now shows the full config file, replacing this:
screenshot from 2017-10-10 14-33-13
with this:
screenshot from 2017-10-10 14-34-00

### DIFF
--- a/config/toml.go
+++ b/config/toml.go
@@ -48,6 +48,8 @@ func writeConfigFile(configFilePath string) {
 	cmn.MustWriteFile(configFilePath, buffer.Bytes(), 0644)
 }
 
+// Note: any changes to the comments/variables/mapstructure
+// must be reflected in the appropriate struct in config/config.go
 const defaultConfigTemplate = `# This is a TOML config file.
 # For more information, see https://github.com/toml-lang/toml
 
@@ -98,7 +100,9 @@ filter_peers = {{ .BaseConfig.FilterPeers }}
 # What indexer to use for transactions
 tx_index = "{{ .BaseConfig.TxIndex }}"
 
+##### advanced configuration options #####
 
+##### rpc server configuration options #####
 [rpc]
 
 # TCP or UNIX socket address for the RPC server to listen on
@@ -111,7 +115,7 @@ grpc_laddr = "{{ .RPC.GRPCListenAddress }}"
 # Activate unsafe RPC commands like /dial_seeds and /unsafe_flush_mempool
 unsafe = {{ .RPC.Unsafe }}
 
-
+##### peer to peer configuration options #####
 [p2p]
 
 # Address to listen for incoming connections
@@ -141,7 +145,7 @@ send_rate = {{ .P2P.SendRate }}
 # Rate at which packets can be received, in bytes/second
 recv_rate = {{ .P2P.RecvRate }}
 
-
+##### mempool configuration options #####
 [mempool]
 
 recheck = {{ .Mempool.Recheck }}
@@ -149,7 +153,7 @@ recheck_empty = {{ .Mempool.RecheckEmpty }}
 broadcast = {{ .Mempool.Broadcast }}
 wal_dir = "{{ .Mempool.WalPath }}"
 
-
+##### consensus configuration options #####
 [consensus]
 
 wal_file = "{{ .Consensus.WalPath }}"
--- a/config/toml.go
+++ b/config/toml.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"strings"
 	"text/template"
 
 	cmn "github.com/tendermint/tmlibs/common"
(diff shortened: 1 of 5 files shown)


config/toml.go
```diff
@@ -17,34 +27,158 @@ func EnsureRoot(rootDir string) {
...
-var defaultConfigTmpl = `# This is a TOML config file.
+const defaultConfigTemplate = `# This is a TOML config file.
 # For more information, see https://github.com/toml-lang/toml
 
-proxy_app = "tcp://127.0.0.1:46658"
-moniker = "__MONIKER__"
-fast_sync = true
-db_backend = "leveldb"
-log_level = "state:info,*:error"
+##### main base config options #####
```
ebuchman: ok, this is amazing, but here's my crazy question .... can we get this code autogenerated from the config struct during a build process so we don't have to worry about redundant commentary here and in the structs??  is that a crazy thing to ask for :P

zramsay: maybe, in the future? I agree with the sentiment but have noticed that generally autogenerated codestuffs get mangled in one way or another. I'll add notices in the appropriate locations.

config/toml.go
```diff
@@ -17,34 +27,158 @@ func EnsureRoot(rootDir string) {
...
+		// TODO: Prompt user to replace if config file already exists.
+	} else {
+		// Ask user for moniker [moniker := cfg.Prompt("Type hostname: ", "anonymous")] (TODO)
+		// Then write the default config from template
+		writeConfigFile(configFilePath)
+	}
+}
+
+// XXX: this func should probably be called by cmd/tendermint/commands/init.go
+// alongside the writing of the genesis.json and priv_validator.json
```
zramsay: thoughts? seems awkward to not have the `config.toml` be initialized alongside `genesis.json` and `priv_validator.json` [here](https://github.com/tendermint/tendermint/blob/master/cmd/tendermint/commands/init.go#L23-L36)

config/toml_test.go
```diff
@@ -49,9 +49,143 @@ func TestEnsureTestRoot(t *testing.T) {
...
 	require.Nil(err)
-	assert.Equal([]byte(testConfig("anonymous")), data)
+	assert.Equal([]byte(defaultConfigHardcoded), data)
 
 	// TODO: make sure the cfg returned and testconfig are the same!
 
 	ensureFiles(t, rootDir, "data", "genesis.json", "priv_validator.json")
 }
+
+const defaultConfigHardcoded = `# This is a TOML config file.
```
melekes: does it mean we'll have to change this every time we change the config?

zramsay: yea.

melekes: can't we generate it on the fly or something?

zramsay: it's only used by the test. The actual config is generated on the fly, but then the test would be redundant, no? Could use a `strings.Contains` instead to check that key elements of the config are present; something like:
```
var elems = []string{"moniker", "seeds", "proxy_app", "fast_sync", "create_empty_blocks"}
for _, e := range elems {
        if !strings.Contains(configFile, e) {
                t.Fail()
       }
}
```

melekes: > Could use a strings.Contains instead to check that key elements of the config are present

I am in favor of doing this

zramsay:
merging to `config` branch

Issue #482, opened 2017-05-03
Support historical abci queries


ethanfrey:
It is a nice use-case to be able to query for the abci state a few blocks ago, or at least to be able to specify the same block height for multiple queries to get a consistent snapshot even if the block progresses in the meantime. Jae also thinks this is very important for removing race conditions from ibc-like communication.

This was at one time considered part of #355 but ended up a much bigger project, and that "just" got proofs from the present block (which already was a huge improvement).  Now that historical queries should soon be [supported in go-merkle](https://github.com/tendermint/go-merkle/pull/24), I want to create this issue to remind us to expose this thoughout the whole stack - merkleeyes, abci, tendermint. So that a client can get a proof of the state from a few blocks ago.

ethanfrey:
@ebuchman Another outstanding issue we have talked about for months... 0.11?  0.12?

ebuchman:
So what changes do we need to make to tendermint to support this? Just add a `height` arg to `/abci_query`  ? Anything else?

ethanfrey:
Yes. And support it in all abci apps that we offer.

ethanfrey:
This is the center point for the various repos I guess

ebuchman:
Ok. Let's denftly get that into tendermint for 0.10. Whether or not it's actually available with the apps  in time with the 0.10 release is a second question, but at least should be soon after

ethanfrey:
Wait.. Is the historical query pr of Paul accepted yet? That is the base functionality we need to expose. I thought that wasn't going to make 0.11.

I think it is good to do rpc, config, logging cleanup together. And in 0.11 add the historical queries, for validators as well. Together they are more useful.

melekes:
```
ABCIQuery(path string, data []byte) { return ABCIQueryWithOptions(path, data, abci.DefaultQueryOptions) }
ABCIQueryWithOptions(path string, data []byte, opts abci.QueryOptions) { ... }

type QueryOptions struct {
  Height uint64
}
```

zramsay:
#734

PR #734, opened 2017-10-10
support historical abci queries

melekes:
See #482

### DIFF
--- a/rpc/client/types.go
+++ b/rpc/client/types.go
@@ -1,0 +1,13 @@
+package client
+
+// ABCIQueryOptions can be used to provide options for ABCIQuery call other
+// than the DefaultABCIQueryOptions.
+type ABCIQueryOptions struct {
+	Height uint64
+	Prove  bool
+}
+
+// DefaultABCIQueryOptions are latest height (0) and prove equal to true.
+func DefaultABCIQueryOptions() ABCIQueryOptions {
+	return ABCIQueryOptions{Height: 0, Prove: true}
+}
--- a/rpc/client/types.go
+++ b/rpc/client/types.go
@@ -3,11 +3,9 @@ package client
 // ABCIQueryOptions can be used to provide options for ABCIQuery call other
 // than the DefaultABCIQueryOptions.
 type ABCIQueryOptions struct {
-	Height uint64
-	Prove  bool
+	Height  uint64
+	Trusted bool
 }
 
 // DefaultABCIQueryOptions are latest height (0) and prove equal to true.
-func DefaultABCIQueryOptions() ABCIQueryOptions {
-	return ABCIQueryOptions{Height: 0, Prove: true}
-}
+var DefaultABCIQueryOptions = ABCIQueryOptions{Height: 0, Trusted: false}
--- a/rpc/client/types.go
+++ b/rpc/client/types.go
@@ -7,5 +7,6 @@ type ABCIQueryOptions struct {
 	Trusted bool
 }
 
-// DefaultABCIQueryOptions are latest height (0) and prove equal to true.
+// DefaultABCIQueryOptions are latest height (0) and trusted equal to false
+// (which will result in a proof being returned).
 var DefaultABCIQueryOptions = ABCIQueryOptions{Height: 0, Trusted: false}
--- a/rpc/core/abci.go
+++ b/rpc/core/abci.go
@@ -41,18 +41,18 @@ import (
 //
 // ### Query Parameters
 //
-// | Parameter | Type   | Default | Required | Description                           |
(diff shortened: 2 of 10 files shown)


rpc/core/routes.go
```diff
@@ -29,7 +29,7 @@ var Routes = map[string]*rpc.RPCFunc{
 	"broadcast_tx_async":  rpc.NewRPCFunc(BroadcastTxAsync, "tx"),
 
 	// abci API
-	"abci_query": rpc.NewRPCFunc(ABCIQuery, "path,data,prove"),
+	"abci_query": rpc.NewRPCFunc(ABCIQuery, "path,data,height,prove"),
```
ebuchman: this is definitely the simplest approach, but I believe @cloudhead was arguing to remove the bool from here and use a separate `abci_query_trusted` method for prove = false, so the default is it's always true. Would that be better ? It might be nice if the options struct was an arg here that could go in through the jsonrpc, but itd be a total pain to call it using the URI - maybe we just shouldn't worry too much about that usecase, I kind of see us eventually deprecating the URI interface eventually due to issues like this

melekes: > be a total pain to call it using the URI

yeah.

As I understood, @ethanfrey and @cloudhead don't care as long as the HttpClient (or LocalClient) defaults to prove=true. And that's what I did: our both clients have this interface

```
ABCIQuery(path string, data []byte) { return ABCIQueryWithOptions(path, data, abci.DefaultQueryOptions) }
ABCIQueryWithOptions(path string, data []byte, opts abci.QueryOptions) { ... }
```

Or this is counterintuitive? I mean when the rpc core method defaults to prove=false, but the official client defaults to prove=true?

rpc/client/types.go
```diff
@@ -0,0 +1,13 @@
...
+// than the DefaultABCIQueryOptions.
+type ABCIQueryOptions struct {
+	Height uint64
+	Prove  bool
+}
+
+// DefaultABCIQueryOptions are latest height (0) and prove equal to true.
+func DefaultABCIQueryOptions() ABCIQueryOptions {
+	return ABCIQueryOptions{Height: 0, Prove: true}
+}
```
cloudhead: It would be nice if this was a `var` instead of a function.

rpc/client/types.go
```diff
@@ -0,0 +1,13 @@
+package client
+
+// ABCIQueryOptions can be used to provide options for ABCIQuery call other
+// than the DefaultABCIQueryOptions.
+type ABCIQueryOptions struct {
+	Height uint64
+	Prove  bool
```
cloudhead: Since we're moving to `Prove: true` as the default, maybe switching this to `Trusted bool`, where `Trusted == !Prove` would be better, so that an empty `ABCIQueryOptions{}` would have the expected defaults.

rpc/core/abci.go
```diff
@@ -45,12 +45,14 @@ import (
 // |-----------+--------+---------+----------+---------------------------------------|
 // | path      | string | false   | false    | Path to the data ("/a/b/c")           |
 // | data      | []byte | false   | true     | Data                                  |
+// | height    | uint64 | 0       | false    | Height (0 means latest)               |
 // | prove     | bool   | false   | false    | Include a proof of the data inclusion |
-func ABCIQuery(path string, data data.Bytes, prove bool) (*ctypes.ResultABCIQuery, error) {
+func ABCIQuery(path string, data data.Bytes, height uint64, prove bool) (*ctypes.ResultABCIQuery, error) {
```
cloudhead: Can this method have a similar signature to the other one, with `ABCIQueryOptions`?

rpc/core/routes.go
cloudhead: I was more concerned with the Go API - but I do think they should all default to prove=true/trusted=false

melekes: "yeah, it will be much easier if we rename prove to trusted"

rpc/core/abci.go
melekes: will be very hard to do

cloudhead:
Good!

rpc/client/types.go
```diff
@@ -0,0 +1,11 @@
+package client
+
+// ABCIQueryOptions can be used to provide options for ABCIQuery call other
+// than the DefaultABCIQueryOptions.
+type ABCIQueryOptions struct {
+	Height  uint64
+	Trusted bool
+}
+
+// DefaultABCIQueryOptions are latest height (0) and prove equal to true.
```
ebuchman: update comment - "trusted equal to false (which will result in a proof being returned)"

Issue #723, opened 2017-10-09
types/Header: ambiguous comment about Hash() returning nil if "required" fields are missing


odeke-em:
So currently in 
https://github.com/tendermint/tendermint/blob/7682ad9a60162dd17fd6f61aeed7049a8635ac78/types/block.go#L172-L188

The comment says:
`// NOTE: hash is nil if required fields are missing.`

But that's a bit unclear/over-generalized. Perhaps later on we should improve that doc to say it returns nil if ValidatorsHash is nil because I'd have thought not including other fields would also cause it to fail. 
```go
package test_it

import (
        "testing"

        "github.com/google/gofuzz"

        "github.com/tendermint/tendermint/types"
)

func TestFuzzMerkleHash(t *testing.T) {
        f := fuzz.New()

        n := 100000
        nonNilCount := uint64(0)
        for i := 0; i < n; i++ {
                hdr := new(types.Header)
                f.Fuzz(hdr)
                hash := hdr.Hash()
                if hash != nil {
                        nonNilCount += 1
                }
        }
        t.Logf("nonNilCount: %d/%d", nonNilCount, n)
}
```

Shooting at it with a fuzzer reveals that most times "randomly" folks can get it right 80% with guesses so that comment perhaps needs an update
```shell
Emmanuels-MacBook-Pro-2:types emmanuelodeke$ go test -v -run=FuzzMerkle
=== RUN   TestFuzzMerkleHash
--- PASS: TestFuzzMerkleHash (3.47s)
	hdr_test.go:595: nonNilCount: 80128/100000
Emmanuels-MacBook-Pro-2:types emmanuelodeke$ go test -v -run=FuzzMerkle
=== RUN   TestFuzzMerkleHash
--- PASS: TestFuzzMerkleHash (3.55s)
	hdr_test.go:595: nonNilCount: 80131/100000
Emmanuels-MacBook-Pro-2:types emmanuelodeke$ go test -v -run=FuzzMerkle
=== RUN   TestFuzzMerkleHash
--- PASS: TestFuzzMerkleHash (3.63s)
	hdr_test.go:595: nonNilCount: 80074/100000
Emmanuels-MacBook-Pro-2:types emmanuelodeke$ go test -v -run=FuzzMerkle
=== RUN   TestFuzzMerkleHash
--- PASS: TestFuzzMerkleHash (4.08s)
	hdr_test.go:595: nonNilCount: 80426/100000
[...]

PR #736, opened 2017-10-11
types: add version string to block header

### LINKED ISSUE #495: Add version string to block header
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):

Feature Request (by ebuchman)

For 0.10 tendermint release

Add one string to the block header with a version string, which is set when we compile the software.  This string will be equal between two nodes iff they are running the same version of the software.  This will allow us to much more easily coordinate upgrades across nodes.

This may contain two parts, like API version enabled, and API version activated

zramsay:
- closes #495, #723

### DIFF
--- a/types/block.go
+++ b/types/block.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/tendermint/tendermint/version"
+
 	wire "github.com/tendermint/go-wire"
 	"github.com/tendermint/go-wire/data"
 	cmn "github.com/tendermint/tmlibs/common"
@@ -22,7 +24,6 @@ type Block struct {
 }
 
 // MakeBlock returns a new block and corresponding part set from the given information
-// TODO: version
 func MakeBlock(height int, chainID string, txs []Tx, commit *Commit,
 	prevBlockID BlockID, valHash, appHash []byte, partSize int) (*Block, *PartSet) {
 	block := &Block{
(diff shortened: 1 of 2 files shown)


types/block.go
```diff
@@ -202,6 +206,7 @@ func (h *Header) StringIndented(indent string) string {
 %s  Data:           %v
 %s  Validators:     %v
 %s  App:            %v
+%s  Version:	    %v
```
melekes: formatting

zramsay: hrm, looks ok here https://github.com/tendermint/tendermint/blob/495-version-in-block-header/types/block.go#L209 and also locally:
screenshot from 2017-10-12 08-27-56

am I missing something?

odeke-em: @zramsay you have a tab in there, if my assumption is that you are using vim in Ubuntu, type `:set list` and you'll see the tab right after `Version:^I`, of
which we aren't using tabs in the entire format string, please remove it and manually line it up with spaces. Once done, `:set nolist` to return to normal.

zramsay: thanks, fixed

melekes:
Don't forget to update the CHANGELOG

odeke-em:
@melekes he updated the changelog, we are ready for flight! Thanks @zramsay!

ebuchman:
I don't want to merge this just yet until there is some discussion on the use of this variable. We need to have a larger discussion about how to roll out upgrades

ebuchman:
Thanks Zach, but gonna close pending further discussion

Issue #713, opened 2017-10-04
replace tendermint/merkleeys/iavl with tendermint/iavl


zramsay:
in rpc/client/rpc_test.go

zramsay:
#768

PR #737, opened 2017-10-11
WIP: rpc: use /iavl repo in test

zramsay:
#713 

Test fails with:
```
Error Trace:    rpc_test.go:178
Error:	        Expected nil, but got: &errors.errorString{s:"Unexpected type byte BE for ptr of untyped thing"}
```

still debugging, maybe @cloudhead knows the fix off hand

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -118,10 +118,8 @@ imports:
   subpackages:
   - data
   - data/base58
-- name: github.com/tendermint/merkleeyes
-  version: 2a93256d2c6fbcc3b55673c0d2b96a7e32c6238b
-  subpackages:
-  - iavl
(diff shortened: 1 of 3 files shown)


melekes:
The problem is that https://github.com/tendermint/tendermint/blob/e2e50bc0fcf81a8580f664a695b425727d90e510/rpc/client/main_test.go#L16 dummy uses an old version! so, if you want this to work, you need to update abci repo as well OR wait until I do it https://github.com/tendermint/abci/pull/113

zramsay:
ah, ok. This can block until abci/113 then, thanks.

ebuchman:
replaced by https://github.com/tendermint/tendermint/pull/768

PR #741, opened 2017-10-13
rpc/client: use compile time assertions instead of methods

odeke-em:


odeke-em:
@ethanfrey please take a look.

rpc/client/httpclient.go
```diff
@@ -39,17 +39,14 @@ func NewHTTP(remote, wsEndpoint string) *HTTP {
...
-
-func (c *HTTP) _assertIsNetworkClient() NetworkClient {
-	return c
-}
-
-func (c *HTTP) _assertIsEventSwitch() types.EventSwitch {
-	return c
-}
+var (
+	// Some compile time assertions to ensure that
```
melekes: don't think this comment is necessary

rpc/client/localclient.go
```diff
@@ -41,13 +41,12 @@ func NewLocal(node *nm.Node) Local {
 	}
 }
 
-func (c Local) _assertIsClient() Client {
-	return c
-}
+var (
+	aLocal = Local{}
```
ebuchman: why this?

ebuchman:
Otherwise lgtm

rpc/client/localclient.go
odeke-em: Because below, we'd end up creating more than one `Local{}`

rpc/client/httpclient.go
odeke-em: Roger that, @melekes I'll update that, thanks.

rpc/client/localclient.go
odeke-em: However, I'll remove it for clarity and instead use `Local{}` all the way.

odeke-em:
I've updated the PR to address both y'all comments, PTAL.

Issue #722, opened 2017-10-06
ValidatorSet: overflowing lengths cause runtime crashes in go-wire


odeke-em:
Found with go-fuzz, the inputs [1] below when fed into `*ValidatorSet.FromBytes` cause a runtime panic [2]

where the running code is
```go
package main

import "github.com/tendermint/tendermint/types"

func main() {
	vs := new(types.ValidatorSet)
	vs.FromBytes([]byte{0x01, 0x01, 0x30, 0x01, 0x06, 0x30, 0x30, 0x30, 0x30, 0x30,0x30})
}
```

[1] Inputs
```input
        inputs := [][]byte{
                {
                        0x01, 0x00, 0x00, 0x30,
                },
                {
                        0x01, 0x01, 0x30, 0x01, 0x06, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30,
                },
                {
                        0x01, 0x01, 0x02, 0x01, 0x01, 0x14, 0x5e, 0xbe, 0x09, 0x30, 0x40, 0xbe, 0x46, 0x15, 0x96, 0x0a,
                },
                {
                        0x01, 0x01, 0x03, 0x01, 0x01, 0x14, 0x89, 0xca, 0x3e, 0xfb, 0x67, 0x05, 0x07, 0xa7, 0x9d, 0xaf,
                        0xd0, 0x24, 0xdd, 0x8d, 0x11, 0xd0, 0x56, 0x16, 0xb3, 0x28, 0x01, 0xd3, 0xab, 0x48, 0xc9, 0x9a,
                        0xbc, 0x8f, 0x23, 0x93, 0xe8, 0xd3, 0x4c, 0x5c, 0x14, 0x41, 0x53, 0xaa, 0x46, 0x4e, 0x92, 0xf9,
                        0x81, 0xb7, 0xf0, 0x72, 0x8c, 0xaf, 0xae, 0xb5, 0x66, 0xe1, 0x25, 0x00, 0x00, 0x00, 0x00, 0x00,
                        0x00, 0x00, 0x64, 0xf2, 0x25, 0x07, 0x5d, 0x00, 0x00, 0x00, 0x64, 0x01, 0x01, 0x14, 0xe1, 0x56,
                        0x43, 0x9c, 0x68, 0x5a, 0x37, 0xf1, 0x75, 0x91, 0xa4, 0xac, 0xc4, 0x2a, 0xb3, 0xbd, 0x4a, 0x0a,
[...]

PR #742, opened 2017-10-13
types: update go-wire dep + test for fuzz malicious vectors

odeke-em:
Closes https://github.com/tendermint/tendermint/issues/722
Requires https://github.com/tendermint/go-wire/pull/38

Updated the go-wire dependency and locked in the test
for the discovered malicous inputs after fuzzing.


odeke-em:
The test is the way it is because we want to check for strict panics i.e either something mentioning `overflow` or `EOF`, any other panic is an indication that the code isn't fixed.

types/validator_set_test.go
```diff
@@ -206,3 +207,63 @@ func BenchmarkValidatorSetCopy(b *testing.B) {
 		vset.Copy()
 	}
 }
+
+func TestFuzzValidatorSetFromBytes(t *testing.T) {
+	// See Issue https://github.com/tendermint/tendermint/issues/722
+	// which was fixed by https://github.com/tendermint/go-wire/pull/38
+	vectors := [][]byte{
+		{
+			0x01, 0x01, 0x30, 0x01, 0x06, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30,
```
ebuchman: yeh not thrilled about this kind of thing in unit tests. way too difficult to maintain.

Perhaps we could start an external testing system that can keep binary encoded things and make sure they deserialize properly for sake of cross implementation testing but this feels to excessive/dirty in the go code.

odeke-em: @ebuchman that's malicious code that would be crafted by an attacker, it was auto-generated by the fuzzer hence doesn't directly translate to normal code and the test is testing FromBytes.

Would you like me instead to put these into some directory such as `testdata` instead?

odeke-em: And for references for what generated them, it was from https://github.com/tendermint/tendermint/commit/2ba16561d00b6675ee996bd92cad734fae29a318.

melekes: Maybe what we should do instead is to run this fuzzier once in a while (before every release) to make sure we didn't introduce new bugs?

ebuchman: Do we even use the FromBytes function anywhere?! Let's just get rid of it :P

ebuchman: Ok well, maybe that's not great, since we obviously deserialize from bytes elsewhere, but maybe we can just use the fuzzers on those, since from what I can tell we don't ever call FromBytes here. Though maybe it's still useful for finding a possible bug like this? 

Also, isn't all this just because we're not using a limit in ReadBinary ?

ebuchman: lets run fuzzers on release; https://github.com/tendermint/tendermint/issues/752

odeke-em: `Maybe what we should do instead is to run this fuzzier once in a while (before every release) to make sure we didn't introduce new bugs?`
Great and interesting idea, but the problem is that hopefully we remember to do this, or perhaps we can put some test directives to the release automation.

`Also, isn't all this just because we're not using a limit in ReadBinary ?`
Yes, a combination of `0` and malicious vectors, of which 0 as a limit is used allover the place manifests in this bug.

`Do we even use the FromBytes function anywhere?! Let's just get rid of it :P`
Sure, we can get rid of it but at least lets update the go-wire dependency since many places use a limit of 0, unless am mistaken.

ebuchman: > perhaps we can put some test directives to the release automation

Yeh we should look to add this.

> at least lets update the go-wire dependency since many places use a limit of 0, 

What's the proposed update? To prevent using 0? Sometimes it's nice like in tests. But maybe it's worth forbidding it from being 0 and just put some effort into the tests to provide a limit

ebuchman:
Closing this. Will remove FromBytes

PR #743, opened 2017-10-13
Update getting-started.rst

zramsay:
removed mention of the deprecated type byte


zramsay:
hrm @codecov-io what is the deal?! .5% increase after removing docs?

PR #744, opened 2017-10-14
adding the trust metric design doc and images

caffix:


caffix:
@odeke-em, I imagined that the trust metric would be added to the peer struct. The way the trust metric is designed currently, it can be used as a trust ranking for anything, not only peers. Just let me know how you'd prefer to interface with the metric though, and I can change it

odeke-em:
Aye aye @caffix. Thing though is that peer isn't a struct, it is an interface, so perhaps we'll add this to the different reactors, amirite @ebuchman?

PR #745, opened 2017-10-15
rpc/lib/server: add handlers tests for params inclusion

odeke-em:
Follow up of PR https://github.com/tendermint/tendermint/pull/724
For https://github.com/tendermint/tendermint/issues/708

Reported initially in #708, this bug was reconfirmed
by the fuzzer.

This fix ensures that:
* if the user doesn't pass in `"id"` that we send them back
a message in an error telling them to send `"id"`. Previously
we let the handler return a 200 with nothing.
* passing in nil `params` doesn't crash
* not passing in `params` doesn't crash
* passing in non-JSON parseable data to `params` doesn't crash

### DIFF
--- a/rpc/lib/server/handlers.go
+++ b/rpc/lib/server/handlers.go
@@ -118,6 +118,8 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, logger log.Logger) http.Han
 		// The Server MUST NOT reply to a Notification, including those that are within a batch request.
 		if request.ID == "" {
 			logger.Debug("HTTPJSONRPC received a notification, skipping... (please send a non-empty ID if you want to call a method)")
+			// Not sending back a response here because according the JSONRPC
+			// specification Section 4.1, we SHOULD NOT one back when "id" == "".
 			return
 		}
 		if len(r.URL.Path) > 1 {
--- a/rpc/lib/server/handlers.go
+++ b/rpc/lib/server/handlers.go
@@ -118,8 +118,6 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, logger log.Logger) http.Han
 		// The Server MUST NOT reply to a Notification, including those that are within a batch request.
 		if request.ID == "" {
 			logger.Debug("HTTPJSONRPC received a notification, skipping... (please send a non-empty ID if you want to call a method)")
-			// Not sending back a response here because according the JSONRPC
-			// specification Section 4.1, we SHOULD NOT one back when "id" == "".
 			return
 		}
 		if len(r.URL.Path) > 1 {
--- a/rpc/lib/server/handlers_test.go
+++ b/rpc/lib/server/handlers_test.go
@@ -40,15 +40,18 @@ func TestRPCParams(t *testing.T) {
 		payload string
 		wantErr string
 	}{
+		// bad
 		{`{"jsonrpc": "2.0", "id": "0"}`, "Method not found"},
 		{`{"jsonrpc": "2.0", "method": "y", "id": "0"}`, "Method not found"},
-		{`{"jsonrpc": "2.0", "method": "c", "id": "0", "params": null}`, ""},
-		{`{"method": "c", "id": "0", "params": {}}`, ""},
 		{`{"method": "c", "id": "0", "params": a}`, "invalid character"},
-		{`{"method": "c", "id": "0", "params": ["a", 10]}`, ""},
 		{`{"method": "c", "id": "0", "params": ["a"]}`, "got 1"},
 		{`{"method": "c", "id": "0", "params": ["a", "b"]}`, "of type int"},
 		{`{"method": "c", "id": "0", "params": [1, 1]}`, "of type string"},
+
+		// good
+		{`{"jsonrpc": "2.0", "method": "c", "id": "0", "params": null}`, ""},
+		{`{"method": "c", "id": "0", "params": {}}`, ""},
+		{`{"method": "c", "id": "0", "params": ["a", 10]}`, ""},
 	}
 
 	for i, tt := range tests {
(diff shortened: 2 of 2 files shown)


rpc/lib/server/handlers.go
```diff
@@ -118,6 +118,7 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, logger log.Logger) http.Han
 		// The Server MUST NOT reply to a Notification, including those that are within a batch request.
 		if request.ID == "" {
 			logger.Debug("HTTPJSONRPC received a notification, skipping... (please send a non-empty ID if you want to call a method)")
+			WriteRPCResponseHTTP(w, types.RPCInvalidParamsError(request.ID, errors.New(`expecting "id"`)))
```
melekes: this violates the JSONRPC specification: `The Server MUST NOT reply to a Notification, including those that are within a batch request.`

rpc/lib/server/handlers_test.go
```diff
@@ -0,0 +1,75 @@
...
+// Ensure that nefarious/unintended inputs to `params`
+// do not crash our RPC handlers.
+// See Issue https://github.com/tendermint/tendermint/issues/708.
+func TestRPCParams(t *testing.T) {
+	funcMap := map[string]*rs.RPCFunc{
+		"c": rs.NewRPCFunc(func(s string, i int) (string, error) { return "foo", nil }, "s,i"),
+	}
+	mux := http.NewServeMux()
+	buf := new(bytes.Buffer)
+	lgr := log.NewTMLogger(buf)
```
melekes: I'd prefer `logger`, but it's your call

rpc/lib/server/handlers_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+		recv := new(types.RPCResponse)
+		assert.Nil(t, json.Unmarshal(blob, recv), "#%d: expecting successful parsing of an RPCResponse:\nblob: %s", i, blob)
+		assert.NotEqual(t, recv, new(types.RPCResponse), "#%d: not expecting a blank RPCResponse", i)
+
+		if tt.wantErr != "" {
+			assert.False(t, statusOK(recv.Error.Code), "#%d: not expecting a 2XX success code", i)
+			// The wanted error is either in the message or the data
+			assert.Contains(t, recv.Error.Message+recv.Error.Data, tt.wantErr, "#%d: expected substring", i)
+			continue
```
melekes: maybe else branch would be better?

rpc/lib/server/handlers.go
odeke-em: Oh, TIL, thanks for that, I did not know about that, I've been in plain HTTP land for too long :)

rpc/lib/server/handlers_test.go
odeke-em: I can def update that for clarity, thanks.

odeke-em:
Alright, thank you @melekes I've updated with your feedback, PTAL.

rpc/lib/server/handlers.go
```diff
@@ -118,6 +118,8 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, logger log.Logger) http.Han
 		// The Server MUST NOT reply to a Notification, including those that are within a batch request.
```
melekes: why don't you want to update this comment? instead of writing a new one

rpc/lib/server/handlers.go
```diff
@@ -118,6 +118,8 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, logger log.Logger) http.Han
...
 		if request.ID == "" {
 			logger.Debug("HTTPJSONRPC received a notification, skipping... (please send a non-empty ID if you want to call a method)")
+			// Not sending back a response here because according the JSONRPC
+			// specification Section 4.1, we SHOULD NOT one back when "id" == "".
```
ebuchman: missing word. `SHOULD NOT send one`

rpc/lib/server/handlers_test.go
```diff
@@ -0,0 +1,81 @@
...
+	buf := new(bytes.Buffer)
+	logger := log.NewTMLogger(buf)
+	rs.RegisterRPCFuncs(mux, funcMap, logger)
+
+	tests := []struct {
+		payload      string
+		wantErr      string
+		notification bool
+	}{
+		{`{"jsonrpc": "2.0"}`, "", true}, // The server SHOULD NOT respond to a notification according to JSONRPC Section 4.1.
```
ebuchman: feels like we should just test the notification separately

rpc/lib/server/handlers_test.go
```diff
@@ -0,0 +1,81 @@
...
+		assert.Nil(t, json.Unmarshal(blob, recv), "#%d: expecting successful parsing of an RPCResponse:\nblob: %s", i, blob)
+		assert.NotEqual(t, recv, new(types.RPCResponse), "#%d: not expecting a blank RPCResponse", i)
+
+		if tt.wantErr == "" {
+			assert.Nil(t, recv.Error, "#%d: not expecting an error", i)
+		} else {
+			assert.False(t, statusOK(recv.Error.Code), "#%d: not expecting a 2XX success code", i)
+			// The wanted error is either in the message or the data
+			assert.Contains(t, recv.Error.Message+recv.Error.Data, tt.wantErr, "#%d: expected substring", i)
+			continue
```
ebuchman: unecessary `continue` ?

rpc/lib/server/handlers.go
odeke-em: I hadn't even seen that, which is what prompted me to firstly erraneously return a response and then later update the comment since with a notification id=="". Thank you for raising that. I'll update the comment up there.

rpc/lib/server/handlers_test.go
odeke-em: Thanks for the catch. That was a vestige of the previous round of commit.

odeke-em:
`Thanks for the tests! Maybe we could make them cleaner with a function that creates the json and just takes whatever strings as args? eg. jsonRequest("2.0", "c", "0", "[5]")`
Could you give me an example of what you mean here? Those strings are the raw payload for the HTTP request, we aren't doing any JSON parsing nor testing JSON responses, just checking for errors returned.

odeke-em:
Thank you @ebuchman for the advice on isolating out the Notification test, I've fixed that, PTAL.

ebuchman:
Thanks odeke!

PR #746, opened 2017-10-15
701 add dev docs in java

srmo:
[docs] added new code to allow creation of collapsible code blocks
[docs] added a java example for each go example in app development section

### DIFF
--- a/docs/_static/custom_collapsible_code.css
+++ b/docs/_static/custom_collapsible_code.css
@@ -1,0 +1,17 @@
+.toggle {
+    padding-bottom: 1em ;
+}
+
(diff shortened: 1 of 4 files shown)


zramsay:
Amazing! Have tested it locally & it works great. I'll probably merge this into a `docs-staging` branch and trigger those builds on read the docs for others to easily see.

srmo:
@zramsay One thing to mention is that this requires ES6 support in the browser, so it might not work with older browsers.

zramsay:
merging to `docs-staging`, @nylira can you review the js code here?

Issue #747, opened 2017-10-15
types.ConsensusParams: document the maximum hard limit on BlockSizeParams.MaxBytes 


odeke-em:
I was just reading through 
https://github.com/tendermint/tendermint/blob/d4634dc6832a7168c2536e7c71bfba4a8ca857be/types/params.go#L82

in order to write tests and noticed that the maximum hard limit is an unexposed constant that doesn't seem to be documented anywhere. Perhaps we should document this hardlimit.

PR #748, opened 2017-10-15
types: ConsensusParams test + document the ranges/limits

odeke-em:
Fixes https://github.com/tendermint/tendermint/issues/747

* Document the unmentioned limits for ConsensusParams.Validate()
* Make the limit for ConsensusParams.BlockSizeParams.MaxBytes
clear at 100MiB

### DIFF
--- a/types/params.go
+++ b/types/params.go
@@ -5,7 +5,7 @@ import (
 )
 
 const (
-	maxBlockSizeBytes = 104857600 // 100MB
+	_100MiB = 104857600
 )
 
 // ConsensusParams contains consensus critical parameters
@@ -69,6 +69,10 @@ func DefaultBlockGossipParams() BlockGossipParams {
 
 // Validate validates the ConsensusParams to ensure all values
 // are within their allowed limits, and returns an error if they are not.
+// Expecting:
+// * BlockSizeParams.MaxBytes > 0
+// * BlockGossipParams.BlockPartSizeBytes > 0
+// * BlockSizeParams.MaxBytes <= 100MiB
 func (params *ConsensusParams) Validate() error {
 	// ensure some values are greater than 0
 	if params.BlockSizeParams.MaxBytes <= 0 {
@@ -79,9 +83,9 @@ func (params *ConsensusParams) Validate() error {
 	}
 
 	// ensure blocks aren't too big
-	if params.BlockSizeParams.MaxBytes > maxBlockSizeBytes {
+	if params.BlockSizeParams.MaxBytes > _100MiB {
 		return errors.Errorf("BlockSizeParams.MaxBytes is too big. %d > %d",
-			params.BlockSizeParams.MaxBytes, maxBlockSizeBytes)
+			params.BlockSizeParams.MaxBytes, _100MiB)
 	}
 	return nil
 }
--- a/types/params.go
+++ b/types/params.go
@@ -5,7 +5,7 @@ import (
 )
 
 const (
-	_100MiB = 104857600
+	maxBlockSizeBytes = 104857600 // 100MB
 )
 
 // ConsensusParams contains consensus critical parameters
@@ -18,7 +18,7 @@ type ConsensusParams struct {
 
 // BlockSizeParams contain limits on the block size.
 type BlockSizeParams struct {
-	MaxBytes int `json:"max_bytes"` // NOTE: must not be 0
+	MaxBytes int `json:"max_bytes"` // NOTE: must not be 0 nor greater than 100MB
 	MaxTxs   int `json:"max_txs"`
 	MaxGas   int `json:"max_gas"`
 }
@@ -69,10 +69,6 @@ func DefaultBlockGossipParams() BlockGossipParams {
 
 // Validate validates the ConsensusParams to ensure all values
 // are within their allowed limits, and returns an error if they are not.
-// Expecting:
-// * BlockSizeParams.MaxBytes > 0
-// * BlockGossipParams.BlockPartSizeBytes > 0
-// * BlockSizeParams.MaxBytes <= 100MiB
 func (params *ConsensusParams) Validate() error {
 	// ensure some values are greater than 0
 	if params.BlockSizeParams.MaxBytes <= 0 {
@@ -83,9 +79,9 @@ func (params *ConsensusParams) Validate() error {
 	}
 
 	// ensure blocks aren't too big
-	if params.BlockSizeParams.MaxBytes > _100MiB {
+	if params.BlockSizeParams.MaxBytes > maxBlockSizeBytes {
 		return errors.Errorf("BlockSizeParams.MaxBytes is too big. %d > %d",
-			params.BlockSizeParams.MaxBytes, _100MiB)
+			params.BlockSizeParams.MaxBytes, maxBlockSizeBytes)
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


types/params.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 const (
-	maxBlockSizeBytes = 104857600 // 100MB
+	_100MiB = 104857600
```
melekes: 👎  we're losing meaning here

types/params.go
```diff
@@ -69,6 +69,10 @@ func DefaultBlockGossipParams() BlockGossipParams {
 
 // Validate validates the ConsensusParams to ensure all values
 // are within their allowed limits, and returns an error if they are not.
+// Expecting:
```
melekes: why duplicate the code? is this what we want from documentation? step by step duplication of the code

melekes: we should probably document `maxBlockSizeBytes` somewhere here https://tendermint.readthedocs.io/en/master/specification/block-structure.html

types/params_test.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/tendermint/tendermint/types"
+)
+
+func TestConsensusParamsValidation(t *testing.T) {
+	c1 := new(types.ConsensusParams)
```
melekes: why not table tests?

types/params.go
odeke-em: The problem though is that a plain maxBlockSizeBytes itself requires having to look it up and isn't documented anywhere, IMHO `_100MiB` is self documenting.

types/params.go
odeke-em: I don't think this is duplicating the code, it saves folks from having to open the code and having to read each line, by line. That's what am trying to avoid. I had to go through the guts of it in order to write the tests themselves.

types/params_test.go
odeke-em: This seemed simpler, usually I'd use table tests but here they seemed simpler. However, I can add table tests since they aren't tightly coupled to the previous lines and don't require saving context of what previously happened. Thanks for pointing this out :)

odeke-em:
@melekes I've made them table tests, PTAL.

types/params.go
melekes: I agree, but how about the following: what if somebody else starts using _100MiB and later you decide to change it to _120MiB. you accidentally change both usages and break the code (or worse).

types/params.go
melekes: > I had to go through the guts of it in order to write the tests themselves.

well, you are a developer, aren't you? you have to go through the guts sometimes

types/params.go
ebuchman: I think having to look it up is fine. We shouldn't ruin the variable name :P 

We do need to document this somewhere though.

types/params.go
ebuchman: Perhaps a better alternative is to include these conditions in the comments on the Params struct itself?

types/params.go
```diff
@@ -79,9 +83,9 @@ func (params *ConsensusParams) Validate() error {
 	}
 
 	// ensure blocks aren't too big
-	if params.BlockSizeParams.MaxBytes > maxBlockSizeBytes {
+	if params.BlockSizeParams.MaxBytes > _100MiB {
```
ebuchman: this isn't good because if we change the limit we have to change the variable name. it really should just be maxBlockSizeBytes

types/params.go
melekes: Could you provide a similar example from Golang source code?

ebuchman: > well, you are a developer, aren't you? you have to go through the guts sometimes
  
LOL I don't think this is fair in this case - these conditions are undocumented, so from the perspective of being able to test the functionality exposed to the user, it's fair to say this was to0 opaque - but for that reason we should put it on the Params struct rather than here on the Validate function

odeke-em: `well, you are a developer, aren't you? you have to go through the guts sometimes`
I believe that the purpose of us making APIs is so make it easy for other people to build off our stuff easily. Imagine you had to read through all the source code of Go's APIs just to be able to use a function? My motivation here it is to try to improve developer experience/productivity and reduce the cognitive overload.

`Could you provide a similar example from Golang source code?`
I can find tonnes of examples:
- https://github.com/golang/go/blob/a6ae01a64a6a5ef17a4825cd486aa31f62481100/src/net/http/request.go#L1146-L1147
- https://github.com/golang/go/blob/a9afa4e933f3eff131f12e24bb0f5b9f3168ca14/src/os/dir.go#L7-L46
https://github.com/golang/go/blob/a9afa4e933f3eff131f12e24bb0f5b9f3168ca14/src/os/env.go#L46 
- https://github.com/golang/go/blob/a9afa4e933f3eff131f12e24bb0f5b9f3168ca14/src/os/exec/exec.go#L127-L139
- https://github.com/golang/go/blob/a6ae01a64a6a5ef17a4825cd486aa31f62481100/src/runtime/mksizeclasses.go#L171-L174
- https://github.com/golang/go/blob/a6ae01a64a6a5ef17a4825cd486aa31f62481100/src/io/io.go#L286-L294
- https://github.com/golang/go/blob/a6ae01a64a6a5ef17a4825cd486aa31f62481100/src/io/io.go#L296-L302
- https://github.com/golang/go/blob/a6ae01a64a6a5ef17a4825cd486aa31f62481100/src/io/io.go#L330-L336
- https://github.com/golang/go/blob/a6ae01a64a6a5ef17a4825cd486aa31f62481100/src/bytes/bytes.go#L115-L119
- https://github.com/golang/go/blob/a6ae01a64a6a5ef17a4825cd486aa31f62481100/src/math/big/rat.go#L463-L468

and many more.

@ebuchman `Perhaps a better alternative is to include these conditions in the comments on the Params struct itself?`
Perhaps, but I think it is very easy for code to go stale that way as it is detached from the actual validation code.

types/params.go
odeke-em: Great point here, thanks.

odeke-em:
`By the way, did you see this: https://github.com/tendermint/tendermint/blob/master/types/genesis_test.go#L72 ?
`
No I hadn't, perhaps we could update the test name, I grepped for `Validate` and no hits in the tests, otherwise I have to go through each file checking

`Is there much new here?`
Just the documentation, plus the actual new test cases testing the extreme ends of the ranges as opposed to `0, 1`, but these can be adapted into that test. Thank you for pointing out that duplicate.

types/params.go
melekes: oh, wow. sorry for being so harsh. admit that I was wrong. great examples! Agree that good API and dev experience/productivity are very important.

odeke-em: Not a biggie @melekes :)

ebuchman:
> perhaps we could update the test name

Yes, or just move that test to params_test.go, which is prob more appropriate. Thanks

types/params.go
odeke-em: Aye aye, I'll revert the variable.

odeke-em:
Alright, thanks for the initial review folks. I've updated the PR with your feedback, PTAL at https://github.com/tendermint/tendermint/pull/748/commits/c6e5fb7f696c45c672b5b5a2fa3a304be9ed45c6.

PR #749, opened 2017-10-15
types: start *Events tests

odeke-em:

Removed the seemingly now defunct EventData*.Assert* methods

Make a sequence logger to fire events then check their
outputs ensuring that they fire in sequence and pump
out data with the respective wanted event name strings.

### DIFF
--- a/types/events.go
+++ b/types/events.go
@@ -140,14 +140,6 @@ type EventDataVote struct {
 	Vote *Vote
 }
 
-func (_ EventDataNewBlock) AssertIsTMEventData()       {}
-func (_ EventDataNewBlockHeader) AssertIsTMEventData() {}
-func (_ EventDataTx) AssertIsTMEventData()             {}
-func (_ EventDataRoundState) AssertIsTMEventData()     {}
-func (_ EventDataVote) AssertIsTMEventData()           {}
-
-func (_ EventDataProposalHeartbeat) AssertIsTMEventData() {}
-
 //----------------------------------------
 // Wrappers for type safety
 
--- a/types/events_test.go
+++ b/types/events_test.go
@@ -1,0 +1,80 @@
+package types_test
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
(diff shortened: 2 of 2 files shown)


types/events.go
```diff
@@ -140,14 +140,6 @@ type EventDataVote struct {
 	Vote *Vote
 }
 
-func (_ EventDataNewBlock) AssertIsTMEventData()       {}
```
ebuchman: nice

ebuchman:
I applaud the effort here but the test feels to fragile. Can we reduce the dependence on hardcoded strings ?

odeke-em:
@ebuchman could you please help with a suggestion to reduce the dependency on strings? Or how would you test FireEvent?

ebuchman:
Can we not use `EventStringXxx()` and json marshalling?

odeke-em:
`Can we not use EventStringXxx() and json marshalling?`
@ebuchman that's exactly what I did for the test https://github.com/tendermint/tendermint/pull/749/files#diff-32c6067f84ad3192c426e89a0eb16cfcR23
screen shot 2017-10-18 at 11 31 15 pm

and those strings are the expected outputs

ebuchman:
Yeh I see, we're mocking the fireable with that. It just feels a bit excessive and difficult to maintain. At the least the strings should be paired with the structs in one list. But why not use the tmlibs/events.EventSwitch and check the outputs from there against the json.Marshals so we don't have to write strings manually?

odeke-em:
@ebuchman thanks for the feedback. Alright, I see, I'll make another update but if I haven't simplified it more, could you please help me take over it? This way we can spend less back and forth on reviews and comments and we can have more throughput in coverage.

ebuchman:
Sure thing - have just been out of pocket this week from travelling. Back on it later today.

Thanks.

melekes:
Once #568 is merged, these tests probably will lose their value

ebuchman:
Oh of course. Let's reconsider them after then.

PR #750, opened 2017-10-16
Cleanup of code and code docs

adrianbrink:
This cleans up some of the code in the state package

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,6 @@ BREAKING CHANGES:
 - Better support for injecting randomness
 - Pass evidence/voteInfo through ABCI
 - Upgrade consensus for more real-time use of evidence
-- func(s *State) LoadABCIResponses() *ABCIResponses now returns nil if now ABCIResponses could be loaded from the database.
-  Previously it would return a pointer to an ABCIResponses object if it didn't find anything in the database.
 
 FEATURES:
 - Peer reputation management
@@ -29,6 +27,8 @@ IMPROVEMENTS:
 BUG FIXES:
 - Graceful handling/recovery for apps that have non-determinism or fail to halt
 - Graceful handling/recovery for violations of safety, or liveness
+
+## 0.12.0 (TBA)
 
 ## 0.11.1 (October 10, 2017)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ BREAKING CHANGES:
 - Better support for injecting randomness
 - Pass evidence/voteInfo through ABCI
 - Upgrade consensus for more real-time use of evidence
+- func(s *State) LoadABCIResponses() *ABCIResponses now returns nil if now ABCIResponses could be loaded from the database.
+  Previously it would return a pointer to an ABCIResponses object if it didn't find anything in the database.
 
 FEATURES:
 - Peer reputation management
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -164,7 +164,7 @@ func (mem *Mempool) TxsFrontWait() *clist.CElement {
 // It blocks if we're waiting on Update() or Reap().
 // cb: A callback from the CheckTx command.
 //     It gets called from another goroutine.
-// NOTE: Either cb will get called, or err returned.
+// CONTRACT: Either cb will get called, or err returned.
 func (mem *Mempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
 	mem.proxyMtx.Lock()
 	defer mem.proxyMtx.Unlock()
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -50,9 +50,10 @@ TODO: Better handle abci client errors. (make it automatically handle connection
 
 const cacheSize = 100000
 
-// Mempool is an ordered in-memory pool for transactions before they are proposed in a consensus round.
-// Transaction validity is checked using the CheckTx abci message before the transaction is added to the pool.
-// The Mempool uses a concurrent list structure for storing transactions that can be efficiently accessed by multiple concurrent readers.
+// Mempool is an ordered in-memory pool for transactions before they are proposed in a consensus
+// round. Transaction validity is checked using the CheckTx abci message before the transaction is
+// added to the pool. The Mempool uses a concurrent list structure for storing transactions that
+// can be efficiently accessed by multiple concurrent readers.
 type Mempool struct {
 	config *cfg.MempoolConfig
 
@@ -78,6 +79,7 @@ type Mempool struct {
 }
 
 // NewMempool returns a new Mempool with the given configuration and connection to an application.
+// TODO: Extract logger into arguments.
 func NewMempool(config *cfg.MempoolConfig, proxyAppConn proxy.AppConnMempool, height int) *Mempool {
 	mempool := &Mempool{
 		config:        config,
@@ -162,7 +164,7 @@ func (mem *Mempool) TxsFrontWait() *clist.CElement {
 // It blocks if we're waiting on Update() or Reap().
 // cb: A callback from the CheckTx command.
 //     It gets called from another goroutine.
-// CONTRACT: Either cb will get called, or err returned.
+// NOTE: Either cb will get called, or err returned.
 func (mem *Mempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
 	mem.proxyMtx.Lock()
 	defer mem.proxyMtx.Unlock()
@@ -287,9 +289,7 @@ func (mem *Mempool) notifyTxsAvailable() {
 	if mem.Size() == 0 {
 		panic("notified txs available but mempool is empty!")
 	}
-	if mem.txsAvailable != nil &&
-		!mem.notifiedTxsAvailable {
-
+	if mem.txsAvailable != nil && !mem.notifiedTxsAvailable {
 		mem.notifiedTxsAvailable = true
 		mem.txsAvailable <- mem.height + 1
 	}
--- a/rpc/lib/client/http_client.go
+++ b/rpc/lib/client/http_client.go
@@ -42,7 +42,8 @@ func makeHTTPDialer(remoteAddr string) (string, func(string, string) (net.Conn, 
 		protocol = "tcp"
 	}
 
-	trimmedAddress := strings.Replace(address, "/", ".", -1) // replace / with . for http requests (dummy domain)
+	// replace / with . for http requests (dummy domain)
+	trimmedAddress := strings.Replace(address, "/", ".", -1)
 	return trimmedAddress, func(proto, addr string) (net.Conn, error) {
 		return net.Dial(protocol, address)
 	}
@@ -67,7 +68,7 @@ type JSONRPCClient struct {
 	client  *http.Client
 }
 
-// NewJSONRPCClient takes an address and returns a pointer to an instance of JSONRPCClient
+// NewJSONRPCClient returns a JSONRPCClient pointed at the given address.
 func NewJSONRPCClient(remote string) *JSONRPCClient {
 	address, client := makeHTTPClient(remote)
 	return &JSONRPCClient{
--- a/rpc/lib/client/http_client.go
+++ b/rpc/lib/client/http_client.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+
 	types "github.com/tendermint/tendermint/rpc/lib/types"
 )
 
@@ -60,12 +61,13 @@ func makeHTTPClient(remoteAddr string) (string, *http.Client) {
 
 //------------------------------------------------------------------------------------
 
-// JSON rpc takes params as a slice
+// JSONRPCClient takes params as a slice
 type JSONRPCClient struct {
 	address string
 	client  *http.Client
 }
 
+// NewJSONRPCClient takes an address and returns a pointer to an instance of JSONRPCClient
 func NewJSONRPCClient(remote string) *JSONRPCClient {
 	address, client := makeHTTPClient(remote)
 	return &JSONRPCClient{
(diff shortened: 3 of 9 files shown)


state/state.go
```diff
@@ -8,11 +8,13 @@ import (
 	"time"
 
 	abci "github.com/tendermint/abci/types"
+
```
odeke-em: Spurious newlines :)

state/state.go
```diff
@@ -8,11 +8,13 @@ import (
...
 	cmn "github.com/tendermint/tmlibs/common"
 	dbm "github.com/tendermint/tmlibs/db"
 	"github.com/tendermint/tmlibs/log"
 
 	wire "github.com/tendermint/go-wire"
+
```
odeke-em: Ditto with removing this newline.

state/state.go
```diff
@@ -87,20 +89,21 @@ func LoadState(db dbm.DB) *State {
 }
 
 func loadState(db dbm.DB, key []byte) *State {
-	s := &State{db: db, TxIndexer: &null.TxIndex{}}
 	buf := db.Get(key)
 	if len(buf) == 0 {
 		return nil
-	} else {
```
odeke-em: Nice, gutting out the unnecessary state instantiation and returning early

state/state.go
```diff
@@ -142,38 +148,43 @@ func (s *State) SaveABCIResponses(abciResponses *ABCIResponses) {
...
-	if len(buf) != 0 {
-		r, n, err := bytes.NewReader(buf), new(int), new(error)
-		wire.ReadBinaryPtr(abciResponses, r, 0, n, err)
-		if *err != nil {
-			// DATA HAS BEEN CORRUPTED OR THE SPEC HAS CHANGED
-			cmn.Exit(cmn.Fmt("LoadABCIResponses: Data has been corrupted or its spec has changed: %v\n", *err))
-		}
-		// TODO: ensure that buf is completely read.
+	if len(buf) == 0 {
+		return nil
```
odeke-em: Unfortunately this might be an unintended change because, previously in the case that `len(buf) == 0` we always reeturned `new(ABCIResponses)` now this change returns `nil`. Hopefully tests can catch this behavior.

state/state.go
```diff
@@ -187,9 +198,11 @@ func (s *State) loadValidators(height int) *ValidatorsInfo {
 	wire.ReadBinaryPtr(v, r, 0, n, err)
 	if *err != nil {
 		// DATA HAS BEEN CORRUPTED OR THE SPEC HAS CHANGED
-		cmn.Exit(cmn.Fmt("LoadValidators: Data has been corrupted or its spec has changed: %v\n", *err))
+		cmn.Exit(cmn.Fmt(`LoadValidators: Data has been corrupted or its spec has changed:
```
odeke-em: Perhaps this is a little nitty about " vs `, but that's alright :)

state/state.go
```diff
@@ -219,8 +232,10 @@ func (s *State) Bytes() []byte {
 	return wire.BinaryBytes(s)
 }
 
-// SetBlockAndValidators mutates State variables to update block and validators after running EndBlock.
-func (s *State) SetBlockAndValidators(header *types.Header, blockPartsHeader types.PartSetHeader, abciResponses *ABCIResponses) {
+// SetBlockAndValidators mutates State variables to update block and validators after running
```
odeke-em: Perhaps we could just break this into 2 almost evenly spread comments, to look even nicer?
```go
// SetBlockAndValidators mutates State variables
// to update block and validators after running EndBlock.
```

state/state.go
```diff
@@ -260,12 +274,11 @@ func (s *State) setBlockAndValidators(
 }
 
 // GetValidators returns the last and current validator sets.
-func (s *State) GetValidators() (*types.ValidatorSet, *types.ValidatorSet) {
+func (s *State) GetValidators() (last *types.ValidatorSet, current *types.ValidatorSet) {
```
odeke-em: Me likey, nice! Makes it self-documenting just looking at the signature.

state/txindex/indexer.go
```diff
@@ -6,28 +6,27 @@ import (
...
-	//
-	// NOTE We do not specify Index method for analyzing a single transaction
+	// AddBatch analyzes, indexes or stores a batch of transactions.
+	// NOTE: We do not specify Index method for analyzing a single transaction
 	// here because it bears heavy perfomance loses. Almost all advanced indexers
 	// support batching.
 	AddBatch(b *Batch) error
 
-	// Tx returns specified transaction or nil if the transaction is not indexed
+	// Get return the transaction specified by hash or nil if the transaction is not indexed
```
odeke-em: Nice catch, thanks!

state/txindex/null/null.go
```diff
@@ -10,12 +10,12 @@ import (
 // TxIndex acts as a /dev/null.
 type TxIndex struct{}
 
-// Tx panics.
+// Get with a hash panics.
```
odeke-em: Perhaps we can explain this more. How about?
```go
// Get on a TxIndex is disabled and panics when invoked
```

state/txindex/null/null.go
```diff
@@ -10,12 +10,12 @@ import (
...
 func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
 	return nil, errors.New(`Indexing is disabled (set 'tx_index = "kv"' in config)`)
 }
 
-// Batch returns nil.
+// AddBatch returns nil.
```
odeke-em: Perhaps:
```go
// AddBatch is a noop and always returns nil.
```

types/block.go
```diff
@@ -14,15 +14,15 @@ import (
...
+// Block defines the atomic unit of a Tendermint blockchain.
 type Block struct {
 	*Header    `json:"header"`
 	*Data      `json:"data"`
 	LastCommit *Commit `json:"last_commit"`
 }
 
-// MakeBlock returns a new block and corresponding part set from the given information
-// TODO: version
+// MakeBlock returns a new block and corresponding part set from the given information.
```
odeke-em: s/part set/partSet/

types/block.go
```diff
@@ -14,15 +14,15 @@ import (
...
+// TODO: Add version information to the Block struct.
```
odeke-em: I think @zramsay got us covered with this in https://github.com/tendermint/tendermint/pull/736 and in particular https://github.com/tendermint/tendermint/pull/736/files#diff-fd289e47592d409909487becb9d38925R38.

rpc/lib/client/http_client.go
```diff
@@ -60,12 +61,13 @@ func makeHTTPClient(remoteAddr string) (string, *http.Client) {
...
 //------------------------------------------------------------------------------------
 
-// JSON rpc takes params as a slice
+// JSONRPCClient takes params as a slice
 type JSONRPCClient struct {
 	address string
 	client  *http.Client
 }
 
+// NewJSONRPCClient takes an address and returns a pointer to an instance of JSONRPCClient
```
odeke-em: I think this comment is unnecessary as it doesn't document anything new, what do you think?
Or perhaps
```go
// NewJSONRPCClient creates a JSONRPCClient after making an
// underlying HTTPClient connected to the address specified by remote
```

mempool/mempool.go
```diff
@@ -162,7 +164,7 @@ func (mem *Mempool) TxsFrontWait() *clist.CElement {
 // It blocks if we're waiting on Update() or Reap().
 // cb: A callback from the CheckTx command.
 //     It gets called from another goroutine.
-// CONTRACT: Either cb will get called, or err returned.
+// NOTE: Either cb will get called, or err returned.
```
ebuchman: why this changed? things which can't be captured in the function signature should be made explicit in a `CONTRACT` comment. We should add that to the coding repo @zramsay

rpc/lib/client/http_client.go
ebuchman: `NewJSONRPCClient returns a JSONRPCClient pointed at the given address` is cool.

The go linter picks up on exposed functions without comment - sometimes there's not much to say, so it's ok if the comment doesn't give us too much info.

state/state.go
ebuchman: we use the new lines to separate blocks of imports. i like them :)

state/state.go
ebuchman: see above

state/state.go
```diff
@@ -219,8 +232,10 @@ func (s *State) Bytes() []byte {
 	return wire.BinaryBytes(s)
 }
 
-// SetBlockAndValidators mutates State variables to update block and validators after running EndBlock.
-func (s *State) SetBlockAndValidators(header *types.Header, blockPartsHeader types.PartSetHeader, abciResponses *ABCIResponses) {
+// SetBlockAndValidators mutates State variables to update block and validators after running
+// EndBlock.
+func (s *State) SetBlockAndValidators(header *types.Header, blockPartsHeader types.PartSetHeader,
```
ebuchman: are these changes all being made automatically? what's the setting? we should make a firm decision on this so we don't have endless PRs making small formatting changes

state/state_test.go
```diff
@@ -30,35 +32,45 @@ func setupTestCase(t *testing.T) (func(t *testing.T), dbm.DB, *State) {
...
 	tearDown, _, state := setupTestCase(t)
 	defer tearDown(t)
+	// nolint: vetshadow
 	assert := assert.New(t)
 
 	stateCopy := state.Copy()
 
 	assert.True(state.Equals(stateCopy),
-		cmn.Fmt("expected state and its copy to be identical. got %v\n expected %v\n", stateCopy, state))
+		cmn.Fmt(`expected state and its copy to be identical. got %v\n expected %v\n`,
```
ebuchman: these newlines feel unecesary ...

state/txindex/indexer.go
```diff
@@ -6,28 +6,27 @@ import (
 	"github.com/tendermint/tendermint/types"
 )
 
-// Indexer interface defines methods to index and search transactions.
+// TxIndexer interface defines methods to index and search transactions.
+// TODO: Add the ability to create an independent TxIndexer.
```
ebuchman: what's this TODO about ?

state/txindex/indexer.go
```diff
@@ -6,28 +6,27 @@ import (
...
 //----------------------------------------------------
 // Txs are written as a batch
 
-// A Batch groups together multiple Index operations you would like performed
-// at the same time. The Batch structure is NOT thread-safe. You should only
-// perform operations on a batch from a single thread at a time. Once batch
-// execution has started, you may not modify it.
+// Batch groups together multiple Index operations you would like performed
+// at the same time.
+// NOTE: Bach is NOT thread-safe and should not be modified after starting its execution.
```
ebuchman: Bach -> Batch

ebuchman: should -> must.

Also lets try to avoid use of `you` in comments. Should be third person

state/state.go
adrianbrink: I'd like to use " instead of `, but that means that I can't line wrap it after 99 characters. :-)

" doesn't support line breaks and I'm getting more and more used to split screens and multiple windows and having line breaks at 100 is great.

mempool/mempool.go
adrianbrink: Oh, I thought we put those under `NOTE:`. The nice thing with `NOTE`, `TODO` or `XXX` is that emacs highlights those.

ebuchman: yeh the highlighting is nice. we can discuss this. but NOTE feels too informal

state/state.go
adrianbrink: Those new lines were fully intended :-) I like to split all imports into blocks so that `github.com/tendermint/tendermint` and `github.com/tendermint/abci` would be in their own block.

state/state.go
adrianbrink: It does, but currently at least nothing within Tendermint relies on this. However now it follows the semantics of `LoadState` in that if there is nothing it will return `nil`. I think we should change it now and broken clients hopefully catch it, because it seems strange to have different semantics in two cases which are very similar.

I've made a note in the changelog.

types/block.go
adrianbrink: Yes, but depending on which one gets merged first let's keep it here. If we remove it here and merge it and then the second one doesn't get merged we would lose the todo.

state/state.go
adrianbrink: I'm following the go convention of lines less than 100. The gometalinter has the linter `lll` which picks this up. I don't think we have to change it all at once, but for all new code we should aim to have a width of less than 100.

It follows the go convention and for a good reason. Working on split screen is much nicer with line breaks at 100.

@zramsay Could we add this to the coding repo?

state/state_test.go
adrianbrink: It depends on whether we want to stick to the 100 line length limit.

state/txindex/indexer.go
adrianbrink: In `func (s *State) Copy() *State` we copy everything except the underlying TxIndexer. It introduces a special case into state, where you can get a completely independent state except for the TxIndexer.

Is it okay if the TxIndexer is never copied and always points to the same underlying one or should we add `Copy()` to the TxIndexer interface to allow for independent ones?

mempool/mempool.go
adrianbrink: Actually, let's stick with `CONTRACT` for now. I'll edit my emacs to also highlight that.

adrianbrink:
@ebuchman I changed NOTE back to CONTRACT which is the last requested change.

state/txindex/indexer.go
ebuchman: I think its ok

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ BREAKING CHANGES:
 - Better support for injecting randomness
 - Pass evidence/voteInfo through ABCI
 - Upgrade consensus for more real-time use of evidence
+- func(s *State) LoadABCIResponses() *ABCIResponses now returns nil if now ABCIResponses could be loaded from the database.
```
ebuchman: this was placed under roadmap. needs to be under next version

ebuchman:
I'll rebase and merge - thanks

Issue #753, opened 2017-10-18
notified txs available but mempool empty


ebuchman:
`E[10-18|00:24:00.478] CONSENSUS FAILURE!!!                         module=consensus err="notified txs available but mempool is empty!`

This happened after sending a tx to a node via `_commit` in fast sync. The tx timed out before it finished syncing. Then once it synced up this error occured.

melekes:
I have not yet reproduced it, but my first guess would be that we're calling 

https://github.com/tendermint/tendermint/blob/develop/mempool/mempool.go#L272

after rechecking txs, while here https://github.com/tendermint/tendermint/blob/develop/mempool/mempool.go#L256 we may be removed the single tx, which left in the mempool => that will lead to `size==0`

PR #755, opened 2017-10-18
WIP: only notify when there are some txs (Refs #753)

melekes:
See #753


melekes:
Have not been able to write a good test for this so far.

ebuchman:
Not thrilled but looks right, don't think we can do much better.

To test we need to make a custom app that can invalidate a tx. Dummy/Counter don't currently allow that. So ie. tx1 is valid in mempool, then tx2 is committed, that makes tx1 invalid, so it will get chucked from the mempool.

PR #760, opened 2017-10-19
change the way to stop making empty blocks.

kidinamoto01:
hey,
I  looked into the problem of enabling no empty block /w Ethermint. EM's AppHash will always change. 
I change the way to stop TM making empty blocks calls by also checking mempool.

Cheers


ebuchman:
Unfortunately we can't do this. We need to make empty blocks sometimes to get commits on the apphashes that result from txs. The problem with ethermint is the app hash updates with every block, even without txs, so it will never stop. We need to make a change in ethermint to fix this.

See https://github.com/tendermint/ethermint/issues/343

Issue #476, opened 2017-05-02
Allow side-channel P2P messaging between nodes


srmo:
We would like to have something like a side-channel for arbitrary messaging between nodes. Such messages should not influence consensus/mempool. One solution is to write a custom p2p layer on top of our ABCI app(s) but why not use the existing p2p network established by the nodes?
We could have a subscribe endpoint (like provided for WSRPC)
Could you implement such a thing? If yes, how difficult would that be?

something like `/broadcast_sidechannel?msg=“Hi There!”`

adrianbrink:
@srmo What is the use case for this feature?

srmo:
@adrianbrink 
We require some kind of toplevel protocol to determine if certain blockchain-persisted transactions should be displayed to clients (client applications). This protocol requires a fully anonymized broadcast exchange of multiple message types. Using the blockchain to run the protocol wastes time and storage. It seems obvious that this should be handled off-chain. The current approach is writing a second P2P between all the chain clients. But why write a new one when we could use the already existing p2p transport layer.

Does that make it clearer or do you need more information?

adrianbrink:
@srmo I don't think that we will implement this in tendermint, as we are focusing in getting tendermint ready for COSMOS. This seems to be a very specific use case that isn't needed for COSMOS and hence I will be closing this. 

@ebuchman Maybe it could be an idea to discuss how to enable users of tendermint to write their own extensions.

srmo:
sadface. @adrianbrink I'm looking forward to a COSMOS road map. We are using tendermint for non-cosmos stuff so we are very interested in seeing what can be expected in terms of new functionality.

ebuchman:
I wouldn't close this so quickly :)

rigelrozanski:
Interesting feature. I bet there is a use case for this communication layer in implementing something like proxy nodes for validators.

srmo:
Any new thoughts on this?

paralin:
If libp2p was in use, then the "side-channel" would just be libp2p pub-sub / encrypted channels between peers.

ebuchman:
It's easy enough to add another Reactor for the tendermint p2p, as was done here: https://github.com/tendermint/tendermint/pull/761

We're interested in this, but it's not priority for launch and we don't really have the time to incorporate it yet. But perhaps we can merge it soon and keep it behind a feature flag so folks can optionally enable if they want for their private networks.

Ivshti:
[...]

PR #761, opened 2017-10-20
#476 P2P Side-Channel Broadcast

wolfposd:
Related to https://github.com/tendermint/tendermint/issues/476

- Created a low priority reactor "TransientBroadcast"
- Take TransientTx from RPC/WSRPC
- Broadcast to Peers
- Re-Broadcast to own subscribed RPC-Clients via EventSwitch


p2p/transient_broadcast_reactor.go
```diff
@@ -0,0 +1,127 @@
+package p2p
+
+import (
+	"bytes"
+	"fmt"
+	"github.com/tendermint/go-wire"
+	"github.com/tendermint/tendermint/types"
+	"reflect"
```
zramsay: should be below `"fmt"` then line break before the tendermint imports

wolfposd:
this is still not complete yet, might have issued this too early.

- gossiping needs to be adressed
- not all peers that are connect receive a message, peer-list=1 even with two connected nodes

adrianbrink:
@wolfposd Is this something that might behave similar to Whisper for Ethereum?

wolfposd:
Yes, Whisper is describing our use-cases for this feature pretty good

> DApps that need to signal to each other in order to ultimately collaborate on a transaction. For example, a currency exchange DApp may use it to coordinate an offer prior to creating one (or two, depending on how the exchange is structured) transactions on the exchange.

wolfposd:
requires https://github.com/tendermint/tmlibs/pull/63/commits

wolfposd:
updated to newest develop, with eventbus

zramsay:
rebase? do we still want this?

ebuchman:
Really appreciate this PR, but unfortunately it would be too much of a distraction to try to get this merged in the short term while we're working towards cosmos launch. we'd need some discussion around design, perhaps ideally by starting with a reactor spec in the same vein as the others under https://github.com/tendermint/tendermint/tree/develop/docs/specification/new-spec/reactors. Of course, it would also need some tests.

I'll close it for now, and hopefully we can revisit later in the year.

Issue #762, opened 2017-10-20
`BlockchainReactor Timeout` error while syncing


melekes:
```
E[10-19|04:55:31.278] Stopping peer for error                      module=p2p peer="Peer{MConn{146.185.173.132:46656} 55AF537EBCFA out}" err="BlockchainReactor Timeout"
E[10-19|04:55:32.408] Failed to add peer                           module=p2p address=104.131.86.166:46656 err="Duplicate peer"
E[10-19|04:55:33.198] Failed to add peer                           module=p2p address=178.62.46.14:46656 err="Duplicate peer"
E[10-19|04:55:35.656] Failed to add peer                           module=p2p address=104.131.86.166:46656 err="Duplicate peer"
E[10-19|04:55:36.735] Failed to add peer                           module=p2p address=178.62.46.14:46656 err="Duplicate peer"
E[10-19|04:55:38.334] Stopping peer for error                      module=p2p peer="Peer{MConn{178.62.46.14:46656} B14916FAF38A out}" err="BlockchainReactor Timeout"
E[10-19|04:55:38.904] Failed to add peer                           module=p2p address=104.131.86.166:46656 err="Duplicate peer"
E[10-19|04:55:39.364] Stopping peer for error                      module=p2p peer="Peer{MConn{104.131.86.166:46656} 7A401C816E0C out}" err="BlockchainReactor Timeout"
E[10-19|04:55:40.276] Failed to add peer                           module=p2p address=178.62.46.14:46656 err="Duplicate peer"
E[10-19|04:55:41.723] Stopping peer for error                      module=p2p peer="Peer{MConn{146.185.173.132:46656} 55AF537EBCFA out}" err="BlockchainReactor Timeout"
E[10-19|04:55:42.022] Stopping peer for error                      module=p2p peer="Peer{MConn{104.131.86.166:46656} 7A401C816E0C out}" err="BlockchainReactor Timeout"
E[10-19|04:55:42.150] Stopping peer for error                      module=p2p peer="Peer{MConn{178.62.46.14:46656} B14916FAF38A out}" err="BlockchainReactor Timeout"
E[10-19|04:55:42.274] Failed to add peer                           module=p2p address=104.131.86.166:46656 err="Duplicate peer"
E[10-19|04:55:43.812] Failed to add peer                           module=p2p address=178.62.46.14:46656 err="Duplicate peer"
[...]

PR #765, opened 2017-10-20
blockchain reactor timeout

melekes:
See #762

### DIFF
--- a/blockchain/pool.go
+++ b/blockchain/pool.go
@@ -69,9 +69,7 @@ func (pool *BlockPool) OnStart() error {
 	return nil
 }
 
-func (pool *BlockPool) OnStop() {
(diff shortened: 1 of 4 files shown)


melekes:
@odeke-em mm, I am not a fan of this approach (I'm referring to `if peer == nil`) and would prefer to avoid it. Such things should be dealt with on a higher level using https://en.wikipedia.org/wiki/Null_object_pattern or something else

ebuchman:
In this case I think we can have the nil checks where the func is called, since we're doing that all over the package. But in general the structure/clarity/safety of the package could probably be improved a lot.

ebuchman:
Looks like this PR introduced a race condition in the tests - @melekes can you look into this please?

PR #767, opened 2017-10-22
Make RPCError an actual error and don't swallow its companion data

silasdavis:
Currently the RPC clients swallow useful nested error data, this PR surfaces that and makes RPCError implement the Go error interface.


ebuchman:
Thanks Silas. This should fix https://github.com/cosmos/gaia/issues/23 :)

PR #768, opened 2017-10-23
Feature/merkleeyes to iavl

ebuchman:
replaces https://github.com/tendermint/tendermint/pull/737


glide.yaml
```diff
@@ -25,12 +25,12 @@ import:
...
   - data
-- package: github.com/tendermint/merkleeyes
-  version: master
+- package: github.com/tendermint/iavl
+  version: develop
   subpackages:
   - iavl
 - package: github.com/tendermint/tmlibs
-  version: ~0.3.2
+  version: develop
```
zramsay: won't these changes make `go get` easily breakable? also contradicts: https://github.com/tendermint/coding/issues/24

melekes: only on develop, which is fine. we will change these back to tags when we'll be doing a release

zramsay: ah gotcha

ebuchman: exactly. we can include that in the guidelines

PR #774, opened 2017-10-24
WIP: Add goreleaser

adrianbrink:


adrianbrink:
@zramsay @Greg-Szabo

zramsay:
this would deprecate part of the release script introduced by @melekes a month or so ago, so this PR should be updated to sync with that script

adrianbrink:
* look into gitian.org
* builds need to be fully reproducible

melekes:
https://github.com/goreleaser/goreleaser/blob/master/pipeline/build/build.go#L89-L102

goreleaser uses `go build`, so it is not hermetic build per se. But I guess it should be fine for other small projects like gaia.

ebuchman:
Let's integrate this into the release script

adrianbrink:
Closed based on conversation at the retreat about build and release process.

Issue #751, opened 2017-10-17
rpc/lib/client: introduce jitter for ExponentialBackoff to avoid potential self DDOS vector


odeke-em:
So in lines https://github.com/tendermint/tendermint/blob/d4634dc6832a7168c2536e7c71bfba4a8ca857be/rpc/lib/client/ws_client.go#L260-L261

we perform exponential backoff by the standard definition which is 1<<n seconds, every client that attempts to reconnect with retry in at least 1<<n seconds.

tl;dr: Let's add randomness to our system in order to avoid potential self DDOS-ing. It might sound far fetched but I've personally dealt with this. A more trustable and prominent web services company AWS also has recognized such problems and added some advice here https://www.awsarchitectureblog.com/2015/03/backoff.html and if I recall right, also historically the ALOHA networking in the 1970s recognized these problems and incorporated randomness into their system.

### Long Wind
The current code is quite deterministic in that if the WSServer were overloaded past capacity, 
and had say X clients all reconnected at the same time, if all the X clients reattempted at the exact same time, this will be an unintentional DDOS.

I mention this because last year, I built a video transcoding system for an internet company and one of the main micro-services running on a very large and beefed up machine(specialized for transcoding video to 3+ formats in parallel).
The transcoder(horizontally scaled on its own) was connected to by multiple clients and had a timeout as well as obviously limits on the hardware. Problem though is that when very many clients uploaded 4K videos to be transcoded to 3 formats in parallel, RAM would blow up and cause problems for many clients, requests would be aborted due to timeouts. But then on retry, all the clients would re-flood it at the same time, leading to a stalemate each time. The solution involved introducing jitter/randomness in the sleeps by:
```go
d := time.Duration(1e9*rand.Float64()) + ((1 << attempts) * time.Second)
time.Sleep(d)
```
and that eased the load a whole lot.

Please feel free to adapt the jitter/randomness as necessary but the main point is that not every client should attempt to reconnect at the same time.

melekes:
I like the idea. Let's do this.

ebuchman:
Cool!

odeke-em:
Fixed by PR https://github.com/tendermint/tendermint/pull/775 that just got merged into branch develop.

PR #775, opened 2017-10-24
rpc/lib/client: add jitter for exponential backoff of WSClient

odeke-em:
Fixes https://github.com/tendermint/tendermint/issues/751.

Adds jitter to our exponential backoff to mitigate a self DDOS
vector. The jitter is a randomly picked percentage of a second
whose purpose is to ensure that each exponential backoff retry
occurs within (1<<attempts) == 2**attempts, but with the delay
each client will have a random buffer time before it tries to
reconnect instead of all at once reconnections that might even
bring back the previous conditions that might have caused the
dial to the WSServer to have failed e.g
* Network outage
* File descriptor exhaustion
* False positives from firewalls
etc

### DIFF
--- a/rpc/lib/client/ws_client.go
+++ b/rpc/lib/client/ws_client.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"math"
+	"math/rand"
 	"net"
 	"net/http"
 	"sync"
@@ -254,11 +254,13 @@ func (c *WSClient) reconnect() error {
 		c.mtx.Unlock()
 	}()
 
+	_1sAsNs := float64(time.Second.Nanoseconds())
 	for {
-		c.Logger.Info("reconnecting", "attempt", attempt+1)
+		jitter := time.Duration(rand.Float64() * _1sAsNs)
+		backoffDuration := jitter + ((1 << uint(attempt)) * time.Second)
 
-		d := time.Duration(math.Exp2(float64(attempt)))
-		time.Sleep(d * time.Second)
+		c.Logger.Info("reconnecting", "attempt", attempt+1, "backoff_duration", backoffDuration)
+		time.Sleep(backoffDuration)
 
 		err := c.dial()
 		if err != nil {
--- a/rpc/lib/client/ws_client.go
+++ b/rpc/lib/client/ws_client.go
@@ -254,10 +254,11 @@ func (c *WSClient) reconnect() error {
 		c.mtx.Unlock()
 	}()
 
-	_1sAsNs := float64(time.Second.Nanoseconds())
+	// 1s == (1e9 ns) == (1 Billion ns)
+	billionNs := float64(time.Second.Nanoseconds())
 	for {
-		jitter := time.Duration(rand.Float64() * _1sAsNs)
-		backoffDuration := jitter + ((1 << uint(attempt)) * time.Second)
+		jitterSeconds := time.Duration(rand.Float64() * billionNs)
+		backoffDuration := jitterSeconds + ((1 << uint(attempt)) * time.Second)
 
 		c.Logger.Info("reconnecting", "attempt", attempt+1, "backoff_duration", backoffDuration)
 		time.Sleep(backoffDuration)
--- a/rpc/lib/client/ws_client_test.go
+++ b/rpc/lib/client/ws_client_test.go
@@ -1,11 +1,14 @@
 package rpcclient
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"regexp"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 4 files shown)


rpc/lib/client/ws_client.go
```diff
@@ -254,11 +254,13 @@ func (c *WSClient) reconnect() error {
 		c.mtx.Unlock()
 	}()
 
+	_1sAsNs := float64(time.Second.Nanoseconds())
 	for {
-		c.Logger.Info("reconnecting", "attempt", attempt+1)
+		jitter := time.Duration(rand.Float64() * _1sAsNs)
```
melekes: how big can jitter be? 1sAsNs is 1000000000, right? so given rand.Float() returned 1.0, max jitter is 1000000000 seconds?

rpc/lib/client/ws_client_test.go
```diff
@@ -191,3 +194,55 @@ func callWgDoneOnResult(t *testing.T, c *WSClient, wg *sync.WaitGroup) {
 		}
 	}
 }
+
+func TestWSClientReconnectWithJitter(t *testing.T) {
```
melekes: good that you wrote a test, but don't think we needed to test that specifically

rpc/lib/client/ws_client.go
silasdavis: maybe `oneSecondAsNanoseconds`... :0?

melekes: yeah, don't like the name too

ebuchman: `billionNS` ?

rpc/lib/client/ws_client_test.go
ebuchman: Agree that this might not be the most useful test, especially given how much code it is. We'd also have to start using `-skip`, and let's not.

That said, I like the idea of adding more advanced tests like this, but let's keep them separate from the unit testing. Perhaps we can come up with a good schema for starting to add all these longer-running, more nuanced tests, including the fuzzer stuff. We can run those tests on release, but they don't need to run every time with the unit tests.

ebuchman:
Test failed, not sure if its related to this change or not, but looks like a port is trying to be reused:

```
I[10-24|09:15:08.558] Added new address                            address=127.0.0.1:41288 total=1^M                                                      
--- FAIL: TestPEXReactorRunning (4.02s)^M                                                                                                                 
panic: Panicked on a Crisis: listen tcp 127.0.0.1:41288: bind: address already in use [recovered]^M                                                       
    panic: Panicked on a Crisis: listen tcp 127.0.0.1:41288: bind: address already in use^M                                                               
^M                                                                                                                                                        
goroutine 88 [running]:^M                                                                                                                                 
testing.tRunner.func1(0xc4201ba870)^M                                                                                                                     
    /usr/local/go/src/testing/testing.go:711 +0x5d9^M                                                                                                     
panic(0x8f48e0, 0xc42003d480)^M                                                                                                                           
    /usr/local/go/src/runtime/panic.go:491 +0x2a2^M                                                                                                       
github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common.PanicCrisis(0x91b2e0, 0xc42006a910)^M                                         
    /go/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common/errors.go:33 +0x120^M                                             
github.com/tendermint/tendermint/p2p.NewDefaultListener(0x97c276, 0x3, 0xc4200173d0, 0xf, 0x1, 0xcc8020, 0xc4201945d0, 0xc42018e400, 0xc420121768)^M      
    github.com/tendermint/tendermint/p2p/_test/_obj_test/listener.go:80 +0xc61^M                                                                          github.com/tendermint/tendermint/p2p.TestPEXReactorRunning(0xc4201ba870)^M                                                                                    /go/src/github.com/tendermint/tendermint/p2p/pex_reactor_test.go:92 +0x4f1^M                                                                          testing.tRunner(0xc4201ba870, 0xab73f8)^M                                                                                                                     /usr/local/go/src/testing/testing.go:746 +0x16d^M                 
```

rpc/lib/client/ws_client_test.go
silasdavis: You can use a build tag like 'integration' (e.g. ithttps://github.com/hyperledger/burrow/blob/master/rpc/tendermint/test/rpc_client_test.go#L1-L2 - note you need the blank line after - run as https://github.com/hyperledger/burrow/blob/master/Makefile#L129)

rpc/lib/client/ws_client.go
odeke-em: `how big can jitter be? 1sAsNs is 1000000000, right? so given rand.Float() returned 1.0, max jitter is 1000000000 seconds?`
So time.Durations units are in ns NOT seconds https://golang.org/pkg/time/#Duration, therefore the max value of `rand.Float() * 1sAsNs is 1e9ns * 1.0 --> 1s` 

Ahaha we've bikeshed on the name. @silasdavis that seems too verbose to me :), I can try to fly with @ebuchman's but I believe I was trying to convey the number of nano seconds within a second without having to whip out a calculator to convert between seconds and nanoseconds.

rpc/lib/client/ws_client_test.go
odeke-em: `good that you wrote a test, but don't think we needed to test that specifically`
`Agree that this might not be the most useful test, especially given how much code it is. We'd also have to start using -skip, and let's not.`
See the problem here is that that code can be changed at anytime and without a comprehensive test, we'd never know what changed at all. Perhaps we need some middle ground, like @ebuchman mentioned, perhaps as a release test.

Great suggestion @silasdavis, thank you.

odeke-em:
@ebuchman this test doesn't bind to any port at all, it just triggers the reconnect routine and since the network is never setup the tests are bound to run until we exceed max attempts. I think that's an unrelated test flake.

odeke-em:
Thank you @melekes @ebuchman and @silasdavis for the first round of review. PTAL, I've updated the code to address your suggestions.

PR #777, opened 2017-10-24
Fix WSClient deadlock in the readRoutine after Stop() is called

silasdavis:
In `WSClient` when `readRoutine` reads a message it blocks sending it to `ResultsCh`. When you call `Stop()` on the `WSClient` this cause it to block indefinitely. This is because `readRoutine` never calls `Done()` on the wait group in `WSClient` because it is blocking on sending to `ResultsCh` and `Stop()` calls `c.wg.Wait()`.

This fixes it by introducing a back channel `writeRoutineQuit` which works in a similar way to the existing `readRoutineQuit`. In both cases the close idiom is used so that any subsequent reads will immediately return which should eliminate any issues if both routines are trying to close each other. We need another channel (rather than using 'readRoutineQuit' for both) because only one goroutine can be responsible for closing a channel without there being a race.

### DIFF
--- a/rpc/lib/client/ws_client.go
+++ b/rpc/lib/client/ws_client.go
@@ -48,10 +48,11 @@ type WSClient struct {
 	onReconnect func()
 
 	// internal channels
-	send            chan types.RPCRequest // user requests
-	backlog         chan types.RPCRequest // stores a single user request received during a conn failure
-	reconnectAfter  chan error            // reconnect requests
-	readRoutineQuit chan struct{}         // a way for readRoutine to close writeRoutine
+	send             chan types.RPCRequest // user requests
+	backlog          chan types.RPCRequest // stores a single user request received during a conn failure
+	reconnectAfter   chan error            // reconnect requests
+	readRoutineQuit  chan struct{}         // a way for readRoutine to close writeRoutine
+	writeRoutineQuit chan struct{}         // a way for writeRoutine to close readRoutine (on <-BaseService.Quit)
 
 	wg sync.WaitGroup
 
@@ -282,6 +283,7 @@ func (c *WSClient) reconnect() error {
 func (c *WSClient) startReadWriteRoutines() {
 	c.wg.Add(2)
 	c.readRoutineQuit = make(chan struct{})
+	c.writeRoutineQuit = make(chan struct{})
 	go c.readRoutine()
 	go c.writeRoutine()
 }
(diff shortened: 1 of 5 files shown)


rpc/lib/client/ws_client.go
```diff
@@ -441,7 +446,12 @@ func (c *WSClient) readRoutine() {
 			continue
 		}
 		c.Logger.Info("got response", "resp", response.Result)
-		c.ResultsCh <- *response.Result
+		// Combine a non-blocking read on writeRoutineQuit with a non-blocking write on ResultsCh to avoid blocking
```
melekes: what about the `c.ErrorsCh`?

melekes:
Thanks! See my comment about ErrorsCh. We'll need to add the same select there. Maybe add a func, which will wrap this.

rpc/lib/client/ws_client.go
silasdavis: Actually this reminds me of an issue with `WSClient`, which might make sense to do something about now. It really would be better if it just exposed a `chan RPCResponse` rather than filtering them into `ResultsCh` and `ErrorsCh` because:

1. It forces callers to maintain and synchronise two goroutines to read a single logical stream of messages.
2. It would be easy to accomplish the same as a caller by launching a goroutine that reads from a single `chan RPCResponse`
3. (And by far the most important) currently callers have no access to the request/response correlation ID `RPCResponse.ID` that is set by the client and is meant to allow the client to map the `Responses`.

There's actually only 13 usages of `ResultsCh` in the Tendermint codebase. Maybe now would be a good time to fix this issue and just provide a `ResponseCh`. I think I could be persuaded to do the work to port it over in the Tendermint codebase, but are there other users we need to worry about a breaking change with?

melekes: can't agree more! I would be in favor of doing that. @ebuchman any objections?

silasdavis:
@melekes well it was quick enough to do so the second commit does that, if you guys want to go that way (which would suit me, since I'd like to get at ID)

ebuchman:
I like the unified responses channel, but it feels wrong to have so many quit channels. Can this not be simplified somehow to just use c.Quit?

silasdavis:
Why yes, we can...

At least my additional channel is unnecessary if we can depend on `BaseService` closing its `Quit` channel as it currently does (probably for this reason) and once closed it will emit unlimited quit signals, so unlike the case I had in mind of a single signal being sent and there being a race between `readRoutine` and `writeRoutine` to see who gets it, they can both be signalled to quit from it.

I had a look to see if we could do without the `readRoutineQuit` channel, but it is legitimately a different signal to `Quit` since it is really a 'reset' signal clearing the way for a reconnect, whereas `Quit` signals permanent irrevocable stopping of the service. We could probably piggy back on some other state to unambiguously detect the reconnect attempt in `writeRoutine` and use that to shut it down (like maybe the `reconnectAfter` channel), but I think it is plainer the way it is.

Last commit removes my extra channel. Let me know if you want a squash.

PR #779, opened 2017-10-24
Update ecosystem.rst

adrianbrink:


ebuchman:
What and why? 

Ideally this is auto-generated from https://github.com/tendermint/aib-data/blob/master/json/ecosystem.json or something

adrianbrink:
@zramsay Can we use the aib-data repo as a data source for readthedocs.io?

It's to update the docs [here](https://tendermint.readthedocs.io/projects/tools/en/master/ecosystem.html)

zramsay:
yeah it's coming. need to sync with @nylira to make this happen

PR #781, opened 2017-10-24
remove unnecessary plus

rudi-cilibrasi:
less confusing this way. same behavior because n is initialized to 0 unconditionally earlier.

### DIFF
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -271,7 +271,7 @@ func (p *peer) CanSend(chID byte) bool {
 func (p *peer) WriteTo(w io.Writer) (n int64, err error) {
 	var n_ int
 	wire.WriteString(p.key, w, &n_, &err)
-	n += int64(n_)
+	n = int64(n_)
 	return
 }
 


ebuchman:
Thanks Rudi. This is ok, but arguably the `+=` was more defensive, ie. if for some reason more things get added to `n` we don't have to worry.

In this case, I might remove the named returns entirely for extra clarity:

```
// WriteTo writes the peer's public key to w.
func (p *peer) WriteTo(w io.Writer) (int64, error) {
	var n int
	var err error
	wire.WriteString(p.key, w, &n, &err)
	return int64(n), err
}
```

Then there's no `n` to be confused about, we just need to type cast

petabytestorage:
I agree @ebuchman  and Ethan F and I figured the same also yesterday. Thanks for confirmation everybody.

ebuchman:
merged here https://github.com/tendermint/tendermint/commit/ceedd4d96852bee9e8bfb166a22fa2edd9107c67

Issue #784, opened 2017-10-25
Better integration/naming of new certifier pkg


ebuchman:
See comments in https://github.com/tendermint/tendermint/pull/776 for some ideas

ebuchman:
lite!

PR #786, opened 2017-10-26
Rename certifier to light (#784) and add godocs

adrianbrink:
The certifier package is renamed to light. This is more descriptive
especially in the wider blockchain context. Moreover we are building
light-clients using the light package.

This also adds godocs to all exported functions.

Furthermore it introduces some extra error handling. I've added one TODO
where I would like someone else's opinion on how to handle the error.


light/inquirer.go
```diff
@@ -14,8 +18,14 @@ type Inquiring struct {
...
 }
 
+// NewInquiring returns a new Inquiring object. It uses the trusted provider to store validated
+// data and the source provider to obtain missing FullCommits.
+//
+// Example: The trusted provider should a CacheProvider, MemProvider or files.Provider. The source
+// provider should be a client.HTTPProvider.
 func NewInquiring(chainID string, fc FullCommit, trusted Provider, source Provider) *Inquiring {
 	// store the data in trusted
+	// TODO: StoredCommit() can return an error and we need to handle this.
```
adrianbrink: @ethanfrey How should we handle this?

ethanfrey:
Ugh, I wish that was two commits... one with godoc and one with the name.

There was no agreement on the name, and I would like some discussion beforehand. See https://github.com/tendermint/tendermint/pull/776#issuecomment-339342861 where I explicitly argue against `light`. There was no reponse except this pr....

adrianbrink:
Discussion for changing the package name from `certifiers` to `light`:
* `light-client` is annoying to type and hence has been discarded
Cons:
* it is not very descriptive

Pros:
* codebases such as Ethereum use the `light` package to group functionality related to building light clients
* even though it isn't very descriptive on the contained types, `light` describes very well when to use this package. The `light` package is used to build light-clients.
* most developers will not associate `certifiers` with the functionality needed to build light clients

@ethanfrey Let me know what you think about these arguments. I agree with you that `certifiers` describes is very descriptive of the actual types. However I think that `light` is more descriptive of when to use the functionality in this package, namely building light clients.

light/client/provider.go
```diff
@@ -28,26 +29,26 @@ type provider struct {
 
 // NewProvider can wrap any rpcclient to expose it as
 // a read-only provider.
-func NewProvider(node SignStatusClient) certifiers.Provider {
+func NewProvider(node SignStatusClient) light.Provider {
 	return &provider{node: node}
 }
 
-// NewProvider can connects to a tendermint json-rpc endpoint
+// NewHTTPProvider can connects to a tendermint json-rpc endpoint
```
odeke-em: can connect.

light/commit.go
```diff
@@ -33,20 +33,23 @@ type FullCommit struct {
...
 
+// NewFullCommit returns a new FullCommit.
 func NewFullCommit(commit Commit, vals *types.ValidatorSet) FullCommit {
 	return FullCommit{
 		Commit:     commit,
 		Validators: vals,
 	}
 }
 
+// Height returns the of the header.
```
odeke-em: // Height returns the height of the commit's Header

ebuchman:
How about `lite` ?

adrianbrink:
@jaekwon 's suggestion is `lite`

ebuchman:
@ethanfrey permission to call this `lite` ? Before you leave and some trigger happy member of the team moves forward without you ;)

adrianbrink:
Fixed up :-) @ebuchman

lite/provider_test.go
```diff
@@ -1,37 +1,58 @@
...
+// missingProvider doens't store anything, always a miss
+// Designed as a mock for testing
+type missingProvider struct{}
+
+// NewMissingProvider returns a provider which does not store anything and always misses.
+func NewMissingProvider() lite.Provider {
+	return missingProvider{}
+}
+
+func (missingProvider) StoreCommit(_ lite.FullCommit) error { return nil }
```
melekes: I think you can drop `_ ` part

adrianbrink: Damn, you are right. I didn't know that.

PR #787, opened 2017-10-26
Initial Trust Metric Implementation

caffix:


p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,394 @@
...
+	History      []float64 `json:"history"`
+}
+
+func loadSaveFromFile(key string, isLoad bool, data *peerHistory) *peerHistory {
+	tmhome, ok := os.LookupEnv("TMHOME")
+	if !ok {
+		return nil
+	}
+
+	filename := filepath.Join(tmhome, "trust_history.json")
```
ebuchman: cool. We should better integrate this with the rest of the package and use a new config.TrustHistory field for this instead of new hard_coded values. Eg. like the AddrBook (and its config https://github.com/tendermint/tendermint/blob/master/config/config.go#L212)

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,394 @@
...
+
+	go store.processRequests()
+}
+
+type peerHistory struct {
+	NumIntervals int       `json:"intervals"`
+	History      []float64 `json:"history"`
+}
+
+func loadSaveFromFile(key string, isLoad bool, data *peerHistory) *peerHistory {
```
ebuchman: this feels weird. Can we just have too separate functions.

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,394 @@
...
+	Resp chan *TrustMetric
+}
+
+type trustMetricStore struct {
+	PeerMetrics map[string]*TrustMetric
+	Requests    chan *peerMetricRequest
+	Disconn     chan string
+}
+
+func init() {
```
ebuchman: lets try to model the trustMetricStore like the AddrBook, rather than just spawning it in the `init()`

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,394 @@
...
+)
+
+type peerMetricRequest struct {
+	Key  string
+	Resp chan *TrustMetric
+}
+
+type trustMetricStore struct {
+	PeerMetrics map[string]*TrustMetric
+	Requests    chan *peerMetricRequest
```
ebuchman: do these channels need to be public? And can we get comments on them please? Same with all the other structs and functions.

caffix:
I have attempted to make the changes requested in the PR comments.

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,558 @@
...
+	tm.integralWeight = config.IntegralWeight
+	tm.intervalLen = config.IntervalLen
+	// The maximum number of time intervals is the tracking window / interval length
+	tm.maxIntervals = int(config.TrackingWindow / tm.intervalLen)
+	// The history size will be determined by the maximum number of time intervals
+	tm.historyMaxSize = intervalToHistoryIndex(tm.maxIntervals) + 1
+	// This metric has a perfect history so far
+	tm.historyValue = 1.0
+	// Setup the channels
+	tm.update = make(chan *updateBadGood, 10)
```
melekes: magic number. better extract this to a `const defaultUpdateChanCapacity = 10)`

melekes:
Cool stuff 👍  This could probably use more tests (to check that the algorithm works)

docs/architecture/adr-006-trust-metric.md
```diff
@@ -0,0 +1,243 @@
...
+
+Instead, having shorter intervals, but keeping a history of interval values, will give our metric the flexibility needed in order to keep the network stable, while also making it resilient against a strategic malicious node in the Tendermint peer-to-peer network. Also, the metric can access trust data over a rather long period of time while not greatly increasing its history size by aggregating older history values over a larger number of intervals, and at the same time, maintain great precision for the recent intervals. This approach is referred to as fading memories, and closely resembles the way human beings remember their experiences. The trade-off to using history data is that the interval values should be preserved in-between executions of the node.
```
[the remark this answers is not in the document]
caffix: agreed. I also noticed during the simulations that, adding the fading memories functionality smoothed out the results a bit

caffix:
Thanks. I have made all the changes requested within your comments. Also, I have run simulations with the trust metric to test the algorithm, but am looking forward to using it within a testnet in order to better understand typical node behavior

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,545 @@
...
+func (tms *TrustMetricStore) OnStop() {
+	tms.mtx.Lock()
+	defer tms.mtx.Unlock()
+
+	// Stop all trust metric goroutines
+	for _, tm := range tms.peerMetrics {
+		tm.Stop()
+	}
+
+	tms.saveToDB()
```
ebuchman: Should we saveToDB more frequently so we don't lose data if there's a crash? Perhaps a separate go-routine that persists this info every minute or so?

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,545 @@
...
+// GetPeerTrustMetric returns a trust metric by peer key
+func (tms *TrustMetricStore) GetPeerTrustMetric(key string) *TrustMetric {
+	tms.mtx.Lock()
+	defer tms.mtx.Unlock()
+
+	tm, ok := tms.peerMetrics[key]
+	if !ok {
+		// If the metric is not available, we will create it
+		tm = NewMetricWithConfig(tms.config)
+		if tm != nil {
```
ebuchman: this should never return nil, so we shouldn't need to check.

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,545 @@
...
+	// Obtain the history data we have so far
+	bytes := tms.db.Get(trustMetricKey)
+	if bytes == nil {
+		return false
+	}
+
+	peers := make(map[string]peerHistoryJSON, 0)
+	err := json.Unmarshal(bytes, &peers)
+	if err != nil {
+		cmn.PanicCrisis(cmn.Fmt("Could not unmarchal Trust Metric Store DB data: %v", err))
```
ebuchman: unmarshal

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,545 @@
...
+		tm := NewMetricWithConfig(tms.config)
+
+		// Restore the number of time intervals we have previously tracked
+		if p.NumIntervals > tm.maxIntervals {
+			p.NumIntervals = tm.maxIntervals
+		}
+		tm.numIntervals = p.NumIntervals
+		// Restore the history and its current size
+		if len(p.History) > tm.historyMaxSize {
+			p.History = p.History[:tm.historyMaxSize]
```
ebuchman: so `History` is most recent first. This means we have to use `append` in a "backwards" way (ie. `append([]xxx{x}, xxx...)` vs. `append(xxx, x)`). Is there a particular reason to do it this way rather than put the most recent elements on the end? Seems like this would be less performant since if I'm not mistaken we'll be reallocating memory every time.

p2p/trust/trustmetric.go
caffix: good catch

p2p/trust/trustmetric.go
caffix: agreed. It will be in the next commit

p2p/trust/trustmetric.go
caffix: okay. History is now most recent last. I have updated the calculations to handle this properly and reduced memory allocations related to the history data. Running some tests right now

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,574 @@
...
+	return config
+}
+
+// Calculates the derivative component
+func (tm *TrustMetric) derivativeValue() float64 {
+	return tm.proportionalValue() - tm.historyValue
+}
+
+// Strengthens the derivative component when the change is negative
+func (tm *TrustMetric) weightedDerivative() float64 {
```
ebuchman: So here `gamma1 = 0` and `gamma2 = 1`, correct? Is that really what we want? And can we make sure it's documented in the proposal.

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,574 @@
...
+	for count, i := 1, first-1; count < tm.historySize; count, i = count+1, i-1 {
+		// The older the data is, the more we spread it out
+		x := math.Pow(2, float64(count))
+		// Two history data values are merged into a single value
+		tm.history[i] = ((tm.history[i] * (x - 1)) + tm.history[i+1]) / x
+	}
+}
+
+// Map the interval value down to an offset from the beginning of history
+func intervalToHistoryOffset(interval int) int {
```
ebuchman: let's add to the comment what's going on here, ie. `// ie. int(floor(log2(interval))`

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,574 @@
...
+}
+
+// Calculates the integral (history) component of the trust value
+func (tm *TrustMetric) calcHistoryValue() float64 {
+	var wk []float64
+
+	// Create the weights.
+	hlen := tm.numIntervals
+	for i := 0; i < hlen; i++ {
+		x := math.Pow(.8, float64(i+1)) // Optimistic weight
```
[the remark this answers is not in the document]
ebuchman: Also, we shouldn't need to recompute these `wk` every time right? Should we keep the `wk` in the `tm` and just append to it if needed?

Same with `wsum`.

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,574 @@
...
+// trust metric tracking window to be large while maintaining a small
+// number of history data values
+func (tm *TrustMetric) updateFadedMemory() {
+	if tm.historySize < 2 {
+		return
+	}
+
+	first := tm.historySize - 1
+	// Keep the most recent history element
+	for count, i := 1, first-1; count < tm.historySize; count, i = count+1, i-1 {
```
ebuchman: this is hard to read. is it possible to simplify?

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,574 @@
...
+	defer tms.mtx.Unlock()
+
+	// Stop all trust metric go-routines
+	for _, tm := range tms.peerMetrics {
+		tm.Stop()
+	}
+
+	// Make the final trust history data save
+	tms.saveToDB()
+	tms.BaseService.OnStop()
```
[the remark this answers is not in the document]
caffix: I wasn't sure. Thank you for letting me know.

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,574 @@
...
+}
+
+// NewTrustMetricStore returns a store that saves data to the DB
+// and uses the config when creating new trust metrics
+func NewTrustMetricStore(db dbm.DB, tmc TrustMetricConfig) *TrustMetricStore {
+	tms := &TrustMetricStore{
+		peerMetrics: make(map[string]*TrustMetric),
+		db:          db,
+		config:      tmc,
+		stop:        make(chan int, 2),
```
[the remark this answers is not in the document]
caffix: I didn't want it to block the other side.

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,574 @@
...
+	mtx sync.Mutex
+
+	// The db where peer trust metric history data will be stored
+	db dbm.DB
+
+	// This configuration will be used when creating new TrustMetrics
+	config TrustMetricConfig
+
+	// This channel is used to stop the store go-routine
+	stop chan int
```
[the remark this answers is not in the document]
caffix: I agree about the int type and will fix it.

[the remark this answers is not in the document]
caffix: I'll check out the BaseService channel as well.

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,574 @@
...
+/* Both of these methods assume the mutex has been acquired, since they write to the map */
+
+var trustMetricKey = []byte("trustMetricStore")
+
+type peerHistoryJSON struct {
+	NumIntervals int       `json:"intervals"`
+	History      []float64 `json:"history"`
+}
+
+// Loads the history data for the Peer identified by key from the store DB.
```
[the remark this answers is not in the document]
caffix: Good catch.

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,574 @@
...
+		tm.historyValue = tm.calcHistoryValue()
+		// Load the peer trust metric into the store
+		tms.peerMetrics[key] = tm
+	}
+	return true
+}
+
+// Saves the history data for all peers to the store DB
+func (tms *TrustMetricStore) saveToDB() {
+	tms.Logger.Info("Saving TrustHistory to DB", "size", tms.size())
```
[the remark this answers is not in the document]
caffix: I'm fixing the history slice during load. I'm not sure what is wrong with the logger info call.

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,574 @@
...
+	bytes, err := json.Marshal(peers)
+	if err != nil {
+		tms.Logger.Error("Failed to encode the TrustHistory", "err", err)
+		return
+	}
+	tms.db.SetSync(trustMetricKey, bytes)
+}
+
+// Periodically saves the trust history data to the DB
+func (tms *TrustMetricStore) periodicSave() {
```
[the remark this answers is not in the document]
caffix: Sure

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,574 @@
...
+func (tm *TrustMetric) Pause() {
+	tm.stop <- false
+}
+
+// Stop tells the metric to stop recording data over time intervals
+func (tm *TrustMetric) Stop() {
+	tm.stop <- true
+}
+
+// BadEvent indicates that an undesirable event took place
```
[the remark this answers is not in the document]
caffix: It was just to create a shorter option. I can remove them if you prefer

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,574 @@
...
+func (tm *TrustMetric) TrustValue() float64 {
+	resp := make(chan float64, 1)
+
+	tm.trustValue <- &reqTrustValue{Resp: resp}
+	return <-resp
+}
+
+// TrustScore gets a score based on the trust value always between 0 and 100
+func (tm *TrustMetric) TrustScore() int {
+	resp := make(chan float64, 1)
```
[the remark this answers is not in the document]
caffix: It was for convenience. One returns the float with a value 0 through 1, and the other returns the int with the value 0 through 100. Only the TrustValue method is necessary

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,574 @@
...
+	return hv
+}
+
+// Calculates the current score for good/bad experiences
+func (tm *TrustMetric) proportionalValue() float64 {
+	value := 1.0
+	// Bad events are worth more in the calculation of our score
+	total := tm.good + math.Pow(tm.bad, 2)
+
+	if tm.bad > 0 || tm.good > 0 {
```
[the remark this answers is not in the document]
caffix: Yes

p2p/trust/trustmetric.go
caffix: Sure

p2p/trust/trustmetric.go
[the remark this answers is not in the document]
caffix: I can definitely call TrustValue, instead of use the channel directly, if that's what you wanted

p2p/trust/trustmetric.go
```diff
@@ -0,0 +1,574 @@
...
+		hv += tm.fadedMemoryValue(i) * weight
+	}
+	return hv
+}
+
+// Calculates the current score for good/bad experiences
+func (tm *TrustMetric) proportionalValue() float64 {
+	value := 1.0
+	// Bad events are worth more in the calculation of our score
+	total := tm.good + math.Pow(tm.bad, 2)
```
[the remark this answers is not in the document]
caffix: This was working well in my tests, but I'm not sure we will want to keep it this way after conducting the testnet tests.

p2p/trust/trustmetric.go
caffix: You are correct. These weights have worked well. The result is that no positive change happens when current behavior is >= previous behavior, yet the penalty is at maximum when the current behavior falls below previous behavior. I will include the information in the proposal, and of course we can play with the numbers in the future if necessary.

p2p/trust/trustmetric.go
caffix: Sure

p2p/trust/trustmetric.go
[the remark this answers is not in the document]
caffix: I landed on the 0.8 value from testing as well. I'll make sure these details are in the proposals as we move forward. As far as the pessimistic weight is concerned; I tried it, and didn't like the results, which caused me to use the optimistic weight. The optimistic weight creates larger weights for newer history data values, while the the pessimistic weight creates larger weights for time intervals with lower scores. I'll optimize the calculation of the history value as suggested.

caffix:
I have made updates to the proposal and implementation that address your questions and suggestions. To answer your question regarding the channels / select approach to handling synchronization, it was presented in all the reading I've done on the Go programming language as "The Go Way," and it has grown on me over time. It also allows all the synchronized access to happen in one place for the requests/updates/timer fires. I'm very comfortable with the alternative approach as well, and can implement it however you prefer.

ebuchman:
Awesome work @caffix !

caffix:
A pleasure to contribute to the Tendermint project!

PR #788, opened 2017-10-26
new pubsub package

melekes:
replaces #568 

Prehistory: https://gist.github.com/melekes/131291e5144e31d34ac3ed3ab7ff2aca


consensus/reactor.go
```diff
@@ -317,24 +321,47 @@ func (conR *ConsensusReactor) FastSync() bool {
 
 //--------------------------------------
 
-// Listens for new steps and votes,
-// broadcasting the result to peers
-func (conR *ConsensusReactor) registerEventCallbacks() {
+// broadcastNewRoundStepsAndVotes subscribes for new round steps and votes
+// using the event bus and broadcasts events to peers upon receiving them.
+func (conR *ConsensusReactor) broadcastNewRoundStepsAndVotes() error {
+	subscriber := "consensus-reactor"
```
melekes: `const subscriber = ...`

consensus/replay_file.go
```diff
@@ -42,7 +44,14 @@ func (cs *ConsensusState) ReplayFile(file string, console bool) error {
 	cs.startForReplay()
 
 	// ensure all new step events are regenerated as expected
-	newStepCh := subscribeToEvent(cs.evsw, "replay-test", types.EventStringNewRoundStep(), 1)
+	newStepCh := make(chan interface{}, 1)
+
+	ctx := context.Background()
+	err := cs.eventBus.Subscribe(ctx, "replay-file", types.EventQueryNewRoundStep, newStepCh)
```
melekes: extract const `const subscriber = "replay-file"`

rpc/client/httpclient.go
```diff
@@ -186,128 +186,114 @@ func (c *HTTP) Validators(height *int) (*ctypes.ResultValidators, error) {
...
-// After being reconnected, it is necessary to redo subscription
-// to server otherwise no data will be automatically received
+// After being reconnected, it is necessary to redo subscription to server
+// otherwise no data will be automatically received.
 func (w *WSEvents) redoSubscriptions() {
-	for event, _ := range w.evtCount {
-		w.subscribe(event)
+	for query, out := range w.subscriptions {
+		// NOTE: no timeout for reconnect
+		w.Subscribe(context.Background(), query, out)
```
melekes: this will return an err since subscription already exists. should be `w.ws.Subscibe(...`

rpc/client/localclient.go
```diff
@@ -62,50 +71,82 @@ func (c Local) ABCIQueryWithOptions(path string, data data.Bytes, opts ABCIQuery
...
+func (Local) Tx(hash []byte, prove bool) (*ctypes.ResultTx, error) {
 	return core.Tx(hash, prove)
 }
+
+func (c *Local) Subscribe(ctx context.Context, query string, out chan<- interface{}) error {
+	q, err := tmquery.New(query)
+	if err != nil {
+		return errors.Wrap(err, "failed to subscribe")
+	}
+	if err = c.EventBus.Subscribe(ctx, "rpclocalclient", q, out); err != nil {
```
melekes: rpc-local-client

rpc/core/events.go
```diff
@@ -71,8 +95,21 @@ func Subscribe(wsCtx rpctypes.WSRPCContext, event string) (*ctypes.ResultSubscri
...
+	logger.Info("Unsubscribe from query", "remote", addr, "query", query)
+	q, ok := wsCtx.DeleteSubscription(query)
+	if !ok {
+		return nil, errors.New("subscription not found")
+	}
+	eventBus.Unsubscribe(context.Background(), addr, q.(*tmquery.Query))
+	return &ctypes.ResultUnsubscribe{}, nil
+}
+
+func UnsubscribeAll(wsCtx rpctypes.WSRPCContext) (*ctypes.ResultUnsubscribe, error) {
```
melekes: write a changelog entry for this

ebuchman:
hmm racy test in `certifiers` @ethanfrey

melekes:
test uncovered a data race. fixed in https://github.com/tendermint/tendermint/pull/788/commits/6642cf407233e66f23c94c37cbf11588a8182e3b

not excited about the solution. maybe I should not extract eventHub to a separate routine after all...

consensus/byzantine_test.go
```diff
@@ -86,6 +79,16 @@ func TestByzantine(t *testing.T) {
 		reactors[i] = conRI
 	}
 
+	defer func() {
```
odeke-em: Why did we move this down here?

consensus/common_test.go
```diff
@@ -267,13 +280,13 @@ func loadPrivValidator(config *cfg.Config) *types.PrivValidatorFS {
 	return privValidator
 }
 
-func fixedConsensusStateDummy() *ConsensusState {
+func fixedConsensusStateDummy(config *cfg.Config, logger log.Logger) *ConsensusState {
```
odeke-em: Nice!

consensus/state.go
```diff
@@ -269,25 +275,22 @@ func (cs *ConsensusState) Wait() {
 }
 
 // OpenWAL opens a file to log all consensus messages and timeouts for deterministic accountability
-func (cs *ConsensusState) OpenWAL(walFile string) (err error) {
-	err = cmn.EnsureDir(filepath.Dir(walFile), 0700)
+func (cs *ConsensusState) OpenWAL(walFile string) (WAL, error) {
+	err := cmn.EnsureDir(filepath.Dir(walFile), 0700)
```
odeke-em: It'd seem to me that this functionality would instead be in `NewWAL` instead.

consensus/state.go
```diff
@@ -1216,9 +1214,12 @@ func (cs *ConsensusState) finalizeCommit(height int) {
...
 	//	* Fire on start up if we haven't written any new WAL msgs
 	//   Both options mean we may fire more than once. Is that fine ?
-	types.FireEventNewBlock(cs.evsw, types.EventDataNewBlock{block})
-	types.FireEventNewBlockHeader(cs.evsw, types.EventDataNewBlockHeader{block.Header})
-	eventCache.Flush()
+	cs.eventBus.PublishEventNewBlock(types.EventDataNewBlock{block})
+	cs.eventBus.PublishEventNewBlockHeader(types.EventDataNewBlockHeader{block.Header})
+	err = txEventBuffer.Flush()
+	if err != nil {
+		cs.Logger.Error("Failed to flush event buffer", "err", err)
```
odeke-em: Did we mean to return after this point?

types/events.go
```diff
@@ -140,112 +139,54 @@ type EventDataVote struct {
 	Vote *Vote
 }
 
-func (_ EventDataNewBlock) AssertIsTMEventData()       {}
-func (_ EventDataNewBlockHeader) AssertIsTMEventData() {}
-func (_ EventDataTx) AssertIsTMEventData()             {}
-func (_ EventDataRoundState) AssertIsTMEventData()     {}
-func (_ EventDataVote) AssertIsTMEventData()           {}
-
+func (_ EventDataNewBlock) AssertIsTMEventData()          {}
```
odeke-em: We don't need these anymore, and perhaps if we still do, let's use type assertions.

types/event_bus_test.go
```diff
@@ -0,0 +1,122 @@
...
+			benchmarkEventBus(bm.numClients, bm.randQueries, bm.randEvents, b)
+		})
+	}
+}
+
+func benchmarkEventBus(numClients int, randQueries bool, randEvents bool, b *testing.B) {
+	// for random* functions
+	rand.Seed(time.Now().Unix())
+
+	eventBus := NewEventBusWithBufferCapacity(0) // set buffer capacity to 0 so we are not testing cache
```
odeke-em: Nice!

types/event_bus_test.go
```diff
@@ -0,0 +1,122 @@
...
+	eventBus.Start()
+	defer eventBus.Stop()
+
+	ctx := context.Background()
+	q := EventQueryNewBlock
+
+	for i := 0; i < numClients; i++ {
+		ch := make(chan interface{})
```
odeke-em: Am curious, in the design of our events, when do we this channel? I ask because perhaps instead eventBus.Subscribe should return the channel as a read-only one? That was it can be closed to indicate when no more events are coming through or on error. We can discuss that later though, that shouldn't block your progress on this PR since this work is moving over the old events package.

types/event_bus.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (b *EventBus) Unsubscribe(ctx context.Context, subscriber string, query tmpubsub.Query) error {
+	return b.pubsub.Unsubscribe(ctx, subscriber, query)
+}
+
+func (b *EventBus) UnsubscribeAll(ctx context.Context, subscriber string) error {
+	return b.pubsub.UnsubscribeAll(ctx, subscriber)
+}
+
+func (b *EventBus) publish(eventType string, eventData TMEventData) error {
```
odeke-em: How come publish is unexported but *Subscribe* are all exported? Just a question, not a blocker :)

rpc/core/events.go
```diff
@@ -33,14 +39,32 @@ import (
...
+	defer cancel()
+	ch := make(chan interface{})
+	err = eventBus.Subscribe(ctx, addr, q, ch)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to subscribe")
+	}
+
+	wsCtx.AddSubscription(query, q)
+
+	go func() {
```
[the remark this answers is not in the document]
melekes: when ch is closed, the loop will stop and so this goroutine https://gobyexample.com/range-over-channels

rpc/lib/server/handlers.go
```diff
@@ -487,6 +478,23 @@ func (wsc *wsConnection) TryWriteRPCResponse(resp types.RPCResponse) bool {
 	}
 }
 
+func (wsc *wsConnection) AddSubscription(query string, data interface{}) {
```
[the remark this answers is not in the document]
melekes: It has to be somewhere. It's either here or inside pubsub. Right now pubsub is kinda dumb and can subscribe the same client to the same query

consensus/byzantine_test.go
melekes: it just felt more natural because the order is always: `file.Open(); defer f.Close()` and not the other way around

consensus/state.go
melekes: not sure. plus the current impl never returns an error. I just thought it will be good to have the interface of eventHub to return an error in case impl will change.

types/event_bus_test.go
melekes: > That was it can be closed to indicate when no more events are coming through or on error.

this is the case right now.

> I ask because perhaps instead eventBus.Subscribe should return the channel as a read-only one? 

I remember going back and forth between the current design and this. The current design allows a subscriber to choose the type of the channel: buffered (we need this in tests) or unbuffered.

consensus/reactor.go
```diff
@@ -317,24 +321,47 @@ func (conR *ConsensusReactor) FastSync() bool {
...
-	types.AddListenerForEvent(conR.evsw, "conR", types.EventStringVote(), func(data types.TMEventData) {
-		edv := data.Unwrap().(types.EventDataVote)
-		conR.broadcastHasVoteMessage(edv.Vote)
-	})
+	// votes
+	votesCh := make(chan interface{})
+	err = conR.eventBus.Subscribe(ctx, subscriber, types.EventQueryVote, votesCh)
+	if err != nil {
+		return errors.Wrapf(err, "failed to subscribe %s to %s", subscriber, types.EventQueryVote)
+	}
```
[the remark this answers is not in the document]
melekes: fixed

ebuchman:
I still don't understand what `wsctx.AddSubscription` is good for. When do we ever do anything with the contents of the `subscriptions` map? Seems we just add to and delete. Is it just so we can return errors?

melekes:
> Is it just so we can return errors?

for the most part, yes. if the client subscribes twice using the same query, he will receive 2 events (not one). Do you think we should not be concerned about it?

consensus/state.go
ebuchman: errors in this function are critical. notice all the panics at the start of this function. `.Flush` should be expected to not error, but it's not critical, so long as nothing relies on it other than the websockets. The current code looks fine (ie just log the error). But if one day Flush actually errors, we may need a better way to get that error back up to the rpc server ...

consensus/types/state.go
```diff
@@ -77,10 +77,20 @@ type RoundState struct {
...
 	edrs := types.EventDataRoundState{
-		Height:     rs.Height,
-		Round:      rs.Round,
-		Step:       rs.Step.String(),
-		RoundState: rs,
+		Height: rs.Height,
+		Round:  rs.Round,
+		Step:   rs.Step.String(),
+		// send only fields needed by makeRoundStepMessages
+		RoundState: &RoundState{
```
ebuchman: we need to have a larger discussion/investigation about the races this is meant to address. AFAIK they only started showing up after a seemingly innocuous change to go-wire, which has been rolled back, so we shouldn't need to do this now. But we will need to investigate and solve it properly soon ...

rpc/client/httpclient.go
```diff
@@ -186,128 +186,115 @@ func (c *HTTP) Validators(height *int) (*ctypes.ResultValidators, error) {
...
 
-/** TODO: more intelligent subscriptions! **/
-func (w *WSEvents) AddListenerForEvent(listenerID, event string, cb events.EventCallback) {
-	// no one listening -> subscribe
-	if w.evtCount[event] == 0 {
-		w.subscribe(event)
+func (w *WSEvents) Subscribe(ctx context.Context, query string, out chan<- interface{}) error {
+	w.mtx.RLock()
+	if _, ok := w.subscriptions[query]; ok {
+		return errors.New("already subscribed")
```
ebuchman: don't we need to unlock in here?

rpc/client/localclient.go
```diff
@@ -33,79 +45,113 @@ type Local struct {
...
+func (Local) Validators(height *int) (*ctypes.ResultValidators, error) {
 	return core.Validators(height)
 }
 
-func (c Local) Tx(hash []byte, prove bool) (*ctypes.ResultTx, error) {
+func (Local) Tx(hash []byte, prove bool) (*ctypes.ResultTx, error) {
 	return core.Tx(hash, prove)
 }
+
+func (c *Local) Subscribe(ctx context.Context, query string, out chan<- interface{}) error {
```
ebuchman: is it weird that suddenly the receiver is a pointer?

rpc/core/events.go
```diff
@@ -33,14 +39,35 @@ import (
...
+	ch := make(chan interface{})
+	err = eventBus.Subscribe(ctx, addr, q, ch)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to subscribe")
+	}
+
+	go func() {
+		for event := range ch {
+			tmResult := &ctypes.ResultEvent{query, event.(tmtypes.TMEventData)}
+			wsCtx.TryWriteRPCResponse(rpctypes.NewRPCSuccessResponse(wsCtx.Request.ID+"#event", tmResult))
```
ebuchman: using TryWrite here means we could lose these events. Now that this is running in its own go-routine, rather than as a callback, should we consider letting it block ?

rpc/core/mempool.go
```diff
@@ -147,20 +150,26 @@ func BroadcastTxSync(tx types.Tx) (*ctypes.ResultBroadcastTx, error) {
...
-		deliverTxResCh <- data.Unwrap().(types.EventDataTx)
-	})
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Millisecond)
+	defer cancel()
+	deliverTxResCh := make(chan interface{})
+	q := types.EventQueryTx(tx)
+	err := eventBus.Subscribe(ctx, "mempool", q, deliverTxResCh)
+	if err != nil {
+		err = errors.Wrap(err, "failed to subscribe to tx")
+		logger.Error("Error broadcasting transaction", "err", err)
```
ebuchman: There was no error broadcasting the tx, we just failed to subscribe

rpc/core/pipe.go
```diff
@@ -51,14 +50,11 @@ var (
 	addrBook         *p2p.AddrBook
 	txIndexer        txindex.TxIndexer
 	consensusReactor *consensus.ConsensusReactor
+	eventBus         *types.EventBus
```
ebuchman: is this not an "external, thread safe interface" ?

rpc/core/routes.go
```diff
@@ -7,8 +7,9 @@ import (
 // TODO: better system than "unsafe" prefix
 var Routes = map[string]*rpc.RPCFunc{
 	// subscribe/unsubscribe are reserved for websocket events.
-	"subscribe":   rpc.NewWSRPCFunc(Subscribe, "event"),
-	"unsubscribe": rpc.NewWSRPCFunc(Unsubscribe, "event"),
+	"subscribe":       rpc.NewWSRPCFunc(Subscribe, "query"),
```
ebuchman: breaking

types/event_bus.go
```diff
@@ -0,0 +1,133 @@
...
+func (b *EventBus) Unsubscribe(ctx context.Context, subscriber string, query tmpubsub.Query) error {
+	return b.pubsub.Unsubscribe(ctx, subscriber, query)
+}
+
+func (b *EventBus) UnsubscribeAll(ctx context.Context, subscriber string) error {
+	return b.pubsub.UnsubscribeAll(ctx, subscriber)
+}
+
+func (b *EventBus) Publish(eventType string, eventData TMEventData) error {
+	if b.pubsub != nil {
```
ebuchman: When is this nil? And can `b.pubsub.PublishWithTags` handle the nil check instead? The other `EventBus` methods don't do this check. Should they?

ebuchman:
Note this is a breaking change for a number of reasons:

- types/events.go has been overhauled
- related changes in `node`
- related changes in `rpc/client`, especially re `WaitForOneEvent` and the switch from `AddListenerForEvent` to `Subscribe`
- changed parameter name `event` to `query` in the Subscribe ws request

rpc/core/events.go
melekes: then the whole pubsub will block because it won't be able to write to that channel again

rpc/client/localclient.go
melekes: why? methods above just delegate to `core.X` static funcs. here we using and mutating the Local's fields.

rpc/core/mempool.go
melekes: This refers to the whole method - broadcastTxCommit. maybe we should change msg to `broadcastTxCommit failed`

rpc/core/events.go
ebuchman: oh duh

rpc/core/mempool.go
ebuchman: sure, or `Error on broadcastTxCommit`

ebuchman:
Seems like we no longer need most of consensus/test_data, except for `many_blocks.cswal`. Should we consider just generating the equivalent of that on the fly during the test, and then we can delete that dir all together?

We can save the `build.sh` in `scripts` so we dont lose it if we think it could still be useful

ebuchman:
> for the most part, yes. if the client subscribes twice using the same query, he will receive 2 events (not one). Do you think we should not be concerned about it?

Can we just prevent this directly in `eventBus.Subscribe` ?

rpc/lib/server/handlers.go
```diff
@@ -445,9 +445,6 @@ func (wsc *wsConnection) OnStart() error {
 
 // OnStop unsubscribes from all events.
```
ebuchman: Is this comment still correct? We no longer unsubscribe from within this function. Should we?

rpc/core/pipe.go
melekes: thread safe object?

melekes:
Do you think it's better to switch to the old mutex-based implementation (events)? Looks like we're not gaining more by switching to channel-based impl running in a separate goroutine.

👍  speed increase by offloading event notifications to a separate goroutine
👎  we can't pass pointers that we're mutating later on (https://github.com/tendermint/tendermint/pull/788#discussion_r147881824)
👎  it is hard to implement https://github.com/tendermint/tendermint/pull/788#issuecomment-340877695

Or maybe we should use 2 - events for internal events (consumed by consensus/reactor), pubsub for external clients.

ebuchman:
>  we can't pass pointers that we're mutating later on

Is this a new problem that didnt happen before? Like when the event was published synchronously, we didn't have to worry about some competing thread manipulating the same data?

melekes:
> Is this a new problem that didnt happen before?

Frey's tests helped to uncover it. it was there from the beginning on this feature

> Like when the event was published synchronously, we didn't have to worry about some competing thread manipulating the same data?

yes, because it was happening in the same thread, so no race.

Issue #740, opened 2017-10-12
Replay: wal.group.Search returned EOF 


rigelrozanski:
We should probably include a mechanism for dealing with the end of the replay which is not an error... It seems to set off a chain reaction of small errors in my case

Reproduction:
 - Connect to running blockchain with a node which has no voting power on that blockchain
reproduction: 

main error:
```
Error on catchup replay. Proceeding to start ConsensusState anyway module=consensus err="Cannot replay height 1048. WAL does not contain #ENDHEIGHT for 1047."
```

Detailed error:
```
I[10-12|20:21:04.872] Executed block                               module=state height=1046 validTxs=0 invalidTxs=0
I[10-12|20:21:04.873] Committed state                              module=state height=1046 txs=0 hash=3CA795AF0D05933A96E01FAC942A3A38D2914496
I[10-12|20:21:04.879] Executed block                               module=state height=1047 validTxs=0 invalidTxs=0
I[10-12|20:21:04.880] Committed state                              module=state height=1047 txs=0 hash=288A1E464C48BF6D65A39B762635547BE61900F6
E[10-12|20:21:04.976] Replay: wal.group.Search returned EOF        module=consensus #ENDHEIGHT=1047
E[10-12|20:21:04.976] Error on catchup replay. Proceeding to start ConsensusState anyway module=consensus err="Cannot replay height 1048. WAL does not contain #ENDHEIGHT for 1047."
E[10-12|20:21:05.379] Attempt to finalize failed. We don't have the commit block. module=consensus height=1048 proposal-block= commit-block=48967C767CB5D1A1117F8CCD110FF25A7C138405
I[10-12|20:21:05.691] Executed block                               module=state height=1048 validTxs=0 invalidTxs=0
I[10-12|20:21:05.692] Committed state                              module=state height=1048 txs=0 hash=9844C65E29C5761D5AC2975A8988C92472EF0E39
E[10-12|20:21:06.166] Attempt to finalize failed. We don't have the commit block. module=consensus height=1049 proposal-block= commit-block=A48EBE3D420413477AAAB52DAF6692882624677D
I[10-12|20:21:06.521] Executed block                               module=state height=1049 validTxs=0 invalidTxs=0
I[10-12|20:21:06.523] Committed state                              module=state height=1049 txs=0 hash=273047DB777C7ECA138E1CCED1F3A9BD6F5F8D09
[...]

PR #791, opened 2017-10-26
dont catchupReplay on wal if we fast synced

ebuchman:
Addresses https://github.com/tendermint/tendermint/issues/740

Not thrilled about it though. Anyone have better ideas?

### DIFF
--- a/blockchain/pool.go
+++ b/blockchain/pool.go
@@ -26,10 +26,10 @@ eg, L = latency = 0.1s
 */
 
 const (
-	requestIntervalMS         = 250
-	maxTotalRequesters        = 300
+	requestIntervalMS         = 100
+	maxTotalRequesters        = 1000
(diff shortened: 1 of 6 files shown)


melekes:
Tests failed with:

```
consensus/byzantine_test.go:104:32: not enough arguments in call to byzR.reactor.SwitchToConsensus
        have (*state.State)
        want (*state.State, int)
consensus/byzantine_test.go:107:23: not enough arguments in call to cr.SwitchToConsensus
        have (*state.State)
        want (*state.State, int)
consensus/reactor_test.go:57:32: not enough arguments in call to reactors[i].SwitchToConsensus
        have (*state.State)
        want (*state.State, int)
```

ebuchman:
Woops pushed some other blockchain pkg changes in there. But think its fine

PR #794, opened 2017-10-27
Kill Tendermint when App dies

ebuchman:
https://github.com/tendermint/tendermint/issues/243#issuecomment-339924812

Tendermint should die if app hash does so OS can handle restarts

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -1204,6 +1204,10 @@ func (cs *ConsensusState) finalizeCommit(height int) {
 	err := stateCopy.ApplyBlock(eventCache, cs.proxyAppConn, block, blockParts.Header(), cs.mempool)
 	if err != nil {
 		cs.Logger.Error("Error on ApplyBlock. Did the application crash? Please restart tendermint", "err", err)
+		err := cmn.Kill()
+		if err != nil {
+			cs.Logger.Error("Failed to kill this process - please do so manually", "err", err)
+		}
 		return
 	}
 


consensus/state.go
```diff
@@ -1204,6 +1204,10 @@ func (cs *ConsensusState) finalizeCommit(height int) {
 	err := stateCopy.ApplyBlock(eventCache, cs.proxyAppConn, block, blockParts.Header(), cs.mempool)
 	if err != nil {
 		cs.Logger.Error("Error on ApplyBlock. Did the application crash? Please restart tendermint", "err", err)
```
odeke-em: Perhaps we should remove this error message then and instead make it
```go
if err != nil {
    if err := cmn.Kill(); err != nil {
        // Maybe log that we successfully killed Tendermint
    } else {
        cs.Logger.Error("Error on ApplyBlock. Did the application crash? Please restart Tendermint manually", "err", err)
    }
    return
}

ebuchman: I think we need to alert the user when the app crashes, and then initiate tear down. If tear down fails, we should log that error too.

odeke-em: Alright, at least let's change the message asking them to start Tendermint yet we are doing it right after logging the message
screen shot 2017-10-28 at 3 44 42 pm

ebuchman: What do you mean? We tell the user to restart tendermint, and then we go and kill it. So then they, or their OS, can restart it!

odeke-em: Oops I was definitely mistaken. For some reason when I read the commit message `Tendermint should die if app hash does so OS can handle restarts` I assumed that Tendermint would be auto-restarted or the OS would handle it exiting, but I re-read everything after your comment. My apologies.

Issue #808, opened 2017-11-06
p2p: code for MakeConnectedSwitches might be invalid every switch connecting to itself


odeke-em:
I was just reading through the MakeConnectedSwitches code and noticed that the 2 nested loops
https://github.com/tendermint/tendermint/blob/ec87c740a7cdc8591e97188f9e5ebfb4ded9d020/p2p/switch.go#L512-L516

pass in i == j thereby connecting every switch to itself. Is this intended? Doesn't seem so at least according to the comment https://github.com/tendermint/tendermint/blob/ec87c740a7cdc8591e97188f9e5ebfb4ded9d020/p2p/switch.go#L523-L525.

This code seems to only be used in tests
```shell
$ grep -Rn 'MakeConnectedSwitches' *
consensus/byzantine_test.go:87:	p2p.MakeConnectedSwitches(config.P2P, N, func(i int, s *p2p.Switch) *p2p.Switch {
consensus/reactor_test.go:47:	p2p.MakeConnectedSwitches(config.P2P, N, func(i int, s *p2p.Switch) *p2p.Switch {
mempool/reactor_test.go:48:	p2p.MakeConnectedSwitches(config.P2P, N, func(i int, s *p2p.Switch) *p2p.Switch {
p2p/CHANGELOG.md:53:- Fixes for MakeConnectedSwitches and Connect2Switches
p2p/switch.go:497:// MakeConnectedSwitches returns n switches, connected according to the connect func.
p2p/switch.go:501:func MakeConnectedSwitches(cfg *cfg.P2PConfig, n int, initSwitch func(int, *Switch) *Switch, connect func([]*Switch, int, int)) []*Switch {
p2p/switch_test.go:96:	switches := MakeConnectedSwitches(config, 2, initSwitch, Connect2Switches)
```
 but nonetheless it is exported code from p2p that's used in consensus tests.

Simple repro at 
https://play.golang.org/p/R5zbM3UqhN
```go
package main

import "fmt"

func connectThem(i, j int) {
	if i == j {
		fmt.Printf("i=%d == j\n", i)
	}
}

func main() {
	n := 10
	for i := 0; i < n; i++ {
		for j := i; j < n; j++ {
			connectThem(i, j)
		}
	}
}
```
which gives 
```shell
i=0 == j
i=1 == j
i=2 == j
i=3 == j
i=4 == j
i=5 == j
i=6 == j
i=7 == j
i=8 == j
i=9 == j
```

melekes:
[...]

PR #812, opened 2017-11-07
MakeConnectedSwitches: connect first switch to others

melekes:
Refs #808


ebuchman:
No tests depend on full connectivity yet. We could add a test that asserts all switches have N-1 peers after running MakeConnectedSwitches. Also a test where the 0 node went down and we expected everyone to still be able to communicate.

Issue #790, opened 2017-10-26
Avoid `time.Sleep` in tests


ebuchman:
We should try to use channels and tickers whenever possible so we can set higher timeouts without impeding the tests.

Sometimes circle fails since these aren't heigh enough.

Eg. https://circleci.com/gh/tendermint/tendermint/1836#artifacts/containers/0 is due to https://github.com/tendermint/tendermint/blob/master/p2p/pex_reactor_test.go#L101

zramsay:
#820

PR #820, opened 2017-11-07
prefer tickers to time.Sleep

melekes:
Refs #790

### DIFF
--- a/consensus/mempool_test.go
+++ b/consensus/mempool_test.go
@@ -141,10 +141,12 @@ func TestRmBadTx(t *testing.T) {
 		}
 
 		// check for the tx
-		txs := cs.mempool.Reap(1)
-		if len(txs) == 0 {
-			emptyMempoolCh <- struct{}{}
-			return
+		for {
+			txs := cs.mempool.Reap(1)
+			if len(txs) == 0 {
+				emptyMempoolCh <- struct{}{}
+			}
+			time.Sleep(10 * time.Millisecond)
 		}
 	}()
 
--- a/consensus/mempool_test.go
+++ b/consensus/mempool_test.go
@@ -124,8 +124,8 @@ func TestRmBadTx(t *testing.T) {
 	app.DeliverTx(txBytes)
 	app.Commit()
 
-	ch := make(chan struct{})
-	cbCh := make(chan struct{})
+	emptyMempoolCh := make(chan struct{})
+	checkTxRespCh := make(chan struct{})
 	go func() {
 		// Try to send the tx through the mempool.
 		// CheckTx should not err, but the app should return a bad abci code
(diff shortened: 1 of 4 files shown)


melekes:
```
BEFORE:
real    2m37.473s
user    3m5.428s
sys     0m27.388s

AFTER:
real    2m35.850s
user    3m11.070s
sys     0m27.979s
```

consensus/mempool_test.go
```diff
@@ -134,28 +134,24 @@ func TestRmBadTx(t *testing.T) {
...
 		// check for the tx
-		for {
-			time.Sleep(time.Second)
-			txs := cs.mempool.Reap(1)
-			if len(txs) == 0 {
-				ch <- struct{}{}
-				return
-			}
-
+		txs := cs.mempool.Reap(1)
```
ebuchman: how can we remove the for loop? what if the tx hasn't been removed yet

ebuchman: would be best if we show the mempool is not empty first, and then it becomes empty when the abci app returns.

node/node_test.go
```diff
@@ -20,19 +20,23 @@ func TestNodeStartStop(t *testing.T) {
...
 	t.Logf("Started node %v", n.sw.NodeInfo())
 
-	// Wait a bit to initialize
-	// TODO remove time.Sleep(), make asynchronous.
-	time.Sleep(time.Second * 2)
+	ticker := time.NewTicker(10 * time.Millisecond)
+	select {
+	case <-ticker.C:
+		if n.IsRunning() {
+			return
```
ebuchman: doesnt this abort the test? maybe you mean `break`?

ebuchman: Also `IsRunning` should be true right away. Maybe we should wait for a block to be produced?

p2p/pex_reactor_test.go
```diff
@@ -98,22 +99,39 @@ func TestPEXReactorRunning(t *testing.T) {
...
 
 	// stop them
 	for _, s := range switches {
 		s.Stop()
 	}
 }
 
+func assertSomePeersWithTimeout(t *testing.T, switches []*Switch, checkPeriod, timeout time.Duration) {
+	ticker := time.NewTicker(checkPeriod)
+	select {
```
ebuchman: needs a loop?

p2p/pex_reactor_test.go
```diff
@@ -98,22 +99,39 @@ func TestPEXReactorRunning(t *testing.T) {
...
+	case <-ticker.C:
+		// check peers are connected
+		allGood := true
+		for _, s := range switches {
+			outbound, inbound, _ := s.NumPeers()
+			if outbound+inbound == 0 {
+				allGood = false
```
ebuchman: can drop allGood and just continue

p2p/switch_test.go
```diff
@@ -131,41 +131,31 @@ func TestSwitches(t *testing.T) {
...
 
-	// Check message on ch2
-	ch2Msgs := s2.Reactor("bar").(*TestReactor).getMsgs(byte(0x02))
-	if len(ch2Msgs) != 1 {
-		t.Errorf("Expected to have received 1 message in ch2")
-	}
-	if !bytes.Equal(ch2Msgs[0].Bytes, wire.BinaryBytes(ch2Msg)) {
-		t.Errorf("Unexpected message bytes. Wanted: %X, Got: %X", wire.BinaryBytes(ch2Msg), ch2Msgs[0].Bytes)
+func assertMsgReceivedWithTimeout(t *testing.T, msg string, channel byte, reactor *TestReactor, checkPeriod, timeout time.Duration) {
+	ticker := time.NewTicker(checkPeriod)
```
ebuchman: loop ...

consensus/mempool_test.go
melekes: looks like mempool is always empty

PR #830, opened 2017-11-09
read all bytes available from wire protocol in MConnection test

petabytestorage:


ebuchman:
hmm what's the story here?

petabytestorage:
reading the source code, i noticed that it seems like the channel would contain the wire-encoded message not the literal message bytes so it was short some bytes for reading on the other end of any ReadWriteCloser. Through guess and check i found it seemed like 4 bytes was the extra space for reasons i am not sure yet @ebuchman however it seemed odd enough to deserve a PR. when i use 5 instead of 4 my test fails so from that i infer i've guessed the right padding.

petabytestorage:
i think this explains the length written: https://github.com/tendermint/tendermint/blob/develop/p2p/connection.go#L663

p2p/connection_test.go
```diff
@@ -44,12 +44,12 @@ func TestMConnectionSend(t *testing.T) {
 	assert.True(mconn.Send(0x01, msg))
 	// Note: subsequent Send/TrySend calls could pass because we are reading from
 	// the send queue in a separate goroutine.
-	server.Read(make([]byte, len(msg)))
+	server.Read(make([]byte, len(wire.BinaryBytes(msg))+4))
```
melekes: have you figured out where 4 bytes are coming from? 1 byte is packetMsgType

ebuchman:
Should I close this for now ?

Issue #829, opened 2017-11-09
[test_integrations] enable logs by default


melekes:
Otherwise, hard to debug. Any objections?

(remove `if CIRCLECI=true`)

PR #834, opened 2017-11-09
Enable logs by default

melekes:
Refs #829


melekes:
> the whole syslog dance

:) 2 lines of code?

I don't care whenever it's syslog or something else. I just want to have access to the logs in case of an error. https://github.com/gliderlabs/logspout is another option, but if we already have syslog why bother adding something else?

PR #835, opened 2017-11-09
Indexing DeliverTx tags

melekes:
Requires https://github.com/tendermint/abci/pull/130
Requires https://github.com/tendermint/tmlibs/pull/84
Requires https://github.com/tendermint/abci/pull/141

## Streaming part
- [x] validate tags? where?
- [x] merge with existing tags

## Long-term indexing
- [x] add a configuration var (`tags_to_index`) - empty by default
- [x] rewrite indexer to be a listener of eventBus
- [x] index txs (only by provided `tags_to_index`)
- [x] add methods for querying the DB (aka query execution)
- [x] add an endpoint for querying transactions
- [x] add new methods to RPC client
-  rewrite indexing service to index a batch of transactions (`it could subscribe to block headers to find out when blocks start and how many txs they have, then wait for those txs and batch the writes.`
TODO: add a note `One day we can have this IndexerService run many go-routines to do the indexing in parallel.`) See #907 
- [x] add docs for `tx_search` RPC endpoint
- benchmark `kvIndexer#Search` See #907 
- [x] modify the dummy to return a tag so we can test that here too!


config/config.go
```diff
@@ -94,7 +97,7 @@ type BaseConfig struct {
 	FilterPeers bool `mapstructure:"filter_peers"` // false
 
 	// What indexer to use for transactions
-	TxIndex string `mapstructure:"tx_index"`
+	TxIndexer string `mapstructure:"tx_index"`
```
ebuchman: should this move into the new `TxIndex` config ?

types/event_bus.go
```diff
@@ -82,7 +82,11 @@ func (b *EventBus) PublishEventVote(vote EventDataVote) error {
 func (b *EventBus) PublishEventTx(tx EventDataTx) error {
 	// no explicit deadline for publishing events
 	ctx := context.Background()
-	b.pubsub.PublishWithTags(ctx, TMEventData{tx}, map[string]interface{}{EventTypeKey: EventTx, TxHashKey: fmt.Sprintf("%X", tx.Tx.Hash())})
+	tags := tx.Tags
+	// add predefined tags (they should overwrite any existing tags)
+	tags[EventTypeKey] = EventTx
```
ebuchman: should we check these here or in the execution and log something if the app returned them so it's notified they were overwritten? At least we need to document this well

types/event_bus.go
```diff
@@ -82,7 +82,11 @@ func (b *EventBus) PublishEventVote(vote EventDataVote) error {
...
+	tags[TxHashKey] = fmt.Sprintf("%X", tx.Tx.Hash())
+	b.pubsub.PublishWithTags(ctx, TMEventData{tx}, tags)
```
ebuchman: <3

types/event_bus.go
melekes: good idea

config/config.go
melekes: yes

node/node.go
```diff
@@ -281,6 +267,30 @@ func NewNode(config *cfg.Config,
...
+	default:
+		txIndexer = &null.TxIndex{}
+	}
+
+	// subscribe for all transactions and index them by tags
+	ch := make(chan interface{})
+	eventBus.Subscribe(context.Background(), "tx_index", types.EventQueryTx, ch)
+	go func() {
+		for event := range ch {
+			// XXX: may be not perfomant to write one event at a time
```
melekes: not sure about this

melekes:
Ok, going to finish this now.

melekes:
OK, as per Frey's suggestion we can index txs

```
tag=value/height/index => hash
...
hash => result
```

but would not it be better to build a btree like Postgresql does for bjson https://www.postgresql.org/docs/9.6/static/datatype-json.html?

melekes:
Now I am thinking about maybe it makes sense to rewrite pubsub/query to use https://github.com/postgrespro/jsquery language. This will free us from the need of transforming query to jsquery in case of Postgresql indexer. This is given jsquery could be more complex (supports complex ops).

melekes:
Ok, so here is what I've come up so far: given DB (levelDB) provides us an ordered map, we can store the following key-value pairs

blank diagram - page 1

melekes:
Query execution is complex stuff :( definitely not one day job. Spend last Sunday working on it. Will continue iterating on it https://github.com/tendermint/tendermint/pull/835/commits/35e0fa5a8aed004012927a9d042b9eefae0a14b6

state/txindex/kv/kv.go
```diff
@@ -46,13 +54,322 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+
+	// extract ranges
+	// if both upper and lower bounds exist, it's better to get them in order not
+	// no iterate over kvs that are not within range.
+	ranges, rangeIndexes := lookForRanges(conditions)
+	if len(ranges) > 0 {
+		skipIndexes = append(skipIndexes, rangeIndexes...)
+	}
+	for _, r := range ranges {
+		hashes2 = txi.matchRange(r, startKeyForRange(r, height, heightIndex > 0))
```
ebuchman: do we need to pass `heightSpecified` ? Isn't checking if the `height == 0` enough ?

state/txindex/kv/kv.go
```diff
@@ -46,13 +54,322 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+	for it.Next() {
+		// no other way to stop iterator other than checking for upperBound
+		switch (r.upperBound).(type) {
+		case int64:
+			v, err := strconv.ParseInt(extractValueFromKey(it.Key()), 10, 64)
+			if err == nil && v == r.upperBound {
+				if r.includeUpperBound {
+					hashes = append(hashes, it.Value())
+				}
+				break
```
ebuchman: I think this only breaks from the switch, not the loop!

state/txindex/kv/kv.go
```diff
@@ -46,13 +54,322 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+}
+
+func (txi *TxIndex) match(c query.Condition, startKey []byte) (hashes [][]byte) {
+	if c.Op == query.OpEqual {
+		it := txi.store.IteratorPrefix(startKey)
+		for it.Next() {
+			hashes = append(hashes, it.Value())
+		}
+	} else if c.Op == query.OpContains {
+		// XXX: full scan
```
ebuchman: why aren't we still starting from the startKey?

state/txindex/indexer.go
```diff
@@ -4,20 +4,24 @@ import (
...
 // TxIndexer interface defines methods to index and search transactions.
 type TxIndexer interface {
 
-	// AddBatch analyzes, indexes or stores a batch of transactions.
-	// NOTE: We do not specify Index method for analyzing a single transaction
-	// here because it bears heavy performance losses. Almost all advanced indexers
-	// support batching.
-	AddBatch(b *Batch) error
+	// AddBatch analyzes, indexes and stores a batch of transactions.
+	AddBatch(b *Batch, allowedTags []string) error
```
ebuchman: would it be better if the `allowedTags` were registered once on the Indexer, rather than with each call ?

state/txindex/kv/kv_test.go
```diff
@@ -11,30 +12,103 @@ import (
...
+	require.NoError(t, err)
+	assert.Equal(t, txResult2, loadedTxResult2)
+}
+
+func TestTxSearch(t *testing.T) {
+	indexer := &TxIndex{store: db.NewMemDB()}
+
+	tx := types.Tx("HELLO WORLD")
+	tags := []*abci.KVPair{
+		&abci.KVPair{Key: "account.number", ValueType: abci.KVPair_INT, ValueInt: 1},
```
ebuchman: drop the struct type prefix and the field names to make this more readable

state/txindex/kv/kv_test.go
```diff
@@ -11,30 +12,103 @@ import (
...
+
+	allowedTags := []string{"account.number", "account.owner", "account.date"}
+	err := indexer.Index(txResult, allowedTags)
+	require.NoError(t, err)
+
+	testCases := []struct {
+		q             string
+		expectError   bool
+		resultsLength int
+		results       []*types.TxResult
```
ebuchman: lets drop `results` and `expectError` here for readability. resultsLength in this test is always {0, 1}, so we can just refer to `txResult` in the test loop, and there's always no error.

types/event_bus.go
```diff
@@ -67,63 +68,106 @@ func (b *EventBus) Publish(eventType string, eventData TMEventData) error {
...
+		switch tag.ValueType {
+		case abci.KVPair_STRING:
+			tags[tag.Key] = tag.ValueString
+		case abci.KVPair_INT:
+			tags[tag.Key] = tag.ValueInt
+		}
+	}
+
+	// add predefined tags
+	if tag, ok := tags[EventTypeKey]; ok {
```
ebuchman: lets make this 3-liner a function, and include the key/value in the error statement

types/event_bus.go
```diff
@@ -67,63 +68,106 @@ func (b *EventBus) Publish(eventType string, eventData TMEventData) error {
...
+
+	if tag, ok := tags[TxHeightKey]; ok {
+		b.Logger.Error("Found predefined tag (value will be overwritten)", "tag", tag)
+	}
+	tags[TxHeightKey] = event.Height
+
+	if tag, ok := tags[TxIndexKey]; ok {
+		b.Logger.Error("Found predefined tag (value will be overwritten)", "tag", tag)
+	}
+	tags[TxIndexKey] = event.Index
```
ebuchman: do we use this for anything ?

state/txindex/kv/kv.go
```diff
@@ -46,13 +54,322 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+	// get a list of conditions (like "tx.height > 5")
+	conditions := q.Conditions()
+
+	// if there is a hash condition, return the result immediately
+	hash, err, ok := lookForHash(conditions)
+	if err != nil {
+		return []*types.TxResult{}, errors.Wrap(err, "error during searching for a hash in the query")
+	} else if ok {
+		res, err := txi.Get(hash)
+		return []*types.TxResult{res}, errors.Wrap(err, "error while retrieving the result")
```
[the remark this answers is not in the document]
melekes: `Wrap(nil, "something") == nil`

state/txindex/kv/kv.go
melekes: yeah, but height == 0 looks kinda fishy...

state/txindex/kv/kv.go
```diff
@@ -46,13 +54,322 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+
+		// no matches
+		if len(hashes2) == 0 {
+			hashes = make(map[string][]byte)
+		} else {
+			// perform intersection as we go
+			for _, h := range hashes2 {
+				k := hashKey(h)
+				if _, ok := hashes[k]; !ok {
+					delete(hashes, k)
```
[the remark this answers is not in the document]
melekes: you're right. basically, we need to intersect 2 sets `{results of querying index A} x {results of querying index B}`

config/config.go
```diff
@@ -412,6 +411,35 @@ func (c *ConsensusConfig) SetWalFile(walFile string) {
...
+// indexer, including tags to index.
+type TxIndexConfig struct {
+	// What indexer to use for transactions
+	//
+	// Options:
+	//   1) "null" (default)
+	//   2) "kv" - the simplest possible indexer, backed by key-value storage (defaults to levelDB; see DBBackend).
+	Indexer string `mapstructure:"indexer"`
+
+	// Comma-separated list of tags to index (by default the only tag is tx hash)
```
melekes: If tx.hash is indexed by default, should we call them `IndexTags` `ExtraIndexTags` then?

proxy/app_conn.go
```diff
@@ -12,12 +12,12 @@ type AppConnConsensus interface {
 	SetResponseCallback(abcicli.Callback)
 	Error() error
 
-	InitChainSync(types.RequestInitChain) (err error)
+	InitChainSync(types.RequestInitChain) error
 
-	BeginBlockSync(types.RequestBeginBlock) (err error)
+	BeginBlockSync(types.RequestBeginBlock) error
```
melekes: `(*types.ResponseBeginBlock, error)`

proxy/app_conn.go
```diff
@@ -12,12 +12,12 @@ type AppConnConsensus interface {
 	SetResponseCallback(abcicli.Callback)
 	Error() error
 
-	InitChainSync(types.RequestInitChain) (err error)
+	InitChainSync(types.RequestInitChain) error
```
melekes: `(*types.ResponseInitChain, error)` https://github.com/tendermint/abci/blob/develop/client/client.go#L47

state/execution.go
```diff
@@ -54,30 +53,25 @@ func execBlockOnProxyApp(txEventPublisher types.TxEventPublisher, proxyAppConn p
...
+				Height: uint64(block.Height),
+				Index:  uint32(txIndex),
 				Tx:     types.Tx(req.GetDeliverTx().Tx),
-				Data:   txResult.Data,
-				Code:   txResult.Code,
-				Log:    txResult.Log,
-				Error:  txError,
-			}
-			txEventPublisher.PublishEventTx(event)
+				Result: *txResult,
```
melekes: I'd prefer copying here ...

state/execution.go
```diff
@@ -54,30 +53,25 @@ func execBlockOnProxyApp(txEventPublisher types.TxEventPublisher, proxyAppConn p
...
+			}})
+
+			abciResponses.DeliverTx[txIndex] = txResult
```
melekes: I think we can remove DeliverTx field from `abciResponses`, no?

state/state_test.go
```diff
@@ -78,9 +78,9 @@ func TestABCIResponsesSaveLoad(t *testing.T) {
 	// build mock responses
 	block := makeBlock(2, state)
 	abciResponses := NewABCIResponses(block)
-	abciResponses.DeliverTx[0] = &abci.ResponseDeliverTx{Data: []byte("foo")}
-	abciResponses.DeliverTx[1] = &abci.ResponseDeliverTx{Data: []byte("bar"), Log: "ok"}
-	abciResponses.EndBlock = abci.ResponseEndBlock{Diffs: []*abci.Validator{
+	abciResponses.DeliverTx[0] = &abci.ResponseDeliverTx{Data: []byte("foo"), Tags: []*abci.KVPair{}}
```
melekes: what if we do not simply include Tags field?

state/txindex/indexer_service.go
```diff
@@ -0,0 +1,48 @@
...
+		for event := range ch {
+			// TODO: may be not perfomant to write one event at a time
+			txResult := event.(types.TMEventData).Unwrap().(types.EventDataTx).TxResult
+			is.idr.Index(&txResult)
+		}
+	}()
+	return nil
+}
+
+func (is *IndexerService) OnStop() {
```
melekes: missing comment

state/txindex/kv/kv.go
```diff
@@ -46,13 +59,321 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+func lookForHeight(conditions []query.Condition) (height uint64, index int) {
+	for i, c := range conditions {
+		if c.Tag == types.TxHeightKey {
+			return uint64(c.Operand.(int64)), i
+		}
+	}
+	return 0, -1
+}
+
+type queryRanges map[string]queryRange
```
melekes: missing comment

state/txindex/kv/kv.go
```diff
@@ -46,13 +59,321 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+	return txi.AddBatch(batch)
+}
+
+// Search performs a search using the given query. It breaks the query into
+// conditions (like "tx.height > 5"). For each condition, it queries the DB
+// index. One special use cases here: (1) if "tx.hash" is found, it returns tx
+// result for it (2) for range queries it is better for the client to provide
+// both lower and upper bounds, so we are not performing a full scan. Results
+// from querying indexes are then intersected and returned to the caller.
+func (txi *TxIndex) Search(q *query.Query) ([]*types.TxResult, error) {
```
melekes: maybe this should be an interface?

```
type Condition interface {
  Key() string
  Operator string
  Value() interface{}
}

type Query interface {
  Conditions() []Condition
}
```

state/txindex/kv/kv.go
```diff
@@ -46,13 +59,321 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+
+	// extract ranges
+	// if both upper and lower bounds exist, it's better to get them in order not
+	// no iterate over kvs that are not within range.
+	ranges, rangeIndexes := lookForRanges(conditions)
+	if len(ranges) > 0 {
+		skipIndexes = append(skipIndexes, rangeIndexes...)
+
+		for _, r := range ranges {
+			if !hashesInitialized {
```
melekes: this messes the whole code

state/txindex/kv/kv.go
```diff
@@ -46,13 +59,321 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+
+func (txi *TxIndex) match(c query.Condition, startKey []byte) (hashes [][]byte) {
+	if c.Op == query.OpEqual {
+		it := txi.store.IteratorPrefix(startKey)
+		defer it.Release()
+		for it.Next() {
+			hashes = append(hashes, it.Value())
+		}
+	} else if c.Op == query.OpContains {
+		// XXX: doing full scan because startKey does not apply here
```
melekes: @ebuchman do you still don't understand why we need to do full scan here?

```
like startKey = "account.owner=an"
search query = "accoutn.owner CONSISTS an"
```

we can't iterate with prefix `account.owner=an` because we might miss keys like `account.owner=Ulan`

state/txindex/kv/kv.go
```diff
@@ -46,13 +59,321 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+		return []byte(fmt.Sprintf("%s/%v/%d/%d", tag.Key, tag.ValueInt, result.Height, result.Index))
+	// case abci.KVPair_TIME:
+	// 	return []byte(fmt.Sprintf("%s/%d/%d/%d", tag.Key, tag.ValueTime.Unix(), result.Height, result.Index))
+	default:
+		panic(fmt.Sprintf("Undefined value type: %v", tag.ValueType))
+	}
+}
+
+///////////////////////////////////////////////////////////////////////////////
+// Utils
```
melekes: these can go into `tmlibs`

config/config.go
ebuchman: if i wanted to index txs the hit my account I might do `account=deadbeef` but if I want to index for all accounts, would I just do `account` ?

state/txindex/kv/kv.go
ebuchman: Seems like a nice idea!

state/txindex/kv/kv.go
```diff
@@ -299,43 +312,20 @@ func startKeyForRange(r queryRange, height uint64, heightSpecified bool) []byte
...
-	return
+func isTagKey(key []byte) bool {
+	return strings.Count(string(key), tagKeySeparator) == 3
 }
 
 func extractValueFromKey(key []byte) string {
-	s := string(key)
-	parts := strings.SplitN(s, "/", 3)
+	parts := strings.SplitN(string(key), tagKeySeparator, 3)
 	return parts[1]
```
ebuchman: I see that this is safe but would be helpful to code it defensively or clarify why

state/execution.go
ebuchman: I believe we still need them for replay if we crash after calling abciApp.Commit() but before saving the state.

https://github.com/tendermint/tendermint/blob/master/consensus/replay.go#L381

state/state_test.go
ebuchman: I think it should be ok to be left empty. isnt it ?

config/config.go
melekes: oh boy :) we can do pattern-matching like "account.*". I just wanted to get smth simple done. If I am not mistaken, the func we have is enough for UI team.

state/execution.go
melekes: oh, right

state/state_test.go
melekes: it is

melekes: hmmm.. this breaks TestABCIResponsesSaveLoad

```
--- FAIL: TestABCIResponsesSaveLoad (0.00s)
        Error Trace:    state_test.go:93
        Error:          Not equal:
                        expected: &types.ResponseDeliverTx{Code:0, Data:[]uint8{0x66, 0x6f, 0x6f}, Log:"", Tags:[]*types.KVPair(nil)}
                        actual: &types.ResponseDeliverTx{Code:0, Data:[]uint8{0x66, 0x6f, 0x6f}, Log:"", Tags:[]*types.KVPair{}}
```

config/config.go
ebuchman: Is there a way to index everything by all tags ? Otherwise I don't think we have enough - ie. we need to be able to run nodes that index on behalf of all accounts, so an LCD can connect and query for all historical transactions related to its account

melekes: > Is there a way to index everything by all tags ?

no, but we can add `*` - a wildcard

rpc/client/rpc_test.go
```diff
@@ -294,3 +295,43 @@ func TestTx(t *testing.T) {
 		}
 	}
 }
+
+func TestTxSearch(t *testing.T) {
```
ebuchman: Sweet. We should probably modify the dummy to return a tag so we can test that here too!

rpc/core/types/responses.go
```diff
@@ -104,18 +104,18 @@ type ResultBroadcastTx struct {
...
 type ResultTx struct {
-	Height   int           `json:"height"`
-	Index    int           `json:"index"`
-	TxResult abci.Result   `json:"tx_result"`
-	Tx       types.Tx      `json:"tx"`
-	Proof    types.TxProof `json:"proof,omitempty"`
+	Height   uint64                 `json:"height"`
+	Index    uint32                 `json:"index"`
+	TxResult abci.ResponseDeliverTx `json:"tx_result"`
+	Tx       types.Tx               `json:"tx"`
```
ebuchman: just a note that this is rendered in base64 since it doesn't use data.Bytes. Will probably annoy someone some day ...

state/txindex/kv/kv.go
```diff
@@ -46,13 +59,322 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+	// 	return []byte(fmt.Sprintf("%s/%d/%d/%d", tag.Key, tag.ValueTime.Unix(), result.Height, result.Index))
+	default:
+		panic(fmt.Sprintf("Undefined value type: %v", tag.ValueType))
+	}
+}
+
+///////////////////////////////////////////////////////////////////////////////
+// Utils
+
+func intersect(as, bs [][]byte) [][]byte {
```
ebuchman: are these slices guaranteed to be sets (ie. no duplicated elements) ? else we could get duplicates in result. Maybe we can just break from the inner loop once we find a match.

state/txindex/kv/kv.go
ebuchman: This looks much nicer. Can you clarify your comment? Does this still `mess the whole code` ? :P

melekes: it is still not as elegant as it could? be, but I guess it's fine

rpc/core/types/responses.go
melekes: Could you please remind me why are storing tx itself? like we should already have it in the block. Most clients will want the receipt (result), not the original tx anyway. see https://github.com/tendermint/tendermint/issues/905

ebuchman: we do need some way to get the tx we want without fetching the whole block. I like #905. But maybe we should wait until after 0.13 is out to address it.

state/txindex/kv/kv.go
melekes: see 66ad366a4fd19c13ce6cfb28d6f45e64810f2271. IteratorPrefix prevents from returning the same hash. We may have troubles with the impl if we decide to introduce `OR`

config/config.go
ebuchman: I think we need too

ebuchman:
```
=== RUN   TestTx
--- PASS: TestTx (0.11s)
	rpc_test.go:274: client 0, case 0
	rpc_test.go:274: client 0, case 1
	rpc_test.go:274: client 0, case 2
	rpc_test.go:274: client 0, case 3
	rpc_test.go:274: client 0, case 4
	rpc_test.go:274: client 0, case 5
	rpc_test.go:274: client 1, case 0
	rpc_test.go:274: client 1, case 1
	rpc_test.go:274: client 1, case 2
	rpc_test.go:274: client 1, case 3
	rpc_test.go:274: client 1, case 4
	rpc_test.go:274: client 1, case 5
=== RUN   TestTxSearch
--- FAIL: TestTxSearch (0.10s)
	rpc_test.go:312: client 0
	rpc_test.go:341: expected a lot of transactions
FAIL
coverage: 79.2% of statements
exit status 1
FAIL	github.com/tendermint/tendermint/rpc/client	5.263s
make: *** [test_integrations] Error 1
```

melekes:
hmm. looks like I broke something along the way

melekes:
oh, old abci version

config/config.go
```diff
@@ -93,9 +96,6 @@ type BaseConfig struct {
 	// so the app can decide if we should keep the connection or not
 	FilterPeers bool `mapstructure:"filter_peers"` // false
 
-	// What indexer to use for transactions
```
odeke-em: I don't have much context into this PR but why are we removing TxIndex from BaseConfig?

melekes: not removing. moving into its own section

consensus/mempool_test.go
```diff
@@ -188,33 +189,41 @@ func (app *CounterApplication) Info(req abci.RequestInfo) abci.ResponseInfo {
...
+	if txValue != uint64(app.mempoolTxCount) {
+		return abci.ResponseCheckTx{
+			Code: abci.CodeType_BadNonce,
+			Log:  fmt.Sprintf("Invalid nonce. Expected %v, got %v", app.mempoolTxCount, txValue)}
+	}
+	app.mempoolTxCount += 1
+	return abci.ResponseCheckTx{Code: abci.CodeType_OK}
 }
 
-func runTx(tx []byte, countPtr *int) abci.Result {
```
odeke-em: Am also curious as to why we broke this code out of this function then duplicate it?

melekes: abci.Result was removed

odeke-em: We could use though the CodeType, right? In each, on success we set `abci.Codetype_OK`.

rpc/client/mock/abci.go
```diff
@@ -38,7 +38,7 @@ func (a ABCIApp) ABCIQueryWithOptions(path string, data data.Bytes, opts client.
 func (a ABCIApp) BroadcastTxCommit(tx types.Tx) (*ctypes.ResultBroadcastTxCommit, error) {
 	res := ctypes.ResultBroadcastTxCommit{}
 	res.CheckTx = a.App.CheckTx(tx)
-	if !res.CheckTx.IsOK() {
+	if res.CheckTx.IsErr() {
```
odeke-em: Nice!

rpc/client/rpc_test.go
```diff
@@ -294,3 +295,50 @@ func TestTx(t *testing.T) {
...
+
+	anotherTxHash := types.Tx("a different tx").Hash()
+
+	for i, c := range GetClients() {
+		t.Logf("client %d", i)
+
+		// now we query for the tx.
+		// since there's only one tx, we know index=0.
+		results, err := c.TxSearch(fmt.Sprintf("tx.hash='%v'", txHash), true)
+		require.Nil(t, err, "%+v", err)
```
odeke-em: Minor nit:
Won't require.Nil post the error it found, so:
```go
require.Nil(t, err)
```

state/txindex/kv/kv.go
```diff
@@ -46,13 +78,328 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+		i++
+	}
+
+	return results, nil
+}
+
+func lookForHash(conditions []query.Condition) (hash []byte, err error, ok bool) {
+	for _, c := range conditions {
+		if c.Tag == types.TxHashKey {
+			decoded, err := hex.DecodeString(c.Operand.(string))
```
odeke-em: Btw did we mean to reuse the `err` here?

state/txindex/kv/kv.go
odeke-em: How big can these slices get?

I ask because the complexity of this code is currently quadratic `O(n^2)` but could easily be made linear `O(n)` by using a map `map[string]bool` (of which the paradigm `map[string(byte)]` won't incur an allocation) so
```go
aElements := make(map[string]bool)
for i := range as {
   aElements[string(as[i])] = true
}
for j := range bs {
   if _, inA := aElements[string(bs[j])]; inA {
      i = append(i, bs[j])
}
```

odeke-em: And btw if they get really big, this is a potential  DOS attack vector as the time is quadratic in the input size.

ebuchman: Made a note of this here https://github.com/tendermint/tendermint/issues/907

PR #839, opened 2017-11-10
Fix nondeterministic tests failures related to pubsub

melekes:
```
panic: interface conversion: interface {} is nil, not types.TMEventData

goroutine 7690 [running]:
github.com/tendermint/tendermint/consensus.waitForAndValidateBlock.func1(0xc427727620, 0x3)
        /go/src/github.com/tendermint/tendermint/consensus/reactor_test.go:292 +0x62b
created by github.com/tendermint/tendermint/consensus.timeoutWaitGroup
        /go/src/github.com/tendermint/tendermint/consensus/reactor_test.go:349 +0xa4
exit status 2
FAIL    github.com/tendermint/tendermint/consensus      38.614s

```

### DIFF
--- a/consensus/reactor_test.go
+++ b/consensus/reactor_test.go
@@ -288,7 +288,12 @@ func TestReactorWithTimeoutCommit(t *testing.T) {
 
 func waitForAndValidateBlock(t *testing.T, n int, activeVals map[string]struct{}, eventChans []chan interface{}, css []*ConsensusState, txs ...[]byte) {
 	timeoutWaitGroup(t, n, func(wg *sync.WaitGroup, j int) {
-		newBlockI := <-eventChans[j]
+		defer wg.Done()
+
+		newBlockI, ok := <-eventChans[j]
+		if !ok {
+			return
+		}
 		newBlock := newBlockI.(types.TMEventData).Unwrap().(types.EventDataNewBlock).Block
 		t.Logf("Got block height=%v validator=%v", newBlock.Height, j)
 		err := validateBlock(newBlock, activeVals)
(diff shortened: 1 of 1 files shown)


melekes:
```
D[11-10|23:27:27.933] setHasVote(LastCommit)                       validator=3 peerRound=0 height=13 round=0 lastCommit=BA{7:XXXXXXX} index=0
D[11-10|23:27:27.933] setHasVote(LastCommit)                       validator=3 peerRound=0 height=13 round=0 lastCommit=BA{7:XXXXXXX} index=4
D[11-10|23:27:27.940] setHasVote(LastCommit)                       validator=3 peerRound=0 height=13 round=0 lastCommit=BA{7:XXXXXXX} index=6
D[11-10|23:27:27.940] setHasVote(LastCommit)                       validator=3 peerRound=0 height=13 round=0 lastCommit=BA{7:XXXXXXX} index=3
D[11-10|23:27:27.941] setHasVote(LastCommit)                       validator=3 peerRound=0 height=13 round=0 lastCommit=BA{7:XXXXXXX} index=0
```

seems like setHasVote has no effect

ebuchman:
I don't think we can just do that. Do we know why this nil is being sent in the first place?

melekes:
@ebuchman apparently, some waitForBlock goroutines are still alive when `eventHub` being stopped, which closes all the channels, resulting in `eventChans` receiving `nil`. Does this make sense?

ebuchman:
Don't we use the wg to wait for all those go-routines to finish? When would we shutdown before reading the new block value ?

ebuchman:
> seems like setHasVote has no effect

Woops, this is always just printing "LastCommit", so looks like nothings changing, even though it is.

melekes:
> Don't we use the wg to wait for all those go-routines to finish? When would we shutdown before reading the new block value ?

```
	done := make(chan struct{})
	go func() {
		wg.Wait()
		close(done)
	}()
```

yes and no. we wait in a separate goroutine, so by the time we call panic `panic("Timed out waiting for all validators to commit a block")` some goroutines may still be blocked

ebuchman:
I see - so this nil issue should only happen if we reached the timeout? ie. if we solve the timeout, we solve the nil event?

I found that increasing the timeout to a minute worked fine. I used to think this was unreasonable, but given that we're running 7 nodes in one process, spewing debug logs, and running in a docker container in a virtual machine on a shared host, this seems more reasonable.

That said, there seem to be a few other non-deterministic test cases now. Have you had success running `make test_race` on the latest develop in a virtual machine (say with 4 `yes &> /dev/null &` processes running on your host)?

melekes:
> so this nil issue should only happen if we reached the timeout? ie. if we solve the timeout, we solve the nil event?

all correct statements. however, I'd prefer this PR merged since this code is more error-prone.

PR #842, opened 2017-11-13
p2p: invoke only a single OnStop per single OnStart

### LINKED ISSUE #841: p2p/MConnection: make OnStop be invoked once per a single OnStart
This is a minor bug report/hardening request, I was writing some test code for MConnection and accidentally invoked MConnection.OnStop twice and it crashed because the internals' channels were already closed.
I proposed that we ensure that each OnStop can only be invoked per OnStart. Go's sync library exposes a primitive called Once that we can use for this https://golang.org/pkg/sync/#Once.

odeke-em:
Fixes #841

Prevent crashes on invoking successive OnStop's by
using a sync.Once protection, reset per OnStart invocation.

The code isn't that critical or for an exploit but it'll
protect any crashes that come from folks many mconn.OnStop
either by error or by copy and paste

Without the protection the test fails with
```shell
$ go test -v -run=Successive
=== RUN   TestMConnectionNonCrashingSuccessiveOnStop
--- FAIL: TestMConnectionNonCrashingSuccessiveOnStop (0.00s)
panic: close of closed channel [recovered]
	panic: close of closed channel

goroutine 20 [running]:
testing.tRunner.func1(0xc4202000f0)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:734 +0x291
panic(0x1350040, 0x15203f0)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/runtime/panic.go:502 +0x229
github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common.(*ThrottleTimer).Stop(0xc42009db00, 0x0)
	/Users/emmanuelodeke/go/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common/throttle_timer.go:71 +0x41
github.com/tendermint/tendermint/p2p.(*MConnection).OnStop(0xc4202001e0)
	/Users/emmanuelodeke/go/src/github.com/tendermint/tendermint/p2p/connection.go:177 +0x36
github.com/tendermint/tendermint/p2p.TestMConnectionNonCrashingSuccessiveOnStop(0xc4202000f0)
	/Users/emmanuelodeke/go/src/github.com/tendermint/tendermint/p2p/connection_test.go:68 +0x12d
testing.tRunner(0xc4202000f0, 0x14f6020)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:769 +0xc0
created by testing.(*T).Run
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:812 +0x2a5
exit status 2
FAIL	github.com/tendermint/tendermint/p2p	0.013s
```

### DIFF
--- a/p2p/connection.go
+++ b/p2p/connection.go
@@ -7,6 +7,7 @@ import (
 	"math"
 	"net"
 	"runtime/debug"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -80,6 +81,7 @@ type MConnection struct {
 	errored     uint32
 	config      *MConnConfig
 
+	onStopOnce   sync.Once
 	quit         chan struct{}
 	flushTimer   *cmn.ThrottleTimer // flush writes as necessary but throttled.
 	pingTimer    *cmn.RepeatTimer   // send pings periodically
@@ -162,6 +164,7 @@ func NewMConnectionWithConfig(conn net.Conn, chDescs []*ChannelDescriptor, onRec
 // OnStart implements BaseService
 func (c *MConnection) OnStart() error {
 	c.BaseService.OnStart()
+	c.onStopOnce = sync.Once{}
 	c.quit = make(chan struct{})
 	c.flushTimer = cmn.NewThrottleTimer("flush", c.config.flushThrottle)
 	c.pingTimer = cmn.NewRepeatTimer("ping", pingTimeout)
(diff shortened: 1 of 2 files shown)


p2p/connection.go
```diff
@@ -173,19 +176,21 @@ func (c *MConnection) OnStart() error {
 
 // OnStop implements BaseService
 func (c *MConnection) OnStop() {
-	c.BaseService.OnStop()
```
melekes: this line can be safely removed

melekes:
single start/stop should be guaranteed by https://github.com/tendermint/tmlibs/blob/master/common/service.go#L125. We need to figure out why it's not the case here

p2p/connection.go
odeke-em: @melekes could we perhaps do that in another PR? I'd like to keep this one focused on just OnStop being invoked once :)

odeke-em:
@melekes it fails on `c.flushTimer.Stop()` not on `BaseServer.Stop()`, sorry my output was skewed because I had edits in that file, so it seems like that code is working as it is.

odeke-em:
@melekes I've updated the commit message with the proper stack trace.

melekes:
I don't understand. You are not supposed to call OnStop/OnStart directly. Instead one should call Stop/Start methods (which are coming from BaseService).

odeke-em:
@melekes I am currently working on fuzzing and testing `p2p`, that's why I was experimenting around with that method.

ebuchman:
Thanks @odeke-em, but I don't think we want to do this, unless we plan to do it for everything that uses cmn.BaseService - but the semantics of BaseService are that you should never call OnStop, only Stop, and it should only be effective once, so we don't really need this extra protection. So I will close this - please open an issue if there is strong disagreement

PR #843, opened 2017-11-13
WIP: begin parallel refactoring: go-wire Write methods and MConnection

petabytestorage:
In conjunction with https://github.com/tendermint/go-wire/pull/47 to initially reduce `io.Writer` dependence and eventually refactor, split, improve and replace about half of `go-wire` and `MConnection`. This PR just does the first smallest step in that direction and is not ready to be merged yet but may become viable for merging once more reflection and tests are in place. This PR introduces the `legacy` namespace prefix to identify and mark the areas to adjust for refactoring while minimizing chance of disruption. Once we have marked all the areas we want to refactor then we can mechanically change each case to our ideal interface in a systematic, consistent, and hopefully more accurate way as a team.


ebuchman:
I think this needs a glide update with go-wire on develop

Issue #851, opened 2017-11-15
panic: runtime error: index out of range


melekes:
```
--- FAIL: TestByzantine (0.10s)
        byzantine_test.go:186: Byzantine: broadcasting conflicting proposals to 0 peers
panic: runtime error: index out of range [recovered]
        panic: runtime error: index out of range

goroutine 23 [running]:
testing.tRunner.func1(0xc420210690)
        /usr/local/go/src/testing/testing.go:711 +0x5d9
panic(0xd12ea0, 0x1344970)
        /usr/local/go/src/runtime/panic.go:491 +0x2a2
github.com/tendermint/tendermint/consensus.TestByzantine(0xc420210690)
        /go/src/github.com/tendermint/tendermint/consensus/byzantine_test.go:117 +0x1f5a
testing.tRunner(0xc420210690, 0xf1d0a8)
        /usr/local/go/src/testing/testing.go:746 +0x16d
created by testing.(*T).Run
        /usr/local/go/src/testing/testing.go:789 +0x569
exit status 2
FAIL    github.com/tendermint/tendermint/consensus      0.153s
```

melekes:
Could it be because of the i+1 change?

```
func MakeConnectedSwitches(cfg *cfg.P2PConfig, n int, initSwitch func(int, *Switch) *Switch, connect func([]*Switch, int, int)) []*Switch {
...
	for i := 0; i < n; i++ {
		for j := i + 1; j < n; j++ {
			connect(switches, i, j)
		}
	}
```

melekes:
Introduced by https://github.com/tendermint/tendermint/commit/49faa79bdce5663894b3febbf4955fb1d172df04

odeke-em:
@melekes huh? 49faa79 doesn't have any change related to that `i + 1` change.

melekes:
```
panic: Error setting deadline while encrypting connection: set pipe: deadline not supported

goroutine 44 [running]:
github.com/tendermint/tendermint/p2p.Connect2Switches.func1(0xc42000b400, 0x12c8400, 0xc42011b030, 0xc42006b200)
        /home/vagrant/go/src/github.com/tendermint/tendermint/p2p/switch.go:523 +0xd2
created by github.com/tendermint/tendermint/p2p.Connect2Switches
        /home/vagrant/go/src/github.com/tendermint/tendermint/p2p/switch.go:526 +0x454
exit status 2
FAIL    github.com/tendermint/tendermint/consensus      0.118s
```

[...]

PR #852, opened 2017-11-15
p2p: use fake net.Pipe since only >=Go1.10 implements SetDeadline

odeke-em:
Fixes https://github.com/tendermint/tendermint/issues/851

Go1.9 and below's net.Pipe did not implement the SetDeadline
method so after commit
https://github.com/golang/go/commit/e2dd8ca946be884bb877e074a21727f1a685a706
this problem was exposed since now we check for errors.

To counter this problem, implement a simple composition for
net.Conn that always returns nil on SetDeadline instead of
tripping out.

Added build tags so that anyone using go1.10 when it is released
will be able to automatically use net.Pipe's net.Conns

### DIFF
--- a/p2p/conn_go110.go
+++ b/p2p/conn_go110.go
@@ -1,0 +1,15 @@
+// +build go1.10
+
+package p2p
+
+// Go1.10 has a proper net.Conn implementation that
+// has the SetDeadline method implemented as per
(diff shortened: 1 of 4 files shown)


odeke-em:
Thanks @melekes for reporting the problem and thanks for the review and merge @ebuchman.

PR #858, opened 2017-11-16
Fix a bug about p2p connection problem

gguoss:
p2p pex enable pattern to connet not enable pattern.

### DIFF
--- a/p2p/connection.go
+++ b/p2p/connection.go
@@ -459,7 +459,11 @@ FOR_LOOP:
 			}
 			channel, ok := c.channelsIdx[pkt.ChannelID]
 			if !ok || channel == nil {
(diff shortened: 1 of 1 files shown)


p2p/connection.go
```diff
@@ -459,7 +459,11 @@ FOR_LOOP:
 			}
 			channel, ok := c.channelsIdx[pkt.ChannelID]
 			if !ok || channel == nil {
-				cmn.PanicQ(cmn.Fmt("Unknown channel %X", pkt.ChannelID))
+				if pkt.ChannelID == PexChannel {
```
melekes: why is PexChannel not a part of `c.channelsIdx`?

ebuchman:
Not sure I understand - this Panic was already removed on develop: https://github.com/tendermint/tendermint/blob/develop/p2p/connection.go#L461

PR #859, opened 2017-11-16
Fix/addrbook

ebuchman:
Started looking at the AddrBook. It's more complicated than it needs to be and at least a little broken.

### DIFF
--- a/p2p/addrbook.go
+++ b/p2p/addrbook.go
@@ -210,37 +210,18 @@ func (a *AddrBook) PickAddress(newBias int) *NetAddress {
 	oldCorrelation := math.Sqrt(float64(a.nOld)) * (100.0 - float64(newBias))
 	newCorrelation := math.Sqrt(float64(a.nNew)) * float64(newBias)
 
+	// pick a random peer from a random bucket
+	var bucket map[string]*knownAddress
 	pickFromOldBucket := (newCorrelation+oldCorrelation)*a.rand.Float64() < oldCorrelation
-	if pickFromOldBucket {
-		var bucket map[string]*knownAddress
-		for len(bucket) == 0 {
+	for len(bucket) == 0 {
+		if pickFromOldBucket {
 			bucket = a.bucketsOld[a.rand.Intn(len(a.bucketsOld))]
-		}
-		// pick a random ka from bucket.
-		randIndex := a.rand.Intn(len(bucket))
-		for _, ka := range bucket {
-			if randIndex == 0 {
-				return ka.Addr
-			}
-			randIndex--
-		}
-		cmn.PanicSanity("Should not happen")
-	} else {
-		var bucket map[string]*knownAddress = nil
-		for len(bucket) == 0 {
+		} else {
 			bucket = a.bucketsNew[a.rand.Intn(len(a.bucketsNew))]
 		}
-		// pick a random ka from bucket.
-		randIndex := a.rand.Intn(len(bucket))
-		for _, ka := range bucket {
-			if randIndex == 0 {
-				return ka.Addr
-			}
-			randIndex--
-		}
-		cmn.PanicSanity("Should not happen")
 	}
-	return nil
+	randIndex := a.rand.Intn(len(bucket))
+	return bucket[randIndex].Addr
 }
 
 // MarkGood marks the peer as good and moves it into an "old" bucket.
--- a/p2p/addrbook.go
+++ b/p2p/addrbook.go
@@ -89,8 +89,8 @@ type AddrBook struct {
 	rand       *rand.Rand
 	ourAddrs   map[string]*NetAddress
 	addrLookup map[string]*knownAddress // new & old
-	addrNew    []map[string]*knownAddress
-	addrOld    []map[string]*knownAddress
+	bucketsOld []map[string]*knownAddress
+	bucketsNew []map[string]*knownAddress
 	nOld       int
 	nNew       int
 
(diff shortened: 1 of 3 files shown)


p2p/addrbook.go
```diff
@@ -79,18 +80,22 @@ const (
...
-	ourAddrs          map[string]*NetAddress
-	addrLookup        map[string]*knownAddress // new & old
-	addrNew           []map[string]*knownAddress
-	addrOld           []map[string]*knownAddress
-	wg                sync.WaitGroup
-	nOld              int
-	nNew              int
+
+	// accessed concurrently
+	mtx        sync.Mutex
```
odeke-em: Nice, great for readability moving this down here next to the variables it protects!

p2p/addrbook.go
```diff
@@ -201,40 +210,34 @@ func (a *AddrBook) PickAddress(newBias int) *NetAddress {
...
-		// pick random New bucket.
-		var bucket map[string]*knownAddress = nil
-		for len(bucket) == 0 {
-			bucket = a.addrNew[a.rand.Intn(len(a.addrNew))]
+	// pick a random peer from a random bucket
+	var bucket map[string]*knownAddress
+	pickFromOldBucket := (newCorrelation+oldCorrelation)*a.rand.Float64() < oldCorrelation
+	if (pickFromOldBucket && a.nOld == 0) ||
+		(!pickFromOldBucket && a.nNew == 0) {
+		return nil
```
odeke-em: Isn't this a behavioral change? The previous code never returned a nil address. It panicked in the worst case.

p2p/pex_reactor.go
```diff
@@ -240,43 +240,26 @@ func (r *PEXReactor) ensurePeers() {
...
+	for i := 0; i < maxAttempts && len(toDial) < numToDial; i++ {
+		try := r.book.PickAddress(newBias)
+		if try == nil {
+			continue
 		}
-		if picked == nil {
+		_, alreadySelected := toDial[try.IP.String()]
+		alreadyDialing := r.Switch.IsDialing(try)
+		alreadyConnected := r.Switch.Peers().Has(try.IP.String())
+		if alreadySelected || alreadyDialing || alreadyConnected {
```
odeke-em: Minor nit but we could avoid the extra dialing by say
```go
if _, selected := toDial[try.IP.String()]; selected {
       continue
}
if _, dialling := r.Switch.IsDialing(try); dialling {
       continue
}
if _, connected := r.Switch.Peers().Has(try.IP.String()); connected {
       continue
}
```

p2p/pex_reactor.go
```diff
@@ -240,43 +240,26 @@ func (r *PEXReactor) ensurePeers() {
 		return
 	}
 
-	toDial := make(map[string]*NetAddress)
+	// bias to prefer more vetted peers when we have fewer connections.
+	// not perfect, but somewhate ensures that we prioritize connecting to more-vetted
+	newBias := cmn.MinInt(numOutPeers, 8)*10 + 10
```
odeke-em: How come we hoisted this out of the loop? Won't that mean that we'll be using the same bias for every pick? This removes the extra randomness per loop iteration.

p2p/addrbook_test.go
```diff
@@ -106,6 +142,10 @@ func TestAddrBookPromoteToOld(t *testing.T) {
 	if len(selection) > book.Size() {
 		t.Errorf("selection could not be bigger than the book")
 	}
+
+	if book.Size() != 100 {
```
odeke-em: Minor nit:
```go
assert.Equal(book.Size(), 100, "expecting book size to be 100")
```

p2p/addrbook.go
ebuchman: This fixes a bug! Used to be we could get stuck in the infinite loop otherwise. And this function returns nil if the book is empty already, so this is another case when we return nil. I'll update the comment - thanks for the keen eye.

p2p/pex_reactor.go
ebuchman: Unless I'm missing something, it does not appear that `newBias` is influenced by anything that changes with each iteration of the loop - ie. it depends only on `numOutPeers` which is already fixed.

p2p/pex_reactor.go
ebuchman: avoid the extra dialing? this looks nice but it's just replacing the extra variable declarations with more instances of the conditional continue, right?

p2p/addrbook.go
```diff
@@ -191,7 +191,12 @@ func (a *AddrBook) size() int {
 	return a.nNew + a.nOld
 }
 
-// PickAddress picks an address to connect to with new/old bias.
+// PickAddress picks an address to connect to.
+// The address is picked randomly from an old or new bucket according
+// to the newBias argument, which must be between [0, 100] (or else is truncated to that range)
+// and determines how biased we are to pick an address from a new bucket.
+// PickAddress returns nil if the AddrBook is empty or if we try to pick
+// from an empty bucket.
```
odeke-em: Nice updates!

p2p/addrbook.go
odeke-em: Awesome, dude the new comments are great!

p2p/pex_reactor.go
odeke-em: My bad, for lack of a proper word I meant to say prevent the unnecessary calls for `IsDialing` and `Peers().Has` in the best case -- if `toDial` matches, no need to invoke `IsDialing`, same thing for if `IsDialing` matches, no need to invoke `Peers().Has`

p2p/pex_reactor.go
odeke-em: Ahaha, my bad. For some reason I looked at that formula and my eyes made me think it was rechoosing the bias. Thank you for the clarification, please disregard my earlier comment.

Issue #850, opened 2017-11-15
p2p: replace *[32]byte key comparison of == check with crypto.PubKey.Equals


odeke-em:
While auditing the code in p2p/secret_connection.go

I noticed that in genNonces that to check for the satisfication of equalization of `locIsLo` whe compared tow *[32]byte values using `==`
in https://github.com/tendermint/tendermint/blob/30f675aafa4cbbb86d887de842a7a93403227548/p2p/secret_connection.go#L70-L71

This looks wrong because those are keys that were generated uniquely and hopefully ephermerally as at
https://github.com/tendermint/tendermint/blob/30f675aafa4cbbb86d887de842a7a93403227548/p2p/secret_connection.go#L51

and

https://github.com/tendermint/tendermint/blob/30f675aafa4cbbb86d887de842a7a93403227548/p2p/secret_connection.go#L54

yet we compare them with `==` in
https://github.com/tendermint/tendermint/blob/30f675aafa4cbbb86d887de842a7a93403227548/p2p/secret_connection.go#L71

Here is an example of what could go terribly wrong
https://play.golang.org/p/rjPMy3Wv0k
```go
package main

import (
	"bytes"
	"fmt"
)

func main() {
	bs1 := [32]byte{0x01, 0x02}
	bs2 := [32]byte{0x01, 0x02}

	bPtr1 := &bs1
	bPtr2 := &bs2
	fmt.Printf("false equals:  %v\n", bPtr1 == bPtr2)
	fmt.Printf("proper equals: %v\n", (bPtr1 == nil && bPtr2 == nil) || (bPtr1 != nil && bPtr2 != nil && bytes.Equal((*bPtr1)[:], (*bPtr2)[:])))
}
```
which gives
```shell
false equals:  false
proper equals: true
```

odeke-em:
False positive (#fakeNews from me) as I read the code without the proper context, that code just checks that
the declared smaller lexicographically sorted public key is the same local ephemeral key.

There is nothing wrong with that code.

PR #876, opened 2017-11-19
p2p: use bytes.Equal for key comparison

odeke-em:
Updates https://github.com/tendermint/tendermint/issues/850
    
My security alarms falsely blarred when I skimmed and noticed
keys being compared with `==`, without the proper context
so I mistakenly filed an issue, yet the purpose of that
comparison was to check if the local ephemeral public key
was just the least, sorted lexicographically.
    
Anyways, let's use the proper bytes.Equal check, to save future labor.


odeke-em:
Yes, SGTM, the lack of that is what initially triggered my alarms.

odeke-em:
Alright, thank you @ebuchman for the review, I've force pushed the change since it was a tiny one, PTAL.

Issue #875, opened 2017-11-19
p2p: Switch.DialSeeds should try to generate a random source before dialling seeds


odeke-em:
I noticed while auditing the code in p2p that DialSeeds will always dial to peers deterministically every single time
https://github.com/tendermint/tendermint/blob/128e2a1d9e88eeb0aa7645973940b23eb7a14803/p2p/switch.go#L318-L326

### Repro
Playground link https://play.golang.org/p/vfcBP2M_Ap
```go
package main

import (
	"fmt"
	"math/rand"
	"time"
)

func dialSeeds(seeds []chan int) {
	perm := rand.Perm(len(seeds))
	for i := 0; i < len(perm); i++ {
		go func(i int) {
			time.Sleep(time.Duration(rand.Int63n(3)) * time.Millisecond)
			j := perm[i]
			seeds[j] <- i
		}(i)
	}
}

func main() {
	seeds := make([]chan int, 10)
	for i := 0; i < len(seeds); i++ {
		seeds[i] = make(chan int, 1)
	}
	dialSeeds(seeds)

	for i := 0; i < len(seeds); i++ {
		si := seeds[i]
		fmt.Printf("#%d: %d\n", i, <-si)
	}
}
```

Will always give
```shell
$ go run main.go 
#0: 5
#1: 7
#2: 2
#3: 6
#4: 1
#5: 9
#6: 3
#7: 8
#8: 4
#9: 0
```
and the danger here is the determinism of which peer/seed will be dialled in whatever sequence, and if any outside attackers has knowledge of when we are dialling.

### Proposed fix
Every invocation of DialSeed should create a new source either seeded with the current time(for simplicity and this time will change) or a mix of `crypto/rand` for the cryptographically securely PRNG and the current time
so something like
```go
rng := rand.New(rand.NewSource(time.Now().Unix()))
perm := rng.Perm(len(netAddrs))
```

PR #877, opened 2017-11-19
p2p: make Switch.DialSeeds use a new PRNG per call

odeke-em:
Fixes https://github.com/tendermint/tendermint/issues/875

Ensure that every DialSeeds call uses a new PRNG seeded from
tendermint/tmlibs/common.RandInt which internally uses
crypto/rand to seed its source.

### DIFF
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -295,7 +295,6 @@ func (sw *Switch) startInitPeer(peer *peer) {
 
 // DialSeeds dials a list of seeds asynchronously in random order.
 func (sw *Switch) DialSeeds(addrBook *AddrBook, seeds []string) error {
-
 	netAddrs, err := NewNetAddressStrings(seeds)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


p2p/switch.go
```diff
@@ -315,11 +314,15 @@ func (sw *Switch) DialSeeds(addrBook *AddrBook, seeds []string) error {
 		addrBook.Save()
 	}
 
+	// Ensure we have a completely undeterministic PRNG.
+	// See Issue https://github.com/tendermint/tendermint/issues/875.
```
ebuchman: Maybe instead of commenting the link to the issue lets just note here that cmn.RandInt64() draws from a seed that's initialized with OS entropy on process start.

odeke-em: Nice suggestion. I've literally copied it verbatim #innerMelaniaTrump #amFinnaPlagiarizeThatSuggestion. I'll force push in a second.

Issue #881, opened 2017-11-20
p2p: race condition in trust metric


ebuchman:
cropped trace:

```
WARNING: DATA RACE
Read at 0x00c42006cd20 by goroutine 8:
  github.com/tendermint/tendermint/p2p/trust.(*TrustMetricStore).saveToDB()
      github.com/tendermint/tendermint/p2p/trust/_test/_obj_test/trustmetric.go:190 +0x31d
  github.com/tendermint/tendermint/p2p/trust.TestTrustMetricStoreSaveLoad()
      /go/src/github.com/tendermint/tendermint/p2p/trust/trustmetric_test.go:61 +0x5c5

Previous write at 0x00c42006cd20 by goroutine 29:
  github.com/tendermint/tendermint/p2p/trust.(*TrustMetric).processRequests()
      github.com/tendermint/tendermint/p2p/trust/_test/_obj_test/trustmetric.go:548 +0x64c

```

Looks like the save routine is racing the process routine. Specifically it's about accessing `tm.history`.

ebuchman:
@caffix

PR #882, opened 2017-11-20
fixed race condition reported in issue #881

caffix:
https://github.com/tendermint/tendermint/issues/881


p2p/trust/trustmetric.go
```diff
@@ -144,6 +144,7 @@ func (tms *TrustMetricStore) loadFromDB() bool {
 	for key, p := range peers {
 		tm := NewMetricWithConfig(tms.config)
 
+		tm.mtx.Lock()
```
odeke-em: Could we perhaps extract the code below into a helper method such as:
```go
func (tms *TrustMetricStore) recalculate(p *peerHistoryJSON) {
   tms.mtx.Lock()
   defer tms.mtx.Unlock()

   // Restore the number of time inteverals we have previously tracked
   ...
```
and then in that loop
```go
for key, p := range peers {
     tm := NewMetricWithConfig(tms.config)
     tm.recalculate(&p)

     // Save the updated peer
     peers[key] = p
```
or someway of already having the pointer to p.

I am making this suggestion to ease with the sparse locking.

p2p/trust/trustmetric.go
```diff
@@ -310,20 +300,44 @@ func (tm *TrustMetric) Stop() {
...
 // GoodEvents indicates that a desirable event(s) took place
 func (tm *TrustMetric) GoodEvents(num int) {
-	tm.update <- &updateBadGood{IsBad: false, Add: num}
+	tm.mtx.Lock()
+	defer tm.mtx.Unlock()
+
+	// Check if this is the first experience with
+	// what we are tracking since being paused
+	if tm.paused {
+		tm.good = 0
```
odeke-em: This code is duplicated above, please extract it to a method if possible, otherwise, no biggie :)

odeke-em:
@caffix @ebuchman might anyone have a repro for the data race https://github.com/tendermint/tendermint/issues/881#issue-275492818 that we can extract to a test to ensure we never regress?

caffix:
@odeke-em, I added the suggested functions/methods

p2p/trust/trustmetric.go
```diff
@@ -181,11 +189,13 @@ func (tms *TrustMetricStore) saveToDB() {
 	peers := make(map[string]peerHistoryJSON, 0)
 
 	for key, tm := range tms.peerMetrics {
+		tm.mtx.Lock()
```
melekes: isn't it better to grab lock outside the for loop once?

caffix: the lock is specific to each metric tm refers to during the loop iterations

p2p/trust/trustmetric.go
```diff
@@ -124,6 +124,37 @@ type peerHistoryJSON struct {
 	History      []float64 `json:"history"`
 }
 
+// Loads the history data for a single peer and takes care of trust metric locking
+func reinstantiateMetric(tm *TrustMetric, ph peerHistoryJSON) {
```
ebuchman: This might be a bit cleaner as an `Init` method on `TrustMetric`. 
Also, since it's only called once and only right after creation, it shouldn't actually need the lock. Comments should reflect this.

p2p/trust/trustmetric.go
ebuchman: Let's try to keep locks at the top of public methods. For instance here we can have `tm.HistoryJSON()` that handles the locking and returns the peerHistoryJSON

p2p/trust/trustmetric.go
```diff
@@ -381,9 +384,7 @@ func NewMetricWithConfig(tmc TrustMetricConfig) *TrustMetric {
 	tm.historyMaxSize = intervalToHistoryOffset(tm.maxIntervals) + 1
 	// This metric has a perfect history so far
 	tm.historyValue = 1.0
-	// Setup the channels
-	tm.update = make(chan *updateBadGood, defaultUpdateChanCapacity)
```
ebuchman: we can get rid of the corresponding constants defined at the top of the file

p2p/trust/trustmetric.go
```diff
@@ -516,24 +530,8 @@ func (tm *TrustMetric) processRequests() {
...
-
-			if bg.IsBad {
-				tm.bad += float64(bg.Add)
-			} else {
-				tm.good += float64(bg.Add)
-			}
-		case rtv := <-tm.trustValue:
-			rtv.Resp <- tm.calcTrustValue()
 		case <-t.C:
+			tm.mtx.Lock()
```
ebuchman: let's move this all into a public function on `tm` and put the locking in there.

p2p/trust/trustmetric.go
```diff
@@ -563,13 +561,17 @@ loop:
 				tm.good = 0
 				tm.bad = 0
 			}
+			tm.mtx.Unlock()
 		case stop := <-tm.stop:
+			tm.mtx.Lock()
```
ebuchman: we don't need the lock until we want to set pause. so we can check `stop`, and if it's false, call `tm.Pause()` which should handle the locking.

p2p/trust/trustmetric_test.go
```diff
@@ -210,7 +210,10 @@ func TestTrustMetricStopPause(t *testing.T) {
 	// Give the pause some time to take place
 	time.Sleep(10 * time.Millisecond)
 
+	tm.mtx.Lock()
```
ebuchman: Perhaps we can introduce a copy function that locks to create a copy of the `tm` and returns that. then we can just do eg. `first := tm.Copy().numIntervals`

ebuchman:
Also please rebase these changes on latest develop and undo the `trustmetric_test.go` rename -> I renamed it to `trustmetric_test.go_` for now so the tests wont run since they're racy.

p2p/trust/trustmetric.go
caffix: will do

caffix:
Thank you, Ethan. All the changes you requested have been made and committed.

PR #884, opened 2017-11-21
consensus/WAL: benchmark WALDecode across data sizes

odeke-em:


odeke-em:
An optional addition, I figured we could add it while auditing `consensus/wal.go`

consensus/benchmark_test.go
```diff
@@ -0,0 +1,56 @@
...
+	enc.Encode(data)
+
+	encoded := buf.Bytes()
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		buf.Reset()
+		buf.Write(encoded)
+		dec := NewWALDecoder(buf)
+		if _, err := dec.Decode(); err == nil {
```
melekes: this returns an error (`"failed to decode data: %v"`) because you're encoding raw bytes, not TimedWALMessage.

we should probably change `Encode(v interface{})` to `Encode(msg *TimedWALMessage)`

odeke-em: Awesome, thank you for the catch @melekes, I'll update it in a second.

odeke-em:
@melekes I've added the benchmarks to `consensus/wal_test.go` PTAL.

consensus/wal_test.go
```diff
@@ -60,3 +62,60 @@ func TestSearchForEndHeight(t *testing.T) {
 	assert.Equal(t, rs.Height, h+1, cmn.Fmt("wrong height"))
 
 }
+
+func init() {
```
melekes: this will be run not only for tests, right? don't think we want this. is there some other way?

odeke-em: It will only be run for tests, we could even throw in Java code in the `_test.go` and it won't be run until we do `go test`.

However, I'll make it clear with an update :)

odeke-em: I've updated it with a sync.Once protection to initialize interface registration only once.

melekes: Omg, I just asked. You didn't have to :)

odeke-em: Hehe, that's alright, it was a great question. I've fixed it so that in the future there won't be any confusion :)

odeke-em:
Thank you for the review @melekes! @ebuchman the floor is yours.

Issue #890, opened 2017-11-23
mempool: Mempool should have a CloseWAL method


odeke-em:
I have been trying to fuzz `mempool` but all my crashers were just false positives because of resource exhaustion e.g `too many open files`, `timed out` etc.

I noticed that underlying we have a WAL whose type is `tmlibs/autofile.AutoFile` https://github.com/tendermint/tendermint/blob/128e2a1d9e88eeb0aa7645973940b23eb7a14803/mempool/mempool.go#L120-L124
which however, just listens for SIGHUP
https://github.com/tendermint/tmlibs/blob/master/autofile/sighup_watcher.go#L22-L29

which won't invoke a closeAll for all other exits except when you destroy the terminal itself.
I raised the signal listening offline in one of the engineering channels, but that's another discussion.

It would be nice for programs that use the WAL to close it after use. This way normal usage won't leave files open.

melekes:
```
// consensus/state.go#267
func (cs *ConsensusState) OnStop() {
...
	// Make BaseService.Wait() wait until cs.wal.Wait()
	if cs.IsRunning() {
		cs.wal.Wait()
	}
```

`Close` or `Start/Stop` by analogy with other services using autofile (e.g. consensus/state)

odeke-em:
Nice, thanks for the suggestion to instead use `Stop`, consistent with our other APIs. I'll update the title to reflect this.

melekes:
Re signals: I think we should add 

```
signal.Notify(c, syscall.SIGTERM)
signal.Notify(c, syscall.SIGINT)
```

odeke-em:
@melekes 100% agree, syscall.SIGQUIT too :), but could we also mention this in the engineering channel for consensus or actually a Github issue on `tmlibs`?

melekes:
```
The SIGQUIT signal is sent when the user at the controlling terminal presses the quit character, which by default is ^\ (Control-Backslash)
```
https://golang.org/pkg/os/signal/

this is not a standard practice to include SIGQUIT. I think it's better without it. Besides who stops programs like this when we have Ctrl-C

PR #892, opened 2017-11-24
mempool: implement Mempool.CloseWAL

odeke-em:
Fixes https://github.com/tendermint/tendermint/issues/890

Add a Stop metho to Mempool to close the underlying
WAL file and then discard it so that further writes
to it will have no effect.

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -110,6 +110,26 @@ func (mem *Mempool) SetLogger(l log.Logger) {
 	mem.logger = l
 }
 
+// CloseWAL closes and discards the underlying WAL file.
+// Any further writes will not be relayed to disk.
+func (mem *Mempool) CloseWAL() bool {
+	if mem == nil {
+		return false
+	}
+
+	mem.proxyMtx.Lock()
+	defer mem.proxyMtx.Unlock()
+
+	if mem.wal == nil {
+		return false
+	}
+	if err := mem.wal.Close(); err != nil && mem.logger != nil {
+		mem.logger.Error("Mempool.CloseWAL", "err", err)
+	}
+	mem.wal = nil
+	return true
+}
+
 func (mem *Mempool) initWAL() {
 	walDir := mem.config.WalDir()
 	if walDir != "" {
--- a/mempool/mempool_test.go
+++ b/mempool/mempool_test.go
@@ -1,8 +1,13 @@
 package mempool
 
 import (
+	"crypto/md5"
 	"crypto/rand"
 	"encoding/binary"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"testing"
 	"time"
 
@@ -13,6 +18,8 @@ import (
 	cfg "github.com/tendermint/tendermint/config"
 	"github.com/tendermint/tendermint/proxy"
 	"github.com/tendermint/tendermint/types"
+
+	"github.com/stretchr/testify/require"
 )
 
 func newMempoolWithApp(cc proxy.ClientCreator) *Mempool {
@@ -57,7 +64,7 @@ func checkTxs(t *testing.T, mempool *Mempool, count int) types.Txs {
 			t.Error(err)
 		}
 		if err := mempool.CheckTx(txBytes, nil); err != nil {
-			t.Fatal("Error after CheckTx: %v", err)
+			t.Fatalf("Error after CheckTx: %v", err)
 		}
 	}
 	return txs
@@ -127,7 +134,7 @@ func TestSerialReap(t *testing.T) {
 			binary.BigEndian.PutUint64(txBytes, uint64(i))
 			err := mempool.CheckTx(txBytes, nil)
 			if err != nil {
-				t.Fatal("Error after CheckTx: %v", err)
+				t.Fatalf("Error after CheckTx: %v", err)
 			}
 
 			// This will fail because not serial (incrementing)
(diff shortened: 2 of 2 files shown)


mempool/mempool.go
```diff
@@ -110,6 +110,20 @@ func (mem *Mempool) SetLogger(l log.Logger) {
 	mem.logger = l
 }
 
+// Stop closes the underlying WAL file and then discards it.
+// After Stop has been invoked, any further writes will not be relayed to disk.
+func (mem *Mempool) Stop() error {
```
melekes: let's include cmn.BaseService into Mempool and implement OnStart/OnStop. this is a standard across our repos. what do you think?

melekes:
Also, don't see where we actually call Stop beside tests. maybe we should do it inside `node.OnStop`?

mempool/mempool.go
odeke-em: @melekes I personally don't have any opinions about that. I was just requesting for a Stop method as the lack of it is cramping up my fuzzing tests and giving false positives allover the place. However, sure, I can implement that.

odeke-em:
Great suggestion @melekes I've made it implement `common.Service` then force pushed the PR for cohesion of the commit message, PTAL and thank you!

mempool/mempool.go
```diff
@@ -110,6 +115,29 @@ func (mem *Mempool) SetLogger(l log.Logger) {
 	mem.logger = l
 }
 
+func (mem *Mempool) OnStop() {
+	mem.BaseService.OnStop()
```
melekes: can be removed

mempool/mempool.go
```diff
@@ -110,6 +115,29 @@ func (mem *Mempool) SetLogger(l log.Logger) {
...
+	mem.Stop()
+}
+
+// Stop closes the underlying WAL file and then discards it.
+// After Stop has been invoked, any further writes will not be relayed to disk.
+func (mem *Mempool) Stop() bool {
```
melekes: https://github.com/tendermint/tmlibs/blob/master/common/service.go#L124

func body should go into onStop

mempool/mempool.go
```diff
@@ -110,6 +115,29 @@ func (mem *Mempool) SetLogger(l log.Logger) {
...
+		return false
+	}
+
+	mem.proxyMtx.Lock()
+	defer mem.proxyMtx.Unlock()
+
+	if mem.wal == nil {
+		return false
+	}
+	_ = mem.wal.Close()
```
melekes: what about `mem.Wait()`?

mempool/mempool.go
```diff
@@ -110,6 +115,29 @@ func (mem *Mempool) SetLogger(l log.Logger) {
...
 
+func (mem *Mempool) OnStop() {
+	mem.BaseService.OnStop()
+	mem.Stop()
+}
+
+// Stop closes the underlying WAL file and then discards it.
+// After Stop has been invoked, any further writes will not be relayed to disk.
+func (mem *Mempool) Stop() bool {
+	if mem == nil {
```
melekes: this should not be possible

mempool/mempool.go
melekes: why not log the error?

ebuchman:
Hmm interesting. Can you clarify what you were fuzzing and how it led to to these errors? I'm not sure we need all the Service stuff in mempool, but might not be a bad idea to expose a `CloseWAL` method. At least the mempool reactor should be able to graceful shut down and should handle closing the mempool's WAL. 

The mempool WAL isn't really even used for anything, so for fuzzing sake we could also consider a way to overwrite it with a NopWAL or something.

odeke-em:
@ebuchman for sure, I was fuzzing CheckTx. Please see https://github.com/tendermint/tendermint/pull/719/commits/cceae360da1d2f127914be2b7d9ccd99100dc694
in which the outputs report errors such as
https://github.com/tendermint/tendermint/pull/719/commits/cceae360da1d2f127914be2b7d9ccd99100dc694#diff-6500d7d7849caaeeb19a47749e88841f
https://github.com/tendermint/tendermint/blob/cceae360da1d2f127914be2b7d9ccd99100dc694/fuzz/mempool/checktx/crashers/101ccfd2d4835d59acaa84e8e3866c762ed03a08.output#L1

but when ran through the repro individually before I was ready to file a bug report
https://github.com/tendermint/tendermint/blob/cceae360da1d2f127914be2b7d9ccd99100dc694/fuzz/mempool/checktx/repro/main.go

produced no crash.

ebuchman:
Cool - so this is because you were creating a thousand new mempools in one process, and none of them get their WALs closed ?

ebuchman:
Simplest/best thing to do to the Mempool is just to expose a `CloseWAL` that you can call in your fuzzing so we dont leak file descriptors.

If you want to also update the reactor to call mempool.CloseWAL in it's OnStop, that'd be cool too (see other reactors for examples) :)

odeke-em:
Yes in deed, I'll go for CloseWAL, exactly the initial method of attack for the bug :)

mempool/mempool.go
odeke-em: No longer relevant.

odeke-em:
Alright, thank you @melekes and @ebuchman for the review. @ebuchman I've implemented CloseWAL(great idea), PTAL.

mempool/mempool.go
```diff
@@ -110,6 +110,26 @@ func (mem *Mempool) SetLogger(l log.Logger) {
...
+	}
+
+	mem.proxyMtx.Lock()
+	defer mem.proxyMtx.Unlock()
+
+	if mem.wal == nil {
+		return false
+	}
+	if err := mem.wal.Close(); err != nil && mem.logger != nil {
+		mem.logger.Error("Mempool.CloseWAL", "err", err)
```
odeke-em: @melekes great suggestion about logging the error. I've addressed this, PTAL.

PR #898, opened 2017-11-27
Pass loggers in the constructor as a first class citizen

adrianbrink:
This is a breaking change for many of the public APIs.

- [ ] Add Changelog

**Questions**
In [tmlibs](https://github.com/tendermint/tmlibs/blob/master/common/service.go) the service has `SetLogger`. What is the purpose of it? Unless you know which concrete object (which implements service) you are dealing with you cannot set a sensible logger. I'd be in favour of removing `SetLogger` from the Service interface.

Related:
abci needs to be updated to follow the same convention: https://github.com/tendermint/abci/commits/feature/logger
tmlibs needs to be updated: https://github.com/tendermint/tmlibs/pull/82


adrianbrink:
@ebuchman I'll have to clean this up some more, but this is the basic implementation of removing `SetLogger`.

zramsay:
keep or close?

melekes:
I'd close this.

adrianbrink:
I agree.

PR #902, opened 2017-11-28
determinisitic linter

zramsay:
plus a few lints


p2p/trust/trustmetric.go
```diff
@@ -139,7 +139,7 @@ func (tms *TrustMetricStore) loadFromDB() bool {
 		return false
 	}
 
-	peers := make(map[string]MetricHistoryJSON, 0)
+	peers := make(map[string]MetricHistoryJSON)
```
odeke-em: Nice catch, I saw this too this morning before bed but forgot to fix it.

zramsay:
new title, PTAL

Makefile
```diff
@@ -24,24 +24,24 @@ build_race:
 dist:
 	@BUILD_TAGS='$(BUILD_TAGS)' sh -c "'$(CURDIR)/scripts/dist.sh'"
 
-test:
-	@echo "--> Running linter"
-	@make metalinter_test
-	@echo "--> Running go test"
-	@go test $(PACKAGES)
+test: tools get_vendor_deps
```
melekes: no, test should not invoke tools or get_vendor_deps. they are both too time consuming! and dev time is damn expensive. Instead, we should add a section in readme (if we don't have it already):
>>>>>>>>>>>>>

## Development

First two command only needed to be executed once. If you are using Vagrant, just go straight to testing (`make test`).

```
make get_vendor_deps
make tools
make test
```

Makefile
```diff
@@ -24,24 +24,24 @@ build_race:
...
 
 test_race:
-	@echo "--> Running go test --race"
-	@go test -v -race $(PACKAGES)
+	echo "--> Running go test --race"
+	go test -v -race $(PACKAGES)
 
-test_integrations:
-	@bash ./test/test.sh
+test_integrations: test
```
melekes: also no

Makefile
```diff
@@ -24,24 +24,24 @@ build_race:
...
+test: tools get_vendor_deps
+	echo "--> Running linter"
+	make metalinter_test
+	echo "--> Running go test"
+	go test $(PACKAGES)
 
 test_race:
-	@echo "--> Running go test --race"
-	@go test -v -race $(PACKAGES)
+	echo "--> Running go test --race"
```
melekes: why remove `@`. the output will be

echo "--> Running go test --race"
--> Running go test --race

I don't think we want that. If you don't like `@`, then we can remove `echo` statements altogether.

zramsay: i got carried away debugging, will revert

Makefile
zramsay: seemed cleaner than adding a line in the `circle.yml`, but I get it, adds too much time

zramsay:
this is blocked because [this line](https://github.com/tendermint/tendermint/pull/902/files#diff-b67911656ef5d18c4ae36cb6741b7965R80) runs in the CI but can't find `gometalinter.v2`, even though the line *right before it* just installed it. I'm at a loss :(

Makefile
```diff
@@ -75,31 +75,31 @@ update_tools:
 	@go get -u $(GOTOOLS)
 
 tools:
-	@echo "--> Installing tools"
-	@go get $(GOTOOLS)
-	@gometalinter --install
+	echo "--> Installing tools"
```
melekes: `@`

Makefile
```diff
@@ -75,31 +75,31 @@ update_tools:
...
 
 ### Formatting, linting, and vetting
 
 metalinter:
-	@gometalinter --vendor --deadline=600s --enable-all --disable=lll ./...
+	gometalinter.v2 --vendor --deadline=600s --enable-all --disable=lll ./...
 
-metalinter_test:
-	@gometalinter --vendor --deadline=600s --disable-all  \
+metalinter_test: tools get_vendor_deps
```
melekes: don't you think the same https://github.com/tendermint/tendermint/pull/902#discussion_r155934459 applies here. I.e. if we want developers actually running this, we want this to be as fast as it can

zramsay: you're right; i was trying to avoid complicating the CI, but it's actually not bad

Makefile
```diff
@@ -77,29 +78,29 @@ update_tools:
...
 	@echo "--> Installing tools"
 	@go get $(GOTOOLS)
-	@gometalinter --install
+	$(GOPATH)/bin/gometalinter.v2 --install
 
 ### Formatting, linting, and vetting
 
 metalinter:
-	@gometalinter --vendor --deadline=600s --enable-all --disable=lll ./...
+	gometalinter.v2 --vendor --deadline=600s --enable-all --disable=lll ./...
```
melekes: `$(GOPATH)/bin/gometalinter.v2`? to be consistent

Makefile
```diff
@@ -1,7 +1,8 @@
 GOTOOLS = \
 					github.com/mitchellh/gox \
 					github.com/tcnksm/ghr \
-					github.com/alecthomas/gometalinter
+					github.com/Masterminds/glide \
```
melekes: do we need glide here? i think i made it that it is installed in get_vendor_deps if missing

zramsay: i think [this line](https://github.com/tendermint/tendermint/pull/902/files/5e677263ceba10c9349cb949b4800ebd8f9b26cb#diff-b67911656ef5d18c4ae36cb6741b7965R69) flubbed in the CI but we'll see in a moment

melekes:
Great work @zramsay

PR #918, opened 2017-12-02
Abci update

ebuchman:


rpc/core/types/responses.go
```diff
@@ -96,9 +96,9 @@ type ResultDumpConsensusState struct {
 }
 
 type ResultBroadcastTx struct {
-	Code abci.CodeType `json:"code"`
-	Data data.Bytes    `json:"data"`
-	Log  string        `json:"log"`
+	Code uint32     `json:"code"`
```
ebuchman: address the `FIXME`s by using the abci response types here. Also probably good to use ResultBroadcastTxAsync, etc.

types/protobuf.go
```diff
@@ -13,9 +13,9 @@ type tm2pb struct{}
 func (tm2pb) Header(header *Header) *types.Header {
```
ebuchman: we should figure out how to get rid of this file completely now. the types should be identical

mempool/mempool_test.go
```diff
@@ -122,10 +124,10 @@ func TestSerialReap(t *testing.T) {
 	mempool := newMempoolWithApp(cc)
 	appConnCon, _ := cc.NewABCIClient()
 	appConnCon.SetLogger(log.TestingLogger().With("module", "abci-client", "connection", "consensus"))
-	if err := appConnCon.Start(); err != nil {
-		t.Fatalf("Error starting ABCI client: %v", err.Error())
-	}
+	err := appConnCon.Start()
+	assert.Nil(t, err)
```
melekes: better to use `require` here cause it stops the test

Issue #785, opened 2017-10-26
Improvements for WAL


melekes:
- [ ] add version
- [x] remove separator

1) By adding a simple string (eg. "v0.11.0\n") we should be able to correctly handle different versions and avoid breaking stuff, therefore, providing a nice user exp.

Other projects doing that:
- Tarantool https://tarantool.org/doc/1.7/dev_guide/internals_index.html#xlog-snap

2) We don't really need a separator unless we have complex structures (rows, cells like RDBMS have https://www.sqlite.org/fileformat.html). Or do we? Need to think more about it.

melekes:
It could be a good idea to add padding https://github.com/facebook/rocksdb/wiki/Write-Ahead-Log-File-Format#log-file-format for faster writes to match underlying disk block size (usually 32KB or 64KB are used).

melekes:
I still would like to add version, but it requires significant effort since we're using auto#Group, which cleans up old files as wal grows, which means we'll need to write version to every group file (but rotating happens on a level down and not exposed to us). I still don't understand why we need a group of files. We need max 2 files: we write to file (1). If its size becomes >= {N} MB, we close it and start writing to file (2). After some time (or height), we remove old file.

ebuchman:
what if we just use directories to handle the versioning ?

melekes:
we can try, but we'll need to implement traversing of directories than (like read `wal_1.1`, than `wal_1.0`, ...) and gradual removing.

ebuchman:
No we shouldn't have to because version should only switch after a replay is over, so a replay should never have to run across multiple versions

jackzampolin:
Is this still something that needs to be implemented?

melekes:
Yes

PR #931, opened 2017-12-07
[consensus] remove WAL separator

melekes:
Refs #785


melekes:
> And it doesn't require any chances to the scripts/ right?

right

melekes:
See my comment about version https://github.com/tendermint/tendermint/issues/785#issuecomment-350045592

Issue #941, opened 2017-12-08
Cannot install on go1.9.2/ubuntu


ethanfrey:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):

BUG REPORT

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):

v0.13.0 and develop/b98098b
(tested and works on v0.12.0 and v0.12.1)

**Environment**:
- **OS** (e.g. from /etc/os-release): Ubuntu 16.04
- **Install tools**: 
- **Others**:

**What happened**:

Running `make all` produces:
```
[INFO]	Replacing existing vendor dependencies
CGO_ENABLED=0 go install -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short HEAD`" ./cmd/tendermint
go install net: open /usr/lib/go-1.9/pkg/linux_amd64/net.a: permission denied
Makefile:15: recipe for target 'install' failed
make: *** [install] Error 1
```

**What you expected to happen**:

tendermint binaries created, tests to pass

**How to reproduce it** (as minimally and precisely as possible):

```
git checkout v0.13.0
make all
```

**Anything else do we need to know**:

the file that caused the error is world readable:
```
# ls -l /usr/lib/go-1.9/pkg/linux_amd64/net.a
-rw-r--r-- 1 root root 1209646 Oct 31 10:38 /usr/lib/go-1.9/pkg/linux_amd64/net.a
# cat /usr/lib/go-1.9/pkg/linux_amd64/net.a > /dev/null
# echo $?
0
```

running go 1.9.2 in standard location:
```
# which go
/usr/bin/go
# go version
go version go1.9.2 linux/amd64
```

go was installed using apt-get from a ppa
```
# apt show golang-1.9-go
Package: golang-1.9-go
Version: 1.9.2-1ubuntu1~longsleep1+xenial
Priority: optional
Section: devel
Source: golang-1.9
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Go Compiler Team <pkg-golang-devel@lists.alioth.debian.org>
Installed-Size: 128 MB
Depends: golang-1.9-src (>= 1.9.2-1ubuntu1~longsleep1+xenial), libc6 (>= 2.4)
Recommends: g++, gcc, golang-1.9-race-detector-runtime, libc6-dev, pkg-config
[...]

PR #942, opened 2017-12-08
From CGO_ENABLED=0 from make install

ethanfrey:
It was writing to stdlib packages net, x/crypto, etc. and failing when it didn't have write access to them (which it often shouldn't)

Fixes issue #941 

Explained in golang issue: https://github.com/golang/go/issues/18981 (fixed in develop/1.10)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,13 +12,13 @@ BUILD_FLAGS = -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`g
 all: get_vendor_deps install test
 
 install:
-	CGO_ENABLED=0 go install $(BUILD_FLAGS) ./cmd/tendermint
+	go install $(BUILD_FLAGS) ./cmd/tendermint
 
 build:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -12,7 +12,7 @@ BUILD_FLAGS = -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`g
 all: get_vendor_deps install test
 
 install:
-	CGO_ENABLED=0 go install $(BUILD_FLAGS) ./cmd/tendermint
+	go install $(BUILD_FLAGS) ./cmd/tendermint
 
 build:
 	CGO_ENABLED=0 go build $(BUILD_FLAGS) -o build/tendermint ./cmd/tendermint/
```
melekes: maybe we should remove it from here too to be consistent?

ethanfrey: It doesn't have the same behavior on go build as per the comment. But sure, I can remove it. Will squish commits together

Issue #916, opened 2017-12-02
p2p/trust: remove sleeps from tests


ebuchman:
Tests are still racey, possibly due to `time.Sleep`. We should try to change the trustmetric so we can run the tests properly without sleeps

ebuchman:
See comment from https://github.com/tendermint/tendermint/pull/914

```
WARNING: DATA RACE
Read at 0x00c4200bc758 by goroutine 42:
  github.com/tendermint/tendermint/p2p/trust.(*TrustMetric).calcTrustValue()
      github.com/tendermint/tendermint/p2p/trust/_test/_obj_test/trustmetric.go:584 +0xa7
  github.com/tendermint/tendermint/p2p/trust.(*TrustMetric).NextTimeInterval()
      github.com/tendermint/tendermint/p2p/trust/_test/_obj_test/trustmetric.go:348 +0xef
  github.com/tendermint/tendermint/p2p/trust.(*TrustMetric).processRequests()
      github.com/tendermint/tendermint/p2p/trust/_test/_obj_test/trustmetric.go:607 +0xdb

Previous write at 0x00c4200bc758 by goroutine 8:
  [failed to restore the stack]

Goroutine 42 (running) created at:
  github.com/tendermint/tendermint/p2p/trust.NewMetricWithConfig()
      github.com/tendermint/tendermint/p2p/trust/_test/_obj_test/trustmetric.go:448 +0x281
  github.com/tendermint/tendermint/p2p/trust.(*TrustMetricStore).loadFromDB()
      github.com/tendermint/tendermint/p2p/trust/_test/_obj_test/trustmetric.go:147 +0x34a
  github.com/tendermint/tendermint/p2p/trust.TestTrustMetricStoreSaveLoad()
      /go/src/github.com/tendermint/tendermint/p2p/trust/trustmetric_test.go:66 +0x685
  fmt.Fscanf()
      /usr/local/go/src/fmt/scan.go:143 +0xf5
  fmt.Sscanf()
      /usr/local/go/src/fmt/scan.go:114 +0xbe
  github.com/tendermint/tendermint/vendor/github.com/syndtr/goleveldb/leveldb/storage.fsParseName()
      /go/src/github.com/tendermint/tendermint/vendor/github.com/syndtr/goleveldb/leveldb/storage/file_storage.go:571 +0x18c
  github.com/tendermint/tendermint/vendor/github.com/syndtr/goleveldb/leveldb/storage.(*fileStorage).List()
[...]

PR #944, opened 2017-12-08
Fixed the problems identified in issue #916

caffix:
https://github.com/tendermint/tendermint/issues/916

### DIFF
--- a/p2p/trust/metric_test.go
+++ b/p2p/trust/metric_test.go
@@ -61,30 +61,32 @@ func TestTrustMetricStopPause(t *testing.T) {
 	tm := NewMetricWithConfig(config)
 
 	// Allow some time intervals to pass and pause
-	time.Sleep(50 * time.Millisecond)
+	tm.NextTimeInterval()
+	tm.NextTimeInterval()
 	tm.Pause()
-	// Give the pause some time to take place
-	time.Sleep(10 * time.Millisecond)
 
 	first := tm.Copy().numIntervals
 	// Allow more time to pass and check the intervals are unchanged
-	time.Sleep(50 * time.Millisecond)
-	assert.Equal(t, first, tm.numIntervals)
+	tm.WaitForTimeIntervalToPass()
+	tm.WaitForTimeIntervalToPass()
+	assert.Equal(t, first, tm.Copy().numIntervals)
 
 	// Get the trust metric activated again
 	tm.GoodEvents(5)
 	// Allow some time intervals to pass and stop
-	time.Sleep(50 * time.Millisecond)
+	tm.NextTimeInterval()
+	tm.NextTimeInterval()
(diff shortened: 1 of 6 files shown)


p2p/trust/metric_test.go
```diff
@@ -61,30 +61,32 @@ func TestTrustMetricStopPause(t *testing.T) {
...
+	tm.NextTimeInterval()
 	tm.Pause()
-	// Give the pause some time to take place
-	time.Sleep(10 * time.Millisecond)
 
 	first := tm.Copy().numIntervals
 	// Allow more time to pass and check the intervals are unchanged
-	time.Sleep(50 * time.Millisecond)
-	assert.Equal(t, first, tm.numIntervals)
+	tm.WaitForTimeIntervalToPass()
```
melekes: why we can't call `tm.NextTimeInterval()` here too similar to above?

p2p/trust/metric_test.go
```diff
@@ -61,30 +61,32 @@ func TestTrustMetricStopPause(t *testing.T) {
...
 	tm.GoodEvents(5)
 	// Allow some time intervals to pass and stop
-	time.Sleep(50 * time.Millisecond)
+	tm.NextTimeInterval()
+	tm.NextTimeInterval()
 	tm.Stop()
-	// Give the stop some time to take place
-	time.Sleep(10 * time.Millisecond)
+	// Wait for the stop to take place
+	tm.WaitForStop()
```
melekes: we can make Stop blocking..

`Wait` is ok too, but it can be simplified imho https://github.com/tendermint/tmlibs/blob/master/common/service.go#L184

btw, is there a specific reason why metric is not a `Service`? i.e. all the things in TM that can be started/stopped follow the same interface and implementation https://github.com/tendermint/tendermint/blob/a2b92c07453f66277f70d0c3979a9dac67bbecea/state/txindex/indexer_service.go#L15

melekes:
Thanks! See my comments. I think we should be able to get rid of waitForStop/waitForX slices in the end as this impl looks super complicated

p2p/trust/metric_test.go
caffix: the wait method will wait until the ticker in the go-routine fires again, which is what the test requires

melekes: ok. I will take a look why ticker is so important here

caffix:
Thank you, @melekes . The Stop method now blocks until the trust metric has completely stopped. I can make it a service as well if you wish

caffix:
Okay, the trust metric is now a service. Additional code impacted has been updated. Also, the channel slices and associated methods have been removed. This has been achieved by adding the TestTicker, which gives us manual control over the passing of time intervals within test routines. All calls to Sleep within the test routines has also been removed.

Issue #945, opened 2017-12-09
Bring back transparent websocket


ebuchman:
Before the new pubsub, we had `types.EventSwitch`, which was an interface.

We had `rpc/client.HTTP` embed that interface via the embedded `rpc/client.WSEvents`, which called `types.NewEventSwitch` in its constructor (https://github.com/tendermint/tendermint/blob/v0.12.0/rpc/client/httpclient.go)

When we setup a WebsocketManager, we would pass in a `types.EventSwitch`, and this was used during `Subscribe`, etc.

This had the really nice property that we could transparently proxy the WS through an intermediate server - ie we could pass a `rpc/client.HTTP` to the `NewWebsocketManager` as a `EventSwitch` - so when `/subscribe` was called, it would hit the `Subscribe` method on the `EventSwitch`, which was actually an HTTP client, that would then call `/subscribe` over the network to its remote endpoint. You can see this in action here: https://github.com/cosmos/cosmos-sdk/blob/master/client/proxy.go

Now, the `WebsocketManager` no longer takes an EventSwitch. Instead it uses a global on  the rpc package. The new EventBus also isn't an interface, so we can't easily use eg an http client to replace it. 

Thus, to proxy the websocket would require some additional work. Unless there's good reason, we should try to bring back this kind of transparency - perhaps the only thing we need to do is provide an interface that EventBus implements and use that in the rpc pipe.

PR #948, opened 2017-12-10
bring back transparent websocket (Refs #945)

melekes:
Requires https://github.com/tendermint/tmlibs/pull/96

### DIFF
--- a/rpc/client/httpclient.go
+++ b/rpc/client/httpclient.go
@@ -323,6 +323,8 @@ func (w *WSEvents) eventListener() {
 				w.Logger.Error("failed to unmarshal response", "err", err)
 				continue
 			}
+			// NOTE: writing also happens inside mutex so we can't close a channel in
+			// Unsubscribe/UnsubscribeAll.
 			w.mtx.RLock()
 			if ch, ok := w.subscriptions[result.Query]; ok {
 				ch <- result.Data
--- a/rpc/client/httpclient.go
+++ b/rpc/client/httpclient.go
@@ -3,7 +3,6 @@ package client
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"sync"
 
 	"github.com/pkg/errors"
@@ -13,6 +12,7 @@ import (
 	rpcclient "github.com/tendermint/tendermint/rpc/lib/client"
 	"github.com/tendermint/tendermint/types"
 	cmn "github.com/tendermint/tmlibs/common"
+	tmpubsub "github.com/tendermint/tmlibs/pubsub"
 )
 
 /*
@@ -204,20 +204,14 @@ type WSEvents struct {
 	endpoint string
 	ws       *rpcclient.WSClient
 
-	subscriptions map[string]chan<- interface{}
 	mtx           sync.RWMutex
-
-	// used for signaling the goroutine that feeds ws -> EventSwitch
-	quit chan bool
-	done chan bool
+	subscriptions map[string]chan<- interface{}
 }
 
 func newWSEvents(remote, endpoint string) *WSEvents {
 	wsEvents := &WSEvents{
 		endpoint:      endpoint,
 		remote:        remote,
-		quit:          make(chan bool, 1),
-		done:          make(chan bool, 1),
 		subscriptions: make(map[string]chan<- interface{}),
 	}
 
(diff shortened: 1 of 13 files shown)


rpc/lib/types/types.go
```diff
@@ -135,10 +137,14 @@ type WSRPCConnection interface {
 	GetRemoteAddr() string
 	WriteRPCResponse(resp RPCResponse)
 	TryWriteRPCResponse(resp RPCResponse) bool
+	GetEventSubscriber() EventSubscriber
+}
 
-	AddSubscription(string, interface{}) error
-	DeleteSubscription(string) (interface{}, bool)
-	DeleteAllSubscriptions()
+// EventSubscriber mirros tendermint/tendermint/types.EventBusSubscriber
```
melekes: or it's ok to use `tmtypes` here?

ebuchman: hmm good question. let's keep the duplication for now and think about how we might improve package structure for things like this

rpc/client/httpclient.go
```diff
@@ -316,34 +309,27 @@ func (w *WSEvents) redoSubscriptions() {
...
-			if ch := w.getSubscription(result.Query); ch != nil {
+			w.mtx.RLock()
+			if ch, ok := w.subscriptions[result.Query]; ok {
 				ch <- result.Data
 			}
-		case <-w.quit:
-			// send a message so we can wait for the routine to exit
-			// before cleaning up the w.ws stuff
-			w.done <- true
+			w.mtx.RUnlock()
```
melekes: probably better to move up a little bit (i.e. unblock before writing to channel)

rpc/core/events.go
```diff
@@ -44,20 +44,19 @@ func Subscribe(wsCtx rpctypes.WSRPCContext, query string) (*ctypes.ResultSubscri
...
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to add subscription")
+		return nil, errors.Wrap(err, "failed to parse query")
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), subscribeTimeout)
 	defer cancel()
 	ch := make(chan interface{})
-	err = eventBus.Subscribe(ctx, addr, q, ch)
+	es := wsCtx.GetEventSubscriber()
```
melekes: ???? unused

rpc/client/httpclient.go
melekes: no. it is intended, so we can't close the channel

melekes: wrote a comment

rpc/client/httpclient.go
```diff
@@ -204,108 +204,101 @@ type WSEvents struct {
...
 	w.mtx.Lock()
-	defer w.mtx.Unlock()
-	ch, ok := w.subscriptions[query]
+	ch, ok := w.subscriptions[q]
 	if ok {
 		close(ch)
-		delete(w.subscriptions, query)
+		delete(w.subscriptions, q)
 	}
+	w.mtx.Unlock()
```
ebuchman: Why not keep the defer? I know there's a small performance price but probably worth it for maintainability

rpc/client/httpclient.go
```diff
@@ -316,34 +309,27 @@ func (w *WSEvents) redoSubscriptions() {
...
-		case resp := <-w.ws.ResponsesCh:
-			// res is json.RawMessage
+		case resp, ok := <-w.ws.ResponsesCh:
+			if !ok {
+				return
+			}
 			if resp.Error != nil {
-				// FIXME: better logging/handling of errors??
-				fmt.Printf("ws err: %+v\n", resp.Error.Error())
+				w.Logger.Error("WS error", "err", resp.Error.Error())
```
ebuchman: Woo!

rpc/core/events.go
```diff
@@ -44,20 +44,19 @@ func Subscribe(wsCtx rpctypes.WSRPCContext, query string) (*ctypes.ResultSubscri
...
+	if es == nil {
+		es = eventBus
```
ebuchman: :o hmm.

Seems a bit too magical.

This whole package is a bit magical with the globals. Should we look to introduce an explicit `RPCPipe` that contains all these guys, and make all these rpc funcs methods on that ?

rpc/core/events.go
```diff
@@ -100,18 +95,31 @@ func Subscribe(wsCtx rpctypes.WSRPCContext, query string) (*ctypes.ResultSubscri
...
+	if err != nil {
+		return nil, err
+	}
 	return &ctypes.ResultUnsubscribe{}, nil
 }
+
+func eventBusFor(wsCtx rpctypes.WSRPCContext) tmtypes.EventBusSubscriber {
+	es := wsCtx.GetEventSubscriber()
+	if es == nil {
+		es = eventBus
```
ebuchman: :o hmm.

So I guess this happens if someone calls NewWebsocketManager and forgets to pass the option to set the eventBus? 

Also maybe its time to make all these funcs methods on a `RPCPipe` struct or something so we can do away with the globals. Agree ?

rpc/lib/server/handlers.go
```diff
@@ -499,28 +503,6 @@ func (wsc *wsConnection) TryWriteRPCResponse(resp types.RPCResponse) bool {
 	}
 }
 
-func (wsc *wsConnection) AddSubscription(query string, data interface{}) error {
```
ebuchman: good riddance!

Issue #929, opened 2017-12-06
consensus: add time fields to signed messages (proposals and votes)


ebuchman:
In preparation for using a BFT time in the header

ethanfrey:
This is just adding the info and not doing anything with it?

Should this be backwards compatible with older versions? Probably not, it is not persisted and all nodes must upgrade tendermint at the same point, right?

ebuchman:
That's right. Once it's in, we can use it to later make BFT timestamps in the block header without a breaking change. So let's just get them in there.

ethanfrey:
Addressed in #953 which was merged. Closing issue.

PR #953, opened 2017-12-11
Add Timestamp to Proposal/Vote

ethanfrey:
Fixes issue #929

Store it as time.Timestamp locally, encode it as RFC3339 with milliseconds
before signing the canonical form.

### DIFF
--- a/types/block.go
+++ b/types/block.go
@@ -420,7 +420,7 @@ func (data *Data) StringIndented(indent string) string {
 
 // BlockID defines the unique ID of a block as its Hash and its PartSetHeader
 type BlockID struct {
-	Hash        data.Bytes    `json:"hash"`
+	Hash        data.Bytes    `json:"hash,omitempty"`
 	PartsHeader PartSetHeader `json:"parts"`
 }
 
--- a/types/block.go
+++ b/types/block.go
@@ -420,7 +420,7 @@ func (data *Data) StringIndented(indent string) string {
 
 // BlockID defines the unique ID of a block as its Hash and its PartSetHeader
 type BlockID struct {
-	Hash        data.Bytes    `json:"hash,omitempty"`
+	Hash        data.Bytes    `json:"hash"`
 	PartsHeader PartSetHeader `json:"parts"`
 }
 
--- a/types/canonical_json.go
+++ b/types/canonical_json.go
@@ -3,13 +3,14 @@ package types
 import (
 	"time"
 
+	wire "github.com/tendermint/go-wire"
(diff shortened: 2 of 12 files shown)


types/block.go
```diff
@@ -420,7 +420,7 @@ func (data *Data) StringIndented(indent string) string {
 
 // BlockID defines the unique ID of a block as its Hash and its PartSetHeader
 type BlockID struct {
-	Hash        data.Bytes    `json:"hash"`
+	Hash        data.Bytes    `json:"hash,omitempty"`
```
ebuchman: ? and why no on partsheader too ?

types/canonical_json.go
```diff
@@ -1,11 +1,16 @@
...
 import (
+	"time"
+
 	"github.com/tendermint/go-wire/data"
 )
 
 // canonical json is go-wire's json for structs with fields in alphabetical order
 
+// timeFormat is RFC3339Millis, used for generating the sigs
+const timeFormat = "2006-01-02T15:04:05.999Z07:00"
```
ebuchman: can we take this from go-wire instead? ie.  wire.RFC3339Millis

types/canonical_json.go
```diff
@@ -22,13 +27,15 @@ type CanonicalJSONProposal struct {
 	POLBlockID       CanonicalJSONBlockID       `json:"pol_block_id"`
 	POLRound         int                        `json:"pol_round"`
 	Round            int                        `json:"round"`
+	Timestamp        string                     `json:"timestamp"`
```
ebuchman: Maybe better to use an int here ?

types/block.go
ethanfrey: Sorry, totally not needed. Added as I had a weird json artifact, but it was in the test code...

types/canonical_json.go
ethanfrey: okay

types/canonical_json.go
ethanfrey: Hmmm....

ethanfrey: Int would not work, it must be int64 to handle milliseconds. Then we define it as what? Milliseconds since unix epoch?

I feel the json encoding is designed to be somewhat human readable. The random int in the rpc interface for header is also quite confusing.

ethanfrey:
Okay guys, I think this is ready for final review.

melekes:
Awesome work!

ethanfrey:
Can I merge it with this approval? Can Anton do it? Or does Bucky do the do?

Issue #668, opened 2017-09-20
Make the ABCI application aware of validator health/participation


mdyring:
Feature request for consideration, from the Rocket Chat.

Short version:
It seems useful to let the ABCI app know the health of the validator set. For instance, a nice metric to have would be number of blocks "missed" (validator in validator set, but didn't vote), so that dead validators could be expired by the app.

There are some security implications by this, so would need careful consideration by the application.

Chat log:
srm 8:19 AM 
@pepesza good questions. Re: throughput limiting factors - yes, it is limit by the number of transitions/transactions (a transition is a new block though) and is correlated with the number of nodes due to quadratic growth of messages to transfer because of the gossip.
How can an abci app be notified about a dropped validator?

mdyring 10:39 AM
@srm yeah i don't think that is possible at the moment, but actually it would be an interesting metric for the app to have.. for instance, how many of the last blocks that a validator "missed" it could be passed in the the commit/endblock invocation
 srm 8:20 AM
How can an abci app be notified about a dropped validator?
then it could be left for the app to decide that after X missed blocks, a validator is dropped

srm 10:54 AM
Exactly. Especially seeing that an offline validator increases each nth round.

mdyring 11:01 AM
@adrian @bucky does the above make sense to you - i.e. should i open a GH issue?

adrian Admin Owner Cosmos 11:16 AM
@mdyring What do you mean by missed?
Also, if in doubt, always open a github issue, so that we can share the conversation with more people
🙂

mdyring 11:19 AM
@adrian "missed": validator is in validator set, but didn't participate (vote) on last X blocks
that leaves it up to the app to decide if some validators should be retired/expire after some threshold it decides

srmo:
As I’ve said, I think this is important to have a consistent blocktime, otherwise dead validators will slow down the overall network.

melekes:
Interesting proposal for the dead validators problem. I think we could send a validator set for the last block with the corresponding heights they are on, so an app can take some measures if some validator is too far behind. The other question is do we want to burden the app with such task or we can implement it on the TM side?

srmo:
I would consider it a burden to the core and a treat for the app. You usually want to have control over the validator set in your app, in my understanding at least. Moving this to core seems to provoke all kinds of wishes on how to configure that behavior. In the end, the app has to decide on the validator set either way, doesn't it?

mdyring:
[...]

PR #954, opened 2017-12-11
Send absent validators

melekes:
Refs #668

### DIFF
--- a/state/execution.go
+++ b/state/execution.go
@@ -76,19 +76,11 @@ func execBlockOnProxyApp(txEventPublisher types.TxEventPublisher, proxyAppConn p
 	}
 	proxyAppConn.SetResponseCallback(proxyCb)
 
-	// determine validators who did not sign last block
+	// determine which validators did not sign last block
 	absentVals := make([]int32, 0)
-	for valA, _ := range lastValidators.Validators {
-		found := false
-		for _, voteB := range block.LastCommit.Precommits {
-			valB := voteB.ValidatorIndex
-			if valA == valB {
-				found = true
-				break
-			}
-		}
-		if !found {
-			absentVals = append(absentVals, int32(valA))
+	for valI, vote := range block.LastCommit.Precommits {
+		if vote == nil {
+			absentVals = append(absentVals, int32(valI))
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


melekes:
Not sure how to test this. Should I write an app, which asserts that it did receive absent vals (given this block and this last commit)

ebuchman:
What makes this breaking ?

ebuchman:
It's in `state` package not covered by semver guarantees: https://github.com/tendermint/tendermint#semver

:D

state/execution_test.go
```diff
@@ -30,16 +32,53 @@ func TestApplyBlock(t *testing.T) {
...
+	// but last commit contains only the first validator
+	prevHash := state.LastBlockID.Hash
+	prevParts := types.PartSetHeader{}
+	prevBlockID := types.BlockID{prevHash, prevParts}
+	lastCommit := &types.Commit{BlockID: prevBlockID, Precommits: []*types.Vote{
+		{ValidatorIndex: 0},
+	}}
+
+	valHash := state.Validators.Hash()
+	block, _ := types.MakeBlock(2, chainID, makeTxs(2), lastCommit,
```
ethanfrey: Nice test, but with all this setup, why not make a table-driven test and verify a few different situations, just loop over thel last 10 lines or so with different variables (none absent, one absent, multiple absent)?

state/execution.go
```diff
@@ -76,12 +76,28 @@ func execBlockOnProxyApp(txEventPublisher types.TxEventPublisher, proxyAppConn p
 	}
 	proxyAppConn.SetResponseCallback(proxyCb)
 
+	// determine validators who did not sign last block
+	absentVals := make([]int32, 0)
+	for valA, _ := range lastValidators.Validators {
```
ethanfrey: Aren't the precommits in order by validator?
I vaguely remember something in the light client, that if slot X was nil, then validator X didn't sign.

This looks correct, but I think it can be made more efficient (O(N*N) here, even if N is only 100, still...)

melekes: Good question

ebuchman: That's right. The `votes` array in a `VoteSet` and the `Precommits` in a `Commit` should always be as large as the validtor set, possibly containing some nils. So we can just look for the nil ones and send that index

PR #955, opened 2017-12-11
p2p: exponential backoff on reconnect. closes #939

### LINKED ISSUE #939: p2p: exponential backoff on reconnecting to persistent peers
https://github.com/tendermint/tendermint/blob/master/p2p/switch.go#L16

ebuchman:


p2p/switch_test.go
```diff
@@ -272,10 +272,10 @@ func TestSwitchReconnectsToPersistentPeer(t *testing.T) {
 	// simulate failure by closing connection
 	peer.CloseConn()
 
-	// TODO: actually detect the disconnection and wait for reconnect
+	// TODO: remove sleep, detect the disconnection, wait for reconnect
```
zramsay: make an issue, link to it in code

p2p/switch.go
```diff
@@ -317,22 +335,24 @@ func (sw *Switch) DialSeeds(addrBook *AddrBook, seeds []string) error {
...
 			j := perm[i]
 			sw.dialSeed(netAddrs[j])
 		}(i)
 	}
 	return nil
 }
 
+// sleep for interval plus some random amount of ms on [0, dialRandomizerIntervalMilliseconds]
+func (sw *Switch) randomSleep(interval time.Duration) {
+	r := time.Duration(sw.rng.Int63n(dialRandomizerIntervalMilliseconds)) * time.Millisecond
```
odeke-em: Am not sure that this is exponential backoff:
It should be `(1 << n) * unitOfTime e.g millisecond` + `jitter from rng`
this one only has the `jitter from rng`

This code instead seems to sleep just a constant amount of times which defeats the purpose of exponentially backing off.

p2p/switch_test.go
ebuchman: ok thanks https://github.com/tendermint/tendermint/issues/956

p2p/switch.go
ebuchman: this is just a function that sleeps an input+rand amount.

its up to the caller to ensure the passed value is exponentially backing off, ie https://github.com/tendermint/tendermint/pull/955/files#diff-44d98df4858f140bb01702ae7982ec31R476

Legit ?

odeke-em: Oh I see, my bad, awesome work!

PR #963, opened 2017-12-11
Remove get_deps, update_deps and list_deps from Makefile

melekes:
See commit description for details.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,27 +54,14 @@ draw_deps:
 	go get github.com/RobotsAndPencils/goviz
 	@goviz -i github.com/tendermint/tendermint/cmd/tendermint -d 3 | dot -Tpng -o dependency-graph.png
 
-list_deps:
-	@go list -f '{{join .Deps "\n"}}' ./... | \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -54,27 +54,14 @@ draw_deps:
...
-update_deps:
-	@echo "--> Updating dependencies"
-	@go get -d -u ./...
-
 get_vendor_deps:
 	@hash glide 2>/dev/null || go get github.com/Masterminds/glide
 	@rm -rf vendor/
 	@echo "--> Running glide install"
-	$(GOPATH)/bin/glide install
+	@$(GOPATH)/bin/glide install
```
ebuchman: wasn't greg asking for us to change this? Ie. what if glide is installed via pkg manager and not go ?

melekes: no, he was asking to change `@hash glide 2>/dev/null` above to `@hash $(GOPATH)/bin/glide 2>/dev/null`. but I am not sure this is correct. my intention was if glide is already present (e.g. user did `brew install glide`) than we do not attempt to install it ourselves

PR #964, opened 2017-12-12
fix gometalinter.v2 automatically

petabytestorage:


Makefile
```diff
@@ -11,9 +11,15 @@ BUILD_FLAGS = -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`g
 
 all: get_vendor_deps install test
 
-install:
+install: .install_precondition.stamp
```
melekes: `metalinter` has nothing to do with `install` target. ie. you don't need metalinter in order for install to work

petabytestorage: it doesn't work without my fix. this hook position works. if you have another place that you think is better please suggest it specifically @melekes   the error without the fix occurs inevitably using the standard build command: ```docker run --rm -v "$PWD":/go/src/github.com/tendermint/tendermint -w /go/src/github.com/tendermint/tendermint -it golang:1.9 make```

Makefile
```diff
@@ -9,7 +9,7 @@ TMHOME = $${TMHOME:-$$HOME/.tendermint}
 
 BUILD_FLAGS = -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short HEAD`"
 
-all: tools get_vendor_deps install test
```
melekes: but test target requires tools. i don't understand

PR #967, opened 2017-12-12
Add TotalTx to block header

ethanfrey:
Addresses one part of issue #952

 - Add extra field to header
 - Make NumTx also an int64 for consistency.
 - Update state to keep track of this info.
 - Change function args as needed.

### DIFF
--- a/consensus/mempool_test.go
+++ b/consensus/mempool_test.go
@@ -108,7 +108,8 @@ func TestTxConcurrentWithCommit(t *testing.T) {
 	for nTxs := 0; nTxs < NTxs; {
 		select {
 		case b := <-newBlockCh:
-			nTxs += b.(types.TMEventData).Unwrap().(types.EventDataNewBlock).Block.Header.NumTxs
+			evt := b.(types.TMEventData).Unwrap().(types.EventDataNewBlock)
+			nTxs += int(evt.Block.Header.NumTxs)
 		case <-ticker.C:
 			panic("Timed out waiting to commit blocks with transactions")
 		}
--- a/lite/helpers.go
+++ b/lite/helpers.go
@@ -112,10 +112,11 @@ func genHeader(chainID string, height int64, txs types.Txs,
 	vals *types.ValidatorSet, appHash []byte) *types.Header {
 
 	return &types.Header{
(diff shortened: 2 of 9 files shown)


lite/helpers.go
```diff
@@ -112,10 +112,11 @@ func genHeader(chainID string, height int64, txs types.Txs,
...
 
 	return &types.Header{
-		ChainID: chainID,
-		Height:  height,
-		Time:    time.Now(),
-		NumTxs:  len(txs),
+		ChainID:  chainID,
+		Height:   height,
+		Time:     time.Now(),
+		NumTxs:   int64(len(txs)),
```
melekes: should this be int64? int32 should be more than enough

melekes: or it's to avoid int64(..) cast?

consensus/mempool_test.go
```diff
@@ -108,7 +108,8 @@ func TestTxConcurrentWithCommit(t *testing.T) {
 	for nTxs := 0; nTxs < NTxs; {
```
melekes: should we make nTxs int64 like `nTxs := int64(0)`

lite/helpers.go
ethanfrey: yup, so it is the same type as TotalTxs. No need to have them different types, just confusion when adding

consensus/mempool_test.go
ethanfrey: I could, but then I would have to change types here as well:
```
	NTxs := 10000
	go deliverTxsRange(cs, 0, NTxs)
```

simpler to cast here once

ethanfrey:
Okay, wrote some block tests that were non-existent. Should work now.

ebuchman:
Great. Thanks for the test.

I wonder if the MakeBlock/ValidateBlock could be better structured, since the args list is a little out of control :P. For instance maybe just take a block and it's parent. Might ripple through the State in ways we don't like though. Just a thought.

ethanfrey:
Yeah, I just wanted to get some test coverage at all.
Probably could be cleaned up, but I don't know all usages of it. Will keep it in mind as I see these functions in use.

PR #972, opened 2017-12-13
Update EndBlock parameters

ethanfrey:
* Update abci dependencies
* Modify references from Diffs to Changes
* Fixes issues #924
* Partial fix for #952 
  * Updates consensus parameter upon EndBlock
  * Add Hash for ConsensusParams to block headers
  * State maintains LastConsensusHash to verify blocks

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -25,7 +25,6 @@ dependencies:
 
 test:
   override:
-    - cd "$PROJECT_PATH" && make tools && make get_vendor_deps && make metalinter_test
     - cd "$PROJECT_PATH" && set -o pipefail && make test_integrations 2>&1 | tee test_integrations.log:
         timeout: 1800
   post:
--- a/state/state.go
+++ b/state/state.go
@@ -251,7 +251,7 @@ func (s *State) SetBlockAndValidators(header *types.Header, blockPartsHeader typ
 	// Update validator accums and set state variables
 	nextValSet.IncrementAccum(1)
 
-	nextParams := s.Params.ApplyChanges(
+	nextParams := applyChanges(s.Params,
 		abciResponses.EndBlock.ConsensusParamChanges)
 
 	s.setBlockAndValidators(header.Height,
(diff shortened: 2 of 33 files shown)


state/state.go
```diff
@@ -251,24 +251,68 @@ func (s *State) SetBlockAndValidators(header *types.Header, blockPartsHeader typ
...
 		header.NumTxs,
 		types.BlockID{header.Hash(), blockPartsHeader},
 		header.Time,
-		prevValSet, nextValSet)
+		prevValSet, nextValSet,
+		nextParams)
+
+}
+
+// applyChanges returns a new param set, overriding any
```
odeke-em: How about?
```go
// applyChanges returns new ConsensusParams
// whose fields are set to any non-zero fields of c.
// If c is nil, it returns p unmodified, as it was passed in.
```

ethanfrey: done!

odeke-em: Thank you @ethanfrey!

ethanfrey:
Need help debugging this.

Make vagrant_test works.
CI times out. 

Whatever

melekes:
```
E[12-14|23:06:21.115] Error signing vote                           module=consensus height=14 round=0 vote="Vote{0:DFCCC3603B2C 14/00/1(Prevote) BE4288EAF2D4 {<nil>} @ 2017-
12-14T23:06:21.115Z}" err="Error signing vote: Step regression"
```

types/block.go
```diff
@@ -91,6 +92,9 @@ func (b *Block) ValidateBasic(chainID string, lastBlockHeight int64,
 	if !bytes.Equal(b.AppHash, appHash) {
 		return errors.New(cmn.Fmt("Wrong Block.Header.AppHash.  Expected %X, got %v", appHash, b.AppHash))
 	}
+	if !bytes.Equal(b.ConsensusHash, consensusHash) {
+		return errors.New(cmn.Fmt("Wrong Block.Header.ConsensusHash.  Expected %X, got %v", consensusHash, b.ConsensusHash))
```
odeke-em: cmn.Fmt is deprecated, in the future I think just `fmt.Sprintf` would suffice ;)

types/params_test.go
```diff
@@ -38,3 +40,49 @@ func TestConsensusParamsValidation(t *testing.T) {
...
+		makeParams(6, 5, 4, 3, 2, 1),
+	}
+
+	hashes := make([][]byte, len(params))
+	for i := range params {
+		hashes[i] = params[i].Hash()
+	}
+
+	// make sure there are no duplicates...
+	// sort, then check in order for matches
```
odeke-em: Neat!

state/state.go
```diff
@@ -251,24 +254,76 @@ func (s *State) SetBlockAndValidators(header *types.Header, blockPartsHeader typ
...
+
+	if c == nil {
+		return p
+	}
+	res := p
+	// we must defensively consider any structs may be nil
+	if c.BlockSize != nil {
+
+		if c.BlockSize.MaxBytes != 0 {
+			res.BlockSize.MaxBytes = int(c.BlockSize.MaxBytes)
```
melekes: should we check this is greater than `0`?

ebuchman:
Thanks. I've got it from here. Let me merge the other PR in and then we can do final review

circle.yml
```diff
@@ -25,7 +25,6 @@ dependencies:
 
 test:
   override:
-    - cd "$PROJECT_PATH" && make tools && make get_vendor_deps && make metalinter_test
```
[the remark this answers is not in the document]
ebuchman: It was failing with out of memory errors. I don't have time for that right now.

Issue #977, opened 2017-12-14
Remove timeout in wal_generator


ebuchman:
ie. https://github.com/tendermint/tendermint/blob/feature/enhance-endblock/consensus/wal_generator.go#L94

It's still biting.

Either remove this, crank it to minutes (Circle man ...), or replace it with perblock event.

Related to https://github.com/tendermint/tendermint/issues/956

PR #981, opened 2017-12-15
enable logging for wal_generator and set timeout to 1 min

melekes:
Refs #977


melekes:
let's see how this goes

Issue #947, opened 2017-12-09
upgrade to circle 2.0


zramsay:
and remove `docker-machine` from circle.yml

zramsay:
the stack is going to jenkins, which is a whole nother issue

PR #983, opened 2017-12-16
Circle testing

ebuchman:
Trying to hunt non-determinism on Circle. 

We've been using CircleCI's version of docker v1.10.0, which is nearly 2 years old. Perhaps it  had a bug in the networking layer ?!


zramsay:
#947 might help somewhat?

ebuchman:
Huh. Seems like mostly we just needed bigger timeouts .... Producing a block really can take > 60s when theres 7 nodes running in docker containers on a single virtual machine ...

PR #991, opened 2017-12-20
cmd/tendermint: fix initialization file creation checks

### LINKED ISSUE #989: installation instructions need updating
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):

0.14.0-88f5f21

**Environment**:
- **Install tools**:
- **Others**:
```shell
$ uname -a
Darwin Emmanuels-MacBook-Pro-2.local 15.6.0 Darwin Kernel Version 15.6.0: Sun Jun  4 21:43:07 PDT 2017; root:xnu-3248.70.3~1/RELEASE_X86_64 x86_64```

**What happened**:
I built tendermint from scratch, starting with erasing my previous ~/.tendermint directory.
I looked at the instructions at http://tendermint.readthedocs.io/projects/tools/en/master/getting-started.html

```shell
$ tendermint init
$ tendermint node --proxy_app=dummy
ERROR: Failed to create node: Couldn't read GenesisDoc file: open /Users/emmanuelodeke/.tendermint/genesis.json: no such file or directory
```
I then pinged the internal engineering team but in vain
```shell
$ tendermint unsafe_reset_all
$ tendermint node --proxy_app=dummy
ERROR: Failed to create node: Couldn't read GenesisDoc file: open /Users/emmanuelodeke/.tendermint/genesis.json: no such file or directory
```
and then with other permutations of the suggestions
```shell
$ tendermint init
$ tendermint node --proxy_app=dummy
ERROR: Failed to create node: Couldn't read GenesisDoc file: open /Users/emmanuelodeke/.tendermint/genesis.json: no such file or directory
```

I expected initialization to work

Delete your ~/.tendermint directory and try to start afresh with the instructions at http://tendermint.readthedocs.io/projects/tools/en/master/getting-started.html

odeke-em:
Fixes #989.

The original initialization sequence started to inexplicably
fail
```shell
tendermint unsafe_reset_all
tendermint init
tendermint node --proxy_app=dummy
```

used to fail with
```shell
ERROR: Failed to create node: Couldn't read GenesisDoc file: open
/Users/emmanuelodeke/.tendermint/genesis.json: no such file or directory
```

because the initialization sequence always assumed that the
genesisDoc would only be set if the privValidator was generated.

However, `tendermint unsafe_reset_all` only created the
`priv_validator.json` file which would mean that then running
`tendermint init` would never create the `genesis.json` file
which if following the recommended sequence would then fail
since the `genesis.json` was absent.

### DIFF
--- a/cmd/tendermint/commands/init.go
+++ b/cmd/tendermint/commands/init.go
@@ -1,8 +1,6 @@
 package commands
 
 import (
-	"os"
-
 	"github.com/spf13/cobra"
 
 	"github.com/tendermint/tendermint/types"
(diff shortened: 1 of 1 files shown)


cmd/tendermint/commands/init.go
```diff
@@ -18,28 +16,29 @@ var InitFilesCmd = &cobra.Command{
...
-			}
+	genFile := config.GenesisFile()
+	if cmn.FileExists(genFile) {
+		logger.Info("Already initialized", "geneisis", genFile)
+	} else {
+		genDoc := types.GenesisDoc{
+			ChainID: cmn.Fmt("test-chain-%v", cmn.RandStr(6)),
 		}
+		genDoc.Validators = []types.GenesisValidator{{
+			PubKey: privValidator.GetPubKey(),
```
melekes: shouldn't we load PrivVal if it exists?
otherwise, genesis file will have "incorrect" priv validator

melekes: there was a TODO somewhere about it..

odeke-em: I think if we used types.LoadOrGenPrivValidatorFS then we'd be set, and the function to use would be
https://github.com/tendermint/tendermint/blob/652d1e3de8a3f216299ca66c595cc66504930d28/types/priv_validator.go#L131-L140

odeke-em: @melekes please take a look at the code above where we load it if it exists lest we generate it. Thank you for that suggestion!

cmd/tendermint/commands/init.go
```diff
@@ -18,28 +16,31 @@ var InitFilesCmd = &cobra.Command{
...
-	if _, err := os.Stat(privValFile); os.IsNotExist(err) {
-		privValidator := types.GenPrivValidatorFS(privValFile)
+	var privValidator *types.PrivValidatorFS
+	if cmn.FileExists(privValFile) {
+		privValidator = types.LoadPrivValidatorFS(privValFile)
+		logger.Info("Already initialized", "priv_validator", privValFile)
+	} else {
+		privValidator = types.GenPrivValidatorFS(privValFile)
 		privValidator.Save()
+		logger.Info("Initialized tendermint", "privValidator", privValFile)
```
melekes: let's change messages to something more clear, like `Initialized private validator` and `Found private validator`

odeke-em:
Awesome, thanks for the review @melekes!

cmd/tendermint/commands/init.go
```diff
@@ -17,29 +15,34 @@ var InitFilesCmd = &cobra.Command{
...
-				PubKey: privValidator.GetPubKey(),
-				Power:  10,
-			}}
-
-			if err := genDoc.SaveAs(genFile); err != nil {
-				panic(err)
-			}
+	// genesis file
+	genFile := config.GenesisFile()
+	if cmn.FileExists(genFile) {
```
melekes: note there can be a situation where there is a genesis file, but no private validator. then genesis will not be modified. but I think it's fine.

Issue #984, opened 2017-12-16
priv validator should just return last sign bytes if h/r/s matches


ebuchman:
since now we have time in the msgs and we might crash between writing the priv val and writing to wal

zramsay:
#993

PR #993, opened 2017-12-20
priv validator returns last sign bytes if h/r/s matches

melekes:
since now we have time in the msgs and we might crash between writing
the priv val and writing to wal.

Refs #984

### DIFF
--- a/types/priv_validator.go
+++ b/types/priv_validator.go
@@ -193,8 +193,8 @@ func (privVal *PrivValidatorFS) Reset() {
 	privVal.Save()
 }
 
-// SignVote signs a canonical representation of the vote, along with the chainID.
-// Implements PrivValidator.
+// SignVote signs a canonical representation of the vote, along with the
+// chainID. Implements PrivValidator.
 func (privVal *PrivValidatorFS) SignVote(chainID string, vote *Vote) error {
 	privVal.mtx.Lock()
 	defer privVal.mtx.Unlock()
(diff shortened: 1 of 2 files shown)


types/priv_validator_test.go
```diff
@@ -99,7 +99,6 @@ func TestSignVote(t *testing.T) {
 	privVal := GenPrivValidatorFS(tempFilePath)
 
 	block1 := BlockID{[]byte{1, 2, 3}, PartSetHeader{}}
-	block2 := BlockID{[]byte{3, 2, 1}, PartSetHeader{}}
```
ebuchman: why are we removing ?

melekes: do you mean why remove the test case? because we removed the check and now test fails with "expected err, but got nil"

types/priv_validator.go
```diff
@@ -211,63 +213,85 @@ func (privVal *PrivValidatorFS) SignVote(chainID string, vote *Vote) error {
...
 	return nil
 }
 
-// signBytesHRS signs the given signBytes if the height/round/step (HRS) are
-// greater than the latest state. If the HRS are equal, it returns the
-// privValidator.LastSignature.
-func (privVal *PrivValidatorFS) signBytesHRS(height int64, round int, step int8, signBytes []byte) (crypto.Signature, error) {
-	sig := crypto.Signature{}
-
+// returns error if HRS regression. returns true if HRS is unchanged
```
melekes: comment does not reflect the code

melekes: it should be more like `returns error if HRS regression or no LastSignBytes. returns true if HRS is unchanged`

PR #994, opened 2017-12-21
Clean/block validation

ebuchman:
Clean up the `MakeBlock` and `ValidateBasic` functions by distinguishing what can be done with just the Block, and what requires the State.

Also make some structure in the State more explicit.

### DIFF
--- a/state/execution_test.go
+++ b/state/execution_test.go
@@ -23,42 +23,51 @@ var (
 	nTxsPerBlock = 10
 )
 
-/*
 func TestValidateBlock(t *testing.T) {
 	state := state()
 	state.SetLogger(log.TestingLogger())
 
-	block := makeBlock(1, state)
-
 	// proper block must pass
-	err = block.ValidateBasic("hello", h-1, 10, lastID, block.Time, appHash, consensusHash)
+	block := makeBlock(state, 1)
+	err := state.ValidateBlock(block)
 	require.NoError(t, err)
 
 	// wrong chain fails
-	err = block.ValidateBasic("other", h-1, 10, lastID, block.Time, appHash, consensusHash)
+	block = makeBlock(state, 1)
+	block.ChainID = "not-the-real-one"
+	err = state.ValidateBlock(block)
 	require.Error(t, err)
 
 	// wrong height fails
-	err = block.ValidateBasic("hello", h+4, 10, lastID, block.Time, appHash, consensusHash)
+	block = makeBlock(state, 1)
+	block.Height += 10
+	err = state.ValidateBlock(block)
 	require.Error(t, err)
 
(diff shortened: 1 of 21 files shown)


state/execution_test.go
```diff
@@ -23,6 +23,43 @@ var (
 	nTxsPerBlock = 10
 )
 
+/*
```
ebuchman: TODO: test

state/state.go
```diff
@@ -27,40 +28,51 @@ func calcValidatorsKey(height int64) []byte {
 	return []byte(cmn.Fmt("validatorsKey:%v", height))
 }
 
+func calcConsensusParamsKey(height int64) []byte {
```
ebuchman: TODO: actually use this to store the historical consensus params.

Alternatively, just stick the ConsensusParams directly in the header. But that's probably ridiculous.

types/block.go
```diff
@@ -15,69 +15,38 @@ import (
...
-// TODO: Add version information to the Block struct.
-func MakeBlock(height int64, chainID string, txs []Tx,
-	totalTxs int64, commit *Commit,
-	prevBlockID BlockID, valHash, appHash, consensusHash []byte,
-	partSize int) (*Block, *PartSet) {
-
-	newTxs := int64(len(txs))
+// MakeBlock returns a new block with an empty header, except what can be computed from itself.
+// It populates the same set of fields validated by ValidateBasic
+func MakeBlock(height int64, txs []Tx, commit *Commit) *Block {
```
ebuchman: maybe send chainID in here too ?

types/block.go
```diff
@@ -15,69 +15,38 @@ import (
...
 		Data: &Data{
 			Txs: txs,
 		},
 	}
 	block.FillHeader()
-	return block, block.MakePartSet(partSize)
+	return block
 }
 
 // ValidateBasic performs basic validation that doesn't involve state data.
```
ebuchman: lol at this comment prior to this PR :P

state/state.go
```diff
@@ -27,40 +28,51 @@ func calcValidatorsKey(height int64) []byte {
...
+	Validators                  *types.ValidatorSet
+	LastValidators              *types.ValidatorSet
 	LastHeightValidatorsChanged int64
 
-	// AppHash is updated after Commit
+	// Consensus parameters used for validating blocks.
+	// Changes returned by EndBlock and updated after Commit.
+	ConsensusParams                  types.ConsensusParams
+	LastConsensusParams              types.ConsensusParams
+	LastHeightConsensusParamsChanged int64
```
melekes: don't see where we actually change it and use it.

ebuchman: still need to save these and use this ...

Issue #988, opened 2017-12-18
Consider using libp2p for p2p layer


paralin:
**Is this a BUG REPORT or FEATURE REQUEST?**: Feature request / RFC

The [libp2p](https://libp2p.io/) library built and used by the [IPFS](https://ipfs.io) team aims to be a modular multi-transport baseline for P2P with multiple applications / protocols communicating over the same bus.

It would be interesting to allow Tendermint to use IPFS's peering in a common bus via libp2p. This would also allow us to combine efforts with that team when improving the P2P layer of Tendermint.

ebuchman:
We're interested in doing this but not before Cosmos launch. 

Spoke with go-ethereum folks at Devcon about working on this as well. 

It's something we can reconsider around mid-2018 once things are a little more stable in both Tendermint and LibP2P.

That said, I'm interested in sooner understanding exactly how this would work and what would change. Are you interested in working on a prototype of the integration, a proof-of-concept?

paralin:
Sure, I can try to replace Tendermint's underlying transport with a libp2p transport. The general plan is:

 1. Keep the existing PEX reactor, peers still need to share peers with each other.
 2. Replace the Dial function with one that attempts to connect to a peer via libp2p.
 3. Replace IPV4 addresses with libp2p multiaddrs.

ebuchman:
Very cool. Excited to see this.

Would also like to better understand how libp2p does peer exchange.

Satisfies `Consensus` interface would be cool, but we currently don't support Rollback!

paralin:
@ebuchman I have a proof of concept done (since a week in the PR). 

libp2p does not necessarily do peer exchange. It is very modular, so you can use or not use whatever components you want, or write your own implementation of the components (routing, discovery, upnp, transports, protocols, etc). For ipfs they use kademia routing with a DHT.

ebuchman:
Thanks for the work here @paralin 

For anyone watching, we've made a lot of updates to p2p layer on latest develop (soon to be released as 0.16). We're not using any libp2p stuff yet, but we're watching it and will perhaps use some down the road. 

Will close this in the meantime.

rairyx:
Any thoughts of when to switch to libp2p?

ebuchman:
Wrote an update here: https://github.com/tendermint/tendermint/issues/3696

rairyx:
Cool thanks @ebuchman

abourget:
[...]

PR #995, opened 2017-12-21
[RFC] Re-base p2p layer on libp2p

paralin:
This commit re-bases the p2p layer on libp2p.

Tests are broken, but the main Tendermint executable compiles and works.

This is a proof-of-concept "prototype" as requested in #988 

Note that "seeds" are now peer IDs. It is also possible to give libp2p a hint as to how to connect to the peers, with multiaddrs, which would let seeds behave the same as before if used. This may not be the desired behavior, though. We want to connect to a specific peer as a seed, not necessarily a specific endpoint - we could change the format to allow specifying the endpoint(s) optionally as a hint.


paralin:
Current work in this area:

 - [ ] Replace UPNP with libp2p upnp.
 - [ ] Replace ListenAddress with ListenAddresses and support multiaddrs.
 - [ ] Seeds should become a list of seed nodes to connect to, define format for this variable.

For seeds, I propose the following:

```
peerid[@192.168.1.1:5000]
OR
peerid[@/ip4/192.168.1.1/tcp/5000] (multiaddr format)
```

Comma separated. `@` and IP information for each peer is optional, as libp2p can do discovery and routing if needed.

In general libp2p is really modular. You can use as little or as much of it as you want. Currently as a proof of concept, I'm going to go all-in on tearing apart the codebase and replacing everything with "libp2p-ified" components. I don't expect you to do this long term, but it can serve as a good reference for how things COULD be done with libp2p in Tendermint, and I'm going to use it personally exclusively.

paralin:
Prototyped the following seeds syntax, works quite nicely:

```
[p2p]
laddr = "tcp://0.0.0.0:46759"
seeds = "QmSyEDdMziN9VaJDQVXNLMKwHDxgMNyNW5pHxPq5s2PabR@/ip4/127.0.0.1/tcp/46656"
```

Address can be specified in the old way too.

paralin:
We are rapidly adding more and more interesting transports to libp2p:

 - [grpc](http://github.com/paralin/go-libp2p-grpc)
 - [kcp](http://github.com/paralin/go-libp2p-kcp)

And other routing strategies as well. I'm finding Tendermint is working quite smoothly on top of libp2p in this initial test.

ebuchman:
Wow very cool.

A few questions:

- is it possible to specify priorities for the different streams on the stream multiplexer?
- how is peer behaviour tracked so we know which peers to connect to and recommend?

ebuchman:
Ok looked a bit more closely. I see neither of my questions are relevant here. 

Is authenticated encryption still being used? I see the `MakeSecretConnection` has been dropped.

paralin:
@ebuchman libp2p internally uses secio to encrypt and authenticate the streams between peers, when enabled. You don't need to worry about that at the Tendermint layer.

The overarching goal of all this is to move 95% of the p2p code out of Tendermint, leaving the focus on the block-chain part and not the p2p management. We can then focus on improving the P2P layer separately, and combine efforts with IPFS and other projects in this area.

ebuchman:
Thanks a lot for doing this - really instructive example. 
We did a lot of work on our p2p layer recently, for instance adding support for IDs and improving the peer exchange. A lot of code moved around in an effort to clean it up and provide some additional modularity.

Perhaps we will adopt some libp2p components down the road. In the meantime I think it makes sense to close this PR. 

Thanks!

ebuchman:
Now would be a great time to write an [ADR](https://github.com/tendermint/tendermint/issues/2313) explaining the how and why of this. Especially useful if it can illustrate how we can start integrating components iteratively, and what it would look like to adopt higher level components like "services" (especially how they would compare to existing Reactor design, and what kind of changes would need to be made to accommodate ...).

We could consider adopting some of the smaller components in the short term, but larger components requiring more integration work are probably farther out.

paralin:
Any interest today?

erikgrinaker:
We're currently refactoring the P2P layer, and decided to drop libp2p for now -- we may start experimenting with it as a separate P2P transport after the refactor. You can read some of the background for this in [ADR-061](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-061-p2p-refactor-scope.md).

liamsi:
I keep coming back to this PR and the related issue #988. 
@paralin one question: is there a specific reason to keep the PEX as stated here: https://github.com/tendermint/tendermint/issues/988#issuecomment-352920005 ? Couldn't most of the consensus reactors message handling be replaced by gossipsub topics (e.g. there could be a consensus topic for proposals, votes, a block data topic for block propagation etc)? 

Block data propagation might be a bit more tricky with gossipsub though and it might be easier/more reasonable to have the gossiping managed by the application instead (using block parts and the simple gossiping approach as before). Everything else seems well suited for gossipsub (mempool, consensus etc). One could easily argue that the block propagation in parts should be optional too. In most SDK apps the blocks are tiny anyways. The complexity of block parts might not be justified in most cases. But that is a whole different story.

paralin:
@liamsi These days libp2p has good support for peer-exchange, so that could be used instead.

faddat:
Libp2p would certainly be good.

liamsi:
I think we will do it [in our fork](https://github.com/celestiaorg/lazyledger-core) eventually to not use two different p2p stacks 😬. I'm curious if @paralin would be interested in helping us out with that?

paralin:
@liamsi I'm not using Tendermint at the moment but will check out your project

PR #999, opened 2017-12-22
Add results hash to header

ethanfrey:
Address issue #952 

This is now maintained in state and save to the db.

TODO:
- [x] add `LastResultsHash data.Bytes` to the header
- [x] It should be the merkle root of the `{code, data}` of the results from the previous block.
- [x] add extra methods to the `BlockStore` for storing all the results for a given block. 
- [x] Then, to prove a result, we can make one call to the blockstore to load the BlockResults and easily compute the merkle proof
- [x] Expose modifications to RPC to make it easy for users to see which transactions in a block were invalid

### DIFF
--- a/rpc/core/blocks.go
+++ b/rpc/core/blocks.go
@@ -9,6 +9,7 @@ import (
 )
 
 // Get block headers for minHeight <= height <= maxHeight.
+// Block headers are returned in descending order (highest first).
 //
 // ```shell
 // curl 'localhost:46657/blockchain?minHeight=10&maxHeight=10'
@@ -314,11 +315,10 @@ func Commit(heightPtr *int64) (*ctypes.ResultCommit, error) {
 }
 
 // BlockResults gets ABCIResults at a given height.
-// If no height is provided, it will fetch the latest block.
+// If no height is provided, it will fetch results for the latest block.
 //
-// Results are for the tx of the last block with the same index.
-// Thus response.results[5] is the results of executing
-// getBlock(h-1).Txs[5]
+// Results are for the height of the block containing the txs.
+// Thus response.results[5] is the results of executing getBlock(h).Txs[5]
 //
 // ```shell
 // curl 'localhost:46657/block_results?height=10'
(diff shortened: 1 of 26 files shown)


melekes:
So now we store ABCI results in two places: blockstore and tx_indexer (given user has enabled it). Did not this seem wasteful to you guys?

ethanfrey:
Yup. But it is quite small (just the code (int32) and data (usually nil)) are stored here.
I guess I could go query the tx_indexer for the tx data one by one to reconstruct them, but I think that is a big performance hit for little space savings. 

If this does bother you still, please change where it saves the data. Hash is calculated in state, that stays, but the save could change to use tx_indexer and the rpc also, and then we save a little space.

odeke-em:
@ethanfrey am curious, what's the purpose of putting "Results" in the header?

ebuchman:
> So now we store ABCI results in two places: blockstore and tx_indexer (given user has enabled it). Did not this seem wasteful to you guys?

The point is to enable cheap queries for all the results in a block.

> what's the purpose of putting "Results" in the header?

It's so light clients can verify Merkle proofs of the results of a transaction.

rpc/core/blocks.go
```diff
@@ -312,3 +312,66 @@ func Commit(heightPtr *int64) (*ctypes.ResultCommit, error) {
...
+
+// BlockResults gets ABCIResults at a given height.
+// If no height is provided, it will fetch the latest block.
+//
+// Results are for the tx of the last block with the same index.
+// Thus response.results[5] is the results of executing
+// getBlock(h-1).Txs[5]
+//
+// ```shell
+// curl 'localhost:46657/block_results?height=10'
```
melekes: re ^^: so here we're fetching results for height=9, right?

ebuchman: changed this. now you get what you asked for

ebuchman:
@melekes want to take another look at this when you get a chance? 

Will merge in the meantime

melekes:
LGTM

Issue #950, opened 2017-12-11
state: enforce val set changes <1/3 per block


ebuchman:
See https://github.com/tendermint/tendermint/issues/377#issuecomment-350089253

We also need to document this so apps know.

zramsay:
#1004

PR #1004, opened 2017-12-25
enforce <1/3 validator updates

melekes:
Refs #950

### DIFF
--- a/state/state_test.go
+++ b/state/state_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	abci "github.com/tendermint/abci/types"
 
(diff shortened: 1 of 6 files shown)


ebuchman:
Good start, but we need to check < 1/3 with respect to voting power, not just number of validators.

Also, the change in voting power must be strictly less than 1/3, not <= 1/3

melekes:
So, the algo should look more like:

```
if (>= 1/3 of vals changed) OR newVP/oldVP * 100 >= 33 {
  error
}
```

ebuchman:
We don't even have to check the number of validators, just the voting power.

Basically we should sum the voting power changes and if thats >=1/3 of the current total, then error.

state/state_test.go
```diff
@@ -130,80 +131,53 @@ func TestValidatorSimpleSaveLoad(t *testing.T) {
...
+	const valSetSize = 7
 	tearDown, _, state := setupTestCase(t)
+	state.Validators = genValSet(valSetSize)
+	state.Save()
 	defer tearDown(t)
-	// nolint: vetshadow
-	assert := assert.New(t)
 
-	// change vals at these heights
-	changeHeights := []int64{1, 2, 4, 5, 10, 15, 16, 17, 20}
```
ebuchman: why remove this test ?

state/state_test.go
```diff
@@ -270,6 +245,23 @@ func makeParams(blockBytes, blockTx, blockGas, txBytes,
...
+	tearDown, _, state := setupTestCase(t)
+	defer tearDown(t)
+
+	height := state.LastBlockHeight + 1
+	block := makeBlock(state, height)
+	abciResponses := &ABCIResponses{
+		Height: height,
+		// 1 val (vp: 10) => less than 3 is ok
+		EndBlock: &abci.ResponseEndBlock{ValidatorUpdates: []*abci.Validator{
+			{PubKey: crypto.GenPrivKeyEd25519().PubKey().Bytes(), Power: 3},
```
ebuchman: we should have more test cases

state/state_test.go
melekes: because it was written with an assumption that there is only one validator. I can try and rewrite it of course

ebuchman: right. I think it's ok since we're specifically testing the "save/load at height" behaviour, and having just one validator was an easy way to write the test, and we have other tests of multiple validators. So we should defntly keep it. If you have a nice way to write it with multiple validators, that'd of course be welcome, though now we have to be mindful that it can only change by < 1/3 :)

melekes:
@ebuchman done. feel free to add more test cases if you want (https://github.com/tendermint/tendermint/pull/1004/files#diff-a1153b96e079da8da43231ad48de57a0R248)

PR #1009, opened 2017-12-27
Spec

ebuchman:


docs/specification/new-spec/blockchain.md
```diff
@@ -0,0 +1,389 @@
...
+block.Header.Height == prevBlock.Header.Height + 1
+```
+
+The height is an incrementing integer. The first block has `block.Header.Height == 1`.
+
+### Time
+
+The median of the timestamps of the valid votes in the block.LastCommit.
+Corresponds to the number of nanoseconds, with millisecond resolution, since January 1, 1970.
+Increments with every new block.
```
milosevic: What do you mean with this? Isn't it defined deterministically as median of the votes in lastCommit?

docs/specification/new-spec/blockchain.md
```diff
@@ -0,0 +1,389 @@
...
+```
+
+The cumulative sum of all transactions included in this blockchain.
+
+The first block has `block.Header.TotalTxs = block.Header.NumberTxs`.
+
+### LastBlockID
+
+```
+parts := MakeParts(block, state.ConsensusParams.BlockGossip.BlockPartSize)
```
milosevic: Should it be previousBlock instead of block?

docs/specification/new-spec/blockchain.md
```diff
@@ -0,0 +1,389 @@
...
+block.HeaderLastBlockID == BlockID{
+    SimpleMerkleRoot(block.Header),
```
milosevic: Should it be previousBlock.Header instead of block.Header?

docs/specification/new-spec/blockchain.md
```diff
@@ -0,0 +1,389 @@
...
+The first block has `block.Header.AppHash == []byte{}`.
+
+### ValidatorsHash
+
+```
+block.ValidatorsHash == SimpleMerkleRoot(state.Validators)
+```
+
+Simple Merkle root of the current validator set that is committing the block.
+This can be used to validate the `LastCommit` included in the next block.
```
milosevic: It might be useful adding a short description of the overall process (probably at the beginning of the document) so it's more clear the relation between blocks, and for example why in the current block we have commit from the previous block.

docs/specification/new-spec/blockchain.md
```diff
@@ -0,0 +1,389 @@
...
+May be updated by the application.
+
+### ConsensusParamsHash
+
+```
+block.ValidatorsHash == SimpleMerkleRoot(state.ConsensusParams)
```
milosevic: Should be block.ConsensusHash instead of block.ValidatorsHash.

docs/specification/new-spec/blockchain.md
```diff
@@ -0,0 +1,389 @@
...
+```
+
+Simple Merkle root of the consensus parameters.
+May be updated by the application.
+
+### Proposer
+
+```
+block.Header.Proposer in state.Validators
```
milosevic: There are no Proposer field in block.Header. What do you mean with this?

docs/specification/new-spec/blockchain.md
ebuchman: Yes it's just a note that we need to ensure that votes include timestamps after the one in the proposed block

docs/specification/new-spec/blockchain.md
ebuchman: woops thanks

docs/specification/new-spec/blockchain.md
milosevic: Is it good idea adding Proposer to the blockchain header? In general, I am not sure if adding more consensus stuff into blockchain header makes sense. It makes it more complex to validate, so it's harder job for light client for example. In principle, you only care about validator set and that enough of them sign that they commit to some block.

ebuchman: I tend to agree, but it's a frequent feature request. That said, it's not something we intend for light clients to be able to authenticate easily any time soon.

It's also the "original proposer for the block", not even the proposer for the round the block got committed. So I'm not sure how useful it is. See comments here https://github.com/tendermint/tendermint/issues/952 and perhaps weight in :)

milosevic: I keep overlooking the reward aspect of blockchain. Although as you said, I am not sure original proposer of the block really matters.

milosevic:
Just to repeat here as a general comment, maybe a bit of big picture process is missing. Something similar to Execution paths you have at the end of two sections. But it probably can be added in different PR (I will try to suggest something). But overall great work!

docs/specification/new-spec/blockchain.md
ebuchman: yes

docs/specification/new-spec/blockchain.md
ebuchman: yep

docs/specification/new-spec/blockchain.md
ebuchman: oh damn. I think this means we do need to keep state.LastConsensusParams

ebuchman:
Thanks for review zarko! Will make the fixes and add a better overview in the Readme!

ebuchman:
@milosevic see latest Readme. Feel free to improve by pushing commits directly to this branch

ebuchman:
Actually I'm going to merge this - we can keep working on it in new PRs

Issue #864, opened 2017-11-17
Distinguish between seeds and manual peers


ebuchman:
- [x] try to connect through addrbook before requesting peers from seeds
- [x] distinguish between seeds and manual peers in the config/flags
    - we only use seeds if we can’t connect to peers in the addrbook.
    - we always connect to nodes given in config/flags

related to https://github.com/tendermint/tendermint/issues/862

melekes:
> try to connect through addrbook before requesting peers from seeds

https://github.com/tendermint/tendermint/blob/master/p2p/pex_reactor.go#L248-L292
we're already doing that, no?

melekes:
> distinguish between seeds and manual peers in the config/flags

so we should add another config option & flag like `--manual_seeds`?

zramsay:
:+1: for config option
:-1: for flag

I guess it depends on how often a dev/user will use this feature but in general i'm against adding new flags unless necessary

ebuchman:
> we're already doing that, no?

Yes but that's after we've done DialSeeds.

We should specifically not dial the seeds if theres addrs we can dial, and only fall back on them if we cant connect to anyone.

> so we should add another config option & flag like --manual_peers

Yes. `--p2p.manual_peers`. This should ultimately replace the current use of `--p2p.seeds`, as seeds will become something else (just where you get your addresses from, not persistent peers).

melekes:
> echo "Test connecting via /dial_seeds"
> bash test/p2p/pex/test_dial_seeds.sh $DOCKER_IMAGE $NETWORK_NAME $N $PROXY_APP

do we still want /dial_seeds to be around?

ebuchman:
Handled in https://github.com/tendermint/tendermint/pull/1030

PR #1030, opened 2017-12-29
Distinguish between seeds and manual peers

melekes:
Refs #864


melekes:
- [ ] update changelog

ebuchman:
Looks like a good start.

We should probably call them "persistent_peers"

Did you see the p2p writeup: https://docs.google.com/document/d/188tgD1t7D0oflZhCZGHuLUaFR3Caf6K_dB5WWgPqZPk/edit

Forgot to check in with that when I called it "manual" in the issues.

Also, I think we want this behaviour in the PEX reactor so that any time we run out we can fall back on seeds, not just the one time.

melekes:
> Did you see the p2p writeup: 

no

ebuchman:
I'll bring this home

node/node.go
```diff
@@ -375,6 +382,27 @@ func (n *Node) OnStart() error {
...
+			outbound, inbound, _ := n.sw.NumPeers()
+			if n.IsRunning() && outbound+inbound == 0 {
+				// TODO: ignore error?
+				n.sw.DialPeersAsync(n.addrBook, seeds, false)
+			}
+		}()
+		return nil
+	}
+
+	return n.sw.DialPeersAsync(n.addrBook, seeds, false)
```
odeke-em: Minor nit, could we reverse the condition here to return early?
```go
if n.config.P2P.PexReactor && n.addrBook.Size() <= 0 {
   return n.sw.DialPeersAsync(n.addrBook, seeds, false)
}

// Give some time for PexReactor to connect us to other peers
```

p2p/pex_reactor_test.go
```diff
@@ -118,16 +119,21 @@ func assertSomePeersWithTimeout(t *testing.T, switches []*Switch, checkPeriod, t
 					allGood = false
 				}
 			}
+			remaining -= checkPeriod
```
ebuchman: why was this needed ?

p2p/pex_reactor_test.go
```diff
@@ -180,6 +186,47 @@ func TestPEXReactorAbuseFromPeer(t *testing.T) {
 	assert.True(r.ReachedMaxMsgCountForPeer(peer.NodeInfo().ListenAddr))
 }
 
+func TestPEXReactorUsesSeedsIfNeeded(t *testing.T) {
+	dir, err := ioutil.TempDir("", "pex_reactor")
+	require.Nil(t, err)
+	defer os.RemoveAll(dir) // nolint: errcheck
+
+	book := NewAddrBook(dir+"addrbook.json", false)
```
ebuchman: filepath.Join

p2p/switch.go
```diff
@@ -315,15 +315,15 @@ func (sw *Switch) startInitPeer(peer *peer) {
...
+func (sw *Switch) DialPeersAsync(addrBook *AddrBook, peers []string, persistent bool) error {
+	netAddrs, errs := NewNetAddressStrings(peers)
 	for _, err := range errs {
-		sw.Logger.Error("Error in seed's address", "err", err)
+		sw.Logger.Error("Error in peer's address", "err", err)
 	}
 
 	if addrBook != nil {
-		// add seeds to `addrBook`
+		// add persistent peers to `addrBook`
```
ebuchman: not necessarily persistent

rpc/core/net.go
```diff
@@ -55,19 +54,31 @@ func NetInfo() (*ctypes.ResultNetInfo, error) {
...
 	logger.Info("DialSeeds", "addrBook", addrBook, "seeds", seeds)
-	err := p2pSwitch.DialSeeds(addrBook, seeds)
+	err := p2pSwitch.DialPeersAsync(addrBook, seeds, false)
 	if err != nil {
 		return &ctypes.ResultDialSeeds{}, err
 	}
 	return &ctypes.ResultDialSeeds{"Dialing seeds in progress. See /net_info for details"}, nil
 }
 
+func UnsafeDialPersistentPeers(persistent_peers []string) (*ctypes.ResultDialPersistentPeers, error) {
```
ebuchman: underscore ?

p2p/pex_reactor_test.go
melekes: no benefit. I believe `dir` already has `/` at the end

p2p/pex_reactor_test.go
melekes: because otherwise it will run forever. if you look closely you will see time.After() always resets in a loop

Issue #417, opened 2017-02-22
go 1.8 failing to build against http in golang.org/x/net/http2/go18 on OS X


joeblew99:
i pulled the latest, go get -u github.com/tendermint/tendermint/cmd/tendermint 

But seems its broken on go 1.8.

````


x-MacBook-Pro:tendermint apple$ go env
GOARCH="amd64"
GOBIN="/Users/apple/workspace/go/bin"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/apple/workspace/go"
GORACE=""
GOROOT="/usr/local/go"
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/wp/ff6sz9qs6g71jnm12nj2kbyw0000gp/T/go-build325127552=/tmp/go-build -gno-record-gcc-switches -fno-common"
CXX="clang++"
CGO_ENABLED="1"
PKG_CONFIG="pkg-config"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"


x-MacBook-Pro:tendermint apple$ pwd
/Users/apple/workspace/go/src/github.com/tendermint/tendermint/cmd/tendermint

x-MacBook-Pro:tendermint apple$ go build
# golang.org/x/net/http2
../../../../../golang.org/x/net/http2/go18.go:15: c.Clone undefined (type *tls.Config has no field or method Clone, but does have tls.clone)
../../../../../golang.org/x/net/http2/go18.go:17: undefined: http.Pusher
../../../../../golang.org/x/net/http2/go18.go:20: undefined: http.PushOptions
../../../../../golang.org/x/net/http2/go18.go:31: h1.IdleTimeout undefined (type *http.Server has no field or method IdleTimeout)
../../../../../golang.org/x/net/http2/go18.go:32: h1.IdleTimeout undefined (type *http.Server has no field or method IdleTimeout)
../../../../../golang.org/x/net/http2/go18.go:41: undefined: http.ErrAbortHandler
[...]

PR #1035, opened 2017-12-29
README: document the minimum Go version

odeke-em:
Solidify in writing, the minimum Go version that
we support, as Go1.9.


odeke-em:
The inspiration was https://github.com/tendermint/tendermint/issues/417#issuecomment-354402018 which I believe should be in writing

Issue #867, opened 2017-11-17
Using the trust metric


ebuchman:
- [ ] The trust metric tracks the quality of the peers. 
- [ ] When a peer exceeds a certain quality (for a certain amount of time?), it is marked as vetted in the addrbook.
- [ ] If a vetted peer's quality degrades sufficiently, it is booted, and must prove itself from scratch.
- [ ] If we need to make room for a new vetted peer, we move the lowest scoring vetted peer back to unvetted.
- [ ] If we need to make room for a new unvetted peer, we remove the lowest scoring unvetted peer - possibly only if its below some absolute minimum ? 

Peer quality is tracked in the connection and across the reactors.
Behaviours are defined as one of: 
    - fatal - something outright malicious. we should disconnect and remember them.
    - bad - any kind of timeout, msgs that dont unmarshal, or fail other validity checks, or msgs we didn't ask for or arent expecting
    - neutral - normal correct behaviour. unknown channels/msg types (version upgrades).
    - good - some random majority of peers per reactor sending us useful messages

Here are some preliminary suggestions for how to mark behaviour in reactors (needs updating):

## PEX
- bad:
    - source of too many unavailable or fatal peers
- good:
    - source of peers we end up vetting

## Mempool
- bad:
    - ?
- good:
    - receiving txs that check_tx OK

## Consensus
- bad:
    - receiving votes/parts for old blocks
- good:
    - new parts and votes

## Blockchain
- bad:
    - blocks we didnt ask for.
    - timeouts
- good:
    - blocks we asked for

ebuchman:
We've merged the MarkGood/StopPeer strategy in https://github.com/tendermint/tendermint/pull/1244

We can revisit using the TrustMetric to provide finer granularity at a future date

PR #1038, opened 2017-12-29
867 Using Trust Metrics

caffix:
https://github.com/tendermint/tendermint/issues/867

This branch contains the work toward integrating the trust metric implementation with all the reactors within Tendermint. The reference to the trust metric store is now within the Switch, so all reactors can reach the trust metrics.

I've run into a roadblock though; some of the code (examples below) that needs to be able to punish peers for bad behavior cannot reach the reactor, and cannot mark peers.

https://github.com/tendermint/tendermint/blob/a1a5cb59a70e3075dbfc58614d2f1b83d5ad2a8b/consensus/types/height_vote_set.go#L119

https://github.com/tendermint/tendermint/blob/a1a5cb59a70e3075dbfc58614d2f1b83d5ad2a8b/blockchain/pool.go#L220

 I was hoping we could discuss the best solution to this current limitation within this PR. Thank you.

### DIFF
--- a/blockchain/reactor.go
+++ b/blockchain/reactor.go
@@ -125,10 +125,31 @@ func (bcR *BlockchainReactor) AddPeer(peer p2p.Peer) {
 	}
 	// peer is added to the pool once we receive the first
 	// bcStatusResponseMessage from the peer and call pool.SetPeerHeight
+
+	// Getting the peer metric initializes it in the store
+	tms := bcR.Switch.MetricStore()
+	tm := tms.GetPeerTrustMetric(peer.String(), BlockchainReactorID)
+}
+
+// MarkPeer implements Reactor.
+// It updates a peer's metric with good or bad events taking place within this reactor
+func (bcR *BlockchainReactor) MarkPeer(peer Peer, good bool, events int) {
+	tms := bcR.Switch.MetricStore()
+	tm := tms.GetPeerTrustMetric(peer.String(), BlockchainReactorID)
+
+	if good {
+		tm.GoodEvents(events)
+	} else {
+		tm.BadEvents(events)
+	}
 }
 
 // RemovePeer implements Reactor by removing peer from the pool.
 func (bcR *BlockchainReactor) RemovePeer(peer p2p.Peer, reason interface{}) {
+	tms := bcR.Switch.MetricStore()
+	// Pause tracking of this peer
+	tms.PeerDisconnected(peer.String(), BlockchainReactorID)
+
 	bcR.pool.RemovePeer(peer.Key())
 }
 
--- a/blockchain/reactor.go
+++ b/blockchain/reactor.go
@@ -127,16 +127,37 @@ func (bcR *BlockchainReactor) AddPeer(peer p2p.Peer) {
 	// bcStatusResponseMessage from the peer and call pool.SetPeerHeight
 
 	// Getting the peer metric initializes it in the store
-	tms := bcR.Switch.MetricStore()
-	tm := tms.GetPeerTrustMetric(peer.String(), BlockchainReactorID)
+	if bcR.Switch != nil {
+		if tms := bcR.Switch.MetricStore(); tms != nil {
+			_ = tms.GetPeerTrustMetric(peer.Key(), BlockchainReactorID)
+		}
+	}
 }
 
 // MarkPeer implements Reactor.
 // It updates a peer's metric with good or bad events taking place within this reactor
-func (bcR *BlockchainReactor) MarkPeer(peer Peer, good bool, events int) {
-	tms := bcR.Switch.MetricStore()
-	tm := tms.GetPeerTrustMetric(peer.String(), BlockchainReactorID)
+func (bcR *BlockchainReactor) MarkPeer(peer p2p.Peer, good bool, events, severity int) {
+	if bcR.Switch == nil {
+		return
+	}
 
+	tms := bcR.Switch.MetricStore()
+	if tms == nil {
+		return
+	}
+
+	// Modify the number of events based on severity
+	num := events
+	switch severity {
+	case p2p.Fatal, p2p.Good:
+		num *= 10
+	case p2p.Bad, p2p.Correct:
+		// These result in events * 1
+	case p2p.Neutral:
+		num *= 0
+	}
+
+	tm := tms.GetPeerTrustMetric(peer.Key(), BlockchainReactorID)
 	if good {
 		tm.GoodEvents(events)
 	} else {
@@ -146,9 +167,12 @@ func (bcR *BlockchainReactor) MarkPeer(peer Peer, good bool, events int) {
 
 // RemovePeer implements Reactor by removing peer from the pool.
 func (bcR *BlockchainReactor) RemovePeer(peer p2p.Peer, reason interface{}) {
-	tms := bcR.Switch.MetricStore()
-	// Pause tracking of this peer
-	tms.PeerDisconnected(peer.String(), BlockchainReactorID)
+	if bcR.Switch != nil {
+		if tms := bcR.Switch.MetricStore(); tms != nil {
+			// Pause tracking of this peer
+			tms.PeerDisconnected(peer.Key(), BlockchainReactorID)
+		}
+	}
 
 	bcR.pool.RemovePeer(peer.Key())
 }
@@ -161,10 +185,13 @@ func (bcR *BlockchainReactor) respondToPeer(msg *bcBlockRequestMessage, src p2p.
 	block := bcR.store.LoadBlock(msg.Height)
 	if block != nil {
 		msg := &bcBlockResponseMessage{Block: block}
+
+		bcR.MarkPeer(src, true, 1, p2p.Correct)
 		return src.TrySend(BlockchainChannel, struct{ BlockchainMessage }{msg})
 	}
 
 	bcR.Logger.Info("Peer asking for a block we don't have", "src", src, "height", msg.Height)
+	bcR.MarkPeer(src, false, 1, p2p.Bad)
 
 	return src.TrySend(BlockchainChannel, struct{ BlockchainMessage }{
 		&bcNoBlockResponseMessage{Height: msg.Height},
@@ -246,6 +273,7 @@ FOR_LOOP:
 			peer := bcR.Switch.Peers().Get(peerID)
 			if peer != nil {
 				bcR.Switch.StopPeerForError(peer, errors.New("BlockchainReactor Timeout"))
+				bcR.MarkPeer(peer, false, 1, p2p.Bad)
 			}
 		case <-statusUpdateTicker.C:
 			// ask for status updates
--- a/blockchain/reactor.go
+++ b/blockchain/reactor.go
@@ -107,6 +107,11 @@ func (bcR *BlockchainReactor) OnStop() {
 	bcR.pool.Stop()
 }
 
+// GetID implements Reactor
+func (bcR *BlockchainReactor) GetID() string {
+	return BlockchainReactorID
+}
+
 // GetChannels implements Reactor
 func (bcR *BlockchainReactor) GetChannels() []*p2p.ChannelDescriptor {
 	return []*p2p.ChannelDescriptor{
@@ -125,44 +130,6 @@ func (bcR *BlockchainReactor) AddPeer(peer p2p.Peer) {
 	}
 	// peer is added to the pool once we receive the first
 	// bcStatusResponseMessage from the peer and call pool.SetPeerHeight
-
-	// Getting the peer metric initializes it in the store
-	if bcR.Switch != nil {
-		if tms := bcR.Switch.MetricStore(); tms != nil {
-			_ = tms.GetPeerTrustMetric(peer.Key(), BlockchainReactorID)
-		}
-	}
-}
-
-// MarkPeer implements Reactor.
-// It updates a peer's metric with good or bad events taking place within this reactor
-func (bcR *BlockchainReactor) MarkPeer(peer p2p.Peer, good bool, events, severity int) {
-	if bcR.Switch == nil {
-		return
-	}
-
-	tms := bcR.Switch.MetricStore()
-	if tms == nil {
-		return
-	}
-
-	// Modify the number of events based on severity
-	num := events
-	switch severity {
-	case p2p.Fatal, p2p.Good:
-		num *= 10
-	case p2p.Bad, p2p.Correct:
-		// These result in events * 1
-	case p2p.Neutral:
-		num *= 0
-	}
-
-	tm := tms.GetPeerTrustMetric(peer.Key(), BlockchainReactorID)
-	if good {
-		tm.GoodEvents(events)
-	} else {
-		tm.BadEvents(events)
-	}
 }
 
 // RemovePeer implements Reactor by removing peer from the pool.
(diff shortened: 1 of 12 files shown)


blockchain/reactor.go
```diff
@@ -122,10 +125,55 @@ func (bcR *BlockchainReactor) AddPeer(peer p2p.Peer) {
 	}
 	// peer is added to the pool once we receive the first
 	// bcStatusResponseMessage from the peer and call pool.SetPeerHeight
+
+	// Getting the peer metric initializes it in the store
+	if bcR.Switch != nil {
+		if tms := bcR.Switch.MetricStore(); tms != nil {
```
melekes: when it can be `nil`?

blockchain/reactor.go
```diff
@@ -122,10 +125,55 @@ func (bcR *BlockchainReactor) AddPeer(peer p2p.Peer) {
 	}
 	// peer is added to the pool once we receive the first
 	// bcStatusResponseMessage from the peer and call pool.SetPeerHeight
+
+	// Getting the peer metric initializes it in the store
+	if bcR.Switch != nil {
```
melekes: same question. is this even possible?

blockchain/reactor.go
```diff
@@ -122,10 +125,55 @@ func (bcR *BlockchainReactor) AddPeer(peer p2p.Peer) {
...
+		if tms := bcR.Switch.MetricStore(); tms != nil {
+			_ = tms.GetPeerTrustMetric(peer.Key(), BlockchainReactorID)
```
melekes: let's wrap this into function `InitPeerMetric(peer)`. better than comment

blockchain/reactor.go
```diff
@@ -122,10 +125,55 @@ func (bcR *BlockchainReactor) AddPeer(peer p2p.Peer) {
...
+
+	// Modify the number of events based on severity
+	num := events
+	switch severity {
+	case p2p.Fatal, p2p.Good:
+		num *= 10
+	case p2p.Bad, p2p.Correct:
+		// These result in events * 1
+	case p2p.Neutral:
+		num *= 0
```
melekes: why not `num = 0`?

p2p/pex_reactor.go
```diff
@@ -108,18 +118,55 @@ func (r *PEXReactor) AddPeer(p Peer) {
...
+		return
+	}
+
+	tms := r.Switch.MetricStore()
+	if tms == nil {
+		return
+	}
+
+	// Modify the number of events based on severity
+	num := events
```
melekes: is this the same for all reactors? maybe we should create a sep function then to avoid copy&paste?

p2p/pex_reactor.go
```diff
@@ -276,8 +324,17 @@ func (r *PEXReactor) ensurePeers() {
 	for _, item := range toDial {
 		go func(picked *NetAddress) {
 			_, err := r.Switch.DialPeerWithAddress(picked, false)
```
melekes: this returns `peer, err`. why not just use it instead of `peer := r.addrToSrcPeer(picked)`?

p2p/trust/store.go
```diff
@@ -135,9 +138,27 @@ func (tms *TrustMetricStore) SaveToDB() {
 
 /* Private methods */
 
+// addPeerTrustMetric takes an existing trust metric and associates it with a peer and reactor ID.
+// This version of the method assumes the mutex has already been acquired
+func (tms *TrustMetricStore) addPeerTrustMetric(peer, reactor string, tm *TrustMetric) {
+	if peer == "" || reactor == "" || tm == nil {
```
melekes: not of a fan of this defensive programming. when tm can be `nil`? when peer == ''?

p2p/trust/store.go
```diff
@@ -175,15 +199,19 @@ func (tms *TrustMetricStore) loadFromDB() bool {
 func (tms *TrustMetricStore) saveToDB() {
 	tms.Logger.Debug("Saving TrustHistory to DB", "size", tms.size())
 
-	peers := make(map[string]MetricHistoryJSON)
+	history := make(map[string]map[string]MetricHistoryJSON)
 
-	for key, tm := range tms.peerMetrics {
-		// Add an entry for the peer identified by key
-		peers[key] = tm.HistoryJSON()
+	for reactor, peers := range tms.reactorPeerMetrics {
```
melekes: `peerMetrics`, not `peers`... peers is confusing

blockchain/reactor.go
caffix: the switch was nil during the tests

blockchain/reactor.go
caffix: the switch was nil while running tests, and the MetricStore doesn' t need to be set

blockchain/reactor.go
caffix: sounds good

blockchain/reactor.go
caffix: I'll change it

p2p/pex_reactor.go
caffix: it's currently the same, but in the future, different reactors could handle it differently as it makes since to support their needs

blockchain/reactor.go
```diff
@@ -137,10 +185,13 @@ func (bcR *BlockchainReactor) respondToPeer(msg *bcBlockRequestMessage, src p2p.
 	block := bcR.store.LoadBlock(msg.Height)
 	if block != nil {
 		msg := &bcBlockResponseMessage{Block: block}
+
+		bcR.MarkPeer(src, true, 1, p2p.Correct)
```
ebuchman: lets mark good when we receive a block we asked for, rather than when we receive a request

blockchain/reactor.go
```diff
@@ -222,6 +273,7 @@ FOR_LOOP:
 			peer := bcR.Switch.Peers().Get(peerID)
 			if peer != nil {
 				bcR.Switch.StopPeerForError(peer, errors.New("BlockchainReactor Timeout"))
+				bcR.MarkPeer(peer, false, 1, p2p.Bad)
```
ebuchman: how does marking bad play with a stopped peer ?

consensus/reactor.go
```diff
@@ -179,10 +228,11 @@ func (conR *ConsensusReactor) Receive(chID byte, src p2p.Peer, msgBytes []byte)
 	_, msg, err := DecodeMessage(msgBytes)
 	if err != nil {
 		conR.Logger.Error("Error decoding message", "src", src, "chId", chID, "msg", msg, "err", err, "bytes", msgBytes)
-		// TODO punish peer?
+		conR.MarkPeer(src, false, 1, p2p.Bad)
 		return
 	}
 	conR.Logger.Debug("Receive", "src", src, "chId", chID, "msg", msg)
+	conR.MarkPeer(src, true, 1, p2p.Correct)
```
ebuchman: probably too ambitious to just mark good because we received an arbitrary consensus message. better if its receiving a vote or block part we needed

mempool/reactor.go
```diff
@@ -60,12 +63,57 @@ func (memR *MempoolReactor) GetChannels() []*p2p.ChannelDescriptor {
...
+			_ = tms.GetPeerTrustMetric(peer.Key(), MempoolReactorID)
+		}
+	}
+
 	go memR.broadcastTxRoutine(peer)
 }
 
+// MarkPeer implements Reactor.
+// It updates a peer's metric with good or bad events taking place within this reactor
+func (memR *MempoolReactor) MarkPeer(peer p2p.Peer, good bool, events, severity int) {
```
ebuchman: can these be deduplicated? seems like they're all the same except the reactor ID that gets passed at the end.

Also, what if we made Mark a method on Peer, rather than on Reactor? Is there some reason that won't work?

mempool/reactor.go
```diff
@@ -84,6 +132,7 @@ func (memR *MempoolReactor) Receive(chID byte, src p2p.Peer, msgBytes []byte) {
 		if err != nil {
 			memR.Logger.Info("Could not check tx", "tx", msg.Tx, "err", err)
 		}
+		memR.MarkPeer(src, true, 1, p2p.Correct)
```
ebuchman: should only happen if there's no error. or better yet, only if the asynchronous CheckTx call returns a success. maybe we can use the CheckTx callback argument for that

p2p/pex_reactor.go
```diff
@@ -108,18 +118,55 @@ func (r *PEXReactor) AddPeer(p Peer) {
 	} else { // For inbound connections, the peer is its own source
 		addr, err := NewNetAddressString(p.NodeInfo().ListenAddr)
 		if err != nil {
-			// peer gave us a bad ListenAddr. TODO: punish
+			// peer gave us a bad ListenAddr
+			r.MarkPeer(p, false, 1, Bad)
 			r.Logger.Error("Error in AddPeer: invalid peer address", "addr", p.NodeInfo().ListenAddr, "err", err)
 			return
 		}
+		r.MarkPeer(p, true, 1, Correct)
```
ebuchman: AddPeer only runs once per connection, so not sure how useful this one is

p2p/pex_reactor.go
```diff
@@ -293,6 +350,18 @@ func (r *PEXReactor) ensurePeers() {
 	}
 }
 
+// addrToSrcPeer return the Peer that provided the address in the param
+func (r *PEXReactor) addrToSrcPeer(addr *NetAddress) Peer {
+	peers := r.Switch.Peers().List()
```
ebuchman: so this will only return active peers.

should we consider marking good for currently inactive peers if they were the source of a peer we just dialed?

p2p/switch.go
```diff
@@ -30,12 +31,22 @@ const (
 	reconnectBackOffBaseSeconds = 3
 )
 
+// The following constants modify the number of events sent to a metric based on severity
+const (
+	Fatal = iota
```
ebuchman: These could use a prefix for clarity when referred to. Maybe `PeerMarkFatal` etc.

p2p/pex_reactor.go
ebuchman: addrToSrcPeer gets the source peer that told us about the peer we're trying to dial.

p2p/trust/store.go
ebuchman: what don't you like? possibly masks bugs?

melekes: exactly because of my questions. it creates additional flow or branches of execution (not sure how to explain this), but every new dev will be asking him/herself these questions: when tm can be nil? when peer == ''? when tm == nil?

I always tend to prefer null object https://github.com/tendermint/tendermint/blob/master/types/nop_event_bus.go in such cases or better not have this nil branch at all.

mempool/reactor.go
caffix: I can move the mark method onto the Peer as long as the reactor is a parameter. This will require the ReactorIDs that are currently constants to be changed into fields within the reactor structs.

p2p/pex_reactor.go
caffix: The trust metric implementation does not continue tracking peers that are no longer active

p2p/switch.go
caffix: Agreed.

mempool/reactor.go
caffix: Or, we could just pass the ReactorID to the peer mark method.

caffix: Actually, we do need the reactor, since it has the reference to the Switch.

p2p/pex_reactor.go
caffix: exactly

p2p/trust/store.go
caffix: I'll make it more clear.

mempool/reactor.go
ebuchman: Why does it need the switch, just for access to the metric store? Maybe there's a nice way to make that accessible within the peer ?

blockchain/reactor.go
caffix: Sure.

blockchain/reactor.go
caffix: Thanks, I'll take care of it.

consensus/reactor.go
caffix: Okay.

p2p/pex_reactor.go
caffix: I'll remove it.

mempool/reactor.go
caffix: Will do.

mempool/reactor.go
caffix: Yes, for access to the metric store. Would you like the peer to have a metric store reference as well? In that case, we would just need to pass the reactor ID to the mark method.

p2p/trust/store.go
caffix: Input validation is a bit of a habit for me, but I can remove the check if you wish.

mempool/reactor.go
ebuchman: yeh I think that might make more sense - peers carry their own metric store

caffix: great! it will be in the next commit

p2p/peer.go
```diff
@@ -9,10 +9,20 @@ import (
...
 	crypto "github.com/tendermint/go-crypto"
 	wire "github.com/tendermint/go-wire"
+	"github.com/tendermint/tendermint/p2p/trust"
 	cmn "github.com/tendermint/tmlibs/common"
 	"github.com/tendermint/tmlibs/log"
 )
 
+// The following constants modify the number of events sent to a metric based on severity
+const (
+	PeerMarkFatal = iota
```
melekes: wanna make a type alias for this?

```
type severity int

const ( 
  PeerMarkFatal severity = iota

Mark(reactorID string, good bool, events int, s severity)
```

p2p/peer.go
```diff
@@ -317,11 +329,47 @@ func (p *peer) NodeInfo() *NodeInfo {
...
+	p.metricStore = tms
+}
+
 // Status returns the peer's ConnectionStatus.
 func (p *peer) Status() ConnectionStatus {
 	return p.mconn.Status()
 }
 
+// It updates a peer's metric with good or bad events taking place within this reactor
+func (p *peer) Mark(reactorID string, good bool, events, severity int) {
```
melekes: I'd create 2 funcs

```
MarkAsGoodNEvents(reactorID string, events int, s severity)
MarkAsBadNEvents(reactorID string, events int, s severity)
```

src.Mark(conR.GetID(), false, 1, p2p.PeerMarkBad) is not the best API because you need to look docs or method signature in order to figure out what false and 1 stands for.

```
src.Mark(memR.GetID(), true, 1, p2p.PeerMarkCorrect)
```

versus

```
src.MarkAsGoodNEvents(memR.GetID(), 1, p2p.PeerMarkCorrect)
```

melekes: Other question is: should peers be aware they are part of the reactors. i.e. maybe this `reactorID string` should be really `peerTrustMetricValue string`?

p2p/peer.go
caffix: Sure.

p2p/peer.go
caffix: Okay, I have created the two separate methods, and renamed the parameter to trustMetricID.

ebuchman:
New simpler approach for this in the sort term is https://github.com/tendermint/tendermint/issues/1147.  We can revisit strategy for using the trust metric later in the year

PR #1039, opened 2017-12-29
Spec of consensus/gossip protocol

milosevic:


docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,210 @@
...
+
+```
+type ProposalHeartbeatMessage struct {
+    Heartbeat *types.Heartbeat
+}
+```
+
+### Heartbeat
+Heartbeat contains validator information (address and index),
+height, round and sequence number (?). It is signed by the private key of the validator.
```
melekes: (?)

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,210 @@
...
+
+Validators in Tendermint communicate by peer-to-peer gossiping protocol. Each validator is connected 
+only to a subset of processes called peers. By the gossiping protocol, a validator send to its peers 
+all needed  information (`ProposalMessage`, `VoteMessage` and `BlockPartMessage`) so they can 
+reach agreement on some block, and also obtain the content of the chosen block (block parts). As part of
+the gossiping protocol, processes also send auxiliary messages that inform peers about the 
+executed steps of the core consensus algorithm (`NewRoundStepMessage` and `CommitStepMessage`), and 
+also messages that inform peers what votes the process has seen (`HasVoteMessage`,
+`VoteSetMaj23Message` and `VoteSetBitsMessage`).
+
```
ebuchman: These opening paragraphs are wonderful <3

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,210 @@
...
+`VoteSetMaj23Message` and `VoteSetBitsMessage`).
+ 
+We now describe the content of each message exchanged during Tendermint consensus protocol.
+
+## ProposalMessage
+ProposalMessage is sent when a new block is proposed. It is a suggestion of what the 
+next block in the blockchain should be.
+```
+type ProposalMessage struct {
+	Proposal *types.Proposal
```
ebuchman: let's drop all pointers and package names.

The goal with the spec is to lay out the data structures in a way that makes their encoding clear. We don't want the fact that it's a pointer to change the encoding, so it shouldn't be relevant (currently it does matter, but we plan to fix that!)

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,210 @@
...
+### Proposal
+Proposal contains height and round for which this proposal is made, BlockID as a unique identifier of proposed
+block, timestamp, and two fields (POLRound and POLBlockID) that are needed for termination of the consensus.
+The message is signed by the validator private key.  
+
+```
+type Proposal struct {
+	Height           int64            
+	Round            int              
+	Timestamp        time.Time
```
ebuchman: int64 (thats what its encoded as)

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,210 @@
...
+
+```
+type Proposal struct {
+	Height           int64            
+	Round            int              
+	Timestamp        time.Time        
+	BlockID          BlockID      
+	POLRound         int              
+	POLBlockID       BlockID          
+	Signature        crypto.Signature
```
ebuchman: Let's drop the pkg name prefixes. This spec is defining the data structures, not referencing the code, and it all lives in one namespace. I defined Signature in blockchain.md

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,210 @@
...
+	BlockID          BlockID      
+	POLRound         int              
+	POLBlockID       BlockID          
+	Signature        crypto.Signature 
+}
+```
+
+NOTE: In the current version of the Tendermint, the consensus value in proposal is represented with
+PartSetHeader, and with BlockID in vote message. It should be aligned as suggested in this spec as 
+BlockID contains PartSetHeader.
```
ebuchman: ++

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,210 @@
...
+VoteMessage is sent to vote for some block (or to inform others that a process does not vote in the current round).
+
+```
+type VoteMessage struct {
+	Vote *types.Vote
+}
+```
+
+### Vote
+Vote contains validator's information (validator address and index), height and round for which the vote is sent,
```
ebuchman: Vote is already defined in blockchain.md

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,210 @@
...
+```
+## BlockPartMessage
+BlockPartMessage is sent when gossipping a piece of the proposed block. It contains height, round 
+and the block part. 
+
+```
+type BlockPartMessage struct {
+	Height int64
+	Round  int
+	Part   *types.Part
```
ebuchman: we should define `Part` in `encoding.md` . that file is for all serialization and crypto-digest definitions

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,210 @@
...
+}
+```
+
+### Heartbeat
+Heartbeat contains validator information (address and index),
+height, round and sequence number (?). It is signed by the private key of the validator. 
+
+```
+type Heartbeat struct {
+	ValidatorAddress data.Bytes
```
ebuchman: []byte

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,210 @@
...
+## NewRoundStepMessage
+NewRoundStepMessage is sent for every step transition during the core consensus algorithm execution. It is 
+used in the gossip part of the Tendermint protocol to inform peers about a current height/round/step 
+a process is in.
+
+```
+type NewRoundStepMessage struct {
+	Height                int64
+	Round                 int
+	Step                  cstypes.RoundStepType
```
ebuchman: lets be clear about the basic type here

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,210 @@
...
+## CommitStepMessage
+CommitStepMessage is sent when an agreement on some block is reached. It contains height for which agreement
+is reached, block parts header that describes the decided block and is used to obtain all block parts,
+and a bit array of the block parts a process currently has, so its peers can know what parts 
+it is missing so they can send them. 
+
+```
+type CommitStepMessage struct {
+	Height           int64
+	BlockPartsHeader types.PartSetHeader
```
ebuchman: lets make it BlockID ?

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,210 @@
...
+	BlockParts       *cmn.BitArray
```
ebuchman: Mmm - we need to define BitArray and how it's encoded!

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,210 @@
...
+}
+```
+
+TODO: Maybe use BlockID instead of BlockPartsHeader for symmetry.
```
ebuchman: ++

docs/specification/new-spec/consensus.md
ebuchman: Though I might add a note about only proposal/vote being signed, and about how the gossip msgs are used to determine how we send data to other peers

docs/specification/new-spec/consensus.md
milosevic: I was wondering what is the sequence field, but it's probably as it says sequence number :)

docs/specification/new-spec/consensus.md
milosevic: What do you mean? Should we define RoundStepType?It's somehow implementation detail and can change, so I am not sure if it should be part of spec.

ebuchman: Let's just give a concrete type - int8

milosevic:
I have tried to address your comment. The encoding part needs to be improved (BitArray and Part), but it's probably easier for someone else to define it. Please take a look @ebuchman @melekes .

docs/specification/new-spec/consensus.md
milosevic: You mean we shouldn't use int type? Because we use it in Vote so I thought that's fine.

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,197 @@
+# Tendermint Consensus
+
+Tendermint consensus is a distributed protocol executed by validator processes to agree on
+the next block to be added to the Tendermint blockchain. The protocol proceeds in rounds, where
```
adrianbrink: `... to be added to a blockchain.` 

Theoretically it could be any blockchain. For example, Polkdadot might use Tendermint consensus.

Also, we should be consistent with naming and always call it `Tendermint Consensus`.

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,197 @@
+# Tendermint Consensus
+
+Tendermint consensus is a distributed protocol executed by validator processes to agree on
+the next block to be added to the Tendermint blockchain. The protocol proceeds in rounds, where 
+each round is a try to reach agreement on the next block. A round starts by having a dedicated 
+process (called proposer) suggesting to other processes what should be the next block with 
+the `ProposalMessage`.
+The processes respond by voting for a block with `VoteMessage` (there are two kinds of vote messages, prevote 
+and precommit votes). Note that a proposal message is just a suggestion what the next block should be; a 
+validator might vote with a `VoteMessage` for a different block. If in some round, enough number
```
adrianbrink: `... in some round enough processes vote for the same block, ... `

docs/specification/new-spec/consensus.md
```diff
@@ -0,0 +1,197 @@
...
+Tendermint consensus is a distributed protocol executed by validator processes to agree on
+the next block to be added to the Tendermint blockchain. The protocol proceeds in rounds, where 
+each round is a try to reach agreement on the next block. A round starts by having a dedicated 
+process (called proposer) suggesting to other processes what should be the next block with 
+the `ProposalMessage`.
+The processes respond by voting for a block with `VoteMessage` (there are two kinds of vote messages, prevote 
+and precommit votes). Note that a proposal message is just a suggestion what the next block should be; a 
+validator might vote with a `VoteMessage` for a different block. If in some round, enough number 
+of processes vote for the same block, then this block is committed and later added to the blockchain. 
+`ProposalMessage` and `VoteMessage` are signed by the private key of the validator.
```
adrianbrink: `... by the private key of each process.` We haven't explained validator before.

PR #1041, opened 2017-12-30
Feature/buildprocess

Greg-Szabo:
Extended the Makefile check_tools with make-compatible checks:
1. Check if go exists
2. Warn if it's not the version specified for deterministic builds
3. Warn if PATH doesn't contain GOPATH/bin
Also:
- Set new build flags to help with deterministic builds
- Minor fix to evaluate static variables only once

Most changes are `make` commands, so we don't rely on external dependencies like bash.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,11 +3,12 @@ GOTOOLS = \
 	github.com/Masterminds/glide \
 	github.com/tcnksm/ghr \
 	gopkg.in/alecthomas/gometalinter.v2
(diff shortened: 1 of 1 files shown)


Greg-Szabo:
Sorry, I opened it against master by accident. I'll reopen for develop.

ebuchman:
Note you can edit the branch without clsoing/reopening the PR. just hit the Edit button!

PR #1042, opened 2017-12-30
Feature/buildprocess

Greg-Szabo:
Extended the Makefile check_tools with make-compatible checks:

1. Check if go exists
1. Warn if it's not the version specified for deterministic builds
1. Warn if PATH doesn't contain GOPATH/bin
Also:
- Set new build flags to help with deterministic builds
- Minor fix to evaluate static variables only once

Most changes are `make` commands, so we don't rely on external dependencies like bash.


Makefile
```diff
@@ -1,13 +1,17 @@
...
-GOTOOLS_CHECK = gox glide ghr gometalinter.v2
+GO_MIN_VERSION := 1.9.2
 PACKAGES=$(shell go list ./... | grep -v '/vendor/')
 BUILD_TAGS?=tendermint
 TMHOME = $${TMHOME:-$$HOME/.tendermint}
-BUILD_FLAGS = -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short HEAD`"
+GOPATH ?= $(shell go env GOPATH)
+GOROOT ?= $(shell go env GOROOT)
+GOGCCFLAGS ?= $(shell go env GOGCCFLAGS)
+BUILD_FLAGS = -gcflags "-trimpath $(GOPATH)" -asmflags "-trimpath $(GOPATH)" -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short=7 HEAD`"
```
melekes: you know this is used by devs to build? building deterministic binary is in https://github.com/tendermint/tendermint/blob/master/scripts/dist_build.sh#L34

not sure we need these flags

Greg-Szabo: I wasn't aware that some builds are built from other scripts, it's not mentioned in the documentation anywhere. Also, if that's the case, then I'm guessing none of our distributed binaries were ever tested. Or does CircleCI use that script too?
According to the docs, this is the place where everyone builds: devs, automation, end-user. I've been using this Makefile to build all the binaries and packages on our testnet.
I think we should stick to this Makefile for builds and that shell script should call this Makefile.
I would like to have one place where the build process is code-documented. It's not fair telling people "it works on my machine" and pointing them to another build process. Especially since we don't have deterministic builds.

Also, please let's not use bash in the build process at all. You can't query its result and you can't segment it properly. Even the `Makefile` with its ancient syntax is a better choice.

I need to have a build process that everyone can point to and say: yes, I've done it that way. It will still be different because people use different operating systems and paths and whatnot, but at least the commands for the build should be the same.

The flags are used to check the settings for deterministic builds. Unfortunately if GOROOT is not set, go will compile the default GOROOT into the binary so it needs to be retrieved and checked. Also unfortunately, the "trimpath" flag can only be used once and it is not perfect. We can use it to remove the GOPATH prefix from the strings. GOPATH had to be read for this or else it would run with empty strings instead of failing.
GOGCCFLAGS is also read for deterministic build checks: the parameter `-fdebug-prefix-map` needs to be available for the C compiler for deterministic builds (also the `-gno-record-gcc-switches` switch but go only checks for the existence of the first one and automatically assumes the second one) or the temporary work directory will be hard-coded in any C binary. (Unfortunately our libraries use C.) So if I want to warn the executor that the build is not going to be deterministic, I need that information and this is easier than checking the C c ompiler available flags directly. (GO dynamically fills up that variable based on the available compiler so this is safer.)

melekes: > I wasn't aware that some builds are built from other scripts, it's not mentioned in the documentation anywhere

sorry about that. our dev documentation pretty much does not exist.

> Or does CircleCI use that script too?

no, but they are using the same code

> I would like to have one place where the build process is code-documented.

See `make dist` command. If you will be able to put the content of scripts/dist.sh and scritps/dist_build.sh into Makefile without making it unreadable, that would be great.

> but at least the commands for the build should be the same.

agree

Greg-Szabo: Thanks for the insight Anton. Here's what I see I can improve if you think it's worth it:
1. Enrich the Makefile to cross-compile any version of the code. `make all` would still only compile the current architecture, while `make dist` would compile to all architectures. (Or the selected ones based on an environment variable, just like in the shell script).
2. I would like to separate packaging (or as it's called here "distribution") and the build process. I think the repository should contain all necessary tools to build the code in the Makefile, while the packaging/distribution/deployment process should be separated (because the community doesn't really benefit from our ways of distribution) in a separate repo. This also allows me to automate all the deployment, including creating ZIP files, deploying to the `/downloads` page, etc.

Does this sound useful?

Greg-Szabo: Few more things (I just checked dist.sh). This is my target:
- The Makefile should be able to build the binary in multiple environments (minimal amount of dependency from the OS) and possibly warn if a build is not going to be deterministic.
- Running the build process in a docker container should ensure deterministic builds. This should be possible with the go<version> official container. (Possibly we shouldn't need to build a container for the build process just run `git clone && make all` in the container.)
- The code-documentation for the docker container should be outside of the tendermint repo, most probably in the `tools` repo. I think the code repository should definitely focus on the minimum and the added tools repo should be able to give examples to the community on how they can use our code. This way if someone just wants a developer build, they do `make` on the tendermint repo, whereas if they want a deterministic build, they do a `make` on the tools repo. This way the additional dependencies that developers don't care about are safely separated.

Greg-Szabo:
I'm closing this pull request with the following take-aways:
1. The feature/buildprocess branch will keep working on deterministic builds and build improvements in general. We need a stable and robust build process to be able to build on top of it. (packaging, CI, etc)
2. It seems that a pull request has a better option to trigger a conversation about how we do things than just stating what's happening in a branch. I hope we can change this so I don't have to keep opening pull requests for input.
3. At one point, a version of these improvements need to be introduced to develop or else we cannot build CI on top of it.

Greg-Szabo:
Apologies for the confusion, I wasn't sure if I can rebase the feature/buildprocess branch properly on develop. It was successful, so we can continue discussions here.
Latest changes involve moving the built binary directly to `$GOPATH/bin`, instead of leaving it in the `build/` folder. This can make the `install` target obsolete which is desirable since `gox` doesn't support it.

Greg-Szabo:
Ok, reopening the PR after fixing the branch to spon off from the latest `develop`.

Issue #1021, opened 2017-12-28
lite: memStoreProvider should use binary/range search for GetByHeight


odeke-em:
Just by reading the code I noticed that the only place that we seem to be mutating is by appending to it then subsequently sorting byHeight at
https://github.com/tendermint/tendermint/blob/d14ec7d7d2f407433070665536994e0a1eafab4b/lite/memprovider.go#L54-L55

However, GetByHeight performs a linear search each time
https://github.com/tendermint/tendermint/blob/d14ec7d7d2f407433070665536994e0a1eafab4b/lite/memprovider.go#L64-L70
which in the code above always assumes the best case. IMO this doesn't do much to guarantee us speed and cuts out the benefits of sorting each time. Let's make this a binary/range search because no matter how many items are inserted in there, the worst case for the binary/range search will always be O(log(n)) while the current code is linear and to show the impact of this change:
If there are are 1million byHeight entries:

n=len(m.byHeight)|maxIterations(Linear search)|maxIterations(Binary search)
---|---|---
10|10|4
100|100|7
1000|1000|10
10000|10000|14
100000|100000|17
1000000|1000000|20

and here is the code that we can use to fix this, it is also shorter and magnitudes more efficent
```go
// Since it is already sorted by largest to lowest in StoreCommit
i := sort.Search(len(m.byHeight), func(j int) bool {
    return m.byHeight[j].Height() <= h
})
```

and for more information, please read https://en.wikipedia.org/wiki/Binary_search_algorithm

ethanfrey:
Yes, Emanuel, you are right.

This was designed for clients that wouldn't have many headers, maybe 500 tops. And I wasn't so worried about in memory iteration performance.

But if you want to make a proper binary search and add a few tests, it would be a great addition. Also maybe optimize the insert. 

My code went for simplicity and correctness first. Performance second. Designed to add it later, but not always implemented.

ethanfrey:
Also the majority of the cases, we will search for the most recent or one of the eg. 5 most recent headers. In this case, I think a reverse linear search may be faster than a binary search

odeke-em:
@ethanfrey thank you for the background information, and in deed you are spot on, when we have little data, linear search shines as you'll see in this PR benchmark shoot out results in PR https://github.com/tendermint/tendermint/pull/1043.

I've also optimized the insertion to lazily sort(so not sorting on every insertion).

[...]

PR #1043, opened 2017-12-30
lite: memStoreProvider GetHeightBinarySearch method + fix ValKeys.signHeaders

odeke-em:
Updates #1021

* Implement a GetHeightBinarySearch method that looks for
the height using the binary search algorithm guaranteeing
worst case iteration time of O(log2(n))
whereas
worst case iteration time of O(n) for the current linear search

So if n we had 500 commits stored by height and sorted, to
trigger the worst case scenario for each, pass in
the most negative height you can find e.g. -1
Linear search: 500 iterations
Binary search: 9 iterations

with n=1000, qHeight = -1
Linear search: 1000 iterations
Binary search: 10 iterations

with n=1e6, qHeight = -1
Linear search: 1e6 iterations
Binary search: 20 iterations

Of course there are realistic expectations e.g. a max of
commits that may be saved so linear search might be useful
for very small size set because it has less preparing overhead
and only ~2 types of comparisons, but nonetheless binary search
shines as soon as we start to hit say 50 commits to search from
as you can see below:

```shell
$ go test -v -run=^$ -bench=MemStore
goos: darwin
goarch: amd64
pkg: github.com/tendermint/tendermint/lite
BenchmarkMemStoreProviderGetByHeightLinearSearch5-4     300000        6491 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightLinearSearch50-4      200000       12064 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightLinearSearch100-4      50000       32987 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightLinearSearch500-4       5000      395521 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightLinearSearch1000-4	       500     2940724 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightBinarySearch5-4     300000        6281 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightBinarySearch50-4      200000       10117 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightBinarySearch100-4     100000       18447 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightBinarySearch500-4      20000       89029 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightBinarySearch1000-4	      5000      265719 ns/op      1600 B/op       15 allocs/op
PASS
ok    github.com/tendermint/tendermint/lite 86.614s
$ go test -v -run=^$ -bench=MemStore
goos: darwin
goarch: amd64
pkg: github.com/tendermint/tendermint/lite
BenchmarkMemStoreProviderGetByHeightLinearSearch5-4     300000        6779 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightLinearSearch50-4      100000       12980 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightLinearSearch100-4      30000       43598 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightLinearSearch500-4       5000      377462 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightLinearSearch1000-4	       500     3278122 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightBinarySearch5-4     300000        7084 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightBinarySearch50-4      200000        9852 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightBinarySearch100-4     100000       19020 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightBinarySearch500-4      20000       99463 ns/op      1600 B/op       15 allocs/op
BenchmarkMemStoreProviderGetByHeightBinarySearch1000-4	      5000      259293 ns/op      1600 B/op       15 allocs/op
PASS
ok    github.com/tendermint/tendermint/lite 86.204s
```

which gives
```shell
$ benchstat old.txt new.txt
name                               old time/op    new time/op    delta
MemStoreProviderGetByHeight5-4       6.63µs ± 2%    6.68µs ± 6%   ~             (p=1.000 n=2+2)
MemStoreProviderGetByHeight50-4      12.5µs ± 4%    10.0µs ± 1%   ~             (p=0.333 n=2+2)
MemStoreProviderGetByHeight100-4     38.3µs ±14%    18.7µs ± 2%   ~             (p=0.333 n=2+2)
MemStoreProviderGetByHeight500-4      386µs ± 2%      94µs ± 6%   ~             (p=0.333 n=2+2)
MemStoreProviderGetByHeight1000-4    3.11ms ± 5%    0.26ms ± 1%   ~             (p=0.333 n=2+2)
```

If need be we can make a hybrid algorithm that switches between the
linear and binary search depending on the number of items.
This is reminiscent of Python's TimSort algorithm.

### DIFF
--- a/lite/helpers.go
+++ b/lite/helpers.go
@@ -78,6 +78,9 @@ func (v ValKeys) signHeader(header *types.Header, first, last int) *types.Commit
 
 	// fill in the votes we want
 	for i := first; i < last; i++ {
+		if i >= len(v) {
+			break
+		}
 		vote := makeVote(header, vset, v[i])
 		votes[vote.ValidatorIndex] = vote
 	}
--- a/lite/helpers.go
+++ b/lite/helpers.go
@@ -77,10 +77,7 @@ func (v ValKeys) signHeader(header *types.Header, first, last int) *types.Commit
 	vset := v.ToValidators(1, 0)
 
 	// fill in the votes we want
-	for i := first; i < last; i++ {
-		if i >= len(v) {
-			break
-		}
+	for i := first; i < last && i < len(v); i++ {
 		vote := makeVote(header, vset, v[i])
 		votes[vote.ValidatorIndex] = vote
 	}
--- a/lite/memprovider.go
+++ b/lite/memprovider.go
@@ -89,7 +89,7 @@ func (m *memStoreProvider) GetByHeightBinarySearch(h int64) (FullCommit, error) 
 	var midFC FullCommit
 	// Our goal is to either find:
 	//   * item ByHeight with the query
-	//   * heighest height with a height <= query
+	//   * greatest height with a height <= query
 	for low <= high {
 		mid = int(uint(low+high) >> 1) // Avoid an overflow
 		midFC = m.byHeight[mid]
--- a/lite/memprovider.go
+++ b/lite/memprovider.go
@@ -60,12 +60,30 @@ func (m *memStoreProvider) StoreCommit(fc FullCommit) error {
 
 // GetByHeight returns the FullCommit for height h or an error if the commit is not found.
 func (m *memStoreProvider) GetByHeight(h int64) (FullCommit, error) {
-	m.mtx.Lock()
-	defer m.mtx.Unlock()
+	// By heuristics, GetByHeight with linearsearch is fast enough
+	// for about 50 keys but after that, it needs binary search.
+	// See https://github.com/tendermint/tendermint/pull/1043#issue-285188242
+	m.mtx.RLock()
+	n := len(m.byHeight)
+	m.mtx.RUnlock()
+
+	if n <= 50 {
+		return m.getByHeightLinearSearch(h)
+	}
+	return m.getByHeightBinarySearch(h)
+}
+
+func (m *memStoreProvider) sortByHeightIfNecessaryLocked() {
 	if !m.sorted {
 		sort.Sort(m.byHeight)
 		m.sorted = true
 	}
+}
+
+func (m *memStoreProvider) getByHeightLinearSearch(h int64) (FullCommit, error) {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+	m.sortByHeightIfNecessaryLocked()
 	// search from highest to lowest
 	for i := len(m.byHeight) - 1; i >= 0; i-- {
 		if fc := m.byHeight[i]; fc.Height() <= h {
@@ -75,14 +93,10 @@ func (m *memStoreProvider) GetByHeight(h int64) (FullCommit, error) {
 	return FullCommit{}, liteErr.ErrCommitNotFound()
 }
 
-// GetByHeight returns the FullCommit for height h or an error if the commit is not found.
-func (m *memStoreProvider) GetByHeightBinarySearch(h int64) (FullCommit, error) {
+func (m *memStoreProvider) getByHeightBinarySearch(h int64) (FullCommit, error) {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()
-	if !m.sorted {
-		sort.Sort(m.byHeight)
-		m.sorted = true
-	}
+	m.sortByHeightIfNecessaryLocked()
 	low, high := 0, len(m.byHeight)-1
 	var mid int
 	var hmid int64
@@ -126,12 +140,13 @@ func (m *memStoreProvider) GetByHash(hash []byte) (FullCommit, error) {
 
 // LatestCommit returns the latest FullCommit or an error if no commits exist.
 func (m *memStoreProvider) LatestCommit() (FullCommit, error) {
-	m.mtx.RLock()
-	defer m.mtx.RUnlock()
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
 
 	l := len(m.byHeight)
 	if l == 0 {
 		return FullCommit{}, liteErr.ErrCommitNotFound()
 	}
+	m.sortByHeightIfNecessaryLocked()
 	return m.byHeight[l-1], nil
 }
--- a/lite/memprovider.go
+++ b/lite/memprovider.go
@@ -14,6 +14,8 @@ type memStoreProvider struct {
 	// btree would be more efficient for larger sets
 	byHeight fullCommits
 	byHash   map[string]FullCommit
+
+	sorted bool
 }
 
 // fullCommits just exists to allow easy sorting
@@ -52,7 +54,7 @@ func (m *memStoreProvider) StoreCommit(fc FullCommit) error {
 	defer m.mtx.Unlock()
 	m.byHash[key] = fc
 	m.byHeight = append(m.byHeight, fc)
-	sort.Sort(m.byHeight)
+	m.sorted = false
 	return nil
 }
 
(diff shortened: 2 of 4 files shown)


lite/helpers.go
```diff
@@ -78,6 +78,9 @@ func (v ValKeys) signHeader(header *types.Header, first, last int) *types.Commit
 
 	// fill in the votes we want
 	for i := first; i < last; i++ {
```
melekes: let's add `&& i < len(v)` instead of if condition

lite/memprovider.go
```diff
@@ -52,25 +54,64 @@ func (m *memStoreProvider) StoreCommit(fc FullCommit) error {
...
+			return midFC, nil
+		case hmid < h:
+			low = mid + 1
+		case hmid > h:
+			high = mid - 1
+		}
+	}
+
+	if high >= 0 {
+		if highFC := m.byHeight[high]; highFC.Height() < h {
```
melekes: > * heighest height with a height <= query

why `<` here then and not `<=`?

lite/performance_test.go
```diff
@@ -115,3 +116,112 @@ func benchmarkCertifyCommit(b *testing.B, keys lite.ValKeys) {
...
+
+var (
+	fcs5, h5       = genFullCommits(nil, nil, 5)
+	fcs50, h50     = genFullCommits(fcs5, h5, 50)
+	fcs100, h100   = genFullCommits(fcs50, h50, 100)
+	fcs500, h500   = genFullCommits(fcs100, h100, 500)
+	fcs1000, h1000 = genFullCommits(fcs500, h500, 1000)
+)
+
+func BenchmarkMemStoreProviderGetByHeightLinearSearch5(b *testing.B) {
```
melekes: I'd rather want to see test instead of benchmark. It's clean for any dev that binary search is faster than linear search

lite/helpers.go
odeke-em: SGTM, thanks!

lite/memprovider.go
odeke-em: If = existed, we'd have already found it inside the binary search.

lite/performance_test.go
odeke-em: Not really though, which is what prompted this change. Please see the discussion for this issue where it is explained when to use the binary search and in fact even the benchmarks illustrate that binary search starts being useful when are at say 50 items but for the common case, linear search will beat it.

odeke-em: @melekes some tests already exist here https://github.com/tendermint/tendermint/blob/edbcd8d684450c182cf8e129cf19a704b422454e/lite/provider_test.go

melekes: cool. still would be nice to have a test

melekes:
https://github.com/tendermint/tendermint/pull/1043#discussion_r160029186

melekes:
```
lite/memprovider.go:92:8:warning: "heighest" is a misspelling of "heights" (misspell)
lite/performance_test.go:201:5:warning: should omit comparison to bool constant, can be simplified to !algo (S1002) (gosimple)
```

ebuchman:
Can we get a test that shows they return the same result for various queries on variously filled stores ?

odeke-em:
Roger that, am on it right now.

odeke-em:
Alright, @melekes @ebuchman, I've added the requested tests with commit https://github.com/tendermint/tendermint/pull/1043/commits/da08b7b2eca3ac81848ca2ffd4526a89b303f023, PTAL.

adrianbrink:
This failed due to a linting error.

odeke-em:
@adrianbrink what's the linting error?

zramsay:
https://circleci.com/gh/tendermint/tendermint/3003?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

`lite/performance_test.go:201:5:warning: should omit comparison to bool constant, can be simplified to !algo (S1002) (gosimple)`

odeke-em:
> lite/performance_test.go:201:5:warning: should omit comparison to bool constant, can be simplified to !algo (S1002) (gosimple)

Thanks for the notice @zramsay. Hmm, that lint warning and suggestion lose the point of readability though. I presume that we could change that to an int, but still I don't think we should be changing up readability to match a linter.

zramsay:
yeah that's why i added `nolint` instead of actually linting it

lite/memprovider.go
```diff
@@ -14,6 +14,8 @@ type memStoreProvider struct {
 	// btree would be more efficient for larger sets
```
ebuchman: comment above needs to be fixed - no longer always sorted ?

odeke-em:
> yeah that's why i added nolint instead of actually linting it
@zramsay oh I see, thank you!

> Seems fine on the surface, but needs to be properly integrated - doesn't make sense to have the two functions exposed, since we're implementing an interface here, and no user should be responsible for choosing.
Great point @ebuchman. Yes in deed, I posted up that perhaps we need an evaluation of when to switch to using the getHeightBinarySearch perhaps after hitting the point when linear search is no longer fast, perhaps derived from the benchmarks. I didn't yet know though if the PR would even be accepted, and since it has been, let me add that. Thank you for the ping.

> As is, the GetByHeightBinarySearch isn't really useful, as it's not exposed on the interface and not implemented for the other provider implementations, so I think we should address that before merging.
100% in deed, that's the last step now that it's been accepted

as I had mentioned in https://github.com/tendermint/tendermint/pull/1043#issue-285188242
> If need be we can make a hybrid algorithm that switches between the
linear and binary search depending on the number of items.
This is reminiscent of Python's TimSort algorithm.

lite/memprovider.go
```diff
@@ -52,25 +54,64 @@ func (m *memStoreProvider) StoreCommit(fc FullCommit) error {
 	defer m.mtx.Unlock()
 	m.byHash[key] = fc
 	m.byHeight = append(m.byHeight, fc)
-	sort.Sort(m.byHeight)
+	m.sorted = false
```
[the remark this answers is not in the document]
odeke-em: > Why not just sort it here?
The other performance boost comes from lazily sorting i.e when needed instead of everytime.

> In LatestCommit method we use byHeight but don't check if its sorted. Does it need to be ?
Wow, thanks for raising this. Shows that our tests ain't that comprehensive or that we don't have LatestCommit tests
```shell
$ grep -Rn 'LatestCommit(' *
client/provider.go:80:func (p *provider) LatestCommit() (fc lite.FullCommit, err error) {
client/provider.go:81:	commit, err := p.GetLatestCommit()
client/provider.go:91:func (p *provider) GetLatestCommit() (*ctypes.ResultCommit, error) {
client/provider_test.go:29:	seed, err := p.LatestCommit()
files/provider.go:105:func (p *provider) LatestCommit() (fc lite.FullCommit, err error) {
memprovider.go:131:func (m *memStoreProvider) LatestCommit() (FullCommit, error) {
provider.go:16:	LatestCommit() (FullCommit, error)
provider.go:90:func (c cacheProvider) LatestCommit() (fc FullCommit, err error) {
provider.go:93:		tfc, err = p.LatestCommit()
provider_test.go:30:func (missingProvider) LatestCommit() (lite.FullCommit, error) {
proxy/certifier.go:18:	fc, err := source.LatestCommit()
proxy/certifier.go:21:	fc, err := trust.LatestCommit()
proxy/query_test.go:64:	latest, err := source.LatestCommit()
```

ebuchman:
Thanks @odeke-em !

odeke-em:
@ebuchman I've pushed a PR updating with the requested changes, PTAL and thank you for the review!

ebuchman:
Thanks @odeke-em !

Issue #673, opened 2017-09-22
PrivValidator Structure


ebuchman:
Continue refactoring of PrivValidator started here: https://github.com/tendermint/tendermint/pull/637

TODO: write a proper ADR for this. Some things to think about:

1) move internal fields into substructs, eg. ID and LastSignedInfo
2) change interface to be Address/PubKey instead of GetAddress/GetPubKey
3) move signBytesHRS to be LastSignedInfo.SignBytesHRS
4) generalize persistence so not stuck with a file

silasdavis:
Here is my copypasta of `PrivValidatorFS` to use the verification logic in my own `PrivValidator`, in case it helps with this ticket:

```go
import (
	"bytes"
	"errors"
	"fmt"
	"sync"

	acm "github.com/hyperledger/burrow/account"
	"github.com/tendermint/go-crypto"
	"github.com/tendermint/go-wire/data"
	tm_types "github.com/tendermint/tendermint/types"
)

const (
	StepError     = -1
	StepNone      = 0 // Used to distinguish the initial state
	StepPropose   = 1
	StepPrevote   = 2
	StepPrecommit = 3
)

func VoteToStep(vote *tm_types.Vote) int8 {
	switch vote.Type {
	case tm_types.VoteTypePrevote:
		return StepPrevote
	case tm_types.VoteTypePrecommit:
		return StepPrecommit
	default:
		return StepError
	}
}

type privValidatorMemory struct {
	privateAccount acm.PrivateAccount
	lastSignedInfo *LastSignedInfo
}

var _ tm_types.PrivValidator = &privValidatorMemory{}

func NewPrivValidatorMemory(privateAccount acm.PrivateAccount) *privValidatorMemory {
	return &privValidatorMemory{
		privateAccount: privateAccount,
		lastSignedInfo: new(LastSignedInfo),
	}
}

func (pvm *privValidatorMemory) GetAddress() data.Bytes {
	return pvm.privateAccount.Address().Bytes()
}

func (pvm *privValidatorMemory) GetPubKey() crypto.PubKey {
	return pvm.privateAccount.PubKey()

}

func (pvm *privValidatorMemory) SignVote(chainID string, vote *tm_types.Vote) error {
	return pvm.lastSignedInfo.SignVote(pvm.privateAccount, chainID, vote)
}

func (pvm *privValidatorMemory) SignProposal(chainID string, proposal *tm_types.Proposal) error {
[...]

PR #1044, opened 2017-12-31
WIP: Priv Validator Refactor

ebuchman:
https://github.com/tendermint/tendermint/issues/673

Hard to see the changes because I busted up the file into its own package, but it should be much easier to read through now.

@silasdavis @wrl @ethanfrey @adrianbrink @melekes


adrianbrink:
This cleans it up a lot. :-)

What is the purpose of `PrivValidatorStore`? Also, shouldn't `CarefulSigner` be rather a struct that implements `PrivValidator`?

Couldn't we remove the Signer interface and just let the client implement `PrivValidator`. This would mean that we'd have a NanoPrivValidator struct. 

I am wondering though how it would work with moving the signing logic into a Thales HSM.

adrianbrink:
I'm wary though how people would use different signers. Ideally wanting to use a NanoLedger or Thales shouldn't require you to fork Tendermint Core. Either we have a finite list of supported signers hardcoded in Core and then try to figure out which one to use through a config option or Core expects a signer to be available behind some socket so that way in the config you can specify a socket address and Tendermint Core will just send the bytes to sign that way.

silasdavis:
> I'm wary though how people would use different signers.
@adrianbrink what are you wary about in particular? Note that we can pass in a `PrivValidator` so we wouldn't need fork Tendermint core to pass in an alternative signer. Nothing to stop you providing configurable default signers, but the flexibility to provide a signer is useful.

I'm not quite following the full idea with `CarefulSigner`, but I don't see any implementation of `GetCarefulSigner` so assume that is coming.

ebuchman:
> What is the purpose of PrivValidatorStore

A place to store the actual priv key, so eg. it can be encrypted on disk. This might go away though.

> Also, shouldn't CarefulSigner be rather a struct that implements PrivValidator
> I'm not quite following the full idea with CarefulSigner

CarefulSigner is the thing that prevents double signing. In this issue, Silas mentioned a `Verifier` interface - that's what this is. It decouples signing from the verification. For instance, you could use a ledger for just the signing, but still use our LastSignedInfo to prevent double signing. Or you could use the ledger to implement both.

> Couldn't we remove the Signer interface and just let the client implement PrivValidator

Possibly, but at least for now that would spoil CarefulSigner

> I am wondering though how it would work with moving the signing logic into a Thales HSM.

The HSM would implement Signer and CarefulSigner

> Either we have a finite list of supported signers hardcoded in Core and then try to figure out which one to use through a config option or Core expects a signer to be available behind some socket so that way in the config you can specify a socket address and Tendermint Core will just send the bytes to sign that way.

Both of these sounds pretty good. Let's start with a list of supported things (unencrypted, encrypted, ledger, thales) and then open it up to a socket.

types/priv_validator/signer.go
```diff
@@ -0,0 +1,25 @@
+package types
+
+import crypto "github.com/tendermint/go-crypto"
+
+//-----------------------------------------------------------------
+
+// DefaultSigner implements Signer.
```
melekes: how about a check: `var _ Signer = (*DefaultSigner)(nil)`

types/priv_validator/types.go
```diff
@@ -0,0 +1,91 @@
...
+
+func voteToStep(vote *types.Vote) int8 {
+	switch vote.Type {
+	case types.VoteTypePrevote:
+		return stepPrevote
+	case types.VoteTypePrecommit:
+		return stepPrecommit
+	default:
+		cmn.PanicSanity("Unknown vote type")
+		return 0
```
melekes: unreachable code

types/validator.go
```diff
@@ -108,13 +108,38 @@ func (vc validatorCodec) Compare(o1 interface{}, o2 interface{}) int {
...
 	votePower := minPower
 	if randPower {
 		votePower += int64(cmn.RandUint32())
 	}
-	val := NewValidator(privVal.GetPubKey(), votePower)
+	val := NewValidator(privVal.PubKey(), votePower)
 	return val, privVal
 }
+
+type Signer interface {
```
melekes: 2 Signer interfaces. Wont this be confusing?

ebuchman: yes this one is just for testing tho. problem of circular imports ...

ebuchman: I'll change it to TestSigner

adrianbrink:
How about calling it `Verifier` instead of `CarefulSigner`. I have a commit with these changes but you are committing faster than me :-) @ebuchman

ebuchman:
It's not just a Verifier though - at the least its a VerifyAndSigner. And it's also something that's expected to have mutable internal state if its going to keep track of things to not double sign. So it seemed CarefulSigner was a good fit for a name.

The CarefulSigner interface isn't actually being used in any useful way right now, but I think it's useful to have, especially for cases where the LastSignedInfo is on disk but the Signer is an HSM. 

The goal is for the DefaultPrivValidator to just use Signer and CarefulSigner interfaces directly, but we need to be able to unmarshall into those, so I'm waiting on changes to go-wire to make that work nicely. In the meantime, it should be a very small amount of code to add a new priv validator - the only thing is you have to handle your own marshal/unmarshal for saving/loading from disk.

ebuchman:
Ok - I went through this again and simplified. 

Please take a look at the ADR and the new simplified code. 

Dropped the CarefulSigner and just put those methods on the LastSignedInfo. 

Shouldn't be any lost functionality really.

Tendermint will jsut support PrivValidatorUnencrypted and PrivValidatorSocket. 

Folks will still be able to override in-process if they want with their own PrivValidator, and should be able to easily use the LastSignedInfo.

types/priv_validator/sign_info.go
```diff
@@ -0,0 +1,238 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	crypto "github.com/tendermint/go-crypto"
+	data "github.com/tendermint/go-wire/data"
+	"github.com/tendermint/tendermint/types"
+)
+
+// TODO: type ?
```
melekes: step is exposed in LastSignedInfo, so don't think this is a good idea because of that

types/priv_validator/sign_info.go
```diff
@@ -0,0 +1,238 @@
...
+const (
+	stepNone      int8 = 0 // Used to distinguish the initial state
```
melekes: iota

types/priv_validator/sign_info.go
```diff
@@ -0,0 +1,238 @@
...
+	info.Signature = sig
+	info.SignBytes = signBytes
+}
+
+// Reset resets all the values.
+// XXX: Unsafe.
+func (info *LastSignedInfo) Reset() {
+	info.Height = 0
+	info.Round = 0
+	info.Step = 0
```
melekes: stepNone

ebuchman:
Closing in favor of fresh new PR without the integrations https://github.com/tendermint/tendermint/pull/1203

PR #1046, opened 2018-01-01
docs: add abci spec

### LINKED ISSUE #1014: Should we include ABCI spec into readthedocs?
People sometime skip the link to https://github.com/tendermint/abci/blob/master/README.md, when it is vital info in order to write an app!

zramsay:
closes #1014 
goes with https://github.com/tendermint/abci/pull/172


ebuchman:
Circle failure should be fixed by https://github.com/tendermint/tendermint/pull/1059

Issue #982, opened 2017-12-16
p2p: peer.Key() uses pubkey but pex_reactor does lookups via address


ebuchman:
Key() is pubkey: https://github.com/tendermint/tendermint/blob/master/p2p/peer.go#L212

but pex reactor uses IP: https://github.com/tendermint/tendermint/blob/master/p2p/pex_reactor.go#L267

This causes the pex to try dialing all the peers we're already connected to.

Either we change the Key() to be the address, or we store the PubKey in the address book too

odeke-em:
@ebuchman how did you find this? Auditing with eyeballing?

ebuchman:
I noticed that peers were dialing eachother in the pex reactor test when they were already connected, and then read the code and found it!

odeke-em:
Oh crap, is there something that we can do to perhaps detect such or perhaps make an integration test later on?

paralin:
Is there a fix?

melekes:
Just looked up how libp2p does this. It seems like they take hash from pubkey for peer identity (there are other methods): https://github.com/libp2p/go-libp2p-peer/blob/master/peer.go#L194

```

In P2P networks which distribute resources of dubious legality, the issue of lack of anonymity becomes apparent. For example, the BitTorrent file sharing system directly exposes the IP address of peers to each other in a swarm. This would allow peers in the swarm to know the identity of other peers who are downloading certain resources, for example. Once the peer's identity is compromised, further attacks, whether physical or legal, can continue to be directed at that specific target.
```
http://www.cse.wustl.edu/~jain/cse571-07/ftp/p2p/#privacy

They use https://github.com/multiformats/go-multiaddr in their addrbook.

melekes:
Also good reading: A Sybilproof Distributed Identity Management for P2P Networks http://ieeexplore.ieee.org/document/4625694/

ebuchman:
Fixed in https://github.com/tendermint/tendermint/pull/1048

Issue #806, opened 2017-11-05
p2p: Persistent node key


ebuchman:
Need a persistent node identity and reasonable certificate format - x509?

Node identity should have some proof of work in it (ie. a pubkey that starts with some number of 0s)

We can also put some proof of work in the handshake.

ebuchman:
Solved by https://github.com/tendermint/tendermint/pull/1048

We decided no PoW on IDs and just use an unencrypted privkey on disk for now

PR #1048, opened 2018-01-02
P2P ID

ebuchman:
Add a persistent peer ID. Addresses:
- https://github.com/tendermint/tendermint/issues/982
- https://github.com/tendermint/tendermint/issues/806


p2p/key.go
```diff
@@ -0,0 +1,122 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"math/big"
+
+	crypto "github.com/tendermint/go-crypto"
+	cmn "github.com/tendermint/tmlibs/common"
+)
+
+type ID string
```
odeke-em: @ebuchman what's the purpose of us defining `type ID string`?
If perhaps it's to add later utility methods like:
* `Validate() error`
* `Equals() bool`
then perhaps that's necessary.

p2p/key.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (nodeKey *NodeKey) SatisfiesTarget(target []byte) bool {
+	return bytes.Compare(nodeKey.id(), target) < 0
+}
+
+// LoadOrGenNodeKey attempts to load the NodeKey from the given filePath,
+// and checks that the corresponding ID is less than the target.
+// If the file does not exist, it generates and saves a new NodeKey
+// with ID less than target.
```
odeke-em: Could you please clarify what `with ID less than target` means?

p2p/key.go
```diff
@@ -0,0 +1,122 @@
...
+	return nodeKey, nil
+}
+
+// generate key with address satisfying the difficult target
+func genPrivKeyEd25519PoW(target []byte) crypto.PrivKeyEd25519 {
+	secret := crypto.CRandBytes(32)
+	var privKey crypto.PrivKeyEd25519
+	for i := 0; ; i++ {
+		privKey = crypto.GenPrivKeyEd25519FromSecret(secret)
+		if bytes.Compare(privKey.PubKey().Address(), target) < 0 {
```
odeke-em: Could you please add a comment specifying what this does?

This looks like something that few people will understand and might be easy to fail, because
even the zero value would pass here ` < 0` because
```go
bytes.Compare([]byte{0x00}, anyValue)
``` 
always returns -1.

p2p/key.go
```diff
@@ -0,0 +1,122 @@
...
+// MakePoWTarget returns a 20 byte target byte array.
+func MakePoWTarget(difficulty uint8) []byte {
+	zeroPrefixLen := (int(difficulty) / 8)
+	prefix := bytes.Repeat([]byte{0}, zeroPrefixLen)
+	mod := (difficulty % 8)
+	if mod > 0 {
+		nonZeroPrefix := byte(1 << (8 - mod))
+		prefix = append(prefix, nonZeroPrefix)
+	}
+	return append(prefix, bytes.Repeat([]byte{255}, 20-len(prefix))...)
```
odeke-em: Could we use `0xFF` instead?
```go
bytes.Repeat([]byte{0xFF}, 20-len(prefix
```

Also this code is going to crash starting at a difficulty of 161 until the end because:
161/8 = 20
161%8 = 1
so len(prefix) = 21
20 - 21 = -1 which finna crash bytes.Repeat

p2p/netaddress_test.go
```diff
@@ -38,6 +38,16 @@ func TestNewNetAddressString(t *testing.T) {
...
+
+		{"deadbeef@127.0.0.1:8080", "", false},
+		{"this-isnot-hex@127.0.0.1:8080", "", false},
+		{"xxxxbeefdeadbeefdeadbeefdeadbeefdeadbeef@127.0.0.1:8080", "", false},
+		{"deadbeefdeadbeefdeadbeefdeadbeefdeadbeef@127.0.0.1:8080", "deadbeefdeadbeefdeadbeefdeadbeefdeadbeef@127.0.0.1:8080", true},
+
+		{"tcp://deadbeef@127.0.0.1:8080", "", false},
+		{"tcp://this-isnot-hex@127.0.0.1:8080", "", false},
+		{"tcp://xxxxbeefdeadbeefdeadbeefdeadbeefdeadbeef@127.0.0.1:8080", "", false},
+		{"tcp://deadbeefdeadbeefdeadbeefdeadbeefdeadbeef@127.0.0.1:8080", "deadbeefdeadbeefdeadbeefdeadbeefdeadbeef@127.0.0.1:8080", true},
```
odeke-em: Could we also try with:
* "tcp://@127.0.0.1:8080"
* "tcp://@"
* ""
* "@"
* "     @"
* "  @ "

p2p/switch.go
```diff
@@ -231,11 +239,15 @@ func (sw *Switch) OnStop() {
 // NOTE: This performs a blocking handshake before the peer is added.
 // NOTE: If error is returned, caller is responsible for calling peer.CloseConn()
 func (sw *Switch) addPeer(peer *peer) error {
+	// Avoid self
+	if sw.nodeKey.ID() == peer.ID() {
```
odeke-em: Me likey!
Although perhaps let's make this a sentinel error that can be checked later on in code e.g
```go
var ErrSelfConnection = errors.New("Ignoring connection from self"

...

switch sw.AddPeer(p) {
case nil:
  // Cool
case ErrSelfConnection:
  // Not severe, so we can ignore
default:
  // Handle this error
}
```

p2p/switch.go
```diff
@@ -244,22 +256,17 @@ func (sw *Switch) addPeer(peer *peer) error {
 		return err
 	}
 
-	// Avoid self
-	if sw.nodeInfo.PubKey.Equals(peer.PubKey().Wrap()) {
-		return errors.New("Ignoring connection from self")
+	// Avoid duplicate
+	if sw.peers.Has(peer.ID()) {
```
odeke-em: Nice!

p2p/key.go
milosevic: From docs (https://github.com/tendermint/tendermint/blob/develop/p2p/docs/peer.md): Peer ID's must come with some Proof-of-Work; that is, they must satisfy peer.PrivKey.Address() < target for some difficulty target. This ensures they are not too easy to generate. To begin, let target == 2^240.

p2p/key.go
ebuchman: it's more so to clarify the allowed type when its being used eg. as a map index. We want to make sure we're not accidentally using the wrong string. If we use a type, the compiler enforces it :) ... `map[ID]whatever`

p2p/key.go
odeke-em: Thanks @milosevic. However, that's still not disambiguated in prose. How about:
```go
// it generates and saves a new NodeKey whose ID's proof work is less than that of target
// thus ensuring that the ID is not too easy to generate.
``` 

What do you think of the above?

p2p/key.go
odeke-em: Aye aye, gotcha. Thanks @ebuchman.

p2p/key.go
milosevic: @odeke-em It's more clear as you suggested; although we are dropping PoW for ID generation if I am not wrong.

p2p/addrbook_test.go
```diff
@@ -188,6 +190,7 @@ func randIPv4Address(t *testing.T) *NetAddress {
 		)
 		port := rand.Intn(65535-1) + 1
 		addr, err := NewNetAddressString(fmt.Sprintf("%v:%v", ip, port))
+		addr.ID = ID(hex.EncodeToString(cmn.RandBytes(20))) // TODO
```
melekes: can be removed

p2p/netaddress.go
```diff
@@ -28,25 +31,45 @@ type NetAddress struct {
...
+
+	var id ID
+	spl := strings.Split(addr, "@")
+	if len(spl) == 2 {
+		idStr := spl[0]
+		idBytes, err := hex.DecodeString(idStr)
+		if err != nil {
+			return nil, errors.Wrap(err, fmt.Sprintf("Address (%s) contains invalid ID", addr))
+		}
+		if len(idBytes) != 20 {
```
melekes: maybe create a `const peerIDSize = 20`?

p2p/pex_reactor.go
```diff
@@ -107,6 +107,7 @@ func (r *PEXReactor) AddPeer(p Peer) {
 		}
 	} else { // For inbound connections, the peer is its own source
 		addr, err := NewNetAddressString(p.NodeInfo().ListenAddr)
+		addr.ID = p.ID() // TODO: handle in NewNetAddress func
```
melekes: can be removed

melekes:
This PR is massive!

p2p/addrbook_test.go
ebuchman: what, the TODO?

p2p/netaddress.go
ebuchman: good catch. So far the addresses we use in go-crypto are 20 bytes but nothing really guarantees that for future addresses. Maybe we will make that convention but for now, I'll simply remove this check

Issue #957, opened 2017-12-11
p2p: Tendermint disconnects peers running unknown reactors/channels


ebuchman:
Ie. if one peer runs the PEX and another doesnt, the latter will disconnect the former when it receives a PEX message: https://github.com/tendermint/tendermint/blob/master/p2p/connection.go#L468

Not sure this is something we want. What accommodations should we make for unknown reactors/channels?

melekes:
> What accommodations should we make for unknown reactors/channels?

drop these messages?

ebuchman:
Yeh, we just need to be careful about spam/DoS

melekes:
Maybe we need to check if peer has pex enabled before sending plus punish peer that is trying to send us pex msg given he knows that pex disabled.

ebuchman:
Lets add `[]Channels` to the NodeInfo struct.

Then we can make peer.Send/TrySend drop msgs for channels the peer the doesnt know about.

Then if a peer sends us a msg for a channel we don't know about, we can mark them as bad and disconnect :)

ebuchman:
closing for PR: https://github.com/tendermint/tendermint/pull/1049

PR #1049, opened 2018-01-02
p2p: add Channels to NodeInfo and don't send for unknown channels

ebuchman:
Solves https://github.com/tendermint/tendermint/issues/957

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -18,7 +18,7 @@ import (
 
 	bc "github.com/tendermint/tendermint/blockchain"
 	cfg "github.com/tendermint/tendermint/config"
-	"github.com/tendermint/tendermint/consensus"
+	cs "github.com/tendermint/tendermint/consensus"
 	"github.com/tendermint/tendermint/evidence"
 	mempl "github.com/tendermint/tendermint/mempool"
 	"github.com/tendermint/tendermint/p2p"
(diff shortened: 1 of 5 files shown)


p2p/types.go
```diff
@@ -51,6 +53,24 @@ func (info *NodeInfo) CompatibleWith(other *NodeInfo) error {
...
+	if len(info.Channels) == 0 {
+		return nil
+	}
+
+	// for each of our channels, check if they have it
+	found := false
+	for _, ch1 := range info.Channels {
+		for _, ch2 := range other.Channels {
+			if ch1 == ch2 {
+				found = true
```
melekes: should we `break` after?

node/node.go
```diff
@@ -551,10 +551,17 @@ func (n *Node) makeNodeInfo(pubKey crypto.PubKey) *p2p.NodeInfo {
 		Moniker: n.config.Moniker,
 		Network: n.genesisDoc.ChainID,
 		Version: version.Version,
+		Channels: []byte{
+			bc.BlockchainChannel,
+			cs.StateChannel, cs.DataChannel, cs.VoteChannel, cs.VoteSetBitsChannel,
+			mempl.MempoolChannel,
+			evidence.EvidenceChannel,
+			p2p.PexChannel,
```
melekes: what if `pex` is disabled?

melekes:
> Then if a peer sends us a msg for a channel we don't know about, we can mark them as bad and disconnect :)

what about this part?

ebuchman:
The old code wasn't actually doing what we wanted. Force pushed  the fix. PTAL

ebuchman:
> what about this part?

This is implemented already since if we receive a msg for an unknown channel we stopPeerForError.

ebuchman:
Saving that for https://github.com/tendermint/tendermint/issues/1148

p2p/node_info.go
```diff
@@ -71,6 +90,24 @@ func (info NodeInfo) CompatibleWith(other NodeInfo) error {
...
+		return nil
+	}
+
+	// for each of our channels, check if they have it
+	found := false
+	for _, ch1 := range info.Channels {
+		for _, ch2 := range other.Channels {
+			if ch1 == ch2 {
+				found = true
+				break // only need one
```
melekes: I think the intention here was to break from outer loop, no?

PR #1051, opened 2018-01-03
docs: Blockchain Reactor Documentation

adrianbrink:


docs/specification/new-spec/blockchain_reactor.md
```diff
@@ -0,0 +1,91 @@
...
+    Height int64
+
+type bcStatusResponseMessage struct {
+    Height int64
+}
+```
+
+## Blockchain Reactor
+
+* coordinates the pool for syncing
```
ebuchman: please try to describe the protocol more clearly. we shouldn't be bothering here with implementation details - no references to code, only to other things well defined in the spec. so things like the block store and the p2p.BaseReactor and "call .Start()" are all irrelevent here. We just want to describe the core behaviour.

PR #792, opened 2017-10-26
Config Improvements

zramsay:
most of #628 
refs: #709, #733

PR #1056, opened 2018-01-04
WIP: Mempool specification

ethanfrey:
Writing a public specification of the mempool module.

* Include overview and configuration options.

### DIFF
--- a/mempool/docs/config.md
+++ b/mempool/docs/config.md
@@ -1,0 +1,59 @@
+# Mempool Configuration
+
+Here we describe configuration options around mempool.
+For the purposes of this document, they are described
+as command-line flags, but they can also be passed in as
+environmental variables or in the config.toml file. The
+following are all equivalent:
+
+Flag: `--mempool.recheck_empty=false`
+
+Environment: `TM_MEMPOOL_RECHECK_EMPTY=false`
+
+Config:
+```
+[mempool]
+recheck_empty = false
+```
+
+
+## Recheck
+
+`--mempool.recheck=false` (default: true)
+
+`--mempool.recheck_empty=false` (default: true)
+
+Recheck determines if the mempool rechecks all pending
+transactions after a block was committed. Once a block
+is committed, the mempool removes all valid transactions
+that were successfully included in the block.
+
+If `recheck` is true, then it will rerun CheckTx on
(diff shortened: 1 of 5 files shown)


mempool/docs/config.md
```diff
@@ -0,0 +1,59 @@
...
+[mempool]
+recheck_empty = false
+```
+
+
+## Recheck
+
+`--mempool.recheck=false` (default: true)
+
+`--mempool.recheck_empty=false` (default: true)
```
ethanfrey: @ebuchman Do we want to make this false by default?

mempool.go:368 suggests it is used for debugging

ebuchman: Maybe. But currently the mempool is supposed to be configured so a correct proposer will only propose valid txs. Apps that use EndBlock could invalidate transactions in the mempool without the mempool knowing, and so a correct proposer could return invalid txs.

We just need to be clear either way. There's def a performance benefit to setting it false (ie. we won't replay all our mempool txs when an empty block is committed ...)

mempool/docs/config.md
```diff
@@ -0,0 +1,59 @@
+# Mempool Configuration
```
melekes: there is https://github.com/tendermint/tendermint/pull/792. let's try to avoid duplication. thanks

melekes: @zramsay maybe you can incorporate some things below into your PR

mempool/docs/messages.md
```diff
@@ -0,0 +1,60 @@
...
+the actual 40-byte tx. Prefix of 350=0x015e byte tx is:
+`0x0102015e...` followed by the actual 350 byte tx.
+
+(Please see the [go-wire repo](https://github.com/tendermint/go-wire#an-interface-example) for more information)
+
+## RPC Messages
+
+Mempool exposes `CheckTx([]byte)` over the RPC interface.
+
+It can be posted via `broadcast_commit`, `broadcast_sync` or
```
melekes: these are not the same things... broadcast_commit calls both CheckTx and waits for DeliverTx result. not sure this should be here. i think this belongs to RPC API!

ebuchman:
Merging, will clean up later

zramsay:
do we want any of these rendered in RTD?

PR #1057, opened 2018-01-04
For Review: P2P docs

ebuchman:
These docs are already on develop but should have gone through PR for review tools


p2p/docs/config.md
```diff
@@ -0,0 +1,39 @@
+# P2P Config
+
+Here we describe configuration options around the Peer Exchange.
+
+## Seed Mode
+
+`--p2p.seed_mode`
+
+The node operates in seed mode. It will kick incoming peers after sharing some peers.
```
milosevic: What about: The node operates in seed mode. In seed mode, a node continuously crawls the network for peers, and upon incoming connection shares some peers and closes the connection.

p2p/docs/config.md
```diff
@@ -0,0 +1,39 @@
...
+`--p2p.persistent_peers “1.2.3.4:46656,2.3.4.5:466656”`
+
+Dial these peers and auto-redial them if the connection fails.
+These are intended to be trusted persistent peers that can help
+anchor us in the p2p network.
+
+Note that the auto-redial uses exponential backoff and will give up
+after a day of trying to connect.
+
+NOTE: If `dial_seeds` and `persistent_peers` intersect,
```
milosevic: Is it `seeds` instead of `dial_seeds`?

p2p/docs/connection.md
```diff
@@ -0,0 +1,116 @@
+## MConnection
+
+`MConnection` is a multiplex connection:
+
+__multiplex__ *noun* a system or signal involving simultaneous transmission of
```
milosevic: I don't understand this definition. If I am not wrong any TCP connection satisfies this definition as you can send several messages over it.

p2p/docs/connection.md
```diff
@@ -0,0 +1,116 @@
...
+`Channel`s.  Each channel has a globally unique byte id.
+The byte id and the relative priorities of each `Channel` are configured upon
+initialization of the connection.
+
+The `MConnection` supports three packet types: Ping, Pong, and Msg.
+
+### Ping and Pong
+
+The ping and pong messages consist of writing a single byte to the connection; 0x1 and 0x2, respectively
+
```
milosevic: . missing at the end of previous sentence.

p2p/docs/connection.md
```diff
@@ -0,0 +1,116 @@
...
+When we haven't received any messages on an `MConnection` in a time `pingTimeout`, we send a ping message.
+When a ping is received on the `MConnection`, a pong is sent in response.
```
milosevic: Do we always send pong or only if there is no msg to be sent? Also we probably want to have some timeout between two consecutive pongs, i.e., we don't want to react upon every ping as this might be DDoS attack.

p2p/docs/connection.md
```diff
@@ -0,0 +1,116 @@
...
+
+If a pong is not received in sufficient time, the peer's score should be decremented (TODO).
+
+### Msg
+
+Messages in channels are chopped into smaller msgPackets for multiplexing.
```
milosevic: Not clear why this is done? TCP already provides some support cutting big messages in smaller chunks. I guess it's part of MConnection way of providing different QoS for different channels?

p2p/docs/connection.md
```diff
@@ -0,0 +1,116 @@
...
+The msgPacket is serialized using go-wire, and prefixed with a 0x3.
+The received `Bytes` of a sequential set of packets are appended together
+until a packet with `EOF=1` is received, at which point the complete serialized message
+is returned for processing by the corresponding channels `onReceive` function.
+
+### Multiplexing
+
+Messages are sent from a single `sendRoutine`, which loops over a select statement that results in the sending
+of a ping, a pong, or a batch of data messages. The batch of data messages may include messages from multiple channels.
+Message bytes are queued for sending in their respective channel, with each channel holding one unsent message at a time.
```
milosevic: Why only one unsent message at a time? We might be missing some high level design goal regarding multiplexing.

p2p/docs/connection.md
```diff
@@ -0,0 +1,116 @@
...
+```go
+func (m MConnection) Send(chID byte, msg interface{}) bool {}
+func (m MConnection) TrySend(chID byte, msg interface{}) bool {}
+```
+
+`Send(chID, msg)` is a blocking call that waits until `msg` is successfully queued
+for the channel with the given id byte `chID`.  The message `msg` is serialized
+using the `tendermint/wire` submodule's `WriteBinary()` reflection routine.
+
+`TrySend(chID, msg)` is a nonblocking call that returns false if the channel's
```
milosevic: What about: is a nonblocking call that queues the message `msg` in the channel with the given id byte `chID` if the queue is not full; otherwise it returns false immediately.

p2p/docs/connection.md
```diff
@@ -0,0 +1,116 @@
...
+for _, peer := range switch.Peers().List() {
+    if peer.IsOutbound() {
+        peer.Send(MyChannelID, "Here's a random message")
+    }
+}
+```
+
+### PexReactor/AddrBook
+
+A `PEXReactor` reactor implementation is provided to automate peer discovery.
```
milosevic: What PEX stands for?

p2p/docs/connection.md
```diff
@@ -0,0 +1,116 @@
...
+
+// Other Reactor methods omitted for brevity
+...
+
+switch := NewSwitch([]Reactor{MyReactor{}})
+
+...
+
+// Send a random message to all outbound connections
+for _, peer := range switch.Peers().List() {
```
milosevic: What is illustrated with the following code snippet?

p2p/docs/node.md
```diff
@@ -0,0 +1,62 @@
...
+to bad peers or their social consensus was invalidated.
+
+## Restarted Full Node
+
+A node checks its address book on startup and attempts to connect to peers from there.
+If it can't connect to any peers after some time, it falls back to the seeds to find more.
+
+Restarted full nodes can run the `blockchain` or `consensus` reactor protocols to sync up
+to the latest state of the blockchain, assuming they aren't too far behind.
+If they are too far behind, they may need to validate a recent `H` and `HASH` out-of-band again.
```
milosevic: Don't understand this. In order to validate recent hash it needs to start from the genesis, right? Isn't it always slower then catching up?

p2p/docs/node.md
```diff
@@ -0,0 +1,62 @@
...
+These nodes require the highest security, and should not accept incoming connections.
+They should maintain outgoing connections to a controlled set of "Sentry Nodes" that serve
+as their proxy shield to the rest of the network.
+
+Validators that know and trust each other can accept incoming connections from one another and maintain direct private connectivity via VPN.
+
+## Sentry Node
+
+Sentry nodes are guardians of a validator node and provide it access to the rest of the network.
+Sentry nodes may be dynamic, but should maintain persistent connections to some evolving random subset of each other.
```
milosevic: Sentry node can also probably connect to Full Nodes.

p2p/docs/peer.md
```diff
@@ -0,0 +1,118 @@
+# Tendermint Peers
+
+This document explains how Tendermint Peers are identified, how they connect to one another,
+and how other peers are found.
+
+## Peer Identity
+
+Tendermint peers are expected to maintain long-term persistent identities in the form of a private key.
```
milosevic: Should it be public key?

p2p/docs/peer.md
```diff
@@ -0,0 +1,118 @@
...
+Each peer has an ID defined as `peer.ID == peer.PrivKey.Address()`, where `Address` uses the scheme defined in go-crypto.
```
milosevic: Should it be `peer.ID == peer.PubKey.Address()`?

p2p/docs/peer.md
```diff
@@ -0,0 +1,118 @@
...
+
+Peer ID's must come with some Proof-of-Work; that is,
```
milosevic: Can be removed as PoW for peer id is dropped.

p2p/docs/peer.md
```diff
@@ -0,0 +1,118 @@
...
+they must satisfy `peer.PrivKey.Address() < target` for some difficulty target.
+This ensures they are not too easy to generate. To begin, let `target == 2^240`.
+
+A single peer ID can have multiple IP addresses associated with it.
+For simplicity, we only keep track of the latest one.
```
milosevic: If I am not wrong, there was a case where we benefit from keeping multiple keys.

p2p/docs/peer.md
```diff
@@ -0,0 +1,118 @@
...
+
+When attempting to connect to a peer, we use the PeerURL: `<ID>@<IP>:<PORT>`.
+We will attempt to connect to the peer at IP:PORT, and verify,
+via authenticated encryption, that it is in possession of the private key
+corresponding to `<ID>`. This prevents man-in-the-middle attacks on the peer layer.
+
+Peers can also be connected to without specifying an ID, ie. just `<IP>:<PORT>`.
+In this case, the peer must be authenticated out-of-band of Tendermint,
+for instance via VPN
```
milosevic: . is missing at the end of the sentence.

p2p/docs/peer.md
```diff
@@ -0,0 +1,118 @@
...
+- compute the Diffie-Hellman shared secret using the peers ephemeral public key and our ephemeral private key
+- generate two nonces to use for encryption (sending and receiving) as follows:
+    - sort the ephemeral public keys in ascending order and concatenate them
+    - RIPEMD160 the result
+    - append 4 empty bytes (extending the hash to 24-bytes)
+    - the result is nonce1
+    - flip the last bit of nonce1 to get nonce2
+    - if we had the smaller ephemeral pubkey, use nonce1 for receiving, nonce2 for sending;
+        else the opposite
+- all communications from now on are encrypted using the shared secret and the nonces, where each nonce
```
milosevic: It looks like sentence is not complete.

p2p/docs/peer.md
```diff
@@ -0,0 +1,118 @@
...
+
+The connection has now been authenticated. All traffic is encrypted.
+
+Note that only the dialer can authenticate the identity of the peer,
+but this is what we care about since when we join the network we wish to
+ensure we have reached the intended peer (and are not being MITMd).
+
+### Peer Filter
+
+Before continuing, we check if the new peer has the same ID as ourselves or
```
milosevic: As we don't assume some kind of central CA to exist for id/key management, it can happen (although with very small probability I guess) that two nodes exist in the network with the same id (public key). Is this a problem?

p2p/docs/peer.md
```diff
@@ -0,0 +1,118 @@
...
+an existing peer. If so, we disconnect.
+
+We also check the peer's address and public key against
+an optional whitelist which can be managed through the ABCI app -
+if the whitelist is enabled and the peer does not qualigy, the connection is
```
milosevic: Typo: qualify instead of qualigy.

p2p/docs/peer.md
```diff
@@ -0,0 +1,118 @@
...
+terminated.
+
+
+### Tendermint Version Handshake
+
+The Tendermint Version Handshake allows the peers to exchange their NodeInfo:
+
+```
+type NodeInfo struct {
+	PubKey     crypto.PubKey `json:"pub_key"`
```
milosevic: Maybe remove json stuff.

p2p/docs/config.md
```diff
@@ -0,0 +1,39 @@
...
+`--p2p.seed_mode`
+
+The node operates in seed mode. It will kick incoming peers after sharing some peers.
+It will continually crawl the network for peers.
+
+## Seeds
+
+`--p2p.seeds “1.2.3.4:466656,2.3.4.5:4444”`
+
+Dials these seeds when we need more peers. They will return a list of peers and then disconnect.
```
melekes: `will -> should`?? i.e. what if one node from this list is not in "seed mode"?

p2p/docs/connection.md
melekes: this is a very good question

p2p/docs/connection.md
melekes: Peer Exchange

p2p/docs/peer.md
```diff
@@ -0,0 +1,118 @@
...
+    - RIPEMD160 the result
+    - append 4 empty bytes (extending the hash to 24-bytes)
+    - the result is nonce1
+    - flip the last bit of nonce1 to get nonce2
+    - if we had the smaller ephemeral pubkey, use nonce1 for receiving, nonce2 for sending;
+        else the opposite
+- all communications from now on are encrypted using the shared secret and the nonces, where each nonce
+- we now have an encrypted channel, but still need to authenticate
+increments by 2 every time it is used
+- generate a common challenge to sign:
```
melekes: why this is called a challenge? i.e. I do not see any challenge here? what's challenging about signing a msg?

p2p/docs/connection.md
ebuchman: This is the stock google definition. 

Afaik, TCP is typically considered a single stream - you read and write one stream of messages. This is an abstraction to support reading/writing on multiple streams

p2p/docs/connection.md
ebuchman: Yes I believe that's right

p2p/docs/connection.md
ebuchman: just an example of using the library. in this case to send a random message to all outbound connections (ie. peers that we dialed)

p2p/docs/node.md
ebuchman: "validate recent hash" is a social excercise. the idea here is if you're far enough behind, you are vulnerable to nothing-at-stake attacks, so when you sync up to some recent block, you need to validate its hash socially to know you're on the right chain

p2p/docs/peer.md
ebuchman: No - that would require a private key collision and we consider those sufficiently improbable. If we had to worry about it here, we'd have to worry about it in the app layer too, and cryptocurrencies wouldn't work

p2p/docs/peer.md
ebuchman: The challenge is: "prove you have the privkey for this pubkey by signing this msg"

ebuchman:
thanks for the feedback folks. see updates in https://github.com/tendermint/tendermint/pull/1076/files

p2p/docs/connection.md
melekes: there is a TODO in code

milosevic:
Looks great!

p2p/docs/pex.md
```diff
@@ -0,0 +1,94 @@
...
+- select a random, unbiased `getSelectionPercent` of the peers
+
+Send the selected peers. Note we select peers for sending without bias for vetted/unvetted.
+
+## Preventing Spam
+
+There are various cases where we decide a peer has misbehaved and we disconnect from them.
+When this happens, the peer is removed from the address book and black listed for
+some amount of time. We call this "Disconnect and Mark".
+Note that the bad behaviour may be detected outside the PEX reactor itseld
```
caffix: Typo: "itseld" was probably meant to be "itself"

p2p/docs/pex.md
```diff
@@ -0,0 +1,94 @@
...
+so it can remove and mark the peer.
+
+In the PEX, if a peer sends us unsolicited lists of peers,
+or if the peer sends too many requests for more peers in a given amount of time,
+we Disconnect and Mark.
+
+## Trust Metric
+
+The quality of peers can be tracked in more fine-grained detail using a
+Proportional-Integral-Derrivative (PID) controller that incorporates
```
caffix: Typo: "Derrivative" should be "Derivative"

p2p/docs/trustmetric.md
```diff
@@ -0,0 +1,16 @@
...
+If a vetted peer's quality degrades sufficiently, it is booted, and must prove itself from scratch.
+If we need to make room for a new vetted peer, we move the lowest scoring vetted peer back to unvetted.
+If we need to make room for a new unvetted peer, we remove the lowest scoring unvetted peer -
+possibly only if its below some absolute minimum ?
+
+Peer quality is tracked in the connection and across the reactors.
+Behaviours are defined as one of:
+    - fatal - something outright malicious. we should disconnect and remember them.
+    - bad - any kind of timeout, msgs that dont unmarshal, or fail other validity checks, or msgs we didn't ask for or arent expecting
+    - neutral - normal correct behaviour. unknown channels/msg types (version upgrades).
```
caffix: We discussed neutral behavior being used for unknown channels, etc, and a separate type of behavior named "Correct" being used as the opposite of Bad. This way, good is reserved for behavior that is exceptionally positive.

p2p/docs/trustmetric.md
```diff
@@ -0,0 +1,16 @@
+
+The trust metric tracks the quality of the peers.
```
caffix: Additional detail could be provided in this file using information from the architecture documents if you wish.

ebuchman:
Thanks @caffix . Fixes in https://github.com/tendermint/tendermint/pull/1123

PR #1059, opened 2018-01-04
priv-val: fix timestamp for signing things that only differ by timestamp

ebuchman:
If we crash after signing but before we hit the WAL, we might later retry signing the same object but with a different timestamp.

This ensures we replace the timestamp with the old one and return the old signature. 

Though I'm realizing now we might be able to make it ok to double sign where the only difference is the timestamp. Just need to make evidence logic aware that this is ok.  That would remove the need for this extra logic in the priv val. But it could make the BFT time piece more difficult


ebuchman:
added to https://github.com/tendermint/tendermint/pull/1059

Issue #819, opened 2017-11-07
Detached Signing Validator Design


MeherRoy:
I'd like to propose an idea to make the life of validators easier.

**Business Problem**

A good validator possesses a large atom stake, is able to setup security infrastructure and validate with low network latency. Individually large atom holders may not be skilled at deploying secure key signing infrastructure. They will need to recruit a Sys Admin to maintain the validating nodes, connect to the network etc. 

For instance, consider a hedge fund manager - financially savvy enough to buy atoms but not skilled at secure key signing infrastructure. A hedge fund manager cannot realistically delegate to an unrelated validator - the operational risk of having their LP funds slashed by an unrelated party is just too big. Hedge fund managers will need to be their own validators.

The interaction between SysAdmins and large atom holders creates a trust issue. Handing over control over private keys to SysAdmin makes the atom holder susceptible to be blackmailed by the SysAdmin - they could threaten to double sign unless a large side-payment is made.

Securing non-computer savvy large atom validators against threat of blackmail is an important problem that needs a satisfactory solution

**A potential solution?**

A detached signing validator design that separates the following two concerns across two different servers might work:

A. Validating server: Connects to the network, maintains the tx mempool, propose blocks and communicates with other nodes on the network.

B. Signing server: Only responsible for signing pre-commit and commit header. Receives "request to sign headers" over RPC, checks that it is not double signing, and then sends signed header over RPC. Does not validate transactions, create blocks or participate in the network.

If the combination of a validating and signing server can be made, our prototypical hedge fund manger could run the simple signing server themselves, and have the SysAdmin run the validating server. The attack surface of the signing server can be made very small, and the machine can also be physically secure against theft.

Is this a feasible design? What are the latency consequences of detached signing?

zmanian:
ToDo: Define an RPC signing protocol that calls out to url to request a signature.

Recipient/hardware wallet code can enforce the invariant that never sign a message type twice at the same height.

This can minimize the surface area by which an attacker can use to black mail an operator.

ebuchman:
This is being worked on in https://github.com/tendermint/tendermint/pull/1044,  https://github.com/tendermint/tendermint/pull/1060, and https://github.com/tendermint/signatory.

Will close this issue in favour of those PRs

PR #1060, opened 2018-01-05
Priv val via sockets

ebuchman:
Support a socket client and server for the priv validator.

Dead simple synchronous server.

@xla @ethanfrey want to work on this ? 

see https://github.com/tendermint/tendermint/pull/1044 for more context

### DIFF
--- a/types/priv_validator/socket.go
+++ b/types/priv_validator/socket.go
@@ -12,8 +12,8 @@ import (
 	"github.com/tendermint/go-wire/data"
 	cmn "github.com/tendermint/tmlibs/common"
 	"github.com/tendermint/tmlibs/log"
-	"golang.org/x/net/netutil"
 
+	"github.com/tendermint/tendermint/p2p"
 	"github.com/tendermint/tendermint/types"
 )
 
@@ -26,7 +26,8 @@ var _ types.PrivValidator = (*PrivValidatorSocketClient)(nil)
 type PrivValidatorSocketClient struct {
 	cmn.BaseService
 
-	conn net.Conn
+	conn    net.Conn
+	privKey *crypto.PrivKeyEd25519
 
 	ID            types.ValidatorID
 	SocketAddress string
@@ -38,11 +39,18 @@ const (
 
 // NewPrivValidatorSocketClient returns an instance of
 // PrivValidatorSocketClient.
-func NewPrivValidatorSocketClient(logger log.Logger, socketAddr string) *PrivValidatorSocketClient {
+func NewPrivValidatorSocketClient(
+	logger log.Logger,
+	socketAddr string,
+	privKey *crypto.PrivKeyEd25519,
+) *PrivValidatorSocketClient {
 	pvsc := &PrivValidatorSocketClient{
 		SocketAddress: socketAddr,
+		privKey:       privKey,
 	}
+
 	pvsc.BaseService = *cmn.NewBaseService(logger, "privValidatorSocketClient", pvsc)
+
 	return pvsc
 }
 
@@ -54,6 +62,7 @@ func (pvsc *PrivValidatorSocketClient) OnStart() error {
 
 	var err error
 	var conn net.Conn
+
 RETRY_LOOP:
 	for {
 		conn, err = cmn.Connect(pvsc.SocketAddress)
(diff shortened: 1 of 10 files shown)


xla:
@ebuchman Can take this on after wrapping up #1056

adrianbrink:
implements #819

types/priv_validator/socket.go
```diff
@@ -0,0 +1,273 @@
...
+
+func readWrite(conn net.Conn, req PrivValidatorSocketMsg) (res PrivValidatorSocketMsg, err error) {
+	b := wire.BinaryBytes(req)
+	_, err = conn.Write(b)
+	if err != nil {
+		return nil, err
+	}
+
+	var n int
+	b = wire.ReadByteSlice(conn, 0, &n, &err) //XXX: no max
```
melekes: check for err after before calling `decodeMsg`

types/priv_validator/socket.go
```diff
@@ -0,0 +1,273 @@
...
+	return decodeMsg(b)
+}
+
+func decodeMsg(bz []byte) (msg PrivValidatorSocketMsg, err error) {
+	n := new(int)
+	r := bytes.NewReader(bz)
+	msgI := wire.ReadBinary(struct{ PrivValidatorSocketMsg }{}, r, 0, n, &err)
+	msg = msgI.(struct{ PrivValidatorSocketMsg }).PrivValidatorSocketMsg
```
melekes: is msgI guaranteed to be PrivValidatorSocketMsg in case of error?

types/priv_validator/socket.go
```diff
@@ -0,0 +1,273 @@
...
+				return // Ignore error from listener closing.
+			}
+			pvss.Logger.Error("Failed to accept connection: " + err.Error())
+			continue
+		}
+
+		pvss.Logger.Info("Accepted a new connection")
+
+		// read/write
+		for {
```
melekes: only one connection at a time?

types/priv_validator/socket.go
```diff
@@ -0,0 +1,273 @@
...
+		addr:    addr,
+		privVal: privVal,
+		chainID: chainID,
+	}
+	pvss.BaseService = *cmn.NewBaseService(logger, "privValidatorSocketServer", pvss)
+	return pvss
+}
+
+func (pvss *PrivValidatorSocketServer) OnStart() error {
+	if err := pvss.BaseService.OnStart(); err != nil {
```
melekes: BaseService#OnStart is empty. no need to call it.

types/priv_validator/socket.go
```diff
@@ -0,0 +1,273 @@
...
+	if err != nil {
+		return err
+	}
+	pvss.listener = ln
+	go pvss.acceptConnectionsRoutine()
+	return nil
+}
+
+func (pvss *PrivValidatorSocketServer) OnStop() {
+	pvss.BaseService.OnStop()
```
melekes: BaseService#OnStop is empty. no need to call it.

types/priv_validator/socket.go
```diff
@@ -0,0 +1,273 @@
...
+	if err := pvss.listener.Close(); err != nil {
+		pvss.Logger.Error("Error closing listener", "err", err)
+	}
+
+	if err := pvss.conn.Close(); err != nil {
```
melekes: check if `pvss.conn != nil`

types/priv_validator/socket.go
xla: @ebuchman Is one connection at a time by design, or should the server be able to hanlde multiple, if so are there any upper bounds we want to enforce, or any other limiting requirements?

ebuchman:
Thanks Alex.

Couple notes here. 

- Server should be on the Tendermint side. The signer should dial Tendermint.
- Should be able to handle multiple connections, but let's cap it at something reasonable (maybe 3). 
- Let's use go-wire for serialization
- Let's use the p2p.SecretConnection for authentication

@tarcieri is working on the signer itself in Rust

ebuchman:
> Should the server exist as it's own binary under cmd/priv_validator or ultimately become a command of the tendermint cli?

No it should run in-process when `tendermint node` boots up. A tendermint node always needs a priv validator - either it's given a priv_validator.json with a private key direcetly (ie. useful for dev/testing) or it starts an in-process listener and waits to receive a connection from an external signing process

> Going to sync with him to see if we can make each others work easier.

Sounds great

xla:
@ebuchman @melekes This is ready for review from my perspective. Tried to address all points.

When working with the protocol between client and server I was dearly missing some form of communicating errors,  but I don't know how settle we are on the current protocol.

ebuchman:
Thanks Alexis. I've moved this over to a [fresh PR](https://github.com/tendermint/tendermint/pull/1204) by rebasing on a fresh copy of the new validator code without integrating to the rest of the codebase. Let's take up review/discussion there

Issue #1020, opened 2017-12-28
lite: NewInquiring's signature should return an error


odeke-em:
I currently see https://github.com/tendermint/tendermint/blob/a2b92c07453f66277f70d0c3979a9dac67bbecea/lite/inquirer.go#L26-L29

in there there is a `TODO` to handle error. Seems to me like this function is a perfect fit for a signature with an error i.e
```go
func NewInquiring(chainID string, fc FullCommit, trusted, source Provider) (*Inquiring, error)
```
How come we had the original signature not returning an error but instead swallowing up the error and we left a TODO?

PR #1063, opened 2018-01-05
NewInquiring returns error instead of swallowing it

adrianbrink:
closes: #1020

### DIFF
--- a/lite/inquirer_test.go
+++ b/lite/inquirer_test.go
@@ -32,7 +32,8 @@ func TestInquirerValidPath(t *testing.T) {
 		vals := keys.ToValidators(vote, 0)
 		h := int64(20 + 10*i)
 		appHash := []byte(fmt.Sprintf("h=%d", h))
-		commits[i] = keys.GenFullCommit(chainID, h, nil, vals, appHash, consHash, resHash, 0, len(keys))
+		commits[i] = keys.GenFullCommit(chainID, h, nil, vals, appHash, consHash, resHash, 0,
+			len(keys))
 	}
 
 	// initialize a certifier with the initial state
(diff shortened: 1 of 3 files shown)


lite/inquirer_test.go
```diff
@@ -32,11 +32,12 @@ func TestInquirerValidPath(t *testing.T) {
...
 		h := int64(20 + 10*i)
 		appHash := []byte(fmt.Sprintf("h=%d", h))
-		commits[i] = keys.GenFullCommit(chainID, h, nil, vals, appHash, consHash, resHash, 0, len(keys))
+		commits[i] = keys.GenFullCommit(chainID, h, nil, vals, appHash, consHash, resHash, 0,
+			len(keys))
 	}
 
 	// initialize a certifier with the initial state
-	cert := lite.NewInquiring(chainID, commits[0], trust, source)
+	cert, _ := lite.NewInquiring(chainID, commits[0], trust, source)
```
melekes: maybe add a `require.NoError(t, err)`?

lite/proxy/certifier.go
```diff
@@ -25,6 +25,11 @@ func GetCertifier(chainID, rootDir, nodeAddr string) (*lite.Inquiring, error) {
 	if err != nil {
 		return nil, err
 	}
-	cert := lite.NewInquiring(chainID, fc, trust, source)
+
+	cert, err := lite.NewInquiring(chainID, fc, trust, source)
```
odeke-em: Minor nit, we can just do
```go
return lite.NewInquiring(chainID, fc, trust, source)
```
instead of the the if-clause

odeke-em:
Thanks @adrianbrink for the fix, and @melekes and @ebuchman for the reviews!

Issue #862, opened 2017-11-17
PEX: seed/crawler mode


ebuchman:
- [x] disconnect from an incoming connection after sending a `pexAddrsMessage` with list of good recent peers
- [x] maintain persistent connections with other known seeds (manually specified in config) 
- [x] continuously crawl for new nodes and to validate old nodes (for now validate can just mean that we can connect to the node and get a list of peers from them. but soon it should be more involved, ie. using the trust metric and vetting them).

ebuchman:
https://github.com/tendermint/tendermint/pull/1128

PR #1067, opened 2018-01-06
seed/crawler mode for the PEX reactor

caffix:
#862


p2p/pex_reactor.go
```diff
@@ -47,18 +52,22 @@ type PEXReactor struct {
 	book              *AddrBook
 	ensurePeersPeriod time.Duration
 
+	// Seed/Crawler Mode
+	runningSeedCrawler bool
```
melekes: let's use either Seed or Crawler, but not both... `seedMode bool` or `crawlerMode bool`

p2p/pex_reactor.go
```diff
@@ -47,18 +52,22 @@ type PEXReactor struct {
...
+
 	// tracks message count by peer, so we can prevent abuse
 	msgCountByPeer    *cmn.CMap
 	maxMsgCountByPeer uint16
 }
 
 // NewPEXReactor creates new PEX reactor.
-func NewPEXReactor(b *AddrBook) *PEXReactor {
+func NewPEXReactor(b *AddrBook, sm bool) *PEXReactor {
```
melekes: `seedMode`. how otherwise one should guess what `sm bool` is? there is no comment

melekes: also. should this be an option for pex reactor? or we should create `PexReactorSeedMode` given how different are normal vs seed behaviours

p2p/pex_reactor.go
```diff
@@ -293,6 +309,160 @@ func (r *PEXReactor) ensurePeers() {
...
 }
 
+// Explores the network searching for more peers. (continuous)
+// Seed/Crawler Mode causes this node to quickly disconnect
+// from peers, except other seed nodes.
+func (r *PEXReactor) seedCrawlerMode() {
+	// Do an initial crawl
+	r.crawlPeers()
+
+	// Fire periodically
```
melekes: code is self describing enough. this comment only duplicates code

p2p/pex_reactor.go
```diff
@@ -293,6 +309,160 @@ func (r *PEXReactor) ensurePeers() {
...
+	return oa
+}
+
+// addrToPeerKey returns the Peer.Key() if the Peer is currently
+// connected and has the RemoteAddr equal to the addr parameter.
+func (r *PEXReactor) addrToPeerKey(addr *NetAddress) string {
+	var id string
+
+	peers := r.Switch.peers.List()
+	for _, p := range peers {
```
melekes: there is a `peers.Has` function

p2p/pex_reactor.go
```diff
@@ -293,6 +309,160 @@ func (r *PEXReactor) ensurePeers() {
...
+	sort.Sort(oa)
+	return oa
+}
+
+// addrToPeerKey returns the Peer.Key() if the Peer is currently
+// connected and has the RemoteAddr equal to the addr parameter.
+func (r *PEXReactor) addrToPeerKey(addr *NetAddress) string {
+	var id string
+
+	peers := r.Switch.peers.List()
```
melekes: let's move this up. i.e. call it once inside `getCrawlStatus` and pass to `peerID`

p2p/pex_reactor.go
```diff
@@ -293,6 +309,160 @@ func (r *PEXReactor) ensurePeers() {
...
+func (oa oldestAttempt) Len() int           { return len(oa) }
+func (oa oldestAttempt) Swap(i, j int)      { oa[i], oa[j] = oa[j], oa[i] }
+func (oa oldestAttempt) Less(i, j int) bool { return oa[i].LastAttempt.Before(oa[j].LastAttempt) }
+
+// getCrawlStatus returns addresses of potential peers that we wish to validate.
+// NOTE: The status information is ordered as described above.
+func (r *PEXReactor) getCrawlStatus() []crawlStatus {
+	var oa oldestAttempt
+
+	addrs := r.book.List()
```
melekes: is there a reason why we want all addresses and not `GetSelection`?

p2p/addrbook.go
```diff
@@ -169,6 +169,18 @@ func (a *AddrBook) OurAddresses() []*NetAddress {
 	return addrs
 }
 
+// List returns a list of new and old addresses.
+func (a *AddrBook) List() []*knownAddress {
```
melekes: Public function with private return args. I think func should be like `GetSelection`

p2p/pex_reactor.go
```diff
@@ -293,6 +309,160 @@ func (r *PEXReactor) ensurePeers() {
...
 
+// Explores the network searching for more peers. (continuous)
+// Seed/Crawler Mode causes this node to quickly disconnect
+// from peers, except other seed nodes.
+func (r *PEXReactor) seedCrawlerMode() {
+	// Do an initial crawl
+	r.crawlPeers()
+
+	// Fire periodically
+	ticker := time.NewTicker(5 * time.Second)
```
melekes: let's make this a const

melekes:
Good work 👍

p2p/addrbook.go
caffix: Since I need the additional information provided by knownAddress, would you like the function to be private instead?

p2p/pex_reactor.go
caffix: Okay.

p2p/pex_reactor.go
caffix: The same type of comment can be found in ensurePeersRoutine(), which is where the idea came from, but I'll remove it.

p2p/pex_reactor.go
caffix: I have changed the parameter name, yet regarding the method to pass the option, I struggled with this as well. I'll gladly change it to whichever method everyone agrees is best.

p2p/pex_reactor.go
caffix: Sure.

p2p/pex_reactor.go
caffix: GetSelection only returns some of the known addresses, instead of allowing the crawler to reach out to all the known addresses and ensure we're increasing its visibility of the network.

p2p/pex_reactor.go
caffix: Good call.

p2p/pex_reactor.go
caffix: Yes, but it performs the lookup using the peer.Key(), not the peer.NodeInfo().RemoteAddr

p2p/pex_reactor.go
melekes: @ebuchman what do you think?

p2p/pex_reactor.go
melekes: Maybe we need to rename `Has` to `FindByID()` and add `FindByRemoteAddr()`?

p2p/addrbook.go
melekes: no. let's make `knownAddress` public. also returning knownAddresses and not NetAddress creates a dissonance (maybe there is a better word for it). i.e. imagine some array ds with API:

```
all() -> []innerElem
some() -> []Element
```

let's rename `List` to `ListOfKnownAddresses()`

p2p/pex_reactor.go
caffix: Done.

p2p/addrbook.go
caffix: Okay.

ebuchman:
rebased in https://github.com/tendermint/tendermint/pull/1128

PR #1068, opened 2018-01-06
STALLED: Generalized Makefile parameters to enable easier copy and customizati…

greg-szabo:
…on to other repos

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,29 +1,36 @@
+# Repository specific variables
+BINARY := tendermint
+TMHOME ?= $(HOME)/.tendermint
+LDFLAGS_VERSIONING := -X github.com/tendermint/tendermint/version.GitCommit=$(shell git rev-parse --short=8 HEAD)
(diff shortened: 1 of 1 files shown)


ebuchman:
What's this stalled on ? And @zach how does this break it doesnt change `make install`. 

Let's sync with @jaekwon on what he's doing for Makefile in SDK

greg-szabo:
There have been a few discussions around the `Makefile` and how it should be changed. This change is obsolete now in three ways:
- Some of the cosmetic changes were incorporated into Anton's or Jae's `Makefile` changes
- Some changes are obsolete now as the technology changed. E.g.: we are getting rid of `gox`.
- Some changes we could not agree on and things turned a different direction (go checks, testing environment, multi-arch builds). These might not be ideal for DevOps automation but for the sake of moving forward (disagree and commit), it's better to go with the flow and get something in place.

Hence I'm closing this change.

PR #1069, opened 2018-01-06
add quick start guide

zramsay:
- quick start guide for Tendermint
- script for quick install on Ubuntu
- default files for quick cluster creation


docs/examples/getting-started.md
```diff
@@ -0,0 +1,125 @@
+# Tendermint
+
+## Overview
+
+Tendermint is ...
+
+## Install
+
+Note: Quick Start on a fresh machine can be done with [this script](https://gist.githubusercontent.com/zramsay/4bca8907a97ad825c30cfdc6f72ba97b/raw/b94bf36e34181a2960f9fe0d3eff260d03b6064e/install_tendermint.sh); tested on Ubuntu 16.04. It is used in the cluster deployment below.
```
melekes: let's add a command to run

```
curl -L https://gist.githubusercontent.com/zramsay/4bca8907a97ad825c30cfdc6f72ba97b/raw/b94bf36e34181a2960f9fe0d3eff260d03b6064e/install_tendermint.sh | bash
```
 or
```
curl -L https://git.io/vNTiV | bash
```

docs/examples/getting-started.md
```diff
@@ -0,0 +1,125 @@
...
+
+Requires: 
+- `go` minimum version 1.9.2
+- `$GOPATH` set and `$GOPATH/bin` on your $PATH (see https://github.com/tendermint/tendermint/wiki/Setting-GOPATH)
+
+To create the `tendermint` binary, run:
```
melekes: `To install Tendermint, run` ?

docs/examples/getting-started.md
```diff
@@ -0,0 +1,125 @@
...
+where the value is returned in hex.
+
+## Cluster of Nodes
+
+First create four Ubuntu cloud machines. The following was testing on Digital Ocean Ubuntu 16.04 x64 (3GB/1CPU, 20GB SSD). We'll refer to their respective IP addresses below as IP1, IP2, IP3, IP4.
+
+Then, `ssh` into each machine, and `curl` then execute [this script](https://gist.githubusercontent.com/zramsay/4bca8907a97ad825c30cfdc6f72ba97b/raw/b94bf36e34181a2960f9fe0d3eff260d03b6064e/install_tendermint.sh):
+
+```
+curl -o tmint.sh https://gist.githubusercontent.com/zramsay/4bca8907a97ad825c30cfdc6f72ba97b/raw/b94bf36e34181a2960f9fe0d3eff260d03b6064e/install_tendermint.sh
```
melekes: `curl -L https://git.io/vNTiV | bash`

docs/examples/install_tendermint.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+# get and unpack golang
+curl -O https://storage.googleapis.com/golang/go1.9.2.linux-amd64.tar.gz
```
melekes: this will not work for x86 and macOS... maybe we need an `if OS == darwin`

docs/examples/install_tendermint.sh
```diff
@@ -0,0 +1,29 @@
...
+tar -xvf go1.9.2.linux-amd64.tar.gz
+
+apt install make
+
+## move go and add binary to path
+mv go /usr/local
+echo "export PATH=\$PATH:/usr/local/go/bin" >> ~/.profile
+
+## create the GOPATH directory, set GOPATH and put on PATH
+mkdir goApps
```
melekes: is this from official guides?

docs/examples/install_tendermint.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+# get and unpack golang
+curl -O https://storage.googleapis.com/golang/go1.9.2.linux-amd64.tar.gz
+tar -xvf go1.9.2.linux-amd64.tar.gz
+
+apt install make
```
melekes: what if no `apt`? what about `yum`?

docs/examples/install_tendermint.sh
```diff
@@ -0,0 +1,29 @@
...
+
+source ~/.profile
+
+## get the code and move into it
+REPO=github.com/tendermint/tendermint
+go get $REPO
+cd $GOPATH/src/$REPO
+
+## build
+git checkout v0.15.0
```
melekes: maybe we need to check for `git` as well or install it above (`apt-get install make git`)

docs/examples/getting-started.md
```diff
@@ -0,0 +1,125 @@
...
+bash tmint.sh
+source ~/.profile
+```
+
+This will install `go` and other dependencies, get the Tendermint source code, then compile the `tendermint` binary.
+
+Next, `cd` into `docs/examples`. Each command below should be run from each node, in sequence:
```
melekes: Can't we just run `tendermint node --home ./nodeN --proxy_app=dummy --p2p.seeds IP1:46656,IP2:46656,IP3:46656,IP4:46656` on each node?

docs/examples/getting-started.md
```diff
@@ -0,0 +1,125 @@
+# Tendermint
+
+## Overview
+
+Tendermint is ...
```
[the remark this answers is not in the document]
zramsay: it's meant to be a one-liner filled in last. It's probably not needed but it's what was asked for

docs/examples/getting-started.md
zramsay: Note: this link will change to refer to the file currently being check-in with this PR, however, that file should point to master branch, so that can wait until release

docs/examples/getting-started.md
zramsay: sure but that's more text, and will lead to more `Error can't connect to peer` than otherwise

docs/examples/install_tendermint.sh
zramsay: i know, that's explicit in this line https://github.com/tendermint/tendermint/pull/1069/files#diff-efa5271fdabd2250b8d2dd60cfc5ff1dR9

This script is meant explicitly for a Ubuntu on a DO box. I'll make it clearer in the script itself

docs/examples/install_tendermint.sh
zramsay: see previous comment. it's not meant a multi-OS check-all-the-things scripts, but rather as a quick and dirty get-started-ASAP-on-Digital-OCean

docs/examples/install_tendermint.sh
zramsay: the official guide doesn't even set a `$GOPATH`, so no. AFAICT there is no standard for naming this directory (I've seen `work`, `gocode`, `go, `goApps` etc, the list goes on)

docs/examples/install_tendermint.sh
zramsay: git is already instaled on Ubnutn 16.04

PR #1073, opened 2018-01-06
Checkin vendor directory

adrianbrink:
This stems from a lengthy discussion on Slack.

Users should have two choices to install Tendermint or any of our other software. They should either install from source or from a binary. This means that we need to publish signed source code as well as signed binaries. 

Checking in the vendor directory allows us to distribute signed source code that can be installed with `go get github.com/tendermint/tendermint`. A user will only need to verify that the latest commit to master is properly signed. Furthermore, by checking in the vendor directory a user can install Tendermint without any other tools, such as `gox`, `xgo` or `glide`. This means that we do not have to audit those tools, as they are not part of the `go get github.com/tendermint/tendermint` build process. 

If all validators would install from source and verify the signature of the latest commit, the network will remain secure as long as <1/3 of validators have a backdoored go toolchain. Once we combine the `installing from **signed source** and verifying signatures` with `downloading **signed binaries** and verifying signatures` we will have an extremely diverse and hence robust installation process. If every validator goes with their own preference we will be protected against some of them having backdoored toolchains and our build server being backdoored. 

Lastly, I strongly advocate making installing from source a standard way to install production software, because it is much easier to audit signed source code than it is to audit signed binaries. If downloading a binary and verifying it is the only displayed way to install Tendermint, our process of binary production is a huge point of failure. I personally would feel much safer if >1/3 of validators would install from source using `go get`.


adrianbrink:
I've regenerated the `glide.yaml` file. Now it reflects accurately the dependencies we currently have within Tendermint. In the future it is advisable to only use `glide get github.com/tendermint/iavl` to add new dependencies to ensure that they are tracked in the yaml file.

ethanfrey:
This came up on slack a week or two ago.

My suggestion, which was also supported by @melekes and @odeke-em was that non-master branches use glide as we currently do, but we check in the vendor dir on master along with every release.

This makes master secure, but also allows us to update deps easily on branches

adrianbrink:
We will still use `glide` on every branch. If you want to update a dependency you update the `glide.yaml` file and run `glide up`. From a developer perspective, nothing changes as far as I can tell.

Greg-Szabo:
I think this is not mutually exclusive, since branches are all in your hands. The only rule we have to make is that vendor/ folders are pointed out if they need to be merged, right?

adrianbrink:
As discussed in length on slack, we do not want to check in the vendor directory.

Issue #1080, opened 2018-01-08
lite/proxy: StartProxy should return the error from service.Start()


odeke-em:
We need to fix this code
https://github.com/tendermint/tendermint/blob/cf426111876e78a341f73fdb459d166dc513d725/lite/proxy/proxy.go#L21

to
```go
if err := c.Start(); err != nil {
       return err
}
```

because even the docs for Service say
```go
// Service defines a service that can be started, stopped, and reset.
type Service interface {
	// Start the service.
	// If it's already started or stopped, will return an error.
	// If OnStart() returns an error, it's returned by Start()
	Start() error
```
https://github.com/tendermint/tmlibs/blob/91b4b534ad78e442192c8175db92a06a51064064/common/service.go#L18-L21

Issue #1079, opened 2018-01-08
lite: GetCertifiedCommit never actually certifies commit due to invalid return ignoring returned err


odeke-em:
I was just reading through the lite/proxy code and noticed this in GetCertifiedCommit

https://github.com/tendermint/tendermint/blob/381fe19335ba8825e04d1d0fefa5cac709bb7178/lite/proxy/query.go#L118-L119

The above I believe was an innocent mistype/copy+paste error but it invalidates the entire purpose of certifying commits and if I am reading that code right, irrespective of the result of certifying always returns a nil error. This to me is a potential vulnerability as long as we can get a commit with the same height as passed in.

Let's avoid this by fixing it to
```go
if err := cert.Certify(commit); err != nil {
    return empty, err
}
return commit, nil
```

and also let's gut out the weird named return signature whose purpose I believe was to have `empty` defined but that's iffy, could we please make it
```
func GetCertifiedCommit(h int64, node rpcclient.Client,
	cert lite.Certifier) (lite.Commit,error) {
 
       var empty lite.Commit
```

adrianbrink:
Yep, probably our first vulnerability. :-)

Issue #1018, opened 2017-12-28
lite: clarify the docs on "Static" and also perhaps use a more intuitive name


odeke-em:
I was just combing through package lite and noticed we have a struct called "Static"

The docs for it say "Static assumes a static set of validators, set on initilization and checks against them. The signatures on every header is checked for > 2/3 votes against the known validator set upon Certify

Good for testing or really simple chains. Building block to support real-world functionality."

Perhaps we could give it a name such as "StaticValidatorsChecker"? I didn't know what to make of "Static"

Also perhaps we could update the docs a little more

"StaticValidatorsChecker assumes a static set of validators, that are set on initilization and it checks against them. The signature on every header is checked for > 2/3 votes against the known validator set upon every invocation of its method Certify

It is good for testing or for use with really simple chains. It is a buuilding block to support real-world functionality." -- let's perhaps add some more content to "real-world functionality"?

odeke-em:
Also a minor nit that we can clear up while here is change
https://github.com/tendermint/tendermint/blob/d14ec7d7d2f407433070665536994e0a1eafab4b/lite/static.go#L13
to
```go
var _ Certifier = (*Static)(nil)
```

ethanfrey:
It was `StaticCertifier`, then I renamed the package when importing into tendermint from light-client, so it was `certifiers.Static` to limit redundancy.

@adrianbrink  then renamed the package to lite, but didn't change the names. Maybe he can finish that cleanup?

Is StaticCertifier / DynamicCertifier / InquiringCertifier clear enough for you?

zramsay:
closed in #1081

PR #1081, opened 2018-01-08
Vulnerability in light client proxy

adrianbrink:
When calling GetCertifiedCommit the light client proxy would call
Certify and even on error return the Commit as if it had been correctly
certified.

Now it returns the error correctly and returns an empty Commit on error.

closes: #1080 #1079 #1018


lite/dynamic.go
```diff
@@ -6,9 +6,9 @@ import (
 	liteErr "github.com/tendermint/tendermint/lite/errors"
 )
 
-var _ Certifier = &Dynamic{}
+var _ Certifier = (*DynamicCertifier)(nil)
```
melekes: 👍  don't forget to rename files (dynamic -> dynamic_certifier)

odeke-em:
@adrianbrink could you point me to the code for the actual fix? I see 16 files edited but can't easily tell what's going on, given the title for the PR being to fix the vulnerability identified.

adrianbrink:
https://github.com/tendermint/tendermint/pull/1081/files#diff-1857e75bc40183b1ffa2ddb870dbd0edR118

Before we never returned the error.

PR #1095, opened 2018-01-10
[p2p] Pong Timeout

melekes:
Refs #804

Original PR: https://github.com/tendermint/tendermint/pull/832

### DIFF
--- a/p2p/conn/connection.go
+++ b/p2p/conn/connection.go
@@ -34,8 +34,8 @@ const (
 	defaultSendRate            = int64(512000) // 500KB/s
 	defaultRecvRate            = int64(512000) // 500KB/s
 	defaultSendTimeout         = 10 * time.Second
-	defaultPingInterval        = 40 * time.Second
-	defaultPongTimeout         = 35 * time.Second
+	defaultPingInterval        = 60 * time.Second
+	defaultPongTimeout         = 45 * time.Second
 )
 
 type receiveCbFunc func(chID byte, msgBytes []byte)
@@ -134,7 +134,7 @@ func NewMConnection(conn net.Conn, chDescs []*ChannelDescriptor, onReceive recei
 // NewMConnectionWithConfig wraps net.Conn and creates multiplex connection with a config
 func NewMConnectionWithConfig(conn net.Conn, chDescs []*ChannelDescriptor, onReceive receiveCbFunc, onError errorCbFunc, config *MConnConfig) *MConnection {
 	if config.pongTimeout >= config.pingInterval {
-		panic("pongTimeout must be less than pingInterval")
+		panic("pongTimeout must be less than pingInterval (otherwise, next ping will reset pong timer)")
 	}
 
 	mconn := &MConnection{
--- a/p2p/conn/connection.go
+++ b/p2p/conn/connection.go
@@ -153,7 +153,7 @@ func NewMConnectionWithConfig(conn net.Conn, chDescs []*ChannelDescriptor, onRec
 		sendMonitor: flow.New(0, 0),
 		recvMonitor: flow.New(0, 0),
 		send:        make(chan struct{}, 1),
-		pong:        make(chan struct{}),
+		pong:        make(chan struct{}, 1),
 		onReceive:   onReceive,
 		onError:     onError,
 		config:      config,
@@ -191,7 +191,7 @@ func (c *MConnection) OnStart() error {
 	c.quit = make(chan struct{})
 	c.flushTimer = cmn.NewThrottleTimer("flush", c.config.FlushThrottle)
 	c.pingTimer = cmn.NewRepeatTimer("ping", c.config.PingInterval)
-	c.pongTimeoutCh = make(chan bool)
+	c.pongTimeoutCh = make(chan bool, 1)
 	c.chStatsTimer = cmn.NewRepeatTimer("chStats", updateStats)
 	go c.sendRoutine()
 	go c.recvRoutine()
@@ -492,8 +492,8 @@ FOR_LOOP:
 			c.Logger.Debug("Receive Pong")
 			select {
 			case c.pongTimeoutCh <- false:
-			case <-c.quit:
-				break FOR_LOOP
+			default:
+				// never block
 			}
 		case packetTypeMsg:
 			pkt, n, err := msgPacket{}, int(0), error(nil)
--- a/p2p/conn/connection.go
+++ b/p2p/conn/connection.go
@@ -2,6 +2,7 @@ package conn
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"io"
 	"math"
@@ -22,6 +23,7 @@ const (
 	minWriteBufferSize = 65536
 	updateStats        = 2 * time.Second
 	pingTimeout        = 40 * time.Second
+	pongTimeout        = 60 * time.Second
 
 	// some of these defaults are written in the user config
 	// flushThrottle, sendRate, recvRate
@@ -84,6 +86,7 @@ type MConnection struct {
 	quit         chan struct{}
 	flushTimer   *cmn.ThrottleTimer // flush writes as necessary but throttled.
 	pingTimer    *cmn.RepeatTimer   // send pings periodically
+	pongTimer    *cmn.ThrottleTimer // close conn if pong not recv in 1 min
 	chStatsTimer *cmn.RepeatTimer   // update channel stats periodically
 
 	created time.Time // time of creation
@@ -170,6 +173,7 @@ func (c *MConnection) OnStart() error {
 	c.quit = make(chan struct{})
 	c.flushTimer = cmn.NewThrottleTimer("flush", c.config.FlushThrottle)
 	c.pingTimer = cmn.NewRepeatTimer("ping", pingTimeout)
+	c.pongTimer = cmn.NewThrottleTimer("pong", pongTimeout)
 	c.chStatsTimer = cmn.NewRepeatTimer("chStats", updateStats)
 	go c.sendRoutine()
 	go c.recvRoutine()
@@ -181,6 +185,7 @@ func (c *MConnection) OnStop() {
 	c.BaseService.OnStop()
 	c.flushTimer.Stop()
 	c.pingTimer.Stop()
+	c.pongTimer.Stop()
 	c.chStatsTimer.Stop()
 	if c.quit != nil {
 		close(c.quit)
@@ -315,7 +320,12 @@ FOR_LOOP:
 			c.Logger.Debug("Send Ping")
 			wire.WriteByte(packetTypePing, c.bufWriter, &n, &err)
 			c.sendMonitor.Update(int(n))
-			c.flush()
+			go c.flush()
+			c.Logger.Debug("Starting pong timer")
+			c.pongTimer.Set()
+		case <-c.pongTimer.Ch:
+			c.Logger.Debug("Pong timeout")
+			err = errors.New("pong timeout")
 		case <-c.pong:
 			c.Logger.Debug("Send Pong")
 			wire.WriteByte(packetTypePong, c.bufWriter, &n, &err)
@@ -454,8 +464,8 @@ FOR_LOOP:
 				// never block
 			}
 		case packetTypePong:
-			// do nothing
 			c.Logger.Debug("Receive Pong")
+			c.pongTimer.Unset()
 		case packetTypeMsg:
 			pkt, n, err := msgPacket{}, int(0), error(nil)
 			wire.ReadBinaryPtr(&pkt, c.bufReader, c.config.maxMsgPacketTotalSize(), &n, &err)
--- a/p2p/conn/connection.go
+++ b/p2p/conn/connection.go
@@ -22,8 +22,6 @@ const (
 	minReadBufferSize  = 1024
 	minWriteBufferSize = 65536
 	updateStats        = 2 * time.Second
-	pingTimeout        = 40 * time.Second
-	pongTimeout        = 60 * time.Second
 
 	// some of these defaults are written in the user config
 	// flushThrottle, sendRate, recvRate
@@ -36,6 +34,8 @@ const (
 	defaultSendRate            = int64(512000) // 500KB/s
 	defaultRecvRate            = int64(512000) // 500KB/s
 	defaultSendTimeout         = 10 * time.Second
+	defaultPingTimeout         = 40 * time.Second
+	defaultPongTimeout         = 60 * time.Second
 )
 
 type receiveCbFunc func(chID byte, msgBytes []byte)
@@ -100,6 +100,9 @@ type MConnConfig struct {
 	MaxMsgPacketPayloadSize int
 
 	FlushThrottle time.Duration
+
+	pingTimeout time.Duration
+	pongTimeout time.Duration
 }
 
 func (cfg *MConnConfig) maxMsgPacketTotalSize() int {
@@ -113,6 +116,8 @@ func DefaultMConnConfig() *MConnConfig {
 		RecvRate:                defaultRecvRate,
 		MaxMsgPacketPayloadSize: defaultMaxMsgPacketPayloadSize,
 		FlushThrottle:           defaultFlushThrottle,
+		pingTimeout:             defaultPingTimeout,
+		pongTimeout:             defaultPongTimeout,
 	}
 }
 
@@ -172,8 +177,8 @@ func (c *MConnection) OnStart() error {
 	}
 	c.quit = make(chan struct{})
 	c.flushTimer = cmn.NewThrottleTimer("flush", c.config.FlushThrottle)
-	c.pingTimer = cmn.NewRepeatTimer("ping", pingTimeout)
-	c.pongTimer = cmn.NewThrottleTimer("pong", pongTimeout)
+	c.pingTimer = cmn.NewRepeatTimer("ping", c.config.pingTimeout)
+	c.pongTimer = cmn.NewThrottleTimer("pong", c.config.pongTimeout)
 	c.chStatsTimer = cmn.NewRepeatTimer("chStats", updateStats)
 	go c.sendRoutine()
 	go c.recvRoutine()
--- a/p2p/conn/connection.go
+++ b/p2p/conn/connection.go
@@ -320,7 +320,7 @@ FOR_LOOP:
 			c.Logger.Debug("Send Ping")
 			wire.WriteByte(packetTypePing, c.bufWriter, &n, &err)
 			c.sendMonitor.Update(int(n))
-			// c.flush
+			// should be c.flush
 			go c.flush()
 			c.Logger.Debug("Starting pong timer")
 			c.pongTimer.Set()
--- a/p2p/conn/connection.go
+++ b/p2p/conn/connection.go
@@ -351,12 +351,7 @@ FOR_LOOP:
 			c.sendMonitor.Update(int(n))
 			c.flush()
 		case <-c.quit:
-			if c.pongTimer != nil {
-				if !c.pongTimer.Stop() {
-					<-c.pongTimer.C
-				}
-				drain(c.pongTimeoutCh)
-			}
+			c.stopPongTimer()
 			break FOR_LOOP
 		case <-c.send:
 			// Send some msgPackets
@@ -482,6 +477,7 @@ FOR_LOOP:
 		switch pktType {
 		case packetTypePing:
 			// TODO: prevent abuse, as they cause flush()'s.
+			// https://github.com/tendermint/tendermint/issues/1190
 			c.Logger.Debug("Receive Ping")
 			select {
 			case c.pong <- struct{}{}:
@@ -490,12 +486,7 @@ FOR_LOOP:
 			}
 		case packetTypePong:
 			c.Logger.Debug("Receive Pong")
-			if c.pongTimer != nil {
-				if !c.pongTimer.Stop() {
-					<-c.pongTimer.C
-				}
-				drain(c.pongTimeoutCh)
-			}
+			c.stopPongTimer()
 		case packetTypeMsg:
 			pkt, n, err := msgPacket{}, int(0), error(nil)
 			wire.ReadBinaryPtr(&pkt, c.bufReader, c.config.maxMsgPacketTotalSize(), &n, &err)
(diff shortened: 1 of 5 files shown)


melekes:
2 peers
4 channels

```
BenchmarkSwitchBroadcast-2         30000             71275 ns/op
--- BENCH: BenchmarkSwitchBroadcast-2
        switch_test.go:339: success: 1, failure: 0
        switch_test.go:339: success: 100, failure: 0
        switch_test.go:339: success: 10000, failure: 0
        switch_test.go:339: success: 30000, failure: 0
```

Tested code:

```
func (sw *Switch) Broadcast(chID byte, msg interface{}) chan bool {
	successChan := make(chan bool, len(sw.peers.List()))
	sw.Logger.Debug("Broadcast", "channel", chID, "msg", msg)
	var wg sync.WaitGroup
	for _, peer := range sw.peers.List() {
		wg.Add(1)
		go func(peer Peer) {
			defer wg.Done()
			success := peer.Send(chID, msg)
			successChan <- success
		}(peer)
	}
	go func() {
		wg.Wait()
		close(successChan)
	}()
	return successChan
}
```

p2p/connection.go
```diff
@@ -176,7 +186,11 @@ func (c *MConnection) OnStart() error {
 	}
 	c.quit = make(chan struct{})
 	c.flushTimer = cmn.NewThrottleTimer("flush", c.config.flushThrottle)
-	c.pingTimer = cmn.NewRepeatTimer("ping", pingTimeout)
+	c.pingTimer = cmn.NewRepeatTimer("ping", c.config.pingInterval)
+	c.pongTimer = time.NewTimer(c.config.pongTimeout)
+	// we start timer once we've send ping; needed here because we use start
+	// listening in recvRoutine
+	_ = c.pongTimer.Stop()
```
caffix: Why the assignment to only a blank identifier?

p2p/connection.go
```diff
@@ -188,6 +202,7 @@ func (c *MConnection) OnStop() {
 	c.BaseService.OnStop()
 	c.flushTimer.Stop()
 	c.pingTimer.Stop()
+	_ = c.pongTimer.Stop()
```
caffix: Same as the question/comment above related to the blank identifier.

p2p/switch.go
```diff
@@ -398,16 +399,22 @@ func (sw *Switch) IsDialing(addr *NetAddress) bool {
 // trying to send for defaultSendTimeoutSeconds. Returns a channel
 // which receives success values for each attempted send (false if times out).
 // NOTE: Broadcast uses goroutines, so order of broadcast may not be preserved.
-// TODO: Something more intelligent.
 func (sw *Switch) Broadcast(chID byte, msg interface{}) chan bool {
```
caffix: Regarding issue #805: What if we kept the implementation similar to this, yet each broadcast was queued, so that broadcast X must finish before broadcast Y? The individual messages could be sent via go-routines, but the order of broadcasts would be preserved. Thoughts?

p2p/connection.go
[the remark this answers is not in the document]
ebuchman: probably for sake of the linter. lets just check the value and return error if its false ?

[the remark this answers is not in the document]
melekes: I used it to indicate that we don't care about return value here. but I can add `// nolint`

p2p/connection.go
```diff
@@ -456,8 +477,10 @@ FOR_LOOP:
 			c.Logger.Debug("Receive Ping")
 			c.pong <- struct{}{}
 		case packetTypePong:
```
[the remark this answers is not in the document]
melekes: nice idea

p2p/switch.go
melekes: Even better to have a queue for each peer. That way we won't be waiting for some peer!

melekes:
I will handle broadcast in a separate PR.

p2p/connection.go
```diff
@@ -319,6 +334,12 @@ FOR_LOOP:
 			legacy.WriteOctet(packetTypePing, c.bufWriter, &n, &err)
 			c.sendMonitor.Update(int(n))
 			c.flush()
+			c.Logger.Debug("Starting pong timer")
+			c.pongTimer.Reset(c.config.pongTimeout)
+		case <-c.pongTimer.C:
+			c.Logger.Debug("Pong timeout")
+			// XXX: should we decrease peer score instead of closing connection?
```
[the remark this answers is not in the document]
zramsay: todo's should be issues, linked in code

ebuchman:
Thanks Anton!

Do you think we could get this rebased on the `addrbook` branch (https://github.com/tendermint/tendermint/pull/1129) ? Alternatively we could try to merge this into that branch - whatever you think will be easier

p2p/conn/connection.go
```diff
@@ -319,6 +341,12 @@ FOR_LOOP:
 			legacy.WriteOctet(packetTypePing, c.bufWriter, &n, &err)
 			c.sendMonitor.Update(int(n))
 			c.flush()
+			c.Logger.Debug("Starting pong timer")
+			c.pongTimer.Reset(c.config.PongTimeout)
+		case <-c.pongTimer.C:
+			c.Logger.Debug("Pong timeout")
+			// XXX: should we decrease peer score instead of closing connection?
```
melekes: ?

ebuchman: no scoring for now, lets just close the conn

p2p/conn/connection.go
```diff
@@ -172,7 +189,11 @@ func (c *MConnection) OnStart() error {
 	}
 	c.quit = make(chan struct{})
 	c.flushTimer = cmn.NewThrottleTimer("flush", c.config.FlushThrottle)
-	c.pingTimer = cmn.NewRepeatTimer("ping", pingTimeout)
+	c.pingTimer = cmn.NewRepeatTimer("ping", c.config.PingInterval)
+	c.pongTimer = time.NewTimer(c.config.PongTimeout)
+	// we start timer once we've send ping; needed here because we use start
+	// listening in recvRoutine
+	_ = c.pongTimer.Stop()
```
ebuchman: seems bad - what if the timer fires before we call stop?  maybe we can just create it the first time we need it ?  Otherwise we probably need the 

```
if !t.Stop() {
	<-t.C
}
```

here

p2p/conn/connection.go
```diff
@@ -452,8 +480,10 @@ FOR_LOOP:
 			c.Logger.Debug("Receive Ping")
 			c.pong <- struct{}{}
 		case packetTypePong:
-			// do nothing
 			c.Logger.Debug("Receive Pong")
+			if !c.pongTimer.Stop() {
```
ebuchman: what if we receive multiple pongs in a row ? wont we stop on the first one, and then this will return false and we'll block forever ?

p2p/conn/connection_test.go
```diff
@@ -116,6 +119,40 @@ func TestMConnectionStatus(t *testing.T) {
 	assert.Zero(status.Channels[0].SendQueueSize)
 }
 
+func TestMConnectionPongTimeoutResultsInError(t *testing.T) {
```
ebuchman: Think we need more tests:

- ping/pong works correctly (ie. we don't disconnect)
- receiving many pongs is benign
- pong timeout results in error

p2p/conn/connection.go
ebuchman: lets just remove the comment

p2p/conn/connection.go
melekes: hmm. it could if pongTimeout is 0 or 1 ms... but it's 45 sec by default

ebuchman: but what about in tests when its ms ? could be a source of non-deterministic failures :(

melekes: sure, you're right. I'll fix. sorry

melekes:
@ebuchman added 2 more tests

p2p/conn/connection.go
```diff
@@ -318,13 +340,23 @@ FOR_LOOP:
 			c.Logger.Debug("Send Ping")
 			legacy.WriteOctet(packetTypePing, c.bufWriter, &n, &err)
 			c.sendMonitor.Update(int(n))
+			c.Logger.Debug("Starting pong timer", "dur", c.config.PongTimeout)
+			c.pongTimer = time.AfterFunc(c.config.PongTimeout, func() {
+				c.pongTimeoutCh <- struct{}{}
```
ebuchman: Cool! Will this leak ? If the conn stops before we receive this, will there be a stuck go-routine? Maybe we need to drain this channel when we call pongTimer.Stop ?

p2p/conn/connection.go
```diff
@@ -452,8 +484,10 @@ FOR_LOOP:
 			c.Logger.Debug("Receive Ping")
 			c.pong <- struct{}{}
```
ebuchman: This pong channel is unbuffered. I think it means we could block here if the sendRoutine exits. This might be reason why StopPeerForError doesn't work sometimes? Might also play into a leak, Would you mind thinking through that and let me know if you agree or if I'm wrong for some reason ?

p2p/conn/connection_test.go
```diff
@@ -116,6 +119,139 @@ func TestMConnectionStatus(t *testing.T) {
...
+		// read ping
+		server.Read(make([]byte, 1))
+		serverGotPing <- struct{}{}
+		// respond with pong
+		_, err = server.Write([]byte{packetTypePong})
+		require.Nil(t, err)
+	}()
+	<-serverGotPing
+
+	pongTimerExpired := mconn.config.PongTimeout + 10*time.Millisecond
```
ebuchman: did this test fail with the old code ? 

my understanding was in this scenario with the previous code, the client should have been deadlocked after the multiple pongs

p2p/conn/connection.go
```diff
@@ -452,8 +484,10 @@ FOR_LOOP:
...
 		case packetTypePong:
-			// do nothing
 			c.Logger.Debug("Receive Pong")
+			if c.pongTimer != nil {
```
ebuchman: won't this race with sendRoutine? If the ping timer fires around when we receive a pong

melekes: it probably will. what's the solution? mutex and another channel?

p2p/conn/connection.go
melekes: hmm.. it will, most probably

p2p/conn/connection.go
melekes: also interesting why go racer does not say anything...

p2p/conn/connection_test.go
melekes: sending multiple pings, not pongs, yes!

p2p/conn/connection.go
melekes: How it prevents StopPeerForError from working? if the code is inside a goroutine. Leak? yes

p2p/conn/connection.go
melekes: https://github.com/tendermint/tendermint/pull/1095/commits/a67865cbd91b3dde2cd54690bcaf77beca6abd01

p2p/conn/connection.go
```diff
@@ -315,13 +337,21 @@ FOR_LOOP:
...
+		case <-c.pongTimeoutCh:
+			c.Logger.Debug("Pong timeout")
+			err = errors.New("pong timeout")
 		case <-c.pong:
 			c.Logger.Debug("Send Pong")
 			wire.WriteByte(packetTypePong, c.bufWriter, &n, &err)
 			c.sendMonitor.Update(int(n))
 			c.flush()
 		case <-c.quit:
+			c.stopPongTimer()
```
ebuchman: if theres an error in this loop, we just break out without processing the closed quit channel so pong timer won't get stopped. but we can't put it in OnStop because it will race. So I think it needs to also be called in the error != nil block below

p2p/conn/connection.go
```diff
@@ -446,15 +476,16 @@ FOR_LOOP:
...
 			c.Logger.Debug("Receive Ping")
 			select {
 			case c.pong <- struct{}{}:
 			case <-c.quit:
 				break FOR_LOOP
 			}
 		case packetTypePong:
-			// do nothing
 			c.Logger.Debug("Receive Pong")
+			c.stopPongTimer()
```
ebuchman: if we receive ping and pong at the same time, wont this race with the `c.pongTimer = ...` in the sendRoutine ? Maybe we need a test where that happens for the race detector to pick it up

p2p/conn/connection.go
```diff
@@ -506,6 +537,15 @@ FOR_LOOP:
 	}
 }
 
+func (c *MConnection) stopPongTimer() {
+	if c.pongTimer != nil {
+		if !c.pongTimer.Stop() {
+			<-c.pongTimer.C
```
ebuchman: suppose we receive two pongs in a row. wont we block here on the second one ?

ebuchman: Here's a diff that proves the point. All I needed to do was add a sleep to the test (without the sleep the pongTimer was just nil which is why it was passing). With the sleep, the test hangs forever

```
diff --git a/p2p/conn/connection.go b/p2p/conn/connection.go
index 40f37d29..b05a36ea 100644
--- a/p2p/conn/connection.go
+++ b/p2p/conn/connection.go
@@ -484,6 +484,7 @@ FOR_LOOP:
                                break FOR_LOOP
                        }
                case packetTypePong:
+                       fmt.Println("GOT PONG")
                        c.Logger.Debug("Receive Pong")
                        c.stopPongTimer()
                case packetTypeMsg:
@@ -538,12 +539,21 @@ FOR_LOOP:
 }
 
 func (c *MConnection) stopPongTimer() {
+       fmt.Println("STOP PONG TIMER. Is nil", c.pongTimer == nil)
        if c.pongTimer != nil {
                if !c.pongTimer.Stop() {
+                       fmt.Println("waiting on .C")
                        <-c.pongTimer.C
+                       fmt.Println("done waiting")
+               } else {
+                       fmt.Println("STOP TIMER WORKED")
                }
+               fmt.Println("draining")
                drain(c.pongTimeoutCh)
+               fmt.Println("Drained")
        }
+
+       fmt.Println("-------------")
 }
 
 type ConnectionStatus struct {
diff --git a/p2p/conn/connection_test.go b/p2p/conn/connection_test.go
index 270b4ae9..6efc6e08 100644
--- a/p2p/conn/connection_test.go
+++ b/p2p/conn/connection_test.go
@@ -173,6 +173,7 @@ func TestMConnectionMultiplePongsInTheBeginning(t *testing.T) {
        err := mconn.Start()
        require.Nil(t, err)
        defer mconn.Stop()
+       time.Sleep(time.Second * 3)
 
        // sending 3 pongs in a row (abuse)
        _, err = server.Write([]byte{packetTypePong})
```

p2p/conn/connection.go
ebuchman: Oh the race detector actually detects this for me

ebuchman: I guess we could just wrap pongTimer in a mutex but that doesn't really feel right.

An alternative is maybe to use a separate go-routine to handle a pongTimer. Then sendRoutine and recvRoutine can just talk to the pongTimerRoutine via channel/s when they need to start or stop the pongTimer.

p2p/conn/connection.go
melekes: You're right. Thanks

p2p/conn/connection.go
melekes: What if we use the same pongTimeoutCh to signal to cancel timeout? See my last commit.

p2p/conn/connection.go
melekes: I don't think this diff is correct (proves anything) because if you add `time.Sleep`, `sendRoutine` will attempt to send ping and block on `flush`. And since we're using a pipe here and you're not reading the ping (see `server.Read` calls), this whole thing will block.

ebuchman: that could be. but do you see what I'm referring to ? It still seems like it will deadlock if it receives multiple pongs

p2p/conn/connection.go
```diff
@@ -506,6 +545,17 @@ FOR_LOOP:
...
 }
 
+// not goroutine-safe
+func (c *MConnection) stopPongTimer() {
+	if c.pongTimer != nil {
+		if !c.pongTimer.Stop() {
+			<-c.pongTimer.C
+		}
+		drain(c.pongTimeoutCh)
+		c.pongTimer = nil
```
melekes: this should prevent a deadlock with multiple stops/pongs

melekes:
Done

p2p/conn/connection.go
```diff
@@ -179,12 +201,12 @@ func (c *MConnection) OnStart() error {
...
 	c.BaseService.OnStop()
-	c.flushTimer.Stop()
-	c.pingTimer.Stop()
-	c.chStatsTimer.Stop()
 	if c.quit != nil {
 		close(c.quit)
 	}
+	c.flushTimer.Stop()
+	c.pingTimer.Stop()
+	c.chStatsTimer.Stop()
```
jaekwon: reason for moving these?

p2p/conn/connection.go
```diff
@@ -315,7 +337,18 @@ FOR_LOOP:
 			c.Logger.Debug("Send Ping")
 			wire.WriteByte(packetTypePing, c.bufWriter, &n, &err)
 			c.sendMonitor.Update(int(n))
+			c.Logger.Debug("Starting pong timer", "dur", c.config.PongTimeout)
+			c.pongTimer = time.AfterFunc(c.config.PongTimeout, func() {
+				c.pongTimeoutCh <- true
```
jaekwon: this could/should be non-blocking.  probably select on .quit

p2p/conn/connection.go
```diff
@@ -503,6 +542,17 @@ FOR_LOOP:
 	}
 }
 
+// not goroutine-safe
+func (c *MConnection) stopPongTimer() {
+	if c.pongTimer != nil {
+		if !c.pongTimer.Stop() {
+			<-c.pongTimer.C
+		}
+		drain(c.pongTimeoutCh)
```
jaekwon: I am fairly sure that we are better off avoiding draining functions.
If you're not convinced, I could try to put more thought into it, but I suspect that draining is not necessary, and consistently avoiding it is better.

Here, we can keep pongTimeoutCh capacity 0, and select on line 342.

p2p/conn/connection.go
```diff
@@ -447,15 +481,20 @@ FOR_LOOP:
...
 				// never block
 			}
 		case packetTypePong:
-			// do nothing
 			c.Logger.Debug("Receive Pong")
+			select {
+			case c.pongTimeoutCh <- false:
+			default:
+				// never block
+			}
```
jaekwon: good use of default

p2p/conn/connection.go
jaekwon: It's OK to leave things in channels, as long as nothing blocks on it.  We can drain defensively, but lets instead make sending to pongTimeoutCh nonblocking, and not drain.

p2p/conn/connection.go
```diff
@@ -315,7 +337,18 @@ FOR_LOOP:
 			c.Logger.Debug("Send Ping")
 			wire.WriteByte(packetTypePing, c.bufWriter, &n, &err)
 			c.sendMonitor.Update(int(n))
+			c.Logger.Debug("Starting pong timer", "dur", c.config.PongTimeout)
+			c.pongTimer = time.AfterFunc(c.config.PongTimeout, func() {
```
jaekwon: I think we need to stop the previous pongTimer, if it exists.

p2p/conn/connection.go
```diff
@@ -81,10 +83,15 @@ type MConnection struct {
...
-	flushTimer   *cmn.ThrottleTimer // flush writes as necessary but throttled.
-	pingTimer    *cmn.RepeatTimer   // send pings periodically
-	chStatsTimer *cmn.RepeatTimer   // update channel stats periodically
+	quit       chan struct{}
+	flushTimer *cmn.ThrottleTimer // flush writes as necessary but throttled.
+	pingTimer  *cmn.RepeatTimer   // send pings periodically
+
+	// close conn if pong is not received in pongTimeout
+	pongTimer     *time.Timer
+	pongTimeoutCh chan bool // true - timeout, false - peer sent pong
```
jaekwon: seems like it would better be,
true ~ timeout, close ~ peer sent pong.

p2p/conn/connection.go
```diff
@@ -315,7 +337,18 @@ FOR_LOOP:
...
+			c.pongTimer = time.AfterFunc(c.config.PongTimeout, func() {
+				c.pongTimeoutCh <- true
+			})
 			c.flush()
+		case timeout := <-c.pongTimeoutCh:
+			if timeout {
+				c.Logger.Debug("Pong timeout")
+				err = errors.New("pong timeout")
+			} else {
+				c.stopPongTimer()
```
jaekwon: weird that we stop it here... i don't think this is the right place to be stopping it.

jaekwon: also, see my first comment way up.

p2p/conn/connection.go
jaekwon: I think

p2p/conn/connection.go
melekes: 1) there is no (little) chance it exists since above I check for `if config.PongTimeout >= config.PingInterval {`
2) that would be bad. what's the reason to start it in the first place if later we stop it and replace with newer timer

p2p/conn/connection.go
melekes: weird?) we need to stop it here to prevent pongTimer from firing (`c.pongTimeoutCh <- true` above). 

```
pongTimerStart ............. (got pong -> stop pongTimer) xxx (pongTimerFires) ..... (pingTimer fires)
pongTimerStart ............. (pongTimer fires)
```

p2p/conn/connection.go
ebuchman: seems correct but does Go guarantee it ? can we check if the pongTimer is still active ? if so, then logically it's timed out (unless we have a bug and fail to stop it when we receive the pong)

p2p/conn/connection.go
ebuchman: hmm - won't that cause us to miss a lot of these ? if we're sending a peer lots of data, but it never responds to pings, we might never detect that.

p2p/conn/connection.go
ebuchman: but then you'd have to keep recreating the channel. why would that be better ?

p2p/conn/connection.go
ebuchman: This is done in https://github.com/tendermint/tendermint/pull/1209, but is it really ok if we're always ignoring those timeouts ?

PR #1107, opened 2018-01-14
better abuse handling in pex

ebuchman:
Instead of counting msgs per peer, only accept Addr msgs if we requested them, and only accept new Request msgs every ensurePeersPeriod / 3 (divided by 3 to be more than lenient with time keeping)

### DIFF
--- a/p2p/pex_reactor.go
+++ b/p2p/pex_reactor.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/pkg/errors"
 	wire "github.com/tendermint/go-wire"
 	cmn "github.com/tendermint/tmlibs/common"
 )
@@ -19,10 +20,6 @@ const (
 	defaultEnsurePeersPeriod = 30 * time.Second
 	minNumOutboundPeers      = 10
 	maxPexMessageSize        = 1048576 // 1MB
-
-	// maximum pex messages one peer can send to us during `msgCountByPeerFlushInterval`
-	defaultMaxMsgCountByPeer    = 1000
-	msgCountByPeerFlushInterval = 1 * time.Hour
 )
 
 // PEXReactor handles PEX (peer exchange) and ensures that an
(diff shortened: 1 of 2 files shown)


p2p/pex_reactor.go
```diff
@@ -147,27 +136,81 @@ func (r *PEXReactor) Receive(chID byte, src Peer, msgBytes []byte) {
...
+	if lastReceived.Equal(time.Time{}) {
+		// first time gets a free pass. then we start tracking the time
+		lastReceived := time.Now()
+		r.lastReceivedRequests.Set(id, lastReceived)
+		return nil
+	}
+
+	now := time.Now()
+	if now.Sub(lastReceived) < r.ensurePeersPeriod/3 {
+		return fmt.Errorf("Peer (%v) is sending too many PEX requests. Disconnecting", src.ID())
```
odeke-em: Nice!

p2p/pex_reactor.go
```diff
@@ -147,27 +136,81 @@ func (r *PEXReactor) Receive(chID byte, src Peer, msgBytes []byte) {
...
+		// initialize with empty time
+		lastReceived := time.Time{}
+		r.lastReceivedRequests.Set(id, lastReceived)
+		return nil
+	}
+
+	lastReceived := v.(time.Time)
+	if lastReceived.Equal(time.Time{}) {
+		// first time gets a free pass. then we start tracking the time
+		lastReceived := time.Now()
```
melekes: `=`

p2p/pex_reactor.go
```diff
@@ -291,24 +311,43 @@ func (r *PEXReactor) ensurePeers() {
...
+		r.Logger.Info("No addresses to dial nor connected peers. Falling back to seeds")
+		r.dialSeed()
 	}
 }
 
-func (r *PEXReactor) flushMsgCountByPeer() {
-	ticker := time.NewTicker(msgCountByPeerFlushInterval)
+func (r *PEXReactor) checkSeeds() error {
+	lSeeds := len(r.config.Seeds)
+	if lSeeds > 0 {
```
melekes: minor: why not exit early?

```
if lSeeds == 0 {
  return nil
}
```

PR #1120, opened 2018-01-19
fix and test config file

ebuchman:


cmd/tendermint/commands/root_test.go
```diff
@@ -22,89 +26,151 @@ const (
...
+		Use:               RootCmd.Use,
 		PersistentPreRunE: RootCmd.PersistentPreRunE,
+		Run:               func(cmd *cobra.Command, args []string) {},
 	}
-	r.AddCommand(cmds...)
-	wr := cli.PrepareBaseCmd(r, "TM", defaultRoot)
-	return wr
+	registerFlagsRootCmd(rootCmd)
+	var l string
+	rootCmd.PersistentFlags().String("log", l, "Log")
```
melekes: new flag?

ebuchman: just random thing for testing

Issue #1110, opened 2018-01-17
Missing binary downloads for tendermint 0.15


srmo:
Ahoi,

we still have no binary download links on the release page for "Leo".
Are they in the pipeline or did I miss a new download location?

Cheers!

srmo:
ping

adrianbrink:
https://github.com/tendermint/tendermint/releases/tag/v0.15.0

Done :-)

@srmo

srmo:
Great! Thanks :)

PR #1122, opened 2018-01-19
Bump version and fix dist_build.sh to new Makefile

adrianbrink:


ebuchman:
This, and more! https://github.com/tendermint/tendermint/pull/1123

adrianbrink:
@ebuchman We should get this and release binaries for 0.15.1, even if we cut 0.16.0 soon. 

People are asking for binaries: https://github.com/tendermint/tendermint/issues/1110

ebuchman:
Just cut the binaries. We don't need to release to do it.

PR #1128, opened 2018-01-21
seed crawler mode

ebuchman:
rebase and update https://github.com/tendermint/tendermint/pull/1067/


p2p/pex_reactor.go
```diff
@@ -357,6 +381,116 @@ func (r *PEXReactor) dialSeed() {
...
+		// Do not attempt to connect with peers we recently dialed
+		if now.Sub(pi.LastAttempt) < defaultCrawlPeerInterval {
+			continue
+		}
+		// Otherwise, attempt to connect with the known address
+		_, err := r.Switch.DialPeerWithAddress(pi.Addr, false)
+		if err != nil {
+			r.book.MarkAttempt(pi.Addr)
+			continue
+		}
```
ebuchman: we should just call RequestAddrs here and remove the below loop

PR #1133, opened 2018-01-21
StopPeerForError in blockchain and consensus

ebuchman:
Implement a few of the "punish peer" TODOs.

There's still lots more but these were low-hanging fruit.

Others may require some design/conversation.


ebuchman:
Tried to write a test but it's not easy. Worth discussion and a serious attempt to up coverage on the reactors at some later time ...

milosevic:
@ebuchman I will take a look at this, and try to write something down about overall approach on dealing with bad peers.

ebuchman:
Very helpful, thanks!

PR #1139, opened 2018-01-22
rename default user from ubuntu to vagrant

melekes:
vagrant docs says it must be vagrant user
https://www.vagrantup.com/docs/boxes/base.html#quot-vagrant-quot-user
If box does not do that, then it's not following vagrant requirements
and should not be used.


adrianbrink:
closed in favour of #1140

PR #1140, opened 2018-01-22
Fix Vagrantfile

adrianbrink:
If you get an error, please run `vagrant box update`.

According to Vagrant policies the user should be `vagrant` and if a box uses a different user it is a shitty box. It seems that the ubuntu box was upgraded in the 2018 version to use `vagrant`.

So if `vagrant box list` gives you something with 2017, you should run `vagrant box update` and it should work.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -21,7 +21,6 @@ Vagrant.configure("2") do |config|
 
     # install base requirements
     apt-get update
-    apt-get upgrade -y
     apt-get install -y --no-install-recommends wget curl jq \
         make shellcheck bsdmainutils psmisc
     apt-get install -y docker-ce golang-1.9-go
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -26,6 +26,7 @@ Vagrant.configure("2") do |config|
     apt-get install -y docker-ce golang-1.9-go
     apt-get install -y language-pack-en
 
+    # cleanup
     apt-get autoremove -y
 
     # needed for docker
@@ -44,7 +45,7 @@ Vagrant.configure("2") do |config|
     chown vagrant:vagrant /home/vagrant/.bash_profile
 
     # get all deps and tools, ready to install/test
-    source /home/vagrant/.bash_profile
-    cd /home/vagrant/go/src/github.com/tendermint/tendermint && make get_tools && make get_vendor_deps
+    su - vagrant  -c 'source /home/vagrant/.bash_profile'
+    su - vagrant -c 'cd /home/vagrant/go/src/github.com/tendermint/tendermint && make get_tools && make get_vendor_deps'
   SHELL
 end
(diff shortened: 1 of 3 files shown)


Vagrantfile
```diff
@@ -21,29 +21,31 @@ Vagrant.configure("2") do |config|
 
     # install base requirements
     apt-get update
+    apt-get upgrade -y
```
melekes: I think u are supposed to upgrade from outside (`vagrant update box`) rather then inside because upgrade might break some vagrant related setup things...

Vagrantfile
```diff
@@ -24,26 +24,27 @@ Vagrant.configure("2") do |config|
...
 
-    chown -R ubuntu:ubuntu /home/ubuntu/go
-    chown ubuntu:ubuntu /home/ubuntu/.bash_profile
+    chown -R vagrant:vagrant /home/vagrant/go
+    chown vagrant:vagrant /home/vagrant/.bash_profile
 
     # get all deps and tools, ready to install/test
-    su - ubuntu -c 'cd /home/ubuntu/go/src/github.com/tendermint/tendermint && make get_tools && make get_vendor_deps'
+    source /home/vagrant/.bash_profile
+    cd /home/vagrant/go/src/github.com/tendermint/tendermint && make get_tools && make get_vendor_deps
```
melekes: the script is executed from `root` account, so this may fail or result in conflicts later on.

Vagrantfile
```diff
@@ -24,26 +24,27 @@ Vagrant.configure("2") do |config|
...
+    ln -s /vagrant /home/vagrant/go/src/github.com/tendermint/tendermint
 
-    chown -R ubuntu:ubuntu /home/ubuntu/go
-    chown ubuntu:ubuntu /home/ubuntu/.bash_profile
+    chown -R vagrant:vagrant /home/vagrant/go
+    chown vagrant:vagrant /home/vagrant/.bash_profile
 
     # get all deps and tools, ready to install/test
-    su - ubuntu -c 'cd /home/ubuntu/go/src/github.com/tendermint/tendermint && make get_tools && make get_vendor_deps'
+    source /home/vagrant/.bash_profile
```
melekes: the script is executed from `root` account. so you're basically loading vagrant profile into root session. not sure we want that

Vagrantfile
```diff
@@ -24,26 +24,27 @@ Vagrant.configure("2") do |config|
     apt-get install -y --no-install-recommends wget curl jq \
         make shellcheck bsdmainutils psmisc
     apt-get install -y docker-ce golang-1.9-go
+    apt-get install -y language-pack-en
```
melekes: good call

Vagrantfile
```diff
@@ -24,26 +24,27 @@ Vagrant.configure("2") do |config|
...
+
+    apt-get autoremove -y
```
melekes:
```
Reading package lists... Done
Building dependency tree
Reading state information... Done
0 upgraded, 0 newly installed, 0 to remove and 18 not upgraded.
```

does it do any good for you?

ebuchman:
`vagrant box update` made it work for me. But now when I `vagrant ssh` I'm not in the tendermint directory. Can we fix that and add a note to the README about using vagrant and possibly running `vagrant box update` ?

adrianbrink:
Done

ebuchman:
Opened an issue for the test failure https://github.com/tendermint/tendermint/issues/1164 (obviously not related to this PR)

Issue #1119, opened 2018-01-19
[p2p] question about mconnection


melekes:
https://github.com/tendermint/tendermint/blob/develop/p2p/connection.go#L143-L152

```
	var channelsIdx = map[byte]*Channel{}
	var channels = []*Channel{}

	for _, desc := range chDescs {
		channel := newChannel(mconn, *desc)
		channelsIdx[channel.desc.ID] = channel
		channels = append(channels, channel)
	}
	mconn.channels = channels
	mconn.channelsIdx = channelsIdx
```

Why we have both a map and a slice here? Do we need order here? Otherwise why we can't just iterate over `channelsIdx`?

ebuchman:
Good point. I think we can probably just iterate over `channelsIdx` and don't actually need `channels`.

ebuchman:
Let's keep both since it's minor and might as well preserve order and have the lookup

PR #1141, opened 2018-01-22
remove channels slice and just use map

melekes:
as far as I know, we don't need order while sending a msg to channels
nor when we report a status.

Refs #1119

### DIFF
--- a/p2p/conn/connection.go
+++ b/p2p/conn/connection.go
@@ -501,7 +501,7 @@ type ConnectionStatus struct {
 	Duration    time.Duration
 	SendMonitor flow.Status
 	RecvMonitor flow.Status
-	Channels    []ChannelStatus
+	Channels    map[byte]ChannelStatus
 }
 
 type ChannelStatus struct {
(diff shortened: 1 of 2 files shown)


ebuchman:
Only concern might be performance of iterating over a map. Can we get a quick benchmark of the difference for say 8, 16, 32, etc. items.

Currently we have about 8 channels I believe. Maybe it will double. I doubt it would be as high as 100 anytime soon but might be good to know

melekes:
```
BenchmarkIterateOverChannelsMap8-2              20000000               114 ns/op
BenchmarkIterateOverChannelsMap16-2              5000000               294 ns/op
BenchmarkIterateOverChannelsMap32-2              3000000               559 ns/op
BenchmarkIterateOverChannelsSlice8-2            200000000                7.71 ns/op
BenchmarkIterateOverChannelsSlice16-2           100000000               14.1 ns/op
BenchmarkIterateOverChannelsSlice32-2            3000000               560 ns/op
```

> Only concern might be performance of iterating over a map

why is this a concern at all if we use map when we send something to a channel???
```
	// Send message to channel.
	channel, ok := c.channels[chID]
```

i.e. we only use slice when reporting status and updating stats

ebuchman:
> why is this a concern at all if we use map when we send something to a channel???

It's just a lookup, we're not iterating over the whole map.

Looks like small slices are ~20x faster to iterate through

p2p/conn/connection.go
```diff
@@ -524,14 +518,16 @@ func (c *MConnection) Status() ConnectionStatus {
 	status.SendMonitor = c.sendMonitor.Status()
 	status.RecvMonitor = c.recvMonitor.Status()
 	status.Channels = make([]ChannelStatus, len(c.channels))
-	for i, channel := range c.channels {
+	i := 0
+	for _, channel := range c.channels {
```
ebuchman: it does mean the channels will be in a different order here every time over the RPC. Could be weird.

melekes: I would change RPC to return a map instead. I.e. if somebody relies on order here, then he/she is obviously doing wrong thing

melekes: Like https://github.com/tendermint/tendermint/pull/1141/commits/da59c46a5a0589803aaaebdcb4624788cba430ff

ebuchman:
Why'd you close it ?

> I would change RPC to return a map instead. I.e. if somebody relies on order here, then he/she is obviously doing wrong thing

I don't disagree, but Is this normal behaviour for an API ? Seems inevitable in Go APIs due to explitly randomized iteration but just curious ...

PR #1142, opened 2018-01-23
Add support for ValidBlock mechanism for the simplest case

milosevic:
With this PR, the idea is to only cover the simplest scenario where having ValidBlock will help with the termination issue. The goal is having the following property assured: during synchronous period (after GST), in a round with a correct proposer, if some correct process locks a block v in round r, at the end of the round r, all correct processes will have lockedBlock = v or lockedBlock = nil, and validBlock = v. This ensures that in the following round, a (correct) proposer will be able to only propose v, and all correct processes will Prevote for v, and hopefully later commit a block. 

Please take a look @ebuchman @melekes. I will add tests after.

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -477,6 +477,9 @@ func (cs *ConsensusState) updateToState(state sm.State) {
 	cs.LockedRound = 0
 	cs.LockedBlock = nil
 	cs.LockedBlockParts = nil
+	cs.ValidRound = 0
+	cs.ValidBlock = nil
+	cs.ValidBlockParts = nil
 	cs.Votes = cstypes.NewHeightVoteSet(state.ChainID, height, validators)
 	cs.CommitRound = -1
 	cs.LastCommit = lastPrecommits
@@ -798,6 +801,9 @@ func (cs *ConsensusState) defaultDecideProposal(height int64, round int) {
 	if cs.LockedBlock != nil {
 		// If we're locked onto a block, just choose that.
 		block, blockParts = cs.LockedBlock, cs.LockedBlockParts
+	} else if cs.ValidBlock != nil {
+		// If there is valid block, choose that.
+		block, blockParts = cs.ValidBlock, cs.ValidBlockParts
 	} else {
 		// Create a new proposal block from state/txs from the mempool.
 		block, blockParts = cs.createProposalBlock()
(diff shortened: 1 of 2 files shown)


consensus/state.go
```diff
@@ -1388,6 +1394,17 @@ func (cs *ConsensusState) addVote(vote *types.Vote, peerID p2p.ID) (added bool,
...
 						cs.eventBus.PublishEventUnlock(cs.RoundStateEvent())
 					}
+					if ok && !blockID.IsZero() && !cs.ValidBlock.HashesTo(blockID.Hash) && vote.Round > cs.ValidRound {
+						// update valid value
+						if cs.ProposalBlock.HashesTo(blockID.Hash) {
+							cs.ValidRound = vote.Round
+							cs.ValidBlock = cs.ProposalBlock
+							cs.ValidBlockParts = cs.ProposalBlockParts
+						}
+						//TODO: We might want to update ValidBlock also in case we don't have that block yet,
```
ebuchman: Oh good point. This might be a bit more complicated.

Well, I guess we just need to update the AddProposalBlockPart logic to check if we have ValidRound > 0 but no ValidBlock, then we know to set it with the complete ProposalBlock

milosevic: It's bit more tricky as ValidBlock might be (and probably will be) different than ProposalBlock. So we would need to extend gossiping to support sending BlockParts for two different blocks. That's why I am proposing addressing only the simplest scenario. We should still be correct with only this change, but the worst case termination time is O(N), where N is number of validators, instead of O(1) if we would be able to support ValidBlock mechanism fully. So it's performance issue, and we can address it after.

consensus/state.go
```diff
@@ -1388,6 +1394,17 @@ func (cs *ConsensusState) addVote(vote *types.Vote, peerID p2p.ID) (added bool,
 						cs.LockedBlockParts = nil
 						cs.eventBus.PublishEventUnlock(cs.RoundStateEvent())
 					}
+					if ok && !blockID.IsZero() && !cs.ValidBlock.HashesTo(blockID.Hash) && vote.Round > cs.ValidRound {
```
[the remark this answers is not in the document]
milosevic: You are right. We don't care about cs.LockedBlock here. I have moved ValidBlock update to a separate block.

melekes:
`TestReactorVotingPowerChange` failed with timeout.

ebuchman:
Rebased on develop

ebuchman:
```
consensus/wal_generator.go:14:2: cannot find package "github.com/tendermint/abci/example/kvstore" in any of:
	/go/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/abci/example/kvstore (vendor tree)
	/usr/local/go/src/github.com/tendermint/abci/example/kvstore (from $GOROOT)
	/go/src/github.com/tendermint/abci/example/kvstore (from $GOPATH)
```

Needs a glide update

xla:
Rebased on latest develop.

PR #1178, opened 2018-01-31
improve error message

### LINKED ISSUE #1158: Difficult to interpret error message caused by undetermined genesis block error
Bug report

During the dry run of setting up external genesis validators, we were seeing errors that eminated from this code block
https://github.com/tendermint/tendermint/blob/master/types/vote_set.go#L172-L176

There isn't an obvious way to track that back to our genesis file and see what went wrong.

zramsay:
closes #1158

### DIFF
--- a/types/vote_set.go
+++ b/types/vote_set.go
@@ -172,7 +172,7 @@ func (voteSet *VoteSet) addVote(vote *Vote) (added bool, err error) {
 	// Ensure that the signer has the right address
 	if !bytes.Equal(valAddr, lookupAddr) {
 		return false, errors.Wrapf(ErrVoteInvalidValidatorAddress,
-			"vote.ValidatorAddress (%X) does not match address (%X) for vote.ValidatorIndex (%d)",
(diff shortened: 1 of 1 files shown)


types/vote_set.go
```diff
@@ -172,7 +172,7 @@ func (voteSet *VoteSet) addVote(vote *Vote) (added bool, err error) {
 	// Ensure that the signer has the right address
 	if !bytes.Equal(valAddr, lookupAddr) {
 		return false, errors.Wrapf(ErrVoteInvalidValidatorAddress,
-			"vote.ValidatorAddress (%X) does not match address (%X) for vote.ValidatorIndex (%d)",
+			"vote.ValidatorAddress (%X) does not match address (%X) for vote.ValidatorIndex (%d)\nEnsure the genesis.json is correct across all validators.",
```
melekes: how about we use "genesis file" instead of "genesis.json" so we won't need to change msg if we change file format

zramsay: fixed

PR #1180, opened 2018-02-01
Add BFT time spec

milosevic:
@ebuchman @melekes @ethanfrey Please check if this makes sense so I can proceed to implementation.

### DIFF
--- a/docs/specification/new-spec/bft-time.md
+++ b/docs/specification/new-spec/bft-time.md
@@ -1,0 +1,42 @@
+# BFT time in Tendermint 
+
+Tendermint provides a deterministic, Byzantine fault-tolerant, source of time. 
+Time in Tendermint is defined with the Time field of the block header. 
+
+It satisfies the following properties:
+
+- Time Monotonicity: Time is monotonically  increasing, i.e., given 
+a header H1 for height h1 and a header H2 for height `h2 = h1 + 1`, `H1.Time < H2.Time`. 
+- Time Validity: Given a set of Commit votes that forms the `block.LastCommit` field, a range of 
+valid values for the Time field of the block header is defined only by  
+Precommit messages (from the LastCommit field) sent by correct processes, i.e., 
+a faulty process cannot arbitrarily increase the Time value.  
+
+In the context of Tendermint, time is of type int64 and denotes UNIX time in milliseconds, i.e., 
+corresponds to the number of milliseconds since January 1, 1970. Before defining rules that need to be enforced by the 
+Tendermint consensus protocol, so the properties above holds, we introduce the following definition:
+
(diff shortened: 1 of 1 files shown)


docs/specification/new-spec/bft-time.md
```diff
@@ -0,0 +1,42 @@
...
+corresponds to the number of milliseconds since January 1, 1970. Before defining rules that need to be enforced by the 
+Tendermint consensus protocol, so the properties above holds, we introduce the following definition:
+
+- median of a set of `Vote` messages is equal to the median of `Vote.Time` fields of the corresponding `Vote` messages
+
+We ensure Time Monotonicity and Time Validity properties by the following rules: 
+  
+- let rs denotes `RoundState` (consensus internal state) of some process. Then 
+`rs.ProposalBlock.Header.Time == median(rs.LastCommit) &&
+rs.Proposal.Timestamp == rs.ProposalBlock.Header.Time`.
```
ebuchman: Why use the median rather than time.Now() here (for the Proposal itself, not for the block header time)?

docs/specification/new-spec/bft-time.md
```diff
@@ -0,0 +1,42 @@
...
+- Furthermore, when creating the `vote` message, the following rules for determining `vote.Time` field should hold: 
+
+    - if `rs.Proposal` is defined then 
+    `vote.Time = max(rs.Proposal.Timestamp + 1, time.Now())`, where `time.Now()` 
+    denotes local Unix time in milliseconds.  
+    
+    - if `rs.Proposal` is not defined and `rs.Votes` contains +2/3 of the corresponding vote messages (votes for the 
+    current height and round, and with the corresponding type (`Prevote` or `Precommit`)), then 
+    
+    `vote.Time = max(median(getVotes(rs.Votes, vote.Height, vote.Round, vote.Type)), time.Now())`,
```
ebuchman: should we include a +1 in here?

milosevic: It's not needed in this case as median is already >= Proposal.Time.

docs/specification/new-spec/bft-time.md
milosevic: rs.Proposal and rs.ProposalBlock are referring to the same thing, so the time should be the same.

milosevic:
@ebuchman  I figured out that we need to add a case for voting for nil in case timeout expires and we haven't received Proposal message. Then we don't have a reference point. But time in Vote messages with nil value shouldn't affect time properties as we only care about not-nil commit messages. Probably for nil based Votes we can just put time.Now(). And maybe we need to check what we do in case of empty blocks.

Issue #1177, opened 2018-01-30
Semantics of "currate too low" error messages and what to do


srmo:
We sometimes see "curRate too low" error messages in the logs, especially during resyncing of a node.
What's the semantic of that? Why is it an error? What to do about it and how can we reason about the state of such a node or the chain at all?
This goes hand in hand with Connection closed for peer errors where it is not clear if such connections have been recovered in a later stage.

Edit: we are still on 0.12 but will see what happens when we manage to migrate to current 0.15

From riot:

> could you create an issue on Github if you can't find any info on curRate
> we should probably output curRate, not only "curRate is low"
> 
> I think it means connection is slow between this node and some peer
> "rate" comes from https://github.com/tendermint/tmlibs/tree/master/flowrate
> curRate - speed with which we get blocks from peer
> msg is badly written anyway

ebuchman:
Basically we set a minimum recv rate in the fast sync to ensure we're receiving blocks from a peer fast enough. It's currently hardcoded to 10kB/s. Perhaps that's too high. If a peer is not sending us data at at least that rate, we consider them to have timedout and we disconnect. 

We should make this more robust and improve the message.

Thanks.

melekes:
Let's take a really old DSL modem - with upload speed 128 Kbps ~ 15 KB/s. https://www.att.net/speedtiers

45ms or less for regional round trips within North America
90ms or less for transatlantic round trips between London and New York.

so (if we assume the above stats from att was measured in US, which seems to be the case), the worst should be transatlantic upload approximately at half speed - 7.5 KB/s.

srmo:
That’s what bothers me. We might see this in local vm network setups or internet/cloud  setups where it is unlikely to hit such low speeds. So there’s an urge to think there’s more going an besides a slow connection. 

The clarification mentioned by ebuchman seems to go in the right direction. A quick bulletpoint list on what an ordinary user could do to analyse/find the root cause would be very helpful (like: measure the connection between peers using xyz with protocol TCP and ...)

melekes:
I changed the message to show current rate (as well as min), so you will be able to see it.

melekes:
To measure conn speed use [iperf](https://github.com/esnet/iperf) https://www.cyberciti.biz/faq/how-to-test-the-network-speedthroughput-between-two-linux-servers/ or netcat (nc).

srmo:
Ok, thank you for the pointer. 
We’ll probably do some testruna

[...]

PR #1194, opened 2018-02-07
improve "curRate too low" message

melekes:
Refs #1177

Note on labels:
KB - 1024
kB - 1000
https://ux.stackexchange.com/questions/13815/files-size-units-kib-vs-kb-vs-kb

### DIFF
--- a/blockchain/pool.go
+++ b/blockchain/pool.go
@@ -127,7 +127,7 @@ func (pool *BlockPool) removeTimedoutPeers() {
 		if !peer.didTimeout && peer.numPending > 0 {
 			curRate := peer.recvMonitor.Status().CurRate
 			// XXX remove curRate != 0
-			if curRate != 0 && curRate < minRecvRate {
+			if curRate < minRecvRate {
 				pool.sendTimeout(peer.id)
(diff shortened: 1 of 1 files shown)


blockchain/pool.go
```diff
@@ -122,7 +126,10 @@ func (pool *BlockPool) removeTimedoutPeers() {
 			// XXX remove curRate != 0
```
zramsay: can this be easily done?

melekes: removed it

blockchain/pool.go
```diff
@@ -120,9 +127,12 @@ func (pool *BlockPool) removeTimedoutPeers() {
 		if !peer.didTimeout && peer.numPending > 0 {
 			curRate := peer.recvMonitor.Status().CurRate
 			// XXX remove curRate != 0
-			if curRate != 0 && curRate < minRecvRate {
```
ebuchman: im not sure why it was here in the beginning. probably something about not removing right away if the curRate is 0 and we havent received anything yet to give it time to ramp up.

Also the fast sync bash test failed here. I can't find this error message in the logs but maybe it's related? could just be non-determinism in this test :(

PR #1198, opened 2018-02-08
SDK: AppOptions -> AppState

rigelrozanski:


types/genesis.go
```diff
@@ -28,7 +28,7 @@ type GenesisDoc struct {
 	ConsensusParams *ConsensusParams   `json:"consensus_params,omitempty"`
 	Validators      []GenesisValidator `json:"validators"`
 	AppHash         cmn.HexBytes       `json:"app_hash"`
-	AppOptions      interface{}        `json:"app_options,omitempty"`
+	AppStateJSON    json.RawMessage    `json:"app_options,omitempty"`
```
ebuchman: We should be consistent with the Go name and JSON name

rigelrozanski:
@ebuchman I forced pushed to (json desc now matches)

types/genesis.go
rigelrozanski: k, Also just renamed to `AppState` unless there is a reason to include JSON in the name? Also what's with `omitempty`? I left it in there

melekes:
What tutorials/docs need to be updated?

ebuchman:
This is a breaking change to the genesis file, though it's not tested. 

We should
1) add a test to types/genesis_test.go to test that we get the right raw JSON for the app_state
2) decide if we want to include it in 0.16, and need to include in changelog

greg-szabo:
Also a note from the PM: if we can avoid it, let's not move the goalpost too much. If we keep adding new issues to the sprint, we'll never finish it.
There is no description to the PR.
There is no issue associated with it.
I literally can't estimate anything about this PR. It's out of the blue.

Are we delivering a broken 0.16 codebase if we omit this PR from it and move it to the next release?
If the answer is no, let's move this to "next release".

ebuchman:
This is coming from recent work on the SDK to just improve the name of the field to be what it actually is :). 

No we don't need it for this release

melekes:
Todo:

- [ ] update changelog
- [ ] check/update documentation

PR #1203, opened 2018-02-09
types/priv_validator package

### LINKED ISSUE #673: PrivValidator Structure

ebuchman:
Replaces #1044 
closes #673 

Note this is just the new code, none of the integration (old priv val is still there).

I didn't feel too good about the integration last time, and now its riddled with conflicts, so maybe we can figure out a simple way to make it work starting a fresh. At least with this we can merge it without affecting anything.

### DIFF
--- a/cmd/priv_val_server/main.go
+++ b/cmd/priv_val_server/main.go
@@ -1,0 +1,47 @@
+package main
+
+import (
+	"flag"
+	"os"
+
+	cmn "github.com/tendermint/tmlibs/common"
+	"github.com/tendermint/tmlibs/log"
+
+	priv_val "github.com/tendermint/tendermint/types/priv_validator"
+)
+
+func main() {
+	var (
+		chainID     = flag.String("chain-id", "mychain", "chain id")
+		listenAddr  = flag.String("laddr", ":46659", "Validator listen address (0.0.0.0:0 means any interface, any port")
+		maxConn     = flag.Int("clients", 3, "maximum of concurrent connections")
+		privValPath = flag.String("priv", "", "priv val file path")
+
+		logger = log.NewTMLogger(log.NewSyncWriter(os.Stdout)).With("module", "priv_val")
+	)
+	flag.Parse()
+
+	logger.Info(
+		"Starting private validator",
+		"chainID", *chainID,
+		"listenAddr", *listenAddr,
+		"maxConn", *maxConn,
+		"privPath", *privValPath,
+	)
+
+	privVal := priv_val.LoadPrivValidatorJSON(*privValPath)
+
+	pvss := priv_val.NewPrivValidatorSocketServer(
+		logger,
+		*chainID,
+		*listenAddr,
+		*maxConn,
+		privVal,
+		nil,
+	)
+	pvss.Start()
+
+	cmn.TrapSignal(func() {
+		pvss.Stop()
+	})
+}
--- a/docs/architecture/adr-008-priv-validator.md
+++ b/docs/architecture/adr-008-priv-validator.md
@@ -8,7 +8,7 @@ For instance, see https://github.com/tendermint/tendermint/issues/673
 
 The goal is to have a clean PrivValidator interface like:
 
-``
+```
 type PrivValidator interface {
 	Address() data.Bytes
 	PubKey() crypto.PubKey
@@ -60,6 +60,7 @@ type ValidatorID struct {
 	Address data.Bytes    `json:"address"`
 	PubKey  crypto.PubKey `json:"pub_key"`
 }
+```
 
 ### LastSignedInfo
 
--- a/docs/architecture/adr-008-priv-validator.md
+++ b/docs/architecture/adr-008-priv-validator.md
@@ -1,0 +1,118 @@
+# ADR 008: PrivValidator
+
+## Context
+
+The current PrivValidator is monolithic and isn't easily reuseable by alternative signers.
+
+For instance, see https://github.com/tendermint/tendermint/issues/673
+
+The goal is to have a clean PrivValidator interface like:
+
+``
+type PrivValidator interface {
+	Address() data.Bytes
+	PubKey() crypto.PubKey
+
+	SignVote(chainID string, vote *types.Vote) error
+	SignProposal(chainID string, proposal *types.Proposal) error
+	SignHeartbeat(chainID string, heartbeat *types.Heartbeat) error
+}
+```
+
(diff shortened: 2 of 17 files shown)


types/priv_validator/json.go
```diff
@@ -0,0 +1,180 @@
...
+	// enable persistence
+	pvj.filePath = filePath
+	return &pvj
+}
+
+// LoadOrGenPrivValidatorJSON loads a PrivValidatorJSON from the given filePath
+// or else generates a new one and saves it to the filePath.
+func LoadOrGenPrivValidatorJSON(filePath string) *PrivValidatorJSON {
+	var pvj *PrivValidatorJSON
+	if _, err := os.Stat(filePath); err == nil {
```
melekes: `cmn.FileExists(filePath)` ?

melekes:
```
# github.com/tendermint/tendermint/types/priv_validator
/tmp/go-build532587797/github.com/tendermint/tendermint/types/priv_validator/_test/_obj_test/unencrypted.go:16:6: undefined: "github.com/tendermint/tendermint/types".ValidatorID
```

ebuchman:
Merged in the new priv validator socket work, but please see review from that PR: https://github.com/tendermint/tendermint/pull/1204#pullrequestreview-97652015

types/priv_validator/json.go
```diff
@@ -0,0 +1,195 @@
...
+	return fmt.Sprintf("PrivValidator{%v %v}", addr, pvj.PrivValidatorUnencrypted.String())
+}
+
+func (pvj *PrivValidatorJSON) Save() {
+	pvj.save()
+}
+
+func (pvj *PrivValidatorJSON) save() {
+	if pvj.filePath == "" {
+		cmn.PanicSanity("Cannot save PrivValidator: filePath not set")
```
melekes: `Panic*` methods are deprecated. Use `panic` instead.

types/priv_validator/sign_info.go
```diff
@@ -0,0 +1,238 @@
...
+//-------------------------------------
+
+// LastSignedInfo contains information about the latest
+// data signed by a validator to help prevent double signing.
+type LastSignedInfo struct {
+	Height    int64            `json:"height"`
+	Round     int              `json:"round"`
+	Step      int8             `json:"step"`
+	Signature crypto.Signature `json:"signature,omitempty"` // so we dont lose signatures
+	SignBytes cmn.HexBytes     `json:"signbytes,omitempty"` // so we dont lose signatures
```
melekes: `sign_bytes`?

docs/architecture/adr-008-priv-validator.md
```diff
@@ -0,0 +1,119 @@
...
+
+LastSignedInfo tracks the last thing we signed:
+
+```
+type LastSignedInfo struct {
+	Height    int64            `json:"height"`
+	Round     int              `json:"round"`
+	Step      int8             `json:"step"`
+	Signature crypto.Signature `json:"signature,omitempty"` // so we dont lose signatures
+	SignBytes data.Bytes       `json:"signbytes,omitempty"` // so we dont lose signatures
```
melekes: `sign_bytes`?

xla:
@ebuchman What's the next step here? Should we start with the actual integration by getting rid of the old priv val, flesh out `cmd/priv_val_server` according to ADR 008, or prepare to merge as is?

Happy to follow-up and tackle any of these.

types/priv_validator/socket.go
```diff
@@ -0,0 +1,467 @@
...
+
+func (pvss *PrivValidatorSocketServer) acceptConnections() {
+	for {
+		conn, err := pvss.listener.Accept()
+		if err != nil {
+			if !pvss.IsRunning() {
+				return // Ignore error from listener closing.
+			}
+			pvss.Logger.Error(
+				"accpetConnections",
```
melekes: `accept`

node/node.go
```diff
@@ -171,6 +173,24 @@ func NewNode(config *cfg.Config,
...
+		var (
+			privKey = crypto.GenPrivKeyEd25519()
+			pvsc    = priv_val.NewSocketClient(
+				logger.With("module", "priv_val"),
+				config.PrivValidatorAddr,
+				&privKey,
+			)
+		)
+
+		if err := pvsc.Start(); err != nil {
```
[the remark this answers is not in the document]
xla: If we `defer pvsc.Stop()` here it would be killed as soon as we exit the function scope. If the client is used it's vital to the node and is kind of bound to the lifecycle of the node process itself, maybe I missed a place where we have teardown when the node shuts down, that would be the place to stop it.

cmd/priv_val_server/main.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/tendermint/tmlibs/log"
+
+	priv_val "github.com/tendermint/tendermint/types/priv_validator"
+)
+
+func main() {
+	var (
+		chainID     = flag.String("chain-id", "mychain", "chain id")
+		listenAddr  = flag.String("laddr", ":46659", "Validator listen address (0.0.0.0:0 means any interface, any port")
+		maxConn     = flag.Int("clients", 3, "maximum of concurrent connections")
```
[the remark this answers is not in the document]
xla: Originally it only accepted a single connection at a time, we had a discussion around it and [decided](https://github.com/tendermint/tendermint/pull/1060#issuecomment-358362939) to go with 3 for now. Usually the relation should be one-to-one, between tendermint process and external signer, as far as I know at least.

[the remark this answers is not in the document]
melekes: Got it. Thanks

node/node.go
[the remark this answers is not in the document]
melekes: https://github.com/tendermint/tendermint/blob/master/node/node.go#L406

ebuchman:
Merging for the sake of progress, will open new issue for remaining points. Thanks @xla !

types/priv_validator/sign_info.go
xla: This would be inconsistent to other occurences of `signbytes`: https://github.com/tendermint/tendermint/blob/master/types/priv_validator.go#L58

PR #1204, opened 2018-02-09
Feature/priv val sockets

ebuchman:
Rebased https://github.com/tendermint/tendermint/pull/1060/ on https://github.com/tendermint/tendermint/pull/1203.

Note the integrations for this in node/node.go are commented out (until types/priv_validator becomes active).

### DIFF
--- a/cmd/priv_val_server/main.go
+++ b/cmd/priv_val_server/main.go
@@ -14,7 +14,7 @@ func main() {
 	var (
 		chainID     = flag.String("chain-id", "mychain", "chain id")
 		listenAddr  = flag.String("laddr", ":46659", "Validator listen address (0.0.0.0:0 means any interface, any port")
-		maxClients  = flag.Int("clients", 3, "number of concurrently connected clients")
+		maxConn     = flag.Int("clients", 3, "maximum of concurrent connections")
 		privValPath = flag.String("priv", "", "priv val file path")
 
 		logger = log.NewTMLogger(log.NewSyncWriter(os.Stdout)).With("module", "priv_val")
@@ -25,7 +25,7 @@ func main() {
 		"Starting private validator",
 		"chainID", *chainID,
 		"listenAddr", *listenAddr,
-		"maxClients", *maxClients,
+		"maxConn", *maxConn,
 		"privPath", *privValPath,
 	)
 
@@ -35,7 +35,7 @@ func main() {
 		logger,
 		*chainID,
 		*listenAddr,
-		*maxClients,
+		*maxConn,
 		privVal,
 		nil,
 	)
--- a/cmd/priv_val_server/main.go
+++ b/cmd/priv_val_server/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"flag"
-	"fmt"
 	"os"
 
 	cmn "github.com/tendermint/tmlibs/common"
@@ -11,20 +10,27 @@ import (
 	priv_val "github.com/tendermint/tendermint/types/priv_validator"
 )
 
-var chainID = flag.String("chain-id", "mychain", "chain id")
-var privValPath = flag.String("priv", "", "priv val file path")
-
 func main() {
+	var (
+		chainID     = flag.String("chain-id", "mychain", "chain id")
+		privValPath = flag.String("priv", "", "priv val file path")
+		socketAddr  = flag.String("socket.addr", ":46659", "socket bind addr")
+
+		logger = log.NewTMLogger(log.NewSyncWriter(os.Stdout)).With("module", "priv_val")
+	)
 	flag.Parse()
 
-	logger := log.NewTMLogger(log.NewSyncWriter(os.Stdout)).With("module", "priv_val")
-	socketAddr := "localhost:46659"
-	fmt.Println(*chainID)
-	fmt.Println(*privValPath)
+	logger.Info("args", "chainID", *chainID, "privPath", *privValPath)
+
 	privVal := priv_val.LoadPrivValidatorJSON(*privValPath)
 
-	pvss := priv_val.NewPrivValidatorSocketServer(logger, socketAddr, *chainID, privVal)
-	pvss.Start()
+	pvss := priv_val.NewPrivValidatorSocketServer(
+		logger,
+		*socketAddr,
+		*chainID,
+		privVal,
+	)
+	// pvss.Start()
 
 	cmn.TrapSignal(func() {
 		pvss.Stop()
--- a/cmd/priv_val_server/main.go
+++ b/cmd/priv_val_server/main.go
@@ -1,0 +1,32 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+
+	cmn "github.com/tendermint/tmlibs/common"
+	"github.com/tendermint/tmlibs/log"
+
+	priv_val "github.com/tendermint/tendermint/types/priv_validator"
+)
+
+var chainID = flag.String("chain-id", "mychain", "chain id")
+var privValPath = flag.String("priv", "", "priv val file path")
+
+func main() {
+	flag.Parse()
+
+	logger := log.NewTMLogger(log.NewSyncWriter(os.Stdout)).With("module", "priv_val")
+	socketAddr := "localhost:46659"
+	fmt.Println(*chainID)
+	fmt.Println(*privValPath)
+	privVal := priv_val.LoadPrivValidatorJSON(*privValPath)
+
+	pvss := priv_val.NewPrivValidatorSocketServer(logger, socketAddr, *chainID, privVal)
+	pvss.Start()
+
+	cmn.TrapSignal(func() {
+		pvss.Stop()
+	})
+}
--- a/cmd/priv_val_server/main.go
+++ b/cmd/priv_val_server/main.go
@@ -13,6 +13,7 @@ import (
 func main() {
 	var (
 		chainID     = flag.String("chain-id", "mychain", "chain id")
+		numClients  = flag.Int("clients", 1, "number of concurrently connected clients")
 		privValPath = flag.String("priv", "", "priv val file path")
 		socketAddr  = flag.String("socket.addr", ":46659", "socket bind addr")
 
@@ -20,17 +21,19 @@ func main() {
 	)
 	flag.Parse()
 
-	logger.Info("args", "chainID", *chainID, "privPath", *privValPath)
+	logger.Info("Reading args privValidatorSocketServer", "chainID", *chainID, "privPath", *privValPath)
 
 	privVal := priv_val.LoadPrivValidatorJSON(*privValPath)
 
 	pvss := priv_val.NewPrivValidatorSocketServer(
 		logger,
-		*socketAddr,
 		*chainID,
+		*socketAddr,
+		*numClients,
 		privVal,
+		nil,
 	)
-	// pvss.Start()
+	pvss.Start()
 
 	cmn.TrapSignal(func() {
 		pvss.Stop()
--- a/cmd/priv_val_server/main.go
+++ b/cmd/priv_val_server/main.go
@@ -13,23 +13,29 @@ import (
 func main() {
 	var (
 		chainID     = flag.String("chain-id", "mychain", "chain id")
-		numClients  = flag.Int("clients", 1, "number of concurrently connected clients")
+		listenAddr  = flag.String("laddr", ":46659", "Validator listen address (0.0.0.0:0 means any interface, any port")
+		maxClients  = flag.Int("clients", 3, "number of concurrently connected clients")
 		privValPath = flag.String("priv", "", "priv val file path")
-		socketAddr  = flag.String("socket.addr", ":46659", "socket bind addr")
 
 		logger = log.NewTMLogger(log.NewSyncWriter(os.Stdout)).With("module", "priv_val")
 	)
 	flag.Parse()
 
-	logger.Info("Reading args privValidatorSocketServer", "chainID", *chainID, "privPath", *privValPath)
+	logger.Info(
+		"Starting private validator",
+		"chainID", *chainID,
+		"listenAddr", *listenAddr,
+		"maxClients", *maxClients,
+		"privPath", *privValPath,
+	)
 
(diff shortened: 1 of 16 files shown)


ebuchman:
https://github.com/tendermint/tendermint/pull/1060/#issuecomment-364201241: 

> When working with the protocol between client and server I was dearly missing some form of communicating errors, but I don't know how settle we are on the current protocol.

We can modify the protocol here to support error handling. Want to propose something?

ebuchman:
Also, looks like Tendermint is responsible for dialing the signer process here. I thought we wanted the signer to dial Tendermint (https://github.com/tendermint/tendermint/pull/1060#issuecomment-358406383). Should we make the SocketServer implement PrivKey and give Tendermint that or is there some better reason this should stay as is?

cmd/priv_val_server/main.go
```diff
@@ -0,0 +1,41 @@
...
+	cmn "github.com/tendermint/tmlibs/common"
+	"github.com/tendermint/tmlibs/log"
+
+	priv_val "github.com/tendermint/tendermint/types/priv_validator"
+)
+
+func main() {
+	var (
+		chainID     = flag.String("chain-id", "mychain", "chain id")
+		numClients  = flag.Int("clients", 1, "number of concurrently connected clients")
```
melekes: how am I supposed to know the number of clients beforehand?

cmd/priv_val_server/main.go
```diff
@@ -0,0 +1,41 @@
...
+		privValPath = flag.String("priv", "", "priv val file path")
+		socketAddr  = flag.String("socket.addr", ":46659", "socket bind addr")
+
+		logger = log.NewTMLogger(log.NewSyncWriter(os.Stdout)).With("module", "priv_val")
+	)
+	flag.Parse()
+
+	logger.Info("Reading args privValidatorSocketServer", "chainID", *chainID, "privPath", *privValPath)
```
melekes: Maybe change to `fmt.Sprintf("Starting private validator socket server on %s", listenAddr), "chainID", *chainID, "privValPath", *privValPath,` ?

cmd/priv_val_server/main.go
```diff
@@ -0,0 +1,41 @@
...
+
+	priv_val "github.com/tendermint/tendermint/types/priv_validator"
+)
+
+func main() {
+	var (
+		chainID     = flag.String("chain-id", "mychain", "chain id")
+		numClients  = flag.Int("clients", 1, "number of concurrently connected clients")
+		privValPath = flag.String("priv", "", "priv val file path")
+		socketAddr  = flag.String("socket.addr", ":46659", "socket bind addr")
```
melekes: Let's keep a single standard (`listenAddr = flag.String("laddr"`)

```
	cmd.Flags().String("p2p.laddr", config.P2P.ListenAddress, "Node listen address. (0.0.0.0:0 means any interface, any port)")
```

cmd/tendermint/commands/run_node.go
```diff
@@ -14,6 +14,9 @@ func AddNodeFlags(cmd *cobra.Command) {
 	// bind flags
 	cmd.Flags().String("moniker", config.Moniker, "Node Name")
 
+	// priv val flags
+	cmd.Flags().String("priv_validator_addr", config.PrivValidatorAddr, "Socket address for PrivValidator")
```
melekes: `private validator`, not `PrivValidator`

types/priv_validator/socket.go
```diff
@@ -34,35 +37,65 @@ const (
...
 
+// OnStart implements cmn.Service.
 func (pvsc *PrivValidatorSocketClient) OnStart() error {
 	if err := pvsc.BaseService.OnStart(); err != nil {
 		return err
 	}
 
-	var err error
-	var conn net.Conn
+	var (
```
melekes: can we extract this into a separate `connect` function?

types/priv_validator/socket.go
```diff
@@ -34,35 +37,65 @@ const (
...
+		conn net.Conn
+	)
+
 RETRY_LOOP:
 	for {
 		conn, err = cmn.Connect(pvsc.SocketAddress)
 		if err != nil {
-			pvsc.Logger.Error(fmt.Sprintf("PrivValidatorSocket failed to connect to %v.  Retrying...", pvsc.SocketAddress))
+			pvsc.Logger.Error(
+				"OnStart",
```
melekes: I think the error msg should be along the lines "Failed to connect to server"

types/priv_validator/socket.go
```diff
@@ -34,35 +37,65 @@ const (
...
+
 			time.Sleep(time.Second * dialRetryIntervalSeconds)
 			continue RETRY_LOOP
 		}
+
+		if pvsc.privKey != nil {
+			conn, err = p2pconn.MakeSecretConnection(conn, pvsc.privKey.Wrap())
+			if err != nil {
+				pvsc.Logger.Error(
+					"OnStart",
```
melekes: same here: "Failed to encrypt connection with server"

types/priv_validator/socket.go
```diff
@@ -71,43 +104,74 @@ func (pvsc *PrivValidatorSocketClient) OnStop() {
...
-func (pvsc *PrivValidatorSocketClient) Address() data.Bytes {
-	pubKey := pvsc.PubKey()
-	return pubKey.Address()
+// Address is an alias for PubKey().Address().
+func (pvsc *PrivValidatorSocketClient) Address() cmn.HexBytes {
+	return pvsc.PubKey().Address()
 }
 
+// PubKey implements PrivValidator.
 func (pvsc *PrivValidatorSocketClient) PubKey() crypto.PubKey {
```
melekes: Should not PubKey return error too by analogy with `SignProposal`?

types/priv_validator/socket.go
```diff
@@ -118,101 +182,142 @@ func (pvsc *PrivValidatorSocketClient) SignHeartbeat(chainID string, heartbeat *
...
-		var err error
-		pvss.conn, err = pvss.listener.Accept()
+		conn, err := pvss.listener.Accept()
 		if err != nil {
 			if !pvss.IsRunning() {
 				return // Ignore error from listener closing.
 			}
-			pvss.Logger.Error("Failed to accept connection: " + err.Error())
+			pvss.Logger.Error(
+				"accpetConnections",
```
melekes: "Failed to accept new connection"

types/priv_validator/socket.go
```diff
@@ -118,101 +182,142 @@ func (pvsc *PrivValidatorSocketClient) SignHeartbeat(chainID string, heartbeat *
...
 		}
 
-		pvss.Logger.Info("Accepted a new connection")
-
-		// read/write
-		for {
-			if !pvss.IsRunning() {
-				return // Ignore error from listener closing.
-			}
+		if err := conn.SetDeadline(time.Now().Add(time.Second)); err != nil {
```
melekes: deadline should be configurable or be a const, no magic numbers please :)

types/priv_validator/socket.go
```diff
@@ -118,101 +182,142 @@ func (pvsc *PrivValidatorSocketClient) SignHeartbeat(chainID string, heartbeat *
...
+			pvss.Logger.Error(
+				"acceptConnetions",
```
melekes: same here

types/priv_validator/socket.go
```diff
@@ -118,101 +182,142 @@ func (pvsc *PrivValidatorSocketClient) SignHeartbeat(chainID string, heartbeat *
...
+
+	for {
+		if !pvss.IsRunning() {
+			return // Ignore error from listener closing.
+		}
+
+		req, err := readMsg(conn)
+		if err != nil {
+			if err != io.EOF {
+				pvss.Logger.Error("handleConnection", "err", err)
```
melekes: same here

cmd/priv_val_server/main.go
melekes: Or this is a maximum number of concurrent clients?

xla: Yes, will make the description more clear.

types/priv_validator/socket.go
xla: While I see the point given the signature of `Error(msg string, keyvals ...interface{})` it seems very hard to build up the context the error was coming from as it would result in: 

```
E[02-13|17:42:57.620] Failed to accept new connection                                      err=foo
```

Maybe I'm not seeing an important aspect of error logging in our codeabse, but that line has no signal in it. Happy to incorporate your comments, just wondering if it is the best way to go abouut it.

types/priv_validator/socket.go
xla: See my [comment](https://github.com/tendermint/tendermint/pull/1204/files#r167945574) for some clarification that'd be important for me. :>

types/priv_validator/socket.go
melekes: You can turn on a tracing logger by providing `--trace` flag https://github.com/tendermint/tendermint/blob/master/cmd/tendermint/commands/root.go#L53

In this mode logger will output full trace if error supports this feature (pkg/errors does). https://github.com/tendermint/tmlibs/blob/master/log/tracing_logger.go#L58

xla: Thanks for pointing that out! This still doesn't really help after the fact, which makes errors emitted without tracing still not very useful.

melekes: Ok. I see your point. So the agreement always was: `meaningful-message key=value`. For example: "Failed to accept connection err=EOF". We're still in a process of improving the messages. Plus we started to use `pkg/errors` not so long ago to accumulate context.

Usually, what we do is just grep by msg ("Failed to accept connection"). Of course, if you have many places in your app where you accept a connection, then msg must be changed to something like "Failed to accept connection from X".

I am happy to hear any suggestions on how to improve the current scheme.

types/priv_validator/socket.go
xla: Thought so too and addressed it in 2a292ef, but would like to have a sanity check from @ebuchman on this. :>

types/priv_validator/socket.go
xla: Very sound approach and I don't mean to start a greater discussion if we already iterating on error handling in other places.

> I am happy to hear any suggestions on how to improve the current scheme.

Usually thinking about in what remote ends logs can end up in typical infrastructure I try to imagine what vital info is needed to make sense of it when consumed in some log infrastructure remote from the origin. Additionally how each log line can interpreted by itself without context of the lines before or after. As the logger is passed from the outside to the inside in the program we can enrich the keyvals with information as well which subsystems don't need to care about. A standard set of fields I try to surface, most of them which can easily be achieved with Dynamic Contextual Values from the go-kit log package:

* **caller**: https://godoc.org/github.com/go-kit/kit/log#Caller
* **timestamp**: https://godoc.org/github.com/go-kit/kit/log#pkg-variables
* **revision**: compile flag: https://github.com/tendermint/tendermint/blob/master/Makefile#L10
* **job/system**: `tendermint` in our case
* **cmd/task**: `node`for the node command
* **service/subsystem**: what we pass to `cmn.NewBaseService(...)` e.g.: `privValidatorSocketServer`
* **err**: set if error present

node/node.go
```diff
@@ -171,6 +184,21 @@ func NewNode(config *cfg.Config,
 	// reload the state (it may have been updated by the handshake)
 	state = sm.LoadState(stateDB)
 
+	/* TODO
+	// Generate node PrivKey
+	privKey := crypto.GenPrivKeyEd25519()
```
zmanian: I've been debating if this should be ephemeral.

In the ephemeral version, I imagine perhaps an ssh like confirm dialog in signatory before signatory starts accepting signing requests for a node.

if this passed in via a config file, then signatory would maintain a white list

types/priv_validator/socket.go
ethanfrey: Very nice

types/priv_validator/socket.go
```diff
@@ -71,148 +90,276 @@ func (pvsc *PrivValidatorSocketClient) OnStop() {
...
 }
 
-func NewPrivValidatorSocketServer(logger log.Logger, socketAddr, chainID string, privVal PrivValidator) *PrivValidatorSocketServer {
+// NewPrivValidatorSocketServer returns an instance of
+// PrivValidatorSocketServer.
+func NewPrivValidatorSocketServer(
+	logger log.Logger,
+	chainID, socketAddr string,
+	maxConnections int,
+	privVal PrivValidator,
```
ebuchman: if SocketServer is what runs in the Tendermint process (ie. it's listening on the socket for a connection from a priv_val_client, but it's also the one that initiates all requests), then it won't have access to a PrivValidator because it doesn't have a key for it - it only gets access to the key by sending messages over the socket

config/config.go
```diff
@@ -103,6 +105,12 @@ type BaseConfig struct {
 	// A custom human readable name for this node
 	Moniker string `mapstructure:"moniker"`
 
+	// TCP or UNIX socket address of the PrivValidator server
+	PrivValidatorAddr string `mapstructure:"priv_validator_addr"`
```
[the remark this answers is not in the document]
xla: See #1231

types/priv_validator/socket.go
```diff
@@ -71,148 +90,276 @@ func (pvsc *PrivValidatorSocketClient) OnStop() {
...
+
+RETRY_LOOP:
+	for retries > 0 {
+		if retries != dialRetryMax {
+			time.Sleep(time.Second * dialRetryIntervalSeconds)
+		}
+
+		retries--
+
+		conn, err := cmn.Connect(pvsc.SocketAddress)
```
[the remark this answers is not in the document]
xla: If it errors we don't need to worry about any cleanup, we do need to defer the closing of it if it succeeds.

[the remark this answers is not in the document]
ebuchman: If this one errors, sure nothing to clean, but what about any of the points below where we `continue`. Eg. if the SecretConnection fails, I think we need to close the conn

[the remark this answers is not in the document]
xla: Absolutely!

PR #1213, opened 2018-02-15
Remove lite/proxy

### LINKED ISSUE #1183: refactor `lite` so iavl dependency is outside tendermint core
`lite` shouldn't depend on `iavl`. It should *only* be concerned with Tendermint related data structures (ie. block headers, votes, validators, etc.).

Ie. see commented out code https://github.com/tendermint/tendermint/blob/develop/lite/proxy/query.go#L86

ethanfrey:
This resolves #1183 
The proxy must be able to validate query proofs, otherwise it loses much meaning, as that is the principle activity is is designed for (as well as validating tx queries). 

I understand the dependency madness, and moved it into cosmos-sdk
(See PR: )

The other longer-term solution, is to have a general format for merkle proofs that is independent of the iavl tree and supports many different types. Design work began already: https://github.com/cosmos/cosmos-sdk/blob/feature/ibc/docs/spec/ibc/protobuf/merkle.proto But feedback and implementation neeeded.


ebuchman:
I was thinking actually that we could keep all this and just not return verified/formatted abci-query proofs - they could just be bytes until we figure out the generic structure. `lite/proxy` is still useful to proxy the tendermint rpc and be able to prove an app-hash and/or tx was committed. We just need to let callers handle the abci query proofs. 

`lite/proxy` could even let callers pass in a function that handles the proof deserialization and verification - then lite/proxy still wouldn't need to import `iavl`, but would execute everything necessary to verify it fully - the iavl dependent code would just be passed in from the SDK.

ethanfrey:
The code remaining in lite does all the header verification. I left that there.
Proxy built upon that to use the headers to verify /tx and /abci_query responses.

We will need to allow custom proof types for each application, especially now that there is multi-store. This is a bigger question and doesn't belong in tendermint. For that reason, I think that proxy doesn't really belong there either. Especially, as requiring "passing in a function" would render `tendermint lite` almost meaningless.

The hard logic is still in lite. Proxy is just some wrappers.

ethanfrey:
What I see is you would provide a proxy with no `query` method? Is that really that useful?

Especially since trying to refine such a proxy in another repo is a recipe for headache (the more you need to touch multiple repos to complete one feature, the more you hate glide).

I like putting it into cosmos-sdk and providing a prover solution that works well with the multi-store, and the proxy and the proofs can evolve side by side *in the same repo*. If someone wants to produce a different prover for a proxy, they can do that in their repo, still using tendermint/lite code.

ebuchman:
No you'd still have `query` (and everything else). The only difference is the query proof isnt verified.  It's still useful because it does all the work of verifying that the root hash was committed too, and leaves it to the client to check the app proof.

ebuchman:
Ok sure - the proxy could be duplicated in SDK. still makes sense to keep it in Tendermint though - eg. lotion uses it

ethanfrey:
But lotion uses it for the queries, no?

ebuchman:
Yeh but the proofs are verified client side ... `tendermint lite` should just check if theres valid tendermint signatures on the app hash, and returns the query proof as a blob for the client to deal with

ethanfrey:
The root hash is not present in the query response, except encoded in the proof.
I can only pass through proofs unverified.

I don't like to have a verifying proxy that passes unverified data, as people will trust it without other checks

ethanfrey:
Proposal. Not a proxy, but a limited verifier.

You can query for blocks, headers, or tx. These will be verified and returned.
All other queries should be made to the remote rpc server.

You have a local api to get the header info. And then can do what you want with it. This would fit inside tendermint and require no iavl. However, all light-client proofs need a root of trust to initialize the provider for the proxy. That is done in `gaia client init` now, how does that happen in tendermint?

ethanfrey:
I would actually make a second binary in the tendermint repo.

`proxy start`, `proxy init` or something like that.
I will add code back in to make it happen, just want a clear design that makes sense.

ebuchman:
I think the above discussion is valuable. See comment in https://github.com/tendermint/tendermint/issues/1183#issuecomment-369822047 that remains the latest update for this issue.

Closing this PR because we don't actually want to remove all this code (just refactor it a bit to remove dependency on iavl)

PR #1239, opened 2018-02-23
Restart pub/sub server on service reset

simonvadee:
Hi fellas,

I encountered an issue when stopping/restarting a `rpc.client.Client` because of:
```golang
// Start implements Service by calling OnStart (if defined). An error will be
// returned if the service is already running or stopped. Not to start the
// stopped service, you need to call Reset.
func (bs *BaseService) Start() error { ... }
```

Therefore I used `func (bs *BaseService) Reset()` but since the `EventBus` does not implement `OnReset()` I got a `PanicSanity("The service cannot be reset")`. This is why I added this method to `EventBus`.

The reason why I want to reset the service is that once it is stopped, the pub/sub server is shutdown and cannot be restarted, so even if subscribers are still able to subscribe to new events, they are not notified.

I don't really know what kind of tests you'd like me to write or doc you want me to update so let me know what you'd like !

Cheers!

### DIFF
--- a/types/event_bus.go
+++ b/types/event_bus.go
@@ -52,6 +52,11 @@ func (b *EventBus) OnStop() {
 	b.pubsub.OnStop()
 }
 
+// OnReset restarts the pub/sub server
+func (b *EventBus) OnReset() error {
+	return b.pubsub.OnReset()
+}
+
 func (b *EventBus) Subscribe(ctx context.Context, subscriber string, query tmpubsub.Query, out chan<- interface{}) error {
 	return b.pubsub.Subscribe(ctx, subscriber, query, out)
 }
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,9 +2,10 @@
 
 
 [[projects]]
+  branch = "master"
   name = "github.com/btcsuite/btcd"
   packages = ["btcec"]
-  revision = "50de9da05b50eb15658bb350f6ea24368a111ab7"
+  revision = "2be2f12b358dc57d70b8f501b00be450192efbc3"
 
 [[projects]]
   name = "github.com/davecgh/go-spew"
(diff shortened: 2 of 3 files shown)


types/event_bus.go
```diff
@@ -52,6 +52,10 @@ func (b *EventBus) OnStop() {
 	b.pubsub.OnStop()
 }
 
+func (b *EventBus) OnReset() error {
+	return b.pubsub.OnStart()
```
melekes: why not `OnReset`?

simonvadee: `pubsub.Sever` does not implement `OnReset()` and `OnStart()` starts the server as intended, do you want me to add the missing method in `tendermint/tmlibs` ? And then should I point glide.yaml to the master branch of tmlibs ?

melekes: > do you want me to add the missing method in tendermint/tmlibs ?

yes, please

```go
func (s *Server) OnReset() error {	return nil }
```

> And then should I point glide.yaml to the master branch of tmlibs ?

no, point it to commit in tmlibs

melekes: Want me to do it?

simonvadee: It's not an urgent matter, i can take care of it when I'll have a few minutes, but thank's for asking :)

melekes:
merged associated PR. @simonvadee could you update this one when u have time?

melekes:
Hmmm.. tests are failing. Could you also rebase your code against latest `develop`?

xla:
@melekes The tests timed out, kicked off a re-run.

xla:
@simonvadee We did some changes to our CI, could you rebase latest `tendermint:develop`?

melekes:
> https://github.com/tendermint/tendermint/pull/1239#issuecomment-370763774

did not help.

simonvadee:
what's the status on this, do you want me to do something about the tests ?

xla:
@simonvadee Looks like the change you made introduced a condition where our tests are blocked, this needs further investigation. Maybe you can reproduce it locally as a start?

ebuchman:
Thanks for this.

Unfortunately there have been a number of issues around Restart and we'll probably be getting rid of it altogether. Instead of restarting, one should just recreate the service and call Start. The `Service` is undergoing refactor and will move into tendermint. See https://github.com/tendermint/tmlibs/pull/114

PR #1243, opened 2018-02-25
Switch to dep from glide for dependency management

zmanian:
Changes dependency resolution from glide to dep and updates `Makefile` to use dep instead of glide.


melekes:
```
tep 10 : RUN make get_vendor_deps
 ---> Running in e155101d118a
--> Running dep
could not find project Gopkg.toml, use dep init to initiate a manifest
Makefile:50: recipe for target 'get_vendor_deps' failed
```

hxzqlh:
Anybody can explain why change pkg management tool from glide to dep?

Issue #1147, opened 2018-01-24
Using addrbook.MarkGood and sw.StopPeerForError


ebuchman:
The long term plan is to use a PID controller to do this, aka the "Trust Metric". See https://github.com/tendermint/tendermint/issues/867

In the meantime, we take the following tack:

- we call `addrbook.MarkGood(peer)` when a peer has contributed unique votes for 10000 blocks and unique block parts for 10000 blocks.
- we call `sw.StopPeerForError` on any kind of bad behaviour - invalid msg decoding or values, unknown reactor channels, timeouts, data that fails to authenticate, etc.

Note we don't want to pass the `addrbook` everywhere - perhaps the switch can expose `MarkGood` and let it flow through to the AddrBook somehow.

For `StopPeerForError`, we can use the event system to get peer errors from inner objects back up to the Reactor, which can then call sw.StopPeerForError.

ebuchman:
Implemented in https://github.com/tendermint/tendermint/pull/1244

See the followup in https://github.com/tendermint/tendermint/issues/1317

PR #1244, opened 2018-02-26
Using MarkGood and StopPeerForError

melekes:
Refs #1147

### DIFF
--- a/blockchain/reactor.go
+++ b/blockchain/reactor.go
@@ -76,8 +76,9 @@ func NewBlockchainReactor(state sm.State, blockExec *sm.BlockExecutor, store *Bl
 			store.Height()))
 	}
 
-	requestsCh := make(chan BlockRequest)
-	errorsCh := make(chan peerError)
+	const cap = 1000 // must be bigger than peers count
+	requestsCh := make(chan BlockRequest, cap)
+	errorsCh := make(chan peerError, cap) // so we don't block in #Receive#pool.AddBlock
 	pool := NewBlockPool(
 		store.Height()+1,
 		requestsCh,
--- a/blockchain/reactor.go
+++ b/blockchain/reactor.go
@@ -72,18 +72,20 @@ func NewBlockchainReactor(state sm.State, blockExec *sm.BlockExecutor, store *Bl
 	fastSync bool) *BlockchainReactor {
 
 	if state.LastBlockHeight != store.Height() {
-		cmn.PanicSanity(cmn.Fmt("state (%v) and store (%v) height mismatch", state.LastBlockHeight,
+		panic(fmt.Sprintf("state (%v) and store (%v) height mismatch", state.LastBlockHeight,
 			store.Height()))
 	}
 
 	const cap = 1000 // must be bigger than peers count
 	requestsCh := make(chan BlockRequest, cap)
 	errorsCh := make(chan peerError, cap) // so we don't block in #Receive#pool.AddBlock
+
 	pool := NewBlockPool(
 		store.Height()+1,
 		requestsCh,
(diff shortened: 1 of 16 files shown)


blockchain/reactor.go
```diff
@@ -22,8 +22,7 @@ const (
 	// BlockchainChannel is a channel for blocks and status updates (`BlockStore` height)
 	BlockchainChannel = byte(0x40)
 
-	defaultChannelCapacity = 1000
```
ebuchman: why not leave this here if we still using it ? bad naming ?

consensus/reactor.go
```diff
@@ -1055,6 +1111,36 @@ func (ps *PeerState) ensureVoteBitArrays(height int64, numValidators int) {
 	}
 }
 
+// RecordVote updates internal statistics for this peer by recording the vote.
+// It returns the total number of votes (1 per block). This essentially means
+// the number of blocks for which peer has been sending us block parts.
```
ebuchman: `sending us block parts` -> `sending us votes`

consensus/reactor.go
```diff
@@ -1055,6 +1111,36 @@ func (ps *PeerState) ensureVoteBitArrays(height int64, numValidators int) {
...
+func (ps *PeerState) RecordVote(vote *types.Vote) int {
+	ps.mtx.Lock()
+	defer ps.mtx.Unlock()
+
+	if ps.stats.lastVoteHeight == vote.Height {
+		return ps.stats.votes
+	}
+
```
ebuchman: dont we need to set lastVoteHeight now ?

consensus/reactor.go
```diff
@@ -1055,6 +1111,36 @@ func (ps *PeerState) ensureVoteBitArrays(height int64, numValidators int) {
...
+// It returns the total number of block parts (1 per block). This essentially means
+// the number of blocks for which peer has been sending us block parts.
+func (ps *PeerState) RecordBlockPart(bp *BlockPartMessage) int {
+	ps.mtx.Lock()
+	defer ps.mtx.Unlock()
+
+	if ps.stats.lastBlockPartHeight == bp.Height {
+		return ps.stats.blockParts
+	}
+
```
ebuchman: dont we need to set lastBlockPartHeight now ?

blockchain/reactor.go
melekes: i think so, yeah. it does not explain a) why it's default? b) why 1000 is a default? c) is it supposed to be used for all channels?

ebuchman:
Let's rebase this so new CI runs ? 

Also which should be merged first this or #1282  ?

melekes:
> Also which should be merged first this or #1282 ?

does not matter

melekes:
@ebuchman rebased

consensus/reactor.go
```diff
@@ -1055,6 +1073,37 @@ func (ps *PeerState) ensureVoteBitArrays(height int64, numValidators int) {
 	}
 }
 
+// RecordVote updates internal statistics for this peer by recording the vote.
+// It returns the total number of votes (1 per block). This essentially means
+// the number of blocks for which peer has been sending us votes.
+func (ps *PeerState) RecordVote(vote *types.Vote) int {
+	ps.mtx.Lock()
+	defer ps.mtx.Unlock()
```
xla: do  we need the mutex here ?

ebuchman:
Follow up in https://github.com/tendermint/tendermint/issues/1317

ebuchman:
Also this needs tests. Might be tough to test given the current setup of consensus testing so worst case we leave that as an open issue

PR #1247, opened 2018-02-27
Integrate CodeCov as change acceptance stage

xla:
As a rough measure to keep quality up we want to integrate our code coverage tooling to the point where it is required for changes to be merged.

Example taken from https://github.com/cosmos/voyager/blob/develop/codecov.yaml


faboweb:
Note: I would make codecov/patch also required. For /project one could lower the coverage in one part and increase it in another which is not desired. But maybe try it out and change this later.

ebuchman:
Happy to merge this but I think we need to make the change in the repo settings, no ?

faboweb:
If you want those checks to be required for a PR to pass, this needs to be turned on per branch (develop / master) in the settings.
image

Issue #1255, opened 2018-02-28
PrivValidator follow up


ebuchman:
- [x] https://github.com/tendermint/tendermint/pull/1203 has been merged but please follow up on Anton's review re `pvsc.Stop()`, `panic`, and `sign_bytes`
- [x] Tendermint is dialing the signer, but we need the signer to dial Tendermint. Perhaps we can have the SocketClient take a `conn` directly to avoid the confusion around a client doing the listening (the ADR did not specify this well, my bad - see https://github.com/tendermint/tendermint/pull/1256 for a start)
- [x] Some more review follow up from https://github.com/tendermint/tendermint/pull/1286#pullrequestreview-103093019
- [ ] priv_val_server needs to be updated 
- [ ] authenticate against `ID@host:port` in the RemoteSigner, check that `id` matches configured 
- [ ] Persist PrivKey the SocketClient uses for external signers to authenticate properly
- [ ] full integration and deprecate the old PrivValidator

xla:
*added TODO*

tarcieri:
If it's too much work for the signing service to connect to Tendermint, we can go back to the original plan of having Tendermint call out to the signer.

Having the signer connect rather than listen is a little bit of defense-in-depth, but so long as both sides are mutually authenticating each other, they should be equivalent from a security perspective, barring some sort of pre-auth vulnerability in the signing service.

xla:
@tarcieri See #1286 we going to stick to the original plan and have the signer(s) connect to Tendermint.

tarcieri:
@xla great!

silasdavis:
I've kind of lost track of where some of this had gone, but at one point we had some extracted signing logic for `PrivValidator` which is now private to `FilePV`. For mean this means I have to maintain this: https://github.com/hyperledger/burrow/blob/develop/consensus/tendermint/sign_info.go so that I can use my in-memory `PrivValidator`. That logic would really much better reside with Tendermint.

I'm not sure why the decision was made to remove `LastSignInfo` from Tendermint, but I think it may have been lost as collateral to: https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-008-priv-validator.md. But making the validation logic its own exported class is orthogonal to supporting just two validators - that decision makes sense, but why prevent others from building from the validation logic which is separable and could live in its own exported type or just exported methods.

ebuchman:
Closing for #2549

Issue #1185, opened 2018-02-03
Proposal: Use Raft (or other) for fault-tolerant validator processes


nickray:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
FEATURE REQUEST

This is a somewhat complementary or even orthogonal to [issue 1136](https://github.com/tendermint/tendermint/issues/1136).

A validator node can fail in many ways, at least two are:
* network connectivity loss (minor protocol penalty for being offline). This could be fixed via issue 1136
* machine breaks (major protocol penalty if double signing, or maybe even for not respecting prevote-the-lock). Issue 1136 does not really resolve this validator obligation.

Is there a promise in the implementation that the state in `/data` is always correct, in the sense of being able to restart a copied node with this serialized state? If so, what are the current conditions of this promise?

Additionally, the production validator node may want to store state on an SDD (which would be lost under machine failure), not some kind of "safe but slow" NAS. A poor man's version could be running on the SSD, but using something like "minio mirror" to sync the SSD to backup. This has failure modes as well though...

So I'm wondering if the idea of "[raftifying](http://thesecretlivesofdata.com/raft/)" the validator node has been thought through. I think it would certainly be feasible. The basic idea is:
* run 3 validator node instances (each with a copy of the HSM key)
* one of them is the leader, participating in Tendermint as "the" validator
*  if it fails then another one takes over via Raft
* perhaps additional nodes not participating in the Raft leader/candidate mechanism could listen and store the state onto backup NAS.

I have been thinking about forking the repo and trying to implement this as an experiment. One problem is that it is not clear to me what exactly the "internal state" of a validator node is (that would be replicated via Raft), as opposed to the "external state" it is involved in according to the Tendermint BFT protocol.

For inspiration, here is a project [doing this for SQLite](https://github.com/rqlite/rqlite).

tarcieri:
This approach sounds roughly analogous to how Google operates their CT logs: they run multiple replicas of the same log in geographically distributed datacenters and use Chubby to elect one to be active at a given point in time. I've heard mixed opinions on using Raft for generalized leader election, though: even though Raft runs an internal leader election protocol, the real goal of Raft is to come to distributed consensus on a log.

There are several important considerations for something like this, like how this will interact with the prospective KMS service I'm working on and how this could be leveraged to prevent double signing.

[...]

PR #1256, opened 2018-02-28
PrivValidatorAddr -> PrivValidatorListenAddr. Update ADR008

ebuchman:
PrivValidator follow up re https://github.com/tendermint/tendermint/issues/1255

### DIFF
--- a/cmd/tendermint/commands/run_node.go
+++ b/cmd/tendermint/commands/run_node.go
@@ -15,7 +15,7 @@ func AddNodeFlags(cmd *cobra.Command) {
 	cmd.Flags().String("moniker", config.Moniker, "Node Name")
 
 	// priv val flags
-	cmd.Flags().String("priv_validator_addr", config.PrivValidatorAddr, "Socket address for private validator")
+	cmd.Flags().String("priv_validator_laddr", config.PrivValidatorListenAddr, "Socket address to listen on for connections from external priv_validator process")
 
 	// node flags
 	cmd.Flags().Bool("fast_sync", config.FastSync, "Fast blockchain syncing")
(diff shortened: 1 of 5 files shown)


ebuchman:
I think maybe we can split the `socketClient` from a `listener` so that we can clarify that Tendermint is the process that actually listens on the socket, but once it has a connection, it functions as the client by sending requests. So maybe `NewSocketClient` is only called once we already have a connection and then it doesn't need to do anything in OnStart

ebuchman:
Really great questions thank you!

> Is it halting any other operations?

Simplest thing to do is probably to halt everything else until we've received a connection. But in an ideal world, we would still boot everything up except maybe the Consensus. I'm not sure exactly how this should be handled, since it might be hard to distinguish between not having received the first connection and the connection failing later. So probably best to just keep it simple at first :). Interested in your thoughts on how we might do it right though.

> Does it expect the signer to connect in a certain timeframe?

I could go either way on this one. Let's flame out for now after a minute or something. Would be valuable to get feedback from validators/users on how they want this to work. It's really a UX question.

> Does it accept at most one connection?

I think we should accept more than 1. It can be a config option. We want folks to be able to set up multiple validators that Tendermint talks to in case one fails. That said, it will be critical that those validators stay synced, since if they don't, they might double signed. So maybe we default to 1 and let folks change it if they want more, but warn them that it means they have to take precaution to keep them synced by running a consensus protocol between the signers (eg. https://github.com/tendermint/tendermint/issues/1185)

> What is the behaviour in case the signer disconnects?

I think we start logging errors and wait a few minutes for it to reconnect. If the signer doesn't come back after a few minutes, perhaps we should panic. 

Maybe @zmanian @tarcieri @nickray want to weigh in on some of this behaviour

tarcieri:
My personal inclination would be for it to continue to run until it receives connection, and perhaps log a warning that there are no signers connected. If it's offering the service, and panicking, it's just one more place for things to go wrong.

xla:
Rebased on latest develop.

PR #1233, opened 2018-03-01
p2p: if we have no peers we should dial seeds right away

xla:
We dial persistent_peers right away, but it takes a while to dial seeds for the first time. If we have no peers, we should dial seeds right away

### DIFF
--- a/p2p/pex/pex_reactor.go
+++ b/p2p/pex/pex_reactor.go
@@ -260,9 +260,17 @@ func (r *PEXReactor) SetEnsurePeersPeriod(d time.Duration) {
 
 // Ensures that sufficient peers are connected. (continuous)
 func (r *PEXReactor) ensurePeersRoutine() {
-	// Randomize when routine starts
-	ensurePeersPeriodMs := r.ensurePeersPeriod.Nanoseconds() / 1e6
-	time.Sleep(time.Duration(rand.Int63n(ensurePeersPeriodMs)) * time.Millisecond)
+	var (
+		seed   = rand.New(rand.NewSource(time.Now().UnixNano()))
+		jitter = seed.Int63n(r.ensurePeersPeriod.Nanoseconds())
+	)
+
+	// Randomize first round of communication to avoid thundering herd.
+	// If no potential peers are present directly start connecting so we guarantee
+	// swift setup with the help of configured seeds.
+	if r.hasPotentialPeers() {
+		time.Sleep(time.Duration(jitter))
+	}
 
 	// fire once immediately.
 	// ensures we dial the seeds right away if the book is empty
@@ -287,9 +295,18 @@ func (r *PEXReactor) ensurePeersRoutine() {
 // the node operator. It should not be used to compute what addresses are
 // already connected or not.
 func (r *PEXReactor) ensurePeers() {
-	numOutPeers, numInPeers, numDialing := r.Switch.NumPeers()
-	numToDial := defaultMinNumOutboundPeers - (numOutPeers + numDialing)
-	r.Logger.Info("Ensure peers", "numOutPeers", numOutPeers, "numDialing", numDialing, "numToDial", numToDial)
+	var (
+		out, in, dial = r.Switch.NumPeers()
+		numToDial     = defaultMinNumOutboundPeers - (out + dial)
+	)
+	r.Logger.Info(
+		"Ensure peers",
+		"numOutPeers", out,
+		"numInPeers", in,
+		"numDialing", dial,
+		"numToDial", numToDial,
+	)
+
 	if numToDial <= 0 {
 		return
 	}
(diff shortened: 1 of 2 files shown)


xla:
@ebuchman @melekes Ready for review.

p2p/pex/pex_reactor_test.go
```diff
@@ -259,44 +226,66 @@ func TestPEXReactorAddrsMessageAbuse(t *testing.T) {
 }
 
 func TestPEXReactorUsesSeedsIfNeeded(t *testing.T) {
+	require := require.New(t)
```
ebuchman: I think we want to avoid this. I don't really mind either way, but for consistency with the rest of the code. Also, it's not great to have variables with the same name as packages and it's not really any trouble to include `t` as an argument to things

p2p/pex/pex_reactor.go
```diff
@@ -260,9 +260,16 @@ func (r *PEXReactor) SetEnsurePeersPeriod(d time.Duration) {
 
 // Ensures that sufficient peers are connected. (continuous)
 func (r *PEXReactor) ensurePeersRoutine() {
-	// Randomize when routine starts
-	ensurePeersPeriodMs := r.ensurePeersPeriod.Nanoseconds() / 1e6
-	time.Sleep(time.Duration(rand.Int63n(ensurePeersPeriodMs)) * time.Millisecond)
+	rand.Seed(time.Now().UnixNano())
```
ebuchman: Doesn't this have global effect ? Maybe we should use a localized seed in case this bites us when we're looking for determinism somewhere later

ebuchman:
Did we not have a corresponding issue for this ?

p2p/pex/pex_reactor.go
melekes: see https://github.com/tendermint/tmlibs/blob/develop/common/random.go

p2p/pex/pex_reactor.go
```diff
@@ -414,6 +431,14 @@ func (r *PEXReactor) crawlPeersRoutine() {
 	}
 }
 
+// hasPotentialPeers indicates if there is a potential peer to connect to, by
+// consulting the Switch as well as the AddrBook.
+func (r *PEXReactor) hasPotentialPeers() bool {
+	o, i, d := r.Switch.NumPeers()
```
melekes: I'd prefer whole words here... `outgoing, incoming, dialing = `

p2p/pex/pex_reactor_test.go
xla: Thought this is the preferred way, looking at most other test cases (especially the ones in the same file). I agree that package names shouldn't be shadowed by vars in the local scop, but also would prefer a consistent style, which currently seems to be this one.

xla:
@ebuchman I used some dark magic to turn this into a PR as it was on the board already so we didn't ave to "track" two things. Maybe too confusing after all?

p2p/pex/pex_reactor.go
xla: Was debating that when writing it, here the lifcycle of the vars is the next line, which was the original idea for single character variables being advocated for in go. Will change, wondering if it will add much for readers to understand better, tho.

ebuchman: I think it's fine the way it is since it gets consumed right away and the words are otherwise large.

That said I think we're supposed to avoid single letter names. Maybe `out, in, dial` would be nice. I guess at that point, adding an `ing` to everything is no biggie. Meh.

p2p/pex/pex_reactor_test.go
ebuchman: Whoops you're right. We've had some back and forth on this. Let's work towards passing `t` explicitly in the calls.

p2p/pex/pex_reactor.go
xla: Will expand them single characters, like the shorter version as well.

ebuchman:
> @ebuchman I used some dark magic to turn this into a PR as it was on the board already so we didn't ave to "track" two things. Maybe too confusing after all?

Dark magic indeed! How did you do it? Also yeh, it's confusing :D

xla:
Githubs API v3 has supports an optional issue field on the pull request endpoint, which transforms the original issue.

xla:
@ebuchman https://developer.github.com/v3/pulls/#alternative-input

p2p/pex/pex_reactor.go
xla: Good catch.

@melekes Unfortunately `common` doesn't offer a method similar to `Int63n`, is worth adding it and using here?

Gonna go for local seed in the meantime.

p2p/pex/pex_reactor_test.go
xla: On board with that. :ship:

xla:
@melekes Feedback addressed.

p2p/pex/pex_reactor.go
```diff
@@ -287,9 +295,18 @@ func (r *PEXReactor) ensurePeersRoutine() {
 // the node operator. It should not be used to compute what addresses are
 // already connected or not.
 func (r *PEXReactor) ensurePeers() {
-	numOutPeers, numInPeers, numDialing := r.Switch.NumPeers()
-	numToDial := defaultMinNumOutboundPeers - (numOutPeers + numDialing)
-	r.Logger.Info("Ensure peers", "numOutPeers", numOutPeers, "numDialing", numDialing, "numToDial", numToDial)
+	var (
+		numOutPeers, numInPeers, numDialing = r.Switch.NumPeers()
```
melekes: why not rename vars here too? `out, in, dialing = `

xla: Good catch, will address.

p2p/pex/pex_reactor.go
```diff
@@ -414,6 +432,14 @@ func (r *PEXReactor) crawlPeersRoutine() {
 	}
 }
 
+// hasPotentialPeers indicates if there is a potential peer to connect to, by
+// consulting the Switch as well as the AddrBook.
+func (r *PEXReactor) hasPotentialPeers() bool {
+	out, in, dial := r.Switch.NumPeers()
```
melekes: `dial` is ambiguous in my opinion

xla: What would you propose?

PR #1245, opened 2018-02-26
Make RPC endpoint handler URLs http/https-agnostic

nickray:
* [ ] Updated all relevant documentation in docs
* [ ] Updated all code comments where relevant
* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

PR #1259, opened 2018-03-01
Protocol-agnostic URLs in RPC endpoint

nickray:
Replaces https://github.com/tendermint/tendermint/pull/1245


ebuchman:
Huh same issue. @nickray do you have some setting in your github or on Circle that could be preventing this from building ?

nickray:
I don't think so, no. I don't have a Circle. This sucks.
Maybe you can just change the two lines directly?

xla:
@nickray Hej Nick, sorry for the inconvenience. We followed-up with circle and see if escalating with them will help with the situation, if we don't resolve it swiftly with them we gonna re-create the PR from one of our accounts so the change goes in soon.

xla:
@nickray Closing this in favour of #1276. Didn't manage to make you proper author, guess not using the right email.

PR #1221, opened 2018-02-19
WIP: Breaking/wire sdk2

ebuchman:
https://github.com/tendermint/tendermint/issues/1186

PR #1265, opened 2018-03-02
Bucky/new wire api

ebuchman:
Start updating for new go-wire API without breaking the underlying encoding scheme.

These changes should be purely cosmetic, but should make actually upgrading later a lot more straight forward and involve less code.

Started cherry-picking commits from https://github.com/tendermint/tendermint/pull/1221.

We can just keep doing so as we please until the whole repo is using the Marshal API

### DIFF
--- a/state/store.go
+++ b/state/store.go
@@ -1,7 +1,6 @@
 package state
 
 import (
-	"bytes"
 	"fmt"
 
 	abci "github.com/tendermint/abci/types"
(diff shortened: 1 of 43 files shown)


state/store.go
```diff
@@ -113,7 +111,11 @@ func NewABCIResponses(block *types.Block) *ABCIResponses {
 
 // Bytes serializes the ABCIResponse using go-wire
 func (a *ABCIResponses) Bytes() []byte {
-	return wire.BinaryBytes(*a)
+	bz, err := wire.MarshalBinary(*a)
+	if err != nil {
+		panic(err)
```
melekes: don't like so many panics. later we'll probably want to do something about them

melekes:
👍

PR #1280, opened 2018-03-06
docs: add 'On Determinism'

zramsay:
This is, admittedly, a half-measure but it gets the ball rolling on having a place to discuss determinism.

closes https://github.com/tendermint/abci/issues/56


melekes:
Nice 👍

Issue #1126, opened 2018-01-20
p2p: private_peers flag


ebuchman:
Add a `--p2p.private_peers` flag and config var that takes a comma separated list of peer IDs (ie. 20 bytes hex-encoded). 

IDs passed through `--p2p.private_peers` should not be added to the AddrBook and thus not gossipped to other peers.

PR #1282, opened 2018-03-06
private peers

melekes:
Refs #1126

* [x] Wrote tests
* [x] Updated CHANGELOG.md

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -421,6 +421,14 @@ func (n *Node) OnStart() error {
 		}
 	}
 
+	// Always connect to private peers, but do not add them to addrbook
+	if n.config.P2P.PrivatePeers != "" {
+		err = n.sw.DialPeersAsync(nil, strings.Split(n.config.P2P.PrivatePeers, ","), true)
+		if err != nil {
+			return err
+		}
+	}
+
 	// start tx indexer
 	return n.indexerService.Start()
 }
--- a/node/node.go
+++ b/node/node.go
@@ -422,38 +422,9 @@ func (n *Node) OnStart() error {
 
 	// Always connect to persistent peers
 	if n.config.P2P.PersistentPeers != "" {
-		// are any of the persistent peers private?
-		persistentPeers := []string{}
-		persistentAndPrivatePeers := []string{}
-		var privatePeerIDs []string
-		if n.config.P2P.PrivatePeerIDs != "" {
-			privatePeerIDs = strings.Split(n.config.P2P.PrivatePeerIDs, ",")
-		}
-	PP_LOOP:
-		for _, peer := range strings.Split(n.config.P2P.PersistentPeers, ",") {
-			spl := strings.Split(peer, "@")
-			if len(spl) == 2 {
-				for _, ppID := range privatePeerIDs {
-					if spl[0] == ppID {
-						persistentAndPrivatePeers = append(persistentAndPrivatePeers, peer)
-						continue PP_LOOP
-					}
-				}
-			}
-			persistentPeers = append(persistentPeers, peer)
-		}
-
-		err = n.sw.DialPeersAsync(n.addrBook, persistentPeers, true)
+		err = n.sw.DialPeersAsync(n.addrBook, strings.Split(n.config.P2P.PersistentPeers, ","), true)
 		if err != nil {
 			return err
-		}
-
-		// if any of the persistent peers are private, do not add them to addrbook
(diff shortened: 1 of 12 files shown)


rpc/core/net.go
```diff
@@ -2,6 +2,7 @@ package core
 
 import (
 	"github.com/pkg/errors"
+	p2p "github.com/tendermint/tendermint/p2p"
```
xla: Import ordering should have newline between external and internal imports.

melekes:
> Is there an issue or some other context to this change?

see PR's description

zmanian:
So it looks like this change set prevents peers specified in outgoing connections form entering the address book for gossip.

How about peers from incoming connections?

I mostly expect validations to dial out to their sentries rather than sentries dialing back to their validators.

melekes:
@zmanian good call! will do

node/node.go
```diff
@@ -415,10 +422,39 @@ func (n *Node) OnStart() error {
 
 	// Always connect to persistent peers
 	if n.config.P2P.PersistentPeers != "" {
-		err = n.sw.DialPeersAsync(n.addrBook, strings.Split(n.config.P2P.PersistentPeers, ","), true)
+		// are any of the persistent peers private?
```
ebuchman: it would be great if this logic didn't need to be in `node`. Maybe we can get the `private_peers` field into the switch so it can figure out when to add peers to the addrbook or not. Also the #1244 puts the addrbook in the Switch so at that point we can probably remove it as an arg to DialPeers. 

In any case, this is fine for now, lets open some issues related to p2p cleanup and improving the separation between switch/addrbook/pex

node/node.go
```diff
@@ -415,10 +422,39 @@ func (n *Node) OnStart() error {
...
+		}
+
+		err = n.sw.DialPeersAsync(n.addrBook, persistentPeers, true)
 		if err != nil {
 			return err
 		}
+
+		// if any of the persistent peers are private, do not add them to addrbook
+		if len(persistentAndPrivatePeers) > 0 {
+			err = n.sw.DialPeersAsync(nil, persistentAndPrivatePeers, true)
```
ebuchman: Private Peers are in general validators that we never dial, so doesn't really make sense for it to be persistent (we will fail to ever re-dial them ...).

I don't think there shoudl ever be overlap between persistent and private peers, right ? Unless folks do want to be able to dial their validators directly... hm

rpc/client/localclient.go
```diff
@@ -88,8 +88,8 @@ func (Local) DialSeeds(seeds []string) (*ctypes.ResultDialSeeds, error) {
 	return core.UnsafeDialSeeds(seeds)
 }
 
-func (Local) DialPeers(peers []string, persistent bool) (*ctypes.ResultDialPeers, error) {
-	return core.UnsafeDialPeers(peers, persistent)
+func (Local) DialPeers(peers []string, persistent bool, private bool) (*ctypes.ResultDialPeers, error) {
```
ebuchman: Don't think it makes sense to dial private peers. Is someone asking to be able to do this ?

p2p/pex/pex_reactor.go
```diff
@@ -241,7 +247,10 @@ func (r *PEXReactor) ReceiveAddrs(addrs []*p2p.NetAddress, src Peer) error {
 
 	srcAddr := src.NodeInfo().NetAddress()
 	for _, netAddr := range addrs {
-		if netAddr != nil {
+		if netAddr == nil {
+			continue
+		}
+		if !isAddrPrivate(netAddr, r.config.PrivatePeerIDs) {
```
odeke-em: Minor nit, as the loop doesn't have much in it except for the if clause and `r.book.AddAddress(netAddr, srcAddr)`
```go
if netAddr != nil && !isAddrPrivate(netAddr, r.config.PrivatePeerIDs) {
```

odeke-em:
@melekes rebase needed but I like this PR, great style and approach!

node/node.go
melekes: right, but someone at some point will add persistent peer to private for whatever reason and will create an issue saying "private peers feature does not work"

melekes: ok, I'll remove this code and leave a comment

melekes:
@ebuchman addressed your comments. PTAL

ebuchman:
https://github.com/tendermint/tendermint/commit/d350da313558480201456d580aca3a8d1fb5e6a9

PR #1283, opened 2018-03-06
Speed up CircleCI builds

xla:
To achieve faster feedback cycles for our feature PRs this change reduces the average buildtime from 35 to ~6min by utilising their new 2.0 offering based on docker and nomad. We make use of parallel build steps wherever possible so that the duration is determined by the slowest test suite (p2p).

This is an intermediate step until we move our CI/CD completely on-premise for more control and added security.

***
The workflow in action: https://circleci.com/workflow-run/1500b172-1104-4920-854c-eb68fe7a816f


Makefile
```diff
@@ -40,7 +40,7 @@ check_tools:
 get_tools:
 	@echo "--> Installing tools"
 	go get -u -v $(GOTOOLS)
-	# @gometalinter.v2 --install
+	@$(GOBIN)/gometalinter.v2 --install
```
melekes: I think we had a long discussion about that. I would prefer for user to add `GOPATH/bin` to `PATH` as https://golang.org/doc/install#install suggests.

Example: If I installed `dep` from other source to `/usr/local/bin`, I still should be able to run `make get_vendor_deps`.

xla: I'd support that sentiment, if we would check for existence of the binary in any location. But our entire Makefile infrastructure leverages the go toolchain to manage binaries.

xla: Additionally we shouldn't take any version of the tool, but one that we can control to avoid incompatibilities or other surprises.

melekes: What happens if I don't have `GOBIN`? Should I set it by hands?

```
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/vagrant/go"
GORACE=""
GOROOT="/usr/lib/go-1.9"
GOTOOLDIR="/usr/lib/go-1.9/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build418845291=/tmp/go-build -gno-record-gcc-switches"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
```

xla: Addressed, no more `$GOBIN` in the Makefile.

test/app/test.sh
```diff
@@ -8,6 +8,7 @@ set -e
 
 # TODO: install everything
 
+export PATH="$GOBIN:$PATH"
```
melekes: do we still need this?

PR #1286, opened 2018-03-07
Invert privVal socket communication

xla:
Follow-up to #1255 aligning with the expectation that the external signing process connects to the node. The SocketClient will block on start until one connection has been established, support for multiple signers connected simultaneously is a planned future extension.

* SocketClient accepts connection
* PrivValSocketServer renamed to RemoteSigner
* extend tests

### DIFF
--- a/types/priv_validator/socket.go
+++ b/types/priv_validator/socket.go
@@ -19,12 +19,16 @@ import (
 
 const (
 	defaultConnDeadlineSeconds = 3
-	defaultDialRetryMax        = 10
+	defaultConnWaitSeconds     = 60
+	defaultDialRetries         = 10
+	defaultSignersMax          = 1
 )
 
 // Socket errors.
 var (
-	ErrDialRetryMax = errors.New("Error max client retries")
+	ErrDialRetryMax    = errors.New("Error max client retries")
+	ErrConnWaitTimeout = errors.New("Error waiting for external connection")
+	ErrConnTimeout     = errors.New("Error connection timed out")
 )
 
 var (
@@ -34,10 +38,16 @@ var (
 // SocketClientOption sets an optional parameter on the SocketClient.
 type SocketClientOption func(*SocketClient)
 
-// SocketClientTimeout sets the timeout for connecting to the external socket
-// address.
-func SocketClientTimeout(timeout time.Duration) SocketClientOption {
-	return func(sc *SocketClient) { sc.connectTimeout = timeout }
+// SocketClientConnDeadline sets the read and write deadline for connections
+// from external signing processes.
+func SocketClientConnDeadline(deadline time.Duration) SocketClientOption {
+	return func(sc *SocketClient) { sc.connDeadline = deadline }
+}
+
+// SocketClientConnWait sets the timeout duration before connection of external
+// signing processes are considered to be unsuccessful.
+func SocketClientConnWait(timeout time.Duration) SocketClientOption {
+	return func(sc *SocketClient) { sc.connWaitTimeout = timeout }
 }
 
 // SocketClient implements PrivValidator, it uses a socket to request signatures
@@ -45,11 +55,13 @@ func SocketClientTimeout(timeout time.Duration) SocketClientOption {
 type SocketClient struct {
 	cmn.BaseService
 
-	conn    net.Conn
-	privKey *crypto.PrivKeyEd25519
+	addr            string
+	connDeadline    time.Duration
+	connWaitTimeout time.Duration
+	privKey         *crypto.PrivKeyEd25519
 
-	addr           string
-	connectTimeout time.Duration
+	conn     net.Conn
+	listener net.Listener
 }
 
 // Check that SocketClient implements PrivValidator2.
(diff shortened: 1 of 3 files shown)


types/priv_validator/socket.go
```diff
@@ -62,24 +74,46 @@ func NewSocketClient(
...
 
-	sc.BaseService = *cmn.NewBaseService(logger, "privValidatorSocketClient", sc)
+	sc.BaseService = *cmn.NewBaseService(logger, "SocketClient", sc)
 
 	return sc
 }
 
 // OnStart implements cmn.Service.
 func (sc *SocketClient) OnStart() error {
 	if err := sc.BaseService.OnStart(); err != nil {
```
melekes: BaseService.OnStart is empty and can be removed

types/priv_validator/socket.go
```diff
@@ -195,166 +249,211 @@ func (sc *SocketClient) SignHeartbeat(chainID string, heartbeat *types.Heartbeat
...
 
-func (pvss *PrivValidatorSocketServer) acceptConnections() {
-	for {
-		conn, err := pvss.listener.Accept()
+func (rs *RemoteSigner) connect() (net.Conn, error) {
+	retries := defaultDialRetries
+
+RETRY_LOOP:
+	for retries > 0 {
+		if retries != defaultDialRetries {
```
melekes: I'd add a comment like `sleep if it's not our first retry`

types/priv_validator/socket.go
```diff
@@ -62,24 +74,37 @@ func NewSocketClient(
...
+	sc.BaseService = *cmn.NewBaseService(logger, "SocketClient", sc)
 
 	return sc
 }
 
 // OnStart implements cmn.Service.
 func (sc *SocketClient) OnStart() error {
-	if err := sc.BaseService.OnStart(); err != nil {
-		return err
+	if sc.listener == nil {
```
ebuchman: dont think we need this check.

types/priv_validator/socket.go
```diff
@@ -195,166 +240,206 @@ func (sc *SocketClient) SignHeartbeat(chainID string, heartbeat *types.Heartbeat
...
-				"sc connect",
-				"addr", sc.addr,
-				"err", errors.Wrap(err, "connection failed"),
-			)
+	if err := conn.SetDeadline(time.Now().Add(sc.connDeadline)); err != nil {
+		return nil, err
+	}
 
-			continue RETRY_LOOP
+	if sc.privKey != nil {
```
ebuchman: lets require the privKey always :)

types/priv_validator/socket.go
```diff
@@ -195,166 +240,206 @@ func (sc *SocketClient) SignHeartbeat(chainID string, heartbeat *types.Heartbeat
...
+	)
+
+	go func(connc chan<- net.Conn, errc chan<- error) {
+		conn, err := sc.acceptConnection()
+		if err != nil {
+			errc <- err
+			return
 		}
 
+		connc <- conn
```
ebuchman: should we include a way to exit this routine if the timeout triggerS ?

types/priv_validator/socket.go
```diff
@@ -195,166 +240,206 @@ func (sc *SocketClient) SignHeartbeat(chainID string, heartbeat *types.Heartbeat
...
+	if err := rs.conn.Close(); err != nil {
+		rs.Logger.Error("OnStop", "err", errors.Wrap(err, "closing listener failed"))
 	}
 }
 
-func (pvss *PrivValidatorSocketServer) acceptConnections() {
-	for {
-		conn, err := pvss.listener.Accept()
+func (rs *RemoteSigner) connect() (net.Conn, error) {
+	retries := defaultDialRetries
```
ebuchman: use rs.connRetries

types/priv_validator/socket.go
```diff
@@ -195,166 +240,206 @@ func (sc *SocketClient) SignHeartbeat(chainID string, heartbeat *types.Heartbeat
...
+
+RETRY_LOOP:
+	for retries > 0 {
```
ebuchman: lets put the `retries := ` and the `retries --` in the one liner

types/priv_validator/socket.go
```diff
@@ -195,166 +240,206 @@ func (sc *SocketClient) SignHeartbeat(chainID string, heartbeat *types.Heartbeat
...
+		// Don't sleep if it is the first retry.
+		if retries != defaultDialRetries {
```
ebuchman: rs.connRetries

types/priv_validator/socket.go
```diff
@@ -195,166 +240,206 @@ func (sc *SocketClient) SignHeartbeat(chainID string, heartbeat *types.Heartbeat
...
+			rs.Logger.Error(
+				"connect",
 				"err", errors.Wrap(err, "setting connection timeout failed"),
 			)
 			continue
 		}
 
-		if pvss.privKey != nil {
-			conn, err = p2pconn.MakeSecretConnection(conn, pvss.privKey.Wrap())
+		if rs.privKey != nil {
```
ebuchman: require it

ebuchman: we also probably want to authenticate. ie. we should be dialing a `ID@host:port` and then after the secret connection is setup, check that the ID (ie. hash of the remote pubkey) of the connection matches the ID we dialed (see how p2p does it)

Issue #1125, opened 2018-01-20
p2p: exponential backoff for addrs we dial in ensurePeers


ebuchman:
Currently we keep trying to connect to addrs from the book. 

We should have some exponential backoff policy for addrs that we're failing to connect to in the address book so we dont see eg. `E[01-20|05:34:50.115] Failed to dial peer                          module=p2p address=73.92.247.82:46656 err="Error creating peer: dial tcp 73.92.247.82:46656: getsockopt: connection refused"` every 30s for every address we can dial

PR #1292, opened 2018-03-08
exponential backoff for addrs in the address book

melekes:
Refs #1125

### DIFF
--- a/p2p/pex/pex_reactor.go
+++ b/p2p/pex/pex_reactor.go
@@ -6,6 +6,7 @@ import (
 	"math/rand"
 	"reflect"
 	"sort"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -59,6 +60,8 @@ type PEXReactor struct {
 	// maps to prevent abuse
 	requestsSent         *cmn.CMap // ID->struct{}: unanswered send requests
 	lastReceivedRequests *cmn.CMap // ID->time.Time: last time peer requested from us
+
+	attemptsToDial sync.Map // dial addr -> number of attempts to dial (for exponential backoff)
 }
 
 // PEXReactorConfig holds reactor specific configuration data.
@@ -338,9 +341,22 @@ func (r *PEXReactor) ensurePeers() {
 		toDial[try.ID] = try
 	}
 
+	// 1s == (1e9 ns) == (1 Billion ns)
+	billionNs := float64(time.Second.Nanoseconds())
+
 	// Dial picked addresses
-	for _, item := range toDial {
+	for _, addr := range toDial {
 		go func(picked *p2p.NetAddress) {
+			// exponential backoff if it's not our first attempt to dial picked address
+			var attempt int
+			if lAttempt, notFirstAttempt := r.attemptsToDial.Load(picked.DialString()); notFirstAttempt {
+				attempt = lAttempt.(int)
+				jitterSeconds := time.Duration(rand.Float64() * billionNs)
+				backoffDuration := jitterSeconds + ((1 << uint(attempt)) * time.Second)
+				r.Logger.Debug(fmt.Sprintf("Dialing %v", picked), "attempt", attempt, "backoff_duration", backoffDuration)
+				time.Sleep(backoffDuration)
+			}
+
 			err := r.Switch.DialPeerWithAddress(picked, false)
 			if err != nil {
 				r.Logger.Error("Dialing failed", "err", err)
@@ -350,8 +366,13 @@ func (r *PEXReactor) ensurePeers() {
 				} else {
 					r.book.MarkAttempt(picked)
 				}
+				// record attempt
+				r.attemptsToDial.Store(picked.DialString(), attempt+1)
+			} else {
+				// cleanup any history
+				r.attemptsToDial.Delete(picked.DialString())
 			}
-		}(item)
+		}(addr)
 	}
 
 	// If we need more addresses, pick a random peer and ask for more.
--- a/p2p/pex/pex_reactor.go
+++ b/p2p/pex/pex_reactor.go
@@ -39,6 +39,8 @@ const (
 	defaultSeedDisconnectWaitPeriod = 2 * time.Minute  // disconnect after this
 	defaultCrawlPeerInterval        = 2 * time.Minute  // dont redial for this. TODO: back-off
 	defaultCrawlPeersPeriod         = 30 * time.Second // check some peers every this
+
+	maxAttemptsToDial = 16 // ~ 35h in total (last attempt - 18h sleep)
 )
 
 // PEXReactor handles PEX (peer exchange) and ensures that an
@@ -61,7 +63,7 @@ type PEXReactor struct {
 	requestsSent         *cmn.CMap // ID->struct{}: unanswered send requests
 	lastReceivedRequests *cmn.CMap // ID->time.Time: last time peer requested from us
 
-	attemptsToDial sync.Map // dial addr -> number of attempts to dial (for exponential backoff)
+	attemptsToDial sync.Map // dial address (string) -> number of attempts (int) to dial (for exponential backoff)
 }
 
 // PEXReactorConfig holds reactor specific configuration data.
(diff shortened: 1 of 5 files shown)


p2p/pex/pex_reactor.go
```diff
@@ -338,9 +341,22 @@ func (r *PEXReactor) ensurePeers() {
 		toDial[try.ID] = try
 	}
 
+	// 1s == (1e9 ns) == (1 Billion ns)
+	billionNs := float64(time.Second.Nanoseconds())
+
 	// Dial picked addresses
-	for _, item := range toDial {
+	for _, addr := range toDial {
 		go func(picked *p2p.NetAddress) {
```
xla: With `ensurePeers` growing further maybe we can break this function out and make it a private method on `PEXReactor` as it only seemed to depend on the reactor itself and the `*p2p.NetAddress`.

p2p/pex/pex_reactor.go
```diff
@@ -338,9 +341,22 @@ func (r *PEXReactor) ensurePeers() {
...
+			// exponential backoff if it's not our first attempt to dial picked address
+			var attempt int
+			if lAttempt, notFirstAttempt := r.attemptsToDial.Load(picked.DialString()); notFirstAttempt {
+				attempt = lAttempt.(int)
+				jitterSeconds := time.Duration(rand.Float64() * billionNs)
```
xla: Isn't `billionNs` equivalent to `float64(time.Second)` here?

p2p/pex/pex_reactor.go
```diff
@@ -350,8 +366,13 @@ func (r *PEXReactor) ensurePeers() {
 				} else {
 					r.book.MarkAttempt(picked)
 				}
+				// record attempt
+				r.attemptsToDial.Store(picked.DialString(), attempt+1)
+			} else {
```
xla: This else is unnecessary.

p2p/pex/pex_reactor.go
```diff
@@ -338,9 +341,22 @@ func (r *PEXReactor) ensurePeers() {
...
+	// 1s == (1e9 ns) == (1 Billion ns)
+	billionNs := float64(time.Second.Nanoseconds())
+
 	// Dial picked addresses
-	for _, item := range toDial {
+	for _, addr := range toDial {
 		go func(picked *p2p.NetAddress) {
+			// exponential backoff if it's not our first attempt to dial picked address
+			var attempt int
+			if lAttempt, notFirstAttempt := r.attemptsToDial.Load(picked.DialString()); notFirstAttempt {
```
xla: Some minor naming nits. Would call `attempt`, `attempts` here as we store the amount of times we attempted to connect. And `notFirstAttempt` -> `attempted` for more concise and inverse of the meaning.

p2p/pex/pex_reactor.go
melekes: how so? there is no return or break

xla: You are right. I assumed you break out in the if condition.

p2p/pex/pex_reactor.go
```diff
@@ -339,19 +342,10 @@ func (r *PEXReactor) ensurePeers() {
...
-				// TODO: detect more "bad peer" scenarios
-				if _, ok := err.(p2p.ErrSwitchAuthenticationFailure); ok {
-					r.book.MarkBad(picked)
-				} else {
-					r.book.MarkAttempt(picked)
-				}
-			}
-		}(item)
+			r.dialPeer(picked)
+		}(addr)
```
xla: Can't we call `go r.dialPeer(addr)` directly?

melekes: https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables ?

xla: Here `addr` is passed to `dialPeer` and evaluate at each iteration and placed on the stack of the routine, similar to how you pass it to the clouse at the moment.

melekes: okay.. thanks

p2p/pex/pex_reactor.go
```diff
@@ -372,6 +364,37 @@ func (r *PEXReactor) ensurePeers() {
...
 }
 
+func (r *PEXReactor) dialPeer(addr *p2p.NetAddress) {
+	// 1s == (1e9 ns) == (1 Billion ns)
+	billionNs := float64(time.Second.Nanoseconds())
+
+	// exponential backoff if it's not our first attempt to dial given address
+	var attempts int
+	if lAttempts, attempted := r.attemptsToDial.Load(addr.DialString()); attempted {
+		attempts = lAttempts.(int)
```
xla: Can we test this behaviour? I think this branch isn't covered with our current tests. It might also surface that we want to make the jitter duration configurable in some form.

melekes: How about now? https://github.com/tendermint/tendermint/pull/1292/commits/f85c8896d9877f034e6c40d11887637150940841

xla: Beautiful and simple!

p2p/pex/pex_reactor.go
```diff
@@ -372,6 +364,37 @@ func (r *PEXReactor) ensurePeers() {
...
+		jitterSeconds := time.Duration(rand.Float64() * billionNs)
```
xla: `billionNs` is the same as `float64(time.Second)`, don't think we need that extra variable, but we might benefit from having it configurable.

p2p/pex/pex_reactor.go
```diff
@@ -372,6 +364,37 @@ func (r *PEXReactor) ensurePeers() {
...
+		backoffDuration := jitterSeconds + ((1 << uint(attempts)) * time.Second)
+		r.Logger.Debug(fmt.Sprintf("Dialing %v", addr), "attempts", attempts, "backoff_duration", backoffDuration)
+		time.Sleep(backoffDuration)
```
ebuchman: Do we have logic that limits the number of attempts per address? 

If this backoffDuration gets big, then these go-routines will just hang around for a while. So maybe we should be using a select block here with a ticker so we can break out on Quit ?

melekes:
@ebuchman thanks for suggestion. implemented in https://github.com/tendermint/tendermint/pull/1292/commits/264bce4ddd6c71808a145e02f28b4ea41f8b96a9. PTAL

p2p/pex/pex_reactor.go
```diff
@@ -372,6 +371,48 @@ func (r *PEXReactor) ensurePeers() {
...
 
+func (r *PEXReactor) dialPeer(addr *p2p.NetAddress) {
+	var attempts int
+	var lastDialed time.Time
+	if lAttempts, attempted := r.attemptsToDial.Load(addr.DialString()); attempted {
+		attempts = lAttempts.(_attemptsToDial).number
+		lastDialed = lAttempts.(_attemptsToDial).lastDialed
+	}
+
+	if attempts > maxAttemptsToDial {
```
ebuchman: Looks like we do have a maxAttempts mechanism already in place: https://github.com/tendermint/tendermint/blob/develop/p2p/pex/known_address.go#L130

Maybe we can consolidate to make sure they're working right together ? feel free to leave this an just open an issue to note the redundancy

p2p/pex/pex_reactor.go
```diff
@@ -372,6 +371,48 @@ func (r *PEXReactor) ensurePeers() {
...
+		return
+	}
+
+	// exponential backoff if it's not our first attempt to dial given address
+	if attempts > 0 {
+		jitterSeconds := time.Duration(rand.Float64() * float64(time.Second)) // 1s == (1e9 ns)
+		backoffDuration := jitterSeconds + ((1 << uint(attempts)) * time.Second)
+		sinceLastDialed := time.Since(lastDialed)
+		if sinceLastDialed < backoffDuration {
+			r.Logger.Debug("Too early to dial", "addr", addr, "backoff_duration", backoffDuration, "last_dialed", lastDialed, "time_since", sinceLastDialed)
```
ebuchman: this is cool - it solves the "hanging go-routine" issue but it does mean we might not end up dialing this peer for much longer than the backoff duration. I suppose that's fine

ebuchman:
See https://github.com/tendermint/tendermint/issues/1304

kaycoinsUSA (commented):
Ok

Issue #1299, opened 2018-03-11
types: Header.Hash should first check if it is nil before dereferencing any field


odeke-em:
was just writing tests for lite and given this test case
```go
package proxy

import (
	"bytes"

	"github.com/pkg/errors"

	"github.com/tendermint/tendermint/lite"
	certerr "github.com/tendermint/tendermint/lite/errors"
	"github.com/tendermint/tendermint/types"
)

func ValidateBlockMeta(meta *types.BlockMeta, check lite.Commit) error {
	if meta == nil {
		return errors.New("expecting a non-nil BlockMeta")
	}
	// TODO: check the BlockID??
	return ValidateHeader(meta.Header, check)
}

func ValidateBlock(meta *types.Block, check lite.Commit) error {
	if meta == nil {
		return errors.New("expecting a non-nil Block")
	}
	err := ValidateHeader(meta.Header, check)
	if err != nil {
		return err
	}
	if !bytes.Equal(meta.Data.Hash(), meta.Header.DataHash) {
		return errors.New("Data hash doesn't match header")
	}
	return nil
}

func ValidateHeader(head *types.Header, check lite.Commit) error {
	if head == nil {
		return errors.New("expecting a non-nil Header")
	}
	// make sure they are for the same height (obvious fail)
	if head.Height != check.Height() {
		return certerr.ErrHeightMismatch(head.Height, check.Height())
	}
	// check if they are equal by using hashes
	chead := check.Header
	if !bytes.Equal(head.Hash(), chead.Hash()) {
		return errors.New("Headers don't match")
	}
	return nil
}
```

giving
```shell
$ go test -v -run=Validate
Tendermint running!
=== RUN   TestValidateBlock
--- FAIL: TestValidateBlock (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xb0 pc=0x149333a]

goroutine 30 [running]:
testing.tRunner.func1(0xc0009421e0)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:753 +0x35a
panic(0x16a6600, 0x1ce5680)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/runtime/panic.go:502 +0x210
github.com/tendermint/tendermint/types.(*Header).Hash(0x0, 0x0, 0x0, 0x0)
[...]

Issue #1017, opened 2017-12-28
lite/proxy: low test coverage on develop tip


odeke-em:
I was just reading the code for lite/client to start fuzzing it and noticed that firstly we have 
1.6% test coverage

```shell
 go test -coverprofile=outf.outf
2017/12/27 23:35:59 req.URL: http://0.0.0.0:38233
Tendermint running!
PASS
coverage: 1.6% of statements
E[12-28|06:35:59.225] RPC HTTP server stopped                      module=rpc-server result="accept tcp [::]:38233: use of closed network connection"
ok  	github.com/tendermint/tendermint/lite/proxy	0.026s
```
and am on develop at tip d14ec7d7d2f407433070665536994e0a1eafab4b

ebuchman:
```
ethanbuchman:lite$ go test -cover ./...
ok  	github.com/tendermint/tendermint/lite	2.244s	coverage: 84.1% of statements
ok  	github.com/tendermint/tendermint/lite/client	5.577s	coverage: 79.1% of statements
ok  	github.com/tendermint/tendermint/lite/errors	0.048s	coverage: 37.5% of statements
ok  	github.com/tendermint/tendermint/lite/files	0.093s	coverage: 84.1% of statements
ok  	github.com/tendermint/tendermint/lite/proxy	0.044s	coverage: 16.8% of statements
```

Could always improve but much better

PR #1300, opened 2018-03-11
lite/proxy: Validation* tests and hardening for nil dereferences

odeke-em:
Updates https://github.com/tendermint/tendermint/issues/1017

Ensure that the Validate* functions in proxy are tests
and cover the case of sneakish bugs that have been encountered
a few times from nil dereferences. The lite package should
theoretically never panic with a nil dereference. It is meant
to contain the certifiers hence it should never panic with such.

Requires the following bugs to be fixed first;
* https://github.com/tendermint/tendermint/issues/1298
* https://github.com/tendermint/tendermint/issues/1299

* [X] Wrote tests

### DIFF
--- a/lite/proxy/block.go
+++ b/lite/proxy/block.go
@@ -11,11 +11,17 @@ import (
 )
 
 func ValidateBlockMeta(meta *types.BlockMeta, check lite.Commit) error {
+	if meta == nil {
+		return errors.New("expecting a non-nil BlockMeta")
+	}
 	// TODO: check the BlockID??
 	return ValidateHeader(meta.Header, check)
 }
 
 func ValidateBlock(meta *types.Block, check lite.Commit) error {
+	if meta == nil {
+		return errors.New("expecting a non-nil Block")
+	}
 	err := ValidateHeader(meta.Header, check)
 	if err != nil {
 		return err
@@ -27,6 +33,9 @@ func ValidateBlock(meta *types.Block, check lite.Commit) error {
 }
 
 func ValidateHeader(head *types.Header, check lite.Commit) error {
+	if head == nil {
+		return errors.New("expecting a non-nil Header")
+	}
 	// make sure they are for the same height (obvious fail)
 	if head.Height != check.Height() {
 		return certerr.ErrHeightMismatch(head.Height, check.Height())
--- a/lite/proxy/validate_test.go
+++ b/lite/proxy/validate_test.go
@@ -12,8 +12,8 @@ import (
 )
 
 var (
-	deabBeefTxs           = types.Txs{[]byte("DE"), []byte("AD"), []byte("BE"), []byte("EF")}
-	deadBeefRipEmd160Hash = deabBeefTxs.Hash()
+	deadBeefTxs           = types.Txs{[]byte("DE"), []byte("AD"), []byte("BE"), []byte("EF")}
+	deadBeefRipEmd160Hash = deadBeefTxs.Hash()
 )
 
 func TestValidateBlock(t *testing.T) {
@@ -91,7 +91,7 @@ func TestValidateBlock(t *testing.T) {
 		{
 			block: &types.Block{
 				Header: &types.Header{Height: 11, DataHash: deadBeefRipEmd160Hash},
-				Data:   &types.Data{Txs: deabBeefTxs},
+				Data:   &types.Data{Txs: deadBeefTxs},
 			},
 			commit: lite.Commit{
 				Header: &types.Header{Height: 11},
@@ -101,20 +101,18 @@ func TestValidateBlock(t *testing.T) {
 		// End Header.Data hash mismatch test
 	}
 
-	assert := assert.New(t)
-
 	for i, tt := range tests {
 		err := proxy.ValidateBlock(tt.block, tt.commit)
 		if tt.wantErr != "" {
 			if err == nil {
-				assert.FailNowf("Unexpectedly passed", "#%d", i)
+				assert.FailNowf(t, "Unexpectedly passed", "#%d", i)
 			} else {
-				assert.Contains(err.Error(), tt.wantErr, "#%d should contain the substring\n\n", i)
+				assert.Contains(t, err.Error(), tt.wantErr, "#%d should contain the substring\n\n", i)
 			}
 			continue
 		}
 
-		assert.Nil(err, "#%d: expecting a nil error", i)
+		assert.Nil(t, err, "#%d: expecting a nil error", i)
 	}
 }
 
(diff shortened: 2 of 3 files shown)


lite/proxy/validate_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/tendermint/tendermint/lite"
+	"github.com/tendermint/tendermint/lite/proxy"
+	"github.com/tendermint/tendermint/types"
+)
+
+var (
+	deabBeefTxs           = types.Txs{[]byte("DE"), []byte("AD"), []byte("BE"), []byte("EF")}
```
melekes: typo

lite/proxy/validate_test.go
```diff
@@ -0,0 +1,256 @@
...
+			},
+			commit: lite.Commit{
+				Header: &types.Header{Height: 11},
+				Commit: &types.Commit{BlockID: types.BlockID{Hash: []byte("DEADBEEF")}},
+			},
+		},
+		// End Header.Data hash mismatch test
+	}
+
+	assert := assert.New(t)
```
melekes: we decided not to use this contruction since it messes with linters

lite/proxy/block.go
```diff
@@ -11,11 +11,17 @@ import (
 )
 
 func ValidateBlockMeta(meta *types.BlockMeta, check lite.Commit) error {
+	if meta == nil {
```
melekes: do we have cases where we expect to receive nil blocks/metas?

odeke-em: I was working on tests and a validation case would be to pass in nil. Here we crash. My goal is so that
dependencies of this package won't be surprised. We've had a couple of embarassing cases where there
were nil dereferences and panics from users.

lite/proxy/validate_test.go
odeke-em: Done, thanks!

lite/proxy/validate_test.go
odeke-em: Oh I see, thank you!

lite/proxy/block.go
```diff
@@ -11,11 +11,17 @@ import (
...
+		return errors.New("expecting a non-nil BlockMeta")
```
xla: Would it make sense to consolidate error handling in lite? There is [lite/errors](https://github.com/tendermint/tendermint/tree/master/lite/errors) which is used in this file further down as `certerr` and there is [lite/proxy/errors.go](https://github.com/tendermint/tendermint/blob/master/lite/proxy/errors.go).

A more rigid approach to errors would enable control flow for callers.

lite/proxy/block.go
```diff
@@ -11,11 +11,17 @@ import (
...
+	}
 	// TODO: check the BlockID??
 	return ValidateHeader(meta.Header, check)
 }
 
 func ValidateBlock(meta *types.Block, check lite.Commit) error {
+	if meta == nil {
+		return errors.New("expecting a non-nil Block")
```
xla: Same as above.

lite/proxy/validate_test.go
```diff
@@ -0,0 +1,252 @@
...
+		// End Headers don't match test
+	}
+
+	for i, tt := range tests {
+		err := proxy.ValidateBlockMeta(tt.meta, tt.commit)
+		if tt.wantErr != "" {
+			if err == nil {
+				assert.FailNowf(t, "Unexpectedly passed", "#%d: wanted error %q", i, tt.wantErr)
+			} else {
+				assert.Contains(t, err.Error(), tt.wantErr, "#%d should contain the substring\n\n", i)
```
xla: Having to do substring matching on error output is brittle and not a proper integration point, this is were we would benefit from proper constructed error types that can be tests against properly.

lite/proxy/block.go
```diff
@@ -27,6 +33,9 @@ func ValidateBlock(meta *types.Block, check lite.Commit) error {
 }
 
 func ValidateHeader(head *types.Header, check lite.Commit) error {
```
xla: `check.Header` can also be nil and can lead to the same nil derefence, see: https://circleci.com/gh/tendermint/tendermint/4228

lite/proxy/block.go
odeke-em: @xla 
>Consolidate error handling in lite

Thank you for raising that, I'll do that.

>A more rigid approach to errors would enable control flow for callers
What does this mean?

lite/proxy/block.go
odeke-em: Roger that, thanks.

lite/proxy/block.go
odeke-em: @xla in deed. This PR is awaiting on
https://github.com/tendermint/tendermint/pull/1301 to be merged first
and I mention this in the commit and first Github message :) :
>Requires the following bugs to be fixed first;
>
>#1298
>#1299

xla: My bad! Disregard.

lite/proxy/block.go
xla: Constructing meaningful errors that the caller can vary on will help to make decisions upstream and will lead to a safer way of interacting with the code, e.g.:

``` go
// meta := ...
// commit := ...
err := proxy.ValidateBlockMeta(meta, commit)
switch errors.Cause(err) {
case proxy.ErrNiBlock:
    // reconsile
case proxy.ErrNilHeader:
    // reconsile
case proxy.ErrHeadersNotMatch:
    // reconsile
default:
    // If we really have to propagate: return err
}
```

If the errors should originate from `proxy`, `lite` or `lite/errors` is something we can play with. Usually the more central the less overhead per set of errors to maintain. A mix is also possible: `proxyErr.Err...`, `liteErr.Err...` as long as we consistently wrap and unpack.

xla: It has sentinesl, but unexported and uses a lot of support boilerplate to test for the errors. We should simplify that package and have proper sentinels.

odeke-em:
>The code changes are fine, though as mentioned would be cool to use sentinels, but that can be a new issue.

@ebuchman gotcha, I didn't want to make edits in another package and then get into another pipeline of what errors can we export etc, that's an entirely new discussion. But sure, I've created issue https://github.com/tendermint/tendermint/issues/1360. 

>As for the tests, it would be great if we could reduce the verbosity and replace the substring matching. 

Without exported errors, this is impossible. Also substring matching is used in places where perhaps the error might be informative with dynamic information, or for unexported errors such as in this case, and as I had previously mentioned, that pattern is also heavily used in the standard library for the same reasons -- just in case wondering where else is it used. Perhaps after https://github.com/tendermint/tendermint/issues/1360 is implemented?

>There's a lot of code duplication (eg. the same header is spelled out in full like 4 times, could instead be a single function).

So I'll pushback on this a little, currently those structs are present for easy inspection/inlined for easy comparison along with their expected values. In the past, in some PRs such as https://github.com/tendermint/tendermint/pull/677, I conceeded and extracted those and used a helper function but I still don't like the taste of having to go navigate through each helper to see what the block will contain. I can make the edit, but I'll state that I don't like the outcome of the code, I believe test cases should be simple to read and not require context in trying to examine exactly what they produce.

lite/proxy/block.go
odeke-em: @xla as per discussions below, that's more comprehensive work that'll be tackled in another issue https://github.com/tendermint/tendermint/issues/1360.

lite/proxy/block.go
odeke-em: And #1301 has been merged, so we should no longer have those failures in CircleCI.

odeke-em:
Ping @ebuchman @xla @melekes

PR #1301, opened 2018-03-11
types: Hash invoked for nil Data and Header should not panic

odeke-em:
Fixes https://github.com/tendermint/tendermint/issues/1298
Fixes https://github.com/tendermint/tendermint/issues/1299

Found while writing tests in https://github.com/tendermint/tendermint/pull/1300

* [X] Wrote tests

### DIFF
--- a/types/block.go
+++ b/types/block.go
@@ -179,7 +179,7 @@ type Header struct {
 // Hash returns the hash of the header.
 // Returns nil if ValidatorHash is missing.
 func (h *Header) Hash() cmn.HexBytes {
-	if len(h.ValidatorsHash) == 0 {
+	if h == nil || len(h.ValidatorsHash) == 0 {
 		return nil
 	}
 	return merkle.SimpleHashFromMap(map[string]merkle.Hasher{
@@ -413,6 +413,9 @@ type Data struct {
 
 // Hash returns the hash of the data
 func (data *Data) Hash() cmn.HexBytes {
+	if data == nil {
+		return (Txs{}).Hash()
+	}
 	if data.hash == nil {
 		data.hash = data.Txs.Hash() // NOTE: leaves of merkle tree are TxIDs
 	}
--- a/types/block_test.go
+++ b/types/block_test.go
@@ -3,7 +3,9 @@ package types
 import (
 	"testing"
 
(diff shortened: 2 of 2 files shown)


types/block.go
```diff
@@ -413,6 +413,9 @@ type Data struct {
 
 // Hash returns the hash of the data
 func (data *Data) Hash() cmn.HexBytes {
+	if data == nil {
```
melekes: I'd argue that such things lead to "hard to debug" errors. One day we'll have a bug with the title "Don't understand why tx hash is empty." Instead of adding `X == nil` everywhere, I think we should locate/change the logic in places where `nil` might come from (basically scan every usage).

odeke-em: Being this close to launch, or even for that matter at anytime, I'd rather avoid nil panics, moreover for untested paths than an endless debug. There have been many runtime panics reported by users and at times @ebuchman has wished we'd write everything in Rust to avoid such.

srmo: Or at least enhance the comment: "Hash returns the hash of the data. If given data was nil it foobarblub"

xla: With @melekes here, if `Hash()` can't satisfy it's contract it should make this explicit and return an additional `error` which can be used by the caller for control flow.

melekes:
https://github.com/tendermint/tendermint/pull/1301#discussion_r173651228
https://github.com/tendermint/tendermint/pull/1301#discussion_r173654820

I think **by default** all our `(* XYZ)` functions should have an unspoken? agreement that `it may crash if the object is nil`. If dev knows that object can be nil at some point (or under some condition), he/she should use https://en.wikipedia.org/wiki/Null_object_pattern (btw, I did not find any article with Golang examples, so it may be a good idea for a post)

types/block.go
ebuchman: Let's delay launch until we re-write the whole thing in rust :stuck_out_tongue_winking_eye:

The `block.Hash` already has these nil checks. I agree it'd be nice not to have to do them, but if we're going to do some, we should probably do them everywhere (in this case, `Commit.Hash` included!).

That said, we should certainly do a better job of clearly articulating/noting under what circumstances these values can be nil.

Issue #1290, opened 2018-03-08
tendermint testnet fail


pj0579:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
bug

**Tendermint version**: 0.16.0
**ABCI app** :0.9.0
**Environment**:
- **OS** macOs13.0.0
- **Install tools**:glide
- **Others**:

**What happened**:
tendermint testnet
panic: Panicked on a Crisis: open mytestnet/mach0/config/080033051: no such file or directory

**What you expected to happen**:
4 nodes

**How to reproduce it** (as minimally and precisely as possible):

**Anything else do we need to know**:
other commands is all right（tendetmint node，tendermint init and so on）

melekes:
1) Replace `--p2p.seeds=$SEEDS; \` with `--p2p.persistent_peers=$SEEDS; \`
2) Looks like you don't have genesis file. Please make sure you have a genesis file in every directory
3) Don't forget to create dir for tendermint `mkdir -p "$TESTNET_ROOT_DIR/$TESTNET_FOLDER/$NODE_0_NAME"`
4) Use `TMHOME` env variable instead of `--home`

pj0579:
thx,  but <code>tendermint testnet</code>  will make four genesis files and four folders actually.

pj0579:
<code>tendermint testnet</code>   need  to add a genesis file by myself?

ArvonHuang:
Hi, have you fixed this problem? I got the same problem as you .

ArvonHuang:
`go get tendermint `

odeke-em:
I can reproduce this on develop just by running
```shell
$ tendermint testnet
```

In the meantime perhaps we need a better error message

odeke-em:
I've mailed out https://github.com/tendermint/tendermint/pull/1303, this'll be fixed in the next release. Thank you very much @pj0579 for reporting this!

odeke-em:
@pj0579 thank you for the report, we've fixed this in PR https://github.com/tendermint/tendermint/pull/1318 which is currently on develop and perhaps might be merged soon to master.

ebuchman:
Thanks @odeke-em - closing as it's on develop. Hopefully will cut a release to master this week :)

PR #1303, opened 2018-03-11
types: PrivValidator.Save should first invoke os.MkdirAll

odeke-em:
Fixes https://github.com/tendermint/tendermint/issues/1290

We invoked cmn.WriteFileAtomic without first checking if
the parent directory existed yet cmn.WriteFileAtomic doesn't
do it first.

This change ensures that we invoke os.MkdirAll first.

With it, we now get
```shell
tendermint testnet
Successfully initialized 4 node directories
```
instead of panicking.

* [X] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/types/priv_validator.go
+++ b/types/priv_validator.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"os"
+	"path/filepath"
 	"sync"
 	"time"
 
@@ -225,7 +227,13 @@ func (privVal *PrivValidatorFS) save() {
 		// `@; BOOM!!!
 		cmn.PanicCrisis(err)
 	}
-	err = cmn.WriteFileAtomic(privVal.filePath, jsonBytes, 0600)
+	outFile := privVal.filePath
+	dir := filepath.Dir(outFile)
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		cmn.PanicCrisis(err)
+	}
+	// We need to mkdirAll the path
+	err = cmn.WriteFileAtomic(outFile, jsonBytes, 0600)
 	if err != nil {
 		// `@; BOOM!!!
 		cmn.PanicCrisis(err)
--- a/types/priv_validator.go
+++ b/types/priv_validator.go
@@ -220,24 +220,22 @@ func (privVal *PrivValidatorFS) Save() {
 
 func (privVal *PrivValidatorFS) save() {
 	if privVal.filePath == "" {
-		cmn.PanicSanity("Cannot save PrivValidator: filePath not set")
+		panic("Cannot save PrivValidator: filePath not set")
 	}
 	jsonBytes, err := json.Marshal(privVal)
 	if err != nil {
-		// `@; BOOM!!!
-		cmn.PanicCrisis(err)
+		panic(err)
 	}
 	outFile := privVal.filePath
 	// We need to ensure that the parent directory exists.
 	// See Issue https://github.com/tendermint/tendermint/issues/1290
 	dir := filepath.Dir(outFile)
 	if err := os.MkdirAll(dir, 0700); err != nil {
-		cmn.PanicCrisis(err)
+		panic(err)
 	}
 	err = cmn.WriteFileAtomic(outFile, jsonBytes, 0600)
 	if err != nil {
-		// `@; BOOM!!!
-		cmn.PanicCrisis(err)
+		panic(err)
 	}
 }
 
--- a/types/priv_validator.go
+++ b/types/priv_validator.go
@@ -228,11 +228,12 @@ func (privVal *PrivValidatorFS) save() {
 		cmn.PanicCrisis(err)
 	}
 	outFile := privVal.filePath
+	// We need to ensure that the parent directory exists.
+	// See Issue https://github.com/tendermint/tendermint/issues/1290
 	dir := filepath.Dir(outFile)
-	if err := os.MkdirAll(dir, 0755); err != nil {
+	if err := os.MkdirAll(dir, 0700); err != nil {
 		cmn.PanicCrisis(err)
 	}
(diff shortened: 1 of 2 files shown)


types/priv_validator.go
```diff
@@ -225,7 +227,13 @@ func (privVal *PrivValidatorFS) save() {
 		// `@; BOOM!!!
 		cmn.PanicCrisis(err)
 	}
-	err = cmn.WriteFileAtomic(privVal.filePath, jsonBytes, 0600)
+	outFile := privVal.filePath
+	dir := filepath.Dir(outFile)
+	if err := os.MkdirAll(dir, 0755); err != nil {
```
melekes: I think permission should also be 0600

odeke-em: 0600 will only allow for Read and Writes for the owner of that directory, if we opened up another terminal and tried to get content out of that directory, we'd get "Permission denied" as 6 for the owner would mean they can't invoke execute which seems to be needed for MkdirAll, let's meet in the middle, how about `0700`? Still then I even wonder if other users in the same group will be able to use the file as they'll have `0` for their access

types/priv_validator.go
```diff
@@ -225,7 +227,14 @@ func (privVal *PrivValidatorFS) save() {
 		// `@; BOOM!!!
 		cmn.PanicCrisis(err)
 	}
-	err = cmn.WriteFileAtomic(privVal.filePath, jsonBytes, 0600)
+	outFile := privVal.filePath
+	// We need to ensure that the parent directory exists.
+	// See Issue https://github.com/tendermint/tendermint/issues/1290
+	dir := filepath.Dir(outFile)
+	if err := os.MkdirAll(dir, 0700); err != nil {
+		cmn.PanicCrisis(err)
```
xla: AFAIK `Panic` methods are deprecated, we use blank `panic`s now.

types/priv_validator_test.go
```diff
@@ -253,3 +255,15 @@ func clipToMS(t time.Time) time.Time {
...
+// See Issue https://github.com/tendermint/tendermint/issues/1290
+func TestPrivValidatorFSSaveDoesNotCrash(t *testing.T) {
+	saveDir, err := ioutil.TempDir("", "priv-validator")
+	require.Nil(t, err, "should have created the tempdir")
+
+	defer os.RemoveAll(saveDir)
+
+	pPath := filepath.Join(saveDir, "out-file", "tf.json")
+	pv := &PrivValidatorFS{filePath: pPath}
+	pv.Save()
```
xla: While it is a regression test, the behaviour we want to ensure is that the file is saved even in nested, non-existing directory structure. Would rename the test accordingly and test that the file with the right content is created for extra verification.

types/priv_validator.go
```diff
@@ -225,7 +227,14 @@ func (privVal *PrivValidatorFS) save() {
 		// `@; BOOM!!!
 		cmn.PanicCrisis(err)
 	}
-	err = cmn.WriteFileAtomic(privVal.filePath, jsonBytes, 0600)
+	outFile := privVal.filePath
+	// We need to ensure that the parent directory exists.
+	// See Issue https://github.com/tendermint/tendermint/issues/1290
+	dir := filepath.Dir(outFile)
+	if err := os.MkdirAll(dir, 0700); err != nil {
```
melekes: Should we move this into `cmn.WriteFileAtomic`? The flow is

1 <...> - 2 <priv_validator.go> - 3 <common.go>

So we should either create the directory at (1) or (3). (2) feels like a halfway.

xla: :+1:

odeke-em: Great point @melekes, in that case I'd rather do it in 3) cmn.WriteFileAtomic. I didn't do so initially because I assumed changing the implementation of cmn.WriteFileAtomic was off-limits.

types/priv_validator.go
odeke-em: I see, thank you. I was using the surrounding code as the guide. However, I'll change that up.

types/priv_validator_test.go
odeke-em: Gotcha, thanks for that @xla, in deed that adds extra verification. In the past I've had pushback whenever I've pushed up rigorous tests, but am glad to get support for such, from you so I'll implement it :)

types/priv_validator.go
odeke-em: Okay now I recall why I put this here, doing MkdirAll in cmn.WriteFileAtomic is quite complex:
Atomic meaning "all or nothing" would require that I have to check the:
a) The least none existent path e.g if we have a/b/c/d/e/f/g/txt.json, if directory b doesn't exist, I'll have to find out firstly what doesn't exist to clean it up if the operation fails
b) Remove the least existent path

I am not sure I'd want to be introducing such  a change in common at a time like this close to launch and moreover when my dev bandwidth is running out. For those reasons, I'll leave this here, someone else can
later move that check into WriteFileAtomic since that had never come up anyways.

types/priv_validator_test.go
odeke-em: Great stuff @xla recommending the rigorous validation, me likey!

odeke-em:
Thank you @melekes and @xla for the reviews, I've addressed all the feedback except for moving the os.MkdirAll out of where it is currently, because, I believe that doing so will expend currently unavailable dev cycles.

xla:
Rebased on latest `develop`.

types/priv_validator.go
melekes: ok, I tend to agree

ebuchman:
Thanks folks but the solution is much simpler: https://github.com/tendermint/tendermint/pull/1318

PrivVal can just require that the folder is already set up - it's the callers duty to initialize. 

This failure is a vestige of the introduction of the config/data folders and the lack of testing the `testnet` command ...

Issue #1296, opened 2018-03-09
Add RPC endpoint for checking health


melekes:
`/health`, which returns 200 OK if this instance is making blocks, otherwise - 500.

melekes:
go ahead, sure!

nickray:
I don't think (longterm) that a health endpoint should check whether blocks are made, since that concerns the health of the chain and not the node. But great to implement to endpoint!

srmo:
Is it possible to give an additional state for "making progress but behind chain"?

r8d8:
It's just health check, only 2 states - `healthy/unhealthy`
There is `\status` endpoint - should be enough for more detailed inspection, or @srmo  I am getting you wrong?

melekes:
Fixed a few thing in https://github.com/tendermint/tendermint/pull/1337

I will leave this issue open because I think we still want to implement @srmo request (showing whenever the node lags behind the rest of the network) based on other peers heights (peerState) and this node height (blockStoreHeight). For that we'll need to add config variable (UnhealthyHeightLag)

ebuchman:
Can we consolidate https://github.com/tendermint/tendermint/issues/726 into here ? 

Lets have the /health endpoint return our height and the height for all our peers. each peer can have its nodeID and moniker. it can also have its validator address once we do #1516

melekes:
> Lets have the /health endpoint return our height and the height for all our peers. each peer can have its nodeID and moniker. it can also have its validator address once we do #1516 

Don't think we should duplicate info. We have other endpoints for that. `/health` needs to be lightweight and only report that something is wrong (node lags behind). then a person will go and use `/net_info` and other RPC endpoints to figure stuff out.

zramsay:
`/health` exists. open a new issue if it needs new features/modification

PR #1306, opened 2018-03-12
Add `\health` rpc endpoint

r8d8:
Related to #1296
* [x] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -454,6 +454,7 @@ func (n *Node) ConfigureRPC() {
 	rpccore.SetConsensusReactor(n.consensusReactor)
 	rpccore.SetEventBus(n.eventBus)
 	rpccore.SetLogger(n.Logger.With("module", "rpc"))
+	rpccore.SetConfig(n.config)
 }
 
 func (n *Node) startRPC() ([]net.Listener, error) {
--- a/node/node.go
+++ b/node/node.go
@@ -454,7 +454,6 @@ func (n *Node) ConfigureRPC() {
 	rpccore.SetConsensusReactor(n.consensusReactor)
 	rpccore.SetEventBus(n.eventBus)
 	rpccore.SetLogger(n.Logger.With("module", "rpc"))
-	rpccore.SetConfig(n.config)
 }
 
 func (n *Node) startRPC() ([]net.Listener, error) {
--- a/rpc/core/health.go
+++ b/rpc/core/health.go
@@ -1,12 +1,7 @@
 package core
 
 import (
-	"time"
-
-	"errors"
-	"fmt"
 	ctypes "github.com/tendermint/tendermint/rpc/core/types"
-	"github.com/tendermint/tendermint/types"
 )
 
 // Get node health. Checks whether new blocks are created.
--- a/rpc/core/health.go
+++ b/rpc/core/health.go
@@ -1,84 +1,50 @@
 package core
 
 import (
-"time"
+	"time"
 
-ctypes "github.com/tendermint/tendermint/rpc/core/types"
-"github.com/tendermint/tendermint/types"
-cmn "github.com/tendermint/tmlibs/common"
+	"errors"
+	"fmt"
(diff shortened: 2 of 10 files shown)


rpc/core/routes.go
```diff
@@ -12,6 +12,7 @@ var Routes = map[string]*rpc.RPCFunc{
 	"unsubscribe_all": rpc.NewWSRPCFunc(UnsubscribeAll, ""),
 
 	// info API
+	"health": 				rpc.NewRPCFunc(Health, ""),
```
melekes: use `go fmt` to format your code or install https://www.jetbrains.com/go/

r8d8: yeap sure, it's still `work in progress`

rpc/core/health.go
```diff
@@ -0,0 +1,84 @@
...
+	if latestHeight != 0 {
+		latestBlockMeta = blockStore.LoadBlockMeta(latestHeight)
+		latestBlockHash = latestBlockMeta.BlockID.Hash
+		latestAppHash = latestBlockMeta.Header.AppHash
+		latestBlockTimeNano = latestBlockMeta.Header.Time.UnixNano()
+	}
+
+	latestBlockTime := time.Unix(0, latestBlockTimeNano)
+
+	return &ctypes.ResultStatus{
```
melekes: height is not supposed to return any result. only status code `200`

```
ctypes.ResultHealth{}, nil
```

xla: What is are the drawbacks from returning additional information? The consumer can decide if to discard the info or not.

r8d8: @xla, it's plain `health` check, just binary state. 
It 's WIP, I will remove unnecessary info soon.

melekes: I think it should be lightweight. The primary purpose of the health check is to give status (working / not working), not provide information (no?). If you want info, there are other API endpoints.

r8d8:
ping @melekes

rpc/core/health.go
```diff
@@ -0,0 +1,51 @@
...
+		latestBlockMeta = blockStore.LoadBlockMeta(latestHeight)
+	}
+
+	latestBlockTimeout := cfg.Consensus.Commit(latestBlockMeta.Header.Time).UnixNano()
+	currentTime := time.Now().UnixNano()
+	blockDelayTime := time.Duration(latestBlockTimeout-currentTime) * time.Nanosecond
+
+	// use single commit timeout as threshold for block creation
+	threshold := time.Duration(cfg.Consensus.TimeoutCommit) * time.Millisecond
+	if blockDelayTime > threshold {
```
melekes: It's a bit more complex than that. If `no_empty_blocks` set to `true`, then this endpoint will report `Unhealthy` state, which is wrong.

I think we should for now just return `return &ctypes.ResultHealth{}, nil`

Later, we can implement @srmo request based on other peers heights (peerState) and this node height (blockStoreHeight). For that we'll need to add config variable (`UnhealthyHeightLag`)

r8d8: @melekes , remove whole check at all? 
I am a bit confused about, how we will know that blocks are been created and node is healthy. So, just return `return &ctypes.ResultHealth{}, nil`, and this will show that node hasn't crashed and still running ?

r8d8: Removed

melekes: yeah, 200 OK response should inform a client that a node is running at the very least.

node/node.go
```diff
@@ -454,6 +454,7 @@ func (n *Node) ConfigureRPC() {
 	rpccore.SetConsensusReactor(n.consensusReactor)
 	rpccore.SetEventBus(n.eventBus)
 	rpccore.SetLogger(n.Logger.With("module", "rpc"))
+	rpccore.SetConfig(n.config)
```
melekes: this is no longer needed

r8d8:
@melekes, looks good to you?

PR #1312, opened 2018-03-14
Net_info should print the ID of peers

zmanian:
It is extremely helpful for network debugging to cross-reference between /net_info and dump_consensus_state. We switched from using ip addresses to describe peers to ID()s but we should have the necessary info the responses to cross-reference

### DIFF
--- a/rpc/core/types/responses.go
+++ b/rpc/core/types/responses.go
@@ -93,6 +93,7 @@ type ResultDialPeers struct {
 
 type Peer struct {
 	p2p.NodeInfo     `json:"node_info"`
+	p2p.ID           `json:"node_id"`
 	IsOutbound       bool                 `json:"is_outbound"`
 	ConnectionStatus p2p.ConnectionStatus `json:"connection_status"`
 }
--- a/rpc/core/net.go
+++ b/rpc/core/net.go
@@ -42,6 +42,7 @@ func NetInfo() (*ctypes.ResultNetInfo, error) {
 	for _, peer := range p2pSwitch.Peers().List() {
 		peers = append(peers, ctypes.Peer{
 			NodeInfo:         peer.NodeInfo(),
+			ID:               peer.ID(),
 			IsOutbound:       peer.IsOutbound(),
 			ConnectionStatus: peer.Status(),
 		})


rpc/core/types/responses.go
```diff
@@ -93,6 +93,7 @@ type ResultDialPeers struct {
 
 type Peer struct {
 	p2p.NodeInfo     `json:"node_info"`
+	p2p.ID           `json:"node_id"`
```
melekes: why not just `id`?

melekes:
```
grouped write of manifest, lock and vendor: error while writing out vendor tree: failed to write dep tree: failed to export github.com/tendermint/go-wire: fatal: failed to unpack tree object fa721242b042ecd4c6ed1a934ee740db4f74e45c
: exit status 128
make: *** [get_vendor_deps] Error 1
```

I think we might have broke dep.

rpc/core/types/responses.go
zmanian: no particular reason?

xla:
> I think we might have broke dep.

@melekes Might it be related to the rename of the repo?

ebuchman:
Need to add go-amino as the source for the go-wire import

melekes:
I am gonna do this right now

ebuchman:
Oh - can we just replace NodeInfo.PubKey with the ID ? It doesn't look like we need or use the ID from there for anything. Or is there a reason we might want that PubKey ? We can always get it again by reflecting on the SecretConn

melekes:
https://github.com/tendermint/tendermint/pull/1314/files

zmanian:
So when I was looking through this, it looks like NodeKey and validatorkeys have nothing to do with each other now?

So we only need to refer valdiators by ID pervasively

xla:
@zmanian This needs to be rebased on latest `develop` in order to resolve the build erorrs. :>

ebuchman:
> So when I was looking through this, it looks like NodeKey and validatorkeys have nothing to do with each other now?

This is correct. But one challenge is we don't associate the two keys. At the peer layer, its all PeerID, but often you want to know which peer corresponds to which validator, something we don't support well yet. 

I'm not 100% what the best way to do this is (especially because it's not relevant to the production net where we want to hide this info explicitly) so maybe someone can help with a proposal :)

Issue #1307, opened 2018-03-12
privVal socket follow up


ebuchman:
Minor points from review: https://github.com/tendermint/tendermint/pull/1286#pullrequestreview-103093019

ebuchman:
Also https://github.com/tendermint/tendermint/issues/1255

ebuchman:
Closed for https://github.com/tendermint/tendermint/issues/1255

PR #1315, opened 2018-03-14
privVal: Improve SocketClient network code

xla:
Follow-up to feedback from #1286, this change simplifies the connection handling in the SocketClient and makes the communication via TCP more robust. It introduces the tcpTimeoutListener to encapsulate accept and i/o timeout handling as well as connection keep-alive, this type could likely be upgraded to handle more fine-grained tuning of the tcp stack (linger, nodelay, etc.) according to the properties we desire. The same methods should be applied to the RemoteSigner which will be overhauled when the priv_val_server is fleshed out.

* require private key
* simplify connect logic
* break out conn upgrades to `tcpTimeoutListener`
* extend test coverage and simplify component setup

***

Belongs to #1255 with the original issue being #1307

### DIFF
--- a/types/priv_validator/socket.go
+++ b/types/priv_validator/socket.go
@@ -11,37 +11,51 @@ import (
 	wire "github.com/tendermint/go-wire"
 	cmn "github.com/tendermint/tmlibs/common"
 	"github.com/tendermint/tmlibs/log"
-	"golang.org/x/net/netutil"
 
 	p2pconn "github.com/tendermint/tendermint/p2p/conn"
 	"github.com/tendermint/tendermint/types"
 )
 
 const (
-	defaultConnDeadlineSeconds = 3
-	defaultConnWaitSeconds     = 60
-	defaultDialRetries         = 10
-	defaultSignersMax          = 1
+	defaultAcceptDeadlineSeconds = 3
+	defaultConnDeadlineSeconds   = 3
+	defaultConnHeartBeatSeconds  = 30
+	defaultConnWaitSeconds       = 60
+	defaultDialRetries           = 10
 )
 
 // Socket errors.
 var (
-	ErrDialRetryMax    = errors.New("Error max client retries")
-	ErrConnWaitTimeout = errors.New("Error waiting for external connection")
-	ErrConnTimeout     = errors.New("Error connection timed out")
+	ErrDialRetryMax    = errors.New("dialed maximum retries")
+	ErrConnWaitTimeout = errors.New("waited for remote signer for too long")
+	ErrConnTimeout     = errors.New("remote signer timed out")
 )
 
 var (
-	connDeadline = time.Second * defaultConnDeadlineSeconds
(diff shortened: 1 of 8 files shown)


types/priv_validator/socket.go
```diff
@@ -240,6 +207,55 @@ func (sc *SocketClient) SignHeartbeat(
...
+	}
+
+	sc.conn = conn
+
+	return nil
+}
+
+// OnStop implements cmn.Service.
+func (sc *SocketClient) OnStop() {
+	sc.BaseService.OnStop()
```
melekes: can be removed too

types/priv_validator/socket.go
```diff
@@ -11,39 +11,53 @@ import (
...
+	defaultDialRetries           = 10
 )
 
 // Socket errors.
 var (
-	ErrDialRetryMax    = errors.New("Error max client retries")
-	ErrConnWaitTimeout = errors.New("Error waiting for external connection")
-	ErrConnTimeout     = errors.New("Error connection timed out")
+	ErrDialRetryMax    = errors.New("dialed maximum retries")
+	ErrConnWaitTimeout = errors.New("waited for remote signer timed out")
```
melekes: maybe "waited for remote signer for too long" or smth else?

types/priv_validator/socket_tcp.go
```diff
@@ -0,0 +1,66 @@
...
+
+// timeoutError can be used to check if an error returned from the netp package
+// was due to a timeout.
+type timeoutError interface {
+	Timeout() bool
+}
+
+// tcpTimeoutListener wraps a *net.TCPListener to standardise protocol timeouts
+// and potentially other tuning parameters.
+// Implements net.Listener.
```
melekes: you can check that in compile time `var _ net.Listener = (*tcpTimeoutListener)(nil)`

types/priv_validator/socket_tcp.go
```diff
@@ -0,0 +1,66 @@
...
+type tcpTimeoutListener struct {
+	*net.TCPListener
+
+	acceptDeadline time.Duration
+	connDeadline   time.Duration
+	period         time.Duration
+}
+
+// tcpTimeoutListener wraps a *net.TCPListener to standardise protocol timeouts
```
melekes: why duplicate comment?

types/priv_validator/socket_test.go
```diff
@@ -90,137 +89,196 @@ func TestSocketClientVote(t *testing.T) {
...
+	SocketClientConnWait(500 * time.Millisecond)(sc)
 
 	go func(sc *SocketClient) {
-		require.NoError(sc.Start())
-		assert.True(sc.IsRunning())
+		defer close(listenc)
 
-		readyc <- struct{}{}
+		require.NoError(t, sc.Start())
+		defer sc.Stop()
```
xla: remove this and add a sleep somewhere

PR #1321, opened 2018-03-18
p2p: NodeInfo.Channels is now common.HexBytes instead of []byte

### LINKED ISSUE #1310: p2p: Channels are JSON encoded as base64 byte string
Channels are `[]byte` because they're literally an array of indivdiual bytes but we want to see the numbers themselves when json encoded not the garbage base64 string.

So maybe we just need to change to `[]int8` or even `[]int` ... 

Ie. see the /status output: `      "channels": "QCAhIiMwOAA=",`

odeke-em:
Fixes #1310

NodeInfo.Channels is now of type common.HexBytes instead of []byte.
[]byte when JSON encoded as represented as base64 bytes. However
base64 hurts readability/inspection. Therefore use common.HexBytes
which can still be assigned to []byte but are encoded as hex bytes e.g
```go
ni.Channels = []byte("foo-bar")
```

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/p2p/node_info.go
+++ b/p2p/node_info.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	crypto "github.com/tendermint/go-crypto"
+	"github.com/tendermint/tmlibs/common"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


ebuchman:
Thanks @odeke-em but this isn't quite what we want.

The `Channels` field does not contain something that should be read as a byte array. It really is a list of independent bytes, and we want to see them like that, ie. `[1, 4, 6, 10]`. 

So we should probably just change it to be a `[]uint8` as discussed in https://github.com/tendermint/tendermint/issues/1310#issue-304736599

odeke-em:
Gotcha gotcha, thanks for the insights @ebuchman, I'll close this.

Issue #1317, opened 2018-03-15
consensus: fix tracking for MarkGood


ebuchman:
As of https://github.com/tendermint/tendermint/pull/1244 we mark peers as good, but the logic there has a bug where peers can send old information that is counted as new. We should only accept new info. 

Also see the other comments from review: https://github.com/tendermint/tendermint/pull/1244#pullrequestreview-104307402
- [x] fix tracking of votes and blockparts to not allow old information
- [x] test the functionality to prove the above is an issue and that it's now fixed
- [x] dont use a magic number for how many blocks we need info from before marking peer as good
- [x] some other things :)

ebuchman:
I'm also worried about changing the panic to a onPeerError here: https://github.com/tendermint/tendermint/pull/1244/files#diff-c46d0aa05757707b7d64e511fbcbe7f5R361

Note that its ok to panic in the onReceive as it's caught in the conn._recover, which does onPeerError.

It might be fine, but we should add a test.

ebuchman:
Booya!

PR #1333, opened 2018-03-19
consensus: fix tracking for MarkGood

melekes:
Refs #1317

### DIFF
--- a/blockchain/reactor.go
+++ b/blockchain/reactor.go
@@ -76,9 +76,9 @@ func NewBlockchainReactor(state sm.State, blockExec *sm.BlockExecutor, store *Bl
 			store.Height()))
 	}
 
-	const cap = 1000 // must be bigger than peers count
-	requestsCh := make(chan BlockRequest, cap)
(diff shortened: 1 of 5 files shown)


melekes:
Should we mark peer as good only once https://github.com/tendermint/tendermint/pull/1333/commits/31f3dd42e7c420ae6a018c7b9a965f22c5736186 or say every 10000 blocks?

ebuchman:
> Should we mark peer as good only once 31f3dd4 or say every 10000 blocks?

I think we need to keep doing it. If enough peers are marked good eventually some will become unmarked, so good to have a force that will continue to cycle them back into good territory!

ebuchman:
Awesome thanks for the quick turn around. I'll open a new issue for the "repeated MarkGood" but let's punt it to the next iteration :)

ebuchman:
lol darn it @melekes that was fast :P 

I was about to merge but I can wait for the new CI run 💃

ebuchman:
Np. PersistentPeer test failed but @xla is working on it now. 

Will merge this. Thanks!

PR #1336, opened 2018-03-19
Add light client spec

milosevic:

### DIFF
--- a/docs/specification/new-spec/light-client.md
+++ b/docs/specification/new-spec/light-client.md
@@ -1,0 +1,114 @@
+# Light client
+
+A light client is a process that connects to the Tendermint Full Node(s) and then tries to verify the Merkle proofs 
+about the blockchain application. In this document we describe mechanisms that ensures that the Tendermint light client 
+has the same level of security as Full Node processes (without being itself a Full Node). 
+
+To be able to validate a Merkle proof, a light client needs to validate the blockchain header that contains the root app hash. 
+Validating a blockchain header in Tendermint consists in verifying that the header is committed (signed) by >2/3 of the 
+voting power of the corresponding validator set. As the validator set is a dynamic set (it is changing), one of the 
+core functionality of the light client is updating the current validator set, that is then used to verify the 
+blockchain header, and further the corresponding Merkle proofs. 
+
+For the purpose of this light client specification, we assume that the Tendermint Full Node exposes the following functions over
+Tendermint RPC:
+
+```golang
+Header(height int64) (SignedHeader, error) // returns signed header for the given height
+Validators(height int64) (ResultValidators, error) // returns validator set for the given height
+LastHeader(valSetNumber int64) (SignedHeader, error)   // returns last header signed by the validator set with the given validator set number
+
+type SignedHeader struct {
+    Header        Header 
+    Commit        Commit
+    ValSetNumber  int64 
+}
+
+type ResultValidators struct {
+    BlockHeight   int64              
+    Validators    []Validator 
+    // time the current validator set is initialised, i.e, time of the last validator change before header BlockHeight 
+    ValSetTime    int64               
+}
+```
+
+We assume that Tendermint keeps track of the validator set changes and that each time a validator set is changed it is 
+being assigned the next sequence number. We can call this number the validator set sequence number. Tendermint also remembers 
+the Time from the header when the next validator set is initialised (starts to be in power), and we refer to this time
+as validator set init time.
+Furthermore, we assume that each validator set change is signed (committed) by the current validator set. More precisely,
+given a block `H` that contains transactions that are modifying the current validator set, the Merkle root hash of the next 
+validator set (modified based on transactions from block H) will be in block `H+1` (and signed by the current validator 
+set), and then starting from the block `H+2`, it will be signed by the next validator set.    
+
+Note that the real Tendermint RPC API is slightly different (for example, response messages contain more data and function 
+names are slightly different); we shortened (and modified) it for the purpose of this document to make the spec more 
+clear and simple. Furthermore, note that in case of the third function, the returned header has `ValSetNumber` equals to 
+`valSetNumber+1`.    
+
+
+Locally, light client manages the following state:
+
+```golang
+valSet        []Validator    // current validator set (last known and verified validator set) 
+valSetNumber  int64          // sequence number of the current validator set 
+valSetHash    []byte         // hash of the current validator set
+valSetTime    int64          // time when the current validator set is initialised  
+```
+
+The light client is initialised with the trusted validator set, for example based on the known validator set hash,
+validator set sequence number and the validator set init time.
+The core of the light client logic is captured by the VerifyAndUpdate function that is used to 1) verify if the given header is valid,
+and 2) update the validator set (when the given header is valid and it is more recent than the seen headers). 
+
+```golang
+VerifyAndUpdate(signedHeader SignedHeader):
+  assertThat signedHeader.valSetNumber >= valSetNumber  
+  if isValid(signedHeader) and signedHeader.Header.Time <= valSetTime + UNBONDING_PERIOD then
+    setValidatorSet(signedHeader)
+    return true
+  else
+    updateValidatorSet(signedHeader.ValSetNumber)
+    return VerifyAndUpdate(signedHeader)
+
+isValid(signedHeader SignedHeader):
+  valSetOfTheHeader = Validators(signedHeader.Header.Height)
+  assertThat Hash(valSetOfTheHeader) == signedHeader.Header.ValSetHash
+  assertThat signedHeader is passing basic validation
+  if votingPower(signedHeader.Commit) > 2/3 * votingPower(valSetOfTheHeader) then return true
+  else 
+    return false
+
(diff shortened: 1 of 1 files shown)


docs/specification/new-spec/light-client.md
```diff
@@ -0,0 +1,111 @@
...
+    votingPower += validator.VotingPower    
+  return votingPower
+  
+updateValidatorSet(valSetNumberOfTheHeader):
+  while valSetNumber != valSetNumberOfTheHeader do
+    signedHeader = LastHeader(valSetNumber)
+    if isValid(signedHeader) then
+      setValidatorSet(signedHeader)
+    else return error
+  return
```
melekes: missed closing "```"

docs/specification/new-spec/light-client.md
```diff
@@ -0,0 +1,111 @@
...
+
+type ResultValidators struct {
+    BlockHeight int64              
+    Validators  []Validator 
+    ValSetTime int64               // time the current validator set is initialised, i.e, time of the last validator change before header BlockHeight 
+}
+```
+
+We assume that Tendermint keeps track of the validator set changes and that each time a validator set is changed it is 
+being assigned the next sequence number. We can this number the validator set sequence number. Tendermint also remembers
```
xla: `We can this number the validator set sequence number.` is very funky, probably needs to be grammatically corrected.

docs/specification/new-spec/light-client.md
```diff
@@ -0,0 +1,111 @@
...
+We assume that Tendermint keeps track of the validator set changes and that each time a validator set is changed it is 
+being assigned the next sequence number. We can this number the validator set sequence number. Tendermint also remembers 
+the Time from the header when the next validator set is initialised (starts to be in power), and we refer to this time
+as validator set init time.
+Furthermore, we assume that each validator set change is signed (committed) by the current validator set. More precisely,
+given a block `H` that contains transactions that are modifying the current validator set, the Merkle root hash of the next 
+validator set (modified based on transactions from block H) will be in block `H+1` (and signed by the current validator 
+set), and then starting from the block `H+2`, it will be signed by the next validator set.    
+
+Note that the real Tendermint RPC api is slightly different (for example, response messages contains more data and function
```
xla: `API` should be in caps as it is an acronym.

xla: `contains` -> `contain`.

docs/specification/new-spec/light-client.md
```diff
@@ -0,0 +1,111 @@
...
+
+
+Locally, light client manages the following state:
+
+```golang
+valSet []Validator     // current validator set (last known and verified validator set) 
+valSetNumber int64     // sequence number of the current validator set 
+valSetHash []byte      // hash of the current validator set
+valSetTime int64       // time when the current validator set is initialised  
+```
```
xla: Improved readbility:

``` golang
valSet       []Validator // current validator set (last known and verified validator set)
valSetNumber int64       // sequence number of the current validator set
valSetHash   []byte      // hash of the current validator set
valSetTime   int64       // time when the current validator set is initialised
```

docs/specification/new-spec/light-client.md
```diff
@@ -0,0 +1,111 @@
...
+    Commit Commit
+    ValSetNumber int64 
+}
+
+type ResultValidators struct {
+    BlockHeight int64              
+    Validators  []Validator 
+    ValSetTime int64               // time the current validator set is initialised, i.e, time of the last validator change before header BlockHeight 
+}
+```
```
xla: Improve formatting:

``` golang
type SignedHeader struct {
	Header       Header
	Commit       Commit
	ValSetNumber int64
}

type ResultValidators struct {
	BlockHeight int64
	Validators  []Validator
	// time the current validator set is initialised, i.e, time of the last validator change before header BlockHeight
	ValSetTime int64
}

```

docs/specification/new-spec/light-client.md
xla: `` ``` ``

docs/specification/new-spec/light-client.md
```diff
@@ -0,0 +1,111 @@
...
+valSetTime int64       // time when the current validator set is initialised  
+```
+
+The light client is initialised with the trusted validator set, for example based on the known validator set hash,
+validator set sequence number and the validator set init time.
+The core of the light client logic is captured by the Certify function that is used to 1) verify if the given header is valid,
+and 2) update the validator set (when the given header is valid and it is more recent than the seen headers). 
+
+```golang
+Certify(signedHeader SignedHeader):
```
xla: I know this is just a description of the current working of the light client, but `Certify` seems to conflate too many things, validation and mutation namely, while the name does not communicate that.

docs/specification/new-spec/light-client.md
milosevic: It's a typo. It should be " We call this number the validator set sequence number".

docs/specification/new-spec/light-client.md
milosevic: This can be seen more as a description of core mechanisms of the light client. Current implementation is quite different. I agree that this is not the best name and that updating validator set is a side effect. My idea was just to have Certify as an entry point where we show how everything works together. The same underlying mechanisms can be used in the context of IBC, pooled security architecture and the light client. Do you have a suggestion how we organise the content differently?

xla: Something that would imply mutation could be a good start e.g.: `Update`

docs/specification/new-spec/light-client.md
```diff
@@ -0,0 +1,113 @@
...
+  assertThat signedHeader.valSetNumber >= valSetNumber  
+  if isValid(signedHeader) and signedHeader.Header.Time <= valSetTime + UNBONDING_PERIOD then
+    setValidatorSet(signedHeader)
+    return true
+  else
+    updateValidatorSet(signedHeader.ValSetNumber)
+    return Certify(signedHeader)
```
xla: This is a recursive call, correct? What happened if the subsequent call to Certify fails even so `updateValidatorSet` has been called already?

docs/specification/new-spec/light-client.md
ebuchman: "We can call this number ..."

docs/specification/new-spec/light-client.md
milosevic: I think that it's not a problem in updating validator set and failing to check next header. With a correct FullNode this should never happen. In general, any call to the FullNode might fail or some of subsequent check could fail. At that point Certify function should fail and light client should decide how to deal with it, for example to connect to other FullNode.

xla: Gotcha, thanks for the clarification. Not sure if it was made explicit somehwere what the guarantees are for the light client.

milosevic: We should make guarantees explicit, this is a first step to agree on the basic mechanisms.

docs/specification/new-spec/light-client.md
```diff
@@ -0,0 +1,114 @@
...
+has the same level of security as Full Node processes (without being itself a Full Node). 
+
+To be able to validate a Merkle proof, a light client needs to validate the blockchain header that contains the root app hash. 
+Validating a blockchain header in Tendermint consists in verifying that the header is committed (signed) by >2/3 of the 
+voting power of the corresponding validator set. As the validator set is a dynamic set (it is changing), one of the 
+core functionality of the light client is updating the current validator set, that is then used to verify the 
+blockchain header, and further the corresponding Merkle proofs. 
+
+For the purpose of this light client specification, we assume that the Tendermint Full Node exposes the following functions over
+Tendermint RPC:
```
odeke-em: What's `Tendermint RPC`?

milosevic: http://tendermint.readthedocs.io/projects/tools/en/master/specification/rpc.html

docs/specification/new-spec/light-client.md
```diff
@@ -0,0 +1,114 @@
...
+
+```golang
+Header(height int64) (SignedHeader, error) // returns signed header for the given height
+Validators(height int64) (ResultValidators, error) // returns validator set for the given height
+LastHeader(valSetNumber int64) (SignedHeader, error)   // returns last header signed by the validator set with the given validator set number
+
+type SignedHeader struct {
+    Header        Header 
+    Commit        Commit
+    ValSetNumber  int64
```
ethanfrey: Why is this number needed?
You return it on the signed header, but you don't return it on ResultValidator.

milosevic: Strictly speaking, it is not needed, but it emphasise that validator set changes are first class citizens. It also simplifies supporting applications that require to strictly track validator set changes. Light client logic also seems simpler. The cost of supporting it does not seem high. We could also add this number in ResultValidator for symmetry.

ebuchman:
I think we need to fix the condition re the Unbonding period (if its longer than the unbonding period, we have to revert to social consensus, we cant trust Tendermint nodes!).

Otherwise I think it looks good.

Notation could be a bit better re mutable state but good enough to merge for now.

PR #1340, opened 2018-03-20
Remove unnecessary bytes.Compare() call

EugeneChung:
It's a simple patch.
bytes.Compare(v.Address, other.Address) doesn't need to be called two times. 

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


xla:
@EugeneChung Could you base your commit on latest tendermint develop, looks like you committed on top of master.

EugeneChung:
@xla It was my mistake. Let me pull another request.

PR #1341, opened 2018-03-20
Remove unnecessary bytes.Compare() call

EugeneChung:
It's a simple patch.
bytes.Compare(v.Address, other.Address) doesn't need to be called two times.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


xla:
Thanks for your contribution!

:+1: :octocat: :shipit:

Issue #1230, opened 2018-02-20
Add retry functionality when verifying seed existence


RobertDiebels:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
FEATURE REQUEST

I would like to suggest that Tendermint implements a retry mechanism when verifying the existence of seeds.

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):
All.

**ABCI app** (name for built-in, URL for self-written if it's publicly available):
None.

**Merkleeyes version** (use `git rev-parse --verify HEAD`, skip if you don't use it):
Unknown,

**Environment**:
- **OS** (e.g. from /etc/os-release): N/A
- **Install tools**: Kubernetes
- **Others**: Kubernetes

**What happened**:
Tendermint crashes if it is unable to locate one of its seeds.

**What you expected to happen**:
I expect Tendermint to retry finding its seeds instead of terminating.

**How to reproduce it** (as minimally and precisely as possible):
If necessary I can put my Kubernetes configuration online for you to test. However I believe the below example is enough.

**Anything else do we need to know**:
Below you can find a description of the use case for this FEATURE REQUEST.

**Use case**
I am trying to create a Kubernetes cluster with Hyperledger Burrow which uses Tendermint under the hood.

In my configuration each Tendermint/Burrow node is represented by a Pod.
Each Pod is tied to a Kubernetes Service which gives it a unique FQDN.

I've written some code that checks the cluster configuration and sets the seeds configuration for each Burrow/Tendermint node to the FQDN of all the other nodes.

This all functions perfectly fine. 

**Problem this feature would solve**
However, the problem with the Kuberntes use case is as follows. Each Burrow/Tendermint node requires their seeds to be up before it the node itself can function.

Because a Kubernetes Pod is unreachable when one of its containers crashes the nodes never find each other. I'll illustrate this with the steps Kubernetes follows when starting Pods.

Lets say I have 3 Tendermint/Burrow nodes which is equivalent to 3 Kubernetes Pods.
Here all the nodes have all nodes as seeds.

1. The network configuration is loaded.
2. Kubernetes starts launching the pods.
3. Pod 1 loads and starts the Tendermint/Burrow container.
4. Pod 2 loads and starts the Tendermint/Burrow container.
5. Pod 1 starts verifying its seeds are up.
6. Pod 2 starts verifying its seeds are up.
[...]

PR #1348, opened 2018-03-21
do not return an error if we fail to lookup IP

melekes:
So we don't fail when e.g. checking seeds (see
pex/pex_reactor.go#checkSeeds)

If there is no IP, Dial/DialTimeout will return an error.

Refs #1230

* [x] Wrote tests
* [x] Updated CHANGELOG.md

### DIFF
--- a/p2p/netaddress.go
+++ b/p2p/netaddress.go
@@ -14,7 +14,12 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+
 	cmn "github.com/tendermint/tmlibs/common"
+)
+
+var (
+	IPNotSetError = errors.New("IP not set. If you are using domain names, check that host is resolvable (host -t a <host>)")
 )
 
 // NetAddress defines information about a peer on the network
@@ -81,12 +86,9 @@ func NewNetAddressString(addr string) (*NetAddress, error) {
 	}
 
 	ip := net.ParseIP(host)
-	if ip == nil {
-		if len(host) > 0 {
-			ips, err := net.LookupIP(host)
-			if err != nil {
-				return nil, err
-			}
+
+	if ip == nil && len(host) > 0 {
+		if ips, err := net.LookupIP(host); err == nil {
 			ip = ips[0]
 		}
 	}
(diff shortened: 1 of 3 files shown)


p2p/netaddress.go
```diff
@@ -14,9 +14,14 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+
 	cmn "github.com/tendermint/tmlibs/common"
 )
 
+var (
+	IPNotSetError = errors.New("IP not set. If you are using domain names, check that host is resolvable (host -t a <host>)")
```
xla: More conventional would be `ErrIPNotSet`.

p2p/netaddress_test.go
```diff
@@ -31,11 +32,11 @@ func TestNewNetAddressString(t *testing.T) {
 		{"127.0.0.1:8080", "127.0.0.1:8080", true},
 		{"tcp://127.0.0.1:8080", "127.0.0.1:8080", true},
 		{"udp://127.0.0.1:8080", "127.0.0.1:8080", true},
-		{"udp//127.0.0.1:8080", "", false},
+		// {"udp//127.0.0.1:8080", "", false},
 		// {"127.0.0:8080", false},
 		{"notahost", "", false},
 		{"127.0.0.1:notapath", "", false},
-		{"notahost:8080", "", false},
+		// {"notahost:8080", "", false},
```
xla: Why are some cases commented?

p2p/netaddress.go
```diff
@@ -168,17 +168,25 @@ func (na *NetAddress) DialString() string {
 	)
 }
 
-// Dial calls net.Dial on the address.
+// Dial calls net.Dial on the address. If no IP is set, it returns
+// IPNotSetError.
 func (na *NetAddress) Dial() (net.Conn, error) {
+	if na.IP == nil {
```
xla: What are the scenarios the `Dial` methods are called when the `IP` is `nil`? Isn't that catched in the constructors up top?

p2p/netaddress_test.go
xla: Shouldn't we remove those cases then?

Issue #1304, opened 2018-03-11
p2p: exponential back-off followup


ebuchman:
See review in https://github.com/tendermint/tendermint/pull/1292

- [x] Document the new functionality in the p2p spec (especially https://github.com/tendermint/tendermint/pull/1292#discussion_r173665125)
- [x] Add TODO in test (needs control over time -
 https://github.com/tendermint/tendermint/pull/1292#commitcomment-28038530)
- [x] consolidate the 2 "max retries" mechanisms (https://github.com/tendermint/tendermint/pull/1292#discussion_r173665070)

melekes:
> consolidate the 2 "max retries" mechanisms (#1292 (comment))

@ebuchman and me agreed it's fine to have both mechanisms a) if we book becomes full, after 3 attempts the address is gone b) if the book is not full, we attempt to connect for ~1.5day.

PR #1349, opened 2018-03-21
Exponential backoff follow up

melekes:
Refs #1304

### DIFF
--- a/p2p/pex/pex_reactor.go
+++ b/p2p/pex/pex_reactor.go
@@ -409,11 +409,15 @@ func (r *PEXReactor) dialPeer(addr *p2p.NetAddress) {
 		// TODO: detect more "bad peer" scenarios
 		if _, ok := err.(p2p.ErrSwitchAuthenticationFailure); ok {
 			r.book.MarkBad(addr)
+			r.attemptsToDial.Delete(addr.DialString())
 		} else {
 			r.book.MarkAttempt(addr)
+			// FIXME: if the addr is going to be removed from the addrbook (hard to
+			// tell at this point), we need to Delete it from attemptsToDial, not
+			// record another attempt.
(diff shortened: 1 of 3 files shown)


p2p/pex/pex_reactor.go
```diff
@@ -409,11 +409,15 @@ func (r *PEXReactor) dialPeer(addr *p2p.NetAddress) {
 		// TODO: detect more "bad peer" scenarios
 		if _, ok := err.(p2p.ErrSwitchAuthenticationFailure); ok {
 			r.book.MarkBad(addr)
+			r.attemptsToDial.Delete(addr.DialString())
 		} else {
 			r.book.MarkAttempt(addr)
+			// FIXME: if the addr is going to be removed from the addrbook (hard to
```
melekes: @ebuchman what do you think about this?

ebuchman: Oooh. Good catch.

So the problem is that the addr might get removed from the AddrBook later once it's full, but it's still being tracked in attemptsToDial, and then will never get deleted ?

So our attemptsToDial map could get bloated with these bad peers.

I think these means we really do nee to unify the MarkBad mechanisms more intelligently.Maybe the AddrBook just needs access to the attemptsToDial somehow? 

Let's open a new issue for it. It's not critical right now but it is important.

melekes:
> then will never get deleted ?

right

> I think these means we really do nee to unify the MarkBad mechanisms more intelligently.Maybe the AddrBook just needs access to the attemptsToDial somehow?

yeah, let's create an issue for it

melekes:
https://github.com/tendermint/tendermint/issues/1356

kaycoinsUSA (commented):
Done

Issue #1275, opened 2018-03-05
p2p: loopbacks should be detected and ignored instead of dialling self infinitely


zramsay:
from: https://github.com/cosmos/gaia/issues/127

odeke-em:
Articulated by @zmanian offline in the engineering channel, this issue is about ensuring that a loopback should be detected and dropped; that is if your own IP is in your list of persistent peers, a node shouldn't dial back to itself infinitely.

ebuchman:
Not just loopback - it might be a public IP and we may not detect that it's us until we see the ID. Once we identify it as ourselves we should record the address as our own and avoid trying to connect to it again

odeke-em:
In deed @ebuchman, not just localhost loopbacks, I was using "loopback" generally to mean a connection back to self.

ebuchman:
Basically we should have the following behaviour:

for each of our addresses:

- we should only dial it once (if at all), after which we will know it's us (because it has the same ID) and we can avoid ever dialling it again :D
- but, it's possible after a lot of time that we change IPs and someone else takes our IP ... so we need to ensure we don't persist our IPs and we just find out every time we restart (I think we already do this)

amrali:
> Not just loopback - it might be a public IP and we may not detect that it's us until we see the ID. Once we identify it as ourselves we should record the address as our own and avoid trying to connect to it again.

If a node detects that the ID belongs to itself would it stop redialing the <IP, port> combination or just the IP address?

melekes:
@amrali combination of IP and port, otherwise we're losing the ability to run many Tendermints on one machine

Issue #1325, opened 2018-03-18
p2p: improve separation of concerns between addrbook/pex/switch 


ebuchman:
There is some confusion as to the relative roles of the addrbook, pex, and switch. We should review the codebase and see if we can improve the separation of concerns. Might also need to consider peer and mconn in the mix.

In particular, there are a bunch of options that complicate certain logic, like toggling the `pex` or the `auth_enc`. We want some things to work regardless of these toggles.

ebuchman:
- It would be great to just remove the options to toggle pex and auth_enc.
  - if not, Switch needs to have the persistent peers and seeds so it can redial them
  - maybe we always run the pex-reactor, but "pex=false" means we don't gossip/request peers using the reactor, just use the persistent peers and seeds ?
- PrivatePeers should be kept in the AddrBook so they can be checked for in AddAddress
- review upgrades to upstream version of AddrBook (lots of changes): https://github.com/btcsuite/btcd/tree/master/addrmgr

ebuchman:
Closing for https://github.com/tendermint/tendermint/issues/2067

PR #1350, opened 2018-03-21
p2p: loopbacks should be detected and ignored instead of dialling self infinitely

melekes:
Refs #1275 

* [x] Wrote tests
* [x] Updated CHANGELOG.md

### DIFF
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -33,13 +33,15 @@ type AddrBook interface {
 
 	// Add our own addresses so we don't later add ourselves
 	AddOurAddress(*p2p.NetAddress)
-
 	// Check if it is our address
 	OurAddress(*p2p.NetAddress) bool
 
 	// Add and remove an address
 	AddAddress(addr *p2p.NetAddress, src *p2p.NetAddress) error
-	RemoveAddress(addr *p2p.NetAddress)
+	RemoveAddress(*p2p.NetAddress)
+
+	// Check if the address is in the book
+	HasAddress(*p2p.NetAddress) bool
 
 	// Do we need more peers?
 	NeedMoreAddrs() bool
@@ -194,6 +196,14 @@ func (a *addrBook) IsGood(addr *p2p.NetAddress) bool {
 	a.mtx.Lock()
 	defer a.mtx.Unlock()
 	return a.addrLookup[addr.ID].isOld()
+}
+
+// HasAddress returns true if the address is in the book.
+func (a *addrBook) HasAddress(addr *p2p.NetAddress) bool {
+	a.mtx.Lock()
+	defer a.mtx.Unlock()
+	ka := a.addrLookup[addr.ID]
+	return ka != nil
 }
 
 // NeedMoreAddrs implements AddrBook - returns true if there are not have enough addresses in the book.
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -34,6 +34,9 @@ type AddrBook interface {
 	// Add our own addresses so we don't later add ourselves
 	AddOurAddress(*p2p.NetAddress)
 
+	// Check if it is our address
+	OurAddress(*p2p.NetAddress) bool
+
 	// Add and remove an address
 	AddAddress(addr *p2p.NetAddress, src *p2p.NetAddress) error
 	RemoveAddress(addr *p2p.NetAddress)
@@ -78,7 +81,7 @@ type addrBook struct {
 	// accessed concurrently
 	mtx        sync.Mutex
 	rand       *rand.Rand
-	ourAddrs   map[string]*p2p.NetAddress
+	ourAddrs   map[string]struct{}
 	addrLookup map[p2p.ID]*knownAddress // new & old
 	bucketsOld []map[string]*knownAddress
 	bucketsNew []map[string]*knownAddress
@@ -93,7 +96,7 @@ type addrBook struct {
 func NewAddrBook(filePath string, routabilityStrict bool) *addrBook {
 	am := &addrBook{
 		rand:              rand.New(rand.NewSource(time.Now().UnixNano())), // TODO: seed from outside
-		ourAddrs:          make(map[string]*p2p.NetAddress),
+		ourAddrs:          make(map[string]struct{}),
 		addrLookup:        make(map[p2p.ID]*knownAddress),
 		filePath:          filePath,
 		routabilityStrict: routabilityStrict,
@@ -154,7 +157,15 @@ func (a *addrBook) AddOurAddress(addr *p2p.NetAddress) {
 	a.mtx.Lock()
 	defer a.mtx.Unlock()
 	a.Logger.Info("Add our address to book", "addr", addr)
-	a.ourAddrs[addr.String()] = addr
+	a.ourAddrs[addr.String()] = struct{}{}
+}
+
+// OurAddress returns true if it is our address.
+func (a *addrBook) OurAddress(addr *p2p.NetAddress) bool {
+	a.mtx.Lock()
+	_, ok := a.ourAddrs[addr.String()]
+	a.mtx.Unlock()
+	return ok
 }
 
 // AddAddress implements AddrBook - adds the given address as received from the given source.
--- a/p2p/pex/pex_reactor_test.go
+++ b/p2p/pex/pex_reactor_test.go
@@ -63,35 +63,45 @@ func TestPEXReactorRunning(t *testing.T) {
 	N := 3
 	switches := make([]*p2p.Switch, N)
 
+	// directory to store address books
 	dir, err := ioutil.TempDir("", "pex_reactor")
 	require.Nil(t, err)
 	defer os.RemoveAll(dir) // nolint: errcheck
-	book := NewAddrBook(filepath.Join(dir, "addrbook.json"), false)
-	book.SetLogger(log.TestingLogger())
+
+	books := make([]*addrBook, N)
+	logger := log.TestingLogger()
 
 	// create switches
 	for i := 0; i < N; i++ {
 		switches[i] = p2p.MakeSwitch(config, i, "127.0.0.1", "123.123.123", func(i int, sw *p2p.Switch) *p2p.Switch {
-			sw.SetLogger(log.TestingLogger().With("switch", i))
+			books[i] = NewAddrBook(filepath.Join(dir, fmt.Sprintf("addrbook%d.json", i)), false)
+			books[i].SetLogger(logger.With("pex", i))
+			sw.SetAddrBook(books[i])
 
-			r := NewPEXReactor(book, &PEXReactorConfig{})
-			r.SetLogger(log.TestingLogger())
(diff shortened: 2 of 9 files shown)


p2p/switch.go
```diff
@@ -148,6 +149,7 @@ func (sw *Switch) IsListening() bool {
 // NOTE: Not goroutine safe.
 func (sw *Switch) SetNodeInfo(nodeInfo NodeInfo) {
 	sw.nodeInfo = nodeInfo
+	sw.nodePublicAddr = nodeInfo.ListenAddr
```
melekes: should this go in `nodeInfo`?

p2p/switch.go
```diff
@@ -522,6 +529,10 @@ func (sw *Switch) addPeer(pc peerConn) error {
 
 	// Avoid self
 	if sw.nodeKey.ID() == peerID {
+		// overwrite current addr to avoid dialing ourselves again
+		// it means original nodePublicAddr was different from public one
```
melekes: possible attack: send us some IP with our ID to prevent us from ever connecting to it

p2p/switch.go
ebuchman: We keep track of our addresses in the AddrBook. See `OurAddress` methods

p2p/switch.go
ebuchman: The ID is authenticated so they could only do this attack if they have our privKey

p2p/switch.go
```diff
@@ -522,6 +529,10 @@ func (sw *Switch) addPeer(pc peerConn) error {
...
+		sw.nodePublicAddr = peerNodeInfo.ListenAddr
```
ebuchman: addrbook.AddOurAddress

p2p/switch.go
melekes: I thought we derive ID from pubkey, which is known and ... public

melekes:
I am a bit concerned how addrbook starts to play a more significant role in Switch but continues to reside inside `pex/` directory. The consequence of that is I have to create addrBookMock. see https://github.com/tendermint/tendermint/pull/1350/commits/f9d6f159f50acba13dac9977c96c85ddae8ec3bf

melekes:
Maybe we should just merge `pex/*` into `p2p`?

ebuchman:
I think MockAddrBook is not a bad thing.

This relates to https://github.com/tendermint/tendermint/issues/1325.

I wouldn't move `pex` into `p2p`. But we do need to address this separation of concerns issue.

p2p/switch_test.go
```diff
@@ -185,6 +178,40 @@ func TestConnAddrFilter(t *testing.T) {
 	assertNoPeersAfterTimeout(t, s2, 400*time.Millisecond)
 }
 
+func TestSwitchFiltersOutItself(t *testing.T) {
+	s1 := MakeSwitch(config, 1, "127.0.0.2", "123.123.123", initSwitchFunc)
+	addr := s1.NodeInfo().NetAddress()
+
+	// add ourselves like we do in node.go#427
```
melekes: also not happy with this

p2p/switch.go
ebuchman: Sure but when we do the AuthEnc handshake, they have to prove they control the privkey relating to the pubkey. They can try to connect with us claiming any ID they want, but as soon as we do the SecretHandshake, we will know for sure if they actually control that ID. That's how we use ID to prevent man-in-the-middle!

melekes: OK, thank you for explaining

p2p/pex/pex_reactor_test.go
```diff
@@ -63,31 +63,44 @@ func TestPEXReactorRunning(t *testing.T) {
...
 		})
 	}
 
 	// fill the address book and add listeners
-	for _, s := range switches {
-		addr := s.NodeInfo().NetAddress()
-		book.AddAddress(addr, addr)
-		s.AddListener(p2p.NewDefaultListener("tcp", s.NodeInfo().ListenAddr, true, log.TestingLogger()))
-	}
+	addr := switches[1].NodeInfo().NetAddress()
```
melekes: can be simplified probably

odeke-em:
Hello @melekes, thank you for working on this bug!

I spent sometime last weekend pondering over how to fix it and realized that it requires more work than we currently have. Currently am not sure that this approach will differentiate "localhost:49957" from "127.0.0.0:49957", which misses the point of detecting if we are on the same that is being dialled. This code uses `p2p.NewNetAddressString` which does a plain address resolution i.e. no canoncialization as in the example below
```go
package main

import (
	"log"
)

func main() {
	log.SetFlags(0)
	addresses := []string{
		"localhost:49957",
		"127.0.0.1:49957",
		"0.0.0.0:49957",
	}

	for i, address := range addresses {
		na, err := p2p.NewNetAddressString(address)
		if err != nil {
			log.Printf("#%d NewNetAddressString address %q error: %v", i, address, err)
			continue
		}
		log.Printf("na: %+v\n", na)
	}
}
```
which prints out
```shell
na: [::1]:49957
na: 127.0.0.1:49957
na: 0.0.0.0:49957
```

I think perhaps we might need to disambiguate addresses i.e. ensure that all those addresses along with my IP address whether external or internal can resolve to the same entity.

odeke-em:
Perhaps the MAC address should be the only differentiator since irrespective of the address it should yield the same but we've just got to make sure that we know that MAC spoofing is really easy to do.

Anyways here is a package for implementing the ARP protocol to help us retrieve a MAC https://github.com/mdlayher/arp

melekes:
@odeke-em thanks for putting thoughts into this. I think it's better if we stick with IPs right now.

```
package main

import (
	"log"

)

func main() {
	log.SetFlags(0)
	addresses := []string{
		"deadbeefdeadbeefdeadbeefdeadbeefdeadbeed@localhost:49957",
		"deadbeefdeadbeefdeadbeefdeadbeefdeadbeed@127.0.0.1:49957",
		"deadbeefdeadbeefdeadbeefdeadbeefdeadbeed@0.0.0.0:49957",
	}

	for i, address := range addresses {
		na, err := p2p.NewNetAddressString(address)
		if err != nil {
			log.Printf("#%d NewNetAddressString address %q error: %v", i, address, err)
			continue
		}
		log.Printf("na: %+v\n", na.String())
	}
}
```

if you use the String method, then localhost will be resolved to 127.0.0.1.

```
na: deadbeefdeadbeefdeadbeefdeadbeefdeadbeed@127.0.0.1:49957
na: deadbeefdeadbeefdeadbeefdeadbeefdeadbeed@127.0.0.1:49957
na: deadbeefdeadbeefdeadbeefdeadbeefdeadbeed@0.0.0.0:49957
```

> Perhaps the MAC address should be the only differentiator since irrespective of the address it should yield the same but we've just got to make sure that we know that MAC spoofing is really easy to do.

we're using node ID (`~/.tendermint/node.json`) for node identification. what's the benefits of using MAC over this?

odeke-em:
>if you use the String method, then localhost will be resolved to 127.0.0.1.

@melekes wow, that's interesting, I didn't realize that. That's exactly what I was hoping for. However, it's weird that that is magic, that "%*v" gives a different representation than "%s" and this doesn't seem to be documented either. 

>we're using node ID (~/.tendermint/node.json) for node identification. what's the benefits of using MAC over this?

Yeah, before I didn't realize that we could get the resolved IP so MAC would have been the common identifier. And yes, node ID is sufficient.

Thank you for taking time to answer my questions @melekes!

p2p/pex/addrbook.go
```diff
@@ -150,7 +153,15 @@ func (a *addrBook) AddOurAddress(addr *p2p.NetAddress) {
...
 	defer a.mtx.Unlock()
 	a.Logger.Info("Add our address to book", "addr", addr)
-	a.ourAddrs[addr.String()] = addr
+	a.ourAddrs[addr.String()] = struct{}{}
+}
+
+// OurAddress returns true if it is our address.
+func (a *addrBook) OurAddress(addr *p2p.NetAddress) bool {
+	a.mtx.Lock()
+	defer a.mtx.Unlock()
```
odeke-em: This might be overkill using a defer here
```go
a.mtx.Lock()
_, ok := a.ourAddrs[addr.String()]
a.mxt.Unlock()
return ok
```

p2p/pex/pex_reactor_test.go
```diff
@@ -63,35 +63,45 @@ func TestPEXReactorRunning(t *testing.T) {
...
 			return sw
 		})
 	}
 
-	// fill the address book and add listeners
-	for _, s := range switches {
-		addr := s.NodeInfo().NetAddress()
-		book.AddAddress(addr, addr)
-		s.AddListener(p2p.NewDefaultListener("tcp", s.NodeInfo().ListenAddr, true, log.TestingLogger()))
+	addOtherNodeAddrToAddrBook := func(switchIndex, otherSwitchIndex int) {
```
odeke-em:
```go
:= func(to, from int) {
```

p2p/switch.go
```diff
@@ -381,6 +383,11 @@ func (sw *Switch) DialPeersAsync(addrBook AddrBook, peers []string, persistent b
 // DialPeerWithAddress dials the given peer and runs sw.addPeer if it connects and authenticates successfully.
 // If `persistent == true`, the switch will always try to reconnect to this peer if the connection ever fails.
 func (sw *Switch) DialPeerWithAddress(addr *NetAddress, persistent bool) error {
+	// do not dial ourselves
+	if sw.addrBook.OurAddress(addr) {
```
odeke-em: Nice!

odeke-em:
@melekes just one minor request, after recompiling tendermint, could you ensure that adding your IP address to the config file like @zramsay had reported in https://github.com/tendermint/tendermint/issues/1275 will properly return that `Err*DialingSelf`? I ask such an integration/end-to-end test that you can do on your terminal, if you haven't already, would help avoid surprises from complex interactions when using the CLI.

Thank you!

p2p/switch.go
ebuchman: Not sure if this is the right spot.

The thing we want to avoid is our address getting into the AddressBook. If it does, then we'll always be trying to dial ourselves, and we'll see logs marking this failure.

If we can ensure we don't get into the addrbook, or that if we do we are soon removed, then I think this doesn't need to be here

melekes:
@ebunchman addressed your comments. PTAL

melekes:
Gonna try to spin up a network with this branch.

p2p/switch.go
```diff
@@ -346,17 +353,20 @@ func (sw *Switch) DialPeersAsync(addrBook AddrBook, peers []string, persistent b
 	for _, err := range errs {
 		sw.Logger.Error("Error in peer's address", "err", err)
 	}
+	if len(errs) > 0 {
+		return errors.New("Errors in peer addresses (see errors above)")
```
ebuchman: hmm, this means if there are lookup errors or typos in the persistent_peers, tendermint will fail to start ? didn't some one ask for us not to do this previously ?

melekes: right right

PR #1353, opened 2018-03-23
Update godep

ebuchman:
We actually shouldn't merge this as is - I updated tmlibs and abci and not sure why the state tests are now failing. Seems to be because of a super minor mismatch in empty struct in the KI64Pair between `[]uint{}` and `[]uint(nil)`  :(

Maybe @xla or @melekes can help figure this out .

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,9 +2,10 @@
 
 
 [[projects]]
+  branch = "master"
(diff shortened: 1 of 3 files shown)


ebuchman:
Replacing with https://github.com/tendermint/tendermint/pull/1354 for better branch name ...

Issue #1130, opened 2018-01-21
p2p: seed more todo


ebuchman:
Seed mode has been implemented. Some more things to do:

- [x] TODO https://github.com/tendermint/tendermint/pull/1128#discussion_r162799294
- [x] Should seeds respond with a bias towards good peers ?
- [x] defaultSeedDisconnectWaitPeriod should be at least as long as we expect it to take for a peer to become MarkGood

ebuchman:
Note MarkGood can take 2-3 hours minimum.

melekes:
> Should seeds respond with a bias towards good peers ?

probably, yes. if node connects to seed, it does not have any trusted peers. I guess (especially in the beginning) node should have more trusted peers than untrusted.

PR #1359, opened 2018-03-24
p2p seed mode follow up

melekes:
Refs #1130 

* [x] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -42,15 +42,19 @@ type AddrBook interface {
 	NeedMoreAddrs() bool
 
 	// Pick an address to dial
-	PickAddress(newBias int) *p2p.NetAddress
+	PickAddress(biasTowardsNewAddrs int) *p2p.NetAddress
 
 	// Mark address
 	MarkGood(*p2p.NetAddress)
 	MarkAttempt(*p2p.NetAddress)
 	MarkBad(*p2p.NetAddress)
 
+	IsGood(*p2p.NetAddress) bool
+
 	// Send a selection of addresses to peers
 	GetSelection() []*p2p.NetAddress
+	// Send a selection of addresses with bias
+	GetSelectionWithBias(biasTowardsNewAddrs int) []*p2p.NetAddress
 
 	// TODO: remove
 	ListOfKnownAddresses() []*knownAddress
@@ -173,6 +177,14 @@ func (a *addrBook) RemoveAddress(addr *p2p.NetAddress) {
 	a.removeFromAllBuckets(ka)
 }
 
+// IsGood returns true if peer was ever marked as good and haven't
+// done anything wrong since then.
+func (a *addrBook) IsGood(addr *p2p.NetAddress) bool {
+	a.mtx.Lock()
+	defer a.mtx.Unlock()
+	return a.addrLookup[addr.ID].isOld()
+}
+
 // NeedMoreAddrs implements AddrBook - returns true if there are not have enough addresses in the book.
 func (a *addrBook) NeedMoreAddrs() bool {
 	return a.Size() < needAddressThreshold
@@ -180,27 +192,27 @@ func (a *addrBook) NeedMoreAddrs() bool {
 
 // PickAddress implements AddrBook. It picks an address to connect to.
 // The address is picked randomly from an old or new bucket according
-// to the newBias argument, which must be between [0, 100] (or else is truncated to that range)
+// to the biasTowardsNewAddrs argument, which must be between [0, 100] (or else is truncated to that range)
 // and determines how biased we are to pick an address from a new bucket.
 // PickAddress returns nil if the AddrBook is empty or if we try to pick
 // from an empty bucket.
-func (a *addrBook) PickAddress(newBias int) *p2p.NetAddress {
+func (a *addrBook) PickAddress(biasTowardsNewAddrs int) *p2p.NetAddress {
 	a.mtx.Lock()
 	defer a.mtx.Unlock()
 
 	if a.size() == 0 {
 		return nil
 	}
-	if newBias > 100 {
-		newBias = 100
+	if biasTowardsNewAddrs > 100 {
+		biasTowardsNewAddrs = 100
 	}
-	if newBias < 0 {
-		newBias = 0
+	if biasTowardsNewAddrs < 0 {
+		biasTowardsNewAddrs = 0
 	}
 
 	// Bias between new and old addresses.
-	oldCorrelation := math.Sqrt(float64(a.nOld)) * (100.0 - float64(newBias))
-	newCorrelation := math.Sqrt(float64(a.nNew)) * float64(newBias)
+	oldCorrelation := math.Sqrt(float64(a.nOld)) * (100.0 - float64(biasTowardsNewAddrs))
+	newCorrelation := math.Sqrt(float64(a.nNew)) * float64(biasTowardsNewAddrs)
 
 	// pick a random peer from a random bucket
 	var bucket map[string]*knownAddress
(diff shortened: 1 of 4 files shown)


melekes:
Need to write more tests for GetSelection/GetSelectionWithBias

melekes:
@odeke-em see https://github.com/tendermint/tendermint/issues/1130. seeds should return more vetted peers because nodes requesting are new to the network.

odeke-em:
>@odeke-em see #1130. seeds should return more vetted peers because nodes requesting are new to the network.

Gotcha, so the crux is then
screen shot 2018-03-27 at 1 29 48 am

I asked because it was hard to notice that change and I mostly saw the rename, thanks for the confirmation

melekes:
What do you mean by hard to see? Is this PR too big or what? https://github.com/tendermint/tendermint/pull/1359/files#diff-8c6b1a350a161ced01d68514ccda7484R198

odeke-em:
Not too big, it's just the naming in changes in `PickAddress` threw me off at the start but then like I show in the screenshot above, I found the crux of the change. Sorry for cluttering your PR with this conversation :) It is all clear now.

melekes:
Ok, cool

p2p/pex/addrbook.go
```diff
@@ -295,6 +307,87 @@ func (a *addrBook) GetSelection() []*p2p.NetAddress {
...
+		a.size()*getSelectionPercent/100)
+	numAddresses = cmn.MinInt(maxGetSelection, numAddresses)
+
+	selection := make([]*p2p.NetAddress, numAddresses)
+
+	oldBucketAndIndex := make(map[int]int)
+	var oldIndex int
+	newBucketAndIndex := make(map[int]int)
+	var newIndex int
+	var bucket map[string]*knownAddress
```
xla: Seems unnecessary given line 349.

p2p/pex/addrbook.go
```diff
@@ -295,6 +307,87 @@ func (a *addrBook) GetSelection() []*p2p.NetAddress {
...
+
+	i := 0
+ADDRS_LOOP:
+	for i < numAddresses {
+		pickFromOldBucket := int((float64(i)/float64(numAddresses))*100) >= biasTowardsNewAddrs
+		bucket = make(map[string]*knownAddress)
+
+		// loop until we pick a random non-empty bucket
+		for len(bucket) == 0 {
```
xla: Won't `bucket` be always of length zero given we construct a new map on line 349?

p2p/pex/addrbook.go
```diff
@@ -295,6 +307,87 @@ func (a *addrBook) GetSelection() []*p2p.NetAddress {
...
+	numAddresses := cmn.MaxInt(
+		cmn.MinInt(minGetSelection, a.size()),
+		a.size()*getSelectionPercent/100)
+	numAddresses = cmn.MinInt(maxGetSelection, numAddresses)
+
+	selection := make([]*p2p.NetAddress, numAddresses)
+
+	oldBucketAndIndex := make(map[int]int)
+	var oldIndex int
+	newBucketAndIndex := make(map[int]int)
```
xla: Why the use of `make` instead of literals (`map[int]int{}`)?

p2p/pex/addrbook.go
```diff
@@ -295,6 +307,87 @@ func (a *addrBook) GetSelection() []*p2p.NetAddress {
...
+				bucket = a.bucketsOld[oldIndex]
+			} else {
+				newIndex = a.rand.Intn(len(a.bucketsNew))
+				bucket = a.bucketsNew[newIndex]
+			}
+		}
+
+		// pick a random index and loop over the map to return that index
+		randIndex := a.rand.Intn(len(bucket))
+		if (pickFromOldBucket && a.nOld > 0) || a.nNew == 0 {
```
xla: This is repeated, maybe we can break it out into its own method.

p2p/pex/addrbook.go
```diff
@@ -295,6 +307,87 @@ func (a *addrBook) GetSelection() []*p2p.NetAddress {
...
+			// check if we already added this address
+			if ind, ok := newBucketAndIndex[newIndex]; ok && ind == randIndex {
+				continue ADDRS_LOOP
+			}
+			newBucketAndIndex[newIndex] = randIndex
+		}
+
+		for _, ka := range bucket {
+			if randIndex == 0 {
+				selection[i] = ka.Addr
```
xla: Why not simply `append`? If the usage `i` changes this might have unintended side-effects.

p2p/pex/pex_reactor.go
```diff
@@ -189,8 +191,11 @@ func (r *PEXReactor) Receive(chID byte, src Peer, msgBytes []byte) {
 
 		// Seeds disconnect after sending a batch of addrs
 		if r.config.SeedMode {
-			// TODO: should we be more selective ?
-			r.SendAddrs(src, r.book.GetSelection())
+			// if node connects to seed, it does not have any trusted peers.
+			// Especially in the beginning, node should have more trusted peers than
+			// untrusted.
+			const biasToSelectNewPeers = 30 // 70 to select good peers
```
xla: Can this be moved to the top with the rest of the constants?

melekes: what for?

p2p/pex/addrbook.go
melekes: we can say the same thing about every loop basically.

p2p/pex/addrbook.go
melekes: yes, thank your

p2p/pex/addrbook.go
melekes: just matter of style

p2p/pex/addrbook.go
melekes: no, because later we do `bucket = a.bucketsXXX[newIndex]`

p2p/pex/addrbook.go
melekes: ok

p2p/pex/pex_reactor.go
ebuchman: good to have all the constants in one place

p2p/pex/addrbook.go
```diff
@@ -295,6 +307,87 @@ func (a *addrBook) GetSelection() []*p2p.NetAddress {
...
+				i++
```
ebuchman: can we add more comments here about what's going on, and where this `i` came from and how we're using it

p2p/pex/pex_reactor.go
melekes: ok, will do. But I am gonna write a blog post and try to convince you that it's a good place if a const is not public and applied in one place.

p2p/pex/addrbook.go
melekes: ok, renamed to `selectionIndex`

p2p/pex/pex_reactor.go
```diff
@@ -30,17 +30,24 @@ const (
...
-	defaultCrawlPeersPeriod         = 30 * time.Second // check some peers every this
+
+	// We want seeds to only advertise good peers. Therefore they should wait at
+	// least as long as we expect it to take for a peer to become good before
+	// disconnecting.
+	// see consensus/reactor.go: blocksToContributeToBecomeGoodPeer
+	// 10000 blocks assuming 1s blocks ~ 2.7 hours.
+	defaultSeedDisconnectWaitPeriod = 3 * time.Hour
+
+	defaultCrawlPeerInterval = 2 * time.Minute // don't redial for this. TODO: back-off. what for?
```
ebuchman: no longer using this ?

melekes: line#562

PR #1370, opened 2018-03-26
Release/0.17.0

melekes:


ebuchman:
Ok, I'm looking through this now and realizing it's actually breaking, at least because of changes in types, eg. the change to use `SignBytes` instead of `WriteSignBytes`.

There's also the change from glide->dep, which technically isn't covered by the semver, but arguably could be (though I made non-breaking releases in other repos with this).

Seems to be a bunch missing from changelog too:
- no mention of priv val stuff
- no mention of the WriteSignBytes->SignBytes changes (breaking)
- mark peer as good and bad stuff 
- other work on p2p like exponential back off 

We should also mention the intro of the `wire` package to ease upgrade to go-amino.

Also there was no version bump, but I just pushed that.

So either we need to make this non breaking in types, or just call it 0.17. I'd be fine with it being 0.17  now, since we're only breaking Golang API stuff, rather than user-level file stuff.

@melekes leaving you to pick up on, I probably won't have time tonight.

Worst case we just tag this as 0.17.0-rc1 or something and release SDK pointing to that.

ebuchman:
And now there's a non-deterministic test failure in the types/priv_validator @xla

melekes:
damn, I forgot to push yesterday...

melekes:
https://circleci.com/gh/tendermint/tendermint/5089

PR #1371, opened 2018-03-27
Process initial validator set and appHash

adrianbrink:


melekes:
```
consensus/replay.go:271:25: res.Validators undefined (type *"github.com/tendermint/tendermint/vendor/github.com/tendermint/abci/types".ResponseInitChain has no field or method Validators)
consensus/replay.go:272:13: res.AppHash undefined (type *"github.com/tendermint/tendermint/vendor/github.com/tendermint/abci/types".ResponseInitChain has no field or method AppHash)
```

you can point dep to a specific revision of `abci` in `Gopkg.toml`

melekes:
Can be closed, I think, since original issue was closed.

ebuchman:
We actually still need this so that the app can set the initial validator set

consensus/replay.go
```diff
@@ -264,10 +264,12 @@ func (h *Handshaker) ReplayBlocks(state sm.State, appHash []byte, appBlockHeight
 			Validators:    validators,
 			AppStateBytes: h.appState,
 		}
-		_, err := proxyApp.Consensus().InitChainSync(req)
+		res, err := proxyApp.Consensus().InitChainSync(req)
 		if err != nil {
 			return nil, err
 		}
+		state.Validators = res.Validators
```
ebuchman: state is not a pointer so non-mutable

consensus/replay.go
```diff
@@ -264,10 +264,12 @@ func (h *Handshaker) ReplayBlocks(state sm.State, appHash []byte, appBlockHeight
...
+		return res.AppHash, nil
```
ebuchman: we still want to run the replay even if we're sending InitChain - the blockchain might have 1000 blocks in it even though the app wants to start from 0 ...

ebuchman:
Thanks - I'll fix this in a separate PR

ebuchman:
See https://github.com/tendermint/tendermint/pull/1701

Issue #1369, opened 2018-03-26
Subscribe on delivertx tag doesn't work


ethanfrey:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):

BUG REPORT

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):

0.16.0-c8a2bdf7

**ABCI app** (name for built-in, URL for self-written if it's publicly available):

https://github.com/confio/weave
(mycoind)

**Merkleeyes version** (use `git rev-parse --verify HEAD`, skip if you don't use it):

**Environment**:
- **OS** (e.g. from /etc/os-release): Ubuntu 16.04.4
- **Install tools**:
- **Others**:

**What happened**:

I add tags to my custom app on DeliverTx, and search for them over rpc.
Works well.

I subscribe to those same tags with `/subscribe` and never get notified.
If I subscribe to all tx with `/subscribe`, it does work...

**What you expected to happen**:

I expected the behavior documented in the godoc:

```
// Tendermint provides a few predefined keys: tm.event, tx.hash and tx.height.
// Note for transactions, you can define additional keys by providing tags with
// DeliverTx response.
//
//		DeliverTx{
//			Tags: []*KVPair{
//				"agent.name": "K",
//			}
//	  }
//
//		tm.event = 'Tx' AND agent.name = 'K'
//		tm.event = 'Tx' AND account.created_at >= TIME 2013-05-03T14:45:00Z
//		tm.event = 'Tx' AND contract.sign_date = DATE 2017-01-01
//		tm.event = 'Tx' AND account.owner CONTAINS 'Igor'
```

I found nice unit tests for search, which worked well, but not for subscriptions.

**How to reproduce it** (as minimally and precisely as possible):

You can work through the demo on https://github.com/confio/weave-js/blob/master/README.md

The error is in the last section "Searching and subscribing to tx", of course, you need to run the server locally and create a few tx first....

If I search with query: `cash='0918829CE2E3B4F0B1B61EB2309E311CF496199D'`, I get a list of transactions that modified the wallet belonging to that address.

If I subscribe with query: `tm.event = 'Tx'` before sending a Tx, I get the tx as desired.

If I subscribe with query `tm.event = 'Tx' AND cash='0918829CE2E3B4F0B1B61EB2309E311CF496199D'`, I never get an event.

[...]

PR #1379, opened 2018-03-27
subscribe does not work for a custom tag

melekes:
Refs #1369

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,13 +27,15 @@ BUG FIXES:
 
 ## 0.17.2 (TBD)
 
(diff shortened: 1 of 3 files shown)


melekes:
TODO: write a test for this https://github.com/tendermint/tendermint/pull/1379/files#diff-044e9bed5a5657ed5429f4f00689bb62R104

melekes:
Please note this is a temporary fix. I am waiting to here from @jaekwon on https://github.com/tendermint/tendermint/issues/1369#issuecomment-377169038. Until then I think we should merge this PR.

Issue #1380, opened 2018-03-27
config: trim whitespace from elements of lists (like persistent_peers)


ebuchman:
If theres whitespace its added to the ID. Needs to be trimmed

PR #1389, opened 2018-03-29
trim whitespace from elements of lists (like `persistent_peers`)

melekes:
Refs #1380


melekes:
Because TrimSpace operates on a byte slice? So would you prefer a small private func in node.go instead?

PR #1390, opened 2018-03-29
Fix lint errors

tomc974:
This PR is just fixing a bunch of lint errors. Each error category has it's own commit so if any one of them doesn't suit your style, it will be easy to revert it.

### DIFF
--- a/benchmarks/proto/test.pb.go
+++ b/benchmarks/proto/test.pb.go
@@ -416,16 +416,16 @@ func (m *ResultStatus) MarshalTo(data []byte) (int, error) {
 	}
 	if m.PubKey == nil {
 		return 0, github_com_gogo_protobuf_proto.NewRequiredNotSetError("pubKey")
+	} else {
+		data[i] = 0x12
+		i++
+		i = encodeVarintTest(data, i, uint64(m.PubKey.Size()))
+		n2, err := m.PubKey.MarshalTo(data[i:])
+		if err != nil {
+			return 0, err
+		}
+		i += n2
 	}
-	data[i] = 0x12
-	i++
-	i = encodeVarintTest(data, i, uint64(m.PubKey.Size()))
-	n2, err := m.PubKey.MarshalTo(data[i:])
-	if err != nil {
-		return 0, err
-	}
-	i += n2
-
 	if m.LatestBlockHash != nil {
 		data[i] = 0x1a
 		i++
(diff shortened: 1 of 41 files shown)


types/priv_validator/sign_info.go
```diff
@@ -149,11 +149,11 @@ func (lsi *LastSignedInfo) SignVote(signer types.Signer, chainID string, vote *t
 // If the HRS are equal and the only thing changed is the timestamp, it sets the timestamp to the previous
 // value and the Signature to the LastSignedInfo.Signature.
 // Else it returns an error.
-func (lsi *LastSignedInfo) SignProposal(signer types.Signer, chainID string, proposal *types.Proposal) error {
+func (info *LastSignedInfo) SignProposal(signer types.Signer, chainID string, proposal *types.Proposal) error {
```
melekes: what's wrong with `lsi`?

config/config.go
```diff
@@ -170,23 +170,23 @@ func (b BaseConfig) ChainID() string {
...
-func (b BaseConfig) NodeKeyFile() string {
-	return rootify(b.NodeKey, b.RootDir)
+func (c BaseConfig) NodeKeyFile() string {
+	return rootify(c.NodeKey, c.RootDir)
 }
 
 // DBDir returns the full path to the database directory
-func (b BaseConfig) DBDir() string {
-	return rootify(b.DBPath, b.RootDir)
+func (c BaseConfig) DBDir() string {
```
melekes: why `cfg` below, but here you are using `c`? should not we be consistent? i.e. `cfg` everywhere

benchmarks/proto/test.pb.go
```diff
@@ -416,16 +416,16 @@ func (m *ResultStatus) MarshalTo(data []byte) (int, error) {
 	}
 	if m.PubKey == nil {
 		return 0, github_com_gogo_protobuf_proto.NewRequiredNotSetError("pubKey")
-	} else {
```
melekes:
```
// Code generated by protoc-gen-gogo.
// source: test.proto
// DO NOT EDIT!
```

types/priv_validator/sign_info.go
tomc974: Nothing's wrong with `lsi`, it's just that some receivers for `LastSignedInfo` were named `lsi` and others were named `info`. I just settle for `info` for all of them. I will switch to `lsi` instead.

melekes: got it. thanks

config/config.go
tomc974: `BaseConfig` type was using `b` and `c` as receiver names and `ConsensusConfig` type was using `c` and `cfg`. I tried to minimize changes by settling on `c` for `BaseConfig` and on `cfg` for `ConsensusConfig`. I will change to `cfg` for both types.

benchmarks/proto/test.pb.go
tomc974: Oh, I miss that! I will revert the changes...

tomc974:
@melekes All comments should be fixed now.

melekes:
🥇 🏎 🍰

Issue #581, opened 2017-07-12
Include voting power in the /status endpoint


adrianbrink:
from slack:
```
antoineherzog [12:15 PM] 
could we include the validator power of the node on the status page?
http://174.138.118.46:46657/status
[12:15] 
it will be helpful (edited)
[12:16] 
like latest_validator_power:
[12:16] 
XX
antoineherzog [12:24 PM] 
for the node we query.
```

PR #1391, opened 2018-03-29
Include validator power in /status

melekes:
Refs #581 

```
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    ....
    "latest_block_height": 1149,
    "latest_block_time": "2018-03-29T11:35:43.112Z",
    "syncing": false,
    "validator_status": {
      "voting_power": 10
    }
  }
}
```

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -720,11 +720,7 @@ func (cs *ConsensusState) needProofBlock(height int64) bool {
 func (cs *ConsensusState) proposalHeartbeat(height int64, round int) {
 	counter := 0
 	addr := cs.privValidator.GetAddress()
-	valIndex, v := cs.Validators.GetByAddress(addr)
-	if v == nil {
-		// not a validator
-		valIndex = -1
-	}
+	valIndex, _ := cs.Validators.GetByAddress(addr)
 	chainID := cs.state.ChainID
 	for {
 		rs := cs.GetRoundState()
--- a/rpc/core/status.go
+++ b/rpc/core/status.go
@@ -1,9 +1,11 @@
 package core
 
 import (
+	"bytes"
 	"time"
 
 	ctypes "github.com/tendermint/tendermint/rpc/core/types"
+	sm "github.com/tendermint/tendermint/state"
 	"github.com/tendermint/tendermint/types"
 	cmn "github.com/tendermint/tmlibs/common"
 )
@@ -48,7 +50,10 @@ import (
 // 			"remote_addr": "",
 // 			"network": "test-chain-qhVCa2",
 // 			"moniker": "vagrant-ubuntu-trusty-64",
-// 			"pub_key": "844981FE99ABB19F7816F2D5E94E8A74276AB1153760A7799E925C75401856C6"
+// 			"pub_key": "844981FE99ABB19F7816F2D5E94E8A74276AB1153760A7799E925C75401856C6",
+//			"validator_status": {
+//				"voting_power": 10
+//			}
 // 		}
 // 	},
 // 	"id": "",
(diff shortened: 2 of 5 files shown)


consensus/state.go
```diff
@@ -720,11 +720,7 @@ func (cs *ConsensusState) needProofBlock(height int64) bool {
 func (cs *ConsensusState) proposalHeartbeat(height int64, round int) {
 	counter := 0
 	addr := cs.privValidator.GetAddress()
-	valIndex, v := cs.Validators.GetByAddress(addr)
-	if v == nil {
-		// not a validator
-		valIndex = -1
-	}
+	valIndex, _ := cs.Validators.GetByAddress(addr)
```
ebuchman: what happened here ? Looks like GetByAddress returns 0 if its not found, not -1 ? 

Also noticing it only checks `idx != len(valSet.Validators)` but should probably use `<` ?

rpc/core/status.go
```diff
@@ -72,12 +76,26 @@ func Status() (*ctypes.ResultStatus, error) {
...
 		LatestBlockHash:   latestBlockHash,
 		LatestAppHash:     latestAppHash,
 		LatestBlockHeight: latestHeight,
 		LatestBlockTime:   latestBlockTime,
-		Syncing:           consensusReactor.FastSync()}, nil
+		Syncing:           consensusReactor.FastSync(),
+	}
+
+	// add ValidatorStatus if node is a validator
+	_, vals := consensusState.GetValidators()
```
ebuchman: I think this is actually racey since the height may have incremented by the time we get here in which case the validator set here may not correspond to the Height we're returning. So I think we need to fetch the validators for the given height to be sure.

rpc/core/types/responses.go
```diff
@@ -54,14 +54,19 @@ func NewResultCommit(header *types.Header, commit *types.Commit,
...
 type ResultStatus struct {
-	NodeInfo          p2p.NodeInfo  `json:"node_info"`
-	PubKey            crypto.PubKey `json:"pub_key"`
-	LatestBlockHash   cmn.HexBytes  `json:"latest_block_hash"`
-	LatestAppHash     cmn.HexBytes  `json:"latest_app_hash"`
-	LatestBlockHeight int64         `json:"latest_block_height"`
-	LatestBlockTime   time.Time     `json:"latest_block_time"`
-	Syncing           bool          `json:"syncing"`
+	NodeInfo          p2p.NodeInfo    `json:"node_info"`
+	PubKey            crypto.PubKey   `json:"pub_key"`
```
ebuchman: shall we move this pubkey under the ValidatorStatus ?

ebuchman: Maybe we should change Status to consist of NodeInfo, SyncInfo, ValidatorInfo.

SyncInfo contains all the `LatestXxx` and the `Syncing` bool. `ValidatorInfo` contains the pubkey and the voting power. 

Thoughts ? No need to do it here but let's open a new issue for the idea if you like it

consensus/state.go
ebuchman: Oh awesome I see you fixed this. Thanks!

types/validator_set.go
```diff
@@ -83,27 +83,31 @@ func (valSet *ValidatorSet) Copy() *ValidatorSet {
...
 func (valSet *ValidatorSet) GetByAddress(address []byte) (index int, val *Validator) {
 	idx := sort.Search(len(valSet.Validators), func(i int) bool {
 		return bytes.Compare(address, valSet.Validators[i].Address) <= 0
 	})
-	if idx != len(valSet.Validators) && bytes.Equal(valSet.Validators[idx].Address, address) {
+	if idx < len(valSet.Validators) && bytes.Equal(valSet.Validators[idx].Address, address) {
 		return idx, valSet.Validators[idx].Copy()
 	} else {
-		return 0, nil
+		return -1, nil
```
ebuchman: think we can drop the `else` since the if returns already (human linter woo!)

rpc/core/types/responses.go
melekes: What we have right now is good enough. I would not make such changes without explicit feedback from users.

melekes: I like the idea. The question is whenever our users will like it or not

rpc/core/status.go
melekes: see the new implementation https://github.com/tendermint/tendermint/blob/ed782e7508f430c2ece4fb2ebce82e2d4c2a6623/rpc/core/status.go#L100-L125

rpc/core/types/responses.go
melekes: https://github.com/tendermint/tendermint/issues/1417

PR #1403, opened 2018-04-02
Add a method for creating an https server

zmanian:
Needed for https://github.com/cosmos/cosmos-sdk/pull/601

### DIFF
--- a/rpc/lib/server/http_server.go
+++ b/rpc/lib/server/http_server.go
@@ -48,6 +48,39 @@ func StartHTTPServer(listenAddr string, handler http.Handler, logger log.Logger)
 	return listener, nil
 }
 
+func StartHTTPAndTLSServer(listenAddr string, handler http.Handler, cert_path string, key_path string, logger log.Logger) (listener net.Listener, err error) {
+	// listenAddr should be fully formed including tcp:// or unix:// prefix
+	var proto, addr string
+	parts := strings.SplitN(listenAddr, "://", 2)
+	if len(parts) != 2 {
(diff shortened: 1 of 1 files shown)


rpc/lib/server/http_server.go
```diff
@@ -48,6 +48,39 @@ func StartHTTPServer(listenAddr string, handler http.Handler, logger log.Logger)
 	return listener, nil
 }
 
+func StartHTTPAndTLSServer(listenAddr string, handler http.Handler, cert_path string, key_path string, logger log.Logger) (listener net.Listener, err error) {
```
melekes: `func StartHTTPAndTLSServer(listenAddr string, handler http.Handler, certFile, keyFile string, logger log.Logger) (listener net.Listener, err error) {`

rpc/lib/server/http_server.go
```diff
@@ -48,6 +48,39 @@ func StartHTTPServer(listenAddr string, handler http.Handler, logger log.Logger)
...
+		// we used to allow addrs without tcp/unix prefix by checking for a colon
+		// TODO: Deprecate
+		proto = types.SocketType(listenAddr)
+		addr = listenAddr
+		// return nil, errors.Errorf("Invalid listener address %s", lisenAddr)
+	} else {
+		proto, addr = parts[0], parts[1]
+	}
+
+	logger.Info(fmt.Sprintf("Starting RPC HTTPS server on %s socket %v", proto, addr))
```
melekes: should we also log `certFile` and `keyFile` here?

melekes:
Addressed the comments above in a separate PR https://github.com/tendermint/tendermint/pull/1405

PR #1404, opened 2018-04-02
p2p: persistent - redial if first dial fails

### LINKED ISSUE #1401: persistent peers doesnt work if we fail to connect the first time
If we fail to connect to a `persistent_peers` peer the first time, it seems we do not try to redial them right away until the addressbook suggests them during an ensurePeers call.

We should fix this so failure to connect the first time is treated the same as the connection being disconnected

tuxcanfly:
Fixes #1401

* [x] Wrote tests
* [x] Updated CHANGELOG.md

### DIFF
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -87,14 +87,13 @@ func newPeer(pc peerConn, nodeInfo NodeInfo,
 type PeerConfig struct {
 	AuthEnc bool `mapstructure:"auth_enc"` // authenticated encryption
 
-	Dial func(addr *NetAddress, config *PeerConfig) (net.Conn, error)
-
 	// times are in seconds
 	HandshakeTimeout time.Duration `mapstructure:"handshake_timeout"`
 	DialTimeout      time.Duration `mapstructure:"dial_timeout"`
 
 	MConfig *tmconn.MConnConfig `mapstructure:"connection"`
 
+	Fail       bool            `mapstructure:"fail"` // for testing
 	Fuzz       bool            `mapstructure:"fuzz"` // fuzz connection (for testing)
 	FuzzConfig *FuzzConnConfig `mapstructure:"fuzz_config"`
 }
@@ -103,10 +102,10 @@ type PeerConfig struct {
 func DefaultPeerConfig() *PeerConfig {
 	return &PeerConfig{
 		AuthEnc:          true,
-		Dial:             dial,
 		HandshakeTimeout: 20, // * time.Second,
 		DialTimeout:      3,  // * time.Second,
 		MConfig:          tmconn.DefaultMConnConfig(),
+		Fail:             false,
 		Fuzz:             false,
 		FuzzConfig:       DefaultFuzzConnConfig(),
 	}
@@ -115,7 +114,7 @@ func DefaultPeerConfig() *PeerConfig {
 func newOutboundPeerConn(addr *NetAddress, config *PeerConfig, persistent bool, ourNodePrivKey crypto.PrivKey) (peerConn, error) {
 	var pc peerConn
 
-	conn, err := config.Dial(addr, config)
+	conn, err := dial(addr, config)
 	if err != nil {
 		return pc, errors.Wrap(err, "Error creating peer")
 	}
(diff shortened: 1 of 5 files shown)


tuxcanfly:
Not sure if handshake failure should be handled too but my gut feeling says this is mainly for dial failures.

tuxcanfly:
The tests are racy, a proper fix would probably require injecting the dialer to the functions.

p2p/peer.go
```diff
@@ -87,6 +87,8 @@ func newPeer(pc peerConn, nodeInfo NodeInfo,
 type PeerConfig struct {
 	AuthEnc bool `mapstructure:"auth_enc"` // authenticated encryption
 
+	Dial func(addr *NetAddress, config *PeerConfig) (net.Conn, error)
```
melekes: I'd rather not have functions inside config.

melekes: maybe we can Have `badAddr`, which returns error on `DialTimeout`. This will require to inroduce NetAddress interface.

p2p/switch.go
```diff
@@ -470,6 +476,7 @@ func (sw *Switch) addOutboundPeerWithConfig(addr *NetAddress, config *PeerConfig
 	peerConn, err := newOutboundPeerConn(addr, config, persistent, sw.nodeKey.PrivKey)
 	if err != nil {
 		sw.Logger.Error("Failed to dial peer", "address", addr, "err", err)
+		go sw.reconnectToPeer(addr)
```
ebuchman: we only want to do this if the `persistent == true`

p2p/peer.go
ebuchman: I think better would be to set a flag on the remotePeer to allow it to reject the the connection in its accept routine. We should try to avoid putting methods in config structs like this

ebuchman:
> Not sure if handshake failure should be handled too but my gut feeling says this is mainly for dial failures.

Yes we want any kind of connection failure on a persistent peer, dial or otherwise, to cause a reconnect attempt

tuxcanfly:
@ebuchman I was wondering about what type of error would cause a handshake failure and would we want to have the same exponential backoff tracking as general failures (such as dial).

Talked to @jaekwon and he seemed to agree. I still need to read about the handshake process so I might be wrong.

p2p/switch.go
tuxcanfly: Done.

p2p/peer.go
tuxcanfly: Maybe we can have a the `Dial` func somewhere else? It seems more convenient to have a configurable dial behavior for testing, for e.g it could be used for slow connections / timeouts.

tuxcanfly: Maybe `Switch`?

kidinamoto01:
why is this PR still blocked?

tuxcanfly:
@kidinamoto01 this it not ideal fix, I'm not sure if this even a complete fix since it doesn't handle handshake related errors (see comments above). I'll be closing this PR if that's OK, don't want to keep it lingering.

kidinamoto01:
@tuxcanfly Thanks for the update, it's very helpful. Hope we can get this fixed soon O(∩_∩)O~

ebuchman:
hey @tuxcanfly sorry for the delay here but I think you're right - we only want the reconnect on the dial, not on the handshake

ebuchman:
On second thought, the handshake includes a timeout, which we might want to try again on, and even if we fail to connect due to compatibility issues, the remote persistent node might get restarted with correct software and then we can connect to it.

So I think unless we do more work on the typing of errors, its fine to just always run reconnectToPeer. Also note that in reconnectToPeer itself, we keep trying no matter what the error. Ill fix this up and merge, thanks!

ebuchman:
Nope, what you've got is fine and straight forward. For now we'll only run the reconnect if we fail to dial the persistent peers. If we lost a connection though, we'll keep retrying even if eg the handshake fails

ebuchman:
merged plus some minor changes: https://github.com/tendermint/tendermint/commit/0cbbb61962df84d1367c3f7da6e93638ae065819

tuxcanfly:
Awesome! :100:

PR #1406, opened 2018-04-02
Update dockerfile and readme

melekes:
Docker fails with:

```
./bin/tendermint
bash: ./bin/tendermint: No such file or directory
```

**DO NOT SQUASH**

### DIFF
--- a/DOCKER/Dockerfile
+++ b/DOCKER/Dockerfile
@@ -1,8 +1,8 @@
 FROM alpine:3.7
 
 # This is the release of tendermint to pull in.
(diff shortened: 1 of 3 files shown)


melekes:
```
# ldd /bin/tendermint
        /lib64/ld-linux-x86-64.so.2 (0x7fb7c201d000)
        libpthread.so.0 => /lib64/ld-linux-x86-64.so.2 (0x7fb7c201d000)
        libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7fb7c201d000)
```

Issue #1393, opened 2018-03-30
`show_node_id` should return error if TM home is not initialized


kidinamoto01:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
FEATURE REQUEST

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):

v0.17.1
**ABCI app** (name for built-in, URL for self-written if it's publicly available):

-
**Merkleeyes version** (use `git rev-parse --verify HEAD`, skip if you don't use it):

**Environment**:
- **OS** (e.g. from /etc/os-release): macOs
- **Install tools**: dep
- **Others**:

**What happened**:
when I set --home to some non-existent folder, node id is still displayed on the screen.

**What you expected to happen**:

**How to reproduce it** (as minimally and precisely as possible):

tendermint show_node_id --home=~/.gaia4

**Logs (you can paste a part showing an error or attach the whole file)**:

**`/dump_consensus_state` output for consensus bugs**

**Anything else do we need to know**:
`show node id`command should give hint on wrong home path rather just generate a random key.

ebuchman:
Hmm, the default behaviour for Tendermint if no home folder exists is to create one with a default config and node key. If you then later run `tendermint init`, it will not override the node_key, but will generate the priv_validator and genesis for you.

Sounds like you're suggesting that all `tendermint` commands (except `tendermint init`) should fail if the home dir doesn't exist.

kidinamoto01:
yes, we must initialize `tendermint`before running other `tendermint` commands

melekes:
Fixed on develop.

PR #1433, opened 2018-04-08
improve `show_node_id`

kidinamoto01:
* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

This is a simple fix, followed the discussion about this [issue](https://github.com/tendermint/tendermint/issues/1393) , the `show_node_id` function will  check the existence of related file first.

### DIFF
--- a/cmd/tendermint/commands/show_node_id.go
+++ b/cmd/tendermint/commands/show_node_id.go
@@ -6,6 +6,9 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/tendermint/tendermint/p2p"
+
+	cmn "github.com/tendermint/tmlibs/common"
+
 )
 
 // ShowNodeIDCmd dumps node's ID to the standard output.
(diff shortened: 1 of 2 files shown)


cmd/tendermint/commands/show_node_id.go
```diff
@@ -16,10 +19,16 @@ var ShowNodeIDCmd = &cobra.Command{
 }
 
 func showNodeID(cmd *cobra.Command, args []string) error {
-	nodeKey, err := p2p.LoadOrGenNodeKey(config.NodeKeyFile())
-	if err != nil {
-		return err
+	if cmn.FileExists(config.GenesisFile())&&cmn.FileExists(config.PrivValidatorFile()) {
```
melekes: I think it should be more like 

```
nodeKey, err := p2p.LoadNodeKey(config.NodeKeyFile())
....
```

notice we use LoadNodeKey, not LoadOrGenNodeKey

kidinamoto01: I think `LoadNodeKey `is a private function.

kidinamoto01: ok, it's done

melekes: you can remove this `if` now

melekes: I was talking about `if cmn.FileExists(config.GenesisFile())&&cmn.FileExists(config.PrivValidatorFile()) {`!

melekes: there is no need to check genesis and priv_validator files. LoadNodeKey will return an error if there is no node_key.json file.

kidinamoto01: got it ^^

melekes:
🥇 🏎 🍰

PR #1443, opened 2018-04-09
Replaced NodeInfo's pubkey to ID

eeonevision:
* [x] Wrote tests
* [ ] Updated CHANGELOG.md


benchmarks/codec_test.go
```diff
@@ -16,17 +16,18 @@ func BenchmarkEncodeStatusWire(b *testing.B) {
 	b.StopTimer()
 	cdc := amino.NewCodec()
 	ctypes.RegisterAmino(cdc)
-	pubKey := crypto.GenPrivKeyEd25519().PubKey()
+	crypto.GenPrivKeyEd25519()
```
melekes: not needed

benchmarks/codec_test.go
```diff
@@ -94,15 +95,20 @@ func BenchmarkEncodeNodeInfoBinary(b *testing.B) {
...
+	someString := "SOMESTRING"
+	otherString := "OTHERSTRING"
 	nodeInfo := proto.NodeInfo{
-		PubKey:     pubKey2,
-		Moniker:    "SOMENAME",
-		Network:    "SOMENAME",
-		ListenAddr: "SOMEADDR",
-		Version:    "SOMEVER",
-		Other:      []string{"SOMESTRING", "OTHERSTRING"},
+		Id:         &proto.ID{Id: &nodeID},
```
melekes: what's up with all these pointers? can't we just use strings?

benchmarks/proto/test.proto
```diff
@@ -16,6 +16,10 @@ message NodeInfo {
   repeated string other = 7;
 }
 
+message ID {
+  optional string id = 1;
```
melekes: required

node/node.go
```diff
@@ -579,13 +579,13 @@ func (n *Node) ProxyApp() proxy.AppConns {
 	return n.proxyApp
 }
 
-func (n *Node) makeNodeInfo(pubKey crypto.PubKey) p2p.NodeInfo {
+func (n *Node) makeNodeInfo(nodeID p2p.ID) p2p.NodeInfo {
```
melekes: `id`

benchmarks/codec_test.go
eeonevision: Fixed

benchmarks/codec_test.go
eeonevision: No because generated proto.go file have *string type for above fields.
Define just "SOME*" string leads to error: "cannot use * (type string) as type *string in field value"

benchmarks/proto/test.proto
eeonevision: Fixed

node/node.go
eeonevision: there will some confusion, what ID we need to use. nodeID explicitly defines, which ID need to send in argument.

melekes:
Could you please rebase against latest develop? I fixed the data race (`test_cover`).

melekes:
🥇 🏎 🍰

PR #1448, opened 2018-04-10
update docs for latest develop

melekes:


docs/install.rst
```diff
@@ -4,7 +4,7 @@ Install Tendermint
 From Binary
 -----------
 
-To download pre-built binaries, see the `Download page <https://tendermint.com/download>`__.
+To download pre-built binaries, see the `Download page <https://tendermint.com/downloads>`__.
```
zramsay: the downloads page just points to github releases ... should we maybe point to the GH releases instead?

melekes: we might switch to something else.. lets keep it for a while

PR #1454, opened 2018-04-12
Greg/private networks

greg-szabo:
Enabled the use of private IPv4 addresses. (RFC1918). This is required to run tendermint on a private network where all the nodes are on the same private network.


melekes:
https://github.com/tendermint/tendermint/issues/1455#issuecomment-381589305

greg-szabo:
I wrote a long dissection of the RFCs on the issue. All in all, I agree: we should remove this kind of check outright.

greg-szabo:
Closing it, as agreed on the meeting.

Issue #1211, opened 2018-02-14
make /validators RPC endpoint report the effective validator set


pdobacz:
**FEATURE REQUEST?** (not sure):

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):
`v0.15.0`

**Environment**:
- **OS** (e.g. from /etc/os-release):
linux

**What happened**:
If I change the validator set in such a way, that the consensus halts, the `validators` rpc endpoint still points to the "last non-halted" set.

I kinda get the point why (there hasn't been a block that would allow the new validator hash to propagate), but it isn't too usable, since in such cases (like halting of the chain) I cannot really check the *effective* validator set, that drives the consensus (or lack thereof).

**What you expected to happen**:
`validator` endpoint report the effective validator set (in this case - the "halting one")

**How to reproduce it** (as minimally and precisely as possible):
use [dummy_persistent](https://github.com/tendermint/abci/tree/master/example/dummy#persistentdummyapplication) example app:
```
tendermint node --log_level "*:info" --proxy_app persistent_dummy
```

Check validators
```
http 'http://localhost:46657/validators'
```

Send a `val:pubkey1/power1` that halts the consensus:
```
http 'http://localhost:46657/broadcast_tx_sync?tx="val:01F0710E97A3C69F853FFA6A3BE415D4C5C5187220D69602794CD7A952FC400D10/1234"'
```

See that consensus halts (validator set changed), but
```
http 'http://localhost:46657/validators'
```
returns old stuff.

adrianbrink:
The validator RPC endpoint should return the latest information from EndBlock. So if the validator set changes in EndBlock and halts consensus, Tendermint RPC still should return the new (and halting) validator set from the rpc.

ebuchman:
@pdobacz can you check my comments in the PR #1463 please ?

In short, we shouldn't change what validator set is returned here. If you need the current set, see dump_consensus_state. But we can include the changes from EndBlock in this end point.

Does that help ? :)

pdobacz:
@ebuchman Yep, looks like `dump_consensus_state` will do the trick, wasn't aware of this, so I went straight for `validators`. Thank you!

PR #1463, opened 2018-04-13
[rpc] /validators returns the effective validator set

melekes:
Refs #1211

```
$ ./build/tendermint node --proxy_app=persistent_dummy
$ curl 'http://localhost46657/broadcast_tx_sync?tx="val:1624de6220e75d17011aa85f07bdc7827bb5e0c29a2bc069cce0c7dbe4d5d8b6b4ce2d820f/1234"'
$ curl 'http://localhost46657/validators'
```

BEFORE:

```
$ curl 'http://localhost46657/validators'
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    "block_height": 9,
    "validators": [
      {
        "address": "9F7581285867CD90AD8148DFBAD7092F67BC1EF3",
        "pub_key": {
          "type": "AC26791624DE60",
          "value": "MdTzAcUYN7YouVbW5UfCFcOh923QTlr9lY/xGPtnq4s="
        },
        "voting_power": 10,
        "accum": 0
      }
    ]
  }
}
```

AFTER:

```
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    "block_height": 4,
    "validators": [
      {
        "address": "9F2A3F031D3D9B988293FED42406E7ACB8448AC3",
        "pub_key": {
          "type": "AC26791624DE60",
          "value": "510XARqoXwe9x4J7teDCmivAaczgx9vk1di2tM4tgg8="
        },
        "voting_power": 1234,
        "accum": -10
      },
      {
        "address": "9F7581285867CD90AD8148DFBAD7092F67BC1EF3",
        "pub_key": {
          "type": "AC26791624DE60",
          "value": "MdTzAcUYN7YouVbW5UfCFcOh923QTlr9lY/xGPtnq4s="
        },
        "voting_power": 10,
        "accum": 10
      }
    ]
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,11 @@ BUG FIXES:
 - Graceful handling/recovery for apps that have non-determinism or fail to halt
 - Graceful handling/recovery for violations of safety, or liveness
 
+## 0.19.1 (TBD)
+
+IMPROVEMENTS:
+- [rpc] `/validators` now returns the effective validator set if no height was given (See #1211)
+
 ## 0.19.0 (TBD)
 
 BREAKING:
--- a/rpc/core/consensus.go
+++ b/rpc/core/consensus.go
@@ -46,12 +46,18 @@ import (
 // }
 // ```
 func Validators(heightPtr *int64) (*ctypes.ResultValidators, error) {
+	// return the effective validator set if no height was given
+	if heightPtr == nil {
+		lastBlockHeight, vals := consensusState.GetValidators()
+		return &ctypes.ResultValidators{lastBlockHeight, vals}, nil
+	}
+
+	// return the stored validator set if height was given
 	storeHeight := blockStore.Height()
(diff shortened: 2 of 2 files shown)


ebuchman:
Not sure about this.

Leads to confusing semantics because the height being returned here does not match the validator set (its the height for the last block with the validators for the next block). It also makes this endpoint different from all the others that allow you to send an empty height and default to returning info from the latest committed block.

IMO the "effective/current" validator set should come from the dump_consensus_state endpoint, where it already is

ebuchman:
Alternatively, maybe `/validators` can return the validator set from the latest committed block along with any *changes* returned from that latest blocks EndBlock.

Thoughts ?

melekes:
I remember I was thinking about reasons why you want to see effective set other than halt (in which case you need to use `/dump_consensus_state`). But now I can't come with one. Yeah, let's close this.

PR #1471, opened 2018-04-16
Greg/devopstools

greg-szabo:
- Updated Dockerfile used by the new automation (and Makefile target added)
- Updated localnode docker image for more permissive container
- Sentry node terraform and ansible scripts and makefile target for validators


Makefile
```diff
@@ -187,16 +195,35 @@ build-linux:
...
+### Remote full-nodes (sentry) using terraform and ansible
+
+# Server management
+server-setup:
+	@if [ -z "$(DO_API_TOKEN)" ]; then echo "DO_API_TOKEN environment variable not set." ; false ; fi
+	@if ! [ -f $(HOME)/.ssh/id_rsa.pub ]; then ssh-keygen ; fi
+	cd networks/remote/terraform && terraform init && terraform apply -var DO_API_TOKEN="$(DO_API_TOKEN)" -var SSH_KEY_FILE="$(HOME)/.ssh/id_rsa.pub"
+	@if ! [ -f $(CURDIR)/build/node0/config/genesis.json ]; then docker run --rm -v $(CURDIR)/build:/tendermint:Z tendermint/localnode testnet --v 0 --n 4 --o . ; fi
+	cd networks/remote/ansible && ANSIBLE_HOST_KEY_CHECKING=False ansible-playbook -i inventory/digital_ocean.py -l remotenet install.yml
+	@echo "Next step: Add your validator setup in the genesis.json and config.tml files and run \"make server-config\". (Public key of validator, chain ID, peer IP and node ID.)"
```
melekes: and where do I get my validator setup?

networks/remote/ansible/roles/config/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
...
+    mode: 0755
+
+- name: Copy config
+  when: item <= N and ansible_hostname == 'remotenet-node' ~ item
+  copy:
+    src: "{{CONFIGDIR}}/node{{item}}/"
+    dest: "/home/{{service}}/.{{service}}/"
+    owner: "{{service}}"
+    group: "{{service}}"
+  loop: [ 0, 1, 2, 3, 4, 5, 6, 7 ]
```
melekes: why 8 if you're creating only 4 machines?

DOCKER/Dockerfile
```diff
@@ -1,45 +1,22 @@
...
+
+#Default home for tendermint. The node command will look for $TMHOME/config/genesis.json at initialization.
+ENV TMHOME /tendermint
+RUN apk update && \
+    apk upgrade && \
+    apk --no-cache add curl jq && \
+    addgroup tmuser && \
+    adduser -S -G tmuser tmuser -h "$TMHOME"
+
+USER tmuser
```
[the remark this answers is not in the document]
greg-szabo: ok, will do. I didn't remove them, just overwritten the old file with one I wrote.

Makefile
```diff
@@ -178,6 +178,14 @@ metalinter_all:
 	@echo "--> Running linter (all)"
 	gometalinter.v2 --vendor --deadline=600s --enable-all --disable=lll ./...
 
+###########################################################
+### Docker image
+
+build-docker:
+	cp build/tendermint DOCKER/tendermint
+	docker build --label=tendermint --tag="tendermint/tendermint" DOCKER
```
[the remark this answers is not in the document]
greg-szabo: It's metadata for the image. We don't use it but Docker Hub can use it to index our image better. So if anyone searches for tendermint, our image comes up with higher ratio.

Makefile
greg-szabo: The use-case is that validators want to set up sentry nodes for their testnet. They should already have a validator running somewhere on their other server. Alternatively, they can run `tendermint init` on another server which creates a validator configuration for testing. They can copy the files over from their.

greg-szabo: Maybe the targets should be renames from server-setup and such to sentry-setup... What do you think?

networks/remote/ansible/roles/config/tasks/main.yml
greg-szabo: The ansible role can take up to 8 nodes (as an upper limit). This means that when you start the script, it uses N=4 as a default and the rest will not be used, but if you want to run 5 nodes, you can still do it. Unfortunately there's no easy way in Ansible to have an unbound upper limit for these kind of loops.

Makefile
melekes: `remotenet-start`. btw, why sentry and not plain network like local? how can I setup remote network without complicated sentry arch?

networks/remote/ansible/roles/config/tasks/main.yml
melekes: strange, but ok.

greg-szabo:
Latest commit contains the requested changes and an update for the ansible README (which was out of date).

Makefile
greg-szabo: You have to define "remote network" better. We decided that the only use-case for remote network is the sentry-node setup - which is a number of full nodes connecting to a previously setup validator node. There's no use-case yet for a set of validators in the cloud managed by one entity. The decision is so far, that if you want to be a validator, you build and deploy yourself.

greg-szabo:
A new feature that we worked on with Adrian (to ship logs to logz.io easily using Ansible) is dependent on this PR to be merged.

I've added the code on top of this PR, please merge it at your convenience.

networks/remote/ansible/logzio.yml
```diff
@@ -0,0 +1,13 @@
+---
+
+- hosts: all
+  user: root
+  any_errors_fatal: true
+  gather_facts: no
+  vars:
+    - service: tendermint
+    - JOURNALBEAT_BINARY: "{{lookup('env', 'GOPATH')}}/bin/journalbeat"
+#    - LOGZIO_TOKEN: ABCDEFGHIJKLMNOPQRSTUVWXYZ012345
```
melekes: leftover?

Makefile
melekes: > from server-setup and such to sentry-setup... What do you think?

let's rename to sentry-* .. (sentry-start, ...). because if we leave `remotenet`, people will think it works exactly like `localnet` but in a cloud, which is not true.

networks/remote/ansible/logzio.yml
greg-szabo: Not really. I kept it there so people understand that it is necessary to run the playbook. I can remove it and add a comment on top to make sure they add it.

Issue #1482, opened 2018-04-19
BUG REPORT: about abci.ABCIQuery parameters mismatch


WeibingChen:
Version = "0.19.0"
IN rpc/core/routes.go line 37
	// abci API
	"abci_query": rpc.NewRPCFunc(ABCIQuery, "path,data,height,prove"),

IN  rpc/core/abci.go line 50
func ABCIQuery(path string, data cmn.HexBytes, height int64, trusted bool) (*ctypes.ResultABCIQuery, error) 

the route defined that the ABCIQuery needs four params: "path,data,height,prove".
In other side the ABCIQuery func defined  four params: "path,data,height,trusted", in which trusted is not required and default value is false.

Problem:
When ABCIQuery call proxyApp the 'trusted' translated into ‘Prove' by operation !trusted.
So the abci app always gets the 'Prove' with true value.

Fix: change the routes definition to:
	// abci API
	"abci_query": rpc.NewRPCFunc(ABCIQuery, "path,data,height,trusted"),

PR #1485, opened 2018-04-22
Fix `Prove` int abci.ABCIQuery

kidinamoto01:
* [ ] Wrote tests
* [ ] Updated CHANGELOG.md
 
As discussed in https://github.com/tendermint/tendermint/issues/1482, we always get the prove when querying ABCI.  fix the query parameter to make sure we get the right result.


melekes:
```
Fix: change the routes definition to:
// abci API
"abci_query": rpc.NewRPCFunc(ABCIQuery, "path,data,height,trusted"),
```

the solution is to change route def. ^^^

kidinamoto01:
@melekes will this work?

melekes:
Yes, thank you 🍏

Issue #1481, opened 2018-04-18
rpc: /dump_consensus_state has some empty fields


ebuchman:
eg:

```
      "Votes": {},
      "CommitRound": -1,
      "LastCommit": {},
``` 

This seems to be a new post-amino issue where for some reason these aren't getting marshalled

ebuchman:
Fixed in https://github.com/tendermint/tendermint/pull/1511

Issue #1480, opened 2018-04-18
rpc: Include validator address under validator_info in /status


ebuchman:
/status now includes "validator_info" but not the address. Let's add the address.

Related to https://github.com/tendermint/tendermint/issues/1479

ebuchman:
Fixed in https://github.com/tendermint/tendermint/pull/1511

Issue #1199, opened 2018-02-08
rpc: node_info may return nil on start


melekes:
```
When we send a status rpc request via websocket immediately after establishing the websocket connection, node info will may be null while later calls will have a non null node info.

Delaying the call 5+ seconds yields expected results.

Looks like the other reactors are not fully up but rpc is already allowed ?
Does that male sense?

PS tendermint 0.12
```

Need to check / try to reproduce and fix.

srmo:
wanted to add the following:
- the Websocket connection is done by the ABCI app in a thread separated from the actual ABCI component
- this happens when it is the first connection between ABCI App and tendermint is established, i.e. consensus connection and websocket are a first connection
- didn't test if this also happens on websocket reconnects after consensus was already established

ebuchman:
Can you check if this is solved in #1511 ?

PR #1511, opened 2018-04-27
Various RPC fixes

ebuchman:
Fixes:

- https://github.com/tendermint/tendermint/issues/1481
- https://github.com/tendermint/tendermint/issues/1480
- https://github.com/tendermint/tendermint/issues/1310
- https://github.com/tendermint/tendermint/issues/1199 (?)

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -402,7 +402,7 @@ func (n *Node) OnStart() error {
 	// Add ourselves to addrbook to prevent dialing ourselves
 	n.addrBook.AddOurAddress(nodeInfo.NetAddress())
 
-	// Run the RPC server first
+	// Start the RPC server before the P2P server
 	// so we can eg. receive txs for the first block
 	if n.config.RPC.ListenAddress != "" {
 		listeners, err := n.startRPC()
@@ -412,7 +412,7 @@ func (n *Node) OnStart() error {
 		n.rpcListeners = listeners
 	}
 
-	// Start the switch
+	// Start the switch (the P2P server).
 	err = n.sw.Start()
 	if err != nil {
 		return err
--- a/node/node.go
+++ b/node/node.go
@@ -382,16 +382,6 @@ func (n *Node) OnStart() error {
 		return err
 	}
 
-	// Run the RPC server first
-	// so we can eg. receive txs for the first block
-	if n.config.RPC.ListenAddress != "" {
-		listeners, err := n.startRPC()
-		if err != nil {
-			return err
-		}
-		n.rpcListeners = listeners
-	}
-
 	// Create & add listener
 	protocol, address := cmn.ProtocolAndAddress(n.config.P2P.ListenAddress)
 	l := p2p.NewDefaultListener(protocol, address, n.config.P2P.SkipUPNP, n.Logger.With("module", "p2p"))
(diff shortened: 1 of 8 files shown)


node/node.go
```diff
@@ -412,6 +402,16 @@ func (n *Node) OnStart() error {
 	// Add ourselves to addrbook to prevent dialing ourselves
 	n.addrBook.AddOurAddress(nodeInfo.NetAddress())
 
+	// Run the RPC server first
+	// so we can eg. receive txs for the first block
```
melekes: first in relation to what?

types/part_set.go
```diff
@@ -264,3 +264,20 @@ func (ps *PartSet) StringShort() string {
 	defer ps.mtx.Unlock()
 	return fmt.Sprintf("(%v of %v)", ps.Count(), ps.Total())
 }
+
+func (ps *PartSet) MarshalJSON() ([]byte, error) {
+	if ps == nil {
+		return []byte("nil-PartSet"), nil
```
melekes: shouldn't this be a valid json? like `null`? or maybe `[]byte("\"nil-PartSet\"")`

types/part_set.go
```diff
@@ -264,3 +264,20 @@ func (ps *PartSet) StringShort() string {
...
+	}
+
+	ps.mtx.Lock()
+	defer ps.mtx.Unlock()
+
+	return cdc.MarshalJSON(struct {
+		CountTotal    string        `json:"count/total"`
```
melekes: why not make them 2 separate fields?

types/vote_set.go
```diff
@@ -445,6 +445,26 @@ func (voteSet *VoteSet) StringIndented(indent string) string {
 		indent)
 }
 
+// Marshal the VoteSet to JSON. Same as String(), just in JSON,
+// and without the height/round/type_ (since its already included in the votes).
+func (voteSet *VoteSet) MarshalJSON() ([]byte, error) {
+	voteStrings := make([]string, len(voteSet.votes))
+	for i, vote := range voteSet.votes {
```
melekes: what about 

```
voteSet.mtx.Lock()
defer voteSet.mtx.Unlock()
```

types/part_set.go
```diff
@@ -264,3 +264,20 @@ func (ps *PartSet) StringShort() string {
...
+		PartsBitArray *cmn.BitArray `json:"parts_bit_array"`
+	}{
+		fmt.Sprintf("%d/%d", ps.Count(), ps.Total()),
+		ps.partsBitArray.Copy(),
```
melekes: now that we have custom marshaller for BitArray's, I think we can drop `.Copy()` https://github.com/tendermint/tmlibs/blob/develop/common/bit_array.go#L327

node/node.go
ebuchman: to the p2p server we start after

types/part_set.go
ebuchman: yes it should. maybe we should just do `{}`

types/part_set.go
ebuchman: Figured there's no need to take up the extra space but maybe that's silly (dump consensus is a big dump already). Also we're typically more interested in the ratio than we are in the two numbers

types/part_set.go
ebuchman: ++

types/vote_set.go
ebuchman: probably right. String() doesn't have it tho

ebuchman:
Thanks for the review @melekes !

Issue #1476, opened 2018-04-17
Do not print errors that are not errors


melekes:
```
E[04-17|18:16:18.170] Failed to add peer                           module=p2p address=bb029f6756c2be3226a964abba495b8d1b8b44b3@159.65.211.28:46656 err="Duplicate peer"
E[04-17|18:16:18.170] Error dialing peer                           module=p2p err="Duplicate peer"
```

```
E[04-17|18:18:46.599] Failed to add new address                    module=p2p err="Cannot add ourselves with address 6905ab36d3afbfbac55da2fd90c5019807bccdf1@188.166.165.4:46656"
```

ebuchman:
We can certainly not log errors when we connect to ourselves or a duplicate peer, but what about all the other possible errors ? if  we connect to apeer with bad chainid or version, should we log an error ?

melekes:
I think this is still an issue though.

xla:
@melekes maybe we add it to the logging improvements planned in #1212

PR #1344, opened 2018-03-20
Notes and TODOs in the p2p layer

ebuchman:
Other things to keep in mind:

- avoid multiple connections to the same IP (from same or different IDs)
- better avoidance of connecting to self (for each of our addresses, we shouldnt end up trying to dial it more than once!)

PR #1519, opened 2018-04-28
Cleanup P2P

### LINKED ISSUE #1501: Clarify MaxNumPeers
MaxNumPeers only affects number of incoming connections. The number of outgoing connections is limited by address book - 1000 (needAddressThreshold). We should either rename it to `MaxNumInPeers` or rewrite the code.

ebuchman:
Replaces https://github.com/tendermint/tendermint/pull/1344

- reduce max pex msg size
- validate nodeInfo.ListenAddr
- clarify maxNumPeers (closes #1501) and update pex spec
- remove trust metric
- more explicit errors for addrbook.AddAddress and NewNetAddressFromString
- give NetAddress json tags in prep for using it in NodeInfo
- allow nodes to differ by minor version
- some cleanup
- a bunch of new comments/notes/todos
- a little #1476

* [ ] Wrote tests
* [x] Updated CHANGELOG.md

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -100,7 +100,6 @@ type Node struct {
 	// network
 	sw       *p2p.Switch  // p2p connections
 	addrBook pex.AddrBook // known peers
-	// trustMetricStore *trust.TrustMetricStore // trust metrics for all peers
 
 	// services
 	eventBus         *types.EventBus // pub/sub for services
--- a/node/node.go
+++ b/node/node.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"net"
 	"net/http"
-	"strings"
 
 	abci "github.com/tendermint/abci/types"
 	amino "github.com/tendermint/go-amino"
@@ -280,14 +279,6 @@ func NewNode(config *cfg.Config,
 		addrBook = pex.NewAddrBook(config.P2P.AddrBookFile(), config.P2P.AddrBookStrict)
 		addrBook.SetLogger(p2pLogger.With("book", config.P2P.AddrBookFile()))
 
-		var seeds []string
-		if config.P2P.Seeds != "" {
-			seeds = strings.Split(config.P2P.Seeds, ",")
-		}
-		var privatePeerIDs []string
-		if config.P2P.PrivatePeerIDs != "" {
-			privatePeerIDs = strings.Split(config.P2P.PrivatePeerIDs, ",")
-		}
 		// TODO persistent peers ? so we can have their DNS addrs saved
 		pexReactor := pex.NewPEXReactor(addrBook,
 			&pex.PEXReactorConfig{
--- a/node/node.go
+++ b/node/node.go
@@ -274,11 +274,11 @@ func NewNode(config *cfg.Config,
 	// if auth_enc=false.
 	//
 	// If PEX is on, it should handle dialing the seeds. Otherwise the switch does it.
+	// Note we currently use the addrBook regardless at least for AddOurAddress
 	var addrBook pex.AddrBook
+	addrBook = pex.NewAddrBook(config.P2P.AddrBookFile(), config.P2P.AddrBookStrict)
+	addrBook.SetLogger(p2pLogger.With("book", config.P2P.AddrBookFile()))
 	if config.P2P.PexReactor {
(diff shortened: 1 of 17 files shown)


p2p/pex/addrbook.go
```diff
@@ -294,29 +295,32 @@ func (a *addrBook) MarkBad(addr *p2p.NetAddress) {
...
 // GetSelection implements AddrBook.
 // It randomly selects some addresses (old & new). Suitable for peer-exchange protocols.
+// Must never return a nil address.
 func (a *addrBook) GetSelection() []*p2p.NetAddress {
 	a.mtx.Lock()
 	defer a.mtx.Unlock()
 
-	if a.size() == 0 {
+	bookSize := a.size()
+	if bookSize == 0 {
```
odeke-em: I get quite nervous when I see `== 0` for checks that result from additions of two entities that are independently subtracted, are ints and moreover don't have full guarantees to always be a minimum of 0.

Perhaps let's make that check `<= 0`?

I personally only trust the Go guarantee of `len` returning a minimum of 0 always since the compiler, runtime and spec guarantee that, but the rest nah.

p2p/pex/addrbook.go
```diff
@@ -338,7 +343,8 @@ func (a *addrBook) GetSelectionWithBias(biasTowardsNewAddrs int) []*p2p.NetAddre
...
```
odeke-em: Ditto about `<= 0`

p2p/pex/addrbook.go
```diff
@@ -487,19 +493,19 @@ func (a *addrBook) getBucket(bucketType byte, bucketIdx int) map[string]*knownAd
 
 // Adds ka to new bucket. Returns false if it couldn't do it cuz buckets full.
 // NOTE: currently it always returns true.
-func (a *addrBook) addToNewBucket(ka *knownAddress, bucketIdx int) bool {
+func (a *addrBook) addToNewBucket(ka *knownAddress, bucketIdx int) {
```
odeke-em: We also need to modify `addToOldBucket` then for consistency, right?

p2p/pex/addrbook.go
```diff
@@ -605,19 +609,25 @@ func (a *addrBook) pickOldest(bucketType byte, bucketIdx int) *knownAddress {
 // adds the address to a "new" bucket. if its already in one,
 // it only adds it probabilistically
 func (a *addrBook) addAddress(addr, src *p2p.NetAddress) error {
+	if addr == nil {
```
odeke-em: why not
```go
if addr == nil || src == nil {
```
since we return the exact same ordering of arguments?

rpc/core/net.go
```diff
@@ -47,9 +48,13 @@ func NetInfo() (*ctypes.ResultNetInfo, error) {
 			ConnectionStatus: peer.Status(),
 		})
 	}
+	// TODO: Should we include PersistentPeers and Seeds in here?
```
melekes: yeah, let's do this

node/node.go
```diff
@@ -99,9 +98,9 @@ type Node struct {
 	privValidator types.PrivValidator // local node's validator key
 
 	// network
-	sw               *p2p.Switch             // p2p connections
-	addrBook         pex.AddrBook            // known peers
-	trustMetricStore *trust.TrustMetricStore // trust metrics for all peers
+	sw       *p2p.Switch  // p2p connections
+	addrBook pex.AddrBook // known peers
+	// trustMetricStore *trust.TrustMetricStore // trust metrics for all peers
```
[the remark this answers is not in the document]
ebuchman: too complex to figure out how to use it effectiely right now . taking a simpler approach (see the pex.md for descriptioN)

p2p/pex/addrbook.go
ebuchman: we actually use the response from addToOldBucket ... but this code could all use a clean up ...

p2p/pex/addrbook.go
ebuchman: originally I used different errors, then consolidated them. thanks

rpc/core/net.go
ebuchman: opened an issue: https://github.com/tendermint/tendermint/issues/1526

Issue #1264, opened 2018-03-02
p2p: prevent multiple connections from same IP


ebuchman:
Currently we allow multiple connections from the same IP because we only enforce uniqueness via the ID.

We shouldn't allow more than one connection from the same IP. Perhaps we should even limit the number of connections from an IP block (we used to do this).

ebuchman:
Reference for why we should do this: https://bitcoinmagazine.com/articles/researchers-explore-eclipse-attacks-ethereum-blockchain/

odeke-em:
For the TL;DR:

<img width="609" alt="screen shot 2018-03-01 at 10 01 53 pm" src="https://user-images.githubusercontent.com/4898263/36885573-9df7e4b4-1d9c-11e8-8c2e-690a485047c0.png">

ebuchman:
Adding PoW to the handshake should actually help with this somewhat (https://github.com/tendermint/tendermint/issues/1045). At least that will make them do work for every connection.

ebuchman:
Started this in #1520

ebuchman:
Done in #1520 thanks @xla

PR #1520, opened 2018-04-28
p2p: prevent connections from same ip

ebuchman:
#1264 

TODO: fix and write tests

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -17,6 +17,7 @@ type Peer interface {
 	cmn.Service
 
 	ID() ID             // peer's cryptographic ID
+	RemoteIP() string   // remote IP of the connection
 	IsOutbound() bool   // did we dial the peer
 	IsPersistent() bool // do we redial this peer when we disconnect
 	NodeInfo() NodeInfo // peer's info
@@ -43,6 +44,15 @@ type peerConn struct {
 // NOTE: Will panic if conn is not *SecretConnection.
 func (pc peerConn) ID() ID {
 	return PubKeyToID(pc.conn.(*tmconn.SecretConnection).RemotePubKey())
+}
+
+// Return the IP from the connection RemoteAddr
+func (pc peerConn) RemoteIP() string {
+	host, _, err := net.SplitHostPort(pc.conn.RemoteAddr().String())
+	if err != nil {
+		panic(err)
+	}
+	return host
 }
 
 // peer implements Peer.
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -17,7 +17,7 @@ type Peer interface {
 	cmn.Service
 
 	ID() ID             // peer's cryptographic ID
-	RemoteIP() string   // remote IP of the connection
+	RemoteIP() net.IP   // remote IP of the connection
 	IsOutbound() bool   // did we dial the peer
 	IsPersistent() bool // do we redial this peer when we disconnect
 	NodeInfo() NodeInfo // peer's info
@@ -38,6 +38,7 @@ type peerConn struct {
 	persistent bool
 	config     *PeerConfig
 	conn       net.Conn // source connection
+	ips        []net.IP
 }
 
 // ID only exists for SecretConnection.
@@ -47,12 +48,24 @@ func (pc peerConn) ID() ID {
 }
 
 // Return the IP from the connection RemoteAddr
-func (pc peerConn) RemoteIP() string {
+func (pc peerConn) RemoteIP() net.IP {
+	if len(pc.ips) > 0 {
+		return pc.ips[0]
+	}
+
 	host, _, err := net.SplitHostPort(pc.conn.RemoteAddr().String())
 	if err != nil {
 		panic(err)
 	}
-	return host
+
+	ips, err := net.LookupIP(host)
+	if err != nil {
+		panic(err)
+	}
+
+	pc.ips = ips
+
+	return ips[0]
 }
 
 // peer implements Peer.
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -3,6 +3,7 @@ package p2p
 import (
 	"fmt"
 	"net"
+	"sync/atomic"
 	"time"
 
 	"github.com/tendermint/go-crypto"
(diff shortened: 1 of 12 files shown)


p2p/peer_set.go
```diff
@@ -76,6 +92,7 @@ func (ps *PeerSet) Get(peerKey ID) Peer {
 func (ps *PeerSet) Remove(peer Peer) {
 	ps.mtx.Lock()
 	defer ps.mtx.Unlock()
+	delete(ps.lookupIP[peer.RemoteIP()])
```
odeke-em: Does this compile?

p2p/peer_set.go
```diff
@@ -61,6 +68,15 @@ func (ps *PeerSet) Has(peerKey ID) bool {
 	return ok
 }
 
+// HasIP returns true iff the PeerSet contains
+// the peer referred to by this IP address.
+func (ps *PeerSet) HasIP(peerIP string) bool {
```
melekes: maybe we should use `net.IP` here?

xla: :+1:

p2p/peer.go
```diff
@@ -45,6 +46,15 @@ func (pc peerConn) ID() ID {
 	return PubKeyToID(pc.conn.(*tmconn.SecretConnection).RemotePubKey())
 }
 
+// Return the IP from the connection RemoteAddr
+func (pc peerConn) RemoteIP() string {
+	host, _, err := net.SplitHostPort(pc.conn.RemoteAddr().String())
+	if err != nil {
+		panic(err)
```
xla: Is this case worth a panic?

p2p/peer.go
```diff
@@ -45,6 +46,15 @@ func (pc peerConn) ID() ID {
 	return PubKeyToID(pc.conn.(*tmconn.SecretConnection).RemotePubKey())
 }
 
+// Return the IP from the connection RemoteAddr
+func (pc peerConn) RemoteIP() string {
```
xla: If we truly want to return an ip address I think we should do a lookup in case `RemoteAddr()` is a dns name.

ebuchman:
:o same test failure as anton is having in https://github.com/tendermint/tendermint/pull/1574#issuecomment-390768376

ebuchman:
Oh well, here at least presumably it's becuase they all have the same IP, so they disconnect when they realize thaT?

ebuchman:
maybe what we need is some kind of "filter" object that we can set that determines what peers we disconnect to ... that way in tests we don't need to apply the same level of filters so we can not worry about connecting to self when we're trying to test something else ... ?

p2p/switch.go
```diff
@@ -534,6 +534,8 @@ func (sw *Switch) addPeer(pc peerConn) error {
 		return err
 	}
 
+	// dont connect to multiple peers on the same IP
```
ebuchman: out of place ?

p2p/peer.go
```diff
@@ -45,6 +50,35 @@ func (pc peerConn) ID() ID {
 	return PubKeyToID(pc.conn.(*tmconn.SecretConnection).RemotePubKey())
 }
 
+// Return the IP from the connection RemoteAddr
+func (pc peerConn) RemoteIP() net.IP {
+	if len(pc.ips) > 0 {
+		return pc.ips[0]
+	}
+
+	if pc.conn == nil || pc.conn.RemoteAddr().String() == "pipe" {
```
ebuchman: is this only nil during test? we should add comment either way that this little section is just for tests.

p2p/peer.go
```diff
@@ -45,6 +50,35 @@ func (pc peerConn) ID() ID {
...
+		pc.ips = []net.IP{
+			net.IP{172, 16, 0, byte(atomic.AddUint32(&testIPSuffix, 1))},
+		}
+
+		return pc.ips[0]
+	}
+
+	host, _, err := net.SplitHostPort(pc.conn.RemoteAddr().String())
```
ebuchman: Does RemoteAddr return host names or only IPs ?

p2p/peer.go
```diff
@@ -37,6 +41,7 @@ type peerConn struct {
 	persistent bool
 	config     *PeerConfig
 	conn       net.Conn // source connection
+	ips        []net.IP
```
ebuchman: are you thinking we'll use this for something ? looks like we only ever use the [0]

p2p/switch.go
```diff
@@ -564,20 +566,21 @@ func (sw *Switch) addPeer(pc peerConn) error {
...
 	}
 
 	// Avoid duplicate
 	if sw.peers.Has(peerID) {
-		return ErrSwitchDuplicatePeer
+		return ErrSwitchDuplicatePeerID{peerID}
+	}
+
+	// check ips for both the connection addr and the self reported addr
+	if sw.peers.HasIP(pc.RemoteIP()) {
```
ebuchman: this is the connection address. Maybe we should check the self-reported address from the NodeInfo as well ?

p2p/peer_set_test.go
```diff
@@ -112,25 +124,38 @@ func TestPeerSetAddDuplicate(t *testing.T) {
...
-	wantErrCount, gotErrCount := n-1, errsTally[ErrSwitchDuplicatePeer]
+	wantErrCount, gotErrCount := n-1, errsTally["duplicateID"]
 	assert.Equal(t, wantErrCount, gotErrCount, "invalid ErrSwitchDuplicatePeer count")
 
-	wantNilErrCount, gotNilErrCount := 1, errsTally[nil]
+	wantNilErrCount, gotNilErrCount := 1, errsTally["other"]
 	assert.Equal(t, wantNilErrCount, gotNilErrCount, "invalid nil errCount")
 }
 
+func TestPeerSetAddDuplicateIP(t *testing.T) {
```
ebuchman: TODO

ebuchman:
Looks good. Some minor comments, but mostly just need to address the test issue before merge.

Note the test failure here actually has a switch with no peers and makes sense why its not working ( I think ). The fail in #1574 might be different

xla:
@ebuchman Thanks for the thorough review.

xla:
@ebuchman @melekes green at last and ready for a final review.

ebuchman:
LGTM!

PR #1529, opened 2018-04-30
Persistence test bash fix for relative folder path

greg-szabo:
This is a bugfix for the persistent tests running in bash. The `txs.sh` file (which is only used by the `test_failure_indices.sh` script) was called with a relative path. When calling the test case using docker, there is no easy way to switch directories. This patch will move the file to the same directory as the executing script, so it's possible to call it using absolute path.


ebuchman:
Can you explain how `$(dirname $0)` is supposed to work ? if I try to run that locally I get issues

greg-szabo:
`$0` - the path and name of the script running
`dirname $0` - the path of the running script
The file (`txs.sh`) invoked was moved to the same folder as the script that is invoking it. $(dirname $0) is supposed to give you the path to the script folder.

What issue do you see?

ebuchman:
I'm going to force push a rebase

PR #1534, opened 2018-05-03
update docs/examples & terraform/ansible

zramsay:
starts (but does not finish) #1533


networks/remote/ansible/README.rst
```diff
@@ -34,7 +34,7 @@ Build
     # Then copy your genesis.json and modify the config.toml as you see fit.
 
     # Reconfig the sentry nodes with a new BINARY and the configuration files from the build folder:
-    ansible-playbook -i inventory/digital_ocean.py -l sentrynet config.yml -e BINARY=`pwd`/build/tendermint -e CONFIGDIR=`pwd`/build
+    ansible-playbook -i inventory/digital_ocean.py -l sentrynet config.yml -e BINARY=$GOPATH/src/github.com/tendermint/tendermint/build/tendermint -e CONFIGDIR=$GOPATH/src/github.com/tendermint/tendermint/docs/examples
```
melekes: we're not necessary inside `GOPATH`.. so it should be `pwd`

networks/remote/ansible/roles/install/templates/systemd.service.j2
```diff
@@ -8,7 +8,7 @@ Restart=on-failure
 User={{service}}
 Group={{service}}
 PermissionsStartOnly=true
-ExecStart=/usr/bin/tendermint node --proxy_app=dummy
+ExecStart=/usr/bin/tendermint node --proxy_app=persistent_kvstore --p2p.persistent_peers=167b80242c300bf0ccfb3ced3dec60dc2a81776e@165.227.41.206:46656,3c7a5920811550c04bf7a0b2f1e02ab52317b5e6@165.227.43.146:46656,303a1a4312c30525c99ba66522dd81cca56a361a@159.89.115.32:46656,b686c2a7f4b1b46dca96af3a0f31a6a7beae0be4@159.89.119.125:46656
```
melekes: where did you get these IPs from? we don't know IPs at this point. /cc @greg-szabo

networks/remote/ansible/README.rst
zramsay: but it doesn't work with `pwd` because you're in `networks/remote/ansible` when running this command.

networks/remote/ansible/roles/install/templates/systemd.service.j2
zramsay: they're what terraform spun up. Agree that they shouldn't be hardcoded - this is placeholder.

networks/remote/ansible/README.rst
melekes: oh, I think that's because Greg copied this from the tools repo. Ok

networks/remote/terraform/cluster/variables.tf
```diff
@@ -5,7 +5,7 @@ variable "name" {
 variable "regions" {
   description = "Regions to launch in"
   type = "list"
-  default = ["AMS2", "FRA1", "LON1", "NYC3", "SFO2", "SGP1", "TOR1"]
+  default = ["TOR1"]
```
melekes: why this change?

zramsay: persistently get an error along the lines of "region not available" but terraform doesn't say which region. I know TOR1 works, so went with that rather than waste time tracking down which region is bunk.

melekes: ok. but we still probably want 4 nodes in 4 different AZ

zramsay: ah, the culprit is `AMS2`, which should be `AMS3`

melekes:
@zramsay are you still working on this?

zramsay:
nope. it's ready. title changed

chinajsstar:
in tendermint/docs/examples/init_testnet.sh,  the line of 36, 44, 52,60 the first ":" should delete. `\"power:\": 10,`  . Otherwise it can `if v.Power == 0 {
			return cmn.NewError("The genesis file cannot contain validators with no voting power: %v", v)
		}`

xla:
@chinajsstar Could you open a separate issue for this, please?

Issue #1165, opened 2018-01-27
Improve SecretConnection/STS KDF?


tarcieri:
SecretConnection presently uses the NaCl `crypto_secretbox` construction for authenticated encryption:

https://github.com/tendermint/tendermint/blob/master/p2p/secret_connection.go#L131

`crypto_secretbox` decomposes to the following:

```
Salsa20Poly1305(HSalsa20(k, nonce), Ø{8}, frame)
```

The `crypto_secretbox` API imposes a 24-byte length requirement on the nonce, and for that reason SecretConnection first pre-hashes all of the cryptographic parameters with RIPEMD160:

https://github.com/tendermint/tendermint/blob/master/p2p/secret_connection.go#L131

So the overall construction ends up looking like this:

```
Salsa20Poly1305(HSalsa20(k, RIPEMD160(lowPubKey || hiPubKey)), Ø{8}, frame)
```

(glossing over the sender/receiver nonce tweaks momentarily)

So, funny thing, my understanding is HSalsa20 is just a PRF and can accept variable length input. The 24-byte restriction is a totally arbitrary one imposed by the `crypto_secretbox` API. So if you wanted to get rid of RIPEMD160 ([see discussion here](https://github.com/cosmos/cosmos-sdk/issues/263#issuecomment-359580833)), you could just do this:

```
Salsa20Poly1305(HSalsa20(k, lowPubKey || hiPubKey), Ø{8}, frame)
```

Tada, no need for RIPEMD160. If you step outside the constraints of the `crypto_secretbox` API, you should just be able to pass everything directly to `HSalsa20`. This should provide a 2<sup>128</sup> security level, as opposed to the 2<sup>80</sup> one provided by RIPEMD160.

All that said, I would probably suggest using HKDF instead of HSalsa20 here. See some of the criticism CodesInChaos makes about the similar nonce-based approach used by CurveCP:

https://codesinchaos.wordpress.com/2012/09/09/curvecp-1/

A more traditional way to build an authenticated key exchange is to compute a transcript hash of the entire key exchange and use that as an input to something like HKDF, and using that to derive symmetric keys.

HKDF is an extremely boring construction with well-understood, provable security properties created by Hugo Krawczyk, who just so happened to be this year's Levchin Award winner at RealWorldCrypto. HKDF is used by both TLS and Noise for key derivation.

[...]

PR #1563, opened 2018-05-13
BREAKING CHANGE Replaces hash based key derivation with hkdf 

zmanian:
Resolves https://github.com/tendermint/tendermint/issues/1165


melekes:
```
+ counter_over_socket
+ rm -rf /home/circleci/.tendermint_app
test/app/test.sh: line 49:   416 Killed                  abci-cli kvstore > /dev/null
+ tendermint init
I[05-14|11:50:05.441] Generated private validator                  module=main path=/home/circleci/.tendermint_app/config/priv_validator.json
I[05-14|11:50:05.443] Generated node key                           module=main path=/home/circleci/.tendermint_app/config/node_key.json
I[05-14|11:50:05.444] Generated genesis file                       module=main path=/home/circleci/.tendermint_app/config/genesis.json
test/app/test.sh: line 49:   397 Killed                  tendermint node > tendermint.log
+ echo 'Starting counter_over_socket'
Starting counter_over_socket
+ pid_counter=497
+ abci-cli counter --serial
+ pid_tendermint=498
+ sleep 5
+ tendermint node
+ echo 'running test'
running test
+ bash counter_test.sh 'Counter over Socket'
... sending tx. expect no error
RESPONSE

-------
TX 00
RESPONSE 
ERROR 
IS_ERR 
----
Got non-zero exit code for 00. 
Exited with code 1
```

one of the tests failed for some reason

melekes:
also this needs rebase (or merge)

zmanian:
I am having trouble localizing what exactly is failing here

p2p/conn/secret_connection.go
```diff
@@ -313,22 +313,19 @@ func shareAuthSignature(sc *SecretConnection, pubKey crypto.PubKey, signature cr
...
+	return res
 }
 
-// We only fill in the first 20 bytes with ripemd160
 func hash24(input []byte) (res *[24]byte) {
-	hasher := ripemd160.New()
-	hasher.Write(input) // nolint: errcheck, gas
-	resSlice := hasher.Sum(nil)
+	hash := sha256.New
+	hkdf := hkdf.New(hash, input, nil, nil)
```
ebuchman: don't we need to pass the key in here as well to actually utilize it as HKDF ? otherwise we might as well just SHA256(input) ? 

Also I think Tony's recommending we use something like "TENDERMINT_SECRETCONNECTION_KDF" for the `info` param

zmanian: Adding a salt like that seems like a good practice.

The KDF is taking the constructor as an argument not the hash object.

p2p/conn/secret_connection.go
```diff
@@ -124,14 +123,18 @@ func (sc *SecretConnection) Write(data []byte) (n int, err error) {
...
 		copy(frame[dataLenSize:], chunk)
 
+		aead, err := xchacha20poly1305.New(sc.shrSecret[:])
+		if err != nil {
+			return n, errors.New("Invalid SecretConnection Key")
+		}
 		// encrypt the frame
-		var sealedFrame = make([]byte, sealedFrameSize)
-		secretbox.Seal(sealedFrame[:0], frame, sc.sendNonce, sc.shrSecret)
+		var sealedFrame = make([]byte, aead.Overhead()+totalFrameSize)
```
ebuchman: is there a way to get this overhead as a constant so we can bring back `sealedFrameSize` ?

zmanian: The AEAD api only makes it available as a function. The AEAD interface only exposes it as function and it's a go standard library interface

tarcieri: For ChaCha20Poly1305 it will always be 16 (the length of a Poly1305 tag): https://github.com/golang/crypto/blob/master/chacha20poly1305/chacha20poly1305.go#L46

ValarDragon:
We need to swap out the use of xchacha for https://github.com/tendermint/go-crypto/pull/135 once thats merged. The reason for this is due to rust having no permissively licensed xchacha implementation.

ebuchman:
> We need to swap out the use of xchacha for tendermint/go-crypto#135 once thats merged. The reason for this is due to rust having no permissively licensed xchacha implementation.

Is this correct? Can't we just let the KMS be GPL ?

tarcieri:
@ebuchman it's something I could probably implement in less than an hour inside of `chacha20_poly1305_aead`. The construction itself is trivial. Here's the libsodium version:

https://github.com/jedisct1/libsodium/blob/d26c8adf84a38086e4d31af593513ce6aba5a968/src/libsodium/crypto_core/hchacha20/core_hchacha20.c#L48-L57

(the rest of it is just handling endianness conversions)

Inside of the Apache 2-licensed pure Rust crate, I believe those `QUARTERROUND` invocations in the C code should map to:

https://github.com/cesarb/chacha20-poly1305-aead/blob/master/src/chacha20.rs#L41

All that said, in the "box" constructions, HChaCha20 is just being used as a KDF (albeit a limited one due to the fixed nonce and output size). HKDF can effectively replace it.

ValarDragon:
For compute shared secret, we're currently using box.Precompute(https://github.com/golang/crypto/blob/master/nacl/box/box.go#L70) which uses `salsa.Hsalsa20`. Should we switch this to HKDF?

zmanian:
Were you think of changing 

```
func computeSharedSecret(remPubKey, locPrivKey *[32]byte) (shrSecret *[32]byte) {
	shrSecret = new([32]byte)
	box.Precompute(shrSecret, remPubKey, locPrivKey)
	return
}
```

to something more like
```
func computeSharedSecret(remPubKey, locPrivKey *[32]byte) (shrSecret *[32]byte) {
        h
	shrSecret = new([32]byte)
        shrKey= new([32]byte)
	curve25519.ScalarMult(shrKey, locPrivKey, remPubKey)
        hash := sha256.New
        hkdf := hkdf.New(hash, shrKey, nil, []byte("TENDERMINT_SECRET_CONNECTION_SHARED_SECRET_GEN")
       io.ReadFull(hkdf, shrSecret[:])
        
	return shrSecret[:]
}
```

ValarDragon:
Yup! Do you think we should do that, to remove the HSalsa primitive?

zmanian:
yeah I like that

tarcieri:
@ValarDragon with that change you can get rid of the `hash24` here (and therefore, altogether), just switching to a 64-byte (or potentially longer) nonce: https://github.com/tendermint/tendermint/blob/5b07b2c818630ad2527422aa71dfc0275d72b6a1/p2p/conn/secret_connection.go#L263

ValarDragon:
Closing this in favor of #2039. We'll probably revive / rebase this branch and update the code when working on that.

Issue #909, opened 2017-11-30
rpc: limit return from TxSearch


ebuchman:
https://github.com/tendermint/tendermint/pull/835/files#diff-73e887135989720e25611c7b8504c023R181

PR #1567, opened 2018-05-14
limit /tx_search output

melekes:
Refs #909

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/rpc/core/tx.go
+++ b/rpc/core/tx.go
@@ -188,7 +188,7 @@ func TxSearch(query string, prove bool, page, perPage int) (*ctypes.ResultTxSear
 	}
 
 	totalCount := len(results)
-	page = validatePage(page)
+	page = validatePage(page, perPage, totalCount)
 	perPage = validatePerPage(perPage)
 	skipCount := (page - 1) * perPage
 
--- a/rpc/core/tx.go
+++ b/rpc/core/tx.go
@@ -6,6 +6,7 @@ import (
 	ctypes "github.com/tendermint/tendermint/rpc/core/types"
 	"github.com/tendermint/tendermint/state/txindex/null"
 	"github.com/tendermint/tendermint/types"
+	cmn "github.com/tendermint/tmlibs/common"
 	tmquery "github.com/tendermint/tmlibs/pubsub/query"
 )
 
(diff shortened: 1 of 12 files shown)


melekes:
> We should also probably do this for /unconfirmed_txs, even though we're bounding the mempool, because the bound will be high.

yes, but only `per_page` / `limit` because mempool is constantly changing. don't see how someone will be able to paginate the unconfirmed txs.

ebuchman:
> Don't see how someone will be able to paginate the unconfirmed txs.

Right. So do we just make it not possible to see all the txs ?

melekes:
> So do we just make it not possible to see all the txs ?

Do we have a choice? The opposite is to allow unlimited txs, which can result in 10000? (mempool limit) entries in a worst case? Someone can probably exploit this. What's the use case for this endpoint? To ensure the tx I've just sent made it to the mempool?

ebuchman:
> What's the use case for this endpoint? To ensure the tx I've just sent made it to the mempool?

Yeh, or to get a diagnostic of whats in the mempool.

Maybe we just need to make sure we can search the mempool for txs by hash and move /unconfirmed_Txs to  /unsafe_unconfirmed_txs

melekes:
mm, actually calling `/unconfirmed_tx?hash` will result in whole list traversal (expensive!). unless we have a `map[hash]*pointer to CElement`. but not sure it's worth creating a map just because of 1 endpoint, which did not prove itself yet

rpc/core/tx.go
```diff
@@ -182,11 +187,15 @@ func TxSearch(query string, prove bool) ([]*ctypes.ResultTx, error) {
...
+	totalCount := len(results)
+	page = validatePage(page)
+	perPage = validatePerPage(perPage)
+	skipCount := (page - 1) * perPage
+
+	apiResults := make([]*ctypes.ResultTx, cmn.MinInt(perPage, totalCount-skipCount))
 	var proof types.TxProof
-	for i, r := range results {
+	for i := 0; i < len(apiResults); i++ {
+		r := results[skipCount+i]
```
ebuchman: this could be out of bounds, since page is not upper bounded

Issue #847, opened 2017-11-15
[pubsub] Prioritise internal subscribers (e.g. reactor) over external (e.g. RPC)


melekes:
Right now when we get the msg, we iterate over map (of queries -> subscribers). If there are 1000 subscribers, consensus reactor could receive his msg last.

Ways to solve this:
1) return back old events system for internal events and proxy events to the new pubsub
2) add some priority/sorting to the new pubsub
3) add special kind of subscribers (VIP) to the new pubsub

ebuchman:
Decision:

- use the old `events` for consensus internal stuff
- use new `pubsub` for RPC

PR #1574, opened 2018-05-15
[pubsub] Prioritise internal subscribers (e.g. reactor) over external (e.g. RPC)

melekes:
Refs #847

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


consensus/reactor.go
```diff
@@ -49,6 +48,12 @@ func NewConsensusReactor(consensusState *ConsensusState, fastSync bool) *Consens
 		conS:     consensusState,
 		fastSync: fastSync,
 	}
+	// XXX: modifing state to send us new round steps, votes and proposal heartbeats
+	consensusState.reactorChs = &reactorChs{
```
melekes: this is not good, isn't it? should we bring `events` pkg to avoid doing this?

melekes:
FOR REVIEW: https://github.com/tendermint/tendermint/pull/1574/files/d74efa6ac7779760e5f7209c7e2f391be9c53f4b..fffc9c707bcfd9c34b5c332584358edf37bea948

consensus/state.go
```diff
@@ -509,6 +520,7 @@ func (cs *ConsensusState) newStep() {
 	// newStep is called by updateToStep in NewConsensusState before the eventBus is set!
 	if cs.eventBus != nil {
 		cs.eventBus.PublishEventNewRoundStep(rs)
+		cs.evsw.FireEvent(types.EventNewRoundStep, &cs.RoundState)
```
ebuchman: why not pass rs here ?

ebuchman:
> I had to alter events package

What required this ?

melekes:
> What required this ?

Not sure how it was done before, but the problem is in order to catch first NewRoundStep I need to subscribe to them before SwitchToConsensus emits it (https://github.com/tendermint/tendermint/pull/1574/commits/30b56886cba7d4f8d608d22e517c07408d566cd8#diff-f50d554d9c7f92cf7686b247b8c91419L85).  See https://github.com/tendermint/tendermint/pull/1574/commits/30b56886cba7d4f8d608d22e517c07408d566cd8

ebuchman:
Sounds good. Test failure:

```
--- FAIL: TestPEXReactorRunning (10.16s)
	pex_reactor_test.go:409: expected all switches to be connected to at least one peer (switches: 0 => {outbound: 0, inbound: 1}, 1 => {outbound: 1, inbound: 1}, 2 => {outbound: 2, inbound: 0}, )
FAIL
coverage: 73.7% of statements
FAIL	github.com/tendermint/tendermint/p2p/pex	19.873s
Exited with code 1
```

I can reproduce locally on this branch but not on develop. What's that about !?

ebuchman:
Oh the message is wrong - it actually expects everyone to be fully connected.  And now I can't reproduce it ...

ebuchman:
Hmm I reran the test on Circle and it timedout of the consensus test.

Ran the consensus tests locally and it failed on `TestMempoolProgressAfterCreateEmptyBlocksInterval`

Looks like this might be introducing some non-determinism ?

melekes:
Hmm.. I will take a look tmrw

melekes:
> Looks like this might be introducing some non-determinism ?

maybe, but how? all that got changed is the way CS reactor gets its events.

ebuchman:
Yeh especially given the test failure in p2p. Maybe the change is just helping surface previously lurking non-determinism. We're also seeing it in https://github.com/tendermint/tendermint/pull/1607

PR #1576, opened 2018-05-15
move spec to its final location

zramsay:
ref: https://github.com/tendermint/coding/issues/58
- makes consistent with Cosmos SDK's spec location
- still ignored by RTD
- a handful of links through both orgs can be updated


ebuchman:
We should keep new-spec/README.md for now with a link to the new location ... I'll add it

Issue #1565, opened 2018-05-14
[docs] Missing diagram


melekes:
https://tendermint.readthedocs.io/en/master/app-architecture.html#custom-abci-server

"The general architecture is shown in the following diagram:"

PR #1577, opened 2018-05-15
docs: small fixes

zramsay:
ref: #1565 

- all existing diagrams are already used, didn't want to duplicate. 
- fixed some lists


docs/app-architecture.rst
```diff
@@ -89,17 +88,20 @@ a message if my balance drops below 500). All "writes" sent to this
 server are proxied via websocket/JSON-RPC to tendermint core. When they
 come back as deliver\_tx over ABCI, they will be written to the data
 store. For "reads", we can do any queries we wish that are supported by
-our architecture, using any web technology that is useful. The general
-architecture is shown in the following diagram:
```
melekes: Let's copy image from https://github.com/tendermint/tendermint/issues/1565#issuecomment-389471110

docs/app-architecture.rst
```diff
@@ -92,14 +91,19 @@ store. For "reads", we can do any queries we wish that are supported by
...
+- Allows much richer, more flexible client-facing API
+- Allows pub-sub, watching certain fields, etc.
 
-Cons: \* Access to ABCI app can be dangerous (be VERY careful not to
-write unless it comes from the validator node) \* No direct access to
-the blockchain headers to verify tx \* You must write your own API (but
-maybe that's a pro...)
+Cons:
+
+- Access to ABCI app can be dangerous (be VERY careful not to write unless it comes from the validator node)
```
wolfposd: isnt it more dangerous to expose the tendermint-rpc endpoint to the world?
this way it can be limitied to localhost with a verified ABCI-App that can check transactions before forwarding them to the rpc-endpoint?

ebuchman: point is that any way to write to the ABCI app that doesn't come through tendermint consensus could cause a consensus failure since other nodes won't know about it.

Tendermint RPC doesnt expose any kind of unsafe or writable state, just some basic info and a way to send txs to the app, so its safe to expose, though one should put a rate limiter and HTTPS infront if its going to be on public internet ...

wolfposd: so maybe it should be "...be VERY careful not to write data which changes the appstate unless it comes over the ABCI-Socket"

ebuchman: updated here - https://github.com/tendermint/tendermint/pull/1636

would love more feedback on how to make the docs better. also i think we're changing this away from how jabci apps are built and towards how we're building the Cosmos stuff

PR #1585, opened 2018-05-17
Mempool cache

melekes:


docs/transactional-semantics.rst
```diff
@@ -0,0 +1,27 @@
...
+Tendermint"<./specification/using-tendermint.html#broadcast-api>`__ we
+discussed different API endpoints for sending transactions and
+differences between them.
+
+What we have not yet covered is transactional semantics.
+
+When you send a transaction using one of the available methods, it
+first goes to the mempool. Currently, it does not provide strong
+guarantees like "if the transaction were accepted, it would be
+eventually included in a block (given CheckTx passes)."
```
ebuchman: we should mention that if you want the guarantees, you can just use `_commit` (which will either be committed or return an error, unlike _sync), but then you have to deal with higher latency. This is the tradeoff.

mempool/mempool.go
```diff
@@ -497,5 +497,6 @@ func (cache *txCache) Push(tx types.Tx) bool {
 func (cache *txCache) Remove(tx types.Tx) {
 	cache.mtx.Lock()
 	delete(cache.map_, string(tx))
+	// so the list grows unboundedly? shouldn't we remove tx from the list here too?
```
ebuchman: the list is capped in `Push`, where we remove from the map and list. So it's ok if we don't remove from the list in Remove, but I don't really see why we wouldn't (maybe performance ?).

docs/transactional-semantics.rst
melekes: I assumed people have read "Using tendermint Broadcast API" where we explain the difference. But I guess it's worth repeating. Will do

mempool/mempool.go
melekes: assuming an attacker creates txs that do not pass CheckTx, list grows unboundedly.

docs/transactional-semantics.rst
ebuchman: Can we not consolidate this doc with that one ?

mempool/mempool.go
ebuchman: I don't understand why another transaction got added after CheckTx fails.

docs/transactional-semantics.rst
melekes: I thought this one will be more technical (mentioning things like WAL, ...)

mempool/mempool.go
melekes: I was wrong.

Issue #1532, opened 2018-05-03
use `make localnet` instead of current p2p bash setup scripts


zramsay:

melekes:
??? why do we need to test that

zramsay:
because it didn't work out of the box for @ebuchman

ebuchman:
probably it should be used for the `test_p2p` instead of the current thing

ebuchman:
but I still can't get it working on my mac :(. Keep coming up against: 

```
node3    | panic: Error initializing DB: sync /tendermint/node3/data/blockstore.db: invalid argument
node3    | 
node3    | goroutine 1 [running]:
node3    | github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/db.NewDB(0xc7f7b2, 0xa, 0xc420035cb0, 0x7, 0xc420961f20, 0x16, 0xc4208ff840, 0x1a)
node3    | 	/Users/ethanbuchman/programming/goApps/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/db/db.go:36 +0x355
```

ebuchman:
I'm on master.

And I've tried setting the permissions in many different ways to no avail. 

Here's what my folders look like locally:

```
ethanbuchman:tendermint$ ls -la build/node0/
total 8
drwxrwxrwx  5 ethanbuchman  staff  170  6 May 11:30 .
drwxrwxrwx  7 ethanbuchman  staff  238  6 May 11:29 ..
drwxrwxrwx  6 ethanbuchman  staff  204  6 May 11:29 config
drwxrwxrwx  3 ethanbuchman  staff  102  6 May 11:31 data
-rwxrwxrwx  1 ethanbuchman  staff  129  6 May 11:31 tendermint.log
ethanbuchman:tendermint$ ls -la build/node0/data/
total 0
drwxrwxrwx  3 ethanbuchman  staff  102  6 May 11:31 .
drwxrwxrwx  5 ethanbuchman  staff  170  6 May 11:30 ..
drwxr-xr-x  5 ethanbuchman  staff  170  6 May 11:31 blockstore.db
ethanbuchman:tendermint$ ls -la build/node0/data/blockstore.db/
total 16
drwxr-xr-x  5 ethanbuchman  staff  170  6 May 11:31 .
drwxrwxrwx  3 ethanbuchman  staff  102  6 May 11:31 ..
[...]

PR #1594, opened 2018-05-18
test cleanup

zramsay:
First pass at updating the bash tests with the goal of replacing [test/p2p/local_testnet_start.sh](https://github.com/tendermint/tendermint/blob/master/test/p2p/local_testnet_start.sh), see #1532. Will make seperate PRs piece by piece


docker-compose.yml
```diff
@@ -8,9 +8,9 @@ services:
       - "46656-46657:46656-46657"
     environment:
       - ID=0
-      - LOG=${LOG:-tendermint.log}
+      - LOG=$${LOG:-tendermint.log}
     volumes:
-      - ${FOLDER:-./build}:/tendermint:Z
+      - ./build:/tendermint:Z
```
ebuchman: @greg-szabo do we need these FOLDER vars?

zramsay: it threw a syntax error with the change as the recommended fix

Issue #1617, opened 2018-05-23
docs: use absolute links


ebuchman:
https://github.com/tendermint/tendermint/pull/1611

PR #1619, opened 2018-05-23
clean up links & spec docs

### LINKED ISSUE #1609: Update/remove consensus/README.md
this file: https://github.com/tendermint/tendermint/blob/master/consensus/README.md
hasn't been touched in nearly 2 years - probably out of date.

@ebuchman can it be `rm`'d?

zramsay:
closes #1609, #1617 

* [N/A] Updated all code comments where relevant
* [N/A] Wrote tests
* [N/A] Updated CHANGELOG.md


docs/spec/reactors/mempool/README.md
```diff
@@ -1,11 +0,0 @@
-# Mempool Specification
```
melekes: why remove? was it moved somewhere?

zramsay: not moved. The only thing in this file is a ToC telling the reader about the 4 files in this directory. For consistency we should either have a README like this in every directory or none at all. The former option doesn't make much sense IMO

docs/spec/README.md
```diff
@@ -14,21 +14,21 @@ please submit them to our [bug bounty](https://tendermint.com/security)!
 
 ### Data Structures
 
-- [Encoding and Digests](./blockchain/encoding.md)
-- [Blockchain](./blockchain/blockchain.md)
-- [State](./blockchain/state.md)
+- [Encoding and Digests](https://github.com/tendermint/tendermint/blob/master/docs/spec/blockchain/encoding.md)
```
ebuchman: oh goodness! I'm sorry, by absolute I meant with respect to the repo. so just `/docs/spec/...` is fine. as long as the first `/` means `top of this repo`. Full links is probably excessive

docs/spec/reactors/mempool/README.md
ebuchman: maybe we can at least try to pattern all the reactor dirs the same way, so its always like `config.md, messages.md, state.md` or something

PR #1628, opened 2018-05-24
evidence: dont send evidence to unsynced peers

### LINKED ISSUE #1624: Evidence reactor disconnects peers if it cant find height for evidence
This is a problem for nodes that aren't caught up - they will receive evidence for blocks they don't have yet, think the evidence is invalid, and thus disconnect the peer: https://github.com/tendermint/tendermint/blob/master/evidence/reactor.go#L99

Maybe we need to limit this behaviour during fast sync ? 

Might be related to https://github.com/tendermint/tendermint/issues/1623

ebuchman:
* only send evidence to peers that are synced enough to validate it all
* closes #1624

* [x] Wrote tests
* [x] Updated CHANGELOG.md


evidence/reactor.go
```diff
@@ -118,21 +119,48 @@ func (evR *EvidenceReactor) broadcastRoutine() {
 		case evidence := <-evR.evpool.EvidenceChan():
 			// broadcast some new evidence
 			msg := &EvidenceListMessage{[]types.Evidence{evidence}}
-			evR.Switch.Broadcast(EvidenceChannel, cdc.MustMarshalBinaryBare(msg))
+			evR.broadcastEvidenceListMsg(msg)
 
-			// TODO: Broadcast runs asynchronously, so this should wait on the successChan
-			// in another routine before marking to be proper.
+			// TODO: the broadcast here is just doing TrySend.
+			// We should make sure the send succeeds before marking broadcasted.
```
melekes: maybe we should use `Send` instead of `TrySend` then?

evidence/reactor.go
```diff
@@ -118,21 +119,48 @@ func (evR *EvidenceReactor) broadcastRoutine() {
...
 			evR.evpool.evidenceStore.MarkEvidenceAsBroadcasted(evidence)
 		case <-ticker.C:
 			// broadcast all pending evidence
 			msg := &EvidenceListMessage{evR.evpool.PendingEvidence()}
-			evR.Switch.Broadcast(EvidenceChannel, cdc.MustMarshalBinaryBare(msg))
+			evR.broadcastEvidenceListMsg(msg)
```
melekes: should we mark evidence as broadcasted here too?

evidence/reactor.go
```diff
@@ -118,21 +119,48 @@ func (evR *EvidenceReactor) broadcastRoutine() {
...
+
+	var maxHeight int64
+	for _, ev := range msg.Evidence {
+		if ev.Height() > maxHeight {
+			maxHeight = ev.Height()
+		}
+	}
+
+	for _, peer := range evR.Switch.Peers().List() {
+		ps := peer.Get(types.PeerStateKey).(PeerState)
```
ebuchman: E[05-26|15:40:21.860] Stopping peer for error                      module=p2p peer="Peer{MConn{13.124.200.97:37036} 0d04322b0637d8bf81c94e263ae7ea056c922d0f in}" err="Error{`recovered panic in MConnection` (cause: interface conversion: interface {} is nil, not *consensus.PeerState)}"

evidence/reactor_test.go
```diff
@@ -130,8 +130,48 @@ func TestReactorBroadcastEvidence(t *testing.T) {
...
+
+type peerState struct {
+	height int64
+}
+
+func (ps peerState) GetHeight() int64 {
+	return ps.height
+}
+
+func TestReactorSelectiveBroadcast(t *testing.T) {
```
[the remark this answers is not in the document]
ebuchman: fixed

PR #1634, opened 2018-05-28
Introduce local option for peerconfig

xla:
In line with other options we use for testing in the peerConfig we add the Local flag which signals to use the LocalAddr from the connection instead of the RemoteAddr. We require this as we piggyback on the loopback (127.0.0/8) addresses to simulate multiple peers.

### DIFF
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -341,6 +341,11 @@ func (sw *Switch) SetAddrBook(addrBook AddrBook) {
 	sw.addrBook = addrBook
 }
 
+// SetPeerConfig sets the PeerConfig used for peer setup.
+func (sw *Switch) SetPeerConfig(config *PeerConfig) {
+	sw.peerConfig = config
+}
+
 // MarkPeerAsGood marks the given peer as good when it did something useful
 // like contributed to consensus.
 func (sw *Switch) MarkPeerAsGood(peer Peer) {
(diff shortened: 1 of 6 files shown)


p2p/switch.go
```diff
@@ -341,6 +341,11 @@ func (sw *Switch) SetAddrBook(addrBook AddrBook) {
 	sw.addrBook = addrBook
 }
 
+// SetPeerConfig sets the PeerConfig used for peer setup.
+func (sw *Switch) SetPeerConfig(config *PeerConfig) {
```
melekes: maybe while you're working on this you can try to tackle https://github.com/tendermint/tendermint/blob/8393d4717fb1cfff28a9eaff2407165cdae7d9a3/p2p/switch.go#L93

would be good to avoid having this setter (SetPeerConfig).

xla: @melekes I'd like to do this in a separate follow-up PR, can we approve this one for now?

ebuchman:
We should add a config option to enable connections to the same IP. Users might want it. We can use that for the tests too

p2p/switch.go
```diff
@@ -341,6 +341,11 @@ func (sw *Switch) SetAddrBook(addrBook AddrBook) {
...
+	sw.peerConfig = config
```
melekes: looks racy to me...

p2p/switch.go
melekes: ok, sure

xla:
Closed in favour of #1646

PR #1636, opened 2018-05-28
gut docs/app-arch

ebuchman:
Motivated by https://github.com/tendermint/tendermint/pull/1577#discussion_r189188468 and other recent comments/work on the docs. This file seemed like it could be much reduced and improved ...

This would eliminate the nice diagram from https://tendermint.readthedocs.io/en/master/app-architecture.html#, but I'm not sure we should recommend that approach since we want people to find the LCD and the Interchain Standards sooner.

Should we bring the fancy network image into the repo, or is that link stable @nylira ?

Note it links to some open PRs and should be updated when theyre merged.

P.S Tendermint docs might be way better if they were organized into higher level sections:

- install
- tendermint core (CLI, RPC, config, genesis, running in production, spec, etc.)
- tendermint networks (local cluster, remote cluster, cosmos testnet)
- application dev (first app, abci, app guides, etc.)

@jolesbi @zramsay

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -139,7 +139,7 @@ jobs:
             for pkg in $(go list github.com/tendermint/tendermint/... | grep -v /vendor/ | circleci tests split --split-by=timings); do
               id=$(basename "$pkg")
 
-              go test -timeout 5m -race -coverprofile=/tmp/workspace/profiles/$id.out -covermode=atomic "$pkg"
+              GOCACHE=off go test -v -timeout 5m -race -coverprofile=/tmp/workspace/profiles/$id.out -covermode=atomic "$pkg"
             done
       - persist_to_workspace:
           root: /tmp/workspace
--- a/docs/app-architecture.rst
+++ b/docs/app-architecture.rst
@@ -1,133 +1,42 @@
 Application Architecture Guide
 ==============================
 
-Overview
---------
+Here we provide a brief guide on the recommended architecture of a Tendermint blockchain
+application.
 
-A blockchain application is more than the consensus engine and the
-transaction logic (eg. smart contracts, business logic) as implemented
-in the ABCI app. There are also (mobile, web, desktop) clients that will
-need to connect and make use of the app. We will assume for now that you
-have a well designed transactions and database model, but maybe this
-will be the topic of another article. This article is more interested in
-various ways of setting up the "plumbing" and connecting these pieces,
-and demonstrating some evolving best practices.
+The following diagram provides a superb example:
 
-Security
---------
+https://lh3.googleusercontent.com/eOT9KuaBjv7xKBg2xN8u2O1nU2Iw1-6PtFXrBSFWW8LYhfAMu84EwXW4RwyOnuiNqCoOoqXxE8Pkhr4Fyq9f=w2559-h1303-rw
 
-A very important aspect when constructing a blockchain is security. The
-consensus model can be DoSed (no consensus possible) by corrupting 1/3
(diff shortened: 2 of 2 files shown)


docs/app-architecture.rst
```diff
@@ -1,133 +1,42 @@
...
-consensus model can be DoSed (no consensus possible) by corrupting 1/3
-of the validators and exploited (writing arbitrary blocks) by corrupting
-2/3 of the validators. So, while the security is not that of the
-"weakest link", you should take care that the "average link" is
-sufficiently hardened.
+The end-user application here is the Cosmos Voyager, at the bottom left.
+Voyager communicates with a REST API exposed by a local Light-Client Daemon.
+The Light-Client Daemon is an application specific program that communciates with
+Tendermint nodes and verifies Tendermint light-client proofs through the Tendermint Core RPC.
+The Tendermint Core process communicates with a local ABCI application, where the
```
melekes: communicates

.circleci/config.yml
```diff
@@ -139,7 +139,7 @@ jobs:
             for pkg in $(go list github.com/tendermint/tendermint/... | grep -v /vendor/ | circleci tests split --split-by=timings); do
               id=$(basename "$pkg")
 
-              go test -timeout 5m -race -coverprofile=/tmp/workspace/profiles/$id.out -covermode=atomic "$pkg"
+              GOCACHE=off go test -v -timeout 5m -race -coverprofile=/tmp/workspace/profiles/$id.out -covermode=atomic "$pkg"
```
xla: Looks like an unrelated change. Also we don't cache the test results, so this will not change anything.

docs/app-architecture.rst
```diff
@@ -1,133 +1,42 @@
...
-need to connect and make use of the app. We will assume for now that you
-have a well designed transactions and database model, but maybe this
-will be the topic of another article. This article is more interested in
-various ways of setting up the "plumbing" and connecting these pieces,
-and demonstrating some evolving best practices.
+The following diagram provides a superb example:
 
-Security
---------
+https://lh3.googleusercontent.com/eOT9KuaBjv7xKBg2xN8u2O1nU2Iw1-6PtFXrBSFWW8LYhfAMu84EwXW4RwyOnuiNqCoOoqXxE8Pkhr4Fyq9f=w2559-h1303-rw
```
xla: Throws a `403` for me.

zramsay:
> Tendermint docs might be way better if they were organized into higher level sections

yeah, let's do it. They've grown a bit since the original layout was made

docs/app-architecture.rst
```diff
@@ -1,133 +1,42 @@
...
-transaction-processing speed is limited by the speed of the abci app and
-the number of validators, one should be able to scale our read
-performance to quite an extent (until the replication process drains too
-many resources from the validator nodes).
+See the following for more extensive documentation:
+- [Interchain Standard for the Light-Client REST API](https://github.com/cosmos/cosmos-sdk/pull/1028)
+- [Tendermint RPC Docs](https://tendermint.github.io/slate/)
+- [Tendermint in Production](https://github.com/tendermint/tendermint/pull/1618)
+- [Tendermint Basics](https://tendermint.readthedocs.io/en/master/using-tendermint.html)
+- [ABCI spec](https://github.com/tendermint/abci/blob/master/specification.rst)
```
zramsay: need rst link format

ebuchman: ugh. spoke to @jolesbi yesterday and sounds like we're moving back to markdown. no one should have to think about rst

PR #1657, opened 2018-05-31
Switch away from ephemeral ports

ValarDragon:
This moves all ports of the form 466xx to 266xx. This is done to address cosmos/cosmos-sdk#932. Essentially the problem was the default ports used by tendermint are in the ephemeral port range. 

This requires tendermint/abci#253


xla:
@ValarDragon We want this to be part of `0.21.0` if I got it right, could you bring it up to par with `develop`? :>

ValarDragon:
Will do!

ValarDragon:
Synced with develop. I accidentally merged abci's port change into develop, so I changed the abci referrence to develop

ValarDragon:
Closed in favor of #1735

PR #1670, opened 2018-06-01
Extract priv_validator into first class package

xla:
This is a maintenance change to move the private validator package out of the types and to a top-level location. There is no good reason to keep it under the types and it will more clearly coommunicate where additions related to the privval belong. It leaves the interface and the mock in types for now as it would introduce circular dependency between privval and types, this should be resolved eventually.

* move priv_validator to privval pkg
* use consistent `privval` as import

Follow-up to #1255


xla:
@ebuchman @melekes Open question is: Should start moving things to `pkg` as hinted at in tendermint/coding#61?

xla:
CI seems to fail for unrelated reasons.

melekes:
> @ebuchman @melekes Open question is: Should start moving things to pkg as hinted at in tendermint/coding#61?

you mean dissolve `types` package? I don't think we should spend our time on this before launch. you disagree?

xla:
@melekes In general refactoring proposals should be done in an incremental way, whenever parts are touched that would be affected by it anyway. As an ongoing process so to say, so not advocating to do the entire move right now, but privval was changed anyway and was special in that it was the only sub-package under types.

PR #1671, opened 2018-06-01
Remove auth_enc config option

### LINKED ISSUE #1518: Remove AuthEnc config option
It would be a nice boost to the cleanliness of some p2p code if we didn't even support this option. 

The consequence is that everyone will have to have IDs along with address in order to connect to peers. This should effectively be true anyways. But maybe some folks wanted to use their own VPN for encrypting connections and not have Tendermint do another layer of encryption ...

I think it's probably worth it to remove the option to turn this off, so that Tendermint always provides authenticated encryption and you need to know the ID of the peers you're dialing.

If you really think you need to set AuthEnc=false, please speak now

xla:
As we didn't hear any voices requesting this feature, we removed the
option to disable it and always have peer connection auth encrypted.

closes #1518
follow-up #1325

* [ ] Wrote tests
* [x] Updated CHANGELOG.md

### DIFF
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -95,7 +95,6 @@ func NewSwitch(config *cfg.P2PConfig) *Switch {
 	sw.peerConfig.MConfig.SendRate = config.SendRate
 	sw.peerConfig.MConfig.RecvRate = config.RecvRate
 	sw.peerConfig.MConfig.MaxPacketMsgPayloadSize = config.MaxPacketMsgPayloadSize
-	sw.peerConfig.AuthEnc = config.AuthEnc
 
 	sw.BaseService = *cmn.NewBaseService(nil, "P2P Switch", sw)
 	return sw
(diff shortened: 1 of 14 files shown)


p2p/switch.go
```diff
@@ -547,13 +542,14 @@ func (sw *Switch) addPeer(pc peerConn) error {
...
-	if pc.config.AuthEnc {
-		connID := pc.ID()
-
-		if peerID != connID {
-			return fmt.Errorf("nodeInfo.ID() (%v) doesn't match conn.ID() (%v)",
-				peerID, connID)
-		}
+	connID := pc.ID()
+
+	if peerID != connID {
```
ebuchman: We can do this before the handshake now !

xla: Not really as we need the `peerNodeInfo`, or is this available without doing the handshake?

ebuchman: oh whoops. I meant some of the checks below. Like checking for connections to self, duplicate peer, filterconnbyid, etc.

ebuchman:
Why is slate failing ?

Issue #1661, opened 2018-05-31
Mempools grow indefinitely under load


ldmberman:
Hi guys,

I would like to know your opinion on a certain problem, which I did not manage to find neither in the issues nor in the codebase.

According to our experiments, if a node keeps posting transactions at high pace, the network does not advance the blockchain - nodes receive the transactions and update their mempool, but do not engage in the consensus algorithm. This way, a single node can hold the blockchain construction off for an indefinite period of time (in our case tens of thousands of transactions ended up in the mempool and kept coming).

I imagine this to be by design, cause:

- no data is lost since everything is written to WAL;
- such an excess load may be prevented by some specific DOS prevention solution;
- some write availability is acquired at the cost of read availability.

How close is it to the truth? Do you have any other helpful insights into the problem?

Thank you!

ebuchman:
A few others have reported similar issue ie. https://github.com/tendermint/tendermint/issues/1642 and https://github.com/tendermint/tendermint/issues/1394

We're looking into this now.

What version of Tendermint are you using? We added a configurable cap to the mempool in v0.19.5 that should help with this, but it sounds like there might be something deeper we need to investigate.

Part of the challenge is to find the right balance between performance and the mempool UX (ie. what guarantees it provides).

ldmberman:
Thank you for the quick response.

I know about https://github.com/tendermint/tendermint/issues/1642 but I think it is unrelated - there the RPC server stops responding when everything else seems fine.

I did not see https://github.com/tendermint/tendermint/issues/1394 - it should be about the same issue indeed. But a single node is actually enough to observe this behaviour.

I was using `0.19.2`. I will try`v0.19.5` out, thank you!

ldmberman:
My colleague, @kansi tried out `0.19.7`:

> it seems to operate fine .i.e. blocks are being created slowly and steadly

So capping works, thank you!

There is now something else I would like to ask to avoid creating extra issues :)

While testing the capper, @kansi noticed that the RPC server stops responding intermittently for the time it commits new blocks. The same is actually happening when Tendermint replays the blocks for the proxy app. Is it a mempool implementation limitation? Do you expect the apps to implement additional buffering if rejecting transactions is not acceptable?

melekes:
> the RPC server stops responding intermittently for the time it commits new blocks. The same is actually happening when Tendermint replays the blocks for the proxy app. Is it a mempool implementation limitation?

[...]

Issue #1642, opened 2018-05-29
Tendermint unresponsive under heavy load


olegomano:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
BUG REPORT

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):
0.19.2

**ABCI app** (name for built-in, URL for self-written if it's publicly available):
BigchainDB 2.0.0a2
Debian 9.4.0 

Hi,
I am running a singe Tendermint node under BigchainDB for testing and evaluation purposes,  using the standard docker-compose setup provided on the BigchainDB github page. Periodically the Tendermint node becomes unresponsive, usually after heavy use.  Running the tm-bench tool to simulate a heavy use scenario will consistently put Tendermint into the unresponsive state. I am able to replicate this on two different machines. I found that it is easier to reproduce when running  tm-bench on the same machine as the Tenderming node, and that if your node is remote it can take a while to experience the unresponsive behavior. 

You can reproduce the issue by doing the follwing
sudo ./tm-bench -v 192.168.1.13:32946

Here are the logs from tm-bench 
```
Running 10s test 
I[05-29|22:32:47.189] Latest block height                          h=3839
I[05-29|22:32:47.189] Time started                                 t=2018-05-29T15:32:47.189606179-07:00
I[05-29|22:32:49.246] sent 1000 transactions                       addr=
 took=64.610214ms
I[05-29|22:32:50.247] sent 1000 transactions                       addr=
took=37.131893ms
I[05-29|22:32:51.248] sent 1000 transactions                       addr=
 took=43.399621ms
I[05-29|22:32:52.249] sent 1000 transactions                       addr=
took=41.565253ms
I[05-29|22:32:53.250] sent 1000 transactions                       addr=
 took=333.455458ms
I[05-29|22:32:55.353] sent 1000 transactions                       addr=
took=2.102869612s
txs send failed: write tcp : i/o timeout. Try reducing the connections count and increasing the rate.
chain@debian:~/Documents/tools/tm-bench$ sudo ./tm-bench -v
Running 10s test @



Status: Post : EOF
```
Here are the logs of Tendermint in the unresponsive state. 
```
I[05-29|21:35:46.228] Finalizing commit of block with 0 txs        module=consensus height=59 hash=4C334418A1D2CA607418539A90433D59183ABDA3 root=
I[05-29|21:35:46.229] Block{
  Header{
[...]

Issue #1394, opened 2018-03-30
Simultaneous requests to different nodes lead to mempool overload (constantly increasing txs)


eeonevision:
Hi!
I'm developing a high load system for tracking users click on different webpages (for advertisers needs) and I found some issue, but first I describe topology of network:
2 validator + 2 non-validator nodes.

When I post a transactions (with broadcast_tx_async) in average 5000 tx/sec (50 seconds) to only 1 non-validator node, the validators processes it right and makes blocks properly.
When I post a transactions in average 5000 tx/sec (50 seconds) to 2 non-validator nodes in parallel, so i got a mempool "**hell**".
The mempool shows me in _/num_unconfirmed_txs_ constantly increasing number of txs.
Now it:
```json
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    "n_txs": 3315247,
    "txs": null
  }
}
```
I post only 500k txs (250k to each of nodes).
Validators also don't make any blocks.

The _/dump_consensus_state_
```json
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    "round_state": {
      "Height": 26,
      "Round": 4,
      "Step": 6,
      "StartTime": "2018-03-30T11:40:09.326471823Z",
      "CommitTime": "2018-03-30T11:40:08.326471823Z",
      "Validators": {
        "validators": [
          {
            "address": "EE2C606BE264959534FAACB996FD592AAE120F81",
            "pub_key": {
              "type": "ed25519",
              "data": "2E66B74BE5C6F752689ED05650FB636681C86D1DF3B38AE11441C3D9F97B3334"
            },
            "voting_power": 10,
            "accum": 0
          },
          {
            "address": "F039F6E2981928FF0DEDDA4F928508E6B1CE6514",
            "pub_key": {
              "type": "ed25519",
              "data": "5AA998CFB71D18941694E9D8C295BB758B150BE7132FDC0659EB37DA0C154764"
            },
            "voting_power": 10,
            "accum": 0
          }
        ],
        "proposer": {
          "address": "F039F6E2981928FF0DEDDA4F928508E6B1CE6514",
          "pub_key": {
            "type": "ed25519",
            "data": "5AA998CFB71D18941694E9D8C295BB758B150BE7132FDC0659EB37DA0C154764"
          },
          "voting_power": 10,
          "accum": 0
        }
      },
      "Proposal": {
        "height": 26,
        "round": 4,
[...]

PR #1673, opened 2018-06-02
fix possible mempool deadlock

ebuchman:
Looking into the reports from #1642, #1394, #1661 .

Running `tm-bench` against a local node will sometimes time out at the RPC level.

Seems occasionally it will also deadlock - the RPC becomes unresponsive when trying to get `/status` or `/consensus_state`.

Passing `SIGABRT` to the tendermint process (`killall -6 <pid>`) dumps the go-routines. Upon inspection, I found all my HTTP request routines blocked on getting the consensus state lock.

Also found consensus blocked trying to Reap from the mempool, and the mempool resCb blocked trying to send on txsAvailable.

Seems the send on txsAvailable could happen multiple times despite the boolean guard. Actually looks like that boolean may have been racy. We just saw a similar kind of thing in the blockchain reactor: https://github.com/tendermint/tendermint/pull/1633

This should prevent us from blocking the mempool on sending txsAvailable.

TODO: update changelog and version before merge to master

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -72,8 +72,8 @@ type Mempool struct {
 	rechecking           int32           // for re-checking filtered txs on Update()
 	recheckCursor        *clist.CElement // next expected response
 	recheckEnd           *clist.CElement // re-checking stops here
-	notifiedTxsAvailable bool            // true if fired on txsAvailable for this height
-	txsAvailable         chan int64      // fires the next height once for each height, when the mempool is not empty
+	notifiedTxsAvailable bool
+	txsAvailable         chan int64 // fires the next height once for each height, when the mempool is not empty
 
 	// Keep a cache of already-seen txs.
 	// This reduces the pressure on the proxyApp.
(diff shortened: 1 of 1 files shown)


ebuchman:
Note I can't replicate this issue with the default config.

I find I need to combine the following:

```
create_empty_blocks = false

skip_timeout_commit = true
```

Either on their own and I can't reproduce ...

mempool/mempool.go
```diff
@@ -327,9 +326,11 @@ func (mem *Mempool) notifyTxsAvailable() {
...
 		panic("notified txs available but mempool is empty!")
 	}
-	if mem.txsAvailable != nil && !mem.notifiedTxsAvailable {
-		mem.notifiedTxsAvailable = true
-		mem.txsAvailable <- mem.height + 1
+	if mem.txsAvailable != nil {
+		select {
+		case mem.txsAvailable <- mem.height + 1:
+		default:
+		}
```
xla: This non-blocking send might break our guarantee on the consumer side, which if I'm not mistaken is only the consensusState. What happens if the select on the other side is never ready when the new height is send, will that break our consensus code?

ebuchman: the channel is buffered with capacity 1 so the first send should go through no matter what.

really we don't want a second send to even happen, which is what the boolean was supposed to prevent before, but on closer inspection is didn't seem quite thread safe, though I couldn't generate a race condition.

Tbh I'm not 100% this is the right solution and I agree its a bit troubling. How can we make it this dependency more explicit as you suggest ?

xla:
@ebuchman We still need the bool, if we want to preserve the current behaviour when it fires. I removed it and the tests pass again. What you think is racy about it?

ebuchman:
Why the rebase ? 

> What you think is racy about it?

its set to false here under protection of lock: https://github.com/tendermint/tendermint/blob/master/mempool/mempool.go#L378

but I don't see how it's under lock when it's set to true

PR #1676, opened 2018-06-02
Collapse PeerConfig into P2PConfig

xla:
As both configs are concerned with the p2p packaage and PeerConfig is only used inside of the package there is no good reason to keep the couple of fields separate, therefore it is collapsed into the more general P2PConifg. This is a stepping stone towards a setup where the components inside of p2p do not have any knowledge about the config.

follow-up to #1325

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,6 +5,15 @@ import (
 	"os"
 	"path/filepath"
 	"time"
+
+	tmconn "github.com/tendermint/tendermint/p2p/conn"
+)
+
+const (
+	// FuzzModeDrop is a mode in which we randomly drop reads/writes, connections or sleep
+	FuzzModeDrop = iota
+	// FuzzModeDelay is a mode in which we randomly sleep
+	FuzzModeDelay
 )
 
 // NOTE: Most of the structs & relevant comments + the
@@ -287,11 +296,24 @@ type P2PConfig struct {
 	// Does not work if the peer-exchange reactor is disabled.
 	SeedMode bool `mapstructure:"seed_mode"`
 
-	// Comma separated list of peer IDs to keep private (will not be gossiped to other peers)
+	// Comma separated list of peer IDs to keep private (will not be gossiped to
+	// other peers)
 	PrivatePeerIDs string `mapstructure:"private_peer_ids"`
 
 	// Toggle to disable guard against peers connecting from the same ip.
 	AllowDuplicateIP bool `mapstructure:"allow_duplicate_ip"`
+
+	// Peer connection configuration.
+	HandshakeTimeout time.Duration      `mapstructure:"handshake_timeout"`
+	DialTimeout      time.Duration      `mapstructure:"dial_timeout"`
+	MConfig          tmconn.MConnConfig `mapstructure:"connection"`
+
+	// Testing params.
+	// Force dial to fail
+	TestDialFail bool `mapstructure:"test_dial_fail"`
+	// FUzz connection
+	TestFuzz       bool            `mapstructure:"test_fuzz"`
+	TestFuzzConfig *FuzzConnConfig `mapstructure:"test_fuzz_config"`
 }
 
 // DefaultP2PConfig returns a default configuration for the peer-to-peer layer
@@ -308,6 +330,12 @@ func DefaultP2PConfig() *P2PConfig {
 		PexReactor:              true,
 		SeedMode:                false,
 		AllowDuplicateIP:        true, // so non-breaking yet
+		HandshakeTimeout:        20 * time.Second,
+		DialTimeout:             3 * time.Second,
+		MConfig:                 tmconn.DefaultMConnConfig(),
+		TestDialFail:            false,
+		TestFuzz:                false,
+		TestFuzzConfig:          DefaultFuzzConnConfig(),
 	}
 }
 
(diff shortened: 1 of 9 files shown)


p2p/test_util.go
```diff
@@ -8,7 +8,7 @@ import (
 	cmn "github.com/tendermint/tmlibs/common"
 	"github.com/tendermint/tmlibs/log"
 
-	cfg "github.com/tendermint/tendermint/config"
+	"github.com/tendermint/tendermint/config"
```
melekes: third change of this sort on my memory. we should make a standard and stick with it

p2p/switch.go
```diff
@@ -486,7 +483,10 @@ func (sw *Switch) listenerRoutine(l Listener) {
 	// cleanup
 }
 
-func (sw *Switch) addInboundPeerWithConfig(conn net.Conn, config *PeerConfig) error {
+func (sw *Switch) addInboundPeerWithConfig(
+	conn net.Conn,
```
melekes: Is this the result of default formatter? or you manually splitting the line

p2p/switch.go
```diff
@@ -90,11 +88,10 @@ func NewSwitch(config *cfg.P2PConfig) *Switch {
 	// Ensure we have a completely undeterministic PRNG.
 	sw.rng = cmn.NewRand()
 
-	// TODO: collapse the peerConfig into the config ?
-	sw.peerConfig.MConfig.FlushThrottle = time.Duration(config.FlushThrottleTimeout) * time.Millisecond
-	sw.peerConfig.MConfig.SendRate = config.SendRate
-	sw.peerConfig.MConfig.RecvRate = config.RecvRate
-	sw.peerConfig.MConfig.MaxPacketMsgPayloadSize = config.MaxPacketMsgPayloadSize
+	sw.config.MConfig.FlushThrottle = time.Duration(cfg.FlushThrottleTimeout) * time.Millisecond
```
melekes: I thought the whole idea was to get rid of this block after merge.

p2p/conn/connection.go
```diff
@@ -121,8 +121,8 @@ func (cfg *MConnConfig) maxPacketMsgTotalSize() int {
 }
 
 // DefaultMConnConfig returns the default config.
-func DefaultMConnConfig() *MConnConfig {
-	return &MConnConfig{
+func DefaultMConnConfig() MConnConfig {
```
melekes: all our `Default..Config` functions return pointers. Maybe we should be consistent

config/config.go
```diff
@@ -287,11 +296,24 @@ type P2PConfig struct {
...
+	// other peers)
 	PrivatePeerIDs string `mapstructure:"private_peer_ids"`
 
 	// Toggle to disable guard against peers connecting from the same ip.
 	AllowDuplicateIP bool `mapstructure:"allow_duplicate_ip"`
+
+	// Peer connection configuration.
+	HandshakeTimeout time.Duration      `mapstructure:"handshake_timeout"`
+	DialTimeout      time.Duration      `mapstructure:"dial_timeout"`
+	MConfig          tmconn.MConnConfig `mapstructure:"connection"`
```
melekes: should this be a pointer too?

p2p/test_util.go
xla: I made it consistent inside of `p2p` in the follow-up PR which comes right after this one is through.

p2p/switch.go
xla: This is how arguments are split when ranged over multiple lines.

p2p/switch.go
xla: Coming next, I first wanted to collapse configs and in the next step get rid of the config dependency completey.

xla: It will another 2-3 distinct PRs, I want to avoid massive changesets for this.

p2p/conn/connection.go
xla: Using pointers turned out to be racy, in general I would challenge that those should be mutable. Having them non-pointers would communicate that better. On other other hand we can always dereference them where they are consumed.

config/config.go
xla: Same point as above, it shouldn't be mutated.

p2p/peer.go
```diff
@@ -300,22 +302,33 @@ func (p *peer) hasChannel(chID byte) bool {
 	}
 	// NOTE: probably will want to remove this
 	// but could be helpful while the feature is new
-	p.Logger.Debug("Unknown channel for peer", "channel", chID, "channels", p.channels)
+	p.Logger.Debug(
+		"Unknown channel for peer",
+		"channel",
+		chID,
```
ebuchman: we should keep the pairs on a single line

p2p/pex/pex_reactor_test.go
```diff
@@ -450,7 +451,7 @@ func teardownReactor(book *addrBook) {
 }
 
 func createSwitchAndAddReactors(reactors ...p2p.Reactor) *p2p.Switch {
-	sw := p2p.MakeSwitch(config, 0, "127.0.0.1", "123.123.123", func(i int, sw *p2p.Switch) *p2p.Switch { return sw })
+	sw := p2p.MakeSwitch(cfg, 0, "127.0.0.1", "123.123.123", func(i int, sw *p2p.Switch) *p2p.Switch { return sw })
```
ebuchman: maybe it would be better to make it more explicit by calling it `testCfg` or something ?

xla: That it is global state is already a smell, it's gonna vanish when all the config traces are gone from the package.

p2p/peer.go
xla: Will include in the next one.

Issue #1659, opened 2018-05-31
Panic in wal after node restart


zmanian:
I've seen this happen to both my nodes on gaia-5001 when I restarted them.

I took a snapshot of the machine so we can debug.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xb1b7e7]

goroutine 33 [running]:
github.com/cosmos/cosmos-sdk/vendor/github.com/tendermint/tendermint/consensus.(*baseWAL).SearchForEndHeight(0xc4279b3590, 0x35c7b, 0xc4254beb90, 0x20, 0xc426500c00, 0x7f92a92c5850, 0x0)
	/home/ubuntu/go/src/github.com/cosmos/cosmos-sdk/vendor/github.com/tendermint/tendermint/consensus/wal.go:179 +0x2a7
github.com/cosmos/cosmos-sdk/vendor/github.com/tendermint/tendermint/consensus.(*ConsensusState).catchupReplay(0xc42691eb00, 0x35c7b, 0x0, 0x0)
	/home/ubuntu/go/src/github.com/cosmos/cosmos-sdk/vendor/github.com/tendermint/tendermint/consensus/replay.go:110 +0xc8
github.com/cosmos/cosmos-sdk/vendor/github.com/tendermint/tendermint/consensus.(*ConsensusState).OnStart(0xc42691eb00, 0xc42c8f9ea0, 0x17)
	/home/ubuntu/go/src/github.com/cosmos/cosmos-sdk/vendor/github.com/tendermint/tendermint/consensus/state.go:255 +0x169
github.com/cosmos/cosmos-sdk/vendor/github.com/tendermint/tmlibs/common.(*BaseService).Start(0xc42691eb00, 0xc4287862e0, 0x9)
	/home/ubuntu/go/src/github.com/cosmos/cosmos-sdk/vendor/github.com/tendermint/tmlibs/common/service.go:130 +0x405
github.com/cosmos/cosmos-sdk/vendor/github.com/tendermint/tendermint/consensus.(*ConsensusReactor).SwitchToConsensus(0xc427cd2680, 0xc4287862e0, 0x9, 0x35c7a, 0xc2, 0xc427ccc8a0, 0x14, 0x14, 0x1, 0xc427ccc8c0, ...)
[...]

PR #1679, opened 2018-06-04
Flush cs.wal on stop

melekes:
Refs #1659
Refs tendermint/tmlibs#217

### DIFF
--- a/consensus/wal.go
+++ b/consensus/wal.go
@@ -106,8 +106,8 @@ func (wal *baseWAL) OnStart() error {
 }
 
 func (wal *baseWAL) OnStop() {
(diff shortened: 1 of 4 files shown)


consensus/wal.go
```diff
@@ -106,8 +106,8 @@ func (wal *baseWAL) OnStart() error {
 }
 
 func (wal *baseWAL) OnStop() {
-	wal.BaseService.OnStop()
 	wal.group.Stop()
+	wal.group.Close()
```
ebuchman: why not make this part of Stop ?

melekes: To keep the parity: OpenGroup/Close, Start/Stop

PR #1687, opened 2018-06-05
do not drain the channel because there is no channel, duh

melekes:
Fixes https://github.com/cosmos/cosmos-sdk/issues/1045

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/p2p/conn/connection_test.go
+++ b/p2p/conn/connection_test.go
@@ -6,9 +6,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/fortytw2/leaktest"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


p2p/conn/connection_test.go
```diff
@@ -6,9 +6,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/fortytw2/leaktest"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"github.com/tendermint/go-amino"
+
+	amino "github.com/tendermint/go-amino"
```
xla: The package is already called `amino`, why the extra alias?

xla: Got it, thanks!

PR #1695, opened 2018-06-06
Enable profiler by default

ValarDragon:
Addresses cosmos/cosmos-sdk#972 and cosmos/cosmos-sdk#1098
* [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## 0.19.10
 
 *TBD*
+PENDING
+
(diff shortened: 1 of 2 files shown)


melekes:
Not sure I understand why a person who wants to profile Tendermint can't just change the config file...

ValarDragon:
My understanding was that we should do this during the testnet, just to make bug hunting potentially easier.

ebuchman:
We want to do this on SDK, not in Tendermint.

Thanks though.

PR #1697, opened 2018-06-06
Improve mempool logging

melekes:


ebuchman:
I like the Totals but I called it TxID in https://github.com/tendermint/tendermint/pull/1795/files#diff-638e837163a1b6a2dbb5f19513153895R61

Issue #1640, opened 2018-05-29
docs: re-organize sections


zramsay:
from: #1636 

Tendermint docs might be way better if they were organized into higher level sections:

- install
- tendermint core (CLI, RPC, config, genesis, running in production, spec, etc.)
- tendermint networks (local cluster, remote cluster, cosmos testnet)
- application dev (first app, abci, app guides, etc.)

To happen once #1636 and #1618 are merged

zramsay:
also, switch to md for upcoming transition to Vue

PR #1698, opened 2018-06-06
docs to markdown

zramsay:
preparation for the documentation overhaul (and #1640). The current RTD will build using markdown and we'll (obviously) keep RTD up until the switch is fully complete (and all links updated). Still a handful of fixes needed -- easier to inspect a PR than using `git diff`

* [N/A] Updated all code comments where relevant
* [N/A] Wrote tests
* [ ] Updated CHANGELOG.md


zramsay:
thank for nothing codecov

zramsay:
ready for review. With the legacy RTD, we're going to lose self-referential links (except the "Previous" and "Next" buttons)...i.e., anything that once ended with `.html` should now be `.md`, if we want the links to the work in the new docs framework. Best solution I can come up with is remove as many self-referenced links within the docs as possible

zramsay:
also made small modifications to the ToC (`index.rst`) based loosely on #1640 but this re-org will take some consideration before diving in

PR #1711, opened 2018-06-08
Bucky/adrs

ebuchman:
Wrote some Architecture Decision Records for 

- the recent changes to the ABCI, including some additional ones we should make. basically complete
- proposed changes to our crypto libs - still needs to be fleshed out and references added

A lot of this information is already captured across issues, especially starting from https://github.com/tendermint/tendermint/issues/1568, but in light of Ismail's HDkeys design doc, thought we should try to catch up here as well. 

We should also add ADRs for the proposed changes to:

- NextValidatorSet
- BFT Time

Note the distinction between an ADR and a spec. We should be using both for basically everything. The ADR provides the context and intuition and justification for a change in architecture, or for the architecture of something new. The spec is much more compressed and streamlined summary of everything as it stands today.

### DIFF
--- a/docs/architecture/adr-009-ABCI-design.md
+++ b/docs/architecture/adr-009-ABCI-design.md
@@ -1,0 +1,205 @@
+# ADR 009: ABCI UX Improvements
+
+## Context
+
+The ABCI was first introduced in late 2015.  It's purpose is to be:
+
+- a generic interface between state machines and their replication engines
+- agnostic to the language the state machine is written in
+- agnostic to the replication engine that drives it
+
+This means ABCI should provide an interface for both pluggable applications and
+pluggable consensus engines.
+
+To achieve this, it uses Protocol Buffers for message types. The dominant
+implementation is in Go.
+
+After some recent discussions with the community on github, the following were
+identified as pain points:
+
+- Amino encoded types
+- Managing validator sets
+- Imports in the protobuf file
+
+See the following relevent github issues:
+- TODO
+
+### Imports
+
+The native Protobuf library in Go generates code that is inellegant and difficult to work with.
+The solution in the Go community is to use a fork of it called `gogoproto`.
+While `gogoproto` is nice, it creates an additional dependency, and compiling
+the protobuf types for other languages has been reported to fail when `gogoproto` is used.
+
+### Amino
+
+Amino is an encoding protocol designed to improve over insufficiencies of protobuf.
+It's goal is to be Protobuf4.
+
+Many people are frustrated by incompatibility with protobuf,
+and with the requirement for Amino to be used at all within ABCI.
+
+We intend to make Amino successful enough that we can eventually use it for ABCI
+message types directly. By then it should be called Protobuf4. In the meantime,
+we want it to be easy to use.
+
+### PubKey
+
+PubKeys were previously encoded using Amino (and before that, go-wire).
+
+### Addresses
+
+The address for an ED25519 pubkey is currently the RIPEMD160 of the Amino
+encoded pubkey.
+
+### Validators
+
+To change the validator set, applications can return a list of validator updates
+with ResponseEndBlock. In these updates, the public key *must* be included,
+because Tendermint requires the public key to verify validator signatures.  This
+means ABCI developers have to work with PubKeys. That said, it would also be
+convenient to work with address information, and for it to be simple to do so.
+
+### AbsentValidators
+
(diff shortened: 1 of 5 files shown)


docs/architecture/adr-009-ABCI-design.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Amino
+
+To simplify ABCI application development in the short term,
+Amino will be completely removed from the ABCI:
+
+- It will not be required for PubKey encoding
+- It will not be required for computing PubKey addresses
+
+That said, we are working to make Amino a huge success, and to become Protobuf4.
```
xla: `Protobuf4` -> `proto4`

docs/architecture/adr-009-ABCI-design.md
```diff
@@ -0,0 +1,260 @@
...
+PubKey will now look like:
+
+```
+message PubKey {
+    string type
+    bytes data
+}
+```
+
+where `type` can be:
```
xla: If `type` is only one of two to three possibilities why are we not using an enum for increased hardening?

docs/architecture/adr-009-ABCI-design.md
```diff
@@ -0,0 +1,260 @@
...
+message RequestInitChain {
+    int64 time
+    string chain_id
+    ConsensusParams consensus_params
+    repeated Validator validators
+    bytes app_state_bytes
+}
+
+message ResponseInitChain {
+    ConsensusParams consensus_params
```
xla: Could be helpful to expand on this field and clarify why we return it back to the caller and the content is.

docs/architecture/adr-010-crypto-changes.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Decision
+
+### Hash Functions
+
+Use the first 20-bytes of the SHA256 hash instead of RIPEMD160 for everything
+
+### Merkle Trees
+
+TODO
```
Liamsi: Will there be changes that will be included here? E.g., SimpleMerkle also uses the truncated sha256 now:
https://github.com/tendermint/go-crypto/blob/master/CHANGELOG.md#070

docs/architecture/adr-009-ABCI-design.md
ebuchman: We actually want `PubKey` to be an interface, where you don't necessarily know all the types that could implement it.

docs/architecture/adr-009-ABCI-design.md
ebuchman: `Chain ResponseInitChain to allow the app to specify the initial validator set
and consensus parameters.` is that better ? ConsensusParams definition doesnt belong here

docs/architecture/adr-010-crypto-changes.md
ebuchman: Yes, would be great if you could help complete it. I left it for now as we were still discussing changes to the internal nodes to prevent 2nd preimages ...

docs/architecture/adr-010-crypto-changes.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Merkle Trees
+
+Tendermint uses a simple Merkle tree to compute digests of large structures like transaction batches
+and even blockchain headers. The Merkle tree length prefixes byte arrays before concatenating and hashing them.
+It uses RIPEMD160.
+
+### Addresses
+
+ED25519 addresses are computed using the RIPEMD160 of the Amino encoding of the public key.
```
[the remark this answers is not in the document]
ebuchman: I doubled down on current tense since we're trying to describe the current state and then make proposals for the new state. I'll add a short README

ebuchman:
Addressed feedback and rebased on develop. Crypto one still needs some work but maybe @Liamsi can get to it as part of preparations for next release with go-crypto merged and any final changes to the merkle tree or other pieces. thanks!

docs/architecture/adr-009-ABCI-design.md
ValarDragon: Since we have to fork anyway to add a new pubkey type, not sure I follow why this couldn't be an enum?

ebuchman: This argument doesn't really hold for Tendermint, but I think it might for an app like the SDK where some module could have a custom implementation of pubkey

PR #1731, opened 2018-06-10
`unsafe_reset_all` also resets addrbook.json

### LINKED ISSUE #1713: unsafe_reset_all should clear addrbook.json
`tendermint unsafe_reset_all` should also clear `addrbook.json`.

adrianbrink:
When executing `unsafe_reset_all` it also clear all IP addresses from
addrbook.json. This is the expected behaviour of `unsafe_reset_all`.

closes #1713

### DIFF
--- a/cmd/tendermint/commands/reset_priv_validator.go
+++ b/cmd/tendermint/commands/reset_priv_validator.go
@@ -27,7 +27,7 @@ var ResetPrivValidatorCmd = &cobra.Command{
 // XXX: this is totally unsafe.
 // it's only suitable for testnets.
 func resetAll(cmd *cobra.Command, args []string) {
-	ResetAll(config.DBDir(), config.PrivValidatorFile(), logger)
+	ResetAll(config.DBDir(), config.P2P.AddrBook, config.PrivValidatorFile(), logger)
 }
 
 // XXX: this is totally unsafe.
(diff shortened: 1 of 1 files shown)


cmd/tendermint/commands/reset_priv_validator.go
```diff
@@ -59,3 +60,13 @@ func resetFilePV(privValFile string, logger log.Logger) {
 		logger.Info("Generated PrivValidator", "file", privValFile)
 	}
 }
+
+// Deletes the addrbook.json file. It will be generated in node.NewNode, which is executed with
+// `tendermint start`.
```
melekes: > `It will be generated in node.NewNode, which is executed with tendermint start`

seems to be out of scope. let's remove

cmd/tendermint/commands/reset_priv_validator.go
```diff
@@ -59,3 +60,13 @@ func resetFilePV(privValFile string, logger log.Logger) {
...
+func resetAddrBook(addrBookFile string, logger log.Logger) {
+	if err := os.Remove(addrBookFile); err == nil {
+		logger.Info("Cleared AddrBook to an empty state", "file", addrBookFile)
```
melekes: `Removed existing address book` is more human than `Cleared AddrBook to an empty state`

cmd/tendermint/commands/reset_priv_validator.go
```diff
@@ -59,3 +60,13 @@ func resetFilePV(privValFile string, logger log.Logger) {
...
+	} else {
+		logger.Info("Failed to clear AddrBook to an empty state", "file", addrBookFile)
```
melekes: `Error removing address book`, `file`, addrBookFile, `err`, err

melekes:
@adrianbrink thank you 👍

PR #1732, opened 2018-06-11
typo fix

maxim-levy:
And typo fix

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


melekes:
@maxim-levy thank you for the PR, but there is no typo.

ebuchman:
Hmm think this actually is a typo

melekes:
Oh, you're right. I was looking at the diff the other way around :(

PR #1735, opened 2018-06-12
Dev/ephemeral port2

ValarDragon:
This moves all ports of the form 466xx to 266xx. This is done to address cosmos/cosmos-sdk#932. Essentially the problem was the default ports used by tendermint are in the ephemeral port range.

This requires tendermint/abci#253

### DIFF
--- a/docs/specification/block-structure.rst
+++ b/docs/specification/block-structure.rst
@@ -129,7 +129,7 @@ For example, a precommit vote might have the following ``sign-bytes``:
 
 .. code:: json
 
-    {"chain_id":"my_chain","vote":{"block_hash":"611801F57B4CE378DF1A3FFF1216656E89209A99","block_parts_header":{"hash":"B46697379DBE0774CC2C3B656083F07CA7E0F9CE","total":123},"height":1234,"round":1,"type":2}}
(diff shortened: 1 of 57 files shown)


docs/specification/block-structure.rst
```diff
@@ -129,7 +129,7 @@ For example, a precommit vote might have the following ``sign-bytes``:
 
 .. code:: json
 
-    {"chain_id":"my_chain","vote":{"block_hash":"611801F57B4CE378DF1A3FFF1216656E89209A99","block_parts_header":{"hash":"B46697379DBE0774CC2C3B656083F07CA7E0F9CE","total":123},"height":1234,"round":1,"type":2}}
+    {"chain_id":"my_chain","vote":{"block_hash":"611801F57B4CE378DF1A3FFF1216656E89209A99","block_parts_header":{"hash":"B26697379DBE0774CC2C3B656083F07CA7E0F9CE","total":123},"height":1234,"round":1,"type":2}}
```
melekes: why change?

ValarDragon: Oh woops, my bad. I was replacing the rst files via the command line, I didn't realize the hash was used in there. I'll revert that.

Issue #986, opened 2017-12-18
Prometheus monitoring


nickray:
I would like to see a Prometheus monitoring endpoint available ([basic example](https://github.com/prometheus/client_golang/blob/master/examples/random/main.go)). The use case is for running a Cosmos Validator, where escrowed value is at risk of improper operations.

Performance considerations permitting (compile time option?), I assume integrating the [client library](https://godoc.org/github.com/prometheus/client_golang/prometheus), exposing the endpoint, and adding basic metrics (such as `latest_block` which are currently available at the `46657/status` endpoint) should be fairly straight forward. From there, the actual metrics could be added on a pull request basis, or Tendermint users could patch in their own (possibly experimental) metrics.

The old [tm-monitor](https://github.com/tendermint/tm-monitor) repository implies that this has been considered before. What was the outcome? In my opinion, Tendermint should expose the information necessary to monitor its operation and state changes, but not be (officially) providing the [tools](https://github.com/tendermint/tools) to monitor them, as project clients may have varying requirements. Moreover, if there are multiple possibilities of implementing a solution, a [(semi)standard](https://coreos.com/blog/monitoring-kubernetes-with-prometheus.html) should be chosen instead of a homegrown solution. 

With the Christmas holiday coming up, I could possibly contribute pull requests. I have professional experience with C++ and Python, however Go and this codebase are new to me.

Looking forward to the discussion!

ebuchman:
Do we need to change anything in Tendermint to support this  ?

nickray:
I apologize for not having sent any pull requests regarding this, and probably will not. It's an architectural decision on your side I think.

As supporting evidence, that go-kit/metrics itself states "It's primarily designed to help you get started with good and robust instrumentation, and to help you migrate from a less-capable system like Graphite to a more-capable system like Prometheus. If your organization has already standardized on an instrumentation system like Prometheus, and has no plans to change, it may make sense to use that system's instrumentation library directly."

What would have to changed AFAIK is
* adding an endpoint for scraping the metrics (configurable whether/where it's exposed)
* linking in the https://github.com/prometheus/client_golang library
* starting to expose relevant metrics (this can be minimal at first, and ideally someone running a node can easily extend those metrics by hacking the code)

xla:
> Do we need to change anything in Tendermint to support this ?

[...]

PR #1737, opened 2018-06-12
metrics

melekes:
Refs #986 

screen shot 2018-06-15 at 15 20 43


docs/architecture/adr-010-monitoring.md
```diff
@@ -0,0 +1,103 @@
...
+
+Same issue as the above.
+
+### 2. opencensus
+
+opencensus provides both metrics and tracing, which may be important in the
+future. It's API looks different from go-kit and Prometheus, but looks like it
+covers everything we need.
+
+Unfortunately, OpenCensus go client does not define any
```
melekes: looks like it's too much work. so maybe we should go with {go-kit + prometheus}

xla:
All in favor of an initial setup with gokit and prometheus. The two concepts are not interchangeable. Metrics, together with logs and tracing make up the larger story of observability. Where logs and metrics have the biggest return for the effort to make. Tracing is very powerful but puts a lot of pressure on the supporting infrastructure where a metrics endpoint in our binaries gives instant value.

consensus/state.go
```diff
@@ -388,6 +392,7 @@ func (cs *ConsensusState) SetProposalAndBlock(proposal *types.Proposal, block *t
 
 func (cs *ConsensusState) updateHeight(height int64) {
 	cs.Height = height
+	cs.metrics.Height.Add(float64(height - cs.Height))
```
xla: Same as with logging, we keep interleaving non-business concerns with our business logic, which makes for harder to maintain code. I know this would require a bigger refactor but this might be a good time to think about better composability. Which could be done with the more rigirous use of interfaces and then layered decoratored implementations for the different concerns (logic, logging, metrics). This [video](https://www.youtube.com/watch?v=xyDkyFjzFVc) gives an intro how that looks and I can give some examples how it is done in production code.

docs/architecture/adr-010-monitoring.md
```diff
@@ -0,0 +1,103 @@
...
+interfaces, so if we want to abstract away metrics we
+will need to write interfaces ourselves.
+
+### List of metrics
+
+|   | Name                                    | Type    |                                                                               |
+| - | --------------------------------------- | ------- | ----------------------------------------------------------------------------- |
+| A | height                                  | Counter |                                                                               |
+| A | validators:<height>                     | Gauge   | Number of validators who signed                                               |
```
xla: `<height>` has a couple of occurrences, which I suspect will lead to high cardinality in our metrics, something which can be a problem for long-running processes. Are we going to cap how many of them we keep around, or do we just expose the latest height?

melekes: I thought of it as a label. Like `validators` with a label `height-11`. Sorry for confusion. What do you think it?

consensus/state.go
melekes: Video is great. I love it.

xla:
Follow-up to https://github.com/tendermint/tendermint/pull/1737#discussion_r195034746: Generally any cardinality beyond 10s of things is going to put a lot of pressure on the collectors. See the [label section](https://prometheus.io/docs/practices/naming/#labels) in the prometheus documentation. I was under the impression that `height` is one of those unbounded set of values.

docs/architecture/adr-011-monitoring.md
```diff
@@ -0,0 +1,105 @@
...
+
+Unfortunately, OpenCensus go client does not define any
+interfaces, so if we want to abstract away metrics we
+will need to write interfaces ourselves.
+
+### List of metrics
+
+|   | Name                                    | Type    | Labels                    | Description                                                                   |
+| - | --------------------------------------- | ------- | ------------------------- | ----------------------------------------------------------------------------- |
+| A | height                                  | Counter | height-X                  |                                                                               |
```
melekes: list needs to be updated

node/node.go
```diff
@@ -81,10 +85,97 @@ func DefaultNewNode(config *cfg.Config, logger log.Logger) (*Node, error) {
 		proxy.DefaultClientCreator(config.ProxyApp, config.ABCI, config.DBDir()),
 		DefaultGenesisDocProviderFunc(config),
 		DefaultDBProvider,
+		DefaultMetricsProvider,
 		logger,
 	)
 }
 
+// MetricsProvider returns a consensus and p2p Metrics.
```
melekes: move to a separate package?

types/block.go
```diff
@@ -135,6 +135,15 @@ func (b *Block) HashesTo(hash []byte) bool {
 	return bytes.Equal(b.Hash(), hash)
 }
 
+// Size returns size of the block in bytes.
+func (b *Block) Size() int {
+	bz, err := cdc.MarshalBinaryBare(b)
```
melekes: expensive to call for each block?

melekes:
Should we add a config option?

node/node.go
```diff
@@ -81,10 +85,102 @@ func DefaultNewNode(config *cfg.Config, logger log.Logger) (*Node, error) {
...
+type MetricsProvider func() (*cs.Metrics, *p2p.Metrics, *mempl.Metrics)
+
+// DefaultMetricsProvider returns consensus, p2p and mempool Metrics build
+// using Prometheus client library.
+func DefaultMetricsProvider() (*cs.Metrics, *p2p.Metrics, *mempl.Metrics) {
```
xla: This seems to be the wrong place for setting those up, as the metrics are observed in the packages themselves they should live there. Also this will grow into a hug construct with more metrics/packages, it would live better in the `metrics.go` int he respective package with the sane parametrisation.

node/node.go
```diff
@@ -122,6 +218,7 @@ func NewNode(config *cfg.Config,
 	clientCreator proxy.ClientCreator,
 	genesisDocProvider GenesisDocProvider,
 	dbProvider DBProvider,
+	metricsProvider MetricsProvider,
```
xla: `NewNode` already takes the config, no need to extend its surface even further we can do the setup of what metric to use for what component down below.

melekes:
@xla thanks for the feedback! greatly appreciated

> the metrics endpoitn should not be on the same port as RPC

agree, changed.

> we should avoid extending the surface of our components. 

you're right. check out my last commits where I changed constructors to accept the list of options aka https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

consensus/state.go
```diff
@@ -1291,6 +1300,43 @@ func (cs *ConsensusState) finalizeCommit(height int64) {
 	// * cs.StartTime is set to when we will start round0.
 }
 
+func (cs *ConsensusState) recordMetrics(height int64, block *types.Block) {
```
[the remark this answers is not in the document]
melekes: But where do we store the state? In the reactor?

node/node.go
melekes: 1. I don't want to bring prometheus package inside our internal packages (cs, mempool, ..)
2. The reason we have *Provider is because some people use Tendermint as a library and for them to use different DB or metrics, it's much easier to only change 1 function - NewNode instead of changing all source code of Tendermint

node/node.go
```diff
@@ -527,6 +650,22 @@ func (n *Node) startRPC() ([]net.Listener, error) {
 	return listeners, nil
 }
 
+// StartPrometheusServer starts a Prometheus HTTP server, listening for metrics
+// collectors on addr.
+func (n *Node) StartPrometheusServer(addr string) *http.Server {
```
xla: Does this need to be public?

node/node.go
xla: That's very reasonable. I still don't believe we need this huge constructor for metrics, check my comment on the state storage above.

consensus/state.go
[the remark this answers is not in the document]
xla: Ideally we don't store the state anywhere instead keep it enclosed in the outter scope. To illustrate:

``` go
type recordMetricsFunc func(int64, *types.Block, *types.ValidatorSet)

func promMetrics() recordMetricsFunc {
	metrics := &Metrics{}

	return func(int64, *types.Block, *types.ValidatorSet) {
		// update metrics
	}
}

func nopMetrics() recordMetricsFunc {
	return func(int64, *types.Block, *types.ValidatorSet) {
		// nop
	}
}
```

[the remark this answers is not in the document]
melekes: that's really smart 👍 let's see what I can do

node/node.go
melekes: no

node/node.go
melekes: moved it to a separate package (`metrics/prometheus`)

consensus/state.go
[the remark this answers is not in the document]
melekes: the problem is not all the metrics can be recorded in this one function. some of the like round increments do not belong to this function... but the idea itself is beautiful.

metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,102 @@
+package prometheus
```
xla: I know we already had a discussion and on second thought I don't agree with the sentiment that this should not go in the respective packages. The declaration of the structs, the default nop implementation and the general ownership of the metrics are already there. We import prometheus anyway in Tendermint and as long as we don't have pressurign use-cases for single packages which require prometheus not to be imported I think we should not make decisions based on it. Even if we decide we want "cleaner" imports we can always make sub-packages with the metric providers. As it stands right now it feels dislocated and is harder to maintain as there are multiple places to keep track, which is unnecessary burden.

consensus/state.go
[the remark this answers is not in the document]
xla: There are couple of ways to go about it, one is to introduce specialised functions and another have the `recordMetrics` take all possible arguments and decide which ones to record based on what has been passed in, e.g. `if block == nil { return }`.

consensus/state.go
```diff
@@ -1291,6 +1310,43 @@ func (cs *ConsensusState) finalizeCommit(height int64) {
...
+		if _, val := cs.Validators.GetByAddress(ev.Address()); val != nil {
+			byzantineValidatorsPower += val.VotingPower
+		}
+	}
+	cs.metrics.ByzantineValidatorsPower.Set(float64(byzantineValidatorsPower))
+
+	if height > 1 {
+		lastBlockMeta := cs.blockStore.LoadBlockMeta(height - 1)
+		cs.metrics.BlockIntervalSeconds.
+			Observe(block.Time.Sub(lastBlockMeta.Header.Time).Seconds())
```
xla: This is an unconventional way of splitting long lines. Preferrably:

``` go
	cs.metrics.BlockIntervalSeconds.Observe(
		block.Time.Sub(lastBlockMeta.Header.Time).Seconds(),
	)
```

consensus/state.go
```diff
@@ -115,10 +115,13 @@ type ConsensusState struct {
...
 	// state only emits EventNewRoundStep, EventVote and EventProposalHeartbeat
 	evsw tmevents.EventSwitch
+
+	// for reporting metrics
+	metrics *Metrics
 }
 
 // NewConsensusState returns a new ConsensusState.
-func NewConsensusState(config *cfg.ConsensusConfig, state sm.State, blockExec *sm.BlockExecutor, blockStore sm.BlockStore, mempool sm.Mempool, evpool sm.EvidencePool) *ConsensusState {
+func NewConsensusState(config *cfg.ConsensusConfig, state sm.State, blockExec *sm.BlockExecutor, blockStore sm.BlockStore, mempool sm.Mempool, evpool sm.EvidencePool, options ...func(*ConsensusState)) *ConsensusState {
```
xla: This is growing quite large can we split it in multiple lines?

types/block.go
xla: Can we make a call on this? Do we know about the cost?

node/node.go
```diff
@@ -208,11 +229,24 @@ func NewNode(config *cfg.Config,
...
+	} else {
+		csMetrics, p2pMetrics, memplMetrics = NopMetricsProvider()
+	}
+
 	// Make MempoolReactor
 	mempoolLogger := logger.With("module", "mempool")
-	mempool := mempl.NewMempool(config.Mempool, proxyApp.Mempool(), state.LastBlockHeight)
-	mempool.InitWAL() // no need to have the mempool wal during tests
+	mempool := mempl.NewMempool(config.Mempool, proxyApp.Mempool(), state.LastBlockHeight,
+		mempl.WithMetrics(memplMetrics))
```
xla: Better to split like:

``` go
mempool := mempl.NewMempool(
    config.Mempool,
    proxyApp.Mempool(),
    state.LastBlockHeight,
    mempl.WithMetrics(memlMetrics),
)
```

melekes:
@xla thanks again for a thorough review. next time feel free to make changes yourself instead of spending time writing comments.

> re https://github.com/tendermint/tendermint/pull/1737#discussion_r196617652

if you're sure about this, please go ahead and implement

types/block.go
melekes: screen shot 2018-06-20 at 11 16 05

~10ms

PR #1738, opened 2018-06-12
Add algorithm for Blockchain Reactor

milosevic:


docs/spec/reactors/block_sync/reactor.md
```diff
@@ -44,10 +44,259 @@ type bcStatusResponseMessage struct {
...
+      sleep requestIntervalMS	        		
+
+  return selectedPeer 	
+```
+sleep for requestIntervalMS
+### Task for creating Requesters
+
+This task is responsible for continuously creating and starting Requester tasks.
+```go
+creteRequestors(pool):
```
xla: `createRequestors`

milosevic: We need to discuss the formatting issues with .md files. I was fighting yesterday 2 hours to align pseudo-code and it apparently still looks bad. Do we have markdown expert, @zramsay maybe? I am wondering how much formatting support we have with markdown so rendering is more deterministic. Having latex support would be awesome. I will update diagram.

milosevic:
@xla Hopefully formatting is better now.

docs/spec/reactors/block_sync/reactor.md
zramsay: We will be implementing a markdown linter in the near future. For now i'd say as long as it is close enough to standard markdown, we're good

milosevic: Thanks @zramsay. I was having issue with formatting, I guess related to tabs and white spaces. The pseudocode is properly formatted in the source file and then after rendering looks bad. I am using Intellij GoLand editor so it could be that it's editor issue. Have you experienced similar issues and do you have some recommendation regarding this?

zramsay: hmm, i use vim and my only plugin is vim-go ... @nylira mentioned using an editor plugin for formatting md files. we could add some info here: https://github.com/tendermint/coding/tree/master/editors

nylira: We're attempting to use Prettier for markdown formatting of all docs and content. https://github.com/prettier/prettier It's a popular and battle tested code formatter that is integrated with most common text editors.

milosevic: Thanks @zramsay @nylira. I will try prettier on some markdown files.

xla:
@milosevic Thanks, way better now!

PR #1742, opened 2018-06-13
Add Proto files for types.Header (incl. BlockId, Time, PartsSetHeader)

Liamsi:
This PR contains proto files for some relevant types in tendermint. The motivation is to a) document the types in the proto3 message format and b) to showcase (and test) amino's proto3 compatibility. 

The added proto files can be handed out to anyone wanting to write an implementation in another language that sends/receives the types in question (they can also use the proto files to simply generate the code in their language). The test-cases are simple regression tests but they also document any differences between amino and protobuf (if any).

cc'ing: @jaekwon as this is related to the topic he raised in the team-meeting:
https://github.com/tendermint/tendermint/pull/1742/files#diff-332798f22074986b29e3257d3510b10eR70
This would need updating if https://github.com/tendermint/tendermint/pull/1965 get's merged first.

depends on https://github.com/tendermint/go-amino/pull/178
needs updating when https://github.com/tendermint/go-amino/pull/179 gets merged

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


types/canonical_json.go
```diff
@@ -3,14 +3,13 @@ package types
...
 
-	"github.com/tendermint/go-amino"
 	cmn "github.com/tendermint/tmlibs/common"
 )
 
 // Canonical json is amino's json for structs with fields in alphabetical order
 
 // TimeFormat is used for generating the sigs
-const TimeFormat = amino.RFC3339Millis
+const TimeFormat = "2006-01-02T15:04:05.000Z" // forced microseconds
```
xla: Can this lead to divergence? Would it make sense to change the `amino` format?

types/proto3_test.go
```diff
@@ -0,0 +1,92 @@
+package types
+
+import (
+	"testing"
+	"github.com/tendermint/tendermint/types/proto3"
+	"github.com/stretchr/testify/assert"
+	"github.com/golang/protobuf/proto"
+	"time"
```
xla: Import order should always be:

``` go
import (
	// stdlib
	//
	// 3rd party
	//
	// internal packages
)
```

types/canonical_json.go
Liamsi: Not related to the actual commits I've made (the proto files). But good question! 
This was recently changed in https://github.com/tendermint/go-amino/pull/175

types/proto3_test.go
Liamsi: Will do 👍 This is still WIP...

Liamsi:
> Besides stating what is going to change it would be very helpful to state the motivation and the anticipated impact in the PR description.

Totally agree! Will add a reasonable description as soon as everything works.

Liamsi:
@xla this is ready for another review

types/proto3/block.proto
```diff
@@ -0,0 +1,48 @@
...
+  bytes ValidatorsHash = 9; // validators for the current block
+  bytes ConsensusHash = 10; // consensus params for current block
+  bytes AppHash = 11; // state after txs from the previous block
+  bytes LastResultsHash = 12; // root hash of all results from the txs from the previous block
+
+  // consensus info
+  bytes EvidenceHash = 13; // evidence included in the block
+}
+
+// Timestamp wraps how amino encodes time. Note that this is different from the protobuf well-known type
```
melekes: what's the reason it's different?

Liamsi: Because we encode them differently in amino. While seconds and nanos are encoded as `varint` in protobuf (see https://github.com/golang/protobuf/tree/master/ptypes/timestamp), we encode them with a fixed length instead `sfixed(32/64)`. I'll open an issue in amino. This doesn't block this PR though.

Liamsi: See: https://github.com/tendermint/go-amino/issues/208

PR #1743, opened 2018-06-13
all: gofmt

whilei:
Run 'gofmt -w .' from project root.

* [n/a] Updated all code comments where relevant
* [n/a] Wrote tests
* [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+* June 13, 2018
+
+SYNTAX
+
+- Isaac runs `gofmt` on the project.
+
+
 ## 0.20.0
 
 *June 6th, 2018*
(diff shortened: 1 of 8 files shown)


whilei:
@xla Hm, interesting... what do you mean generally? Is there an ambiguity in `gofmt`?

xla:
@whilei I'm not talking about the changes in this PR and that the order of the imports is not always correct.

whilei:
Can you point to an example?

As far as I see, these proposed changes only alphabetize a few _internal packages_ sections.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+* June 13, 2018
+
+SYNTAX
```
melekes: No need in the changelog entry for such changes. Let 's spare the time of the people who read it.

xla:
@whilei Thanks for your contribution! Ultimately we want to have even more consistency for the imports, which is out-of-scope for this PR.

PR #1759, opened 2018-06-17
update README

ebuchman:
- fixed link from rocket to riot
- added note on production readiness

NOTE: we should provide an actual contact point for folks looking to run it in production ...

Also need to update /security page to point to the live bounty!


ebuchman:
great idea!

PR #1808, opened 2018-06-23
crypto: Abstract pubkey / signature size when known to constants

ValarDragon:
* Created PubKeyEd25519Size as 32
* Created PubkeySecp256k1Size as 33
* Created SignatureEd25519Size as 64

Closes https://github.com/tendermint/go-crypto/issues/81

* [ ] Updated all relevant documentation in docs - is there any I should be updating?
* [X] Updated all code comments where relevant - n/a, constant names are self-descriptive
* [X] Wrote tests - n/a
* [X] Updated CHANGELOG.md - Did I update the right changelog?

Note, I am not updating secret connection in this PR, in favor of https://github.com/tendermint/tendermint/pull/1563 being settled first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,6 @@
 
 ## TBD
 
-BREAKING CHANGES:
- - [crypto] Remove xchacha20poly1305 in favor of hkdfchachapoly1305
-
 IMPROVEMENTS:
  - [crypto] Make public key size into public constants
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## TBD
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## TBD
 
+BREAKING CHANGES:
+ - [crypto] Remove xchacha20poly1305 in favor of hkdfchachapoly1305
```
melekes: why are you adding it here? was it done before?

ValarDragon: I had never added it to the changelog previously. I can remove it though

melekes: Please do

PR #1818, opened 2018-06-26
Adding IPv6 support to peering

kph7:
Adding IPv6 support to the p2p layer.  Tested between two IPv6 nodes.

### DIFF
--- a/p2p/listener.go
+++ b/p2p/listener.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/tendermint/tendermint/p2p/upnp"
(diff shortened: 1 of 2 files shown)


p2p/listener.go
```diff
@@ -159,6 +161,15 @@ func (l *DefaultListener) ExternalAddress() *NetAddress {
 	return l.extAddr
 }
 
+func (l *DefaultListener) ExternalAddressToString() string {
```
alexanderbez: Thoughts on adding a godoc for this guy?

mdyring:
Thanks for the work on this.

Is there a way to advertise both ipv4 and ipv6 addresses via external_address (config.toml)?

mdyring:
I can confirm this also worksfor RPC:

```
# netstat -l
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State
tcp6       0      0 [::]:26656              [::]:*                  LISTEN
tcp6       0      0 [::]:26657              [::]:*                  LISTEN
```

Issue #1765, opened 2018-06-19
Incorrect nil check


360CodeSafe:
https://github.com/tendermint/tendermint/blob/c84be3b8ddf46040567897f7d8163d35aa561a25/p2p/trust/metric.go#L261
Before check whether `tm` is `nil`,  there is `tm.mtx.Lock()`, which may cause a nil pointer error.

PR #1819, opened 2018-06-26
p2p/trust: Fix nil pointer error on TrustMetric Copy()

ricardohsd:
Fix the following issue: https://github.com/tendermint/tendermint/issues/1765


melekes:
Awesome. Thank you!

PR #1820, opened 2018-06-26
Adding configurability via config file to address book

kph7:
This diff adds the ability to configure elements in the address book via the config file.  For now, only adding the "Need address threshold" to the config file, but this would allow any configuration items to be added.  Also reduces number of elements we need to pass to NewAddrBook

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -311,6 +311,9 @@ type P2PConfig struct {
 	// FUzz connection
 	TestFuzz       bool            `mapstructure:"test_fuzz"`
 	TestFuzzConfig *FuzzConnConfig `mapstructure:"test_fuzz_config"`
+
+	// Addresses under which the address manager will claim to need more addresses
+	NeedAddressThreshold int `mapstructure:"need_address_threshold"`
 }
 
 // DefaultP2PConfig returns a default configuration for the peer-to-peer layer
@@ -332,6 +335,7 @@ func DefaultP2PConfig() *P2PConfig {
 		TestDialFail:            false,
 		TestFuzz:                false,
 		TestFuzzConfig:          DefaultFuzzConnConfig(),
+		NeedAddressThreshold:    1000,
 	}
 }
 
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -15,6 +15,7 @@ import (
 	crypto "github.com/tendermint/go-crypto"
 	"github.com/tendermint/tendermint/p2p"
 	cmn "github.com/tendermint/tmlibs/common"
+	"github.com/tendermint/tendermint/config"
 )
 
 const (
@@ -73,6 +74,8 @@ var _ AddrBook = (*addrBook)(nil)
 type addrBook struct {
 	cmn.BaseService
 
+	config     *config.P2PConfig
+
 	// immutable after creation
 	filePath          string
 	routabilityStrict bool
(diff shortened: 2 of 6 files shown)


p2p/pex/addrbook.go
```diff
@@ -208,7 +212,7 @@ func (a *addrBook) HasAddress(addr *p2p.NetAddress) bool {
 
 // NeedMoreAddrs implements AddrBook - returns true if there are not have enough addresses in the book.
 func (a *addrBook) NeedMoreAddrs() bool {
-	return a.Size() < needAddressThreshold
```
melekes: could you also remove `needAddressThreshold` const?

p2p/pex/addrbook.go
```diff
@@ -93,13 +96,14 @@ type addrBook struct {
 
 // NewAddrBook creates a new address book.
 // Use Start to begin processing asynchronous address updates.
-func NewAddrBook(filePath string, routabilityStrict bool) *addrBook {
+func NewAddrBook(cfg *config.P2PConfig) *addrBook {
```
melekes: Should we pass the whole config here or rather create a 

```
AddrBookConfig struct {
  Filepath string
  Strict bool
  NeedAddressThreshold int
}
```

P2P config is gonna grow over time and it might become inconvenient to create it every time say you need a book. Plus addrbook now depends on `config` package 👎 

I feel like 
```
NewAddrBook(filename string, config *AddrBookConfig) {
}
```
is a better looking API. Thoughts?

xla: In general it seems undesirable to have entities holding handles to (mutable) configuration structs. It leads to leakage of environmental concerns into business logic, e.g.: having a test related struct field checked in the middle of an algorithm. It leads to lazy implemenations and non-solid boundaries.

Instead APIs should express dependencies, implementations should come with sane defaults and configuration can be handled with options demonstrated in other parts of the codebase.

config/config.go
```diff
@@ -332,6 +335,7 @@ func DefaultP2PConfig() *P2PConfig {
 		TestDialFail:            false,
 		TestFuzz:                false,
 		TestFuzzConfig:          DefaultFuzzConnConfig(),
+		NeedAddressThreshold:    1000,
```
xla: How have you decided on this number?

p2p/pex/addrbook.go
```diff
@@ -15,6 +15,7 @@ import (
 	crypto "github.com/tendermint/go-crypto"
 	"github.com/tendermint/tendermint/p2p"
 	cmn "github.com/tendermint/tmlibs/common"
+	"github.com/tendermint/tendermint/config"
```
xla: Imports should format as follows:

``` go
package main

import (
	// stdlib

	// 3rd party

	// interal
)
```

p2p/pex/addrbook_test.go
```diff
@@ -11,9 +11,17 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/tendermint/tendermint/p2p"
 	cmn "github.com/tendermint/tmlibs/common"
+	"github.com/tendermint/tendermint/config"
```
xla: See comment on imports above.

p2p/pex/pex_reactor_test.go
```diff
@@ -84,7 +84,9 @@ func TestPEXReactorRunning(t *testing.T) {
 	// create switches
 	for i := 0; i < N; i++ {
 		switches[i] = p2p.MakeSwitch(cfg, i, "testing", "123.123.123", func(i int, sw *p2p.Switch) *p2p.Switch {
-			books[i] = NewAddrBook(filepath.Join(dir, fmt.Sprintf("addrbook%d.json", i)), false)
+			books[i] = NewAddrBook(&config.P2PConfig{
+				AddrBook: 					filepath.Join(dir, fmt.Sprintf("addrbook%d.json", i)),
```
xla: Formatting seems off.

p2p/pex/pex_reactor_test.go
```diff
@@ -84,7 +84,9 @@ func TestPEXReactorRunning(t *testing.T) {
...
+				AddrBookStrict: 	false})
```
xla: All struct fields should be terminated with a comma and a newline.

config/config.go
melekes: needAddressThreshold const

xla:
Hej @kph7, we really appreciate your contribution, but would like to address the issue you identified correctly in a more general change of the p2p package structure. To avoid spending too much time in the future we would like to encourage everyone to deliver a design document before actual implementation to identify if it is going to be a good fit for addition. For inspiration checkout the [ADR section](https://github.com/tendermint/tendermint/tree/master/docs/architecture#architecture-decision-records-adr).

Issue #1754, opened 2018-06-15
Tendermint stops building blocks


srmo:
**BUG REPORT**

**Tendermint version** 
0.20.0-27bd1dea

**ABCI app**
Custom closed Source. Based on jABCI (https://github.com/jTendermint/jabci)

**Environment**:
- OSX
- Binary release package
- 4 nodes, 3 validators, all running localhost
- UPDATE: 1 Validator is entered as seed/persistent peer in all other nodes.
- create_empty_blocks == false

**What happened**:
After several blocks, tendermint no longer creates new blocks on new transactions.
CheckTx runs successful and a `added good tx` message is found in the logs but nothing else happens.
There are several tx types. Most tx are around 10kb in size.
The tx that is tried to be sent is ~280bytes
Also, tendermint reports several older tx to be as duplicate in the mempool, no idea where that is coming from.

Please see the following dump consensus output of the validator node,  to which the tx is sent:

```
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    "round_state": {
      "height": 187,
      "round": 0,
      "step": 4,
      "start_time": "2018-06-15T17:11:48.984420411+02:00",
      "commit_time": "0001-01-01T00:00:00Z",
      "validators": {
        "validators": [
          {
            "address": "377218B1A5419768569B7829CE95C774F125DB35",
            "pub_key": {
              "type": "AC26791624DE60",
              "value": "8xEJkW3ghtHrLg0JBXvRi4xzUEy4KkuqMGtgODlkE/U="
            },
            "voting_power": 10,
            "accum": -20
          },
          {
            "address": "7F529D2F03B68924E5BB5F880C828CF8D134B8BD",
            "pub_key": {
              "type": "AC26791624DE60",
              "value": "eRmCyNdnKbsJWyFkSxLWcyduXKKdu/ZG2eue3ud3dA0="
            },
            "voting_power": 10,
            "accum": 10
          },
          {
            "address": "B5A8E53B84C94FB98C94C487FC19B0CC4F490DEA",
            "pub_key": {
              "type": "AC26791624DE60",
              "value": "xZYfYQGfrDi4dFF1ETY6qRGFfyZ6HUg7OpdRJpS1BAY="
            },
            "voting_power": 10,
            "accum": 10
          }
        ],
        "proposer": {
          "address": "377218B1A5419768569B7829CE95C774F125DB35",
          "pub_key": {
            "type": "AC26791624DE60",
[...]

PR #1824, opened 2018-06-27
Fix emptyPacketMsgSize calc. plus limit max request body size and other mics. changes

melekes:
Refs #1762 , #1754

TODO:
- [x] update changelog

**PLEASE DO NOT SQUASH WHEN MERGING**


srmo:
@melekes does it make sense for us to test this branch?

melekes:
no, not yet.

melekes:
@srmo can you give this branch a try now?

p2p/conn/connection.go
```diff
@@ -788,17 +788,16 @@ func (mp PacketMsg) String() string {
 // - EOF field key + byte = 2 bytes
 // - Bytes field key = 1 bytes
 // - Uvarint length of MustMarshalBinary(bytes) = 1 or 2 bytes
-// - Struct terminator = 1 byte
-// = up to 14 bytes overhead for the packet.
+// = up to 13 bytes overhead for the packet.
 
-func emptyPacketMsgSize() int {
+func emptyPacketMsgSize(maxPayloadSize int) int {
```
melekes: still feels like a dirty hack

melekes: OK, I am going to revert this to Jae's original solution with a small tweak

rpc/lib/server/http_server.go
```diff
@@ -202,3 +208,13 @@ func (w *ResponseWriterWrapper) WriteHeader(status int) {
 func (w *ResponseWriterWrapper) Hijack() (net.Conn, *bufio.ReadWriter, error) {
 	return w.ResponseWriter.(http.Hijacker).Hijack()
 }
+
+type maxBytesHandler struct {
+	h http.Handler
+	n int64
+}
+
+func (h maxBytesHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
alexanderbez: Could we add a quick godoc -- something simple like stating that `maxBytesHandler` implements the `http.Handler interface`?

config/config.go
```diff
@@ -299,9 +299,8 @@ type P2PConfig struct {
 	// Time to wait before flushing messages out on the connection, in ms
 	FlushThrottleTimeout int `mapstructure:"flush_throttle_timeout"`
 
-	// Maximum size of a message packet, in bytes
-	// Includes a header, which is ~13 bytes
-	MaxPacketMsgSize int `mapstructure:"max_packet_msg_size"`
+	// Maximum size of a message packet payload, in bytes
+	MaxPacketMsgPayloadSize int `mapstructure:"max_packet_msg_payload_size"`
```
ebuchman: We just changed this. Changing it back ?

consensus/state.go
```diff
@@ -932,7 +932,7 @@ func (cs *ConsensusState) createProposalBlock() (block *types.Block, blockParts
 	}
 
 	// Mempool validated transactions
-	txs := cs.mempool.Reap(cs.config.MaxBlockSizeTxs)
+	txs := cs.mempool.Reap(cs.state.ConsensusParams.BlockSize.MaxTxs)
```
ebuchman: Note this means this param will be set at genesis time and can't be changed. It also has an abnormally large default

config/config.go
melekes: yes. my bad

Jae's original code:

| -------- 1024 bytes ----------- | amino encoding 15 bytes |
limit: 1024 + 15
a) if amino changes, you need to change the `defaultAminoOverhead` const
b) 2 TM versions with different versions of amino become incompatible

My first version

| -------- 1024 bytes ----------- | (includes amino encoded header)
limit: 1024 + 15

although this version gives you control over the size of the whole package (not just payload), if you make a mistake in the calculation (like I did with emptyPacketMsgSize, it results in "maxSize is ..., but amino unmarshal 3+...." error.

My second version

| -------- 1024 bytes ----------- | amino encoding 15 bytes |
note: limit is set to 1024+15+10. 10 bytes protect us from errors above ^^ and give the opportunity to change something in amino (make it more compact / less compact)

melekes:
> Should we try to include this in release v0.22.0 ie. #1840 ?

I think we should.

consensus/state.go
melekes: > It also has an abnormally large default

set to 10000

PR #1827, opened 2018-06-28
config: rename skip_upnp to upnp

### LINKED ISSUE #1806: config: upnp false by default and flip flag
Most full nodes aren't interested in this and it just wastes time at start. Let's turn it off by default and change `--p2p.skip_upnp` to `--p2p.upnp`

ValarDragon:
Change default option to enable upnp.

Closes #1806

* [ ] Updated all relevant documentation in docs - I didn't see any 
* [ ] Wrote tests - n/a I think?
* [X] Updated CHANGELOG.md

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -285,7 +285,7 @@ type P2PConfig struct {
 	PersistentPeers string `mapstructure:"persistent_peers"`
 
 	// Skip UPNP port forwarding
-	SkipUPNP bool `mapstructure:"skip_upnp"`
+	UPNP bool `mapstructure:"upnp"`
 
 	// Path to address book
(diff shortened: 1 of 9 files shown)


melekes:
> Change default option to enable upnp.

it should be off by default

config/config.go
```diff
@@ -341,6 +341,7 @@ type P2PConfig struct {
 func DefaultP2PConfig() *P2PConfig {
 	return &P2PConfig{
 		ListenAddress:        "tcp://0.0.0.0:26656",
+		UPNP:                 false,
```
xla: This is superfluous as the zero value of bools is already `false`.

PR #1828, opened 2018-06-28
[p2p] add godoc comments to Listener

melekes:


xla:
@melekes Two nodes have been connected, but what about implications on addrbook and pex?

melekes:
> what about implications on addrbook and pex?

should be good. addrbook is using `NetAddress.String()`, which calls `IP.String()`. maybe we can write a test though

p2p/listener.go
```diff
@@ -147,21 +159,26 @@ func (l *DefaultListener) listenRoutine() {
...
 }
 
+// ExternalAddress returns the external NetAddress (publicly available,
+// determined using either UPnP or local resolver).
 func (l *DefaultListener) ExternalAddress() *NetAddress {
 	return l.extAddr
 }
 
-func (l *DefaultListener) ExternalAddressToString() string {
+// TODO: remove the need for this function by supporting IPv6 in NetAddress.
```
melekes: actually, we may already support IPv6 in NetAddress. `net.IP` can hold either IPv4 or IPv6.

melekes:
@xla what's exactly you don't like in this PR? changelog entry? I can remove it

xla:
@melekes If all components are supporting ipv6 then I all my concerns are adressed.

xla:
Sorry bad wording, what I meant is that this should be fine as nothing seem to break from it, so good to goz

PR #1832, opened 2018-06-29
Update ISSUE_TEMPLATE to mention using pastebin

### LINKED ISSUE #1826: meta: recommend pasting logs in pastebin or similar for issues
Many issues have extremely long output which we ask for in the issue template, which is necessary for debugging. However, these issues being extremely long makes it hard to scroll the issue / read the surrounding context. Could the issue template be changed to recommend putting long logs / config files inside of a pastebin / ghostbin / equivalent service? 

If we recommend pastebin, then we can easily update issues to snip out logs into pastebins. I think its more work to try to take a section of logs, and then save it to a file and then attaching it to the github issue.

ValarDragon:
closes #1826

* [X] Updated all relevant documentation in docs - n/a
* [X] Wrote tests - n/a
* [X] Updated CHANGELOG.md - n/a i think?

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -33,7 +33,7 @@ in a case of bug.
 
 **How to reproduce it** (as minimally and precisely as possible):
 
-**Logs (you can paste a small part showing an error or link a pastebin containing more of the log file)**:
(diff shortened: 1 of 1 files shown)


melekes:
Thanks!

PR #1840, opened 2018-07-01
Release/v0.22.0

ebuchman:
- Merge in go-crypto
- Merge in abci
- Merge in tmlibs
- Prometheus metrics
- Stability fixes
- Various other ...


ebuchman:
Note we may want to include https://github.com/tendermint/tendermint/pull/1824

ebuchman:
Also need to update the spec
- new hash functions
- Updated the Amino crypto table

consensus/state.go
```diff
@@ -320,10 +320,7 @@ func (cs *ConsensusState) OnStop() {
 
 	cs.timeoutTicker.Stop()
 
-	// Make BaseService.Wait() wait until cs.wal.Wait()
```
melekes: what is the motivation behind this change?

ebuchman: SDK tests are failing non-deterministically with "cant open wal". Those tests do a lot of stopping and starting. So I went looking into this and noticed that the code here actually never ran and we never stopped the WAL - unless I'm missing something.

PR #1842, opened 2018-07-02
Bucky/merge tmlibs

ebuchman:
Merges all of tmlibs into Tendermint.

Thought it would make sense to get all the repo merges into v0.22.0.


melekes:
```
libs/autofile/cmd/logjack.go:16:1:warning: sleepSeconds is unused (deadcode)
libs/db/debug_db.go:10:1:warning: _fmt is unused (deadcode)
libs/bech32/bech32_test.go:28:5:warning: should use !bytes.Equal(data, sum[:]) instead (S1004) (gosimple)
libs/common/async.go:79:3:warning: should use a simple channel send/receive instead of select with a single case (S1000) (gosimple)
libs/common/errors_test.go:20:4:warning: redundant return statement (S1023) (gosimple)
```

PR #1843, opened 2018-07-02
Bucky/external address

ebuchman:
Develop tests fail locally for me with:

```
ethanbuchman:tendermint$ go test ./p2p/
--- FAIL: TestListener (0.00s)
        listener_test.go:18: Could not connect to listener address [fe80::1]:8001
FAIL
FAIL    github.com/tendermint/tendermint/p2p    2.530s
```

Not sure exactly why, but I changed the behaviour here so that if you don't specify an IP address, it defaults to IPv4. I think we should still have defaults as IPv4 for now, but let people enable IPv6 by specifying it specifically.

Also, while at it, I added ExternalAddress to the config so folks can specify where peers should dial.

### DIFF
--- a/p2p/listener.go
+++ b/p2p/listener.go
@@ -7,9 +7,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/tendermint/tendermint/p2p/upnp"
+	"github.com/tendermint/tendermint/config"
 	cmn "github.com/tendermint/tendermint/libs/common"
 	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/p2p/upnp"
 )
 
 // Listener is a network listener for stream-oriented protocols, providing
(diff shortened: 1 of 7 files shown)


p2p/listener.go
```diff
@@ -88,17 +92,29 @@ func NewDefaultListener(protocol string, lAddr string, UPNP bool, logger log.Log
 		panic(err)
 	}
 
+	inAddrAny := lAddrIP == "" || lAddrIP == "0.0.0.0"
+
 	// Determine external address...
 	var extAddr *NetAddress
-	if UPNP {
+	if cfg.UPNP {
```
melekes: these two `if` statements can be combined

p2p/listener.go
```diff
@@ -88,17 +92,29 @@ func NewDefaultListener(protocol string, lAddr string, UPNP bool, logger log.Log
...
 		// If the lAddrIP is INADDR_ANY, try UPnP
-		if lAddrIP == "" || lAddrIP == "0.0.0.0" {
+		if inAddrAny {
 			extAddr = getUPNPExternalAddress(lAddrPort, listenerPort, logger)
 		}
 	}
+
+	if cfg.ExternalAddress != "" {
```
melekes: I think this should be the first check. And only if external_address is "", we should proceed with getUPNPExternalAddress

```
if cfg.ExternalAddress != "" {
  ...
} else if cfg.UPNP && inAddrAny {
  ...
}
```

p2p/listener.go
```diff
@@ -59,8 +61,10 @@ func splitHostPort(addr string) (host string, port int) {
 
 // NewDefaultListener creates a new DefaultListener on lAddr, optionally trying
 // to determine external address using UPnP.
-func NewDefaultListener(protocol string, lAddr string, UPNP bool, logger log.Logger) Listener {
-	// Local listen IP & port
+func NewDefaultListener(cfg *config.P2PConfig, logger log.Logger) Listener {
```
melekes: see https://github.com/tendermint/tendermint/pull/1820#discussion_r198367012 discussion

ebuchman:
Thanks @melekes all comments addressed

Issue #1890, opened 2018-07-03
Tendermint can hang forever if abci app dies


ethanfrey:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
 
BUG REPORT

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):

0.20 (verified with unit test on 0.22+develop)

**ABCI app** (name for built-in, URL for self-written if it's publicly available):

**Environment**:
- **OS** (e.g. from /etc/os-release):
- **Install tools**:
- **Others**:

**What happened**:

Some tendermint nodes got stuck, in that they responded to rpc for status, etc, but they did not add more blocks not did they pass on any proposed transactions to other nodes. These were non-validating proxy nodes exposing rpc to the outside.

Digging through the log files, I found this on the tendermint fullnode logs:

(There were many such websocket messages, but this seems to be the first one with an error)
```
I[07-01|18:54:21.044] We need more addresses. Sending pexRequest to random peer module=p2p peer="Peer{MConn{10.24.2.65:46656} 5c6e058cafccbdc156557b4570a52bda9c23f7a0 out}
I[07-01|18:54:51.041] We need more addresses. Sending pexRequest to random peer module=p2p peer="Peer{MConn{10.24.10.10:46656} 778cfa655a5d70a2c356947f90b19bfab83b8c7c out}
E[07-01|18:54:51.948] Error closing connection                     module=rpc-server protocol=websocket remote=127.0.0.1:60496 err="close tcp 127.0.0.1:46657->127.0.0.1:60496: use of closed network connection"
E[07-01|18:54:55.743] Stopping abci.socketClient for error: EOF    module=abci-client connection=consensus
E[07-01|18:54:55.743] Stopping abci.socketClient for error: EOF    module=abci-client connection=mempool
E[07-01|18:54:55.743] Stopping abci.socketClient for error: EOF    module=abci-client connection=query
...
E[07-01|18:55:37.139] Error closing connection                     module=rpc-server protocol=websocket remote=127.0.0.1:44472 err="close tcp 127.0.0.1:46657->127.0.0.1:44472: use of closed network connection" 
E[07-01|18:55:41.348] Connection failed @ sendRoutine              module=p2p peer=10.24.5.59:46656 conn=MConn{10.24.5.59:46656} err="pong timeout"
[...]

PR #1891, opened 2018-07-03
abciClient.BeginBlockSync should not hang on crashed server 

ethanfrey:
* [x] Wrote tests
* [x] Updated CHANGELOG.md

Addressing issue #1890

Reproduced the errant behavior, and pushing to demo that.
Working on a solution.

### DIFF
--- a/abci/client/socket_client_test.go
+++ b/abci/client/socket_client_test.go
@@ -5,7 +5,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/tendermint/tendermint/abci/client"
+	"github.com/tendermint/tendermint/abci/server"
+	"github.com/tendermint/tendermint/abci/types"
 )
 
 func TestSocketClientStopForErrorDeadlock(t *testing.T) {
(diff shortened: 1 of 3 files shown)


abci/client/socket_client_test.go
```diff
@@ -26,3 +33,94 @@ func TestSocketClientStopForErrorDeadlock(t *testing.T) {
...
+		require.True(t, ok, "Must not close channel")
+		assert.Error(t, err, "We should get EOF error")
+	}
+}
+
+func setupClientServer(t *testing.T, app types.Application) (
+	cmn.Service, abcicli.Client) {
+	// some port between 20k and 30k
+	port := 20000 + cmn.RandInt32()%10000
+	addr := fmt.Sprintf("localhost:%d", port)
```
melekes: can't we just use `:0` as port here, so server would pick ephemeral port for us?

melekes: I think the server should expose `Addr()` method, but this is for another PR. thanks

```
func (s *Server) Addr() net.Addr
Addr returns the server's listener's network address, either a *TCPAddr or *UnixAddr.
```

abci/client/socket_client_test.go
```diff
@@ -26,3 +33,94 @@ func TestSocketClientStopForErrorDeadlock(t *testing.T) {
...
+	defer c.Stop()
+
+	resp := make(chan error, 1)
+	go func() {
+		// Start BeginBlock and flush it
+		reqres := c.BeginBlockAsync(types.RequestBeginBlock{})
+		flush := c.FlushAsync()
+		// wait 20 ms for all events to travel socket, but
+		// no response yet from server
+		time.Sleep(20 * time.Millisecond)
```
ebuchman: this seems non-deterministic. is there some way we can test this without depending on sleeps ?

abci/client/socket_client_test.go
```diff
@@ -26,3 +33,94 @@ func TestSocketClientStopForErrorDeadlock(t *testing.T) {
...
+		// kill the server, so the connections break
+		s.Stop()
+
+		// wait for the response from BeginBlock
+		fmt.Println("waiting for begin block")
```
ebuchman: can we get rid of the print statements please?

abci/client/socket_client_test.go
ethanfrey: I couldn't think of any way without major changes, as we wait for another goroutine to process an event and there is no clean way to hook into some event system.

Good thing to keep in mine for a refactor of this code.

abci/client/socket_client_test.go
ethanfrey: Yeah, my bad.

I'm away from coding machines for a few days. Can you just push that onto this branch? I think I gave maintainers push access.

melekes: Done in https://github.com/tendermint/tendermint/pull/1950

PR #1898, opened 2018-07-04
state: err if 0 power validator is added to the validator set

### LINKED ISSUE #1893: state:  err if 0 power validator is added to the validator set
We recently caught a bug in the SDK where a validator update was returned for a validator with 0 power that wasn't already part of the validator set (https://github.com/cosmos/cosmos-sdk/issues/1470). Tendermint ended up adding this validator to the set but with 0 power. 

It's the responsibility of the application to not do this (it needs to track the validator correctly), but Tendermint can help by throwing an error right away rather than allowing a 0 power validator.

melekes:
Closes #1893

* [x] Wrote tests
* [ ] ~~Updated CHANGELOG.md~~

### DIFF
--- a/state/execution.go
+++ b/state/execution.go
@@ -284,17 +284,17 @@ func updateValidators(currentSet *types.ValidatorSet, abciUpdates []abci.Validat
 
 		address := valUpdate.Address
 		_, val := currentSet.GetByAddress(address)
-		if val == nil && valUpdate.VotingPower != 0 {
-			// add val
-			added := currentSet.Add(valUpdate)
-			if !added {
-				return fmt.Errorf("Failed to add new validator %v", valUpdate)
-			}
-		} else if valUpdate.VotingPower == 0 {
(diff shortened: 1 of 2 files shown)


state/execution.go
```diff
@@ -278,9 +278,13 @@ func updateValidators(currentSet *types.ValidatorSet, abciUpdates []abci.Validat
...
 	// these are tendermint types now
 	for _, valUpdate := range updates {
+		if valUpdate.VotingPower < 0 {
+			return fmt.Errorf("Voting power can't be negative %v", valUpdate)
+		}
+
 		address := valUpdate.Address
 		_, val := currentSet.GetByAddress(address)
-		if val == nil {
+		if val == nil && valUpdate.VotingPower != 0 {
```
xla: If we expect it to be larger than `0` why not check for it instead of unqueality check?

melekes: 👍 will do

Issue #1839, opened 2018-07-01
consensus: WAL isn't shut down properly


ebuchman:
See https://github.com/tendermint/tendermint/blob/f760c24ff0df7cd150ca1e10a16cdc70110fac79/consensus/state.go#L301

By this point, IsRunning will always be false, so the cs.wal.Wait is never called. But it also looks like the WAL is never even stopped? 

This came up because the SDK tests fails non-deterministically sometimes because the autofile panics. Those tests start and stop Tendermint nodes in process.

melekes:
Done in https://github.com/tendermint/tendermint/commit/af703620d4423478968ab602716a2a51cbb02ac6#diff-80343e80650fe43a4394fc26afd5cb3b

melekes:
What about this stop?

https://github.com/tendermint/tendermint/blob/f760c24ff0df7cd150ca1e10a16cdc70110fac79/consensus/state.go#L605

Do you still have a stack trace of the panic mentioned above?

ebuchman:
Seems I missed that. I may have been wrong to add the Stop where I did - it's not even clear that was actually the issue anymore. But it does seem the IsRunning is always false by then, so that Wait() call never happens. I believe removing the IsRunning caused it to hang, too.

melekes:
> But it does seem the IsRunning is always false by then, so that Wait() call never happens

true, but removing it does not achieve anything other then deleting 3 lines of code.

PR #1899, opened 2018-07-04
Properly shutdown consensus WAL

melekes:
Refs #1839


consensus/state.go
```diff
@@ -314,20 +310,8 @@ func (cs *ConsensusState) startRoutines(maxSteps int) {
...
-
 	cs.timeoutTicker.Stop()
-
-	cs.wal.Stop()
-}
-
-// Wait waits for the the main routine to return.
-// NOTE: be sure to Stop() the event switch and drain
-// any event channels or this may deadlock
-func (cs *ConsensusState) Wait() {
```
melekes: unused

PR #1906, opened 2018-07-04
clean up abci/libs directories

### LINKED ISSUE #1793: Consolidate metadata for merged repos
After the merge of projects like abci, crypto and tmlibs we should consolidate licenses, build infrastructure, tests, docs and commands.

#1782
#1790 
#1842

zramsay:
closes #1793 

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


zramsay:
@melekes see new changes since you approved

melekes:
I am a bit concerned with Tendermint Makefile getting too big. Plus it will be hard to decouple makefiles back once this is done. Maybe it will be better if some subrepos retain parts of their original Makefiles (get_protoc, install_protoc, etc..). Thoughts?

zramsay:
i somewhat agree that it's getting too big although it probably shouldn't grow much past this size at all. There a bunch of outdated stuff that can probably go and a lot of of the networks stuff will move to networks/Makefile (already the case in SDK). protoc is used by two sub directories so for deduplication I think it makes sense for it to be in the root Makefile

xla:
I'd say having consolidation outweighs size. Maintainability is largely determined by structure, so as long as we enfore that and are adamant about our Makefile organisation we should get away with it.

PR #1915, opened 2018-07-05
Release/v0.22.1

ebuchman:
* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


ebuchman:
This is mostly about getting the docs improvements to master before more breaking changes hit develop (like https://github.com/tendermint/tendermint/pull/1815).

#1815 will have a significant effect on apps that change the validator set.

As for the "error on adding 0-validator", see this for another take https://github.com/cosmos/cosmos-sdk/issues/1541 - we may want to undue it if we want to support app devs having 0-power validators so they can track full nodes ...

xla:
@ebuchman LGTM. Should we get this out?

PR #1925, opened 2018-07-09
Peng/deprecate aib data

nylira:
* [x] Updated CHANGELOG.md

We're deprecating the `tendermint/aib-data` repo and the Tendermint software ecosystem data should be moved here. It cannot be in the Tendermint website repo because that is closed source.


nylira:
Whoops, shouldn't be merging to master...

PR #1927, opened 2018-07-09
limit the number of requests in flights for Prometheus server

### LINKED ISSUE #1804: Limit max open connections for Prometheus server
Same thing we did here https://github.com/tendermint/tendermint/pull/1781

Also, do we have any other servers we haven't wrapped yet with `LimitedListener`?

melekes:
Closes #1804

Default to 1 because usually there's just one collector.

* [ ] ~~Wrote tests~~
* [x] Updated CHANGELOG.md

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -620,7 +620,7 @@ func DefaultInstrumentationConfig() *InstrumentationConfig {
 	return &InstrumentationConfig{
 		Prometheus:           false,
 		PrometheusListenAddr: ":26660",
-		MaxOpenConnections:   1,
+		MaxOpenConnections:   3,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


docs/tendermint-core/configuration.md
```diff
@@ -209,4 +209,10 @@ prometheus = false
 
 # Address to listen for Prometheus collector(s) connections
 prometheus_listen_addr = ":26660"
+
+# Maximum number of simultaneous connections.
+# If you want to accept more significant number than the default, make sure
```
zramsay: accept `a` more significant

config/config.go
```diff
@@ -614,6 +620,7 @@ func DefaultInstrumentationConfig() *InstrumentationConfig {
 	return &InstrumentationConfig{
 		Prometheus:           false,
 		PrometheusListenAddr: ":26660",
+		MaxOpenConnections:   1,
```
xla: This default should be higher, given that a typical HA setup for Prometheus is identical active/active servers.

melekes: Right. Thanks!

PR #1928, opened 2018-07-09
Unexport DecodeMessage

### LINKED ISSUE #646: all reactors: unexport DecodeMessage and defend against empty input
While working on https://github.com/orijtech/tendermint/commit/8c13a65bdd0396888a3dba7db8cf7d1396fbc098, which is a fix for https://github.com/tendermint/tendermint/issues/514, and at develop https://github.com/tendermint/tendermint/commit/54c63726b03cbba067d10fa6b69b885c85678d17, I have to write a test for the new change. However, I noticed that inside `blockchain/reactor.go` there is an exported function `DecodeMessage` https://github.com/tendermint/tendermint/blob/54c63726b03cbba067d10fa6b69b885c85678d17/blockchain/reactor.go#L291-L302.

That function is untested but also doesn't seem to be used anywhere in the other repos.
Grep-ing for `\.DecodeMessage` outside of this repo, gives 686 spurious hits that can be examined at https://gist.github.com/odeke-em/657adf5c63e0b4866f67c9508e25eb40, which you'll notice are spurious hits because the return signatures don't match up

I suggest that we unexport this function, to avoid API bloat as it isn't used anywhere elese but in the same file, and moreover isn't unit tested.

melekes:
Closes #646 

"let's unexport all of them and ensure their inputs aren't empty (even though reactors should guarantee the input is not empty)"

the second part is now handled by Amino.

* [x] Wrote tests
* [ ] ~~Updated CHANGELOG.md~~

### DIFF
--- a/p2p/pex/pex_reactor_test.go
+++ b/p2p/pex/pex_reactor_test.go
@@ -134,11 +134,11 @@ func TestPEXReactorReceive(t *testing.T) {
 
 	size := book.Size()
 	addrs := []*p2p.NetAddress{peer.NodeInfo().NetAddress()}
-	msg := cdc.MustMarshalBinary(&pexAddrsMessage{Addrs: addrs})
+	msg := cdc.MustMarshalBinaryBare(&pexAddrsMessage{Addrs: addrs})
 	r.Receive(PexChannel, peer, msg)
 	assert.Equal(t, size+1, book.Size())
 
-	msg = cdc.MustMarshalBinary(&pexRequestMessage{})
+	msg = cdc.MustMarshalBinaryBare(&pexRequestMessage{})
 	r.Receive(PexChannel, peer, msg) // should not panic.
 }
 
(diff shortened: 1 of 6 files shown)


p2p/pex/pex_reactor_test.go
```diff
@@ -134,11 +134,11 @@ func TestPEXReactorReceive(t *testing.T) {
 
 	size := book.Size()
 	addrs := []*p2p.NetAddress{peer.NodeInfo().NetAddress()}
-	msg := cdc.MustMarshalBinary(&pexAddrsMessage{Addrs: addrs})
+	msg := cdc.MustMarshalBinaryBare(&pexAddrsMessage{Addrs: addrs})
```
xla: How are the two funcs differrent?

xla: Couldn't find it in the docs.

melekes: MustMarshalBinary adds the uvarint length of the whole binary (before the actual data).

PR #1947, opened 2018-07-10
Add OSX install script

### LINKED ISSUE #1931: docs: Make mac osx installation script
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): BUG REPORT

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source): v 0.22.0

**Environment**:
- **OS** (e.g. from /etc/os-release): Mac oSX
- **Install tools**:
- **Others**:

**What happened**:
errors, unable to tendermint

what dev said would happen

run sudo sh ./installation_script_ubuntu.sh on mac osx
**Logs (you can paste a small part showing an error or link a pastebin, gist, etc. containing more of the log file)**:
https://pastebin.com/AgYgMs1B
mac osx

jackzampolin:
This PR adds an OSX install script. Instead of installing `golang` and other deps from scratch it prompts users to do so if they have not done so already.

closes #1931

### DIFF
--- a/scripts/install/install_tendermint_osx.sh
+++ b/scripts/install/install_tendermint_osx.sh
@@ -3,9 +3,6 @@
 # XXX: this script is intended to be run from
 # an MacOS machine
 
-# upon its completion, you must either reset
-# your terminal or run `source ~/.profile`
-
 # as written, this script will install
 # tendermint core from master branch
 REPO=github.com/tendermint/tendermint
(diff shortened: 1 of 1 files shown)


scripts/install/install_tendermint_osx.sh
```diff
@@ -0,0 +1,48 @@
...
+# build & install
+git checkout $BRANCH
+# XXX: uncomment if branch isn't master
+# git fetch origin $BRANCH
+make get_tools
+make get_vendor_deps
+make install
+
+# the binary is located in $GOPATH/bin
+# run `source ~/.profile` or reset your terminal
```
melekes: copy pasta?

jackzampolin: Ah didn't remove that. I'll ge tit.

jackzampolin:
closes https://github.com/tendermint/tendermint/issues/1931

melekes:
`brew install` requires sudo. Other commands should not require sudo.

xla:
Homebrew assumes `/usr/local` is owned by the user, so that all its commands dont need to be run with sudo. @melekes Is that not the case here?

melekes:
I am not sure, but if you have 2 user accounts: 1 with root privileges and 1 regular user, who don't have access to /usr/local, you might need some magic.

melekes:
Not sure. Let's just merge this for now.

xla:
@melekes Mulit-user setup is indeed not accounted for in Homebrew, but is a mild disaster in general on os x (just recently had this issue). Don't think something to optimise for.

Issue #1957, opened 2018-07-11
crypto: completely remove dependency on amino for signatures


ValarDragon:
Ref https://github.com/tendermint/go-crypto/issues/125, https://github.com/tendermint/go-crypto/issues/121

The reason we should upgrade: ~Currently the amino coding is path dependent (hence things broke when we switched the repo), this is bad.~ We don't want to use amino / protobuf, since that is less space efficient and upgradability doesn't make sense here. Most cryptosystems define canonical encoding formats anyway. Upgradability doesn't matter here, since if there is something that changes how the pubkeys are internally encoding, that means the respective crypto libraries need to be updated. For systems with security parameters, either their are already publically agreed upon parameters (i.e. elliptic curves), so changing those requires a new key type, or they are included in the pubkey canonical encoding (i.e. RSA modulus size). If there is a critical change that needs to upgrade the encoding format, I argue that it should have a different enum type, rather than using the existing enum type. 

My proposal for how we should do pubkey encoding:

We make an enum for pubkey types. (value being a uint). We use var uint encoding, and we marshal / unmarshal these pubkeys as 
`varuint(PUBKEY_TYPE_CONSTANT) || canonical pubkey encoding`
We also include the pubkey type constant when hashing to get addresses. 

Signatures can be encoded as raw byteslices, and be parsed by their respective pubkey. (This means removing the Signature struct / interface)

Privkeys can be encoded in the native format. Since the privkeys satisfy interfaces, this can later accommodate future caching for efficiency purports. 

/cc @cwgoes , @Liamsi

Liamsi:
Quick remark to:

>  Currently the amino coding is path dependent (hence things broke when we switched the repo), this is bad.

How so? The prefix bytes are derived by the registered name (not the path of the struct in question). See RegisterConcrete / RegisterInterface in codec.go. 

That said, I'm generally in favour of moving to standard encoding formats for keys (not for sending them around but for actually storing them).

> Privkeys can be encoded in the native format. 

What is the native format here?

ValarDragon:
> How so? The prefix bytes are derived by the registered name (not the path of the struct in question). See RegisterConcrete / RegisterInterface in codec.go.

Oh, my bad. I thought it was caused by the interface locations changing. You're right then. 

> What is the native format here?

[...]

PR #1960, opened 2018-07-12
[ADR] Proposal for multisignature encoding

ValarDragon:
This proposal is partially tied to the resolution of #1957. Link to markdown rendered version: https://github.com/tendermint/tendermint/blob/00f2225ba3ee3af6aa93002da1e4823621dda6d4/docs/architecture/adr-012-multisigs.md

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


docs/architecture/adr-012-multisigs.md
```diff
@@ -0,0 +1,135 @@
...
+ASM's don't have constant size pubkeys. 
+The same log stands for length prefixing signatures.
+
+\*\* The reason this doesn't use `log_8(n)` bytes per signer is because that heavily optimizes for the case where a very small number of signers are required.
+e.g. for `n` of size `24`, that would only be more space efficient for `k < 3`. 
+This seems less likely, and that it should not be the case optimized for.
+
+\*\*\* Important decision: 
+There is advocacy for using amino encoding of an `[][]byte`, instead of having a signature blob and parsing by reading the length prefix. (Note both are doing the same internally) 
+Protobuf requires having a "field number" on each signature (this is a design choice in protobuf encoding of slices), which means an extra byte on all signatures relative to the proposed encoding scheme, in addition to a constant overhead.
```
Liamsi: Haven't read this in detail to have an informed opinion yet. Just to let you know: while in proto3 a fieldnumber is mandatory, the amino API allows to encode (length-prefixed) raw bytes, without a field number. Not sure if this good or bad (my first intuition would be that this shouldn't be doable with amino). We even make sue of this already in secret connection (no fields / structs involved):
https://github.com/tendermint/tendermint/blob/2aa2b63cadc42cca1071c36adfd2f2ce14e1aa8f/p2p/conn/secret_connection.go#L199-L213

zmanian:
I am huge fan of Pact's implementation here and it might be good to review.
https://github.com/kadena-io/pact/blob/master/docs/pact-reference.rst#keysets-and-authorization

ValarDragon:
Their implementation is cool! My understanding is that its limited in the sense that you don't get the ability to encode more information within the key set though. You can only encode more information in your own custom predicate. We want to be able to have k of n multisigs under the same predicate, which means that `k` needs to be encoded with the keys themselves, so we couldn't use that method.

ValarDragon:
I need to update this proposal now that we've decided to not use amino

ValarDragon:
This is now ready for review!

docs/architecture/adr-013-multisigs.md
```diff
@@ -0,0 +1,148 @@
...
+func SigToMultisig(sig Signature, n int)
+func GetIndex(pk crypto.Pubkey, []crypto.Pubkey])
+func AddSignature(sig Signature, index int, multiSig *Multisignature)
+```
+The multisignature will be converted to an `[]byte` using amino.MarshalBinaryBare. \*
+
+#### Bit Array
+We would be using a new implementation of a bitarray. The struct it would be encoded/decoded from is 
+```golang
+type CryptoBitArray struct {
```
zmanian: Maybe just call it CompactBitArray. Maybe the other bit array can be refactored to FastBitArray

docs/architecture/adr-013-multisigs.md
```diff
@@ -0,0 +1,148 @@
...
+
+It is important to recall that each private key will output a signature on the provided message itself.
+So no signing algorithm ever outputs the multisignature.
+The UI will take a signature, cast into a multisignature, and then keep adding
+new signatures into it, and when done marshal into `[]byte`.
+This will require the following helper methods:
+```golang
+func SigToMultisig(sig Signature, n int)
+func GetIndex(pk crypto.Pubkey, []crypto.Pubkey])
+func AddSignature(sig Signature, index int, multiSig *Multisignature)
```
Liamsi: Is `Signature` a type alias for []byte here? Or, is this not updated to reflect the assumed changes from #1957)?

docs/architecture/adr-013-multisigs.md
```diff
@@ -0,0 +1,148 @@
...
+	Elems            []byte    `json:"elems"`
+}
+```
+The reason for not using the BitArray currently implemented in `libs/common/bit_array.go`
+is that it is less space efficient, due to a space / time trade-off.
+Evidence for this is outlined in [this issue](https://github.com/tendermint/tendermint/issues/2077). 
+
+In the multisig, we will not be performing arithmetic operations,
+so there is no performance increase with the current implementation,
+and just loss of spatial efficiency.
```
Liamsi: Am I the only one who is  not entirely sure what "spatial efficiency" in this context is?

docs/architecture/adr-013-multisigs.md
```diff
@@ -0,0 +1,148 @@
...
+
+#### Weighted threshold signature
+
+The pubkey is the following struct:
+
+```golang
+type WeightedThresholdMultisignatureKey struct {
+	Weights []uint             `json:"weights"`
+	Threshold uint             `json:"threshold"`
+	Pubkeys []crypto.Pubkey    `json:"pubkeys"`
```
Liamsi: Not sure where these `WeightedThresholdMultisignatureKey` will be used or why they are needed, but I guess this has sth to do with staking? 

I'm wondering if we shouldn't have a type `WeightedPubkeys` instead of two loosely coupled arrays? "Weights and Pubkeys must be of the same length." would be autom. enforced by that. What do you think?

docs/architecture/adr-013-multisigs.md
```diff
@@ -0,0 +1,148 @@
...
+#### Footnotes
+\* I advocate that we should actually include the fields from CryptoBitArray into the multisignature directly. This saves space in encoding, despite breaking abstractions slightly. I think it is worth the space savings, however I've put this into a footnote since this is quite minor. 
+
+### Future points of discussion
+
+If desired, we can use ed25519 batch verification for all ed25519 keys.
+This is a future point of discussion, but would be backwards compatible as this information won't need to be marshalled.
+(There may even be cofactor concerns without ristretto)
+Aggregation of pubkeys / sigs in Schnorr sigs / BLS sigs is not backwards compatible, and would need to be a new ASM type.
+
```
Liamsi: Will we have simple, non-aggregated (non-schnorr, non-BLS) multi-sigs pre launch?

docs/architecture/adr-013-multisigs.md
```diff
@@ -0,0 +1,148 @@
...
+We store that difference in ExtraBitsStored.
+This allows for us to have an unbounded number of signers, and is more space efficient than what is currently used in `libs/common`.
+Again the implementation of this space saving feature is straight forward.
+
+### Encoding the structs
+
+We will use straight forward amino encoding. This is chosen for ease of compatibility in other languages. 
+
+#### Footnotes
+\* I advocate that we should actually include the fields from CryptoBitArray into the multisignature directly. This saves space in encoding, despite breaking abstractions slightly. I think it is worth the space savings, however I've put this into a footnote since this is quite minor.
```
Liamsi: Not sure what this means: You want basically:
```golang
type Multisignature struct {
	BitArray    CryptoBitArray // Documented later
	Sigs        [][]byte       
        // Bit array:
        ExtraBitsStored  byte      `json:"extra_bits"` // The number of extra bits in elems. 
	Elems            []byte    `json:"elems"`
}
```
Is that right? I think it's semantically OK to include the bitmask into the multi-sig.

docs/architecture/adr-013-multisigs.md
zmanian: This is a desired prelaunch feature

docs/architecture/adr-013-multisigs.md
ValarDragon: It would be []byte, didn't update this for that

docs/architecture/adr-013-multisigs.md
ValarDragon: The reason I proposed it like this is that it saves space in encoding. Using a WeightedPubkey struct would then have to encode both of the field numbers for each pubkey.

Weighted Threshold Multisignatures are intended for use in the cosmos hub.

docs/architecture/adr-013-multisigs.md
ValarDragon: Yes, that was basically what I was advocating for. (But no BitArray there)
On further thought though, I don't think this is necessary. We can just keep it as a "CompactBitArray" there, and if we ever need more space efficiency, we encode in a method that is different than amino in the signature.

docs/architecture/adr-013-multisigs.md
ValarDragon: I will reword this to "it just takes more space."

zmanian: Cause `[1]uint64` == `[4]byte` in length so CompactBitArray allows for more space efficient bit arrays

docs/architecture/adr-013-multisigs.md
ebuchman: What does `both of the field numbers` mean here?

docs/architecture/adr-013-multisigs.md
```diff
@@ -0,0 +1,148 @@
...
+Every ASM will then have its own struct, implementing the crypto.pubkey interface.
+
+This ADR assumes that [replacing crypto.Signature with []bytes](https://github.com/tendermint/tendermint/issues/1957) has been accepted.
+
+#### K of N threshold signature
+
+The pubkey is the following struct:
+
+```golang
+type ThresholdMultisignatureKey struct { // K of N threshold multisig
```
ebuchman: should probably be `ThresholdMultiSignaturePubKey`

docs/architecture/adr-013-multisigs.md
```diff
@@ -0,0 +1,148 @@
...
+The reason this doesn't use `log_8(n)` bytes per signer is because that heavily optimizes for the case where a very small number of signers are required.
+e.g. for `n` of size `24`, that would only be more space efficient for `k < 3`. 
+This seems less likely, and that it should not be the case optimized for.
+
+#### Weighted threshold signature
+
+The pubkey is the following struct:
+
+```golang
+type WeightedThresholdMultisignatureKey struct {
```
ebuchman: `WeightedThresholdMultiSignaturePubKey`

docs/architecture/adr-013-multisigs.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Positive
+* Supports multisignatures, in a way that won't require any special cases in our downstream verification code. 
+* Easy to serialize / deserialize
+* Unbounded number of signers
+
+### Negative
+* Larger codebase, however this should reside in a subfolder of tendermint/crypto, as it provides no new interfaces. (Ref #https://github.com/tendermint/go-crypto/issues/136)
+* Space inefficient due to utilization of amino encoding
+* Suggested implementation requires a new struct for every ASM.
```
ebuchman: A new struct for every ASM ?

docs/architecture/adr-013-multisigs.md
ValarDragon: Every time you encoded a Pubkey then, the amino encoded version would have two field numbers, one for the weight, one for the actual pubkey. 

Having a single "Weights" slice and "pubkeys" slice allows saving the field number from being encoded for each weight/pubkey pair.

docs/architecture/adr-013-multisigs.md
ValarDragon: This is required since we're registering another concrete type with amino for each ASM we want to support.

PR #1964, opened 2018-07-13
tools/tmbench: Improve accuracy with large tx sizes.

ValarDragon:
At larger tx sizes (e.g. > 10000) we were spending non-neglible
amounts of time in tx creation, due to making the final bytes random.
The slower the send loop, the less accurate it is at measuring the time
tendermint took. (As we can't reach the promised contract of the given rate)

There really isn't much need for that randomness, so this PR makes it
such that only the txNumber gets bumped between txs from the same
connection, thereby improving sendloop speed and accuracy.

This PR is intendeded for develop, just aimed at `dev/tm_bench_ignore_first_block` so the diff is easier to view. 

* [X] Updated all relevant documentation in docs - this doesn't conflict with whats written in the doc.
* [ ] Updated all code comments where relevant - I think I did
* [X] Wrote tests
* [ ] Updated CHANGELOG.md - not sure this warrants a changelog update?


tools/tm-bench/transacter_test.go
```diff
@@ -0,0 +1,104 @@
...
+				"Length of expected Tx differed. tc #%d, i #%d", tcIndex, i)
+			require.Equal(t, tc.txSize*2, len(expectedHex),
+				"Length of expected hex differed. tc #%d, i #%d", tcIndex, i)
+			require.Equal(t, tc.txSize*2, len(updatedHex),
+				"Length of updated hex differed. tc #%d, i #%d", tcIndex, i)
+		}
+	}
+}
+
+func BenchmarkIterationOfSendLoop(b *testing.B) {
```
ValarDragon: this currently has code duplication because we don't yet have an implementation of a mock web socket. (As far as I've been able to find) 

The websocket connection isn't an interface sadly, so making a mock one is non-trivial as far I can discern. We have to initialize the websocket.Dialer to a net.Conn which meets the websocket requirements. (Probably not _too_ hard, but does have to do more than return nil everywhere)

melekes:
Without randomness, is there still an upper limit on how many transactions we can create without sacrificing accuracy? Maybe we should enforce it or at least document it?

zramsay:
lets add these new tests to the CI

ValarDragon:
> Without randomness, is there still an upper limit on how many transactions we can create without sacrificing accuracy? Maybe we should enforce it or at least document it?

This upper limit would be a function of the network latency, and how sensitive we are to sacrificing accuracy. Not really sure how to model network latency, or at what point we consider it to be negligible. (i.e. suppose creating the tx took half of the time the loop. That clearly skews tests,  but is the same true when it takes 10% of the time. Its still skewing, but to a lesser extent)

If we came up with a model for the websocket connections latency, then we could.

ValarDragon:
> lets add these new tests to the CI

This is currently covered in `test_cover`

ValarDragon:
Confirmed, it was tested in container #3, in `test_cover`
```
build_15225_step_105_container_3.txt
292458:=== RUN   TestGenerateTxUpdateTxConsistentency
292459:--- PASS: TestGenerateTxUpdateTxConsistentency (14.05s)
```

Issue #1955, opened 2018-07-11
 crypto/symmetric: Change algo from xsalsa20poly1305 to xchacha20poly1305 


ValarDragon:
**Feature Request**
Ref: https://github.com/tendermint/go-crypto/issues/138

Basically the title. Currently we encrypt our private keys created gaiacli via nacl's secretbox which is xsalsa20poly1305. We should switch this to our hkdfchacha20poly1305 AEAD (#135)? This also lets us remove the nacl dependency. Discussed with @zmanian, we thought it would be a good idea. This is an upgrade that can happen in a backwards compatible manner. (i.e. we can deprecate a standard, and upgrade it over the course of a release)

ValarDragon:
I think this makes more sense to switch to xchacha20poly1305, due to standardization.

ValarDragon:
Sorry I know I was advocating for doing this prelaunch earlier, but if we want to just adopt a standard for private key storage, then I think it makes sense to upgrade to the standard all at once postlaunch.

marbar3778:
@ValarDragon is this applicable in tendermint? Seems to be SDK related?

Note: they are absorbing the keys from tendermint used in the SDK.

ValarDragon:
You're right, its SDK related! And also obsolete with the usage of builtin key management.

(Used to be relevant in tendermint as we'd provide the crypto primitive here)

PR #1966, opened 2018-07-13
crypto: Refactor to move files out of the top level directory

### LINKED ISSUE #1956: crypto: Refactor to reduce number of files in top level directory
Ref https://github.com/tendermint/go-crypto/issues/136

ValarDragon:
Currently the top level directory contains basically all of the code
for the crypto package. This should be moved into sub-directories.
There are a couple of open questions, how do we want to do a symmetric
interface, and how do we want to structure the encoding. The first commit
currently names the encoding package "amino", but this has a clear
naming conflict and will probably be confusing. I propose "encoding" 
as another name for that package. (And it will also be accurate once we 
remove amino)

The commit doesn't update all of tendermint yet, hence the failing
tests.

Closes #1956

* [ ] Wrote tests
* [X] Updated CHANGELOG.md


xla:
The surface of the crupto package isn’t that big, what are the gains from the package isolation?

Could you also expand on what is meant by symetric interface?

@ValarDragon Thanks for picking this up!

ValarDragon:
> Could you also expand on what is meant by symetric interface?

Currently we just have two functions "Encrypt Symmetric" and "Decrypt Symmetric", which is basically a wrapper on an AEAD. (The particular AEAD rn is xsalsa20poly1305)

I would like to switch this xsalsa20poly1305 usage to hkdfchachapoly1305 (https://github.com/tendermint/tendermint/issues/1955), which means we should have an interface here so its easy to swap out. (Even if we don't want to use hkdfchachapoly, it should still be easy to swap out the underlying symmetric algo). 

> The surface of the crypto package isn’t that big, what are the gains from the package isolation?

Thats true. I currently don't think our crypto package is well organized. I'd prefer if we had a single interfaces file, and a file for each cryptosystem so its easy to find. I think that is necessary for code discovery. Currently if I'm working on Ed25519, I have to flip back and forth between 3-6 different files, and constantly keep scrolling back to the top to see the interface definition. (6 includes test) 

I originally made this issue when the ledger crypto code was being included in go-crypto, and at that point the number of top level directory files was at a point which I thought was too high. 

You're right, having all of these files in the top level directory wouldn't have that much of a difference at the moment. I think its just better for cryptosystems that we're playing with / doing custom stuff with to be isolated into their own package. (E.g. hkdfchachapoly) This makes a nice distinction in other code too, where you can see if you're depending on an interface, or a particular instantiation. Then if/when we do pubkey decompressing caching for ed25519, we will need a separate file for that set of keys interface. That increase in file number could be handled well by ed25519 being a package. (And since its custom, by my reasoning that warrants a subpackage)

There is a good argument for not splitting these into sub-packages though, the decoding process is more complicated now. (You have to import `tendermint/tendermint/crypto/{amino / encoding}` to just go from bytes to crypto.pubkey)

xla:
> I would like to switch this xsalsa20poly1305 usage to hkdfchachapoly1305 (#1955), which means we should have an interface here so its easy to swap out. (Even if we don't want to use hkdfchachapoly, it should still be easy to swap out the underlying symmetric algo).

Sound, as long as we assume it's the same semantics across algos.

> With that in mind, I think it would also make sense to keep ed25519 and secp in the top module. (Still using the same reorganization) Any thoughts on what would be better (moving ed25519 + secp to isolated modules, or keeping them in the top module)?

Thanks for the thorough write-up of your thoughts, great overall motivation and conclusion. A lot of your proposal reminds me of how encoding is structured in the stdlib.  We can keep the top-level for the interface(s) and the sub-packages for the concrete implementations. This might make for a clear design and nice usuability downstream. Readers disgression adviced, pseudo-code following:

``` go
// crypto top-level

// Decrypter...
type Decrypter interface {
	Decrypt([]byte) ([]byte, error)
}

// Encrypter...
type Encrypter interface {
	Encrypt([]byte) ([]byte, error)
}
```

And then in the usage side:

``` go
import (
	// ...
	
	"github.com/tendermint/tendermint/crypto/ed25519"
)

func myFunc(encrypter crypto.Encrypter) {
	// ...
}

func main() {
	myFunc(ed25519.NewEncrypter())
}
```

ValarDragon:
I was thinking of the stdlib when I made the issue! Also regarding API, I was thinking of doing the same. Do you think using a seperate "encoding" module to handle going from marshalled crypto object to pubkey (where you don't care about what type it is) makes sense to do? (The 'amino' package in the current latest commit)

xla:
If encoding is pluggable and agnostic to the implementation we could follow the stdlib example here as well and have `crypto/encoding/amino`.

> For the symmetric interface, I think that should be done in a seperate PR, just so that this is easier to review.

Full ack.

ValarDragon:
Last commit updates all of tendermint to use this. If you want to look at just the changes within crypto, look at the source tree one commit back from latest.

ValarDragon:
Tests are now passing. (Only codecov/patch is failing)

crypto/xsalsa20Symmetric/symmetric.go
```diff
@@ -3,10 +3,13 @@ package crypto
 import (
 	"errors"
 
-	. "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/crypto"
+	"github.com/tendermint/tendermint/libs/common"
```
melekes: `cmn`

ValarDragon:
I'll squash the merge commits once review is done!

cmd/priv_val_server/main.go
```diff
@@ -37,7 +37,7 @@ func main() {
 		*chainID,
 		*addr,
 		pv,
-		crypto.GenPrivKeyEd25519(),
+		ed25519.GenPrivKeyEd25519(),
```
Liamsi: suggestion: now that this is prefixed with a proper package name (`ed25519` instead of `crypto`) don't you think this should be named `ed25519.GenPrivKey()` instead?

crypto/ed25519/ed25519.go
```diff
@@ -0,0 +1,253 @@
...
+)
+
+var cdc = amino.NewCodec()
+
+func init() {
+	// NOTE: It's important that there be no conflicts here,
+	// as that would change the canonical representations,
+	// and therefore change the address.
+	// TODO: Add feature to go-amino to ensure that there
+	// are no conflicts.
```
Liamsi: This are not new TODOs, right? It's a bit odd that this shows up as new file instead of showing changes to the old one.

crypto/secp256k1/secp256k1.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+var cdc = amino.NewCodec()
+
+func init() {
+	// NOTE: It's important that there be no conflicts here,
+	// as that would change the canonical representations,
+	// and therefore change the address.
+	// TODO: Add feature to go-amino to ensure that there
+	// are no conflicts.
```
Liamsi: This comment is added everywhere. Not sure it is necessary everywhere. Also there should be an open issue in amino instead for (Add feature to go-amino to ensure that there are no conflicts).

crypto/secp256k1/secp256k1.go
```diff
@@ -0,0 +1,205 @@
...
+	priv__, _ := secp256k1.PrivKeyFromBytes(secp256k1.S256(), privKey[:])
+	sig__, err := priv__.Sign(crypto.Sha256(msg))
+	if err != nil {
+		return nil, err
+	}
+	return SignatureSecp256k1(sig__.Serialize()), nil
+}
+
+func (privKey PrivKeySecp256k1) PubKey() crypto.PubKey {
+	_, pub__ := secp256k1.PrivKeyFromBytes(secp256k1.S256(), privKey[:])
```
Liamsi: didn't you remove the double underscores in another PR?

crypto/ed25519/ed25519.go
ValarDragon: It's showing that because the file was moved + edited for comments. It is an old to do

cmd/priv_val_server/main.go
ValarDragon: Agreed

crypto/secp256k1/secp256k1.go
ValarDragon: The other pr is based on this branch

crypto/secp256k1/secp256k1.go
ValarDragon: Agreed. Do you want that done in this same refactor or? That came from moving the file

crypto/ed25519/ed25519.go
Liamsi: Thanks for the clarification.

ValarDragon:
Fixed the GenPrivKey methods. If tests pass, I'll squash the relevant commits, and this should be good to merge!

ValarDragon:
Actually not sure how to squash this. Is it fine to merge as is? `git rebase -i` with squashing all intermediate merge commits had merge conflicts

melekes:
Let's go with a merge commit then. Not biggie.

PR #1979, opened 2018-07-15
common/rand: Remove exponential distribution functions

ValarDragon:
We were computing these functions incorrectly. I'm not sure what distribution
these numbers are, but it isn't the normal definition of exponential distribution. (We're making
the probability of getting a number of a particular bitlength equal, but the number in 
that bitlength thats gets chosen is uniformly chosen)

We weren't using these functions anywhere in our codebase, and they
had a nomenclature error. (There aren't exponentially distributed
integers, instead they would be geometrically distributed) While they were
implicitly part of RandTime, and rand.Time, neither of those functions were
used anywhere.

* [ ] Updated all relevant documentation in docs - don't think these were in the docs?
* [ ] ~Updated all code comments where relevant~
* [ ] ~Wrote tests~
* [X] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 BREAKING CHANGES:
 - [genesis] removed deprecated `app_options` field.
 - [types] Genesis.AppStateJSON -> Genesis.AppState
+- [libs/common] remove exponentially distributed random numbers
 
 FEATURES:
 - [tools] Merged in from github.com/tendermint/tools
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 BREAKING CHANGES:
 - [genesis] removed deprecated `app_options` field.
 - [types] Genesis.AppStateJSON -> Genesis.AppState
+- [libs/common] remove exponentially distributed random numbers
```
melekes: needs to be moved up

PR #1980, opened 2018-07-16
Switch usage of math/rand to cmn's rand

### LINKED ISSUE #1343: all: standardize use of rand
Standardize use of rand in the codebase. 

We should exclusively use https://github.com/tendermint/tmlibs/blob/develop/common/random.go

See eg. p2p where we make various use of different rand sources.

ValarDragon:
This commit switches all usage of math/rand to cmn's rand. The only
exceptions are within the random file itself, the tools package, and the
crypto package. In tools/tm-bench you don't want it to lock between the go-routines.
The crypto package doesn't use it so the crypto package has no other
dependencies within tendermint/tendermint for easier portability.

Crypto/rand usage is unadjusted.

Closes #1343

* [ ] Updated all code comments where relevant - n/a?
* [ ] Wrote tests - n/a?
* [ ] Updated CHANGELOG.md  - not a user-facing change

### DIFF
--- a/blockchain/pool_test.go
+++ b/blockchain/pool_test.go
@@ -1,7 +1,6 @@
 package blockchain
 
 import (
-	"math/rand"
 	"testing"
 	"time"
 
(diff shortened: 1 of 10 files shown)


melekes:
Short description in the beginning helps. I think I'll be doing it too now.

PR #1987, opened 2018-07-16
Static marshaler

silasdavis:
This change makes importing ABCI types (in my case the block `Header`) into gogoproto-based files approximately 327 times more possible by making marshalling methods that can handle the use of `(gogoproto.nullable) = false` in types.proto (see https://github.com/gogo/protobuf/issues/386 for some context).

For this to work the same option must be added in the dependee proto file and then GRPC does not end up using the reflective marshalling in google/protobuf. Without this you will see errors like:

```go
panic: message/group field types.Validator:bytes without pointer [recovered]
	panic: message/group field types.Validator:bytes without pointer
```

A side effect of using these marshaller extensions is you get faster marshalling. Given you are using gogoproto here (and it introduces the issues) I can't think of a reason not to have these extensions. I've verified that with these extensions enabled I can use a type that includes `types.Header` (provided I use gogoproto - which I want to anyway).

The other think you could consider which also works is to drop all use of the nullable option and GRPC/vanilla protobuf will work.


Makefile
```diff
@@ -36,7 +36,7 @@ protoc_abci:
 	## See https://stackoverflow.com/a/25518702
 	protoc $(INCLUDE) --gogo_out=plugins=grpc:. abci/types/*.proto
 	@echo "--> adding nolint declarations to protobuf generated files"
-	@awk '/package abci/types/ { print "//nolint: gas"; print; next }1' abci/types/types.pb.go > abci/types/types.pb.go.new
```
silasdavis: Looks like this must have got merged in an non-working state since is no `package types`

Makefile
```diff
@@ -24,7 +24,7 @@ build:
 build_race:
 	CGO_ENABLED=0 go build -race $(BUILD_FLAGS) -tags '$(BUILD_TAGS)' -o build/tendermint ./cmd/tendermint
 
-install:
+install:aw
```
melekes: what's aw?

Makefile
melekes: https://github.com/tendermint/tendermint/blob/257622cf6b1edd30ef7cdb30c3d0c2cb7b951562/abci/types/types.pb.go#L47

`@awk '/package types/ { print "//nolint: gas"; print; next }1' abci/types/types.pb.go > abci/types/types.pb.go.new` should work

silasdavis: yeah that's what I changed it to, comment on wrong line

Makefile
silasdavis: um... probably 'append' 'word' oh shit I'm in the wrong window and insert mode :)

melekes:
```
abci/types/types.pb.go:4709:1:warning: sozTypes is unused (deadcode)
libs/common/types.pb.go:198:1:warning: sozTypes is unused (deadcode)
rpc/grpc/types.pb.go:367:1:warning: sozTypes is unused (deadcode)
```

looks like we'll have to change `// nolint: gas` to just `// nolint`

silasdavis:
I've added a plain `\\nolint` and I've tried to dry up the makefile with a pattern rule for all proto files

silasdavis:
Did I do this...?
```
=== RUN   TestWriter
--- FAIL: TestWriter (0.48s)
	io_test.go:147: w.Status(1)
		expected: {true 2018-07-17 13:12:29.36 +0000 UTC 500ms 100ms 100 5 200 200 200 200 0 0s 100.000%}
		got     : {true 2018-07-17 13:12:29.36 +0000 UTC 500ms 80ms 100 5 200 200 200 200 0 0s 100.000%}
```

silasdavis:
circle appears to let me trigger rerun but then does nothing

melekes:
> Did I do this...? Looks like a timing sensitive test 

no, it's not you.

silasdavis:
I can enable `gogoproto.populate` and `gogoproto.testgen` that will generate tests and probably have a positive effect on coverage if you would like to avoid damaging that metric.

melekes:
@silasdavis that would be great! let's try

silasdavis:
It would be nice if each job in circle config had a plain `make` equivalent to test this stuff locally

melekes:
> It would be nice if each job in circle config had a plain make equivalent to test this stuff locally

it does, I think. see test_* targets in Makefile.

```
test_integrations:
	make build_docker_test_image
	make get_tools
	make get_vendor_deps
	make install
	make test_cover
	make test_apps
	make test_abci_apps
	make test_abci_cli
	make test_libs
	make test_persistence
	make test_p2p
```

silasdavis:
@melekes was looking at

```yaml
  test_cover:
    <<: *defaults
    parallelism: 4
    steps:
      - attach_workspace:
          at: /tmp/workspace
      - restore_cache:
          key: v2-pkg-cache
      - restore_cache:
          key: v2-tree-{{ .Environment.CIRCLE_SHA1 }}
      - run: mkdir -p /tmp/logs
      - run:
          name: Run tests
          command: |
            for pkg in $(go list github.com/tendermint/tendermint/... | grep -v /vendor/ | circleci tests split --split-by=timings); do
              id=$(basename "$pkg")

              GOCACHE=off go test -v -timeout 5m -race -coverprofile=/tmp/workspace/profiles/$id.out -covermode=atomic "$pkg" | tee "/tmp/logs/$id-$RANDOM.log"
            done
      - persist_to_workspace:
          root: /tmp/workspace
          paths:
            - "profiles/*"
      - store_artifacts:
          path: /tmp/logs
```

And yay it passes

ValarDragon:
`go test -v -cover -race  ./...` will do that too, not sure if theres currently a command for it.
(You probably would need a higher timeout / no timeout since circle CI is using a 5 minute timeout per test)

silasdavis:
Oh right no I missed one. Not much we can do about the coverage beyond the test gen.

silasdavis:
So I took the liberty of adding ignore to your codecov.yml for protobuf files: `**/*pb.go`

And lo, the test do pass... If you want to include generated code then let me know and I will remove but seems fair enough to exclude them to me.

melekes:
LGTM

melekes:
> I've verified that with these extensions enabled I can use a type that includes types.Header (provided I use gogoproto - which I want to anyway).

and if someone does not use gogo, will these changes make their life harder?

silasdavis:
@melekes no quite the contrary - it will make their lives possible. Currently golang/protobuf will panic trying to serialise the `(gogoproto.nullable) = true` Validator (and other types). This changes adds a generated `Marshal`/`MarshalTo` method that mainline protobuf will follow (into the gogoproto library). This is actually what fixes GRPC since it imports golang/protobuf and without the static marshal methods it tries to use its own reflective marshalling that does not work for the nullable types.

On the wire these protobuf files are compatible (for example I am using them with the node.js libraries just fine) - this just deals with incompatibility between generated go types.

silasdavis:
@melekes my pleasure, nice to put something back into Tendermint.

p.s. my remark you quote is a little confusing without me giving enough context. What I mean by 'provided I use gogoproto'. Is for _my_ gogoproto enabled types (custom types, cast types etc) that include `Header` I need to enable marshal methods also - that introduces a compile time error since they expect `Header` to have them too. Come to think of it 'provided I use gogoproto' is really a misspeaking on my part - it's actually 'given I use gogoproto elsewhere - I must enable marshallers in container types'. I'm 90% sure that a straight protobuf/GRPC user would just not need to know about gogoproto (other than they would transitively need to import) and their generated code would fall through to the marhsalling methods just for `Header`. I can check this.

melekes:
> I'm 90% sure that a straight protobuf/GRPC user would just not need to know about gogoproto (other than they would transitively need to import) and their generated code would fall through to the marhsalling methods just for Header. I can check this.

please do

PR #1991, opened 2018-07-17
A link fixed

maxim-levy:
To address the relocation of terraform-and-ansible.md under networks/, as per https://github.com/tendermint/tendermint/commit/e54c0f804faa3d4a6c0c1ada0288c07d2a49ce50 

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


melekes:
@maxim-levy hi! Thank you for this patch. You don't have to create a PR for every link. You can make a few changes and open just one PR.

PR #1996, opened 2018-07-17
tools: Remove redundant grep -v vendors/

ValarDragon:
This was used in conjunction with `go list <path>`, however `go list`
already ignores the vendor directory. (It appears that this wasn't always the case, haven't been able to track down which version of go introduced this) This makes the `grep -v` redundant.

* [ ] Updated all code comments where relevant - n/a?
* [ ] Wrote tests - n/a
* [ ] Updated CHANGELOG.md - not user facing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ GOTOOLS = \
 	github.com/gogo/protobuf/protoc-gen-gogo \
 	github.com/gogo/protobuf/gogoproto \
 	github.com/square/certstrap
-PACKAGES=$(shell go list ./...')
+PACKAGES=$(shell go list ./...)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -5,7 +5,7 @@ GOTOOLS = \
 	github.com/gogo/protobuf/protoc-gen-gogo \
 	github.com/gogo/protobuf/gogoproto \
 	github.com/square/certstrap
-PACKAGES=$(shell go list ./... | grep -v '/vendor/')
+PACKAGES=$(shell go list ./...')
```
melekes: forgot to remove one `'`

Issue #1999, opened 2018-07-17
crypto: keys generated from ed25519 generate have insufficient security


ValarDragon:
Currently the ed25519 `Generate` code for deterministic key generation is the following: 
```
func (privKey PrivKeyEd25519) Generate(index int) PrivKeyEd25519 {
	bz := cdc.MustMarshalBinaryBare(struct {
		PrivKey [64]byte
		Index   int
	}{privKey, index})
	newBytes := crypto.Sha256(bz)
	newKey := new([64]byte)
	copy(newKey[:32], newBytes)
	return PrivKeyEd25519(*newKey)
}
```

Its not clear if this is our own custom scheme. It seems weird to me that the last 32 bytes of the produced new private key are nil. These last 32 bytes are definitely used within signing, and knowing that these are null bytes makes signature forgery significantly easier. (the `r` term in the ed25519 equation would be public) Haven't ran through the equation yet, but it appears to me that this definitely enables signature forgery.

cc @Liamsi

ValarDragon:
Just confirmed, this does indeed enable a trivial signature forgery attack.

ValarDragon:
I believe this enables a private key recovery attack after a single signature.

ebuchman:
Last 32 bytes are supposed to be the pubkey.

ValarDragon:
EDIT to original post: This wouldn't enable attacks after a single signature, I forgot that eddsa hashes your private key before doing anything with it. That being said, this does halve the security going into the hash function.

> Last 32 bytes are supposed to be the pubkey.

This doesn't make sense to me. The public key is public, therefore the last 32 bytes could easily be derived from the first 32 bytes / the original public key.

PR #2002, opened 2018-07-18
crypto/ed25519: Update the godocs

ValarDragon:
This commit updates the godocs for the package, and adds an optimization
to the privkey.Pubkey() method.

The optimization is that in golang, the private key (due to interface
compatibility reasons) has a copy of the public key stored inside of it.
Therefore if this copy has already been computed, there is no need to
recompute it.

* [X] Updated all relevant documentation in docs - no API changes
* [X] Wrote tests - n/a
* [X] Updated CHANGELOG.md - not user facing

### DIFF
--- a/crypto/ed25519/ed25519.go
+++ b/crypto/ed25519/ed25519.go
@@ -44,21 +44,40 @@ func init() {
 		Ed25519SignatureAminoRoute, nil)
 }
 
-// Implements crypto.PrivKey
+// PrivKeyEd25519 implements crypto.PrivKey.
 type PrivKeyEd25519 [64]byte
 
+// Bytes marshals the privkey using amino encoding.
 func (privKey PrivKeyEd25519) Bytes() []byte {
 	return cdc.MustMarshalBinaryBare(privKey)
 }
 
+// Sign produces a signature on the provided message.
 func (privKey PrivKeyEd25519) Sign(msg []byte) (crypto.Signature, error) {
 	privKeyBytes := [64]byte(privKey)
 	signatureBytes := ed25519.Sign(&privKeyBytes, msg)
 	return SignatureEd25519(*signatureBytes), nil
 }
 
+// PubKey gets the corresponding public key from the private key.
 func (privKey PrivKeyEd25519) PubKey() crypto.PubKey {
 	privKeyBytes := [64]byte(privKey)
+	initialized := false
+	// If the latter 32 bytes of the privkey are all zero, compute the pubkey
+	// otherwise privkey is initialized and we can use the cached value inside
+	// of the private key.
+	for _, v := range privKeyBytes[32:] {
+		if v != 0 {
+			initialized = true
+			break
+		}
+	}
+	if initialized {
+		var pubkeyBytes [PubKeyEd25519Size]byte
+		copy(pubkeyBytes[:], privKeyBytes[32:])
+		return PubKeyEd25519(pubkeyBytes)
+	}
+
 	pubBytes := *ed25519.MakePublicKey(&privKeyBytes)
 	return PubKeyEd25519(pubBytes)
 }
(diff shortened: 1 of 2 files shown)


crypto/ed25519/ed25519.go
```diff
@@ -117,16 +151,19 @@ func GenPrivKeyEd25519FromSecret(secret []byte) PrivKeyEd25519 {
...
+// PubKeyEd25519 implements crypto.PubKey for the Ed25519 signature scheme
 type PubKeyEd25519 [PubKeyEd25519Size]byte
 
 // Address is the SHA256-20 of the raw pubkey bytes.
 func (pubKey PubKeyEd25519) Address() crypto.Address {
 	return crypto.Address(tmhash.Sum(pubKey[:]))
 }
 
+// Bytes marshals the PubKey using amino encoding.
+// Amino encoding was chosen due its portability
```
melekes: not sure if it's the best place to talk about our reasoning

crypto/ed25519/ed25519.go
```diff
@@ -47,18 +47,37 @@ func init() {
 // Implements crypto.PrivKey
 type PrivKeyEd25519 [64]byte
 
+// Bytes marshals the privkey using amino encoding
```
Liamsi: Nitpicking: I think we want full sentences with a full stop at the end.

crypto/ed25519/ed25519.go
```diff
@@ -47,18 +47,37 @@ func init() {
...
 }
 
+// Sign produces a signature on the provided message.
 func (privKey PrivKeyEd25519) Sign(msg []byte) (crypto.Signature, error) {
 	privKeyBytes := [64]byte(privKey)
 	signatureBytes := ed25519.Sign(&privKeyBytes, msg)
 	return SignatureEd25519(*signatureBytes), nil
 }
 
+// PubKey gets the corresponding public key from the private key
```
Liamsi: Nit: full stop

crypto/ed25519/ed25519.go
```diff
@@ -73,42 +92,57 @@ func (privKey PrivKeyEd25519) Equals(other crypto.PrivKey) bool {
 	}
 }
 
-func (privKey PrivKeyEd25519) ToCurve25519() *[32]byte {
+// ToCurve25519 takes a private key and returns its representation on
+// Curve25519. Curve25519 is birationally equivalent to Edwards25519,
+// which Ed25519 uses internally. This method is for intended use in
```
Liamsi: and is internally used by Ed25519?

crypto/ed25519/ed25519.go
```diff
@@ -73,42 +92,57 @@ func (privKey PrivKeyEd25519) Equals(other crypto.PrivKey) bool {
...
+// Diffie Hellman key exchanges. (i.e. X25519)
```
Liamsi: This method is intended to be used in ...

crypto/ed25519/ed25519.go
```diff
@@ -73,42 +92,57 @@ func (privKey PrivKeyEd25519) Equals(other crypto.PrivKey) bool {
...
+func (privKey PrivKeyEd25519) ToCurve25519() *[PubKeyEd25519Size]byte {
 	keyCurve25519 := new([32]byte)
 	privKeyBytes := [64]byte(privKey)
 	extra25519.PrivateKeyToCurve25519(keyCurve25519, &privKeyBytes)
 	return keyCurve25519
 }
 
-// Deterministically generates new priv-key bytes from key.
+// Generate deterministically generates a new priv-key bytes from key.
```
Liamsi: s/generates/derives (?)

crypto/ed25519/ed25519.go
```diff
@@ -73,42 +92,57 @@ func (privKey PrivKeyEd25519) Equals(other crypto.PrivKey) bool {
...
+// The privkey is generated as Sha256(amino_encode({privkey, index}))
+// Note that we are using the non-standard golang definition for ed25519
+// privkeys that includes the pubkey appended to it
```
Liamsi: "Note that similar to golang/x/crypto/ed25519 we include the private key ..."
Maybe, even add a link to: https://github.com/golang/crypto/blob/a49355c7e3f8fe157a85be2f77e6e269a0f89602/ed25519/ed25519.go#L48-L52 ?

The way we derive the key here is non-standard anyways?

crypto/ed25519/ed25519.go
```diff
@@ -73,42 +92,57 @@ func (privKey PrivKeyEd25519) Equals(other crypto.PrivKey) bool {
...
 		Index   int
 	}{privKey, index})
-	if err != nil {
-		panic(err)
-	}
 	newBytes := crypto.Sha256(bz)
 	newKey := new([64]byte)
 	copy(newKey[:32], newBytes)
+	// Caches the pubkey in the latter 32 bytes of the privkey
+	// (this is a mutative function)
```
Liamsi: does this clarify / help? It does not mutate the underlying `privKey`? (What is a "mutative function" and why is it relevant in that context?) I'd propose to remove this.

crypto/ed25519/ed25519.go
Liamsi: I agree.

crypto/ed25519/ed25519.go
ValarDragon: This does mutate the private key object. (I found this quite confusing tbh, it is what caused #1999)

Perhaps changing it to 
" this mutates the last 32 bytes of the privkey" would be more clear?

Liamsi: It is confusing indeed. But wait, it does not mutate the private on which it is called on (which is represented by the var `privKey`, right?), it does indeed "mutate" the freshly created private-key though which is called `newKey`. Not sure if "mutative" makes sense in this context then (because of course the method "mutates" the object which it creates). Or am I getting this wrong?

ValarDragon: `ed25519.MakePublicKey(newKey)` alters the last 32 bytes of `newKey`

ValarDragon:
It was the right base, since thats where I refactored into an ed25519 file. I squashed that branch, which is whats causing all of the merge conflicts here. I don't really know how to fix this tho. Should I open a new PR on develop, once `dev/refactor_crypto` is merged?

Liamsi:
> Should I open a new PR on develop that cherry-picks these commits, once dev/refactor_crypto is merged?

I think you can either rebase this branch on (origin) dev/refactor_crypto or merge dev/refactor_crypto into this branch. In both cases you'd need to resolve the merge conflicts above.

ValarDragon:
I'm amazed this worked, but I ran `git rebase -i HEAD~10 --onto dev/refactor_crypto`, and dropped all commits except the ones I wanted, and squashed those together into one.

PR #2004, opened 2018-07-18
Compatible old version

needkane:
Now update version must change code： https://github.com/tendermint/tendermint/commit/2fa99628351bd6d4ba14226bb7336135d655ff40.

There is a problem, if  already run old version nodes，and many user use them everyday. I can't upgrade one node which use new version code. 

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -617,11 +617,6 @@ func (sw *Switch) addPeer(pc peerConn) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


needkane:
Why

ebuchman:
See some notes on how we're thinking about this here https://github.com/tendermint/tendermint/pull/1983

PR #2011, opened 2018-07-19
compatible old version

needkane:
Now update version must change code： https://github.com/tendermint/tendermint/commit/2fa99628351bd6d4ba14226bb7336135d655ff40.

There is a problem, if  already run old version(major  version different) nodes，and many user use them everyday. I can't upgrade one node which use new version code. 


melekes:
This is related to https://github.com/tendermint/tendermint/pull/1983

PR #2012, opened 2018-07-19
Update version.go

needkane:
Avoid change version should modify two places

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


melekes:
It must be a string because we are using it to detect build version in our scripts https://github.com/tendermint/tendermint/blob/40342bfa4af115d7f06605225583a990b5771624/scripts/dist.sh#L9.

PR #2016, opened 2018-07-20
crypto/ed25519: Remove privkey.Generate method

### LINKED ISSUE #2000: crypto: remove pubkey from hash inside of ed25519 generate
This is in regards to the generate function for determinstic ed25519 privkey generation. 

The latter 32 bytes of the privkey are just the cached pubkey. Currently we're hashing the entire privkey, which includes the pubkey being hashed. This seems unneccessary. The canonical form of the private keys is just these first 32 bytes. (Therefore if we use our current method, it may not fit with other language implementations) I believe that the pubkey is cached in the private key in order to make it fit the relevant API. 

/cc @Liamsi , @ebuchman

ValarDragon:
The privkey.Generate method here was a custom-made method for deriving
a private key from another private key. This function is currently
not used anywhere in our codebase, and has not been reviewed enough
that it would be secure to use. This removes that method. We should
adopt the official ed25519 HD derivation once that has been standardized,
in order to fulfill this need.

closes #2000

* [X] Wrote tests - n/a
* [ ] Updated CHANGELOG.md - not sure if I should. While this is an API change, it wasn't used anywhere.

### DIFF
--- a/crypto/ed25519/ed25519.go
+++ b/crypto/ed25519/ed25519.go
@@ -103,27 +103,6 @@ func (privKey PrivKeyEd25519) ToCurve25519() *[PubKeyEd25519Size]byte {
 	return keyCurve25519
 }
 
-// Generate deterministically derives a new priv-key bytes from key.
-// The privkey is generated as Sha256(amino_encode({privkey, index}))
-// Note that we append the public key to the private key, the same way
-// that golang/x/crypto/ed25519 does. See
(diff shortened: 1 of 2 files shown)


ValarDragon:
As far as I can tell, there is no referrence to this in the documentation. (Determined by looking at all instances of `generate` case agnostic, in the docs)

PR #2017, opened 2018-07-20
crypto/secp256k1: Add godocs, remove indirection in privkeys

ValarDragon:
The following was previously done for creating secp256k1 private keys:

First obtain privkey bytes. Then create a private key in the
underlying library, with scalar exponent equal to privKeyBytes.
(The method called was secp256k1.PrivKeyFromBytes,
https://github.com/btcsuite/btcd/blob/fb90c334dffc6c91843d8d36ac31e7f7eb6c7594/btcec/privkey.go#L21)

Then the private key was serialized using the underlying library, which just
returns back the bytes that comprised the scalar exponent, but padded to be
exactly 32 bytes.
https://github.com/btcsuite/btcd/blob/fb90c334dffc6c91843d8d36ac31e7f7eb6c7594/btcec/privkey.go#L70

Thus the entire indirection of calling the underlying library can be avoided
by just ensuring that we pass in a 32 byte value. This also avoids doing a more
expensive public key creation. (Point scalar multiplication on the generator)

A test case has even be written to show the correctness more clearly in review.

Note I have reviewed the source myself, to ensure nothing is going on. When the input array is not 32 bytes, it left pads with 0 bytes for it to become 32 bytes long, or it truncates from the right if the input was too long.

* [X] Updated all relevant documentation in docs - basically a godoc update
* [X] Wrote tests
* [x] Updated CHANGELOG.md - not a user facing change


crypto/secp256k1/secp256k1.go
```diff
@@ -73,54 +73,46 @@ func (privKey PrivKeySecp256k1) PubKey() crypto.PubKey {
...
-	} else {
-		return false
 	}
+	return false
 }
 
-/*
-// Deterministically generates new priv-key bytes from key.
-func (key PrivKeySecp256k1) Generate(index int) PrivKeySecp256k1 {
-	newBytes := cdc.BinarySha256(struct {
```
ValarDragon: This was removed as its purpose is superceded by HD derivation code.

crypto/secp256k1/secp256k1.go
```diff
@@ -73,54 +73,46 @@ func (privKey PrivKeySecp256k1) PubKey() crypto.PubKey {
...
+// that 32 byte output to create the private key.
 // NOTE: secret should be the output of a KDF like bcrypt,
 // if it's derived from user input.
 func GenPrivKeySecp256k1FromSecret(secret []byte) PrivKeySecp256k1 {
-	privKey32 := crypto.Sha256(secret) // Not Ripemd160 because we want 32 bytes.
-	priv, _ := secp256k1.PrivKeyFromBytes(secp256k1.S256(), privKey32)
-	privKeyBytes := [32]byte{}
-	copy(privKeyBytes[:], priv.Serialize())
-	return PrivKeySecp256k1(privKeyBytes)
+	privKey32 := sha256.Sum256(secret)
```
ValarDragon: Uses the default sha256.Sum256, since that provides a fixed length output. (Not a slice)

crypto/secp256k1/secp256k1.go
```diff
@@ -139,22 +136,22 @@ func (pubKey PubKeySecp256k1) Bytes() []byte {
 	return bz
 }
 
-func (pubKey PubKeySecp256k1) VerifyBytes(msg []byte, sig_ crypto.Signature) bool {
+func (pubKey PubKeySecp256k1) VerifyBytes(msg []byte, _sig crypto.Signature) bool {
```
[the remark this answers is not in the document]
ValarDragon: A single leading underscore is allowed user the rules of golint. (It has to do with checking if the var is used)

I'll try to remove the underscore entirely though.

crypto/secp256k1/secp256k1.go
```diff
@@ -73,54 +77,46 @@ func (privKey PrivKeySecp256k1) PubKey() crypto.PubKey {
...
+// PubKeySecp256k1Size is comprised of 32 bytes for one field element
+// (the x-coordinate), plus one byte for the parity of the y-coordinate.
 const PubKeySecp256k1Size = 33
 
-// Implements crypto.PubKey.
-// Compressed pubkey (just the x-cord),
-// prefixed with 0x02 or 0x03, depending on the y-cord.
+// PubKeySecp256k1 implements crypto.PubKey.
+// Compressed pubkey (just the x-coordinate),
+// prefixed with 0x02 or 0x03, depending on the y-coordinate.
```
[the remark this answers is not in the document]
ValarDragon: Good point

crypto/secp256k1/secpk256k1_test.go
```diff
@@ -63,3 +65,15 @@ func TestSignAndValidateSecp256k1(t *testing.T) {
 
 	assert.False(t, pubKey.VerifyBytes(msg, sig))
 }
+
+func TestSecp256k1LoadPrivkeyAndSerializeIsIdentity(t *testing.T) {
```
Liamsi: This test is cool to show the reviewer that what you are doing is OK. Not sure, if in two years from now people will wonder why it is included. Maybe add a comment why this is useful.

Also, did you experiment with an edge case: What if the random bytes already contain zeroes? I think the padding was partly to make sure that leading zeroes in `pk` be "ignored" by `new(big.Int).SetBytes(pk)` in https://github.com/btcsuite/btcd/blob/fb90c334dffc6c91843d8d36ac31e7f7eb6c7594/btcec/privkey.go#L21-L31
This might be even less likely when using the output of sha256. But it would be great that this didn't change with this PR.

Liamsi: Clarified this with @ValarDragon on slack. This isn't an issue here, as `Serialize` is still called.

ValarDragon:
I think I addressed the concerns pointed out in your review + what was in the other crypto refactor PR. If it looks good, we can squash this and then merge.

PR #2018, opened 2018-07-20
abci/types:Make response can be read by humans

MrXJC:
For solving the Human Unreadable Bug in CosmosSDK [issue#1604](https://github.com/cosmos/cosmos-sdk/issues/1604),  I think this bug should be fixed in Tendermint. 
So I  add a function to the `Response`  and make `Response`  readable.

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md


melekes:
It's not a very good idea to change a file, which is autogenerated. This will require us to copy-paste these functions every time protobuf changes.

MrXJC:
Thank you very much. Maybe I can do it in CosmoSDK by go-refect!

Issue #1920, opened 2018-07-06
Transactions are stuck in mempool and not getting propagated to other nodes


yuomii:
BUG REPORT

**Tendermint version** 
0.22.0-fa05b039

**ABCI app**
Custom Internal

**Environment**:
- Ubuntu 64-bit
- 4 Validator Nodes
- Tx size ~ 80KB

**What happened**:
After a not deterministic time period transactions which were successfully validated by CheckTx are not getting gossiped or deleted from mempool and the node cant be used anymore.
**What you expected to happen**:
After a tx is validated by CheckTx it should be gossiped to other nodes and the proposer node calls BeginBlock.
**How to reproduce it**:
We can reproduce this consistently without a deterministic time period with the setup described above 

**Config**:
empty blocks=false
p2p-params as described in "running in production"

**`/dump_consensus_state` output for consensus bugs**
consensus node 1:
https://gist.github.com/yuomii/5acbd97d2935b7d1dd61d98e2ce5adb8
consensus node 2:
https://gist.github.com/yuomii/ab11b2914753080b39cd3923b07f5d6d
consensus node 3:
https://gist.github.com/yuomii/23f74831f9031b213186c00d83fd35eb
consensus node 4:
https://gist.github.com/yuomii/215862af5e899092987b41852e13018c

**Logs**:
Full Logs (log level debug):
node 1:
https://gist.github.com/yuomii/918ed5271a909613b37649f1b8d152f9
node 2:
https://gist.github.com/yuomii/d80a783cdb088039e1cc78689917c697
node 3:
https://gist.github.com/yuomii/96b0e482d697d46e27648d3e73c36ae8
node 4:
https://gist.github.com/yuomii/2082b674c996386499144c819785a164

Seems to be related to #1875 

EDIT:
dump_consensus was called 2h later. After ~2h we noticed that blocks were beeing build again.

ebuchman:
> We can reproduce this consistently without a deterministic time period with the setup described above

Can you clarify with some details around how many txs you are sending and to which nodes and at what rate?

ebuchman:
Also it seems some of the logs are out of order? Height and time flow backwards at points. Is this raw output from Tendermint?

yuomii:
[...]

PR #2021, opened 2018-07-20
#1920 try to fix race condition on proposal height for published txs

srmo:
Potential fix for https://github.com/tendermint/tendermint/issues/1920

My observation is, that there's a race condition between the actual mempool height used in the propose step and the height used to publish new transactions via `notifyTxsAvailable`

It's possible that `notifyTxsAvailable` publishes a tx with a height, that is already finalized (via votes/short before commiting/`Update()` - thought that's where my understanding ends).

If this happens, the receiveRoutine will drain the tx from the channel and and then call `handleTxsAvailable` which in turn triggers propose. Now the height is just wrong and propose will throw a `enterPropose(229/0): Invalid args. Current step: 230/0/RoundStepNewHeight module=consensus height=229 round=0` and the tx is now stuck in the mempool.
Interestingly, additional broadcasts also do not get published.
Probably because the channel is now full and we hit the `default` branch as the `case` fails to publish.

Calling `handleTx` with the next height (current mempool height + 1) fixes the problem.
I've exposed mempool.height via a new method `Height()` on mempool and added a defensive lock around it - I assume that one has think about remaining race conditions surrounding that.

The root cause analysis is hard for me.
I'm happy to discuss this here / on the issue / ...

- related to create_empty_blocks=false
- published height for accepted tx can be wrong (too low)
- use the actual mempool height + 1 for the proposal
- expose Height() on mempool

* [ ] Wrote tests
* [x] Updated CHANGELOG.md


srmo:
and no idea how to increase the MempoolMock coverage

consensus/state.go
```diff
@@ -572,7 +572,9 @@ func (cs *ConsensusState) receiveRoutine(maxSteps int) {
 
 		select {
 		case height := <-cs.mempool.TxsAvailable():
-			cs.handleTxsAvailable(height)
+			actualHeight := cs.mempool.Height() + 1
```
melekes: can't we use `cs.Height` here?

melekes: I meant `cs.Height + 1`, sorry

srmo:
Using `cs.Height + 1` didn't work at all. I'm a bit stumped by that but using just `cs.Height` seems to work, at least in a short one-node run. I'll keep the node running over night.

consensus/state.go
```diff
@@ -571,8 +571,11 @@ func (cs *ConsensusState) receiveRoutine(maxSteps int) {
 		var mi msgInfo
 
 		select {
-		case height := <-cs.mempool.TxsAvailable():
-			cs.handleTxsAvailable(height)
+		case txAvailable := <-cs.mempool.TxsAvailable():
+			// TODO why don't we have a "default" branch for this drain? Similar to the publish, wouldn't this lock otherwise?
```
melekes: this won't lock because we have other cases below

consensus/state.go
```diff
@@ -683,11 +686,13 @@ func (cs *ConsensusState) handleTimeout(ti timeoutInfo, rs cstypes.RoundState) {
 
 }
 
-func (cs *ConsensusState) handleTxsAvailable(height int64) {
+func (cs *ConsensusState) handleTxsAvailable() {
 	cs.mtx.Lock()
 	defer cs.mtx.Unlock()
 	// we only need to do this for round 0
-	cs.enterPropose(height, 0)
+	heightToPropose := cs.Height
```
melekes: no need to create a separate variable

srmo: Idea here was being paranoid, as there's still time passing between the call to cs.Height in the logging statement and the actual passing. (and I'm coming from Java where we usually don't access fields but methods, so multiple calls should be consolidated into one variable.
Will remove this.

consensus/state.go
srmo: Ah, OK, understood.

srmo:
I don't know if I need to write an explicit test for this. Any hints or can I tick the checkbox?

PR #2029, opened 2018-07-23
Release/0.22.5

melekes:
CHANGELOG: https://github.com/tendermint/tendermint/blob/948b91e62ec2d46825fde843d0403e4fcc4510fb/CHANGELOG.md#0225


libs/common/random.go
```diff
@@ -175,146 +166,95 @@ MAIN_LOOP:
 	return string(chars)
 }
 
-// It is not safe for cryptographic usage.
```
ebuchman: We should air on the side of caution with warnings like this. Better to keep them than remove!

p2p/switch.go
```diff
@@ -281,8 +281,11 @@ func (sw *Switch) StopPeerForError(peer Peer, reason interface{}) {
 	sw.stopAndRemovePeer(peer, reason)
 
 	if peer.IsPersistent() {
-		// NOTE: this is the self-reported addr, not the original we dialed
-		go sw.reconnectToPeer(peer.NodeInfo().NetAddress())
+		addr := peer.OriginalAddr()
+		if addr == nil {
+			panic(fmt.Sprintf("persistent peer %v with no original address", peer))
```
ebuchman: is it documented somewhere that only outbound peers can be persistent on our end?

rpc/core/doc.go
```diff
@@ -39,8 +39,6 @@ curl 'localhost:26657/broadcast_tx_sync?tx="abc"'
 }
 ```
 
-The first entry in the result-array (`96`) is the method this response correlates with. `96` refers to "ResultTypeBroadcastTx", see [responses.go](https://github.com/tendermint/tendermint/blob/master/rpc/core/types/responses.go) for a complete overview.
```
ebuchman: omg thank you

state/execution.go
```diff
@@ -86,7 +86,7 @@ func (blockExec *BlockExecutor) ApplyBlock(state State, blockID types.BlockID, b
 	fail.Fail() // XXX
 
 	// update the state with the block and responses
-	state, err = updateState(state, blockID, block.Header, abciResponses)
+	state, err = updateState(state, blockID, &block.Header, abciResponses)
```
ebuchman: this one needs a pointer?

libs/common/random.go
melekes: They were moved to the `Rand` struct documentation (not removed). https://github.com/tendermint/tendermint/blob/948b91e62ec2d46825fde843d0403e4fcc4510fb/libs/common/random.go#L14-L20

p2p/switch.go
melekes: No. Actually, should we try and dial self-reported addr if persistent peer is "inbound"?

state/execution.go
melekes: I did not want to remove all the pointers (can result in poor performance, no?)

p2p/switch.go
melekes: https://github.com/tendermint/tendermint/pull/2031

rpc/core/doc.go
zramsay: :trollface:

crypto/priv_key.go
```diff
@@ -1,164 +0,0 @@
...
-		panic(err)
-	}
-	newBytes := Sha256(bz)
-	newKey := new([64]byte)
-	copy(newKey[:32], newBytes)
-	ed25519.MakePublicKey(newKey)
-	return PrivKeyEd25519(*newKey)
-}
-
-func GenPrivKeyEd25519() PrivKeyEd25519 {
```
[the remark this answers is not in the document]
ValarDragon: If the only concern is the sdk, I can go ahead and update it. 

We _can't_ deprecate this method, since that would cause import cycles.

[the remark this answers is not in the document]
ValarDragon: Updated in https://github.com/cosmos/cosmos-sdk/pull/1798, depends on #2034  though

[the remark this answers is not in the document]
ebuchman: ok merged #2034 into this. thanks!

PR #2031, opened 2018-07-23
reconnect to self-reported address if persistent peer is inbound

melekes:


p2p/switch.go
```diff
@@ -283,7 +283,8 @@ func (sw *Switch) StopPeerForError(peer Peer, reason interface{}) {
 	if peer.IsPersistent() {
```
ebuchman: from what I can tell, this will only be true for an outbound peer. I think that's probably fine, but its cool to have this change instead of panicing just in case it changes. Worth adding a comment about this though!

PR #2054, opened 2018-07-24
secret connection update

### LINKED ISSUE #2039: Secret connection plan
This issue is meant to summarize / consolidate the secret connection conversation. (Its currently across 3 different locations) This is per discussion in the tendermint meeting today.

The current issue in secret connection is that we are relying on the salsa primitive, which has basically been abandoned in favor of chacha in many applications, such as TLS. We're also relying on the RipeMD160 primitive as a sub-component of nacl secret box. 

The proposed solution is the following: 

In the handshake, use a single HKDF-Sha2 invocation on the shared DH secret, local eph pubkey and remote eph pubkey to generate the secret key, send nonce, recv nonce, and challenges. 

Use chacha20poly1305 as a stream cipher, via incrementing the send and recv nonces. 

The only remaining question I have is that is there a need for send nonce or recv nonce to be derived from hkdf? Could it instead be that one of the two participants has their recv nonce initially as 1, and the other has their send nonce initially as 2**96 -1 and decrements their nonce. The reason I'm wondering that is that that would minimize the chance of collisions (since we're using the nonce as a stream here, and not one of the randomized nonce variants)

/cc @zmanian @tarcieri

ValarDragon:
This PR updates secret connection to remove the ripemd primitive, and the salsa primitive. This gives us a resulting secret connection that looks alot more `noise` like. 

This switches the sender and receiver to use different keys, similar to what noise does. We now use a single hkdf-sha2 invocation on the diffie hellman secret to generate the two keys and a challenge. 

Closes #2039 

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/docs/spec/p2p/peer.md
+++ b/docs/spec/p2p/peer.md
@@ -27,27 +27,24 @@ Both handshakes have configurable timeouts (they should complete quickly).
 ### Authenticated Encryption Handshake
 
 Tendermint implements the Station-to-Station protocol
-using ED25519 keys for Diffie-Helman key-exchange and NACL SecretBox for encryption.
+using X25519 keys for Diffie-Helman key-exchange and chacha20poly1305 for encryption.
 It goes as follows:
-- generate an emphemeral ED25519 keypair
+- generate an ephemeral X25519 keypair
 - send the ephemeral public key to the peer
 - wait to receive the peer's ephemeral public key
 - compute the Diffie-Hellman shared secret using the peers ephemeral public key and our ephemeral private key
-- generate two nonces to use for encryption (sending and receiving) as follows:
-    - sort the ephemeral public keys in ascending order and concatenate them
-    - RIPEMD160 the result
-    - append 4 empty bytes (extending the hash to 24-bytes)
-    - the result is nonce1
-    - flip the last bit of nonce1 to get nonce2
-    - if we had the smaller ephemeral pubkey, use nonce1 for receiving, nonce2 for sending;
-        else the opposite
-- all communications from now on are encrypted using the shared secret and the nonces, where each nonce
-increments by 2 every time it is used
+- generate two keys to use for encryption (sending and receiving) and a challenge for authentication as follows:
+    - create a hkdf-sha256 instance with the key being the diffie hellman shared secret, and info parameter as
+      `TENDERMINT_SECRET_CONNECTION_KEY_AND_CHALLENGE_GEN`
+    - get 96 bytes of output from hkdf-sha256
+    - if we had the smaller ephemeral pubkey, use the first 32 bytes for the key for receiving, the second 32 bytes for sending; else the opposite
+    - use the last 32 bytes of output for the challenge
+- use a seperate nonce for receiving and sending. Both nonces start at 0, and should support the full 96 bit nonce range
+- all communications from now on are encrypted in 1024 byte frames,
+using the respective secret and nonce. Each nonce is incremented by one after each use. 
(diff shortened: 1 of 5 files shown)


docs/spec/p2p/peer.md
```diff
@@ -27,27 +27,24 @@ Both handshakes have configurable timeouts (they should complete quickly).
 ### Authenticated Encryption Handshake
 
 Tendermint implements the Station-to-Station protocol
-using ED25519 keys for Diffie-Helman key-exchange and NACL SecretBox for encryption.
+using Ed25519 keys for Diffie-Helman key-exchange and chacha20poly1305 for encryption.
```
zmanian: The Montgomery keys are typically referred to as x25519

docs/spec/p2p/peer.md
```diff
@@ -27,27 +27,24 @@ Both handshakes have configurable timeouts (they should complete quickly).
...
 It goes as follows:
-- generate an emphemeral ED25519 keypair
+- generate an emphemeral Ed25519 keypair
```
zmanian: X25519

docs/spec/p2p/peer.md
ValarDragon: Good catch! Thanks

p2p/conn/secret_connection.go
```diff
@@ -230,9 +237,33 @@ func shareEphPubKey(conn io.ReadWriteCloser, locEphPub *[32]byte) (remEphPub *[3
...
-	box.Precompute(shrSecret, remPubKey, locPrivKey)
+func deriveSecretAndChallenge(dhSecret *[32]byte, locIsLeast bool) (recvSecret, sendSecret *[aeadKeySize]byte, challenge *[32]byte) {
+	hash := sha256.New
+	hkdf := hkdf.New(hash, dhSecret[:], nil, []byte("TENDERMINT_SECRET_CONNECTION_KEY_AND_CHALLENGE_GEN"))
+	// get enough date for 2 aead keys, and a 32 byte challenge
+	res := new([96]byte)
+	io.ReadFull(hkdf, res[:])
+
+	challenge = new([32]byte)
+	recvSecret = new([32]byte)
```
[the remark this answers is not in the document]
ValarDragon: My concern with aeadKeySize in this function was that it made something relatively simple suddenly seem pretty complicated. (I changed it here, and changed it back)

p2p/conn/secret_connection.go
```diff
@@ -230,9 +237,37 @@ func shareEphPubKey(conn io.ReadWriteCloser, locEphPub *[32]byte) (remEphPub *[3
...
 
-func computeSharedSecret(remPubKey, locPrivKey *[32]byte) (shrSecret *[32]byte) {
-	shrSecret = new([32]byte)
-	box.Precompute(shrSecret, remPubKey, locPrivKey)
+func deriveSecretAndChallenge(dhSecret *[32]byte, locIsLeast bool) (recvSecret, sendSecret *[aeadKeySize]byte, challenge *[32]byte) {
+	hash := sha256.New
+	hkdf := hkdf.New(hash, dhSecret[:], nil, []byte("TENDERMINT_SECRET_CONNECTION_KEY_AND_CHALLENGE_GEN"))
+	// get enough data for 2 aead keys, and a 32 byte challenge
+	res := new([2*aeadKeySize + 32]byte)
+	io.ReadFull(hkdf, res[:])
```
Liamsi: Do not skip the error here, or, explain why we can assume that no error can happen:
https://golang.org/pkg/io/#ReadFull

Issue #2053, opened 2018-07-24
v0.22.5 is broken due to amino v0.11.1  


ebuchman:
v0.22.5 doesn't work (ie. `tendermint init && tendermint node --proxy_app=dummy` will panic immediately.

It seems to be due to the upgrade from v0.10.1 to v0.11.1 of amino.

ValarDragon:
These amino errors showed up in my secret connection PR basically all tests failed, until I downgraded to 0.10.1. So I guess circle did catch these on new PR's?

ebuchman:
So, it turns out I was using `dep ensure` rather than `make get_vendor_deps`. Using the latter, it actually works, the difference being the latter doesnt update Gopkg.lock, just uses the existing one. But I don't understand why amino v0.11.1 works if we do `make get_vendor_deps` but only v0.10.1 if we do `dep ensure` !

ebuchman:
So it turns out this is only half true. Using `make get_vendor_deps`, it works fine. The problem is the toml updated for v0.11, but the lock file didn't, and the make command ignores the toml.

ebuchman:
Not a real issue for folks using `make get_vendor_deps` and fixed in #2056 anyways

ebuchman:
Tagging @Liamsi @jaekwon to take note that apparently some work will be required to upgrade Tendermint to Amino v0.11

PR #2055, opened 2018-07-25
makefile: Add `make check_dep` and remove `make ensure_deps`

ValarDragon:
This adds a new makefile command, which is used in CI linting, `make check_dep`.
This ensures the toml is in sync with the lock, and that were not pinning to a
branch in any repository.

This also adapts `make get_vendor_deps` to check the lock, in addition to
populating the vendor directory. This removes the need for `make ensure_deps`.

This makes `make get_vendor_deps` consistent between tendermint and the sdk.

This would stop problems like #2053 from happening. 

* [ ] Updated all relevant documentation in docs - anything I should be updating
* [x] Updated all code comments where relevant - i think so
* [x] Wrote tests - n/a
* [x] Updated CHANGELOG.md

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,6 +77,12 @@ jobs:
             set -ex
             export PATH="$GOBIN:$PATH"
             make metalinter
+      - run:
+          name: check_dep
+          command: |
+            set -ex
+            export PATH="$GOBIN:$PATH"
+            make check_dep
 
   test_abci_apps:
     <<: *defaults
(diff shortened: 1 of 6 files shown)


ValarDragon:
I'm not sure why this is failing `test_p2p`. Its failing when running `make get_vendor_deps` in the docker file, but I don't see why this would fail, since it should have had the toml and lock.

melekes:
```
--> Running dep
no dirs contained any Go code
```

ValarDragon:
ohh that makes sense. I've just re-ordered the lines for copying in the files in the dockerfile! That fixes it.

ValarDragon:
Presuming tests pass, this should be fine to merge. (Could this be squashed via github)

PR #2058, opened 2018-07-25
crypto: Switch hkdfchacha back to xchacha

ValarDragon:
hkdfchachapoly was a construction we came up with. There is no longer any
reason to use it. We should instead just use xchacha for the remaining use
cases we have. (Symmetrically encrypting the keys in the sdk keys package)

This basically restores zaki's old code, but also includes testing randomized cases.

* [X] Wrote tests
* [x] Updated CHANGELOG.md


crypto/xchacha20poly1305/xchachapoly.go
```diff
@@ -0,0 +1,256 @@
...
+func (c *xchacha20poly1305) Overhead() int {
+	return TagSize
+}
+
+func (c *xchacha20poly1305) Seal(dst, nonce, plaintext, additionalData []byte) []byte {
+	if len(nonce) != NonceSize {
+		panic("xchacha20poly1305: bad nonce length passed to Seal")
+	}
+
+	if uint64(len(plaintext)) > (1<<38)-64 {
```
melekes: MaxPlaintextSize

crypto/xchacha20poly1305/xchachapoly.go
```diff
@@ -0,0 +1,256 @@
...
+	copy(subNonce[4:], nonce[16:])
+
+	return chacha20poly1305.Seal(dst, subNonce[:], plaintext, additionalData)
+}
+
+func (c *xchacha20poly1305) Open(dst, nonce, ciphertext, additionalData []byte) ([]byte, error) {
+	if len(nonce) != NonceSize {
+		return nil, fmt.Errorf("xchacha20poly1305: bad nonce length passed to Open")
+	}
+	if uint64(len(ciphertext)) > (1<<38)-48 {
```
melekes: MaxCiphertextSize

crypto/xchacha20poly1305/xchachapoly.go
```diff
@@ -0,0 +1,256 @@
...
+	HChaCha20(&subKey, &hNonce, &c.key)
+
+	chacha20poly1305, _ := chacha20poly1305.New(subKey[:])
+
+	copy(subNonce[4:], nonce[16:])
+
+	return chacha20poly1305.Open(dst, subNonce[:], ciphertext, additionalData)
+}
+
+// TODO: Ask zaki what codebase the below is due to.
```
Liamsi: Looks like this is based on this code base?
https://github.com/aead/chacha20poly1305
https://github.com/aead/chacha20poly1305/blob/master/LICENSE

ValarDragon:
Sorry for not providing the diff. I'll definitely include one in the future!

crypto/xchacha20poly1305/xchachapoly_test.go
```diff
@@ -57,7 +15,7 @@ func TestVector(t *testing.T) {
 // license that can be found at the bottom of this file.
 func TestRandom(t *testing.T) {
 	// Some random tests to verify Open(Seal) == Plaintext
-	for i := 0; i < 256; i++ {
+	for i := 0; i < 1024; i++ {
```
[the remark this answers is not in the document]
ValarDragon: It was just that 256 felt insufficiently large, and the time delay wasn't significant (0.08 seconds) However upon further thought, doing more than just a couple random tests isn't that useful. So I'll revert this back to 256. Thanks!

PR #2061, opened 2018-07-25
crypto: Add benchmarking code for signature schemes

### LINKED ISSUE #1984: crypto: Add Benchmarks for all signature schemes
We should have benchmarks in tendermint/tendermint/crypto, just so that we can easily compare speeds.

ValarDragon:
This does a slight refactor for the key generation code. It now calls a
seperate unexported method to allow generation from a reader. I think this
will actually reduce time in generation, due to no longer initializing an
extra slice. This was needed in order to enable benchmarking.

This uses an internal package for the benchmarking code, so that this can
be standardized without being exported in the public API. The benchmarking
code is derived from agl/ed25519's benchmarking code, and has copied the
license over.

Closes #1984

* [ ] Updated all relevant documentation in docs - unsure if anything should be updated
* [x] Wrote tests
* [x] Updated CHANGELOG.md - no user facing change

### DIFF
--- a/crypto/secp256k1/secp256k1.go
+++ b/crypto/secp256k1/secp256k1.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha256"
 	"crypto/subtle"
 	"fmt"
+	"io"
 
 	secp256k1 "github.com/btcsuite/btcd/btcec"
 	amino "github.com/tendermint/go-amino"
@@ -80,8 +81,13 @@ func (privKey PrivKeySecp256k1) Equals(other crypto.PrivKey) bool {
 // It uses OS randomness in conjunction with the current global random seed
 // in tendermint/libs/common to generate the private key.
 func GenPrivKey() PrivKeySecp256k1 {
+	return genPrivKey(crypto.CReader())
+}
+
+// genPrivKey generates a new secp256k1 private key using the provided reader.
+func genPrivKey(reader io.Reader) PrivKeySecp256k1 {
 	privKeyBytes := [32]byte{}
-	copy(privKeyBytes[:], crypto.CRandBytes(32))
+	io.ReadFull(reader, privKeyBytes[:])
 	// crypto.CRandBytes is guaranteed to be 32 bytes long, so it can be
 	// casted to PrivKeySecp256k1.
 	return PrivKeySecp256k1(privKeyBytes)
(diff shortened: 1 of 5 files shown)


crypto/secp256k1/secp256k1.go
```diff
@@ -80,8 +81,13 @@ func (privKey PrivKeySecp256k1) Equals(other crypto.PrivKey) bool {
 // It uses OS randomness in conjunction with the current global random seed
 // in tendermint/libs/common to generate the private key.
 func GenPrivKey() PrivKeySecp256k1 {
+	return genPrivKey(crypto.CReader())
+}
+
+// genPrivKey generates a new secp256k1 private key using the provided reader.
+func genPrivKey(reader io.Reader) PrivKeySecp256k1 {
```
Liamsi: can we rename `reader` to `rand`?

crypto/secp256k1/secp256k1.go
```diff
@@ -80,8 +81,13 @@ func (privKey PrivKeySecp256k1) Equals(other crypto.PrivKey) bool {
...
 	privKeyBytes := [32]byte{}
-	copy(privKeyBytes[:], crypto.CRandBytes(32))
+	io.ReadFull(reader, privKeyBytes[:])
```
Liamsi: This can return an `error` (and I don't see why we can expect it not to happen here). I think we should panic in that case (e.g. there is not enough entropy available to create a private key -> panic).

crypto/secp256k1/secp256k1.go
ValarDragon: Sure!

crypto/secp256k1/secp256k1.go
ValarDragon: My bad, forgot that io.readfull errors

ValarDragon:
So far, just found concretely the speed difference between ed25519 and secp. 
```
pkg: github.com/tendermint/tendermint/crypto/ed25519
BenchmarkKeyGeneration-8   	   20000	     99581 ns/op
BenchmarkSigning-8         	   10000	    101777 ns/op
BenchmarkVerification-8    	    5000	    275005 ns/op
```

For Secp256k1
```
pkg: github.com/tendermint/tendermint/crypto/secp256k1
BenchmarkKeyGeneration-8   	20000000	       117 ns/op
BenchmarkSigning-8         	    5000	    242135 ns/op
BenchmarkVerification-8    	    3000	    513049 ns/op
```

Also our ed25519 implementation's speed is _basically_ the same as the underlying library, within the expected randomness of a single benchmark of both, which is good news.

Issue #2067, opened 2018-07-25
P2P Refactor


xla:
In its current implementation the separation of concerns betweeen different components (AddrBook, Peer, PEX, Switch) is not solid, for reference see #1325, #1356. This comes at no surprise as this is one of the oldest packages in the codebase with almost 4 years of history. Over time different requirements have been tried to fit in and the scope of the package expanded, especially given its center role in the entire codebase. Additionally, a range of potential security issues surfaced (#2046, #2047) through external audits which could be addressed with structural changes and better separation.

In here we track the progress of the P2P improvements and try to gather all relevant issues, designs and docs.

ebuchman:
Some original thoughts about this in https://github.com/tendermint/tendermint/issues/1325.
Some of that has already been addressed, some is just pointers to take another look at the AddrBook and to clearing up the interplay between pex/seeds/persistent-peers/addrbook

ebuchman:
Replacing #2046 by noting here that we need to be able to write better tests for that.

ValarDragon:
We discussed to post potential future p2p improvements here as well. Here are some from the top of my head that are worth doing / considering.

* Use Noise for SecretConnection
* Send mempool txs in batches, and then use compression to mitigate the bandwidth. (The p2p change being adding compression for sufficiently large msgs)
* Change tx gossip algorithm https://github.com/tendermint/tendermint/issues/1798
* Enhance the peer scoring metrics
* Adapt block propagation to not send the block, just an ordered list of tx hashes, with optimistic sends of a few txs you think they may not have. Then have nodes request missing tx hashes from peers. (https://bitcoincore.org/en/2016/06/07/compact-blocks-faq/)
* Until we implement compact-blocks (which will be a while), compress block data before splitting into parts
* Use QUIC instead of TCP
* #Very postlaunch, consider modifying QUIC to not need message ordering. We'd have to switch from chacha to xchacha, but thats not a big deal.

ValarDragon:
I had the chance to talk to @rargulati at SFBW. He gave some extremely helpful insight into switching to libp2p. One of the examples he gave was making our address book align with the libp2p peer store's addrbook interface. Then we could just switch that piece over quite easily in the future (https://github.com/libp2p/go-libp2p-peerstore). We could then switch over the peer-store in isolation.

[...]

PR #2069, opened 2018-07-25
p2p: Propose PeerTransport ADR

xla:
First iteration in #2067 and an attempt to liberate the Switch from connection handling and low-level network concenrs.

### DIFF
--- a/docs/architecture/adr-012-peer-transport.md
+++ b/docs/architecture/adr-012-peer-transport.md
@@ -1,0 +1,93 @@
+# ADR 012: PeerTransport
+
+## Context
+
+One of the more apparent problems with the current architecture in the p2p
+package is that there is no clear separation of concerns between different
+components. Most notably the `Switch` is currently doing physical connection
+handling. An artifact is the dependency of the Switch on `config.P2PConfig`.
+
+Addresses #2046 #2047
+
+First iteraton in #2067
+
+## Decision
+
+### PeerTransport
+
+Responsible for emitting and connecting to Peers. The implementation of `Peer`
+is left to the transport, which implies that the chosen transport dictates the
+characteristics of the implementation handed back to the `Switch`. It is the
+place where we enforce low-level guards, like dropping connections from our own
+node.
+
+``` go
+// PeerTransport proxies incoming and outgoing peer connections.
+type PeerTransport interface {
+	// Accept returns a newly connected Peer.
+	Accept(peerConfig) (Peer, error)
+
+	// Dial connects to a Peer.
+	Dial(NetAddress, peerConfig) (Peer, error)
+}
+
+// EXAMPLE OF DEFAULT IMPLEMENTATION
+
+// multiplexTransport accepts tcp connections and upgrades to multiplexted
+// peers.
+type multiplexTransport struct {
+	listener net.Listener
+
+	acceptc chan accept
+	closec  <-chan struct{}
+	listenc <-chan struct{}
+
+	addr             NetAddress
+	dialTimeout      time.Duration
+	handshakeTimeout time.Duration
+	nodeInfo         NodeInfo
+	nodeKey          NodeKey
+
+	// TODO(xla): Remove when MConnection is refactored into mPeer.
+	mConfig conn.MConnConfig
+}
+
+var _ PeerTransport = (*multiplexTransport)(nil)
+
+// NewMTransport returns network connected multiplexed peers.
+func NewMTransport(
+	nodeAddr NetAddress,
+	nodeInfo NodeInfo,
+	nodeKey NodeKey,
+) *multiplexTransport 
+```
+
+### Switch
+
+From now the Switch will depend on a fully setup `PeerTransport` to
+retrieve/reach out to its peers. As the more low-level concerns are pushed to
+the transport, we can omit passing the `config.P2PConfig` to the Switch.
+
+``` go
+func NewSwitch(transport PeerTransport, opts ...SwitchOption) *Switch
+```
+
+## Status
+
+Proposed.
+
+## Consequences
+
+### Positive
+
+* free Switch from transport concerns
+* pluggable Peer implementation
+* dependency injection during tests
+
+### Negative
+
+* more setup for tests which depend on Switches
+
+### Neutral
+
+* multiplexed will be the default implementation


docs/architecture/adr-012-peer-transport.md
```diff
@@ -0,0 +1,96 @@
...
+First iteraton in [#2067](https://github.com/tendermint/tendermint/issues/2067)
+
+## Decision
+
+### PeerTransport
+
+Responsible for emitting and connecting to Peers. The implementation of `Peer`
+is left to the transport, which implies that the chosen transport dictates the
+characteristics of the implementation handed back to the `Switch`. It is the
+place where we enforce low-level guards, like dropping connections from our own
```
ValarDragon: Do we have a list of low level guards we want to enforce anywhere? Its not clear to me what the scope of "low-level" guards is.

docs/architecture/adr-012-peer-transport.md
```diff
@@ -0,0 +1,108 @@
...
+* handshake fails
+* upgrade to secret connection fails
+* protocol/app/... version doesn't match
+
+
+``` go
+// PeerTransport proxies incoming and outgoing peer connections.
+type PeerTransport interface {
+	// Accept returns a newly connected Peer.
+	Accept() (Peer, error)
```
ValarDragon: I'm having trouble understanding `Accept`. Is `Accept` supposed to evaluate an incoming connection? If so why doesn't have connection as a parameter

docs/architecture/adr-012-peer-transport.md
```diff
@@ -0,0 +1,108 @@
...
+Transport concerns will be handled by a new component (`PeerTransport`) which
+will provide Peers at its boundary to the caller. In turn `Switch` will use
+this new component accept new `Peer`s and dial them based on `NetAddress`.
+
+### PeerTransport
+
+Responsible for emitting and connecting to Peers. The implementation of `Peer`
+is left to the transport, which implies that the chosen transport dictates the
+characteristics of the implementation handed back to the `Switch`. It is the
+place where we enforce low-level guards[0]:
```
ValarDragon: Thanks for listing out the intended low-level guards! I still don't really understand what low-level means here. 

Are these the set of checks that can only be determined from information in the handshake and the local nodes' configuration? (No knowledge of other peers that this is node is connected to, or how many peers it is currently connected to?)

ValarDragon: More concretely, I'm confused about why the following aren't here:

"Ensure this isn't a duplicate connection from the same IP" or "Ensure that we don't already have too many peers"

xla: Very good point. Indeed the use of *low-level* might be missleading as it actually should say domain-specific. As long as we only handle aspects that belong to the transport domain it should be clear. This might result in that protocol/app version check is down somewhere else.  Most certainly the check if we already are connected to too many peers belongs in the switch.

Will add the duplicate ip check and rephrase low-level.

docs/architecture/adr-012-peer-transport.md
ValarDragon: > Accept is the way for the caller to signal that they are ready to deal with one more incoming Peer.

Ohh I see. Perhaps this may be more clearly named as "GetPeer"?

xla: Trying to avoid getter semantics when possible, in the domain of transport accept and dial are very well understood and the signature already communicates what is returned.

ValarDragon: Sounds good! Thank you for the explanations :)

docs/architecture/adr-012-peer-transport.md
```diff
@@ -0,0 +1,110 @@
...
+is left to the transport, which implies that the chosen transport dictates the
+characteristics of the implementation handed back to the `Switch`. Each
+transport implementation is responsible to filter establishing peers specific
+to its domain, for the default multiplexed implementation the following will
+apply:
+
+* connections from our own node
+* handshake fails
+* upgrade to secret connection fails
+* prevent duplicate ip
```
ebuchman: Should also filter based on:
- duplicate ID
- node info compatibility

could also clarify that it doesn't handle max-peers, thats for the switch

docs/architecture/adr-012-peer-transport.md
```diff
@@ -0,0 +1,110 @@
...
+// multiplexTransport accepts tcp connections and upgrades to multiplexted
+// peers.
+type multiplexTransport struct {
+	listener net.Listener
+
+	acceptc chan accept
+	closec  <-chan struct{}
+	listenc <-chan struct{}
+
+	addr             NetAddress
```
melekes: `nodeAddr`

docs/architecture/adr-012-peer-transport.md
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// EXAMPLE OF DEFAULT IMPLEMENTATION
+
+// multiplexTransport accepts tcp connections and upgrades to multiplexted
+// peers.
+type multiplexTransport struct {
+	listener net.Listener
+
+	acceptc chan accept
```
melekes: I prefer `acceptCh`, but feel free to ignore

docs/architecture/adr-012-peer-transport.md
```diff
@@ -0,0 +1,110 @@
+# ADR 012: PeerTransport
+
+## Context
+
+One of the more apparent problems with the current architecture in the p2p
+package is that there is no clear separation of concerns between different
+components. Most notably the `Switch` is currently doing physical connection
```
Liamsi: Can you link to some code (ideally) with a perma-link? I'm not familiar with the current implementation and it makes it easier to read and understand this. e.g. [`Switch`](https://github.com/tendermint/tendermint/blob/05a76fb517f50da27b4bfcdc7b4cf185fc61eff6/p2p/switch.go#L51-L78)

docs/architecture/adr-012-peer-transport.md
Liamsi: If you don't ignore: also, closeCh and listenCh for consistency.

docs/architecture/adr-012-peer-transport.md
```diff
@@ -0,0 +1,110 @@
...
+## Status
+
+In Review.
+
+## Consequences
+
+### Positive
+
+* free Switch from transport concerns - simpler implementation
+* pluggable transport implementation - simpler test setup
```
Liamsi: I know what you mean but could we use another word here? [pluggable transports](https://www.pluggabletransports.info/) often means sth. else.

docs/architecture/adr-012-peer-transport.md
```diff
@@ -0,0 +1,110 @@
+# ADR 012: PeerTransport
+
+## Context
+
+One of the more apparent problems with the current architecture in the p2p
+package is that there is no clear separation of concerns between different
+components. Most notably the `Switch` is currently doing physical connection
+handling. An artifact is the dependency of the Switch on `config.P2PConfig`.
```
Liamsi: Link would be helpful again.

docs/architecture/adr-012-peer-transport.md
Liamsi: Will there be another ADR explaining *why* these are the right choices?

xla: @Liamsi Which choices are you referring to?

docs/architecture/adr-012-peer-transport.md
xla: What would you prefer?

docs/architecture/adr-012-peer-transport.md
xla: Link to what in that section would you like to see?

docs/architecture/adr-012-peer-transport.md
Liamsi: The choices of what was previously called "guards".

docs/architecture/adr-012-peer-transport.md
Liamsi: [`config.P2PConfig`](https://github.com/tendermint/tendermint/blob/05a76fb517f50da27b4bfcdc7b4cf185fc61eff6/config/config.go#L272-L339)

docs/architecture/adr-012-peer-transport.md
Liamsi: For me the words pluggable transport is overloaded by what I linked above (means tor obfuscate tor traffic as regular https traffic). But yeah in this context it describes correctly what this is about: the ability to plug in different transport implementations. So that's fine.

docs/architecture/adr-012-peer-transport.md
ebuchman: Would be valuable to include some explanation in here as to the "why" for sure

docs/architecture/adr-012-peer-transport.md
ebuchman: While I appreciate this point, we need to strike a balance in these ADRs between being an efficient document for design and including enough information for an outsider to keep up. I think to some extent, readers of the ADR should be expected to have some familiarity with the codebase, so that we don't need to worry too much about adding too many links.

Issue #1958, opened 2018-07-11
crypto: Handling of Secp256k1 signature malleability


ValarDragon:
TL;DR: Do we want the following to be an invariant of signature.Equal: It returns true for two signatures which are both valid for a given public key and message pair. If we want this to be an invariant we have to mildly adapt the equality implementation for secp256k1.

Secp256k1's signing algorithm is ecdsa. In ecdsa, the signature is of the form (r,s). However anyone given a valid signature on a message can produce a second valid signature (r, -s). So both of these signatures are valid, but have different byte representations.

Currently we load secp256k1 signatures directly from its bytes, and then check equality by checking if the byte representations are equal. However two valid signatures may not be considered equal currently. I think it may be easiest for future functionality to consider these as equal. This is also related to #1664. As far as I can tell, the downsides of this are that we've deviated slightly from the standard, but I'm unsure of how important that is in this case, and that decoding bytes to signature takes a bit more time. The upside is its one less attack / bug vector to be concerned about in downstream usage.

Other alternatives are to adapt the verify method to reject an ecdsa signature whose s is not in canonical form, or to have all validators force convert all s parameters into canonical form. (Canonical as being the smaller value within the finite field)

Of course, this only needs to be changed if this malleability is an issue. (I.e. Perhaps if a certain signature hash is expected)

/cc @ebuchman

ValarDragon:
This is in progress. The changes in the upstream library are in https://github.com/tendermint/btcd/pull/1. Once that is merged in, we just need to update crypto/secp256k1 to use that API. (That PR also changes the secp256k1 encoding, so that we match #2121)

melekes:
btw, are you planning to contribute to the original library?

ValarDragon:
They would reject this change because it is backwards incompatible. I'll contribute any speed improvements or constant time changes tho

melekes:
Merged to develop.

PR #2091, opened 2018-07-27
ADR: Fix malleability problems in Secp256k1 signatures

ValarDragon:
Previously you could not assume that your transaction hash would
appear on chain.

Ref #1958

### DIFF
--- a/docs/architecture/adr-014-secp-malleability.md
+++ b/docs/architecture/adr-014-secp-malleability.md
@@ -1,0 +1,53 @@
+# ADR 014: Secp256k1 Signature Malleability
+
+## Context
+
+Secp256k1 has two layers of malleability.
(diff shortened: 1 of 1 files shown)


zmanian:
Given that our system is fundamentally account based, I can't currently an imagine a protocol that requires signing transaction hashes.

But ECDSA signatures malleability is annoying. Apparently we don't DER encode ECDSA signatures and instead just encode r||s. So all we need to do is protect against inverting s instead of all the DER encoding malleability BItcoin had to deal with.

zmanian:
It sounds like our encoding scheme is like Ethereum without the precomputation byte for fast pub key recovery.

zmanian:
Bitcoin and Ethereum both only accept signatures with the upper s value as proposed in the ADR

Issue #1134, opened 2018-01-21
Add proposer to header


ebuchman:
This should be the original proposer of the block. 

When a validator makes a block to propose, they will include their own address in the header.

Will also need to be reflected in the ABCI header so the app gets access to this too.

ebuchman:
We also need to include the Round in BeginBlock somehow so apps can incentivize based on it.

Not sure if we should do `LastCommitRound` in the next header or just include the `CommitRound` in the BeginBlockRequest

ebuchman:
Closed #2106 and #2156

PR #2106, opened 2018-07-30
add proposer address to block's Header

melekes:
Refs #1134

* [x] Wrote tests
* [x] Updated CHANGELOG.md


docs/spec/blockchain/blockchain.md
```diff
@@ -56,8 +56,8 @@ type Header struct {
     ConsensusParamsHash []byte  // SimpleMerkle of the ConsensusParams
 
     // consensus
-    Proposer            []byte  // Address of the block proposer
     EvidenceHash        []byte  // SimpleMerkle of []Evidence
+    ProposerAddress     []byte  // Address of the original proposer of the block
```
melekes: It's a bit strange to call it just `Proposer` because it's really an address (without pubkey or power).

docs/spec/blockchain/blockchain.md
```diff
@@ -56,8 +56,8 @@ type Header struct {
     ConsensusParamsHash []byte  // SimpleMerkle of the ConsensusParams
 
     // consensus
-    Proposer            []byte  // Address of the block proposer
```
melekes: I think `ProposerAddress` better reflects the field's value. It's really an address (without pubkey or power).

melekes:
> Would also be really great if we could write a consensus test for this so we can test that the right proposer is included when

Isn't it obvious from the code? Or you want to make sure we are not overwriting it?

types/block_test.go
```diff
@@ -44,6 +44,7 @@ func TestBlockValidateBasic(t *testing.T) {
 
 	block := MakeBlock(h, txs, commit, evList)
 	require.NotNil(t, block)
+	block.ProposerAddress = valSet.GetProposer().Address
```
ebuchman: I think if we're going to require the proposer in ValidateBasic, we need to accommodate it in MakeBlock. Any block without it will fail validation.

Maybe the right middle ground is to change ValidateBasic to allow an empty proposer, but if its given it must be tmhash.Size

ebuchman:
Ok I'll wrap this up - will remove the check from ValidateBasic and add a ProposerAddress to state.MakeBlock

ebuchman:
Turns out the test I used to show we weren't including the right proposer was causing races in other tests (since it depended on the createProposalBlock method doing a lookup on the validator set, which was not thread safe). So I removed it. Definitely continues to be a big issue for us that we can't easily test consensus relevant changes like this.

ebuchman:
Going to squash these commits and rebase on develop with #1815 merged

melekes:
Thank you for spotting the issue with the incorrect address! And you're right, tests should be required, but in the same time "Definitely continues to be a big issue for us that we can't easily test consensus relevant changes like this." 😞

PR #2113, opened 2018-07-30
libs: AtomicWriteFile filename collision fixes

### LINKED ISSUE #2109: libs/common: Temporary File Name Collision
https://github.com/tendermint/tendermint/blob/40d6dc2ee55a34a075e7b7e73c3525c84ae705ea/libs/common/os.go#L129-L136

> Although unlikely, RandStr() may return the same value during two concurrent executions of WriteFileAtomic(), causing the same temporary file name to be used. This may corrupt the contents within the temporary file.

Recommendation:
> use Go's `TempFile` function to generate a temporary file

ValarDragon:
This uses the stdlib's method of creating a tempfile in our write
file atomimc method, with a few modifications. We use a 64 bit number
rather than 32 bit, and therefore a corresponding LCG. This is to
reduce collision probability. (Note we currently use 24 bytes of entropy,
so switching to 32 bits is a significant change. 
We can move from 64 bits to 32 bits in the future)

We handle reseeding the LCG in such a way that multiple threads are
even less likely to reuse the same seed. 
Before if two processes were spawned that needed write file atomics at the same time, then there is a decent chance of seed collision. This is now less likely.

I'm working on writing a test case to demonstrate all the behaviors are working correctly.

Closes #2109 

* [x] Wrote tests
* [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -22,6 +22,6 @@ IMPROVEMENTS:
 - [common] bit array functions which take in another parameter are now thread safe
 
 BUG FIXES:
-- [common] \#2109 Safely handle cases where atomic write files already exist
+- [common] Safely handle cases where atomic write files already exist [#2109](https://github.com/tendermint/tendermint/issues/2109)
 - [privval] fix a deadline for accepting new connections in socket private
   validator.
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -22,5 +22,6 @@ IMPROVEMENTS:
 - [common] bit array functions which take in another parameter are now thread safe
 
 BUG FIXES:
+- [common] \#2109 Safely handle cases where atomic write files already exist
 - [privval] fix a deadline for accepting new connections in socket private
   validator.
--- a/libs/common/tempfile.go
+++ b/libs/common/tempfile.go
@@ -1,28 +1,117 @@
 package common
 
 import (
+	fmt "fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
 )
 
-// WriteFileAtomic creates a temporary file with data and the perm given and
+const (
+	atomicWriteFilePrefix = "write-file-atomic-"
+	// Maximum number of atomic write file conflicts before we start reseeding
+	// (reduced from golang's default 10 due to using an increased randomness space)
+	atomicWriteFileMaxNumConflicts = 5
+	// Maximum number of attempts to make at writing the write file before giving up
+	// (reduced from golang's default 10000 due to using an increased randomness space)
+	atomicWriteFileMaxNumWriteAttempts = 1000
+	// LCG constants from Donald Knuth MMIX
+	// This LCG's has a period equal to 2**64
+	lcgA = 6364136223846793005
+	lcgC = 1442695040888963407
+	// Create in case it doesn't exist and force kernel
+	// flush, which still leaves the potential of lingering disk cache.
+	// Never overwrites files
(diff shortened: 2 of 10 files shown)


libs/common/tempfile.go
```diff
@@ -0,0 +1,127 @@
...
+	// Close the file before renaming it, otherwise it will cause "The process
+	// cannot access the file because it is being used by another process." on windows.
+	f.Close()
+
+	return os.Rename(f.Name(), filename)
+}
+
+//--------------------------------------------------------------------------------
+
+func Tempfile(prefix string) (*os.File, string) {
```
melekes: I think we should get rid of these two methods and use underlying `ioutil` methods. For what I see, the only thing they do is panic on error which a) undocumented b) not really needed anywhere

ValarDragon:
This is ready for review! This should be easiest to review commit by commit (I think).

The reason this is not hitting the goal code coverage (unless latest commit fixed that, idk yet), is because I moved Tempfile and Tempdir. Those files never had test cases, and seem slated for removal. (See Melekes' comment) All of the actual new code I wrote _should_ be tested, except for 1000 files in a row already existing. 
(Thats both logistically hard to test due to writing code in a way to avoid that scenario, and not that useful of a thing to test imo)

libs/common/tempfile.go
```diff
@@ -0,0 +1,149 @@
...
+		f   *os.File
+	)
+
+	nconflict := 0
+	// Limit the number of attempts to create a file. Something is seriously
+	// wrong if it didn't get created after 1000 attempts, and we don't want
+	// an infinite loop
+	i := 0
+	for ; i < atomicWriteFileMaxNumWriteAttempts; i++ {
+		name := filepath.Join(dir, atomicWriteFilePrefix+randWriteFileSuffix())
```
melekes: can't we just use `ioutil.TempFile` here? As far as I can see we're doing the same thing ioutil.TempFile does.

ValarDragon: We need different flags on the atomic write file. We want it to be write only, and it to be written synchronously, the tempfile doesn't provide us that.

melekes: Oh, right. Thanks

CHANGELOG_PENDING.md
```diff
@@ -22,5 +22,6 @@ IMPROVEMENTS:
 - [common] bit array functions which take in another parameter are now thread safe
 
 BUG FIXES:
+- [common] \#2109 Safely handle cases where atomic write files already exist
```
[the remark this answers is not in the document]
ValarDragon: # renders as markdown in some places. It just lowers the chance of this rendering weirdly in some markdown editor somewhere

libs/common/tempfile.go
ValarDragon: PR already made :) #2114

libs/common/tempfile.go
ValarDragon: The idea behind the write file is we only write to this file, and then just rename it. There is no need for opening with Read + Write. We also require that we write to this file synchronously. (So its better not to have Read, to avoid any blocking)

libs/common/tempfile.go
Liamsi: Haha, you are a machine :-D

libs/common/tempfile.go
Liamsi: Thanks so much for the clarification!

xla: @ValarDragon  This is spot on and I dunno how but you inferred it from the code without any comments, maybe we can leave this reational in the code so people understand the full context and reasoning behind not using exisiting APIs.

ValarDragon:
This should be good to merge now! I left the refactor as a seperate commit, as I think commits that have refactors + changes make it harder to see the actual change when going through history.

PR #2115, opened 2018-07-31
p2p/pex: Connect to peers from a seed node immediately

### LINKED ISSUE #2093: p2p: start dialing peers sooner
A new node with no peers will dial a seed and get a list of addresses, but then it seems to wait a while before it dials any of them. 

Would be much nicer if it knew to dial them right away

ValarDragon:
This is to reduce wait times when initially connecting. This still runs checks
such as whether you still want additional peers.

A test case has been created, which fails if this change is not included.

This is to reduce wait times when initially connecting. This still runs checks
such as whether you still want additional peers.

A test case has been created, which fails if this change is not included.

Closes #2093 

* [ ] Updated all relevant documentation in docs - should I update the docs here?
* [X] Wrote tests
* [X] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -20,7 +20,7 @@ IMPROVEMENTS:
     - only process one block at a time to avoid starving
 - [crypto] Switch hkdfchachapoly1305 to xchachapoly1305
 - [common] bit array functions which take in another parameter are now thread safe
-- [p2p] \#2093 begin connecting to peers as soon a seed node provides them to you
+- [p2p] begin connecting to peers as soon a seed node provides them to you ([#2093](https://github.com/tendermint/tendermint/issues/2093))
 
 BUG FIXES:
 - [privval] fix a deadline for accepting new connections in socket private
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -20,6 +20,7 @@ IMPROVEMENTS:
     - only process one block at a time to avoid starving
 - [crypto] Switch hkdfchachapoly1305 to xchachapoly1305
 - [common] bit array functions which take in another parameter are now thread safe
+- [p2p] \#2093 begin connecting to peers as soon a seed node provides them to you
 
 BUG FIXES:
 - [privval] fix a deadline for accepting new connections in socket private
--- a/p2p/pex/pex_reactor.go
+++ b/p2p/pex/pex_reactor.go
@@ -74,6 +74,8 @@ type PEXReactor struct {
 	requestsSent         *cmn.CMap // ID->struct{}: unanswered send requests
 	lastReceivedRequests *cmn.CMap // ID->time.Time: last time peer requested from us
 
+	seedAddrs []*p2p.NetAddress
+
 	attemptsToDial sync.Map // address (string) -> {number of attempts (int), last time dialed (time.Time)}
 }
 
@@ -120,9 +122,12 @@ func (r *PEXReactor) OnStart() error {
 
 	// return err if user provided a bad seed address
 	// or a host name that we cant resolve
-	if err := r.checkSeeds(); err != nil {
+	seedAddrs, err := r.checkSeeds()
+	if err != nil {
 		return err
 	}
+
+	r.seedAddrs = seedAddrs
 
 	// Check if this node should run
 	// in seed/crawler mode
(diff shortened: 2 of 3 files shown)


melekes:
> should I update the docs here?

Ideally, you should update p2p pex spec https://github.com/tendermint/tendermint/blob/develop/docs/spec/reactors/pex/pex.md

CHANGELOG_PENDING.md
```diff
@@ -20,6 +20,7 @@ IMPROVEMENTS:
     - only process one block at a time to avoid starving
 - [crypto] Switch hkdfchachapoly1305 to xchachapoly1305
 - [common] bit array functions which take in another parameter are now thread safe
+- [p2p] \#2093 begin connecting to peers as soon a seed node provides them to you
```
xla: Let's create a markdown link to the issue and have it at the end.

p2p/pex/pex_reactor_test.go
```diff
@@ -417,14 +427,77 @@ func assertPeersWithTimeout(
...
+				nPeers, numPeersStr,
 			)
 			return
 		}
 	}
 }
 
+// Creates a seed which knows about the provided addresses / source address pairs.
+// Starting and stopping the seed is left to the caller
+func createSeed(dir string, id int, knownAddrs, srcAddrs []*p2p.NetAddress) *p2p.Switch {
```
xla: We should make clear this is for testing purposes with a `test` prefix to avoid collision.

p2p/pex/pex_reactor_test.go
```diff
@@ -417,14 +427,77 @@ func assertPeersWithTimeout(
...
+	)
+	seed.AddListener(
+		p2p.NewDefaultListener("tcp://"+seed.NodeInfo().ListenAddr, "", false, log.TestingLogger()),
+	)
+	return seed
+}
+
+// Creates a peer which knows about the provided seed.
+// Starting and stopping the peer is left to the caller
+func createPeerWithSeed(dir string, id int, seed *p2p.Switch) *p2p.Switch {
```
xla: Same as above.

p2p/pex/pex_reactor.go
```diff
@@ -297,6 +296,15 @@ func (r *PEXReactor) ReceiveAddrs(addrs []*p2p.NetAddress, src Peer) error {
 
 		err := r.book.AddAddress(netAddr, srcAddr)
 		r.logErrAddrBook(err)
+
+		// If this address came from a seed node, try to connect to it without waiting.
+		// TODO: Change this to an O(log(N)) check on a sorted seedAddr list
+		seedAddrs, _ := p2p.NewNetAddressStrings(r.config.Seeds)
```
xla: As seeds are not changing throughout the lifecycle of the process wouldn't it make sense to create a lookup (`map[NetAddress]struct{}`) once and use that instead.

ValarDragon: Great point!

CHANGELOG_PENDING.md
ValarDragon: Not sure what you mean by "have it at the end". Do you mean make this a bug fix?

xla: Sorry I was ambiguous. Make the issue reference an actual link and have it at the end of the line, like this:

``` md
[p2p] begin connecting to peers as soon a seed node provides them to you ([#2093](https://github.com/tendermint/tendermint/issues/2093))
```

ValarDragon: :o thanks!

p2p/pex/pex_reactor.go
ValarDragon: I'll change it to this for now, but on further thought I'm worried that the constant overhead of a hash function may make this less efficient in the expected case of a few seeds. But this is definitely on the order of things we can optimize post launch!

xla: Which hash function are you referring to?

ValarDragon: the internal hash function which will be used in the hashmap. (Namely its `memhash` here on 64 bit systems: https://golang.org/src/runtime/hash64.go)

xla: I don't think this is something we need to optimise for, even post-launch.

ValarDragon: Just ran the golang source runtime benchmarks. Slightly skewed speeds on my system, since it has the very nice hardware AES hash :)
```
BenchmarkHash5-8                          	200000000	         9.12 ns/op	 548.11 MB/s
BenchmarkHash16-8                         	200000000	         9.59 ns/op	1668.44 MB/s
BenchmarkHash64-8                         	100000000	        13.4 ns/op	4777.66 MB/s
BenchmarkHash1024-8                       	20000000	        76.0 ns/op	13467.29 MB/s
BenchmarkHash65536-8                      	  300000	      4340 ns/op	15097.94 MB/s
```
For a small number of seeds, a linear scan would probably be faster. But I'll switch to the hashmap here anyway, optimizing linear / logarithmic scans vs hashmaps is something we can defo do later. I'll write this up into an issue though, since it seems useful

Issue #1969, opened 2018-07-14
Tendermint no longer producing blocks after many large txs


ValarDragon:
**BUG REPORT**

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):
0.22.2-dev-1dbe7b7e
(dev/tmbench_improve_sendloop_efficiency)

**Environment**:
- **OS** (e.g. from /etc/os-release): Ubuntu 16.04
- **Install tools**:
- **Others**:

**What happened**:
Tendermint stopped building blocks, replied with `Internal msg queue is full. Using a go-routine module=consensus` a bunch of times, then bumping the round number and repeating the process.

**What you expected to happen**:
Tendermint to continue block production 

**How to reproduce it** (as minimally and precisely as possible):
Clean Tendermint install. Run 
`tm-bench -v -c 10 -T 3 -r 50 -s 100000 localhost:26657`

**Logs (you can paste a small part showing an error or link a pastebin, gist, etc. containing more of the log file)**:

```
I[07-14|01:35:08.804] Internal msg queue is full. Using a go-routine module=consensus 
I[07-14|01:35:08.804] Internal msg queue is full. Using a go-routine module=consensus 
I[07-14|01:35:08.804] Signed proposal                              module=consensus height=8 round=25 proposal="Proposal{8/25 2265:033EB133F138 (24,:0:000000000000) /CCE9C3412B8F.../ @ 2018-07-14T01:35:08.747Z}"
I[07-14|01:35:08.830] Received proposal                            module=consensus proposal="Proposal{8/25 2265:033EB133F138 (24,:0:000000000000) /CCE9C3412B8F.../ @ 2018-07-14T01:35:08.747Z}"
```

**Config (you can paste only the changes you've made)**:
Default

ValarDragon:
Just repeated the above process on a clean install, and was able to get tendermint to basically hang. 
Logs:
https://pastebin.com/qgpcxZs0
(Note that last output kept repeating, no blocks got produced, and no new txs were accepted)

Lsof: 
https://pastebin.com/phkqcNxX

kill -6:
https://pastebin.com/1mHABgvM

melekes:
It's a known issue. 1) we should enforce max block size & max tx size 2) clients need to setup rate limiting (nginx / caddy can do that)

ValarDragon:
Sorry, didn't realize this was a known problem.

[...]

PR #2123, opened 2018-08-01
WIP: mempool: Switch Reap to use bytes instead of number of txs

### LINKED ISSUE #2035: consensus: remove MaxTxs for MaxBytes
We currently use MaxTxs to reap txs from the mempool when proposing a block, but enforce MaxBytes when unmarshalling a block, so we could easily propose a block thats too large to be valid.

We should probably just remove MaxTxs all together and stick with MaxBytes, and have a mempool.ReapBytes

ValarDragon:
This closes #2035. This also solves #1969.
The latter error was tendermint not producing blocks if the data in
a block got larger than max block size. I have tested that this fix
works through tm-bench. This also removes Max_Txs from
blocks.

* [ ] Wrote tests 
* [ ] Updated CHANGELOG.md

### DIFF
--- a/mempool/mempool_test.go
+++ b/mempool/mempool_test.go
@@ -56,10 +56,11 @@ func ensureFire(t *testing.T, ch <-chan struct{}, timeoutMS int) {
 	}
 }
 
-func checkTxs(t *testing.T, mempool *Mempool, count int) types.Txs {
+func checkTxs(t *testing.T, mempool *Mempool, count int) (types.Txs, int) {
 	txs := make(types.Txs, count)
+	bytesPerTx := 20
 	for i := 0; i < count; i++ {
-		txBytes := make([]byte, 20)
+		txBytes := make([]byte, bytesPerTx)
 		txs[i] = txBytes
 		_, err := rand.Read(txBytes)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


state/validation.go
```diff
@@ -42,6 +42,12 @@ func validateBlock(stateDB dbm.DB, state State, block *types.Block) error {
 		return fmt.Errorf("Wrong Block.Header.TotalTxs. Expected %v, got %v", state.LastBlockTotalTx+newTxs, block.TotalTxs)
 	}
 
+	// Don't risk recomputing, since marshalling may be slow.
+	blockSize := block.Size()
+	if blockSize > state.ConsensusParams.BlockSize.MaxBytes {
+		return fmt.Errorf("Block is too large. Got %d, maximum is %d", blockSize, state.ConsensusParams.BlockSize.MaxBytes)
```
ValarDragon: Is this the correct place to ensure that the received block isn't too large? It would be more ideal to ensure this before decoding the block.

ValarDragon: The race conditions this causes makes it clear that this is not the correct place to check this.

ValarDragon: melekes pointed out that this already ensured here: https://github.com/tendermint/tendermint/blob/dde96b75ce3684283f980b968b9edb29ecdff973/consensus/state.go#L1420

PR #2124, opened 2018-08-01
abci: Update readme for building protoc

ValarDragon:
* [X] Wrote tests - n/a
* [X] Updated CHANGELOG.md - seems unnecessary

### DIFF
--- a/abci/README.md
+++ b/abci/README.md
@@ -29,7 +29,7 @@ The two guides to focus on are the `Application Development Guide` and `Using AB
 To compile the protobuf file, run:
 
 ```
-make protoc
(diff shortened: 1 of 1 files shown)


abci/README.md
```diff
@@ -29,7 +29,7 @@ The two guides to focus on are the `Application Development Guide` and `Using AB
 To compile the protobuf file, run:
 
 ```
-make protoc
```
Liamsi: To recompile the proto files one currently still has to delete `abci/types/types.pb.go` or not? `make protoc_abci` doesn't have any effect, if the file already exists?

melekes: no, if there're some changes, it will apply them

PR #2126, opened 2018-08-01
[WIP] PeerTransport

xla:
**WORK IN PROGRESS**

### DIFF
--- a/p2p/transport.go
+++ b/p2p/transport.go
@@ -1,0 +1,309 @@
+package p2p
+
+import (
+	"errors"
+	"net"
+	"time"
+
+	"github.com/tendermint/tendermint/config"
+	crypto "github.com/tendermint/tendermint/crypto"
+	"github.com/tendermint/tendermint/p2p/conn"
+)
+
+// accept is the container to carry the upgrade connection and nodeInfo from
+// asynchronously running routine to the Accept method.
+type accept struct {
+	conn     net.Conn
+	nodeInfo NodeInfo
+	err      error
+}
+
+// peerConfig is used to bundle data we need to fully setup a Peer with an
+// MConn, provided by the caller of Accept and Dial.
+type peerConfig struct {
+	chDescs              []*conn.ChannelDescriptor
+	onPeerError          func(Peer, interface{})
+	outbound, persistent bool
+	reactorsByCh         map[byte]Reactor
+}
+
+// Transport emits and connects to Peers. The implementation of Peer is left to
+// the transport. Each transport is also responsible to filter establishing
+// peers specific to its domain.
+type Transport interface {
+	// Accept returns a newly connected Peer.
+	Accept(peerConfig) (Peer, error)
+
+	// Dial connects to the Peer for the address.
+	Dial(NetAddress, peerConfig) (Peer, error)
+}
+
+// transportLifecycle bundles the methods for callers to control start and stop
+// behaviour.
+type transportLifecycle interface {
+	Close() error
+	Listen(NetAddress) error
+}
+
+// multiplexTransport accepts and dials tcp connections and upgrades them to
+// multiplexed peers.
+type multiplexTransport struct {
+	listener net.Listener
+
+	acceptc chan accept
+	closec  chan struct{}
+
+	// Lookup table for duplicate ip and id checks.
+	peers map[net.Conn]NodeInfo
+
+	dialTimeout      time.Duration
+	handshakeTimeout time.Duration
+	nodeAddr         NetAddress
+	nodeInfo         NodeInfo
+	nodeKey          NodeKey
+
+	// TODO(xla): Those configs are still needed as we parameterise peerConn and
+	// peer currently. All relevant options should be refactored into options with
+	// sane defaults.
+	mConfig   conn.MConnConfig
+	p2pConfig config.P2PConfig
+}
+
+// Test multiplexTransport for interface completeness.
+var _ Transport = (*multiplexTransport)(nil)
+var _ transportLifecycle = (*multiplexTransport)(nil)
+
+// NewMultiplexTransport returns a tcp conencted multiplexed peers.
+func NewMultiplexTransport() *multiplexTransport {
+	return &multiplexTransport{
+		acceptc: make(chan accept),
+		closec:  make(chan struct{}),
+	}
+}
+
+func (mt *multiplexTransport) Accept(cfg peerConfig) (Peer, error) {
+	select {
+	case a := <-mt.acceptc:
+		if a.err != nil {
+			return nil, a.err
+		}
+
+		err := mt.filterPeer(a.conn, a.nodeInfo)
+		if err != nil {
+			return nil, errors.New("peer filtered")
+		}
+
+		cfg.outbound = false
+
+		return mt.wrapPeer(a.conn, a.nodeInfo, cfg), nil
+	case <-mt.closec:
+		return nil, errors.New("transport is closed")
+	}
+}
+
+func (mt *multiplexTransport) Dial(
+	addr NetAddress,
+	cfg peerConfig,
+) (Peer, error) {
+	c, err := addr.DialTimeout(mt.dialTimeout)
+	if err != nil {
+		return nil, err
+	}
+
+	uc, ni, err := upgrade(c, mt.handshakeTimeout, mt.nodeKey.PrivKey, mt.nodeInfo)
+	if err != nil {
+		return nil, errors.New("upgrade failed")
+	}
+
+	err = mt.filterPeer(uc, ni)
+	if err != nil {
+		return nil, errors.New("peer filtered")
+	}
+
+	cfg.outbound = true
+
+	return mt.wrapPeer(uc, ni, cfg), nil
+}
+
+func (mt *multiplexTransport) Close() error {
+	close(mt.closec)
+
+	return mt.listener.Close()
+}
+
+func (mt *multiplexTransport) Listen(addr NetAddress) error {
+	ln, err := net.Listen("tcp", addr.DialString())
+	if err != nil {
+		return err
+	}
+
+	mt.listener = ln
+	mt.nodeAddr = addr
+
+	go mt.acceptPeers()
+
+	return nil
+}
+
+func (mt *multiplexTransport) acceptPeers() {
+	for {
+		c, err := mt.listener.Accept()
+		if err != nil {
+			// Close has been called so we silently exit.
+			if _, ok := <-mt.closec; !ok {
+				return
+			}
+
+			mt.acceptc <- accept{err: err}
+			return
+		}
+
+		go func(conn net.Conn) {
+			c, ni, err := upgrade(
+				c,
+				mt.handshakeTimeout,
+				mt.nodeKey.PrivKey,
+				mt.nodeInfo,
+			)
+
+			select {
+			case mt.acceptc <- accept{c, ni, err}:
+				// Make the upgraded peer available.
+			case <-mt.closec:
+				// Give up if the transport was closed.
+				_ = c.Close()
+				return
+			}
+		}(c)
+	}
+}
+
+func (mt *multiplexTransport) filterPeer(c net.Conn, ni NodeInfo) error {
+	// TODO(xla): Implement IP filter.
+	// TODO(xla): Implement ID filter.
+	// TODO(xla): Check NodeInfo compatibility.
+	return nil
+}
+
+func (mt *multiplexTransport) wrapPeer(
+	c net.Conn,
+	ni NodeInfo,
+	cfg peerConfig,
+) Peer {
+	pc := peerConn{
+		conn:       c,
+		config:     &mt.p2pConfig,
+		outbound:   cfg.outbound,
+		persistent: cfg.persistent,
+	}
+
+	mt.peers[c] = ni
+
+	return &multiplexPeer{
+		onStop: func() {
+			delete(mt.peers, c)
+		},
+		peer: newPeer(
+			pc,
+			mt.mConfig,
+			ni,
+			cfg.reactorsByCh,
+			cfg.chDescs,
+			cfg.onPeerError,
+		),
+	}
+}
+
+func handshake(
+	c net.Conn,
+	timeout time.Duration,
+	nodeInfo NodeInfo,
+) (NodeInfo, error) {
+	if err := c.SetDeadline(time.Now().Add(timeout)); err != nil {
+		return NodeInfo{}, err
+	}
+
+	var (
+		errc = make(chan error, 2)
+
+		peerNodeInfo NodeInfo
+	)
+
+	go func(errc chan<- error, c net.Conn) {
+		_, err := cdc.MarshalBinaryWriter(c, nodeInfo)
+		errc <- err
+	}(errc, c)
+	go func(errc chan<- error, c net.Conn) {
+		_, err := cdc.UnmarshalBinaryReader(
+			c,
+			&peerNodeInfo,
+			int64(MaxNodeInfoSize()),
+		)
+		errc <- err
+	}(errc, c)
+
+	for i := 0; i < cap(errc); i++ {
+		err := <-errc
+		if err != nil {
+			return NodeInfo{}, err
+		}
+	}
+
+	return peerNodeInfo, c.SetDeadline(time.Time{})
+}
+
+func secretConn(
+	c net.Conn,
+	timeout time.Duration,
+	privKey crypto.PrivKey,
+) (net.Conn, error) {
+	if err := c.SetDeadline(time.Now().Add(timeout)); err != nil {
+		return nil, err
+	}
+
+	c, err := conn.MakeSecretConnection(c, privKey)
+	if err != nil {
+		return nil, err
+	}
+
+	return c, c.SetDeadline(time.Time{})
+}
+
+func upgrade(
+	conn net.Conn,
+	timeout time.Duration,
+	privKey crypto.PrivKey,
+	nodeInfo NodeInfo,
+) (sc net.Conn, ni NodeInfo, err error) {
+	defer func() {
+		if err != nil {
+			_ = conn.Close()
+		}
+	}()
+
+	c, err := secretConn(conn, timeout, privKey)
+	if err != nil {
+		return nil, NodeInfo{}, err
+	}
+
+	ni, err = handshake(c, timeout, nodeInfo)
+	if err != nil {
+		return nil, NodeInfo{}, err
+	}
+
+	return c, ni, nil
+}
+
+// multiplexPeer is the Peer implementation returned by the multiplexTransport
+// and wraps the default peer implementation for now, with an added hook for
+// shutdown of the a peer. It should ultimately grow into the proper
+// implementation.
+type multiplexPeer struct {
+	*peer
+	onStop func()
+}
+
+func (mp *multiplexPeer) OnStop() {
+	mp.onStop()
+	mp.peer.OnStop()
+}
--- a/p2p/transport.go
+++ b/p2p/transport.go
@@ -2,6 +2,7 @@ package p2p
 
 import (
 	"errors"
+	"fmt"
 	"net"
 	"time"
 
@@ -10,7 +11,12 @@ import (
 	"github.com/tendermint/tendermint/p2p/conn"
 )
 
-// accept is the container to carry the upgrade connection and nodeInfo from
+const (
+	defaultDialTimeout      = time.Second
+	defaultHandshakeTimeout = 3 * time.Millisecond
+)
+
+// accept is the container to carry the upgraded connection and NodeInfo from an
 // asynchronously running routine to the Accept method.
 type accept struct {
 	conn     net.Conn
@@ -63,8 +69,8 @@ type multiplexTransport struct {
 	nodeKey          NodeKey
 
 	// TODO(xla): Those configs are still needed as we parameterise peerConn and
-	// peer currently. All relevant options should be refactored into options with
-	// sane defaults.
+	// peer currently. All relevant configuration should be refactored into options
+	// with sane defaults.
 	mConfig   conn.MConnConfig
 	p2pConfig config.P2PConfig
 }
@@ -74,10 +80,15 @@ var _ Transport = (*multiplexTransport)(nil)
 var _ transportLifecycle = (*multiplexTransport)(nil)
 
 // NewMultiplexTransport returns a tcp conencted multiplexed peers.
-func NewMultiplexTransport() *multiplexTransport {
+func NewMultiplexTransport(nodeKey NodeKey) *multiplexTransport {
 	return &multiplexTransport{
-		acceptc: make(chan accept),
-		closec:  make(chan struct{}),
+		acceptc:          make(chan accept),
+		closec:           make(chan struct{}),
+		dialTimeout:      defaultDialTimeout,
+		handshakeTimeout: defaultHandshakeTimeout,
+		mConfig:          conn.DefaultMConnConfig(),
+		nodeKey:          nodeKey,
+		peers:            map[net.Conn]NodeInfo{},
 	}
 }
 
@@ -158,6 +169,10 @@ func (mt *multiplexTransport) acceptPeers() {
 			return
 		}
 
+		// Connection upgrade should be asynchronous to avoid Head-of-line blocking[0]
+		// Reference:  https://github.com/tendermint/tendermint/issues/2047
+		//
+		// [0] https://en.wikipedia.org/wiki/Head-of-line_blocking
 		go func(conn net.Conn) {
 			c, ni, err := upgrade(
 				c,
@@ -201,7 +216,9 @@ func (mt *multiplexTransport) wrapPeer(
 
 	return &multiplexPeer{
 		onStop: func() {
+			fmt.Println("onStop")
 			delete(mt.peers, c)
+			_ = c.Close()
 		},
 		peer: newPeer(
 			pc,
@@ -283,12 +300,12 @@ func upgrade(
 
 	c, err := secretConn(conn, timeout, privKey)
 	if err != nil {
-		return nil, NodeInfo{}, err
+		return nil, NodeInfo{}, fmt.Errorf("secrect conn failed: %v", err)
 	}
 
 	ni, err = handshake(c, timeout, nodeInfo)
 	if err != nil {
-		return nil, NodeInfo{}, err
+		return nil, NodeInfo{}, fmt.Errorf("handshake failed: %v", err)
 	}
 
 	return c, ni, nil
@@ -296,7 +313,7 @@ func upgrade(
 
 // multiplexPeer is the Peer implementation returned by the multiplexTransport
 // and wraps the default peer implementation for now, with an added hook for
-// shutdown of the a peer. It should ultimately grow into the proper
+// shutdown of the peer. It should ultimately grow into the proper
 // implementation.
 type multiplexPeer struct {
 	*peer
--- a/p2p/transport_test.go
+++ b/p2p/transport_test.go
@@ -1,0 +1,121 @@
+package p2p
+
+import (
+	"net"
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/tendermint/tendermint/crypto/ed25519"
+)
+
+func TestTransportAcceptDial(t *testing.T) {
+	mt := NewMultiplexTransport(NodeKey{
+		PrivKey: ed25519.GenPrivKey(),
+	})
+
+	addr, err := NewNetAddressStringWithOptionalID("127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = mt.Listen(*addr)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	errc := make(chan error)
+
+	go func() {
+		dialer := NewMultiplexTransport(NodeKey{
+			PrivKey: ed25519.GenPrivKey(),
+		})
+
+		addr, err := NewNetAddressStringWithOptionalID(mt.listener.Addr().String())
+		if err != nil {
+			errc <- err
+			return
+		}
+
+		_, err = dialer.Dial(*addr, peerConfig{})
+		if err != nil {
+			errc <- err
+			return
+		}
+
+		close(errc)
+	}()
+
+	if err := <-errc; err != nil {
+		t.Errorf("connection failed: %v", err)
+	}
+
+	p, err := mt.Accept(peerConfig{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = p.Start()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = p.Stop()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if have, want := len(mt.peers), 0; have != want {
+		t.Errorf("have %v, want %v", have, want)
+	}
+}
+
+func TestTransportHandshake(t *testing.T) {
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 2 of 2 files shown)


p2p/switch.go
```diff
@@ -196,25 +180,21 @@ func (sw *Switch) OnStart() error {
...
-	// Stop listeners
-	for _, listener := range sw.listeners {
-		listener.Stop()
-	}
-	sw.listeners = nil
 	// Stop peers
-	for _, peer := range sw.peers.List() {
-		peer.Stop()
-		sw.peers.Remove(peer)
+	for _, p := range sw.peers.List() {
```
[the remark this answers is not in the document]
melekes: I'd personally say no if the loop / func body is less than 5-10 lines

p2p/errors.go
```diff
@@ -5,6 +5,85 @@ import (
 	"net"
 )
 
+// ErrRejected indicates that a Peer was rejected carrying additional
+// information as to the reason.
+type ErrRejected struct {
```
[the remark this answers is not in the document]
ebuchman: Ah, I think it might be because we want to easily check later on if an err is ErrRejected, rather than 6 possible types of ErrRejected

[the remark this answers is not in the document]
xla: Indeed, it's divorce cause from reason and have the caller deal with the complexity, but can when necessary. See how the switch handles rejected for self cases: https://github.com/tendermint/tendermint/blob/1c341f758988b340ad1d075cdd4475e2164ce8b5/p2p/switch.go#L451-L468

p2p/switch.go
```diff
@@ -196,25 +180,21 @@ func (sw *Switch) OnStart() error {
...
+	go sw.acceptRoutine()
+
 	return nil
 }
 
-// OnStop implements BaseService. It stops all listeners, peers, and reactors.
+// OnStop implements BaseService. It stops all peers and reactors.
 func (sw *Switch) OnStop() {
-	// Stop listeners
-	for _, listener := range sw.listeners {
```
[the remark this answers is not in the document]
xla: Not having the Switch care about transport lifecycle maks it leaner, additionally some implementations of transport might not even need to be started/stopped in some for, for that reason I broke out the lifecycle methods into its own interface: https://github.com/tendermint/tendermint/blob/1c341f758988b340ad1d075cdd4475e2164ce8b5/p2p/transport.go#L58-L63

In our production stack the node code will be responsible for setup and teardown of the transport.

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+	outbound, persistent bool
+	reactorsByCh         map[byte]Reactor
+}
+
+// Transport emits and connects to Peers. The implementation of Peer is left to
+// the transport. Each transport is also responsible to filter establishing
+// peers specific to its domain.
+type Transport interface {
+	// Accept returns a newly connected Peer.
+	Accept(peerConfig) (Peer, error)
```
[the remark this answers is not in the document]
xla: This is temporrary, I will make it more clear in the comment on `peerConfig`. The reason it's still present is that `*peer` depends on `ch` and `reactor` info and the switch needs to have the error callback hooked up. It will all vanish when we make reactor setup less dynamic and have `PeerBehaviour` as concept esablished.

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+		// Reference:  https://github.com/tendermint/tendermint/issues/2047
+		//
+		// [0] https://en.wikipedia.org/wiki/Head-of-line_blocking
+		go func(c net.Conn) {
+			var (
+				ni NodeInfo
+				sc *conn.SecretConnection
+			)
+
+			err := mt.filterConn(c)
```
ebuchman: Does filterConn need to be async? These should be quick local checks (granted could be over ABCI, but that should be quick too).

Also, can/should a conn filter closure do the maxPeers check (currently happening in the switch)?

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+	for _, f := range mt.peerFilters {
+		errc := make(chan error)
+
+		go func(f PeerFilterFunc, p Peer, errc chan<- error) {
+			errc <- f(mt.peers, p)
+		}(f, p, errc)
+
+		select {
+		case err := <-errc:
+			return &ErrRejected{err: err, id: p.ID(), isFiltered: true}
```
ebuchman: same, need to check `err != nil`

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+
+	return c.Close()
+}
+
+func (mt *MultiplexTransport) filterConn(c net.Conn) error {
+	// We acquire a single read lock for all filters.
+	mt.connsMux.RLock()
+
+	if _, ok := mt.conns[c.RemoteAddr().String()]; ok {
+		return &ErrRejected{conn: c, isDuplicate: true}
```
ebuchman: these returns don't unlock the mutex.

we should break out the "read" section of this into its own function and use `defer`

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+
+	return nil
+}
+
+func (mt *MultiplexTransport) filterPeer(p Peer) error {
+	// We acquire a single read lock for all filters.
+	mt.peersMux.RLock()
+
+	if _, ok := mt.peers[p.ID()]; ok {
+		return &ErrRejected{id: p.ID(), isDuplicate: true}
```
ebuchman: same - peersMux doesn't unlock on these errors

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+
+// MultiplexTransport accepts and dials tcp connections and upgrades them to
+// multiplexed peers.
+type MultiplexTransport struct {
+	listener net.Listener
+
+	acceptc chan accept
+	closec  chan struct{}
+
+	// Lookup table for duplicate ip and id checks.
```
ebuchman: can't we just use the existing PeerSet for this purpose?

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+func MultiplexTransportPeerFilters(
+	filters ...PeerFilterFunc,
+) MultiplexTransportOption {
+	return func(mt *MultiplexTransport) { mt.peerFilters = filters }
+}
+
+// MultiplexTransport accepts and dials tcp connections and upgrades them to
+// multiplexed peers.
+type MultiplexTransport struct {
+	listener net.Listener
```
ebuchman: noting that we likely want to support listening on multiple addresses, as we currently do. Not sure if that should be multiple listeners in the transport or multiple transports in the switch. An example would be running both an IPv4 and IPv6 listener

p2p/switch.go
```diff
@@ -551,131 +519,82 @@ func (sw *Switch) addOutboundPeerWithConfig(
...
+func (sw *Switch) addPeer(p Peer) error {
 	// Avoid duplicate
-	if sw.peers.Has(peerID) {
-		return ErrSwitchDuplicatePeerID{peerID}
+	if sw.peers.Has(p.ID()) {
+		return ErrSwitchDuplicatePeerID{p.ID()}
 	}
 
-	// Check for duplicate connection or peer info IP.
-	if !sw.config.AllowDuplicateIP &&
```
ebuchman: noting that we may want to keep the ability to allow duplicate ips, eg. for local testing scenarios (though this probably better solved through a "dev" mode for tendermint than this specific config option ...) 

Also noting/asking we probably want to enforce the duplicate IP check on both the connection IP and the nodeInfo.IP

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+		}
+
+		for _, knownIPs := range im {
+			for _, known := range knownIPs {
+				for _, new := range newIPs {
+					if new.Equal(known) {
+						return &ErrRejected{
+							conn:        c,
+							err:         fmt.Errorf("IP<%v> already connected", new),
+							isDuplicate: true,
```
ebuchman: doesn't really seem right for filters to be expected to return ErrRejected if we want to support filters implemented outside this package, they won't have access to `is*` variables

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+	// with sane defaults.
+	mConfig   conn.MConnConfig
+	p2pConfig config.P2PConfig
+}
+
+// Test multiplexTransport for interface completeness.
+var _ Transport = (*MultiplexTransport)(nil)
+var _ transportLifecycle = (*MultiplexTransport)(nil)
+
+// NewMultiplexTransport returns a tcp conencted multiplexed peers.
```
ebuchman: `connected multiplexed peer`

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+	// Lookup table for duplicate ip and id checks.
+	conns    map[string]net.Conn
+	connsMux sync.RWMutex
+	peers    map[ID]Peer
+	peersMux sync.RWMutex
+
+	dialTimeout      time.Duration
+	filterTimeout    time.Duration
+	handshakeTimeout time.Duration
+	nodeAddr         NetAddress
```
ebuchman: do we need the `nodeAddr`?

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+	nodeKey          NodeKey
+
+	connFilters []ConnFilterFunc
+	peerFilters []PeerFilterFunc
+
+	// TODO(xla): Those configs are still needed as we parameterise peerConn and
+	// peer currently. All relevant configuration should be refactored into options
+	// with sane defaults.
+	mConfig   conn.MConnConfig
+	p2pConfig config.P2PConfig
```
ebuchman: don't seem to be setting/using this `p2pConfig`

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+func (mt *MultiplexTransport) Dial(
+	addr NetAddress,
+	cfg peerConfig,
+) (Peer, error) {
+	c, err := addr.DialTimeout(mt.dialTimeout)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(xla): Evaluate if we should apply filters if we explicitly dial.
```
ebuchman: Good question. We probably want to prevent ourselves from dialing duplicates as much as possible in the first place, but at least for now it might happen if mechanisms like PEX and persistentPeers are concurrently trying to dial the same peers

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+	go mt.acceptPeers()
+
+	return nil
+}
+
+func (mt *MultiplexTransport) acceptPeers() {
+	for {
+		c, err := mt.listener.Accept()
+		if err != nil {
+			// Close has been called so we silently exit.
```
ebuchman: "If Close has been called, silently exit"

p2p/switch.go
```diff
@@ -459,85 +439,73 @@ func (sw *Switch) IsDialingOrExistingAddress(addr *NetAddress) bool {
...
+func (sw *Switch) acceptRoutine() {
+	for {
+		p, err := sw.transport.Accept(peerConfig{
+			chDescs:      sw.chDescs,
+			onPeerError:  sw.StopPeerForError,
+			reactorsByCh: sw.reactorsByCh,
+		})
+		if err != nil {
+			switch err.(type) {
+			case *ErrRejected:
```
melekes: Is there a particular reason why these errors are pointers?

p2p/switch.go
```diff
@@ -459,85 +439,73 @@ func (sw *Switch) IsDialingOrExistingAddress(addr *NetAddress) bool {
...
+				rErr := err.(*ErrRejected)
```
melekes: wouldn't it be nicer to have a `IsErrRejected(err error) bool` function

Examples in std lib: os.IsNotExist, os.IsTimeout

```go
// IsTimeout returns a boolean indicating whether the error is known
  // to report that a timeout occurred.
  func IsTimeout(err error) bool {
  	terr, ok := underlyingError(err).(timeout)
  	return ok && terr.Timeout()
  }
  
  // underlyingError returns the underlying error for known os error types.
  func underlyingError(err error) error {
  	switch err := err.(type) {
  	case *PathError:
  		return err.Err
  	case *LinkError:
  		return err.Err
  	case *SyscallError:
  		return err.Err
  	}
  	return err
  }
```

p2p/errors.go
```diff
@@ -47,6 +126,12 @@ func (e ErrSwitchAuthenticationFailure) Error() string {
 	)
 }
 
+type ErrTransportClosed struct{}
```
melekes: godoc is missing

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(xla): Evaluate if we should apply filters if we explicitly dial.
+	if err := mt.filterConn(c); err != nil {
+		return nil, err
+	}
+
+	sc, ni, err := mt.upgrade(c)
```
melekes: what are `sc` and `ni` stand for?

p2p/errors.go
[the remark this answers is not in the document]
xla: Because then you'd need to assert for all errors that indicate rejected, unless I missunderstand your concern.

p2p/transport_test.go
```diff
@@ -0,0 +1,1016 @@
...
+
+func TestTransportMultiplexConnFilterTimeout(t *testing.T) {
+	mt := NewMultiplexTransport(
+		NodeInfo{},
+		NodeKey{
+			PrivKey: ed25519.GenPrivKey(),
+		},
+	)
+
+	MultiplexTransportFilterTimeout(5 * time.Millisecond)(mt)
```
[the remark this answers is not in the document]
xla: Your first option expands the surface the transport and the second will introduce a struct which will grow significantly with the options introduced and mixes different configurations. It takes the [functional option](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) (which seem to have better self-documenting character)  approach and instead of a variadic param has them as standalone calls, which are chainable. Either seems fine, I prefer the standalone version.

p2p/transport_test.go
```diff
@@ -0,0 +1,1016 @@
...
+		}
+	} else {
+		t.Errorf("expected ErrRejected")
+	}
+}
+
+func TestTransportMultiplexAcceptMultiple(t *testing.T) {
+	var (
+		pv = ed25519.GenPrivKey()
+		mt = NewMultiplexTransport(
```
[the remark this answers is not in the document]
xla: Which part? I'm hesitant as we have a lot of obfuscation in our tests in the project which doesn't necessarily help. I prefer a very verbose order of execution so the test is self-contained. @ebuchman pointed at a good resource regarding the [wrong abstraction](https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction) and I think this could be one of those, unless we have very strong arguments what the gains would be, besides a couple of lines spared.

p2p/transport_test.go
```diff
@@ -0,0 +1,1016 @@
...
+
+		close(errc)
+	}()
+
+	if err := <-errc; err != nil {
+		t.Errorf("connection failed: %v", err)
+	}
+
+	_, err = mt.Accept(peerConfig{})
+	if err, ok := err.(*ErrRejected); ok {
```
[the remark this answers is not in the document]
xla: Can you expand what you'd like to see addressed here?

p2p/switch.go
xla: No, good catch.

p2p/switch.go
```diff
@@ -459,85 +439,73 @@ func (sw *Switch) IsDialingOrExistingAddress(addr *NetAddress) bool {
...
-		inConn, ok := <-l.Connections()
-		if !ok {
 			break
 		}
 
 		// Ignore connection if we already have enough peers.
 		_, in, _ := sw.NumPeers()
 		if in >= sw.config.MaxNumInboundPeers {
 			sw.Logger.Info(
-				"Ignoring inbound connection: already have enough inbound peers",
```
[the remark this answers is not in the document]
xla: It's redundant as it is already mentioned in `Ignoring inbound connection`.

p2p/switch.go
xla: Those concerns have been moved to the Transport filters. We allow for duplicate IPs by default unless the filter has been passed as option to the transport.

Wit the addition of filters on transport we have now two layers one on raw network connection and one on incoming peers. The check for the nodeinfo.IP sounds like should be best handled as a peer filter.

p2p/transport.go
xla: If this is a desired property, I'd advocate for supporting multiple listeners in the transport. Is there strong evidence we want to support it, would like to reference requests for it when making the change.

p2p/transport.go
xla: Great suggestion, will investigate!

p2p/transport.go
xla: Depends on how we want to proceed with the multiple listeners, maybe it's not necessary to keep it around, if the Transport just expects a list of listeners.

p2p/transport.go
xla: It's necessary to set up the peer: https://github.com/tendermint/tendermint/blob/ffde3960131f99a07d555ac2c54c56a423a9998a/p2p/transport.go#L464-L474

p2p/transport.go
xla: `secretConn` and `nodeInfo` respectively.

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+			mt.acceptc <- accept{err: err}
+			return
+		}
+
+		// Connection upgrade and filtering should be asynchronous to avoid
+		// Head-of-line blocking[0].
+		// Reference:  https://github.com/tendermint/tendermint/issues/2047
+		//
+		// [0] https://en.wikipedia.org/wiki/Head-of-line_blocking
+		go func(c net.Conn) {
```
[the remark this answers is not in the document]
xla: We can either create a pool for the routines in the transport and have it push back here, or we limit the listeners by wrapping them in something like [LimitListener](https://godoc.org/golang.org/x/net/netutil#LimitListener). The latter gives up more control and depends on the caller passing in well-behaved listeners. If we assume this to be a serious attack vector the transport should guarantee that resource starvation is prevented.

p2p/transport.go
xla: > Does filterConn need to be part of this async routine? These should be quick local checks (granted could be over ABCI, but that should be quick too).

As the filters are handed from the outside and potentially are side-effectful (network) I advise to be as defensive as possible, as in executing them asynchronously and guard against long running operations.

> Also, can/should a conn filter closure do the maxPeers check (currently happening in the switch)?

Very valid point, the more the transport shapes up the more of the peer regulation seems to fit there. But currently the switch has still a lot of logic regarding peer managament, so I'd leave there for now.

p2p/transport.go
xla: Also this mutex madness can be prevented if we move to `PeerSet` and create a similar datatype for the conns, which are threadsafe. Think this is a good direction?

p2p/transport.go
xla: Currently we don't support multiple listeners unless I missed something: https://github.com/tendermint/tendermint/blob/d419fffe18531317c28c29a292ad7d253f6cafdf/node/node.go#L432-L438

p2p/switch.go
xla: We can introduce such a method, yet to get to the methods on the `ErrRejected` you have to assert anyway.

p2p/errors.go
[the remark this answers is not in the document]
ebuchman: the `IsX` methods check the booleans - instead they could do a type assertion on the `e.err` and return the `ok`

p2p/switch.go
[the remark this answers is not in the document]
melekes: But It changes the meaning. Correct me if I am wrong but "already have enough peers" is not the same as already "have enough inbound peers"

p2p/transport_test.go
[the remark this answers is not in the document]
melekes: You're creating the same transport, so I thought it can be extracted into `createMultiplexTransport` func.

```
func createMultiplexTransport(listenAddr string) {
pv = ed25519.GenPrivKey()
		mt = NewMultiplexTransport(
			NodeInfo{
				ID:         PubKeyToID(pv.PubKey()),
				ListenAddr: "127.0.0.1:0",
				Moniker:    "transport",
				Version:    "1.0.0",
			},
			NodeKey{
				PrivKey: pv,
			},
		)
}
```

p2p/transport_test.go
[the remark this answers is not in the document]
melekes: Sorry, I was mostly talking about https://github.com/tendermint/tendermint/pull/2126#discussion_r216231110

p2p/transport_test.go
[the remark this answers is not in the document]
xla: We can sure do that. My point is then in order to understand any given test-case you have to consult the implementation of `createMultiplexTransport`. Is the main motivation here to reduce repetition?

p2p/transport.go
xla: It doesn't seem to interfer if filters provided by the `p2p` package return specialised errors. Alternatively we can think of exposing a way for callers to create the right errors.

p2p/transport_test.go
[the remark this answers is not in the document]
melekes: > Is the main motivation here to reduce repetition?

yes

> My point is then in order to understand any given test-case you have to consult the implementation of createMultiplexTransport.

I understand, but that can be said about anything in programming. And that does not mean Tendermint should be 1 function 50000 lines long 😆

p2p/transport.go
melekes: such code is hard to read the first time. @ebuchman is right. we should avoid short names generally and use them only in loops and function no more than 10 lines of code where one can immediately get the context. i.e.

```
func GetTransportInfo() {
  t := getTransport() // it is obvious that getTransport returns transport
  doSomething(t)
}
```
if GetTransportInfo is more than 10 lines of code, we should use `transport := getTransport()`

```
sc, ni, err := mt.upgrade(c) // I have to consult with upgrade docs in order to understand what ni is
secretConn, nodeInfo, err := mt.upgrade(conn) // better
```

p2p/transport.go
```diff
@@ -0,0 +1,579 @@
...
+	for _, f := range mt.connFilters {
+		errc := make(chan error)
+
+		go func(f ConnFilterFunc, c net.Conn, errc chan<- error) {
+			errc <- f(mt.conns, c)
+		}(f, c, errc)
+
+		select {
+		case err := <-errc:
+			return &ErrRejected{conn: c, err: err, isFiltered: true}
```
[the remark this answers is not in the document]
ebuchman: Think we still need to fix this

p2p/transport.go
ebuchman: We should add this to https://github.com/tendermint/coding/blob/master/go/coding_standard.md

p2p/switch.go
```diff
@@ -109,6 +119,16 @@ func NewSwitch(cfg *config.P2PConfig, options ...SwitchOption) *Switch {
 	return sw
 }
 
+// SwitchFilterTimeout sets the timeout used for peer filters.
+func SwitchFilterTimeout(timeout time.Duration) SwitchOption {
+	return func(sw *Switch) { sw.filterTimeout = timeout }
+}
+
+// SwitchPeerFilters sets the filters for rejection of new peers.
+func SwitchPeerFilters(filters ...PeerFilterFunc) SwitchOption {
```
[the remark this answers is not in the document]
xla: We should standardize, I'm strongly against the current form, as it is ambigious, from a callers perspective it looks like this `p2p.WithMetrics` which would make one believe this is on the package level. Ideally it, even if more verbose, options hould have the `SwitchOption` prefix, so that other components in the same package can expose their options without danger of collision. Would that make sense? /cc @melekes

p2p/transport.go
```diff
@@ -0,0 +1,520 @@
...
+	cfg peerConfig,
+) (Peer, error) {
+	c, err := addr.DialTimeout(mt.dialTimeout)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(xla): Evaluate if we should apply filters if we explicitly dial.
+	if err := mt.filterConn(c); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
xla: Addressed in cleanup in `filterConn`: 2d8e2e6

rpc/core/pipe.go
```diff
@@ -34,13 +34,16 @@ type Consensus interface {
...
-	NumPeers() (outbound, inbound, dialig int)
-	NodeInfo() p2p.NodeInfo
+type transport interface {
+	Listeners() []string
 	IsListening() bool
+	NodeInfo() p2p.NodeInfo
+}
+
+type peers interface {
 	DialPeersAsync(p2p.AddrBook, []string, bool) error
```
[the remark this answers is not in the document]
xla: Ideally it should be on the pex which informs the Switch of what to dial and what point, currently it is still a method on the Switch and needs to be reworked. Would say it's outside of the scope of this changeset, going to leave a TODO on the Switch.

node/node.go
```diff
@@ -296,14 +299,109 @@ func NewNode(config *cfg.Config,
...
+		peerFilters = []p2p.PeerFilterFunc{}
+	)
+
+	if !config.P2P.AllowDuplicateIP {
+		connFilters = append(connFilters, p2p.ConnDuplicateIPFilter())
+	}
+
+	// Filter peers by addr or pubkey with an ABCI query.
+	// If the query return code is OK, add peer.
+	// XXX: Query format subject to change
```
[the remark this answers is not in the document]
xla: Anything we can do about it in the context of this PR?

node/node.go
```diff
@@ -671,22 +715,44 @@ func (n *Node) ProxyApp() proxy.AppConns {
 	return n.proxyApp
 }
 
-func (n *Node) makeNodeInfo(nodeID p2p.ID) p2p.NodeInfo {
+//------------------------------------------------------------------------------
+
+func (n *Node) Listeners() []string {
```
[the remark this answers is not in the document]
xla: The main reason to move those onto `node` is that it manages the lifecycle of the transport and it does not appear why we should be forced to extend the `Transport` interface to accomodate for this information when the node is well aware of it and glues those two pieces together anyway. What way forward would you propose to satisfy the need for status information in the rpc?

p2p/switch_test.go
```diff
@@ -151,35 +148,6 @@ func assertMsgReceivedWithTimeout(t *testing.T, msgBytes []byte, channel byte, r
 	}
 }
 
-func TestConnAddrFilter(t *testing.T) {
```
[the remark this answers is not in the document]
xla: Connection filtering is tested in transport_test.go

p2p/conn_set.go
```diff
@@ -0,0 +1,86 @@
+package p2p
+
+import (
+	"net"
+	"sync"
+)
+
+// ConnSet is a lookup table for connections and all their ips.
+type ConnSet interface {
+	Get(string) net.Conn
```
melekes: would be good to add argument name or godoc, so it's clear that one should pass `addr`

melekes: also, Get and Set methods are asymmetric. Should probably be `Get(addr string)` and `Set(addr string, net.Conn, []net.IP)`. Alternatively, you can rename `Get` to `GetByAddr(string)`

p2p/switch.go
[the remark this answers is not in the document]
melekes: If there are multiple publicly exposed structs within a package, it makes sense to prefix options. SwitchPeerFilters looks OK to me

p2p/switch.go
```diff
@@ -547,135 +533,119 @@ func (sw *Switch) addInboundPeerWithConfig(
...
-		return err
-	}
-
-	// Exchange NodeInfo on the conn
-	peerNodeInfo, err := pc.HandshakeTimeout(sw.nodeInfo, time.Duration(sw.config.HandshakeTimeout))
-	if err != nil {
-		return err
+func (sw *Switch) filterPeer(p Peer) error {
+	// Avoid duplicate
+	if sw.peers.Has(p.ID()) {
```
melekes: Should we also check IP?

node/node.go
```diff
@@ -696,34 +762,29 @@ func (n *Node) makeNodeInfo(nodeID p2p.ID) p2p.NodeInfo {
...
+	lAddr := config.P2P.ExternalAddress
 
-	if !n.sw.IsListening() {
-		return nodeInfo
+	if lAddr == "" {
+		lAddr = config.P2P.ListenAddress
 	}
 
-	p2pListener := n.sw.Listeners()[0]
-	p2pHost := p2pListener.ExternalAddressHost()
```
melekes: Does the new version work with IPv6?

node/node.go
```diff
@@ -462,6 +495,17 @@ func (n *Node) OnStart() error {
...
+	// Start the transport.
+	addr, err := p2p.NewNetAddressStringWithOptionalID(n.config.P2P.ListenAddress)
+	if err != nil {
+		return err
+	}
+	if err := n.transport.Listen(*addr); err != nil {
+		return err
+	}
+
+	n.isListening = true
```
melekes: Why do we need this? For RPC status?

p2p/transport.go
melekes: https://github.com/tendermint/tendermint/issues/2317

p2p/transport.go
```diff
@@ -0,0 +1,488 @@
...
+
+	return nil
+}
+
+func (mt *MultiplexTransport) acceptPeers() {
+	for {
+		c, err := mt.listener.Accept()
+		if err != nil {
+			// If Close() has been called, silently exit.
+			if _, ok := <-mt.closec; !ok {
```
melekes: won't this block on reading?

p2p/switch_test.go
```diff
@@ -213,41 +187,6 @@ func assertNoPeersAfterTimeout(t *testing.T, sw *Switch, timeout time.Duration)
 	}
 }
 
-func TestConnIDFilter(t *testing.T) {
```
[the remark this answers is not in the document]
xla: Revived in 9aa19e1

p2p/transport.go
```diff
@@ -0,0 +1,488 @@
...
+		mt.mConfig,
+		ni,
+		cfg.reactorsByCh,
+		cfg.chDescs,
+		cfg.onPeerError,
+	)
+
+	// Wait for Peer to Stop so we can cleanup.
+	go func(c net.Conn) {
+		select {
```
melekes: you don't need select if there's just one blocking case.

xla: facepalm

node/node.go
xla: Indeed, rpc expects to call `IsLIstening` on something for the status endpoint

node/node.go
xla: Could you be more specific which ipv6 compatibility you refer to?

p2p/conn_set.go
xla: Removed the `Get` method completely as it is not used.

p2p/transport.go
xla: @melekes Regarding #2317, I see this as an improvement in our configuration and reported information in the NodeInfo we exchange, external addresses are not basis for `bind` calls on the local machine of the node. It needs to be addressed, but this changeset isn't the right context for that.

p2p/transport.go
xla: Yeah found this as well, going to fix.

node/node.go
melekes: https://github.com/tendermint/tendermint/pull/1818

mdyring: Related to slack chat, we've just transitioned to a IPv6-only SNA. Please ensure IPv6 is supported moving forward. 

TIA :-)

xla: @mdyring Rest assured we follow up to harden those concerns, Thanks for staying on top of it.

PR #2128, opened 2018-08-02
crypto: Remove interface from crypto.Signature

ValarDragon:
Signatures are now []byte, which saves on the number of bytes after
amino encoding. This is 4 - 5 less bytes in all signatures :)

Ref #2121. All that remains for that ADR's implementation is forking btced. (Which will definitely be done in a separate PR)

* [ ] Updated all relevant documentation in docs - any spec / docs I should update? I searched for `crypto.Signature` and replaced all instances of it.
* [X] Wrote tests
* [X] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -8,6 +8,7 @@ BREAKING CHANGES:
 - [p2p] Remove salsa and ripemd primitives, in favor of using chacha as a stream cipher, and hkdf
 - [abci] Changed time format from int64 to google.protobuf.Timestamp
 - [abci] Changed Validators to LastCommitInfo in RequestBeginBlock
+- [crypto] Switch crypto.Signature from interface to []byte for space efficiency
 
 FEATURES:
 - [tools] Added `make check_dep`
--- a/crypto/ed25519/ed25519.go
+++ b/crypto/ed25519/ed25519.go
@@ -11,7 +11,6 @@ import (
 	amino "github.com/tendermint/go-amino"
 	"github.com/tendermint/tendermint/crypto"
 	"github.com/tendermint/tendermint/crypto/tmhash"
-	cmn "github.com/tendermint/tendermint/libs/common"
 )
 
 //-------------------------------------
@@ -19,9 +18,11 @@ import (
 var _ crypto.PrivKey = PrivKeyEd25519{}
 
 const (
-	Ed25519PrivKeyAminoRoute   = "tendermint/PrivKeyEd25519"
-	Ed25519PubKeyAminoRoute    = "tendermint/PubKeyEd25519"
-	Ed25519SignatureAminoRoute = "tendermint/SignatureEd25519"
+	Ed25519PrivKeyAminoRoute = "tendermint/PrivKeyEd25519"
+	Ed25519PubKeyAminoRoute  = "tendermint/PubKeyEd25519"
+	// Size of an Edwards25519 signature. Namely the size of a compressed
+	// Edwards25519 point, and a field element. Both of which are 32 bytes.
+	SignatureEd25519Size = 64
 )
 
 var cdc = amino.NewCodec()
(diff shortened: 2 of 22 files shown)


crypto/ed25519/ed25519.go
```diff
@@ -159,15 +156,15 @@ func (pubKey PubKeyEd25519) Bytes() []byte {
...
-func (pubKey PubKeyEd25519) VerifyBytes(msg []byte, sig_ crypto.Signature) bool {
+func (pubKey PubKeyEd25519) VerifyBytes(msg []byte, sig_ []byte) bool {
 	// make sure we use the same algorithm to sign
-	sig, ok := sig_.(SignatureEd25519)
-	if !ok {
+	if len(sig_) != SignatureEd25519Size {
 		return false
 	}
+	sig := new([SignatureEd25519Size]byte)
+	copy(sig[:], sig_)
```
melekes: why copy here?

ValarDragon: Is there a way to do this without copying? I need to go from `[]byte` to `[32]byte`

melekes: oh, right. no, copy is the only way

CHANGELOG_PENDING.md
```diff
@@ -8,6 +8,7 @@ BREAKING CHANGES:
 - [p2p] Remove salsa and ripemd primitives, in favor of using chacha as a stream cipher, and hkdf
 - [abci] Changed time format from int64 to google.protobuf.Timestamp
 - [abci] Changed Validators to LastCommitInfo in RequestBeginBlock
+- [crypto] Switch crypto.Signature from interface to []byte for space efficiency
```
Liamsi: Note to myself: investigate if this means we need to be update the kms code as well (doesn't seem so as we already just treat signatures as bytes there) /cc @zmanian

crypto/encoding/amino/encode_test.go
```diff
@@ -53,24 +55,24 @@ func ExamplePrintRegisteredTypes() {
...
 			privKey:  ed25519.GenPrivKey(),
 			privSize: 69,
 			pubSize:  37,
+			sigSize:  65,
 		},
 		{
 			privKey:  secp256k1.GenPrivKey(),
 			privSize: 37,
 			pubSize:  38,
+			sigSize:  71, // can also be 72
```
Liamsi: I'm confused by this comment: `// can also be 72`. Also, the `sigSize` field doesn't seem to be used/read in the test below?

crypto/encoding/amino/encode_test.go
```diff
@@ -84,13 +86,11 @@ func TestKeyEncodings(t *testing.T) {
...
 
 		// Check (de/en)codings of Signatures.
-		var sig1, sig2, sig3 crypto.Signature
+		var sig1, sig2 []byte
 		sig1, err := tc.privKey.Sign([]byte("something"))
 		assert.NoError(t, err, "tc #%d", tcIndex)
 		checkAminoBinary(t, sig1, &sig2, -1) // Signature size changes for Secp anyways.
 		assert.EqualValues(t, sig1, sig2, "tc #%d", tcIndex)
-		checkAminoJSON(t, sig1, &sig3, false) // TODO also check Prefix bytes.
-		assert.EqualValues(t, sig1, sig3, "tc #%d", tcIndex)
```
Liamsi: Question: why delete this check?

ValarDragon: Because this is now `[]byte`, which we already assume amino can handle

crypto/encoding/amino/encode_test.go
ValarDragon: This was for the purpose of having the sig size listed. Once we remove the DER encoding on secp, signatures will be constant size, and then I was planning on making this used.

Liamsi: If you don't mind, I would suggest to remove these and reintroduce when needed.

ebuchman:
Squashed and merged into https://github.com/tendermint/tendermint/pull/2152

PR #2129, opened 2018-08-02
p2p/pex: Allow configured seed nodes to not be resolvable over DNS

### LINKED ISSUE #1716: If you only have a seed, and its down, tendermint p2p crashes
I only had one seed in the config.toml, 
`seeds = "80a35a46ce09cfb31ee220c8141a25e73e0b239b@cosmos.cryptium.ch:46656"`

I ran `gaiad start`. The seed was down, and I got the following error message:
```
I[06-08|09:08:46.084] Starting AddrBook                            module=p2p book=/root/.gaiad/config/addrbook.json impl=AddrBook
ERROR: Error{`failed to start PEXReactor` (cause: Error looking up host (cosmos.cryptium.ch): lookup cosmos.cryptium.ch on 67.207.67.2:53: no such host)}
```
Apparently, tendermint shouldn't have crashed here, according to adrian.

ValarDragon:
Previously you couldn't startup tendermint if a seed node couldn't be resolved from DNS.
This now allows you to startup tendermint, as long as all seed node addresses
are formatted correctly. In the event that all seed nodes' IP's can't be resolved,
and the address book is empty, then it crashes with an informative error msg.
(This case doesn't occur if no seeds were specified)

Closes #1716

* [ ] Updated all relevant documentation in docs - I don't think theres anything in the docs that needs to be updated here?
* [ ] Wrote tests - Didn't write test for all seed nodes down + addr book empty, since I figured we didn't want to have outbound network connections in tests. (Only way I know to test "host not found" errors)
* [X] Updated CHANGELOG.md


p2p/pex/addrbook.go
```diff
@@ -223,6 +226,13 @@ func (a *addrBook) NeedMoreAddrs() bool {
 	return a.Size() < needAddressThreshold
 }
 
+// Empty implements AddrBook - returns true if there are no addresses in the address book.
+// Does not count the peer appearing in its own address book, or private peers.
+func (a *addrBook) Empty() bool {
+	// Size doesn't include self address or private address
```
melekes: duplicate of godoc above

p2p/pex/pex_reactor.go
```diff
@@ -121,10 +122,13 @@ func (r *PEXReactor) OnStart() error {
 	}
 
 	// return err if user provided a bad seed address
```
melekes: this should go into checkSeeds godoc

p2p/pex/pex_reactor.go
```diff
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/pkg/errors"
```
melekes:
```
go standard lib
\n
3rd party repos
\n
(repo) internal packages
```

p2p/pex/pex_reactor.go
```diff
@@ -7,12 +7,12 @@ import (
...
 	"time"
 
-	"github.com/pkg/errors"
 	amino "github.com/tendermint/go-amino"
 	cmn "github.com/tendermint/tendermint/libs/common"
-
 	"github.com/tendermint/tendermint/p2p"
 	"github.com/tendermint/tendermint/p2p/conn"
+
+	"github.com/pkg/errors"
```
melekes: ordering is wrong

```
go standard lib
\n
3rd party repos
\n
(repo) internal packages
```

p2p/pex/pex_reactor.go
```diff
@@ -120,12 +121,16 @@ func (r *PEXReactor) OnStart() error {
 		return err
 	}
 
-	// return err if user provided a bad seed address
-	// or a host name that we cant resolve
-	seedAddrs, err := r.checkSeeds()
+	// error if any of the seeds are formatted incorrectly.
```
melekes: excessive comment

p2p/pex/pex_reactor.go
```diff
@@ -120,12 +121,16 @@ func (r *PEXReactor) OnStart() error {
...
+	numOnline, seedAddrs, err := r.checkSeeds()
 	if err != nil {
 		return err
 	}
+	// error if seeds are in the config, they are all offline,
```
melekes: excessive comment. error below explains condition

p2p/pex/pex_reactor.go
```diff
@@ -120,11 +121,11 @@ func (r *PEXReactor) OnStart() error {
...
 	}
 
-	// return err if user provided a bad seed address
-	// or a host name that we cant resolve
-	seedAddrs, err := r.checkSeeds()
+	numOnline, seedAddrs, err := r.checkSeeds()
 	if err != nil {
 		return err
+	} else if numOnline == 0 && r.book.Empty() {
+		return errors.New("Address book is empty, and could not connect to any seed nodes")
```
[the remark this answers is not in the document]
ValarDragon: This behaviour makes sense to me. If no seed node was online, it may take awhile for an operator to notice. If the seed node wasn't reachable the first ping, you likely want to configure your seed node listing / make sure you're connected to the internet.

I guess we could try connecting to the seeds like 3 times, but I definitely don't think we should keep trying indefinitely long.

p2p/pex/pex_reactor.go
```diff
@@ -478,19 +479,27 @@ func (r *PEXReactor) dialPeer(addr *p2p.NetAddress) {
 	}
 }
 
-// check seed addresses are well formed
-func (r *PEXReactor) checkSeeds() ([]*p2p.NetAddress, error) {
+// checkSeeds checks that addresses are well formed.
+// Returns number of seeds we can connect to, along with all seeds addrs.
+// return err if user provided any badly formatted seed addresses.
+// Its fine if we can't resolve a host name that we cant resolve.
```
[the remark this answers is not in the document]
ValarDragon: Thanks for pointing that out, I'll fix this comment!

PR #2139, opened 2018-08-03
WIP: Simple merkle proof RFC compatibility

ValarDragon:
per request of @jaekwon at the libs meeting, indicate what files need to be changed and how 
to make this RFC compatible. This seems like it should be fairly
straightforward to switch. I didn't comment all instances of "Hasher",
since converting to byter is fairly straight forward

* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

### DIFF
--- a/crypto/merkle/simple_map.go
+++ b/crypto/merkle/simple_map.go
@@ -27,6 +27,7 @@ func (sm *simpleMap) Set(key string, value Hasher) {
 	// The value is hashed, so you can
 	// check for equality with a cached value (say)
 	// and make a determination to fetch or not.
+	// TODO: Change this to tmhash(0x00 || value.Bytes())
 	vhash := value.Hash()
 
 	sm.kvs = append(sm.kvs, cmn.KVPair{
@@ -66,6 +67,7 @@ func (sm *simpleMap) KVPairs() cmn.KVPairs {
 // then hashed.
 type KVPair cmn.KVPair
 
+// TODO: Change this to Bytes()
 func (kv KVPair) Hash() []byte {
 	hasher := tmhash.New()
 	err := encodeByteSlice(hasher, kv.Key)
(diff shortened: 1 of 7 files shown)


crypto/merkle/simple_map.go
```diff
@@ -27,6 +27,7 @@ func (sm *simpleMap) Set(key string, value Hasher) {
 	// The value is hashed, so you can
 	// check for equality with a cached value (say)
 	// and make a determination to fetch or not.
+	// TODO: Change this to tmhash(0x00 || value.Bytes())
```
Liamsi: Did we discuss to only prefix leaf nodes? According to the RFC in question, we would also prefix non-leaf nodes (with `0x01`): https://tools.ietf.org/html/rfc6962#section-2.1

crypto/merkle/simple_tree.go
```diff
@@ -51,6 +56,7 @@ func simpleHashFromHashes(hashes [][]byte) []byte {
 	case 1:
 		return hashes[0]
 	default:
+		// TODO: Change how we split the left and right nodes
```
Liamsi: Wasn't this the part @jaekwon was sceptical about? 

According to the [rfc][1] this would change to:

(n is the ordered number of inputs d(0), ..., d(n-1) )

> The hash of a list with one entry (also known as a leaf hash) is:
> MTH({d(0)}) = SHA-256(0x00 || d(0)).

>    For n > 1, let k be the largest power of two smaller than n (i.e.,
>   k < n <= 2k).  The Merkle Tree Hash of an n-element list D[n] is then
>   defined recursively as
>
>   MTH(D[n]) = SHA-256(0x01 || MTH(D[0:k]) || MTH(D[k:n])),
>
 >  where || is concatenation and D[k1:k2] denotes the list {d(k1),
 >  d(k1+1),..., d(k2-1)} of length (k2 - k1).  (Note that the hash
 >  calculations for leaves and nodes differ.  This domain separation is
 >  required to give second preimage resistance.)

This shouldn't be difficult to change, right?

[1]: https://tools.ietf.org/html/rfc6962#section-2.1

ValarDragon:
We discussed in the libs meeting that we want to go ahead with implementation.

PR #2140, opened 2018-08-03
crypto: Add compact bit array for intended usage in the multisig

ValarDragon:
This is in a separate PR from the rest of the multisig code for ease of review.

* [ ] Updated all relevant documentation in docs - will write as part of a later multisig spec
* [X] Wrote tests
* [ ] Updated CHANGELOG.md - will update as part of a later multisig changelog update

### DIFF
--- a/crypto/multisig/compact_bit_array.go
+++ b/crypto/multisig/compact_bit_array.go
@@ -37,6 +37,8 @@ func (bA *CompactBitArray) Size() int {
 	} else if bA.ExtraBitsStored == byte(0) {
 		return len(bA.Elems) * 8
 	}
+	// num_bits = 8*num_full_bytes + overflow_in_last_byte
+	// num_full_bytes = (len(bA.Elems)-1)
 	return (len(bA.Elems)-1)*8 + int(bA.ExtraBitsStored)
 }
 
--- a/crypto/multisig/compact_bit_array.go
+++ b/crypto/multisig/compact_bit_array.go
@@ -1,0 +1,218 @@
+package multisig
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// CompactBitArray is an implementation of a space efficient bit array.
+// This is used to ensure that the encoded data takes up a minimal amount of
+// space after amino encoding.
+// This is not thread safe, and is not intended for concurrent usage.
+type CompactBitArray struct {
+	ExtraBitsStored byte   `json:"extra_bits"` // The number of extra bits in elems.
+	Elems           []byte `json:"bits"`
+}
+
+// NewCompactBitArray returns a new compact bit array.
+// It returns nil if the number of bits is zero.
+func NewCompactBitArray(bits int) *CompactBitArray {
+	if bits <= 0 {
+		return nil
+	}
+	return &CompactBitArray{
+		ExtraBitsStored: byte(bits % 8),
+		Elems:           make([]byte, (bits+7)/8),
+	}
+}
+
+// Size returns the number of bits in the bitarray
+func (bA *CompactBitArray) Size() int {
+	if bA == nil {
+		return 0
+	} else if bA.ExtraBitsStored == byte(0) {
+		return len(bA.Elems) * 8
+	}
+	return (len(bA.Elems)-1)*8 + int(bA.ExtraBitsStored)
+}
+
+// GetIndex returns the bit at index i within the bit array.
+// The behavior is undefined if i >= bA.Size()
+func (bA *CompactBitArray) GetIndex(i int) bool {
+	if bA == nil {
+		return false
+	}
+	if i >= bA.Size() {
+		return false
+	}
+	return bA.Elems[i>>3]&(uint8(1)<<uint8(7-(i%8))) > 0
+}
+
+// SetIndex sets the bit at index i within the bit array.
+// The behavior is undefined if i >= bA.Size()
+func (bA *CompactBitArray) SetIndex(i int, v bool) bool {
+	if bA == nil {
+		return false
+	}
+	if i >= bA.Size() {
+		return false
+	}
+	if v {
+		bA.Elems[i>>3] |= (uint8(1) << uint8(7-(i%8)))
+	} else {
+		bA.Elems[i>>3] &= ^(uint8(1) << uint8(7-(i%8)))
+	}
+	return true
+}
+
+// Copy returns a copy of the provided bit array.
+func (bA *CompactBitArray) Copy() *CompactBitArray {
+	if bA == nil {
+		return nil
+	}
+	c := make([]byte, len(bA.Elems))
+	copy(c, bA.Elems)
+	return &CompactBitArray{
+		ExtraBitsStored: bA.ExtraBitsStored,
+		Elems:           c,
+	}
+}
+
+// String returns a string representation of CompactBitArray: BA{<bit-string>},
+// where <bit-string> is a sequence of 'x' (1) and '_' (0).
+// The <bit-string> includes spaces and newlines to help people.
+// For a simple sequence of 'x' and '_' characters with no spaces or newlines,
+// see the MarshalJSON() method.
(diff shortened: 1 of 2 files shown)


crypto/multisig/compact_bit_array.go
```diff
@@ -0,0 +1,218 @@
...
+// For a simple sequence of 'x' and '_' characters with no spaces or newlines,
+// see the MarshalJSON() method.
+// Example: "BA{_x_}" or "nil-BitArray" for nil.
+func (bA *CompactBitArray) String() string {
+	return bA.StringIndented("")
+}
+
+// StringIndented returns the same thing as String(), but applies the indent
+// at every 10th bit, and twice at every 50th bit.
+func (bA *CompactBitArray) StringIndented(indent string) string {
```
Liamsi: Nit: it would be cool to have this tested, too

crypto/multisig/compact_bit_array.go
```diff
@@ -0,0 +1,218 @@
...
+	if bA == nil {
+		return 0
+	} else if bA.ExtraBitsStored == byte(0) {
+		return len(bA.Elems) * 8
+	}
+	return (len(bA.Elems)-1)*8 + int(bA.ExtraBitsStored)
+}
+
+// GetIndex returns the bit at index i within the bit array.
+// The behavior is undefined if i >= bA.Size()
```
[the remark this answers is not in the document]
ValarDragon: I'm not sure. I was using the same functionality as the normal bit array. I think its easier if we say its undefined, and then we don't have to worry about the functionality there / testing those cases.

crypto/multisig/compact_bit_array.go
ValarDragon: This is tested through the JSON code.

crypto/multisig/compact_bit_array.go
```diff
@@ -0,0 +1,218 @@
...
+	*bA = *bA2
+	return nil
+}
+
+// CompactMarshal is a space efficient encoding for CompactBitArray.
+// It is not amino compatible.
+func (bA *CompactBitArray) CompactMarshal() []byte {
+	size := bA.Size()
+	if size <= 0 {
+		return []byte("null")
```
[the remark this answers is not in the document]
ValarDragon: I was just using the same case as the normal bit array. Encoding as empty seems like it would be confusing. (If this is null, something has gone unexpectedly wrong)

crypto/multisig/compact_bit_array.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// Size returns the number of bits in the bitarray
+func (bA *CompactBitArray) Size() int {
+	if bA == nil {
+		return 0
+	} else if bA.ExtraBitsStored == byte(0) {
+		return len(bA.Elems) * 8
+	}
+	return (len(bA.Elems)-1)*8 + int(bA.ExtraBitsStored)
```
melekes: why `-1`?

ValarDragon: `num_bits = 8*num_full_bytes + overflow_in_last_byte`, I'll add a comment for this.

crypto/multisig/compact_bit_array.go
```diff
@@ -0,0 +1,220 @@
...
+	"strings"
+)
+
+// CompactBitArray is an implementation of a space efficient bit array.
+// This is used to ensure that the encoded data takes up a minimal amount of
+// space after amino encoding.
+// This is not thread safe, and is not intended for concurrent usage.
+type CompactBitArray struct {
+	ExtraBitsStored byte   `json:"extra_bits"` // The number of extra bits in elems.
+	Elems           []byte `json:"bits"`
```
ebuchman: can we align the JSON and field names?

crypto/multisig/compact_bit_array.go
```diff
@@ -0,0 +1,220 @@
...
+// GetIndex returns the bit at index i within the bit array.
+// The behavior is undefined if i >= bA.Size()
+func (bA *CompactBitArray) GetIndex(i int) bool {
+	if bA == nil {
+		return false
+	}
+	if i >= bA.Size() {
+		return false
+	}
+	return bA.Elems[i>>3]&(uint8(1)<<uint8(7-(i%8))) > 0
```
ebuchman: So this counts bits from left to right. But the libs BitArray counts from right to left. Is that right? Is that because it just makes more sense to humans or is there another reason?

crypto/multisig/compact_bit_array.go
```diff
@@ -0,0 +1,220 @@
...
+			bA2.SetIndex(i, true)
+		}
+	}
+	*bA = *bA2
+	return nil
+}
+
+// CompactMarshal is a space efficient encoding for CompactBitArray.
+// It is not amino compatible.
+func (bA *CompactBitArray) CompactMarshal() []byte {
```
ebuchman: what is this for?

crypto/multisig/compact_bit_array_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	cmn "github.com/tendermint/tendermint/libs/common"
+)
+
+func randCompactBitArray(bits int) (*CompactBitArray, []byte) {
+	numBytes := (bits + 7) / 8
+	src := cmn.RandBytes((bits + 7) / 8)
+	bA := NewCompactBitArray(bits)
```
ebuchman: can't we just loop over bits, flip a coin, and Set if its true?

crypto/multisig/compact_bit_array_test.go
```diff
@@ -0,0 +1,169 @@
...
+	for i := uint8(0); i < 8-uint8(bA.ExtraBitsStored); i++ {
+		bA.SetIndex(numBytes*8+int(i), src[numBytes-1]&(uint8(1)<<(8-i)) > 0)
+	}
+	return bA, src
+}
+
+func TestNewBitArrayNeverCrashesOnNegatives(t *testing.T) {
+	bitList := []int{-127, -128, -1 << 31}
+	for _, bits := range bitList {
+		_ = NewCompactBitArray(bits)
```
ebuchman: might as well add the Nil check in here ...

crypto/multisig/compact_bit_array.go
ValarDragon: Sure!

crypto/multisig/compact_bit_array.go
ValarDragon: I think this just makes more sense. I didn't realize that the normal BitArray goes from right to left, I think that should be left to right as well. Theres no efficiency difference AFAICT

ebuchman:
I also don't understand what ExtraBitsStored is for actually (even from reading ADR). Can you help clarify?

crypto/multisig/compact_bit_array_test.go
ValarDragon: Sure, the intended use of this was that it'd be faster, but I guess theres no speed concern. (RandBool internally calls Int63)

ValarDragon:
Sure! In the ideal world, we would encode as `number_of_bits || elements`. However we can't do this with amino, as it length prefixes by bytes. (`compact_marshal` is made to do the above. I was going to suggest using `compact_marshal` later, as it will be more efficient.)

So to get the equivalent functionality with amino, there are two approaches. Either specify the total number of bits as a separate variable, or specify the number of bits in the last byte. The former is what is done by the libs bit array implementation. This ends up taking more space due to varint encoding. So instead here, I've opted to just list the number of bits in the last byte. (i.e. 0-8)

So the total size is `(len(bytes) - 1) * 8 + ExtraBitsStored`. I'll rename ExtraBitsStored to `NumBitsInFinalByte`

crypto/multisig/compact_bit_array.go
ValarDragon: Described in below reply :)

ebuchman:
I see. It's just for computing Size

ValarDragon:
Yup!

PR #2146, opened 2018-08-04
fix errors in Makefile

b00f:
* remove gogoproto from tools (NOTE; gogoproto is not an executable. look here: https://github.com/gogo/protobuf/blob/master/gogoproto/Makefile)
* fix make `get_check` and `get_vendor_deps` errors

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -15,7 +15,7 @@
   name = "github.com/btcsuite/btcd"
   packages = ["btcec"]
   pruneopts = "UT"
-  revision = "f5e261fc9ec3437697fb31d8b38453c293204b29"
+  revision = "cf05f92c3f815bbd5091ed6c73eff51f7b1945e8"
 
 [[projects]]
   digest = "1:1d8e1cb71c33a9470bbbae09bfec09db43c6bf358dfcae13cd8807c4e2a9a2bf"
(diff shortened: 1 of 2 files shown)


Gopkg.lock
```diff
@@ -15,7 +15,7 @@
   name = "github.com/btcsuite/btcd"
```
melekes: can you exclude the updates from this PR?

melekes:
@b00f thank you for checking the Makefile 👍

Gopkg.lock
b00f: what updates?

melekes: fix_errors_in_makefile_by_b00f_ _pull_request__2146_ _tendermint_tendermint_

b00f: I had error on running `dep ensure`. Removing `Gopkg.lock` and running the `dep ensure` fixed the errors.

ebuchman:
Fixed in https://github.com/tendermint/tendermint/pull/2152/commits/c5c2b9601f02747d50e70f54a71f40270a659956, thanks

b00f:
@ebuchman I still have same errors. Could you please check this command: `make check`
There is no `gogoproto` as executable in PATH.
Please check here: https://github.com/gogo/protobuf/blob/master/gogoproto/Makefile

melekes:
@b00f sorry, fixed in https://github.com/tendermint/tendermint/pull/2198

Issue #2015, opened 2018-07-19
types: allow genesis file to have 0 validators


ebuchman:
InitChain allows the app to return the initial validator set. This is nice because it allows the initial validator set to be a function of application state, for instance some set of bonds/delegations. 

However, Tendermint currently requires the genesis to have at least one validator. We should loosen this requirement, but then ensure that the validator set after InitChain has at least one validator ...

Note we create the state before calling InitChain and then update it after InitChain.

b00f:
@ebuchman I am going to work on this issue to resolve it.
But I have a question, why tendermint cannot work without validator? I mean in this case the node is just helping the eco-system by broadcasting transactions without validating proposals.

b00f:
I can't build the code. I got several errors:

``` 
$ make build
CGO_ENABLED=0 go build -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short=8 HEAD`" -tags 'tendermint' -o build/tendermint ./cmd/tendermint/
# github.com/tendermint/tendermint/abci/types
abci/types/types.pb.go:89:29: undefined: "github.com/tendermint/tendermint/vendor/github.com/gogo/protobuf/proto".InternalMessageInfo
abci/types/types.pb.go:511:33: undefined: "github.com/tendermint/tendermint/vendor/github.com/gogo/protobuf/proto".InternalMessageInfo
abci/types/types.pb.go:557:34: undefined: "github.com/tendermint/tendermint/vendor/github.com/gogo/protobuf/proto".InternalMessageInfo
abci/types/types.pb.go:597:33: undefined: "github.com/tendermint/tendermint/vendor/github.com/gogo/protobuf/proto".InternalMessageInfo
abci/types/types.pb.go:646:38: undefined: "github.com/tendermint/tendermint/vendor/github.com/gogo/protobuf/proto".InternalMessageInfo
abci/types/types.pb.go:704:38: undefined: "github.com/tendermint/tendermint/vendor/github.com/gogo/protobuf/proto".InternalMessageInfo
abci/types/types.pb.go:782:34: undefined: "github.com/tendermint/tendermint/vendor/github.com/gogo/protobuf/proto".InternalMessageInfo
abci/types/types.pb.go:854:39: undefined: "github.com/tendermint/tendermint/vendor/github.com/gogo/protobuf/proto".InternalMessageInfo
[...]

PR #2148, opened 2018-08-05
Issue 2015

b00f:
* [X ] Updated all code comments where relevant
* [ ] Wrote tests
* [ ] Updated CHANGELOG.md

Here is a fix for issue #2015 
Note: for testing this issue I used my local testnet with 4 nodes. you can find out a copy of it here: https://github.com/b00f/tm-testnet

I manually changed the genesis file in new node and removed all validators. Then I updated the BaseApplication (InitChain phase) for updating the state. It worked properly.

### DIFF
--- a/abci/types/application.go
+++ b/abci/types/application.go
@@ -63,26 +63,6 @@ func (BaseApplication) Query(req RequestQuery) ResponseQuery {
 
 func (BaseApplication) InitChain(req RequestInitChain) ResponseInitChain {
 	return ResponseInitChain{}
-	/*
-		This part is for code review. How I update validator list?
-		///
-		var vals []Validator
-
-		//test1:
-		//  3 vals, should fail
-		//err := json.Unmarshal([]byte("[{\"pub_key\":{\"type\":\"ed25519\",\"data\":\"slwynMSR1dfUG+gSniy9byVq9MipYOH3IGysvURwAIE=\"},\"power\":10,\"name\":\"validator_2\"},{\"pub_key\":{\"type\":\"ed25519\",\"data\":\"60azWyk4YzJF4Eiuyz9nh3KLQR1g8z5M+9t63bUXuiU=\"},\"power\":10,\"name\":\"validator_3\"},{\"pub_key\":{\"type\":\"ed25519\",\"data\":\"825sPf9cslugMqCQqzmuhZqVUVbVXpbd/ou8xqaS4HA=\"},\"power\":10,\"name\":\"validator_4\"}]"), &vals)
-
-		// tes2:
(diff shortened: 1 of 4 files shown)


types/genesis_test.go
```diff
@@ -14,14 +14,10 @@ import (
 func TestGenesisBad(t *testing.T) {
 	// test some bad ones from raw json
 	testCases := [][]byte{
-		[]byte{},                                           // empty
-		[]byte{1, 1, 1, 1, 1},                              // junk
-		[]byte(`{}`),                                       // empty
-		[]byte(`{"chain_id":"mychain"}`),                   // missing validators
-		[]byte(`{"chain_id":"mychain","validators":[]}`),   // missing validators
```
melekes: can we just move these test cases into TestGenesisGood? in order to test that they don't result in error

melekes: can we not remove these, but move them to TestGenesisGood instead?

abci/types/application.go
```diff
@@ -63,6 +63,26 @@ func (BaseApplication) Query(req RequestQuery) ResponseQuery {
 
 func (BaseApplication) InitChain(req RequestInitChain) ResponseInitChain {
 	return ResponseInitChain{}
+	/*
+		This part is for code review. How I update validator list?
```
melekes: Can you clarify your question? Do you want to update a running chain (https://tendermint.com/docs/app-dev/abci-spec.html#endblock)?

b00f: I am so sorry. I didn't mean to ask. It's not a question. I wanted to show that how I tested it... ignore the question mark! ;)

abci/types/application.go
```diff
@@ -63,6 +63,28 @@ func (BaseApplication) Query(req RequestQuery) ResponseQuery {
 
 func (BaseApplication) InitChain(req RequestInitChain) ResponseInitChain {
 	return ResponseInitChain{}
+	/*
```
melekes: can you remove this comment?

types/genesis_test.go
b00f: They supposed to be bad genesis json? shall I move them to `GoodGenesisTest`?
I added new test with wrong public key.

b00f: Done. Please check the latest commit.

melekes: Oh god. I think you misunderstood me. I wanted you to only move "missing validators" lines, not the whole `TestGenesisBad` test. The reason for moving is that "missing validators" are now correct use-case, so imho should be inside `TestGenesisGood`

b00f: Please check it now

PR #2149, opened 2018-08-05
[ADR] ABCI Validators

ebuchman:
Some minor changes to improve the usability of the ABCI, and to properly include the round number.

See markdown at: https://github.com/tendermint/tendermint/blob/279259ec8efb75d4dd2ebd5095c06892ff327804/docs/architecture/adr-018-ABCI-Validators.md

### DIFF
--- a/docs/architecture/adr-018-ABCI-Validators.md
+++ b/docs/architecture/adr-018-ABCI-Validators.md
@@ -1,0 +1,95 @@
+# ADR 018: ABCI Validator Improvements
+
+## Changelog
+
+05-08-2018: Initial draft
+
+## Context
+
+ADR 009 introduced major improvements to the ABCI around validators and the use
+of Amino. Here we follow up with some additional changes to improve the naming
+and expected use of Validator messages.
+
+We also fix how we communicate the commit round - there is no defined commit
+round, as validators can commit the same block in different rounds, so we
+should communicate the round each validator committed in.
+
+## Decision
+
+### Validator
+
+Currently a Validator contains address and `pub_key`, and one or the other is
+optional/not-sent depending on the use case. Instead, we should have a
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-018-ABCI-Validators.md
```diff
@@ -0,0 +1,95 @@
...
+
+```
+message Validator {
+    bytes address
+    int64 power
+}
+
+message ValidatorUpdate {
+    PubKey pub_key
+    int64 power
```
ValarDragon: If you are meant to refer to validators by address, why not pass in the address here as well. Then the semantics become more clear in my opinion. I.e. it becomes 

"ValidatorUpdate updates the validator at address addr to have the given pubkey and power".

ValarDragon: I agree that the validator struct should be unaltered. I was just suggesting adding an address field to validator updates. I think this makes it clearer. (You're right though, the intended method to keep track of it is pretty obvious though)

I am probably biased here though, since I think #postlaunch we should consider supporting other address derivation methods. (I.e. a global validator index)

ebuchman: I struggled with whether or not the address ought to be required and correct there, so opted for leaving it out. 

I think it could make sense to add it back, but if we do then it should be required and correct. For now that means Tendermint would fail if the address is empty or doesnt match the tmhash of the pubkey. In the future, perhaps it will have a different meaning.

As for BeginBlock, if we're not going to send the pubkey, we should use a message that doesn't have a pubkey

ValarDragon: Makes sense! Thanks for the explanation

Issue #2156, opened 2018-08-05
ABCI: Change SigningValidator to VoteInfo and add commit round


ebuchman:
SigningValidator is a bad name since it conflates validator info with voting info.

A better name is `VoteInfo`.

We should also remove the CommitRound from LastCommitInfo and instead add it to the VoteInfo, since we have a round for each vote in the commit. See the ADR in https://github.com/tendermint/tendermint/pull/2149 for details

ebuchman:
Done in #2159

PR #2159, opened 2018-08-06
Bucky/abci validators

ebuchman:
Ref #2156 

* [ ] Wrote tests


state/execution.go
```diff
@@ -241,17 +240,22 @@ func getBeginBlockValidatorInfo(block *types.Block, lastValSet *types.ValidatorS
...
+			if vote != nil {
+				commitRound = vote.Round
+			}
 		}
-		val := abci.SigningValidator{
-			Validator:       types.TM2PB.ValidatorWithoutPubKey(val),
-			SignedLastBlock: vote != nil,
+		voteInfo := abci.VoteInfo{
+			Validator:       types.TM2PB.Validator(val),
+			SignedLastBlock: vote != nil,        // XXX: should we replace with commitRound == -1 ?
```
melekes: why send an additional bool if `commit_round=-1` tells u the same thing

types/protobuf_test.go
```diff
@@ -104,32 +104,29 @@ func TestABCIEvidence(t *testing.T) {
...
+	assert.Panics(t, func() { TM2PB.NewValidatorUpdate(nil, 10) })
+	assert.Panics(t, func() { TM2PB.NewValidatorUpdate(pubKeyEddie{}, 10) })
 }
 
 func TestABCIValidatorWithoutPubKey(t *testing.T) {
 	pkEd := ed25519.GenPrivKey().PubKey()
 
-	abciVal := TM2PB.ValidatorWithoutPubKey(&Validator{
+	abciVal := TM2PB.Validator(&Validator{
 		Address:     pkEd.Address(),
```
melekes: can be omitted ...

melekes:
Also, ABCI spec https://tendermint.com/docs/app-dev/abci-spec.html needs to be updated.

state/execution.go
ebuchman: No good reason.

But I realized I was actually wrong. We do have a single commit round. Its true validators can see commits in different rounds subjectively but when a proposer selects a commit to include in the next block, all teh votes must come from the same round. So we should really revert this

types/protobuf_test.go
ebuchman: what can?

melekes: the address field

melekes: Right, sorry

melekes:
Linter failed because abci `types.pb.go` does not contain `//nolint`.

There's must be something wrong with:
```
	@awk '/^\s*package \w+/ { print "//nolint" }1' $@ > $@.tmp && mv $@.tmp $@
```

state/execution.go
```diff
@@ -218,15 +217,15 @@ func execBlockOnProxyApp(logger log.Logger, proxyAppConn proxy.AppConnConsensus,
 
 	logger.Info("Executed block", "height", block.Height, "validTxs", validTxs, "invalidTxs", invalidTxs)
 
-	valUpdates := abciResponses.EndBlock.ValidatorUpdates
-	if len(valUpdates) > 0 {
-		logger.Info("Updates to validators", "updates", abci.ValidatorsString(valUpdates))
+	if len(abciResponses.EndBlock.ValidatorUpdates) > 0 {
```
ebuchman: fix

abci/types/types.pb.go
```diff
@@ -1,7 +1,52 @@
 // Code generated by protoc-gen-gogo. DO NOT EDIT.
 // source: abci/types/types.proto
 
-//nolint
+/*
```
ebuchman: We need to lock these binary dependencies down or this stuff will keep happening.

types/protobuf_test.go
```diff
@@ -41,26 +41,26 @@ func TestABCIValidators(t *testing.T) {
...
+	abciVal = TM2PB.ValidatorUpdate(tmVal)
+	tmVals, err = PB2TM.ValidatorUpdates([]abci.ValidatorUpdate{abciVal})
 	assert.Nil(t, err)
 	assert.Equal(t, tmValExpected, tmVals[0])
 
-	// val with incorrect address
-	abciVal = TM2PB.Validator(tmVal)
-	abciVal.Address = []byte("incorrect!")
-	tmVals, err = PB2TM.Validators([]abci.Validator{abciVal})
+	// val with incorrect pubkey daya
```
ebuchman: data

abci/types/types.pb.go
ebuchman: https://github.com/tendermint/tendermint/issues/2238

ebuchman:
> Linter failed because 

Fixed in https://github.com/tendermint/tendermint/pull/2159/commits/e1ff0306fe5c7664d130c486cf96363058625551 :)

abci/types/types.proto
```diff
@@ -271,12 +272,18 @@ message PartSetHeader {
 // Validator
 message Validator {
   bytes address = 1;
-  PubKey pub_key = 2 [(gogoproto.nullable)=false];
+  //PubKey pub_key = 2 [(gogoproto.nullable)=false];
   int64 power = 3;
```
melekes: `3` -> `2`

docs/app-dev/abci-spec.md
```diff
@@ -111,14 +111,16 @@ See below for more details on the message types and how they are used.
...
+  - `Validators ([]ValidatorUpdate)`: Initial genesis validators.
   - `AppStateBytes ([]byte)`: Serialized initial application state. Amino-encoded JSON bytes.
 - **Response**:
   - `ConsensusParams (ConsensusParams)`: Initial
     consensus-critical parameters.
-  - `Validators ([]Validator)`: Initial validator set.
+  - `Validators ([]ValidatorUpdate)`: Initial validator set.
 - **Usage**:
   - Called once upon genesis.
+  - List of validators in the response will be used as the initial validator set
```
melekes: what if it (validators list in the response) is empty? or nil?

docs/spec/software/abci.md
```diff
@@ -52,14 +52,13 @@ objects in the `ResponseBeginBlock`:
 
 ```
 message Validator {
```
melekes: `ValidatorUpdate`. also need to add `Validator (address, power)` struct

types/protobuf.go
```diff
@@ -214,17 +211,13 @@ func (pb2tm) Validators(vals []abci.Validator) ([]*Validator, error) {
 		}
 		// If the app provided an address too, it must match.
 		// This is just a sanity check.
-		if len(v.Address) > 0 {
+		/*if len(v.Address) > 0 {
```
melekes: why not remove?

abci/types/types.proto
ebuchman: Left this on purpose since it makes it non-breaking and honestly wouldn't be too surprised if pubkeys made their way back !

docs/app-dev/abci-spec.md
ebuchman: Good catch will update

docs/spec/software/abci.md
ebuchman: This is just some convenience notes that we might be able to remove completely now that ABCI has been merged into Tendermint. Note it should all be pretty redundant with what's in the actual ABCI spec, but it's also replicated here for now. 

We could add notes on the BeginBlock stuff here, but not sure it's worth it since we probably want to deprecate this file in favour of the actual ABCI spec. 

Thoughts?

ebuchman:
@melekes comments addressed - thanks!

docs/spec/software/abci.md
melekes: 👍 for deprecating this stuff

Issue #1135, opened 2018-01-21
Add version to block header


ebuchman:
We have discussed adding a TendermintVersion and AppVersion to the block header. 

Perhaps we really only need one `Version string` field. 

Additionally, we discussed prefixing the serialized block with a single byte representing the version.

We should discuss the actual requirements and needs here and settle on something.

adrianbrink:
@rigelrozanski @sunnya97 and I just talked about it.

Putting the version number in the block makes a lot of sense, since it forces all validators to upgrade the software in a timely manner. Moreover it makes every change a hardfork which is a good thing.

@jessysaurusrex and I also talked about this as a way to incentivise validators to run the latest software.

rigelrozanski:
@sunnya97 also suggested that we include a... hash or piece of the hash in order to eliminate certain problems associated with hard forks? (what if you have two forks with same version and different hashes right)

sunnya97:
Yeah, what my suggestion was is that the chain id should include somehow the Tendermint version and app version like @ebuchman suggested, but also have it somehow incorporate the like hash of the genesis block or something.  (Btw, I define the genesis block as first block in chain that doesn't have a justification from the previous block).  That way even if the software versions don't change, but you need to fork the validator set, it still gets a new chain id.

b00f:
Can we merge this issue with #1855 ?

ebuchman:
Don't think so - we want to be explicit about this being version numbers.

b00f:
@ebuchman Let me know if you haven't assigned it to anyone. I did same for app_data here: https://github.com/tendermint/tendermint/pull/1953

b00f:
`TendermitVersion` and `AppVersion` of proposer?

PR #2160, opened 2018-08-06
[WIP] Versions

ebuchman:
Refs 
- https://github.com/tendermint/tendermint/issues/1135
- [ADR 016](https://github.com/tendermint/tendermint/blob/bucky/versions/docs/architecture/adr-016-protocol-versions.md)

* [ ] Wrote tests


ebuchman:
Started implementing the ADR but realized that we don't expose the result of the handshake anywhere, and that's where Tendermint will hear about the app version, so need to separate the handshake functionality out so it can be called explicitly and its return value utilized. 

Currently the handshaker is passed into the proxy.NewAppConns constructor and we never see it again. Instead, we should start the proxyApp such that it doesn't deal with the handshake, and we call the Handshake explicitly (passing in the proxyApp). This way we have direct access to the result of the handshake (eg. the app version) where we need it. This provides better separation between the proxy app and any messages/protocols carried out using it.

Made some progress towards that, but needs to be continued.

ebuchman:
I'm going to break this up into smaller PRs :)

Issue #1916, opened 2018-07-05
WebsocketEvents for ValidatorSet changes


zmanian:
We don't currently have any websocket events for validator set changes.

https://github.com/tendermint/tendermint/blob/c84be3b8ddf46040567897f7d8163d35aa561a25/types/events.go#L12-L32

It looked `EventBond` | `EventUnbond` | `EventRebond` where intended for this purpose but never implemented.

melekes:
Do we need all these unbond, rebond events? Can't we just emit `EventValidatorSetUpdates`?

zmanian:
I think 
```
type EventDataValidatorSetUpdates struct {
  updates []struct{
      prev *Validator
      cur *Validator
  }
}

melekes:
Released in 0.23.1

PR #2161, opened 2018-08-06
send ValidatorSetUpdates event when validator set changes

melekes:
Refs #1916

* [ ] Wrote tests

**REBASE AGAINST DEVELOP ONCE #2159 MERGED**

### DIFF
--- a/docs/app-dev/subscribing-to-events-via-websocket.md
+++ b/docs/app-dev/subscribing-to-events-via-websocket.md
@@ -26,3 +26,39 @@ more information on query syntax and other options.
 You can also use tags, given you had included them into DeliverTx
 response, to query transaction results. See [Indexing
 transactions](./indexing-transactions.md) for details.
+
+### ValidatorSetUpdates
+
+When validator set changes, ValidatorSetUpdates event is published. The
+event carries a list of pubkey/power pairs. The list is the same
+Tendermint receives from ABCI application (see [EndBlock
+section](https://tendermint.com/docs/app-dev/abci-spec.html#endblock) in
+the ABCI spec).
+
+Response:
+
+```
+{
+    "jsonrpc": "2.0",
+    "id": "0#event",
+    "result": {
+        "query": "tm.event='ValidatorSetUpdates'",
+        "data": {
+            "type": "tendermint/event/ValidatorSetUpdates",
+            "value": {
+              "validator_updates": [
+                {
+                  "address": "09EAD022FD25DE3A02E64B0FE9610B1417183EE4",
+                  "pub_key": {
+                    "type": "tendermint/PubKeyEd25519",
+                    "value": "ww0z4WaZ0Xg+YI10w43wTWbBmM3dpVza4mmSQYsd0ck="
+                  },
+                  "voting_power": "10",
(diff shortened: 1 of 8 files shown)


docs/app-dev/subscribing-to-events-via-websocket.md
```diff
@@ -26,3 +26,34 @@ more information on query syntax and other options.
...
+```
+{
+    "jsonrpc": "2.0",
+    "id": "0#event",
+    "result": {
+        "query": "tm.event='ValidatorSetUpdates'",
+        "data": {
+            "type": "tendermint/event/ValidatorSetUpdates",
+            "value": {
+              "validator_updates": [
```
fedekunze: @melekes does it contains only the new validators and validators that changed their `power` or the updated list with the new validator set ?

melekes: Neither. It's the list of updates. So if the current validator set is A, B, C and application includes D, then `ValidatorSetUpdates` will contain only D.

Is that not enough? Given you track validators from the beginning (or get them using RPC `/validators`), you should be able to construct the updated set.

fedekunze: And if the current validator set is A, B, C and and application includes D and drops A ? will `ValidatorSetUpdates` only include D as well ?

melekes: No. It will be `A (pubkeyA, power: 0), D (pubkeyD, power: X)`

ebuchman:
```
+ make metalinter
--> Running linter
abci/types/types.pb.go:6768:1:warning: randUnrecognizedTypes is unused (deadcode)
abci/types/types.pb.go:7637:1:warning: sozTypes is unused (deadcode)
Makefile:221: recipe for target 'metalinter' failed
make: *** [metalinter] Error 1
Exited with code 2
```

types/events.go
```diff
@@ -4,26 +4,28 @@ import (
...
+	EventNewRoundStep        = "NewRoundStep"
+	EventPolka               = "Polka"
+	EventProposalHeartbeat   = "ProposalHeartbeat"
+	EventRelock              = "Relock"
+	EventTimeoutPropose      = "TimeoutPropose"
+	EventTimeoutWait         = "TimeoutWait"
+	EventTx                  = "Tx"
+	EventUnlock              = "Unlock"
+	EventValidatorSetUpdates = "ValidatorSetUpdates"
+	EventVote                = "Vote"
```
ebuchman: maybe we should note in the comments that these are alphabetical for the poor soul that adds an event to this list without realizing :P

types/events.go
```diff
@@ -92,6 +86,10 @@ type EventDataVote struct {
 
 type EventDataString string
 
+type EventDataValidatorSetUpdates struct {
+	ValidatorUpdates []abci.ValidatorUpdate `json:"validator_updates"`
```
ebuchman: is it weird that we're going to be passing an ABCI validator over the RPC here ? ie. it will be different than the usual representation of pubkeys using AminoJSON

melekes:
Released in 0.23.1

Issue #2163, opened 2018-08-06
Implement multisig ADR


ValarDragon:
This issue is for tracking this in the current project board. See #1960 for proposal.

ValarDragon:
Ref https://github.com/tendermint/tendermint/pull/2164#pullrequestreview-145289993
What remains:
* For the threshold multisig, complete switching everything to table driven tests, with explicit checks on the recursive case. Adapt ADR to note the recursive case is allowed
* Implement weighted multisig
* Consider using compact bit array marshalling for marshalling the multisignature object.
* Make the compact bit array JSON elements and struct element names align.

alexanderbez:
Note, getting a `Pubkey` back from a bech32-ified string does not work in the Cosmos SDK because the bech32 string is too long (> 90 chars).

PR #2164, opened 2018-08-07
Threshold Multisignature implementation

ValarDragon:
This adds an implementation of the threshold multisig. I also merged develop in, for the change to the crypto.Signature interface. This is the only commit with new code changes: https://github.com/tendermint/tendermint/pull/2164/commits/e7dd76c28d9eb3ff9b30f0dd1303e25bd16bca9b. 

This may be easier to review if we merged the compact bitmap PR first, and then I aimed this PR at develop / remade this PR.

* [x] Wrote tests


crypto/multisig/threshold_multisig_test.go
```diff
@@ -45,15 +45,15 @@ func TestMultiSigPubkeyEquality(t *testing.T) {
...
+	require.True(t, multisigKey.Equals(unmarshalledMultisig))
 
 	// Ensure that reordering pubkeys is treated as a different pubkey
 	pubkeysCpy := make([]crypto.PubKey, 5)
 	copy(pubkeysCpy, pubkeys)
 	pubkeysCpy[4] = pubkeys[3]
 	pubkeysCpy[3] = pubkeys[4]
 	multisigKey2 := NewThresholdMultiSignaturePubKey(2, pubkeysCpy)
-	require.NotEqual(t, multisigKey, multisigKey2)
+	require.False(t, multisigKey.Equals(multisigKey2))
```
Liamsi: Interesting, I would also have expected this to call the underlying `Equal` instead

ValarDragon:
Absolutely! Let's merge the compact bit array pr first, that should simplify rebasing

crypto/multisig/threshold_multisig.go
```diff
@@ -0,0 +1,78 @@
...
+// The multisig uses a bitarray, so multiple signatures for the same key is not
+// a concern.
+func (pk *ThresholdMultiSignaturePubKey) VerifyBytes(msg []byte, marshalledSig []byte) bool {
+	var sig *Multisignature
+	err := cdc.UnmarshalBinary(marshalledSig, &sig)
+	if err != nil {
+		return false
+	}
+	size := sig.BitArray.Size()
+	if len(sig.Sigs) < int(pk.K) || len(pk.Pubkeys) != size || sig.BitArray.NumOfTrueBitsBefore(size) < int(pk.K) {
```
ebuchman: Should we also do a consistency check that `len(sig.Sigs) < size`?

Maybe it would be better if we split this up and evaluate the concerns separately.

Like first we check the size of the bitarray against the number of sigs and the number of pubkeys.

Then we handle the K stuff.

crypto/multisig/threshold_multisig_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/tendermint/tendermint/crypto"
+	"github.com/tendermint/tendermint/crypto/ed25519"
+	"github.com/tendermint/tendermint/crypto/secp256k1"
+)
+
+func TestThresholdMultisig(t *testing.T) {
```
ebuchman: I think this test is ripe for table-driven testing.

We should be testing multiple values of N, and make sure we get expected behaviour when N<=0 and when N=1.

We should be testing multiple values of K, including that we get expected behaviour in the edges, eg. when K<=0, K=1, K=N-1, K=N, K>N

ebuchman: Same with the other tests really ...

ebuchman: We should also test what happens with recursive multisig pubkeys. Or if we want to restrict it, then test it fails.

crypto/multisig/wire.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	amino "github.com/tendermint/go-amino"
+	"github.com/tendermint/tendermint/crypto"
+	"github.com/tendermint/tendermint/crypto/ed25519"
+	"github.com/tendermint/tendermint/crypto/secp256k1"
+)
+
+// TODO: Figure out API for others to either add their own pubkey types, or
+// to make verify / marshal accept a cdc.
```
ebuchman: Huh. We don't really expose this as an API at all, which perhaps suggests that PubKey really could be a `oneof`

crypto/multisig/threshold_multisig_test.go
ebuchman: We should also test unmarshalling

crypto/multisig/wire.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+var cdc = amino.NewCodec()
+
+func init() {
+	cdc.RegisterInterface((*crypto.PubKey)(nil), nil)
+	cdc.RegisterConcrete(ThresholdMultiSignaturePubKey{},
+		ThresholdPubkeyAminoRoute, nil)
+	cdc.RegisterConcrete(ed25519.PubKeyEd25519{},
+		ed25519.Ed25519PubKeyAminoRoute, nil)
```
ebuchman: oh, should these just become`PubKeyAminoRoute` since it's an anti-pattern to have the double prefix (pkg and var name)?

crypto/multisig/wire.go
```diff
@@ -0,0 +1,26 @@
...
+	amino "github.com/tendermint/go-amino"
+	"github.com/tendermint/tendermint/crypto"
+	"github.com/tendermint/tendermint/crypto/ed25519"
+	"github.com/tendermint/tendermint/crypto/secp256k1"
+)
+
+// TODO: Figure out API for others to either add their own pubkey types, or
+// to make verify / marshal accept a cdc.
+const (
+	ThresholdPubkeyAminoRoute = "tendermint/ThresholdMultisigPubkey"
```
ebuchman: We use eg. `tendermint/PubKeyEd25519` so this should be `tendermint/PubKeyMultisigThreshold`

crypto/multisig/threshold_multisig.go
```diff
@@ -0,0 +1,78 @@
...
+
+// ThresholdMultiSignaturePubKey implements a K of N threshold multisig.
+type ThresholdMultiSignaturePubKey struct {
+	K       uint            `json:"threshold"`
+	Pubkeys []crypto.PubKey `json:"pubkeys"`
+}
+
+var _ crypto.PubKey = &ThresholdMultiSignaturePubKey{}
+
+// NewThresholdMultiSignaturePubKey returns a new ThresholdMultiSignaturePubKey.
```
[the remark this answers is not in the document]
ebuchman: I saw your comment about Go not having these panics. I think that's bad on their part. We should be as clear as possible to users about when we panic.

crypto/multisig/threshold_multisig_test.go
ValarDragon: Table driven sounds good. I do think we should enable recursive multisigs. 

Not really sure why we would test unmarshalling. We assume that amino works, and all variables are public?

crypto/multisig/threshold_multisig.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Address returns tmhash(ThresholdMultiSignaturePubKey.Bytes())
+func (pk *ThresholdMultiSignaturePubKey) Address() crypto.Address {
+	return crypto.Address(tmhash.Sum(pk.Bytes()))
+}
+
+// Equals returns true iff pk and other both have the same number of keys, and
+// all constituent keys are the same, and in the same order.
+func (pk *ThresholdMultiSignaturePubKey) Equals(other crypto.PubKey) bool {
+	if otherKey, ok := other.(*ThresholdMultiSignaturePubKey); ok {
```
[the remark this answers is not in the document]
ValarDragon: Sounds good! Didn't consider `ok` being overloaded, lol

crypto/multisig/multisignature.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Multisignature is used to represent the signature object used in the multisigs.
+// Sigs is a list of signatures, sorted by corresponding index.
+type Multisignature struct {
+	BitArray *CompactBitArray
+	Sigs     [][]byte
+}
+
+// NewMultisig returns a new Multisignature of size n.
+func NewMultisig(n int) *Multisignature {
```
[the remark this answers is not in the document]
ValarDragon: The object shouldn't be immutable? You want to call `AddSignature` to it repeatedly.

crypto/multisig/multisignature.go
```diff
@@ -0,0 +1,68 @@
...
+	"errors"
+
+	"github.com/tendermint/tendermint/crypto"
+)
+
+// Multisignature is used to represent the signature object used in the multisigs.
+// Sigs is a list of signatures, sorted by corresponding index.
+type Multisignature struct {
+	BitArray *CompactBitArray
+	Sigs     [][]byte
```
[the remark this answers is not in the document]
ValarDragon: If you think that improves clarity, then sure. I didn't really see it as a problem.

ebuchman:
Some review of the compact bit array PR: https://github.com/tendermint/tendermint/pull/2140#pullrequestreview-145310378

crypto/multisig/threshold_multisig.go
ebuchman: but we allow `len(sigs.Sigs) > k`, so we should set an upper bound.

crypto/multisig/multisignature.go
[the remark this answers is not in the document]
ebuchman: Leave for now. Maybe we'll open an issue for it later

crypto/multisig/multisignature.go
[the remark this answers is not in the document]
ebuchman: Was suggesting AddSignature to return a new Multisignature, but yeh it's silly

crypto/multisig/threshold_multisig.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"github.com/tendermint/tendermint/crypto"
+	"github.com/tendermint/tendermint/crypto/tmhash"
+)
+
+// ThresholdMultiSignaturePubKey implements a K of N threshold multisig.
+type ThresholdMultiSignaturePubKey struct {
+	K       uint            `json:"threshold"`
+	Pubkeys []crypto.PubKey `json:"pubkeys"`
```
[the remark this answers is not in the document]
ebuchman: Ok. Let's add tests for it.

Will we have some way to expose how deep we've gone in the recursion tho? Or will the apps just have to restrict total sig/tx size ?

crypto/multisig/threshold_multisig_test.go
ebuchman: Right

crypto/multisig/wire.go
ebuchman: Yes I wasn't suggesting you do anything. But it goes back to the larger conversation of "do we really need pubkey to be an any-one-can-implement-it-without-permission interface" ie. the kind of thing Amino was built for, or would pure protobuf (using protobufs union type, called `oneof`) do just fine, since there isn't really a nice way to add new pubkey types right now without changing our crypto lib

Note this is unlike Msg in the SDK which really does want to be an interface.

ValarDragon: Oh, I think pure protobuf is better here. We have to register it in all the codecs, so its not really like anyone can just implement it.

crypto/multisig/threshold_multisig.go
ValarDragon: Oh woops, your totally right, great catch!

crypto/multisig/threshold_multisig.go
```diff
@@ -0,0 +1,78 @@
...
+			}
+			sigIndex++
+		}
+	}
+	return true
+}
+
+// Bytes returns the amino encoded version of the ThresholdMultiSignaturePubKey
+func (pk *ThresholdMultiSignaturePubKey) Bytes() []byte {
+	return cdc.MustMarshalBinary(pk)
```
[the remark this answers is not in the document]
ValarDragon: Good point! I'll make this marshal bare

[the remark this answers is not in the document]
ValarDragon: Actually, we probably want to switch all pubkeys to use bare at once, lets move this to a seperate issue.

crypto/multisig/threshold_multisig.go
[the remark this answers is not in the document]
ValarDragon: We can expose how deep we're going if we want to make a seperate "MultisigPubkey" interface, or a "Size()" method on the main pubkey interface. (Its easy to add such a method, the problem is how to best expose it. The method would just be iterating thru all pubkeys, and add 1 to size if its not a multisig, and to call .size if its a threshold or weighted multisig)

I'm not really sure whats the ideal way to do this. From the SDK's side, we probably want a method to calculate gas cost based off number of signatures, and weight secp and ed25519 differently.

crypto/multisig/threshold_multisig.go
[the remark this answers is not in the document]
ValarDragon: Oh other pubkeys do already do bare. I'll switch this to bare then. Would you like to make an issue for switching amino default in general?

ValarDragon:
Follow up comment made in an open issue!

crypto/multisig/compact_bit_array.go
```diff
@@ -71,6 +71,21 @@ func (bA *CompactBitArray) SetIndex(i int, v bool) bool {
 	return true
 }
 
+// NumOfTrueBitsBefore returns the location of the given index, among the
+// values in the bit array that are set to true.
+// e.g. if bA = _XX_X_X, NumOfTrueBitsBefore(4) = 2, since
+// the value at index 4 of the bit array is the third
+// value that is true. (And it is 0-indexed)
+func (bA *CompactBitArray) NumOfTrueBitsBefore(index int) int {
```
jaekwon: minor point, I'd just call this `NumTrueBitsBefore(index int)`.  I think "of" is implied.

crypto/multisig/compact_bit_array.go
```diff
@@ -71,6 +71,21 @@ func (bA *CompactBitArray) SetIndex(i int, v bool) bool {
 	return true
 }
 
+// NumOfTrueBitsBefore returns the location of the given index, among the
```
jaekwon: "returns the location of the given index" is not the right wording since it doesn't return a location, but number of bits true.

crypto/multisig/compact_bit_array.go
```diff
@@ -71,6 +71,21 @@ func (bA *CompactBitArray) SetIndex(i int, v bool) bool {
...
+// values in the bit array that are set to true.
+// e.g. if bA = _XX_X_X, NumOfTrueBitsBefore(4) = 2, since
```
jaekwon: "since ... value at index 4 .... is true" 
"since" seems like a strange modifier and the following statement a bit orthogonal.

jaekwon: Maybe it make sense to let the user specify 0 to count the total number of true bits?
And then we can comment that, which makes it clear that you don't want to enter NumTrueBitsBefore(0) when you really mean NumTrueBitsBefore(1).

ValarDragon: I feel like making edge cases like letting 0 counting the number of true bits makes it more confusing. I do definitely agree with your point about the comments here!

crypto/multisig/threshold_multisig.go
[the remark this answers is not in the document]
jaekwon: Separate to the discussion of depth, I think we should also add a "IsComposite()" to PubKey, and maybe do this: https://github.com/cosmos/cosmos-sdk/issues/2019

crypto/multisig/threshold_multisig.go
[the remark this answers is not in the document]
ebuchman: https://github.com/tendermint/go-amino/issues/219

crypto/multisig/wire.go
ebuchman: Addressed in https://github.com/tendermint/tendermint/pull/2205

ebuchman:
Sorry I forgot to squash everything. 

Follow up is in https://github.com/tendermint/tendermint/issues/2163#issuecomment-412177158

Issue #2130, opened 2018-08-02
p2p/pex: Connect to more than 10 peers


melekes:
https://github.com/tendermint/tendermint/blob/eaa137512cddbb691bb301102c8355ea2a4e47e9/p2p/pex/pex_reactor.go#L364

PEX should respect `min`, but try to connect to as many peers as possible (up to `max`)

### What you are seeing
It's been reported that in normal mode (!seed, pex enabled) a node stops trying to connect to other peers after it has =10 outbound connections.

### What you expect to see
Node should connect up to `MaxNumPeers` and only then stop any attempts.

melekes:
Of course! Sorry

ebuchman:
Sorry I didn't really check in with this earlier.

There's a good reason we only dial min and not all the way to max - it's so we can leave a lot of room to accept new connections.

We might want to expose minNumOutbound peers to the config, but I don't think we should get rid of it. That will just make it harder for new peers to join the network.

Is there some reason we think we need to keep dialing more than some min amount which is less than the max? Or was the reasoning just not clearly explained?

mdyring:
I discussed this with @melekes on Riot.

I have configured max_num_peers = 256 for gaia-7005 and would expect to see the number of peers increase until the limit is reached or no known, unconnected peers exist.

It is unclear from config.toml, that there is a minimum number of peers as well, after which nothing will happen.

Stopping at some lower number of peers seems counter-intuitive, there might be good reason for wanting to connect to as many peers as possible.

IMHO it makes sense to just use the max_num_peers and then remove the minimum. This will provide maximum control for the user.

ValarDragon:
Why not have a `maxOutboundPeers` and `maxInboundPeers` in addition to max peers. That seems like it will give a high degree of configurability while also being easy to understand.

mdyring:
@ValarDragon yeah just edited my reply after I re-read (and understood ;-) what @ebuchman commented.

melekes:
> There's a good reason we only dial min and not all the way to max - it's so we can leave a lot of room to accept new connections.

This is an internal knowledge and should be visible to the end user either in config or logs.

Although we're designing Tendermint for Cosmos, I don't think we should make any assumptions about desired network topology, but instead, give the ability to regulate it or/and make it public knowledge.

mdyring:
Looks great - thanks!

ebuchman:
Believe this was addressed by #2169

PR #2169, opened 2018-08-07
[p2p/pex] connect to more than 10 peers

melekes:
Refs #2130

* [ ] Wrote tests

### DIFF
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -26,10 +26,6 @@ const (
 	// ie. 3**10 = 16hrs
 	reconnectBackOffAttempts    = 10
 	reconnectBackOffBaseSeconds = 3
-
-	// keep at least this many outbound peers
-	// TODO: move to config
-	DefaultMinNumOutboundPeers = 10
 )
 
 //-----------------------------------------------------------------------------
@@ -268,6 +264,11 @@ func (sw *Switch) NumPeers() (outbound, inbound, dialing int) {
 	return
 }
 
+// MaxNumPeers returns a maximum number of peers.
+func (sw *Switch) MaxNumPeers() int {
+	return sw.config.MaxNumPeers
+}
+
 // Peers returns the set of peers that are connected to the switch.
 func (sw *Switch) Peers() IPeerSet {
 	return sw.peers
@@ -491,11 +492,14 @@ func (sw *Switch) listenerRoutine(l Listener) {
 			break
 		}
 
-		// ignore connection if we already have enough
-		// leave room for MinNumOutboundPeers
-		maxPeers := sw.config.MaxNumPeers - DefaultMinNumOutboundPeers
-		if maxPeers <= sw.peers.Size() {
-			sw.Logger.Info("Ignoring inbound connection: already have enough peers", "address", inConn.RemoteAddr().String(), "numPeers", sw.peers.Size(), "max", maxPeers)
+		// Ignore connection if we already have enough peers.
+		if sw.config.MaxNumPeers <= sw.peers.Size() {
+			sw.Logger.Info(
+				"Ignoring inbound connection: already have enough peers",
+				"address", inConn.RemoteAddr().String(),
+				"numPeers", sw.peers.Size(),
+				"max", sw.config.MaxNumPeers,
+			)
 			inConn.Close()
 			continue
 		}
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -264,9 +264,9 @@ func (sw *Switch) NumPeers() (outbound, inbound, dialing int) {
 	return
 }
 
-// MaxNumPeers returns a maximum number of peers.
-func (sw *Switch) MaxNumPeers() int {
-	return sw.config.MaxNumPeers
+// MaxNumOutboundPeers returns a maximum number of outbound peers.
+func (sw *Switch) MaxNumOutboundPeers() int {
+	return sw.config.MaxNumOutboundPeers
 }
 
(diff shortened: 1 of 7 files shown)


melekes:
```go
func TestPEXReactorRunning(t *testing.T) {
	N := 18
	switches := make([]*p2p.Switch, N)

	// directory to store address books
	dir, err := ioutil.TempDir("", "pex_reactor")
	require.Nil(t, err)
	defer os.RemoveAll(dir) // nolint: errcheck

	books := make([]*addrBook, N)
	logger := log.TestingLogger()

	// create switches
	for i := 0; i < N; i++ {
		switches[i] = p2p.MakeSwitch(cfg, i, "testing", "123.123.123", func(i int, sw *p2p.Switch) *p2p.Switch {
			books[i] = NewAddrBook(filepath.Join(dir, fmt.Sprintf("addrbook%d.json", i)), false)
			books[i].SetLogger(logger.With("pex", i))
			sw.SetAddrBook(books[i])

			sw.SetLogger(logger.With("pex", i))

			r := NewPEXReactor(books[i], &PEXReactorConfig{})
			r.SetLogger(logger.With("pex", i))
			r.SetEnsurePeersPeriod(250 * time.Millisecond)
			sw.AddReactor("pex", r)

			return sw
		})
	}

	addOtherNodeAddrToAddrBook := func(switchIndex, otherSwitchIndex int) {
		addr := switches[otherSwitchIndex].NodeInfo().NetAddress()
		books[switchIndex].AddAddress(addr, addr)
	}

	addOtherNodeAddrToAddrBook(0, 1)
	addOtherNodeAddrToAddrBook(1, 0)
	addOtherNodeAddrToAddrBook(2, 1)
	addOtherNodeAddrToAddrBook(3, 1)
	addOtherNodeAddrToAddrBook(4, 1)
	addOtherNodeAddrToAddrBook(5, 1)
	addOtherNodeAddrToAddrBook(6, 1)
	addOtherNodeAddrToAddrBook(7, 1)
	addOtherNodeAddrToAddrBook(8, 1)
	addOtherNodeAddrToAddrBook(9, 1)
	addOtherNodeAddrToAddrBook(10, 1)
	addOtherNodeAddrToAddrBook(11, 1)
	addOtherNodeAddrToAddrBook(12, 1)
	addOtherNodeAddrToAddrBook(13, 1)
	addOtherNodeAddrToAddrBook(14, 1)
	addOtherNodeAddrToAddrBook(15, 1)
	addOtherNodeAddrToAddrBook(16, 1)
	addOtherNodeAddrToAddrBook(17, 1)

	for i, sw := range switches {
		sw.AddListener(p2p.NewDefaultListener("tcp://"+sw.NodeInfo().ListenAddr, "", false, logger.With("pex", i)))

		err := sw.Start() // start switch and reactors
		require.Nil(t, err)
	}

	assertPeersWithTimeout(t, switches, 10*time.Millisecond, 10*time.Second, N-1)

	// test at least one switch has more than 10 outbound peers
	moreThan10OutboundPeers := false
	for _, sw := range switches {
		out, _, _ := sw.NumPeers()
		if out > 10 {
			moreThan10OutboundPeers = true
		}
	}
	if !moreThan10OutboundPeers {
		t.Fatal("None of the switches has more than 10 outbound peers")
	}

	// stop them
	for _, s := range switches {
		s.Stop()
	}
}
```

I used modified version of TestPEXReactorRunning to test the expected behaviour.

p2p/switch.go
```diff
@@ -491,11 +492,14 @@ func (sw *Switch) listenerRoutine(l Listener) {
 			break
 		}
 
-		// ignore connection if we already have enough
-		// leave room for MinNumOutboundPeers
-		maxPeers := sw.config.MaxNumPeers - DefaultMinNumOutboundPeers
```
ValarDragon: Should we have some bias towards accepting incoming peers depending on our current inbound / outbound peer ratio? Thats something we can do postlaunch though.

melekes: I thought about it a little bit. We can give greater control by introducing MaxNumInboundPeer/MaxNumOutboundPeers or InOutPeerRatio or something else. But it's perfectly fine to do it later.

ValarDragon: Sounds good to me! Should we make an issue for that?

ebuchman:
I'm not sure we should do this - see https://github.com/tendermint/tendermint/issues/2130#issuecomment-412153064

We should fix the log statement to include the min outbound peers since I agree its confusing that we don't say it there.

melekes:
Replaced MaxNumPeers with MaxNumInboundPeers/MaxNumOutboundPeers.

p2p/switch.go
```diff
@@ -491,11 +492,15 @@ func (sw *Switch) listenerRoutine(l Listener) {
...
-		if maxPeers <= sw.peers.Size() {
-			sw.Logger.Info("Ignoring inbound connection: already have enough peers", "address", inConn.RemoteAddr().String(), "numPeers", sw.peers.Size(), "max", maxPeers)
+		// Ignore connection if we already have enough peers.
+		_, in, _ := sw.NumPeers()
+		if in >= sw.config.MaxNumInboundPeers {
```
melekes: note: if MaxNumInboundPeers is negative, number of conns is unlimited => which can cause consensus crash due to not enough file descriptors. Should we check and panic?

ebuchman: We should probably catch issues in the config much sooner. Maybe we add ValidateBasic methods to the config struct we can call after parsing?

melekes: I like this idea

Issue #2127, opened 2018-08-01
ABCI / mempool: Add a "Recheck Tx" function


ValarDragon:
In addition to CheckTx, we should also have a "Recheck Tx". The functionality would basically be the same, but it would be intended for use on things that have already passed a CheckTx once before. The reason for this is that the app could then skip certain checks, such as signature verification. Regardless of state changes, for most txs the signatures will remain valid. If there was a "MsgChangePubkey" this could be checked in the application logic, and in that case assoc. signature re-checked. (Given that this doesn't yet exist, doesn't seem like a concern right now)

Then we would call recheck in the mempool instead of check.

Right now rechecking every tx on each new block, and therefore doing each signature verification is going to be a significant hit on performance if you have a non-trivially filled mempool. Recall that our current Secp signature verification implementation is `505039 ns/op`. Right now the bitcoin mempool is at 15MB. (https://www.blockchain.com/charts/mempool-size). The average bitcoin tx is ~300 bytes, which means that there are `52428` txs in the mempool. At signature verification time we have, this means a recheck would take 26 seconds. Thus the proposed scheme seems to me be a very important concern, as the mempool rechecking processing time between blocks would be 26 seconds. (Perhaps divided by a small constant factor due to parallelization)

ValarDragon:
cc @jackzampolin

ebuchman:
Nice idea. Though technically this could be performed app side right now by keeping an in-memory map of txs we've successfully run CheckTx for, yeh? It would be like replicating the mempool cache in the app.

ValarDragon:
Sure it could, but that breaks abstraction barriers. We want handling of sorting and priorities to belong in the mempool. (We don't want the app maintaining an _additional_ map of all txs) That also seems more wasteful in terms of resources. 

I feel like having a separate ABCI message (or alternatively additional boolean in CheckTx) is more optimal. I'm not opposed to adding a boolean in CheckTx to signal if were rechecking or not.

ebuchman:
Right. Maybe the time has finally come for a RequestCheckTx type

ValarDragon:
That sounds super useful to me!

ValarDragon:
I'd prefer "CheckTxTypes", `0` being unchecked, `1` being naive "this tx has been verified before", with further room for more, and an additional `[]byte` sent along the wire as "additional data".

[...]

PR #2173, opened 2018-08-07
add recheckTx ,ABCI / mempool: Add a "Recheck Tx" function #2127

bradyjoestar:
I try to solve the #2127  issue to reduce the verification of signature as ValarDragon said.

So I modity the `RequestCheckTx` message in the ABCI/types/types.proto ,I add a `recheck` flag to prove the tx is from mempool of tendermint.

I add  "RecheckTxAysnc" and "RecheckTxSync" function in the `ABCI/client/client.go`  `ABCI/client/grpc_client.go`, `ABCI/client/local_client.go` ," `ABCI/client/socket_client.go`".

However in the local_client , `RecheckTxAsync` doesnt work because I found it called the "CheckTx(tx []byte)" of types.BaseApplication to communicate with the ABCIAPP directly . I will continue to solve it.

* [ ] Wrote tests


abci/client/client.go
```diff
@@ -42,6 +43,7 @@ type Client interface {
 	SetOptionSync(types.RequestSetOption) (*types.ResponseSetOption, error)
 	DeliverTxSync(tx []byte) (*types.ResponseDeliverTx, error)
 	CheckTxSync(tx []byte) (*types.ResponseCheckTx, error)
+	RecheckTxSync(tx []byte) (*types.ResponseCheckTx, error)
```
ValarDragon: Good idea! Sync recheck tx makes sense if we already had a new tx in our mempool. To make this work we'd have to make this recheck share the same state as the normal check tx, which shouldn't be too complex to do. May want to save implementation of this for a separate pr

bradyjoestar: Thanks for your review!! I found it's difficult to process the `checkTx`  function on the application. I will close this pull request and create a new pull request when I solve this issue completely.

PR #2179, opened 2018-08-07
Reduce Commit to Signatures and Timestamps

### LINKED ISSUE #1648: Change `Commit` to consist of just signatures
Currently, every block includes a LastCommit which is the set of +2/3 Precommit votes for the last block.

Note votes include validator address, index, blockID, as well as signature.

All of that information is redundant except for the signature - thus we could make blocks significantly smaller if we only included the signature in the Commit, instead of the whole vote.

jlandrews:
Closes #1648 
* [x] Wrote tests

### DIFF
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -629,7 +629,7 @@ OUTER_LOOP:
 			// Load the block commit for prs.Height,
 			// which contains precommit signatures for prs.Height.
 			commit := conR.conS.blockStore.LoadBlockCommit(prs.Height)
-			commit.AddAddresses(conR.conS.Validators.GetAddresses())
+			commit.SetAddresses(conR.conS.Validators.GetAddresses())
 			if ps.PickSendVote(commit) {
 				logger.Debug("Picked Catchup commit to send", "height", prs.Height)
 				continue OUTER_LOOP
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -629,6 +629,7 @@ OUTER_LOOP:
 			// Load the block commit for prs.Height,
 			// which contains precommit signatures for prs.Height.
 			commit := conR.conS.blockStore.LoadBlockCommit(prs.Height)
+			commit.AddAddresses(conR.conS.Validators.GetAddresses())
 			if ps.PickSendVote(commit) {
 				logger.Debug("Picked Catchup commit to send", "height", prs.Height)
 				continue OUTER_LOOP
--- a/consensus/replay_test.go
+++ b/consensus/replay_test.go
@@ -550,12 +550,10 @@ func makeBlockchainFromWAL(wal WAL) ([]*types.Block, []*types.Commit, error) {
 			if p.Type == types.VoteTypePrecommit {
 				thisBlockCommit = &types.Commit{
 					BlockID: p.BlockID,
-					Precommits: []*types.CommitSig{
-						&types.CommitSig{
-							Signature: p.Signature,
-							Timestamp: p.Timestamp,
-						},
-					},
+					Precommits: []*types.CommitSig{{
+						Signature: p.Signature,
+						Timestamp: p.Timestamp,
+					}},
 					HeightNum: p.Height,
 					RoundNum:  p.Round,
 				}
--- a/consensus/replay_test.go
+++ b/consensus/replay_test.go
@@ -531,7 +531,7 @@ func makeBlockchainFromWAL(wal WAL) ([]*types.Block, []*types.Commit, error) {
 				if block.Height != height+1 {
 					panic(fmt.Sprintf("read bad block from wal. got height %d, expected %d", block.Height, height+1))
 				}
-				commitHeight := thisBlockCommit.Precommits[0].Height
+				commitHeight := thisBlockCommit.Height()
 				if commitHeight != height+1 {
 					panic(fmt.Sprintf("commit doesnt match. got height %d, expected %d", commitHeight, height+1))
 				}
@@ -549,8 +549,15 @@ func makeBlockchainFromWAL(wal WAL) ([]*types.Block, []*types.Commit, error) {
 		case *types.Vote:
 			if p.Type == types.VoteTypePrecommit {
 				thisBlockCommit = &types.Commit{
-					BlockID:    p.BlockID,
-					Precommits: []*types.Vote{p},
+					BlockID: p.BlockID,
+					Precommits: []*types.CommitSig{
+						&types.CommitSig{
+							Signature: p.Signature,
+							Timestamp: p.Timestamp,
+						},
+					},
+					HeightNum: p.Height,
+					RoundNum:  p.Round,
 				}
 			}
 		}
@@ -564,7 +571,7 @@ func makeBlockchainFromWAL(wal WAL) ([]*types.Block, []*types.Commit, error) {
 	if block.Height != height+1 {
 		panic(fmt.Sprintf("read bad block from wal. got height %d, expected %d", block.Height, height+1))
 	}
-	commitHeight := thisBlockCommit.Precommits[0].Height
+	commitHeight := thisBlockCommit.Height()
 	if commitHeight != height+1 {
 		panic(fmt.Sprintf("commit doesnt match. got height %d, expected %d", commitHeight, height+1))
 	}
--- a/consensus/state_test.go
+++ b/consensus/state_test.go
@@ -7,7 +7,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/tendermint/abci/example/counter"
 	cstypes "github.com/tendermint/tendermint/consensus/types"
+	dbm "github.com/tendermint/tendermint/libs/db"
 	"github.com/tendermint/tendermint/libs/log"
 	tmpubsub "github.com/tendermint/tendermint/libs/pubsub"
 	"github.com/tendermint/tendermint/types"
@@ -289,6 +292,50 @@ func TestStateFullRoundNil(t *testing.T) {
 
 	// should prevote and precommit nil
 	validatePrevoteAndPrecommit(t, cs, round, 0, vss[0], nil, nil)
+}
+
+func TestStateFullRound1Reload(t *testing.T) {
+	blockDB := dbm.NewMemDB()
+	state, privVals := randGenesisState(1, false, 10)
+	vss := make([]*validatorStub, 1)
+	cs := newConsensusStateWithConfigAndBlockStore(config, state, privVals[0], counter.NewCounterApplication(true), blockDB)
+	vss[0] = NewValidatorStub(privVals[0], 0)
+	incrementHeight(vss[1:]...)
+	height, round := cs.Height, cs.Round
+
+	// NOTE: buffer capacity of 0 ensures we can validate prevote and last commit
+	// before consensus can move to the next height (and cause a race condition)
+	cs.eventBus.Stop()
+	eventBus := types.NewEventBusWithBufferCapacity(0)
+	eventBus.SetLogger(log.TestingLogger().With("module", "events"))
+	cs.SetEventBus(eventBus)
+	eventBus.Start()
+
+	voteCh := subscribe(cs.eventBus, types.EventQueryVote)
+	propCh := subscribe(cs.eventBus, types.EventQueryCompleteProposal)
+	newRoundCh := subscribe(cs.eventBus, types.EventQueryNewRound)
+
+	startTestRound(cs, height, round)
+
+	<-newRoundCh
+
+	// grab proposal
+	re := <-propCh
+	propBlockHash := re.(types.EventDataRoundState).RoundState.(*cstypes.RoundState).ProposalBlock.Hash()
+
+	<-voteCh // wait for prevote
+	validatePrevote(t, cs, round, vss[0], propBlockHash)
+
+	<-voteCh // wait for precommit
+
+	// we're going to roll right into new height
+	<-newRoundCh
+
+	validateLastPrecommit(t, cs, vss[0], propBlockHash)
+
+	// try reloading the consensus state from the DB
+	ncs := newConsensusStateWithConfigAndBlockStore(config, cs.state, nil, counter.NewCounterApplication(true), blockDB)
+	require.NotNil(t, ncs, "newConsensusState should not fail")
 }
 
 // run through propose, prevote, precommit commit with two validators
--- a/consensus/state_test.go
+++ b/consensus/state_test.go
@@ -294,6 +294,7 @@ func TestStateFullRoundNil(t *testing.T) {
 	validatePrevoteAndPrecommit(t, cs, round, 0, vss[0], nil, nil)
 }
 
+// Tests that consensus state can be saved and reloaded from DB after committing a block.
 func TestStateFullRound1Reload(t *testing.T) {
 	blockDB := dbm.NewMemDB()
 	state, privVals := randGenesisState(1, false, 10)
--- a/consensus/types/round_state_test.go
+++ b/consensus/types/round_state_test.go
@@ -16,7 +16,7 @@ func BenchmarkRoundStateDeepCopy(b *testing.B) {
 	// Random validators
 	nval, ntxs := 100, 100
 	vset, _ := types.RandValidatorSet(nval, 1)
-	precommits := make([]*types.Vote, nval)
+	precommits := make([]*types.CommitSig, nval)
 	blockID := types.BlockID{
 		Hash: cmn.RandBytes(20),
 		PartsHeader: types.PartSetHeader{
@@ -25,11 +25,9 @@ func BenchmarkRoundStateDeepCopy(b *testing.B) {
 	}
 	sig := make([]byte, ed25519.SignatureSize)
 	for i := 0; i < nval; i++ {
-		precommits[i] = &types.Vote{
-			ValidatorAddress: types.Address(cmn.RandBytes(20)),
-			Timestamp:        time.Now(),
-			BlockID:          blockID,
-			Signature:        sig,
+		precommits[i] = &types.CommitSig{
+			Timestamp: time.Now(),
+			Signature: sig,
 		}
 	}
 	txs := make([]types.Tx, ntxs)
--- a/consensus/types/round_state_test.go
+++ b/consensus/types/round_state_test.go
@@ -26,7 +26,7 @@ func BenchmarkRoundStateDeepCopy(b *testing.B) {
 	sig := make([]byte, ed25519.SignatureSize)
 	for i := 0; i < nval; i++ {
 		precommits[i] = &types.CommitSig{
-			Timestamp: time.Now(),
+			Timestamp: time.Now().UTC(),
 			Signature: sig,
 		}
 	}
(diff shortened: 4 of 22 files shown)


jlandrews:
Not sure why `test_p2p` is stalling out here, will investigate and write a unit test to catch the behavior earlier.

jlandrews:
`test_p2p` was stalling because I forgot to have `VoteSet.MakeCommit` remove any votes with different `BlockID`. It seems in `test_p2p`, the first committed block only had 3/4 votes for the same `BlockID` with the 4th being a nil vote. When trying to verify block 2, the `LastCommit` would fail when verifying the signature for the nil vote, since it was incorrectly reconstructed with the majority `BlockID`.

lite/helpers.go
```diff
@@ -71,20 +71,25 @@ func (pkz privKeys) ToValidators(init, inc int64) *types.ValidatorSet {
 
 // signHeader properly signs the header with all keys from first to last exclusive.
 func (pkz privKeys) signHeader(header *types.Header, first, last int) *types.Commit {
-	votes := make([]*types.Vote, len(pkz))
+	cs := make([]*types.CommitSig, len(pkz))
```
melekes: `cs` is a common name for consensus. Can I suggest a different name? `sigs` or `commitSigs` or `precommits` perhaps

types/block.go
```diff
@@ -296,49 +296,38 @@ type Commit struct {
 	// NOTE: The Precommits are in order of address to preserve the bonded ValidatorSet order.
 	// Any peer with a block can gossip precommits by index with a peer without recalculating the
 	// active ValidatorSet.
-	BlockID    BlockID `json:"block_id"`
-	Precommits []*Vote `json:"precommits"`
+	BlockID    BlockID      `json:"block_id"`
+	Precommits []*CommitSig `json:"precommits"`
+	RoundNum   int
```
melekes: Why the `Num` suffix?

types/block.go
```diff
@@ -368,8 +357,16 @@ func (commit *Commit) BitArray() *cmn.BitArray {
 }
 
 // GetByIndex returns the vote corresponding to a given validator index
-func (commit *Commit) GetByIndex(index int) *Vote {
-	return commit.Precommits[index]
+func (com *Commit) GetByIndex(index int) *Vote {
```
melekes: `com -> commit` for consistency

types/vote_set.go
```diff
@@ -539,13 +539,25 @@ func (voteSet *VoteSet) MakeCommit() *Commit {
 	if voteSet.maj23 == nil {
 		cmn.PanicSanity("Cannot MakeCommit() unless a blockhash has +2/3")
 	}
+	blockID := *voteSet.maj23
 
 	// For every validator, get the precommit
 	votesCopy := make([]*Vote, len(voteSet.votes))
 	copy(votesCopy, voteSet.votes)
+	precommits := make([]*CommitSig, len(voteSet.votes))
```
melekes: Shouldn't we set length to 0 and capacity to `en(voteSet.votes)` and use append if some votes can be nil? Or it's ok to have some `precommits` eq `nil`?

types/block.go
melekes: if this is because of the `Height/Round()` methods, can't we just get rid of them?

consensus/replay_test.go
```diff
@@ -551,8 +551,15 @@ func makeBlockchainFromWAL(wal WAL) ([]*types.Block, []*types.Commit, error) {
 		case *types.Vote:
 			if p.Type == types.VoteTypePrecommit {
 				thisBlockCommit = &types.Commit{
-					BlockID:    p.BlockID,
-					Precommits: []*types.Vote{p},
+					BlockID: p.BlockID,
+					Precommits: []*types.CommitSig{
+						&types.CommitSig{
```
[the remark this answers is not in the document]
jlandrews: :+1:

lite/helpers.go
jlandrews: :+1:

types/block.go
jlandrews: :+1:

types/block.go
jlandrews: It is because of the getter functions. They are needed for`Commit` to implement `VoteSetReader`, which allows the consensus reactor to pull votes from `LastCommit` for gossiping to peers.
https://github.com/tendermint/tendermint/blob/b202edca9b42e5b457556d418300f19c0b0d0608/consensus/reactor.go#L617-L624

types/vote_set.go
jlandrews: It is ok to have some precommits be `nil`. We use the full array so that the `CommitSigs` have matching indices to the validators who signed them. Then when verifying we just skip any that are `nil`. I imagine we could make this work without any `nil` values by using the bit array and counting to determine indices at verification time, but I don't know if that would end up using less total space when serialized.

docs/spec/blockchain/blockchain.md
```diff
@@ -107,6 +108,29 @@ There are two types of votes:
...
+## Commit
+
+A Commit contains enough data to verify that a BlockID was committed to at a certain
+height and round by a sufficient set of validators.
+
+```go
+type Commit struct {
+	BlockID    BlockID
+	Precommits []*CommitSig
+	RoundNum   int
```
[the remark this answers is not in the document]
melekes: > Did we have some good reason not to use int64 for rounds?

less space?) if Round number goes above 2147483647, then something is def wrong

jlandrews:
@ebuchman @melekes How would you feel about removing `ValidatorAddress` entirely from `Vote`, it looks like it's only being used for sanity checking and printing votes. Both of those use cases could be replaced by using the `ValidatorIndex` to query the address from the `ValidatorSet`, and removing it would save a fair bit of space in vote gossiping.

docs/spec/blockchain/blockchain.md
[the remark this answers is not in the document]
ValarDragon: We should probably note this somewhere then. 

I presume `int` was chosen vs `int32` to get the varint encoding for better space efficiency?

jlandrews:
It was pretty easy to do, I already had most of it done, just wanted to get feedback on the concept before pushing the commit.

jlandrews:
Interesting problem I'm running into here... There's no way to find all the places that a `String()` function is being called from, since it's an implicit conversion. I've changed the method signatures and all the tests pass, but I'm sure there are places where blocks, votes or commits are being printed or logged that I can't find. It seems like the only option is to manually look through areas where those objects are being used and look for implicit conversions, or just not have the validator address in the string representation.

ebuchman:
@jlandrews can we revert removing the Address for now? https://github.com/tendermint/tendermint/issues/2226#issuecomment-413941227

consensus/state_test.go
```diff
@@ -291,6 +294,50 @@ func TestStateFullRoundNil(t *testing.T) {
 	validatePrevoteAndPrecommit(t, cs, round, 0, vss[0], nil, nil)
 }
 
+func TestStateFullRound1Reload(t *testing.T) {
```
melekes: Could you add a few sentences above as for what this test checks?

consensus/types/round_state_test.go
```diff
@@ -25,11 +25,9 @@ func BenchmarkRoundStateDeepCopy(b *testing.B) {
 	}
 	sig := make([]byte, ed25519.SignatureSize)
 	for i := 0; i < nval; i++ {
-		precommits[i] = &types.Vote{
-			ValidatorAddress: types.Address(cmn.RandBytes(20)),
-			Timestamp:        time.Now(),
-			BlockID:          blockID,
-			Signature:        sig,
+		precommits[i] = &types.CommitSig{
+			Timestamp: time.Now(),
```
melekes: `time.Now().UTC()`?

types/block.go
```diff
@@ -386,30 +407,6 @@ func (commit *Commit) ValidateBasic() error {
...
-		if precommit == nil {
-			continue
-		}
-		// Ensure that all votes are precommits.
-		if precommit.Type != VoteTypePrecommit {
-			return fmt.Errorf("Invalid commit vote. Expected precommit, got %v",
-				precommit.Type)
-		}
-		// Ensure that all heights are the same.
-		if precommit.Height != height {
```
melekes: shouldn't we still check this once for `commit`?

types/validator_set.go
```diff
@@ -270,34 +278,21 @@ func (vals *ValidatorSet) VerifyCommit(chainID string, blockID BlockID, height i
...
 
 	talliedVotingPower := int64(0)
-	round := commit.Round()
+	commit.AddAddresses(vals.GetAddresses())
 
 	for idx, precommit := range commit.Precommits {
 		if precommit == nil {
 			continue // OK, some precommits can be missing.
 		}
-		if precommit.Height != height {
```
melekes: shouldn't we still check this once for `commit`?

types/validator_set.go
```diff
@@ -349,40 +344,27 @@ func (vals *ValidatorSet) VerifyFutureCommit(newSet *ValidatorSet, chainID strin
 	// Check old voting power.
 	oldVotingPower := int64(0)
 	seen := map[int]bool{}
-	round := commit.Round()
 
 	for idx, precommit := range commit.Precommits {
 		if precommit == nil {
 			continue
 		}
-		if precommit.Height != height {
```
melekes: shouldn't we still check this once for `commit`?

types/block.go
jlandrews: Here the check is only working with data stored in the `Commit`, so it has no height/round to check against. The previous code was just checking that all contained precommits had the same values, but now they automatically do implicitly.

types/validator_set.go
jlandrews: We do check that the commit has height/blockID which match the function parameters, but that code is outside the context here. https://github.com/tendermint/tendermint/blob/e5965ababe944d7470b4c21c52d09af93a8640b6/types/validator_set.go#L272-L283

types/validator_set.go
jlandrews: Here we don't check the commit height/blockID in the function, but it does make a call to `VerifyCommit` with the appropriate height/blockID, so it's doing the check there.

https://github.com/tendermint/tendermint/blob/e5965ababe944d7470b4c21c52d09af93a8640b6/types/validator_set.go#L338-L343

docs/spec/blockchain/blockchain.md
```diff
@@ -107,6 +108,29 @@ There are two types of votes:
...
 a *precommit* has `vote.Type == 2`.
 
+## Commit
+
+A Commit contains enough data to verify that a BlockID was committed to at a certain
+height and round by a sufficient set of validators.
+
+```go
+type Commit struct {
+    HeightNum  int64
```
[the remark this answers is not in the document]
jlandrews: I would agree with using the standard variable names, but we need the `Height()` getters to implement the `VoteSetReader` interface so that `Commit`s can have votes pulled from them for gossiping via the same process as in progress vote sets. I would be open to alternative names tho, as `HeightNum` is a bit awkward.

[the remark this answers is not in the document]
melekes: What if we make them private and use `NewCommit` (or create additional constructors) to create `Commit`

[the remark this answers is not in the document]
ebuchman: In the spec we don't have to worry about collisions with method names, we can just call it Height.

consensus/reactor.go
```diff
@@ -629,6 +629,7 @@ OUTER_LOOP:
 			// Load the block commit for prs.Height,
 			// which contains precommit signatures for prs.Height.
 			commit := conR.conS.blockStore.LoadBlockCommit(prs.Height)
+			commit.SetAddresses(conR.conS.Validators.GetAddresses())
```
ebuchman: I don't think this is the correct validator set - it's using our current validator set but the commit is coming from the peers height, which is less than ours in this `if` clause.

The ease of getting something like this wrong is a bit uncomfortable. I wonder if there's still a better way - maybe trying to stuff both vote and commit into the VoteSetReader isn't the best design?

ebuchman: Maybe instead we can define a method that takes a blockStore and a state and loads the full vote set (ie. the commits plus the addresses)

consensus/types/round_state_test.go
```diff
@@ -40,7 +38,7 @@ func BenchmarkRoundStateDeepCopy(b *testing.B) {
 	block := &types.Block{
 		Header: types.Header{
 			ChainID:         cmn.RandStr(12),
-			Time:            time.Now(),
+			Time:            time.Now().UTC(),
```
ebuchman: will conflcit with bft time pr

consensus/reactor.go
jaekwon: yeah this is the wrong validatorset... we're fetching the commit for height prs.Height (a peer's height) and setting the current local validatorset...

jlandrews: Yup, good catch @ebuchman. Are we currently storing the historical validator sets anywhere? Because it not we'll have to start doing that to be able to reconstruct votes with addresses for arbitrary heights like this.

types/validator_set.go
```diff
@@ -270,34 +278,21 @@ func (vals *ValidatorSet) VerifyCommit(chainID string, blockID BlockID, height i
 	}
 
 	talliedVotingPower := int64(0)
-	round := commit.Round()
+	commit.SetAddresses(vals.GetAddresses())
```
ebuchman: This one looks right, though we don't actually need it here because the only thing we do with the Vote is get the SignBytes

consensus/state_test.go
```diff
@@ -291,6 +294,51 @@ func TestStateFullRoundNil(t *testing.T) {
...
+
+	startTestRound(cs, height, round)
+
+	<-newRoundCh
+
+	// grab proposal
+	re := <-propCh
+	propBlockHash := re.(types.EventDataRoundState).RoundState.(*cstypes.RoundState).ProposalBlock.Hash()
+
+	<-voteCh // wait for prevote
```
jaekwon: All new tests should use the new ensureVote() format as defined in consensus/common_test.go (and we should migrate old tests to this new format).
Just receiving events from a channel like this without validating the height/round is an absolute nightmare to debug.
We need it for all subscriptions like for propCh and newRoundCh as well.

consensus/state_test.go
```diff
@@ -291,6 +294,51 @@ func TestStateFullRoundNil(t *testing.T) {
...
+	validatePrevote(t, cs, round, vss[0], propBlockHash)
+
+	<-voteCh // wait for precommit
+
+	// we're going to roll right into new height
+	<-newRoundCh
```
jaekwon: This doesn't actually test that we roll into a new height. (newRoundCh just ensures that we go to a new round).
We should actually test that we go to a new height... and I'm not sure that we need the other subscription tests above (since going to a new height pre-supposes prevotes and precommits, and this test is testing reloading... all this setup is making the point of the test less obvious.

types/vote_set.go
```diff
@@ -539,14 +539,18 @@ func (voteSet *VoteSet) MakeCommit() *Commit {
...
 	}
+	blockID := *voteSet.maj23
 
 	// For every validator, get the precommit
 	votesCopy := make([]*Vote, len(voteSet.votes))
 	copy(votesCopy, voteSet.votes)
-	return &Commit{
-		BlockID:    *voteSet.maj23,
-		Precommits: votesCopy,
+	precommits := make([]*CommitSig, len(voteSet.votes))
```
ebuchman: voteSet.Size()

docs/spec/blockchain/blockchain.md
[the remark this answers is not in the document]
jaekwon: I think it should be Height, and the method can be GetHeight().  Adding "Num" due to a method name conflict is surprising.  Adding a "Get" prefix is already convention.

consensus/reactor.go
ebuchman: Yes you can get them from the state! Let's make a function like `GetVoteSetFromCommit(blockStore, state) VoteSet`

types/validator_set.go
```diff
@@ -270,34 +270,27 @@ func (vals *ValidatorSet) VerifyCommit(chainID string, blockID BlockID, height i
...
+		precommitSignBytes := baseVote.SignBytes(chainID)
 		if !val.PubKey.VerifyBytes(precommitSignBytes, precommit.Signature) {
 			return fmt.Errorf("Invalid commit -- invalid signature: %v", precommit)
 		}
 		// Good precommit!
-		if blockID.Equals(precommit.BlockID) {
-			talliedVotingPower += val.VotingPower
-		} else {
-			// It's OK that the BlockID doesn't match.  We include stray
-			// precommits to measure validator availability.
```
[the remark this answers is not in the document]
ebuchman: @jaekwon just pointed out we might need more here, since currently we wouldn't be including votes for validators that correctly precommited for nil. 

We could rectify that by including precommits for nil (note this is distinct from nil precommits, ie. not receiving a precommit for anything from that validator), and including a bit array indicating which votes are for nil and which are for the block id

consensus/reactor.go
jlandrews: Mmk that makes sense, I can work that up.

ebuchman:
@jlandrews seems like there's still a bit of work to do here. Would you mind writing up a small ADR that describes what we're trying to achieve and how, and touches on some of the issues that came up in this PR? Would be good to have the rationale around this vote vs. commit stuff well documented.

consensus/state_test.go
jlandrews: Yeah, this test was basically just copied from the tests above, with the addition of doing the state reload. I'll read through the stuff you referenced in common_test and rework this to be more straightforward.

types/validator_set.go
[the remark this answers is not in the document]
jaekwon: I think we should include nil precommits, since precluding those would have adverse economic consequences and could lead to unexpected incentive distortions.

consensus/reactor.go
jaekwon: Could we include a Commit.Addresses ([]address) and Commit.IsNils (bitarray) that are in the first iteration of this software mandatory to include, and we can work on making them optional after launch?  This way we won't introduce any bugs like the above, at least until we start working on optimizing Commit further, but I'm not sure we even need to do that any time soon.  20*100 = 2KB, compared to the signatures it's not that much more, and it helps with debugging commits, so we'll want it for the time being anyways.

jlandrews:
Sure I can do that. That will likely add more latency to getting this actually merged, but better to have clarity around what we're doing than do it quickly but incorrectly.

ebuchman:
> Sure I can do that. That will likely add more latency to getting this actually merged, but better to have clarity around what we're doing than do it quickly but incorrectly.

That's ok, I don't think it's a rush to do this say even for the next release. If you think there's a smaller piece of this that should be merged sooner to make the total PR size smaller, that's cool, but this guy definitely deserves some clearer design records.

Can you link to https://github.com/tendermint/tendermint/issues/2226 as well in the ADR? I'm going to close the issue but worth mentioning.

ebuchman:
Closing for now, tracking in https://github.com/tendermint/tendermint/issues/1648#issuecomment-419952424. Let's see if we can bring this back in a series of smaller PRs, starting with an ADR

PR #2184, opened 2018-08-08
MaxBytes

melekes:
Refs #2035

* [ ] Wrote tests


consensus/state.go
```diff
@@ -1438,7 +1456,14 @@ func (cs *ConsensusState) addProposalBlockPart(msg *BlockPartMessage, peerID p2p
 	}
 	if added && cs.ProposalBlockParts.IsComplete() {
 		// Added and completed!
-		_, err = cdc.UnmarshalBinaryReader(cs.ProposalBlockParts.GetReader(), &cs.ProposalBlock, int64(cs.state.ConsensusParams.BlockSize.MaxBytes))
+		maxDataSize := cs.state.ConsensusParams.BlockSize.MaxTxsBytes
+		// header + evidence + last commit
+		const maxOtherPartsSize = 5000000 // 5 MB
```
melekes: this is ugly. and it should be lower

mempool/mempool.go
```diff
@@ -368,7 +383,8 @@ func (mem *Mempool) notifyTxsAvailable() {
 
 // Reap returns a list of transactions currently in the mempool.
 // If maxTxs is -1, there is no cap on the number of returned transactions.
-func (mem *Mempool) Reap(maxTxs int) types.Txs {
+// If maxTxsBytes is -1, there is no cap on the size of returned transactions.
+func (mem *Mempool) Reap(maxTxs, maxTxsBytes int) types.Txs {
```
melekes: having `maxTxs` is convenient for testing and `/unconfirmed_txs?limit=N` endpoint, but it can be removed still. But then I am not sure what to do with `/unconfirmed_txs?limit=N` endpoint. Should we limit by total bytes?

state/state_test.go
```diff
@@ -328,7 +328,7 @@ func TestConsensusParamsChangesSaveLoad(t *testing.T) {
 	params[0] = state.ConsensusParams
 	for i := 1; i < N+1; i++ {
 		params[i] = *types.DefaultConsensusParams()
-		params[i].BlockSize.MaxBytes += i
+		params[i].BlockSize.MaxTxsBytes += i
```
melekes: not sure what's going on here

mempool/mempool.go
ValarDragon: I think we could just make a separate `ReapTxs` function, which should only be used by the endpoint and testing. I don't think that sort of code repitition will hurt us, as long as we don't use it in consensus.

consensus/state.go
ValarDragon: We should also calculate the exact amino overhead. (Perhaps we make a method in amino to reflect through all the structs and tell you how big it would be if you were to marshall it)

mempool/mempool.go
melekes: Nice idea!

mempool/mempool.go
```diff
@@ -393,33 +393,34 @@ func (mem *Mempool) Reap(maxTxs, maxTxsBytes int) types.Txs {
 		time.Sleep(time.Millisecond * 10)
 	}
 
-	txs := mem.collectTxs(maxTxs, maxTxsBytes)
+	var cur int
+	txs := make([]types.Tx, 0, mem.txs.Len())
```
ValarDragon: I think we will get a performance boost if we have a good estimate of avg size per tx, and set the initial capacity based off of that. (We could under-esitmate avg size per tx a bit) In the case of max=-1, we could just call reapTxs(-1) for simplicity.

mempool/mempool.go
```diff
@@ -393,33 +393,34 @@ func (mem *Mempool) Reap(maxTxs, maxTxsBytes int) types.Txs {
...
-	if maxTxs == -1 {
-		maxTxs = mem.txs.Len()
+	for atomic.LoadInt32(&mem.rechecking) > 0 {
+		// TODO: Something better?
+		time.Sleep(time.Millisecond * 10)
 	}
 
-	txs := make([]types.Tx, 0, cmn.MinInt(mem.txs.Len(), maxTxs))
-	for e := mem.txs.Front(); e != nil && len(txs) < maxTxs; e = e.Next() {
+	txs := make([]types.Tx, 0, mem.txs.Len())
```
ValarDragon: Can we still do the min between mem.txs.Len() and max? (and keep the maxTxs == -1 check)

mempool/mempool.go
melekes: Indeed, but how big of a boost? It's better to make such optimizations after we know current performance profile. It may be that the time to calculate avg size will hurt us more (overweights gains).

ValarDragon: Great point!

consensus/state.go
```diff
@@ -1328,6 +1336,13 @@ func (cs *ConsensusState) finalizeCommit(height int64) {
 
 	fail.Fail() // XXX
 
+	// update mempool filter because consensus params might have changed
+	cs.mempool.SetFilter(func(tx types.Tx) bool {
```
ValarDragon: Is the idea behind using a filter here that we can more easily use any consensus param, rather than just the maxbytes? The only other thing I can imagine us doing prelaunch is enforcing a minfee. I think both of these would be more well suited inside of the mempool itself. This would also avoid us having to use the mutex on the consensus param everytime a new tx is added. I think this wall cause blocking issues, plus the defer is a performance hit.

ValarDragon:
Do you want to enforce max gas in this pr as well, or should we save that for a separate pr?

consensus/state.go
melekes: > more well suited inside of the mempool itself

but consensus params can be changed in any block (https://tendermint.com/docs/app-dev/abci-spec.html#endblock). you mean we should just copy them over if they change. That actually not a bad idea.

melekes:
> Do you want to enforce max gas in this pr as well, or should we save that for a separate pr?

separate PR

consensus/state.go
ebuchman: we should be able to be more precise, and we should do the opposite (ie. subtract from maxBlockSize to get maxDataSize)

melekes:
ADR: https://github.com/tendermint/tendermint/blob/b19bb3a51b76f088edab5a2cb44a87f40f4b065c/docs/architecture/adr-020-block-size.md

docs/architecture/adr-020-block-size.md
```diff
@@ -0,0 +1,65 @@
...
+  * MaxEvidenceBytes
+  * MaxLastCommitBytes
+4) Use these variables to limit the appropriate parts (header, evidence, ...)
+https://github.com/tendermint/tendermint/blob/8a1a79257e8e8b309cd35bb1fe40bf9b3330fd7d/types/block.go#L195
+
+Proposed default values are:
+
+- MaxHeaderBytes - 512 bytes (~200 bytes hashes + 200 bytes - 50 UTF-8 encoded symbols of chain ID)
+- MaxEvidenceBytes - 4096 bytes (10 DuplicateVoteEvidence ~350 bytes each)
+- MaxLastCommitBytes - 163840 (160 KB) (1000 votes ~150 bytes each)
```
ValarDragon: Not sure this one should be a parameter. Seems completely determined by number of validators and tendermint protocol right?

docs/architecture/adr-020-block-size.md
```diff
@@ -0,0 +1,65 @@
...
+3) Introduce Max\* variables for header, evidence and last commit:
+  * MaxHeaderBytes
+  * MaxEvidenceBytes
+  * MaxLastCommitBytes
+4) Use these variables to limit the appropriate parts (header, evidence, ...)
+https://github.com/tendermint/tendermint/blob/8a1a79257e8e8b309cd35bb1fe40bf9b3330fd7d/types/block.go#L195
+
+Proposed default values are:
+
+- MaxHeaderBytes - 512 bytes (~200 bytes hashes + 200 bytes - 50 UTF-8 encoded symbols of chain ID)
```
ValarDragon: This is probably my lack of familiarity, but why are there 200 bytes of hashes? I thought it would just be one merkelized hash at the root of the header.

docs/architecture/adr-020-block-size.md
```diff
@@ -0,0 +1,65 @@
...
+- MaxEvidenceBytes - 4096 bytes (10 DuplicateVoteEvidence ~350 bytes each)
```
ValarDragon: Why don't we make this maximum "MaxEvidencesPerBlock", and then derive the assoc. bytes directly from that.

docs/architecture/adr-020-block-size.md
melekes: Correct, but we still want to limit the block size. We know that TM currently does not support more than 300 validators (even less), so 1000 should be fine.

docs/architecture/adr-020-block-size.md
melekes: nah, tendermint header will be identical to ABCI header

docs/architecture/adr-020-block-size.md
melekes: Good idea

docs/architecture/adr-020-block-size.md
ValarDragon: Why not go the other way around?

i.e. Set a `MaxBlockSize`, and then derive tx size as follows:

```
MaxLastCommitBytes = function of number of validators currently enabled
MaxEvidenceBytes = function of consensus param: "maxNumEvidences"
MaxHeaderBytes = a consensus param
```
Then `MaxTxSize = MaxBlockSize - amino overhead - MaxLastCommitBytes - MaxEvidenceBytes - MaxHeaderBytes`. This makes MaxBlockSize the constant in consensus, not commit / tx size. (Max TxSize will decrease automatically as num validators increases, until block size is increased)

docs/architecture/adr-020-block-size.md
ValarDragon: Your right, thanks for the clarification! (I hadn't read through the ABCI header definition before, this makes sense now)

docs/architecture/adr-020-block-size.md
melekes: > function of consensus param: "maxNumEvidences"

and `MaxEvidenceBytes` - max size of a single evidence

> function of number of validators currently enabled

and `MaxVoteBytes`, right?

> This makes MaxBlockSize the constant in consensus, not commit / tx size. (Max TxSize would then decrease automatically as num validators increases)

Right, so it's either a) we have a sum and block size is growing when number of validators increasing OR b) we have a fixed `BlockSize.MaxBytes` and number of txs is decreasing while number of validators increases

ValarDragon: I prefer the latter, fixed block size and space for txs decreases as validators increases. Fixing the block size rather than tx data size feels more natural to me, and allows for better predictions of storage requirements, hence my preference for it. 

MaxEvidenceBytes and MaxVoteBytes are calculated constants, not part of the config right? If so, then yes.

melekes:
Updated ADR.

docs/architecture/adr-020-block-size.md
```diff
@@ -0,0 +1,66 @@
...
+not for the txs inside the block. There's extra amino overhead + the actual
+headers on top of the actual transactions + evidence + last commit.
+
+## Proposed solution
+
+Therefore, we should
+
+1) Get rid of MaxTxs.
+2) Rename MaxTxsBytes to MaxBytes.
+3) Add MaxHeaderBytes cs parameter (TODO: https://github.com/tendermint/tendermint/blob/8a1a79257e8e8b309cd35bb1fe40bf9b3330fd7d/types/block.go#L195)
```
ValarDragon: I'm confused as to why this is a consensus parameter.

AFAICT, everything here can be upper bounded from type definition:https://github.com/tendermint/tendermint/blob/8a1a79257e8e8b309cd35bb1fe40bf9b3330fd7d/types/block.go#L197-L222

All of the hashes are tmhashes, time.Time surely has a max size, we can bound chain-ID (256 bits b64 encoded? This is moreso dependent on what our plan for chain-ID is). We can use the max size for "int" in PartSetHeader (used within BlockID). So we should be able to derive the MaxHeaderBytes just from its definition. We can just amino encode it as well, so we don't have to guess at the amino overhead for the header.

melekes: Right. Thanks.

ValarDragon:
ADR looks great to me! Thanks for writing it!

abci/types/types_test.go
```diff
@@ -21,7 +21,7 @@ func TestConsensusParams(t *testing.T) {
 
 	// get values with real fields
 	assert.NotNil(params.GetBlockSize())
```
melekes: not sure I understand the point of this test

consensus/reactor_test.go
```diff
@@ -194,9 +194,9 @@ func newMockEvidencePool(val []byte) *mockEvidencePool {
 	}
 }
 
-func (m *mockEvidencePool) PendingEvidence() []types.Evidence {
+func (m *mockEvidencePool) PendingEvidence(limit int) []types.Evidence {
```
melekes: should be `uint` really, but not sure if it's the right time

evidence/pool.go
```diff
@@ -58,8 +58,8 @@ func (evpool *EvidencePool) PriorityEvidence() []types.Evidence {
 }
 
 // PendingEvidence returns all uncommitted evidence.
```
melekes: update description

evidence/store.go
```diff
@@ -87,8 +88,9 @@ func (store *EvidenceStore) PriorityEvidence() (evidence []types.Evidence) {
 }
 
 // PendingEvidence returns all known uncommitted evidence.
```
melekes: update description

mempool/mempool.go
```diff
@@ -57,6 +57,8 @@ var (
 	ErrMempoolIsFull = errors.New("Mempool is full")
 )
 
+// type Filter func(types.Tx) bool
```
melekes: remove

types/block.go
```diff
@@ -13,6 +13,14 @@ import (
 	cmn "github.com/tendermint/tendermint/libs/common"
 )
 
+const (
+	// MaxHeaderBytes is a maximum header size (including amino overhead).
+	MaxHeaderBytes = 476
+
+	// AminoOverheadForBlock - amino overhead to encode the block (whole block + data field).
+	AminoOverheadForBlock = 5
```
melekes: do we need a script to get these values?

melekes:
TODO: limit chain ID (since now we have MaxHeaderBytes, calculated with max 50 UTF-8 symbols) to 50 symbols max

consensus/state.go
```diff
@@ -153,6 +153,13 @@ func NewConsensusState(
 	cs.setProposal = cs.defaultSetProposal
 
 	cs.updateToState(state)
+
+	// filter mempool txs based on consensus params
+	maxBytes := cs.state.ConsensusParams.TxSize.MaxBytes
```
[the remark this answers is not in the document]
melekes: > better to concentrate the "changes due to a block commit" in one place

agree

You mean https://github.com/tendermint/tendermint/blob/d580006173994da163262c47a808aaabc5b3950b/state/execution.go#L148, right?

abci/types/types.proto
```diff
@@ -207,8 +207,8 @@ message ConsensusParams {
 // BlockSize contains limits on the block size.
 message BlockSize {
   int32 max_bytes = 1;
-  int32 max_txs = 2;
-  int64 max_gas = 3;
+  int64 max_gas = 2;
+  int32 max_num_evidences = 3;
```
ebuchman: do we need to expose this in here? maybe we can set some internal heuristic and let all just be bounded by `max_bytes` ?

consensus/state.go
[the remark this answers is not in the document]
ebuchman: yes - if there's more parameters the mempool will depend on beyond the height, we should pass those in there too

docs/architecture/adr-020-block-size.md
```diff
@@ -0,0 +1,67 @@
...
+
+Proposed default values are:
+
+- MaxNumEvidences - 10
+
+When we need to ReapMaxBytes from the mempool, we calculate the upper bound as follows:
+
+```
+MaxLastCommitBytes = {number of validators currently enabled} * {MaxVoteBytes}
+MaxEvidenceBytes = {MaxNumEvidences} * {MaxEvidenceBytes}
```
ebuchman: i feel like we can set an internal MaxEvidenceBytes and check if evidence is available, take up to that amount, and then subtract precisely the amount of evidence we're including rather than potentially leaving a lot of room here with every block (since most blocks will have no evidence)

docs/architecture/adr-020-block-size.md
```diff
@@ -0,0 +1,67 @@
...
+## Consequences
+
+### Positive
+
+* one way to limit the size of a block
+* more granular control over the size of block parts (header, evidence, ...)
+
+### Negative
+
+* Neccessity to configure additional variables
```
ebuchman: Will actually reduce the need to configure more variables potentially!

Also we should probably add something about gas in here so that we limit by both bytes and gas (which ever limit is reached first)

types/block.go
ebuchman: At the very least a test should fail if the Header changes and MaxHeaderBytes doesn't change with it.

ebuchman:
We should probably add something about gas to the ADR too so that we limit by both bytes and gas (which ever limit is reached first)

docs/architecture/adr-020-block-size.md
ValarDragon: In order to do that, we would have to acquire the evidence before reaping the txs, or do 2 tx reaps. I'm not opposed to either of those possibilities.

Setting `MaxEvidenceBytes` vs `MaxNumEvidences` is independent of that however. I do think `MaxNumEvidences` is the better variable in the config.

abci/types/types.proto
ValarDragon: What do you mean by this? Don't we want this to be a consensus parameter, and since it pertains to parts of a block size it belongs in this object?

If I understand correctly, your suggesting we have a function inside of the code that calculates `max_num_evidence = f(max_bytes)`. This seems harder to configure, and as a user of the protocol I would want an easy way to alter this function. (At which point it would then have to go back into the config)

docs/architecture/adr-020-block-size.md
melekes: > rather than potentially leaving a lot of room here with every block

Yeah, did not think of that. Thank you! Let's do that.

> In order to do that, we would have to acquire the evidence before reaping the txs, or do 2 tx reaps.

we can do it in one pass
https://github.com/tendermint/tendermint/blob/29a51a87a15710c27a991241df94db22dc9f072d/consensus/state.go#L957-L958

consensus/state.go
```diff
@@ -946,14 +947,25 @@ func (cs *ConsensusState) createProposalBlock() (block *types.Block, blockParts
 		return
 	}
 
+	maxBytes := cs.state.ConsensusParams.BlockSize.MaxBytes
+	// bound evidence to 1/10th of the block
+	evidence := cs.evpool.PendingEvidence(maxBytes / 10)
```
melekes: note: incoming block evidence is unbound

evidence/store.go
```diff
@@ -78,26 +78,34 @@ func NewEvidenceStore(db dbm.DB) *EvidenceStore {
...
+func (store *EvidenceStore) PendingEvidence(maxBytes int) (evidence []types.Evidence) {
+	return store.ListEvidence(baseKeyPending, maxBytes)
 }
 
-// ListEvidence lists the evidence for the given prefix key.
+// ListEvidence lists the evidence for the given prefix key up to maxBytes.
 // It is wrapped by PriorityEvidence and PendingEvidence for convenience.
-func (store *EvidenceStore) ListEvidence(prefixKey string) (evidence []types.Evidence) {
+// If maxBytes is -1, there's no cap on the size of returned evidence.
+func (store *EvidenceStore) ListEvidence(prefixKey string, maxBytes int) (evidence []types.Evidence) {
```
melekes: make private?

mempool/mempool.go
```diff
@@ -378,19 +393,39 @@ func (mem *Mempool) Reap(maxTxs int) types.Txs {
...
 
-// maxTxs: -1 means uncapped, 0 means none
-func (mem *Mempool) collectTxs(maxTxs int) types.Txs {
-	if maxTxs == 0 {
-		return []types.Tx{}
-	} else if maxTxs < 0 {
-		maxTxs = mem.txs.Len()
+// ReapMaxTxs reaps up to max transactions from the mempool.
+// If max is negative, function panics.
+func (mem *Mempool) ReapMaxTxs(max int) types.Txs {
```
melekes: uint?

types/block_test.go
```diff
@@ -118,7 +120,7 @@ func TestBlockMakePartSetWithEvidence(t *testing.T) {
 
 	partSet := MakeBlock(h, txs, commit, evList).MakePartSet(1024)
 	assert.NotNil(t, partSet)
-	assert.Equal(t, 3, partSet.Total())
+	assert.Equal(t, 2, partSet.Total())
```
melekes: ??

types/tx.go
```diff
@@ -11,6 +11,11 @@ import (
 	cmn "github.com/tendermint/tendermint/libs/common"
 )
 
+const (
+	// MaxAminoOverheadForTx - amino overhead to encode a transaction.
```
melekes: `maximum amino overhead` seems to range from 1 to 4 bytes

types/block.go
```diff
@@ -13,6 +13,15 @@ import (
 	cmn "github.com/tendermint/tendermint/libs/common"
 )
 
+const (
+	// MaxHeaderBytes is a maximum header size (including amino overhead).
+	MaxHeaderBytes = 478
+
+	// MaxAminoOverheadForBlock - maximum amino overhead to encode a block (up to
+	// MaxBlockSizeBytes in size).
```
melekes: not including it's parts (only varint len + fields without data)

types/block.go
```diff
@@ -13,6 +13,15 @@ import (
...
+	MaxAminoOverheadForBlock = 4
```
melekes: verify

types/tx.go
```diff
@@ -11,6 +11,12 @@ import (
 	cmn "github.com/tendermint/tendermint/libs/common"
 )
 
+const (
+	// MaxAminoOverheadForTx - maximum amino overhead to encode a transaction
+	// (ranges from 1 to 4 bytes).
+	MaxAminoOverheadForTx = 4
```
melekes: verify

types/block.go
melekes: I know. Should be 4 + 1 byte for each field (4) probably

> We'd have to calculate a maximum number of txs to create the MaxAminoOverheadForBlock

no. see what I've done with MaxAminoOverheadForTx. In short: when we add txs, we count len(tx)+MaxAminoOverheadForTx

melekes: It's still seems a bit complex. If you or anybody else have ideas how to simplify this stuff, I'd be happy to hear them

mempool/mempool.go
```diff
@@ -378,19 +393,39 @@ func (mem *Mempool) Reap(maxTxs int) types.Txs {
...
 
-	txs := mem.collectTxs(maxTxs)
+	var cur int
+	// TODO: we will get a performance boost if we have a good estimate of avg
+	// size per tx, and set the initial capacity based off of that.
+	// txs := make([]types.Tx, 0, cmn.MinInt(mem.txs.Len(), max/mem.avgTxSize))
+	txs := make([]types.Tx, 0, mem.txs.Len())
+	for e := mem.txs.Front(); e != nil; e = e.Next() {
+		memTx := e.Value.(*mempoolTx)
+		if max > 0 && cur+len(memTx.tx)+types.MaxAminoOverheadForTx > max {
```
melekes: it does not have to be Max. We can create a amino#EstimateSize(o interface{}) function

consensus/state.go
ebuchman: We should state this ratio more explicitly

mempool/mempool.go
```diff
@@ -378,19 +396,41 @@ func (mem *Mempool) Reap(maxTxs int) types.Txs {
...
+	// txs := make([]types.Tx, 0, cmn.MinInt(mem.txs.Len(), max/mem.avgTxSize))
+	txs := make([]types.Tx, 0, mem.txs.Len())
+	for e := mem.txs.Front(); e != nil; e = e.Next() {
+		memTx := e.Value.(*mempoolTx)
+		// amino.UvarintSize is not used here because it won't be possible to reuse buf
+		aminoOverhead := binary.PutUvarint(buf[:], uint64(len(memTx.tx)))
+		if max > 0 && cur+len(memTx.tx)+aminoOverhead > max {
+			return txs
+		}
+		cur += len(memTx.tx) + aminoOverhead
```
ebuchman: txs in the mempool are amino encoded ?

mempool/mempool.go
```diff
@@ -378,19 +396,41 @@ func (mem *Mempool) Reap(maxTxs int) types.Txs {
...
 }
 
-// maxTxs: -1 means uncapped, 0 means none
-func (mem *Mempool) collectTxs(maxTxs int) types.Txs {
-	if maxTxs == 0 {
-		return []types.Tx{}
-	} else if maxTxs < 0 {
-		maxTxs = mem.txs.Len()
+// ReapMaxTxs reaps up to max transactions from the mempool.
+// If max is negative, function panics.
```
ebuchman: why does this panic while the other returns unbounded?

mempool/mempool.go
```diff
@@ -411,6 +451,10 @@ func (mem *Mempool) Update(height int64, txs types.Txs) error {
 	mem.height = height
 	mem.notifiedTxsAvailable = false
 
+	if filter != nil {
+		mem.filter = filter
```
ebuchman: interesting

state/execution.go
```diff
@@ -145,7 +145,9 @@ func (blockExec *BlockExecutor) Commit(block *types.Block) ([]byte, error) {
 		"appHash", fmt.Sprintf("%X", res.Data))
 
 	// Update mempool.
-	if err := blockExec.mempool.Update(block.Height, block.Txs); err != nil {
+	maxBytes := state.ConsensusParams.TxSize.MaxBytes
+	filter := func(tx types.Tx) bool { return len(tx) <= maxBytes }
```
ebuchman: Maybe better to define this and export it.

types/block_test.go
```diff
@@ -233,6 +234,40 @@ func TestCommitValidateBasic(t *testing.T) {
...
+	for i := 0; i < MaxChainIDLen; i++ {
+		maxChainID += "𠜎"
+	}
+
+	h := Header{
+		ChainID:            maxChainID,
+		Height:             10,
+		Time:               time.Now().UTC(),
+		NumTxs:             100,
+		TotalTxs:           200,
```
ebuchman: Aren't these varints? What if we had `TotalTxs: 20000` - then the test fails because the header is 479 bytes

types/evidence_test.go
```diff
@@ -89,10 +93,27 @@ func TestEvidenceList(t *testing.T) {
...
 }
 
+func TestMaxEvidenceBytes(t *testing.T) {
+	val := NewMockPV()
+	blockID := makeBlockID(tmhash.Sum([]byte("blockhash")), 1000, tmhash.Sum([]byte("partshash")))
+	blockID2 := makeBlockID(tmhash.Sum([]byte("blockhash2")), 1000, tmhash.Sum([]byte("partshash")))
+	const chainID = "mychain"
+	ev := &DuplicateVoteEvidence{
+		PubKey: secp256k1.GenPrivKey().PubKey(), // use secp because it's pubkey is longer
+		VoteA:  makeVote(val, chainID, 0, 10, 2, 1, blockID),
```
ebuchman: same

node/node.go
```diff
@@ -248,6 +248,8 @@ func NewNode(config *cfg.Config,
 		mempl.WithMetrics(memplMetrics),
 	)
 	mempool.SetLogger(mempoolLogger)
+	maxBytes := state.ConsensusParams.TxSize.MaxBytes
+	mempool.SetFilter(func(tx types.Tx) bool { return len(tx) <= maxBytes })
```
ebuchman: should we make this a mempool option and set it on the constructor instead?

abci/types/types.proto
ebuchman: >  This seems harder to configure, and as a user of the protocol I would want an easy way to alter this function.

Do you really though? Seems like it should basically be as large as possible - if there's evidence, Tendermint should prioritize including it. Not clear that it's really something that needs to be exposed to users right now.

state/execution.go
ebuchman: Also do we want to allow a negative or zero MaxBytes to indicate unbounded?

ebuchman: It seems we're not checking TxSize.MaxBytes for txs in the block, just in the mempool. Technically that would make it not a consensus param, because a proposer running different mempool code could include txs larger than TxSize.MaxBytes

I wonder if we even need a TxSize.MaxBytes in the consensus params -  it might be simpler to just use the computed maxDataBytes. 

We'd still need `Update` to take a filter, but we'd be able to remove the TxSize configuration and just do it all through blocks.

ebuchman: > though I can't find a super convincing reason as to why.

Right. We have a lot of surface area. When we don't have a really good reason to have something, it's probably worth removing it. Helps simplify things over all and reduce the possibility for errors.

node/node.go
ValarDragon: I think it makes sense to add to the mempool config.

state/execution.go
melekes: > Maybe better to define this and export it.

That was an initial implementation, but the problem is in https://github.com/tendermint/tendermint/blob/fef6ab6b79dc5d8fdfa1307cee8cb394fc00f05d/state/services.go#L26 it's undesirable to import mempool package (just for `mempl.Filter`)

melekes: > We'd still need Update to take a filter, but we'd be able to remove the TxSize configuration and just do it all through blocks.

Should this be done along with enforcing MaxGas?

consensus/state.go
melekes: Spec for how we construct a block?

Issue #2180, opened 2018-08-07
Mempool: tx cache can be filled with invalid txs until memory is exhausted.


ValarDragon:
Found through hacker one!

In mempool.CheckTx() we add any incoming transaction to the mempool and cache, and after that we run check tx. CheckTx has a callback to delete the tx from the mempool and cache if check tx failed. However its not deleting it from the mempool's caches' linked list. This means that an attacker can spam garbage txs at the p2p layer, and they will clog up everyones mempool. Due to our default settings, each of these invalid Txs can be up to 1 MB, and the default mempool would take 100k of these. This means you can cause Out of Memory errors through the p2p layer. (As this allows for up to 100GB of RAM to be used up given enough time)

This can be quickly patched by making the callback delete from the linked list and write a test for callbacks deleting from all three locations. We should iterate from the back of the linked list to delete the element from there.

Additionally we should lower the default max tx size from 1 MB. I suggest 100KB. This should be a parameter in the config though.

Additionally we change the CheckTx functionality to only add it to the cache after checkTx passes. This could be done #postlaunch, but it needs to be done. If not, this same wasted space vector exists on txs that take forever to pass through checktx. (E.g. multisig with many accts) This attack vector is much less impactful, and likely would never crash a node, but no reason to let someone make you waste extra memory / computation.

melekes:
> However its not deleting it from the mempool's caches' linked list.

I raised this issue a long time ago. I will double check, but this should not be a problem since old txs get evicted when we push new txs (linked list has a fixed size)

> Additionally we change the CheckTx functionality to only add it to the cache after checkTx passes.

the sole purpose of the cache is to prevent the same txs entering `app.CheckTx(tx []byte)`

melekes:
And yes, we should enforce https://github.com/tendermint/tendermint/blob/d07eaab76ad294e20b6b907021db89a930fe21ec/types/params.go#L68

ValarDragon:
> the sole purpose of the cache is to prevent the same txs entering  `app.CheckTx(tx []byte)`

[...]

PR #2188, opened 2018-08-08
mempool: Keep cache hashmap and linked list in sync

ValarDragon:
This removes bugs with the linked list being full, but hashmap empty.

Ref #2180 

* [x] Wrote tests

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -2,7 +2,6 @@ package mempool
 
 import (
 	"bytes"
-	"container/list"
 	"fmt"
 	"sync"
 	"sync/atomic"
@@ -488,8 +487,8 @@ type txCache interface {
 type mapTxCache struct {
 	mtx  sync.Mutex
 	size int
-	map_ map[string]struct{}
-	list *list.List // to remove oldest tx when cache gets too big
+	map_ map[string]*clist.CElement
+	list *clist.CList // to remove oldest tx when cache gets too big
 }
 
 var _ txCache = (*mapTxCache)(nil)
@@ -498,15 +497,15 @@ var _ txCache = (*mapTxCache)(nil)
 func newMapTxCache(cacheSize int) *mapTxCache {
 	return &mapTxCache{
 		size: cacheSize,
-		map_: make(map[string]struct{}, cacheSize),
-		list: list.New(),
+		map_: make(map[string]*clist.CElement, cacheSize),
+		list: clist.New(),
 	}
 }
 
 // Reset resets the cache to an empty state.
 func (cache *mapTxCache) Reset() {
 	cache.mtx.Lock()
-	cache.map_ = make(map[string]struct{}, cache.size)
+	cache.map_ = make(map[string]*clist.CElement, cache.size)
 	cache.list.Init()
 	cache.mtx.Unlock()
 }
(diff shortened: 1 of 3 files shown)


mempool/mempool.go
```diff
@@ -528,16 +527,26 @@ func (cache *mapTxCache) Push(tx types.Tx) bool {
...
+	cache.map_[string(tx)] = cache.list.Back()
 	return true
 }
 
 // Remove removes the given tx from the cache.
 func (cache *mapTxCache) Remove(tx types.Tx) {
 	cache.mtx.Lock()
-	delete(cache.map_, string(tx))
+	stx := string(tx)
+	// TODO: Can we get value and index at the same time?
```
melekes: nope

ValarDragon: Aww okay. #postlaunch it may be worth experimenting with a "prehashed input hashmap. (Alot simpler than it sounds) and then we just hash once with the same fast hash.

ValarDragon:
The benefit of this change is space savings + fixing bugs with cache eviction. Computation time will increase due to usage of the clist, so making a benchmark is a good idea. 

Is there a good way to benchmark size? Just println unsafe size of?

melekes:
`go test -memprofile`?

ValarDragon:
:o cool, didn't know about that

ValarDragon:
before:
```
BenchmarkCacheInsertTime-8   	 5000000	       385 ns/op	      88 B/op	       3 allocs/op
BenchmarkCacheRemoveTime-8   	10000000	       152 ns/op	       0 B/op	       0 allocs/op
```
after:
```
BenchmarkCacheInsertTime-8   	 1000000	      1027 ns/op	     360 B/op	       7 allocs/op
BenchmarkCacheRemoveTime-8   	 2000000	       860 ns/op	     112 B/op	       2 allocs/op
```
So it looks like we're getting a performance hit due to the concurrent list. However this performance hit is neglible compared to the signature verification of saving a checkTx, and we are now saving memory from elements we remove from the cache. (The before case didn't remove it at all from the linked list. If we used the O(N) method to do that properly, it would certainly be slower)

melekes:
> However this performance hit is neglible compared to the signature verification of saving a checkTx, and we are now saving memory from elements we remove from the cache.

I tend to agree. Plus, if there are usually no "bad" txs in the network (e.g. private deployment), cache can be always disabled.

mempool/mempool.go
```diff
@@ -488,8 +487,8 @@ type txCache interface {
 type mapTxCache struct {
 	mtx  sync.Mutex
 	size int
-	map_ map[string]struct{}
-	list *list.List // to remove oldest tx when cache gets too big
+	map_ map[string]*clist.CElement
+	list *clist.CList // to remove oldest tx when cache gets too big
```
ebuchman: is there some reason we needed to switch to Clist?

mempool/mempool.go
```diff
@@ -524,20 +523,26 @@ func (cache *mapTxCache) Push(tx types.Tx) bool {
 	if cache.list.Len() >= cache.size {
 		popped := cache.list.Front()
 		poppedTx := popped.Value.(types.Tx)
-		// NOTE: the tx may have already been removed from the map
-		// but deleting a non-existent element is fine
```
ebuchman: so we knew we weren't removing from the list in .Remove but we thought it was ok?

mempool/mempool.go
ValarDragon: Yes. CheckTxs are called in parallel, and can have different execution times. I can describe problematic situations more if you'd like. (I ran through a couple in my head that I thought would eventually cause big problems before making the clist switch)

mempool/mempool.go
ValarDragon: I'm guessing noone considered the lingering memory or the issues with how it affects cache eviction of good txs. (Full linked list, empty hashmap)

mempool/mempool.go
ebuchman: But is there some new requirement that only now we need a clist? or you think we always needed it?

CheckTx actually don't run in parallel, they run sequentially, but responses come back asynchronously (though in order)

ValarDragon: Oh your right, we don't need it here. The reason we don't need it (and something I failed to consider earlier) is that we have a mutex on all deletions and additions. I forgot that we needed that for the hashmap still. I'll switch it back, thanks for pointing this out :D

ValarDragon:
Now that we're back to normal lists, here are the new benchmarks:
```
BenchmarkCacheInsertTime-8   	 5000000	       398 ns/op	      88 B/op	       3 allocs/op
BenchmarkCacheRemoveTime-8   	10000000	       179 ns/op	       0 B/op	       0 allocs/op
```
basically just adds 20ns to removal time.

PR #2195, opened 2018-08-09
Peng/lint md

nylira:
I've added two lint rules to the documentation linter. `write-good` checks the documentation for well-written English. `stop-words` prevents us from using words like `investor` or `atom distribution` in writing.

I've also run `yarn lint` in the `./docs` folder, so a lot of documentation files have been updated to follow a standardized format based on `Prettier`'s rules.

Closes https://github.com/tendermint/tendermint.com/issues/172

@zramsay once this PR goes through we should do the same for the Cosmos SDK docs.

Here are some sample `write-good` suggestions. They're set to `warning` and not `error` so we can  fix the documentation over time.

screen shot 2018-08-09 at 2 02 14 pm


docs/architecture/adr-017-chain-versions.md
```diff
@@ -91,10 +90,4 @@ When a chain forks or upgrades but continues the same history, it takes a new Ch
...
 ```
 
-Where
-    - ChainID is the ChainID from the previous ChainDescription (ie. its hash)
-    - `x` denotes that a change occured
-    - `Height` is the height the change occured
-    - ForkDescription has the same form as ChainDescription but for the fork
-        - this allows forks to specify new versions for tendermint or the app, as well as arbitrary changes to the state or validator set
-
+Where - ChainID is the ChainID from the previous ChainDescription (ie. its hash) - `x` denotes that a change occured - `Height` is the height the change occured - ForkDescription has the same form as ChainDescription but for the fork - this allows forks to specify new versions for tendermint or the app, as well as arbitrary changes to the state or validator set
```
ValarDragon: I'm confused, we want this to be a list

docs/spec/blockchain/encoding.md
```diff
@@ -48,33 +48,33 @@ spec](https://github.com/tendermint/go-amino#computing-the-prefix-and-disambigua
...
 ( while || stands for byte concatenation here).
 
-| Type | Name | Prefix | Length | Notes |
-| ---- | ---- | ------ | ----- | ------ |
-| PubKeyEd25519 | tendermint/PubKeyEd25519 | 0x1624DE64 | 0x20 |  |
-| PubKeySecp256k1 | tendermint/PubKeySecp256k1 | 0xEB5AE987 | 0x21 |  |
-| PrivKeyEd25519 | tendermint/PrivKeyEd25519 | 0xA3288910 | 0x40 |  |
-| PrivKeySecp256k1 | tendermint/PrivKeySecp256k1 | 0xE1B0F79B | 0x20 |  |
-| SignatureEd25519 | tendermint/SignatureEd25519 | 0x2031EA53 | 0x40 |  |
+| Type               | Name                          | Prefix     | Length   | Notes |
```
ValarDragon: I think we want this to be spaced out in the same way as our output right? Though I guess we could just put this into a code block

docs/architecture/adr-017-chain-versions.md
nylira: Yeah, the Prettier linter formatted the list into a paragraph because the word `Where` is breaking the parser. We'd have to fix this one manually.

docs/spec/blockchain/encoding.md
nylira: I'm not sure what you mean. The automatic table formatting seems correct in this case?

ValarDragon: We want the output of this to match what the code returns exactly (I assume)


If so, this function has the sample code output 
https://github.com/tendermint/tendermint/blob/013b9cef642f875634c614019ab13b17570778ad/crypto/encoding/amino/encode_test.go#L50

docs/architecture/adr-016-protocol-versions.md
```diff
@@ -58,19 +58,15 @@ to connect to peers with older version.
 
 ### BlockVersion
 
-- All tendermint hashed data-structures (headers, votes, txs, responses, etc.).
-	- Note the semantic meaning of a transaction may change according to the AppVersion,
-		but the way txs are merklized into the header is part of the BlockVersion
-- It should be the least frequent/likely to change.
-	- Tendermint should be stabilizing - it's just Atomic Broadcast.
-	- We can start considering for Tendermint v2.0 in a year
+- All tendermint hashed data-structures (headers, votes, txs, responses, etc.). - Note the semantic meaning of a transaction may change according to the AppVersion,
```
melekes: we should either remove `-` or add a newline.

melekes:
Also, should we run documentation linter on CircleCI?

docs/architecture/adr-016-protocol-versions.md
```diff
@@ -58,19 +58,15 @@ to connect to peers with older version.
...
+  but the way txs are merklized into the header is part of the BlockVersion
+- It should be the least frequent/likely to change. - Tendermint should be stabilizing - it's just Atomic Broadcast. - We can start considering for Tendermint v2.0 in a year
```
ebuchman: same, seems maybe these shouldnt have been collapsed?

docs/spec/reactors/block_sync/reactor.md
```diff
@@ -46,256 +46,256 @@ type bcStatusResponseMessage struct {
 
 ## Architecture and algorithm
 
-The Blockchain reactor is organised as a set of concurrent tasks: 
-    - Receive routine of Blockchain Reactor 
-    - Task for creating Requesters
-    - Set of Requesters tasks and 
-    - Controller task.        
+The Blockchain reactor is organised as a set of concurrent tasks: - Receive routine of Blockchain Reactor - Task for creating Requesters - Set of Requesters tasks and - Controller task.
```
ebuchman: why does this keep happening?

docs/spec/reactors/block_sync/reactor.md
```diff
@@ -46,256 +46,256 @@ type bcStatusResponseMessage struct {
...
 
-Requester data structure is used to track assignment of request for `block` at position `height` to a 
-peer with id equals to `peerID`. 
+Requester data structure is used to track assignment of request for `block` at position `height` to a
+peer with id equals to `peerID`.
 
 ```go
 type Requester {
-  mtx          Mutex		
+  mtx          Mutex
```
ebuchman: oh man this spec needs to be re-written. Way too much implementation specific detail in here

docs/spec/software/abci.md
```diff
@@ -61,12 +60,9 @@ message PubKey {
   string type = 1;
   bytes  data = 2;
 }
-
 ```
 
-The `pub_key` currently supports two types:
-    - `type = "ed25519" and `data = <raw 32-byte public key>`
-    - `type = "secp256k1" and `data = <33-byte OpenSSL compressed public key>`
+The `pub_key` currently supports two types: - `type = "ed25519" and`data = <raw 32-byte public key>`-`type = "secp256k1" and `data = <33-byte OpenSSL compressed public key>`
```
ebuchman: ah

nylira:
> Also, should we run documentation linter on CircleCI?

Yes, we can run the lint step on CircleCI. We'd have to add something like this to `.circleci/config.yml`:

```
  lint-docs:
    docker:
      # specify the version you desire here
      - image: circleci/node:9.4.0-browsers

    working_directory: ~/repo

    steps:
      - checkout

      # Download and cache dependencies
      - restore_cache:
          keys:
          - v1-dependencies-{{ checksum "package.json" }}
          # fallback to using the latest cache if no exact match is found
          - v1-dependencies-

      - run: yarn

      - save_cache:
          paths:
            - yarn.lock
            - node_modules
          key: v1-dependencies-{{ checksum "package.json" }}

      - run: yarn lint
```

> @nylira how do I run this linter? write-good *.md? Could you create a Makefile target?

This is how you run the linter:

```
cd docs
yarn
yarn lint
```

I could use help with these two tasks because I'm not too familiar with CircleCI or Makefiles.

zramsay:
lets get this in? see also https://github.com/cosmos/cosmos-sdk/issues/1655 and https://github.com/tendermint/devops/issues/62

docs/tendermint-core/running-in-production.md
```diff
@@ -135,33 +135,34 @@ Tendermint, replay will fail with panic.
 
 Recovering from data corruption can be hard and time-consuming. Here are two approaches you can take:
 
-1) Delete the WAL file and restart Tendermint. It will attempt to sync with other peers.
-2) Try to repair the WAL file manually:
+1. Delete the WAL file and restart Tendermint. It will attempt to sync with other peers.
+2. Try to repair the WAL file manually:
 
-  1. Create a backup of the corrupted WAL file:
+1) Create a backup of the corrupted WAL file:
```
melekes: shouldn't this be `1.`?

docs/tendermint-core/running-in-production.md
```diff
@@ -135,33 +135,34 @@ Tendermint, replay will fail with panic.
...
 
 ```
 cp "$TMHOME/data/cs.wal/wal" > /tmp/corrupted_wal_backup
 ```
 
-  2. Use `./scripts/wal2json` to create a human-readable version
+2. Use `./scripts/wal2json` to create a human-readable version
```
melekes: or this be `2)`?

Issue #1954, opened 2018-07-11
Revive json2wal tool


melekes:
Refs https://github.com/tendermint/tendermint/pull/1934#discussion_r201574829

I am not sure we will be able to find it in the Git history, but anyway should be easy to write it from the ground up.

ebuchman:
json2wal? You mean wal2json? It's still in `scripts`

melekes:
No, I mean `json2wal`. A script that turns json into binary wal (opposite of `wal2json`).

ebuchman:
What's the use case again ?

ebuchman:
Oh so we can compose test wals?

melekes:
1) tool is mentioned in https://github.com/tendermint/tendermint/blob/e398962d873d5474e3ce0f4ba3eaed26b670fbf3/docs/tendermint-core/running-in-production.md#wal-corruption it's supposed to help user restore corrupted WAL file
2) compose test wals

melekes:
Merged to develop.

PR #2196, opened 2018-08-09
add json2wal & fix wal2json

bradyjoestar:
I am try to solve the issue #1954 

In wal2json file , it use json.Marshal to convert msg to json.

If We want to unMarshal json to TimedWALMessage,we'd better use go-amino to convert msg to json file.

* [Y] Updated all relevant documentation in docs
* [Y ] Updated all code comments where relevant
* [] Wrote tests


scripts/json2wal/main.go
```diff
@@ -0,0 +1,75 @@
...
+func init() {
+	cs.RegisterConsensusMessages(cdc)
+	cs.RegisterWALMessages(cdc)
+	types.RegisterBlockAmino(cdc)
+}
+
+
+func main() {
+	if len(os.Args) < 3 {
+		fmt.Println("missing arguments: Usage:json2wal <path-to-JSON>  <path-to-wal>")
```
melekes: `fmt.Fprintln(os.Stderr, "missing`

scripts/json2wal/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+	for {
+		msgJson, _, err := br.ReadLine()
+		if err == io.EOF{
+			break
+		}else if err != nil {
+			panic(fmt.Errorf("failed to read file: %v", err))
+		}
+		// ignore the ENDHEIGHT in json.File
+		if strings.HasPrefix(string(msgJson),"ENDHEIGHT"){
```
melekes: why do we ignore it? it should be present in the output file, no?

scripts/json2wal/main.go
```diff
@@ -0,0 +1,75 @@
...
+		os.Exit(1)
+	}
+
+	f, err := os.Open(os.Args[1])
+	if err != nil {
+		panic(fmt.Errorf("failed to open WAL file: %v", err))
+	}
+	defer f.Close()
+
+	walFile, err := os.OpenFile(os.Args[2],os.O_APPEND|os.O_WRONLY|os.O_CREATE,0666)
```
melekes: don't think we should append. better crash if file already exists - `O_EXCL`

scripts/json2wal/main.go
bradyjoestar: ok,I will update.

bradyjoestar: ok.

scripts/json2wal/main.go
bradyjoestar: ok.

scripts/json2wal/main.go
bradyjoestar: I found the following code in the wal2json.file
```
if err == nil {
			if end, ok := msg.Msg.(cs.EndHeightMessage); ok {
				_, err = os.Stdout.Write([]byte(fmt.Sprintf("ENDHEIGHT %d\n", end.Height))) // nolint: errcheck, gas
			}
		}
```
so I ignore the "ENDHEIGHT"

scripts/json2wal/main.go
```diff
@@ -39,7 +39,7 @@ func main() {
 	}
 	defer f.Close()
 
-	walFile, err := os.OpenFile(os.Args[2],os.O_APPEND|os.O_WRONLY|os.O_CREATE,0666)
+	walFile, err := os.OpenFile(os.Args[2],os.O_EXCL|os.O_WRONLY,0666)
```
melekes: why did you remove `O_CREATE`?

scripts/json2wal/main.go
melekes: Have you tried converting existing wal to json, converting json to wal and running it? Does it work?

It should be more like:

```go
if strings.HasPrefix(string(msgJson),"ENDHEIGHT"){
h := extractHeight(string(msgJson))
err = dec.Encode(&cs.EndHeightMessage{h})
		if err != nil{
			panic(fmt.Errorf("failed to encode msg: %v", err))
		}
}
```

scripts/json2wal/main.go
bradyjoestar: oh,my careless

scripts/json2wal/main.go
bradyjoestar: Yeah, I use 'md5sum' command to check the hash of existing wal and generated wal on ubuntu,and it matched.I will upload my wal right now.

bradyjoestar: If there are EndHeightMessage in json file， go-amino will help us  write it into generated wal.
I just ignore the prefix of 'ENDHEIGHT',not 
```
{"time":"2018-08-08T11:28:00.372359633Z","msg":{"type":"tendermint/wal/EndHeightMessage","value":{"height":"2"}}}

```

bradyjoestar: generated json file slice
```
{"time":"2018-08-08T11:28:00.285064707Z","msg":{"type":"tendermint/wal/EventDataRoundState","value":{"height":"2","round":"0","step":"RoundStepCommit"}}}
{"time":"2018-08-08T11:28:00.372359633Z","msg":{"type":"tendermint/wal/EndHeightMessage","value":{"height":"2"}}}
ENDHEIGHT 2
```

melekes: Oh, right. Awesome. Thank you

melekes:
I think you've accidentally checked-in scripts/wal2JsonTest binary

ValarDragon:
Want to add an entry to the changelog? (Just to let ppl know its back)

melekes:
I'll add a changelog entry myself.

PR #2198, opened 2018-08-09
Remove gogoproto from Makefile's TOOLS

melekes:


melekes:
> any important changes between protobuf 3.4.1 and 3.6.1?

Not for Golang. For other languages, yes.

PR #2199, opened 2018-08-09
libs: Remove usage of custom Fmt, in favor of fmt.Sprintf

### LINKED ISSUE #2193: libs/common: Remove Fmt
Currently we have the following function in `libs/common/string.go`, which is actually used throughout our codebase:

```
// Like fmt.Sprintf, but skips formatting if args are empty.
var Fmt = func(format string, a ...interface{}) string {
	if len(a) == 0 {
		return format
	}
	return fmt.Sprintf(format, a...)
}
```
This doesn't make sense to use for efficiency purports. You should never be calling `fmt.Sprintf` if you have no arguments. (Nil is an argument that gets formatted). (This is the exact reason theres `fmt.Printf` and `Println`, and `errors.New` and `fmt.Errorf`) Therefore there isn't a good use case for this to replace. We should just use `fmt.Sprintf` directly and avoid the indirection.

There isn't really a conflation between when to use `fmt.Sprint` and `fmt.Sprintf` (one is format string, the other isn't). There may be some confusion between `fmt.Errorf` and `errors.New` (hence this being an explicit thing golint warns about), however this should be taken care of within our errors library, and not have an assoc. publically exported API which is used directly throughout our codebase.

ValarDragon:
Closes #2193

* [X] Wrote tests - n/a

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -9,6 +9,7 @@ BREAKING CHANGES:
   top-level
 - [abci] Added address of the original proposer of the block to Header.
 - [abci] Change ABCI Header to match Tendermint exactly
+- [libs] Remove cmn.Fmt, in favor of fmt.Sprintf
 
 FEATURES:
 
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -244,7 +244,7 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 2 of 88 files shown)


ValarDragon:
As further evidence that this is a good idea, there was a message which didn't have format string chars matched, but now that we've switched to the normal method this was caught at compile time:
`consensus/state.go:1088: Sprintf format % is missing verb at end of string`

PR #2205, opened 2018-08-10
crypto: Remove unnecessary prefixes from amino route variable names

ValarDragon:
*REBASE ONTO DEVELOP ONCE #2164 IS MERGED*

* [x] Updated CHANGELOG_PENDING.md - should this be updated? I don't think anyone was using these variables.


melekes:
> Updated CHANGELOG_PENDING.md - should this be updated? I don't think anyone was using these variables.

you never know. let's add an entry

ebuchman:
I got this

ebuchman:
Should we do the same thing for eg. `ed25519.PubKeyEd25519` -> `ed25519.PubKey` ?

ValarDragon:
I was considering that. (I actually already have that commit in my local env as well) The reason I didn't is that it may make some amino debugging harder. AFAICT, amino prints the struct name, not the path to it. (So you wouldn't see the `ed25519`)

PR #2208, opened 2018-08-11
clist: Speedup functions

ValarDragon:
We used unnecessary function calls, defers, and extra mutexes.
These are not our friends for writing fast code in our libs.

`defer()` was removed because these functions only had one or two code paths, so the minute increase in code complexity seems justified to me. I do think defers should be used outside of libs or in libs whenever there are several code paths. 

Before:
```
BenchmarkDetaching-8   	 5000000	       263 ns/op	       0 B/op	       0 allocs/op
BenchmarkRemoved-8     	20000000	        98.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkPushBack-8    	 2000000	       637 ns/op	     328 B/op	       5 allocs/op
```
After:
```
BenchmarkDetaching-8   	20000000	       111 ns/op	       0 B/op	       0 allocs/op
BenchmarkRemoved-8     	20000000	        61.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkPushBack-8    	 3000000	       438 ns/op	     328 B/op	       5 allocs/op
```
Originally this commit was just for detaching, then I realized defers were used in the same fashion throughout the file. I think the benchmarks for detaching are indicative that this is a good idea, but if desired I can benchmark more of the functions. `BenchmarkRemoved` and `BenchmarkPushback` should be indicative of the gains experienced by most of the functions with this change. 

* [X] Updated all relevant documentation in docs - nothing changes as a user of the API
* [ ] Updated all code comments where relevant - anything that warrants a comment?
* [X] Wrote tests
* [ ] Updated CHANGELOG_PENDING.md - do we update changelog for this sort of stuff?

### DIFF
--- a/libs/clist/clist.go
+++ b/libs/clist/clist.go
@@ -115,36 +115,36 @@ func (e *CElement) Next() *CElement {
 // Nonblocking, may return nil if at the end.
 func (e *CElement) Prev() *CElement {
 	e.mtx.RLock()
-	defer e.mtx.RUnlock()
-
-	return e.prev
+	prev := e.prev
+	e.mtx.RUnlock()
+	return prev
 }
 
 func (e *CElement) Removed() bool {
 	e.mtx.RLock()
-	defer e.mtx.RUnlock()
-
-	return e.removed
+	isRemoved := e.removed
+	e.mtx.RUnlock()
+	return isRemoved
 }
 
 func (e *CElement) DetachNext() {
-	if !e.Removed() {
+	e.mtx.Lock()
+	if !e.removed {
+		e.mtx.Unlock()
 		panic("DetachNext() must be called after Remove(e)")
 	}
-	e.mtx.Lock()
-	defer e.mtx.Unlock()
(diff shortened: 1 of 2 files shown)


libs/clist/clist.go
```diff
@@ -115,43 +115,42 @@ func (e *CElement) Next() *CElement {
...
+	isRemoved := e.removed
+	e.mtx.RUnlock()
+	return isRemoved
 }
 
 func (e *CElement) DetachNext() {
-	if !e.Removed() {
+	e.mtx.Lock()
+	if !e.removed {
+		e.mtx.Unlock()
```
melekes: should we even bother unlocking before `panic`?

ValarDragon: I'm not sure. It depends on if we plan on ever catching panics. My thinking was that this may come up in testing, and may just be annoying to have it lock. (i.e. `require.Panics(t, <invalid thing>)`)

PR #2212, opened 2018-08-13
fix #1544

### LINKED ISSUE #1544: Pass in NodeKey to NewNode
Yes please Tendermint:

> `// TODO: pass in like privValidator`

https://github.com/tendermint/tendermint/blob/master/node/node.go#L392

We have tooling to centralise the management of our validator keys within a pool, it would be useful to us if we could provide `NodeKey` in a similar way we can provide `PrivValidator`. Though I can see it will not really be possible to use just a `Signer` interface since you use the private key directly for the STS protocol stuff. 

Since these keys are now needed to set up seeds when authenticated encryption is enabled it would be much better for us if we could handle the generation and provisioning of these and pass them in to `NewNode` rather than having to write them to a volume.

bradyjoestar:
fix #1544 
```
	// Generate node PrivKey
	// TODO: pass in like privValidator
	nodeKey, err := p2p.LoadOrGenNodeKey(n.config.NodeKeyFile())
	if err != nil {
		return err
	}
```

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -119,6 +119,7 @@ type Node struct {
 	// network
 	sw       *p2p.Switch  // p2p connections
 	addrBook pex.AddrBook // known peers
+	nodeKey  *p2p.NodeKey // our node privkey
 
 	// services
 	eventBus         *types.EventBus // pub/sub for services
@@ -294,6 +295,13 @@ func NewNode(config *cfg.Config,
 	sw.AddReactor("CONSENSUS", consensusReactor)
 	sw.AddReactor("EVIDENCE", evidenceReactor)
 
+	// Generate node PrivKey
+	nodeKey, err := p2p.LoadOrGenNodeKey(config.NodeKeyFile())
+	if err != nil {
+		return nil,err
+	}
+	p2pLogger.Info("P2P Node ID", "ID", nodeKey.ID(), "file", config.NodeKeyFile())
+
 	// Optionally, start the pex reactor
 	//
 	// TODO:
@@ -392,6 +400,7 @@ func NewNode(config *cfg.Config,
 
 		sw:       sw,
 		addrBook: addrBook,
+		nodeKey: nodeKey,
 
 		stateDB:          stateDB,
 		blockStore:       blockStore,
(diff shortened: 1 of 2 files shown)


node/node.go
```diff
@@ -294,6 +295,13 @@ func NewNode(config *cfg.Config,
 	sw.AddReactor("CONSENSUS", consensusReactor)
 	sw.AddReactor("EVIDENCE", evidenceReactor)
 
+	// Generate node PrivKey
+	nodeKey, err := p2p.LoadOrGenNodeKey(config.NodeKeyFile())
```
melekes: this needs to be moved into `DefaultNewNode`

```
nodeKey, err := p2p.LoadOrGenNodeKey(config.NodeKeyFile())
if err != nil {
  return nil, err 
}
NewNode(config,
		privval.LoadOrGenFilePV(config.PrivValidatorFile())
                 nodeKey,
```

log statement should stay here

melekes:
Note this is technically breaking

node/node.go
bradyjoestar: Roger that!

bradyjoestar:
I will check the error and review the code more careful ,then submit pr.

node/node.go
```diff
@@ -80,8 +80,14 @@ type NodeProvider func(*cfg.Config, log.Logger) (*Node, error)
 // PrivValidator, ClientCreator, GenesisDoc, and DBProvider.
 // It implements NodeProvider.
 func DefaultNewNode(config *cfg.Config, logger log.Logger) (*Node, error) {
+	// Generate node PrivKey
+	nodeKey, err := p2p.LoadOrGenNodeKey(config.NodeKeyFile())
+	if err != nil {
+		return nil,err
```
melekes: please format https://golang.org/cmd/gofmt/

bradyjoestar: formatted, thanks~

melekes: you forgot about node/node.go

melekes:
can you add a changelog entry to the CHANGELOG_PENDING.md? `- [node] NewNode now accepts a *p2p.NodeKey`

melekes:
I'll do it myself.

melekes:
Thank you for contributing to Tendermint Core! 🍏

PR #2213, opened 2018-08-13
blockchain: fix register concrete name.

peerlink:
* [x] Wrote tests


ebuchman:
:o thanks for catching this

ValarDragon:
Did this get into a release? If so we probably should add this to breaking changes right?

ebuchman:
Fixed in https://github.com/tendermint/tendermint/pull/2224/files

PR #2219, opened 2018-08-14
// TODO: Print out all the sub-commands available

bradyjoestar:
* [ ] Wrote tests

### DIFF
--- a/abci/cmd/abci-cli/abci-cli.go
+++ b/abci/cmd/abci-cli/abci-cli.go
@@ -479,6 +479,14 @@ func muxOnCommands(cmd *cobra.Command, pArgs []string) error {
 func cmdUnimplemented(cmd *cobra.Command, args []string) error {
 	msg := "unimplemented command"
 
+	if len(args) > 0 {
+		msg += fmt.Sprintf(" args: [%s]", strings.Join(args, " "))
+	}
+	printResponse(cmd, args, response{
+		Code: codeBad,
+		Log:  msg,
+	})
+
 	fmt.Println("Available commands:")
 	fmt.Printf("%s: %s\n", echoCmd.Use, echoCmd.Short)
 	fmt.Printf("%s: %s\n", infoCmd.Use, infoCmd.Short)
@@ -487,14 +495,8 @@ func cmdUnimplemented(cmd *cobra.Command, args []string) error {
 	fmt.Printf("%s: %s\n", queryCmd.Use, queryCmd.Short)
 	fmt.Printf("%s: %s\n", commitCmd.Use, commitCmd.Short)
 	fmt.Printf("%s: %s\n", setOptionCmd.Use, setOptionCmd.Short)
+	fmt.Println("Use \"[command] --help\" for more information about a command.")
 
-	if len(args) > 0 {
-		msg += fmt.Sprintf(" args: [%s]", strings.Join(args, " "))
-	}
-	printResponse(cmd, args, response{
-		Code: codeBad,
-		Log:  msg,
-	})
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


melekes:
Hi! Thank you for your contribution! 

Is there an issue? Could you expand on what this PR does?

bradyjoestar:
Sure.You are welcome,it's my pleasure.

no,there isn't a issue.
I searched all the "TODO" in the code and found it in "abci-cli.go"

The following picture is we use `abci-cli console` and then executed `info` and `nana`(unimplement command):



I think the `TODO: Print out all the sub-commands available` want to print the all available command in the console.
```
func cmdUnimplemented(cmd *cobra.Command, args []string) error {
	// TODO: Print out all the sub-commands available
	msg := "unimplemented command"
	if err := cmd.Help(); err != nil {
		msg = err.Error()
	}
	if len(args) > 0 {
		msg += fmt.Sprintf(" args: [%s]", strings.Join(args, " "))
	}
	printResponse(cmd, args, response{
		Code: codeBad,
		Log:  msg,
	})
	return nil
}
```

so I changed it, after changed:

abci/cmd/abci-cli/abci-cli.go
```diff
@@ -477,9 +503,8 @@ func muxOnCommands(cmd *cobra.Command, pArgs []string) error {
 }
 
 func cmdUnimplemented(cmd *cobra.Command, args []string) error {
-	// TODO: Print out all the sub-commands available
 	msg := "unimplemented command"
-	if err := cmd.Help(); err != nil {
+	if err := muxonCmd.Help(); err != nil {
```
melekes: I think we can do it without any additional commands. What if we just cycle through available commands and print their use and short description

```
fmt.Println("Available commands:")
fmt.Printf("%s: %s\n", echoCmd.Use, echoCmd.Short)
...
```

bradyjoestar: I had made it more complicated.  lol

Thank you!
By the way，I'm reading the code of consensus which is really awesome!!

abci/cmd/abci-cli/abci-cli.go
```diff
@@ -477,11 +477,17 @@ func muxOnCommands(cmd *cobra.Command, pArgs []string) error {
 }
 
 func cmdUnimplemented(cmd *cobra.Command, args []string) error {
-	// TODO: Print out all the sub-commands available
 	msg := "unimplemented command"
-	if err := cmd.Help(); err != nil {
-		msg = err.Error()
-	}
+
+	fmt.Println("Available commands:")
```
melekes: let's move this section after `printResponse`

abci/cmd/abci-cli/abci-cli.go
```diff
@@ -477,11 +477,17 @@ func muxOnCommands(cmd *cobra.Command, pArgs []string) error {
...
+	fmt.Printf("%s: %s\n", echoCmd.Use, echoCmd.Short)
+	fmt.Printf("%s: %s\n", infoCmd.Use, infoCmd.Short)
+	fmt.Printf("%s: %s\n", checkTxCmd.Use, checkTxCmd.Short)
+	fmt.Printf("%s: %s\n", deliverTxCmd.Use, deliverTxCmd.Short)
+	fmt.Printf("%s: %s\n", queryCmd.Use, queryCmd.Short)
+	fmt.Printf("%s: %s\n", commitCmd.Use, commitCmd.Short)
+	fmt.Printf("%s: %s\n", setOptionCmd.Use, setOptionCmd.Short)
```
melekes: Also, would be good to print at the end

```
fmt.Println("Use \"[command] --help\" for more information about a command.")
```

bradyjoestar: Yes sir!

Issue #2207, opened 2018-08-11
libs/clist: Make remove automatically detachprev and detach next


ValarDragon:
Ref:
```golang
// CONTRACT: Caller must call e.DetachPrev() and/or e.DetachNext() to avoid memory leaks.
// NOTE: As per the contract of CList, removed elements cannot be added back.
func (l *CList) Remove(e *CElement) interface{} {
```
The caller should not have to call detach on the previous element, or the next element. This is confusing, and I can't see any reason why we wouldn't want these to detach. For reference, the golang standard list automatically detaches prev and next. AFAICT  _all_ current usages of our CList have memory leaks. They all only call `DetachPrev` and don't call `DetachNext`.

JayT106:
@alexanderbez @ValarDragon just looking at this issue, the mempool V0/V1 uses CList.
 Do you think it's still true for detaching the prev and the next element when removing an element?

alexanderbez:
Yes @JayT106. If you want to give a PR a shot, I'd be happy to review :)

JayT106:
> Yes @JayT106. If you want to give a PR a shot, I'd be happy to review :)

Sure!

PR #2235, opened 2018-08-15
mempool,pool: Fix potential memory leak

ValarDragon:
We weren't calling detachNext() on elements removed from the clist.
This means it is possible that a pointer to the next node would linger
along, which could eventually cause memory leaks.

ref #2207 

* [ ] Wrote tests - n/a?


ValarDragon:
Oh huh, maybe I missed something here. I'll close this, and we can relook at it post launch. (Should have ran tests locally before committing first)

PR #2236, opened 2018-08-16
ABCIAPPClient conn close

bradyjoestar:
Now we called `abci/cilent/grpc_client.go`  `OnStop`,we didn't close the conn.
```
// TODO: how to close conn? its not a net.Conn and grpc doesn't expose a Close()
	/*if cli.client.conn != nil {
		cli.client.conn.Close()
	}*/
```
I try to solve it by this issue:
[Close gprcClient defined by service in the .proto file.](https://github.com/grpc/grpc-go/issues/2264)


abci/client/grpc_client.go
```diff
@@ -80,10 +82,10 @@ func (cli *grpcClient) OnStop() {
 	cli.BaseService.OnStop()
 	cli.mtx.Lock()
```
melekes: I think we don't need a mutex here

bradyjoestar: sure deleted :-D

PR #2239, opened 2018-08-16
crypto/secp256k1: Fix signature malleability, adopt more efficient en…

### LINKED ISSUE #1958: crypto: Handling of Secp256k1 signature malleability

ValarDragon:
…coding

This removes signature malleability per ADR 14, and makes secp match
the encoding in ADR 15.

Closes #1958

* [ ] Wrote tests


Gopkg.toml
```diff
@@ -85,6 +85,10 @@
   name = "github.com/btcsuite/btcutil"
   revision = "d4cc87b860166d00d6b5b9e0d3b3d71d6088d4d4"
 
+[[constraint]]
+  name = "github.com/tendermint/btcd"
+  revision = "e5840949ff4fff0c56f9b6a541e22b63581ea9df"
```
Liamsi: I guess we won't do releases for forked libs, as this would be confusing?

melekes:
Looks like another non-deterministic failure.

```
--- FAIL: TestMempoolProgressAfterCreateEmptyBlocksInterval (0.08s)
panic: We should be stuck waiting, not moving to the next step [recovered]
	panic: We should be stuck waiting, not moving to the next step

goroutine 678 [running]:
testing.tRunner.func1(0xc420fc8000)
panic(0xf01b20, 0x1213000)
github.com/tendermint/tendermint/consensus.ensureNoNewStep(0xc420f29200)
	/go/src/github.com/tendermint/tendermint/consensus/common_test.go:314 +0x14e
github.com/tendermint/tendermint/consensus.TestMempoolProgressAfterCreateEmptyBlocksInterval(0xc420fc8000)
	/go/src/github.com/tendermint/tendermint/consensus/mempool_test.go:50 +0x3c5
testing.tRunner(0xc420fc8000, 0x118afd8)
created by testing.(*T).Run
FAIL	github.com/tendermint/tendermint/consensus	3.982s
```

PR #2243, opened 2018-08-18
Auto group process ticks

bradyjoestar:
In the `autofile.go`&`group.go` of `libs/autofile` ,

`processTicks` only run once .

So I add a loop to continue to listen the time.Ticker:

```
for {
		select {
		case <-af.ticker.C:
			af.closeFile()
		case <-af.tickerStopped:
			return
		}
	}
```


bradyjoestar:
circlecI told me that
```
#!/bin/bash -eo pipefail
bash test/app/test.sh
bash: test/app/test.sh: No such file or directory
Exited with code 127
```

But what I commited didn't included those files.Be grateful for your help!

PR #2255, opened 2018-08-21
Cleanup up Multisig naming

ebuchman:
This is follow up from https://github.com/tendermint/tendermint/issues/2163.

This PR standardizes the naming used in the multisig package and adds the multisig pubkey to the crypto/encoding/amino codec.

It also moves the compact bit array into it's own package.

* [ ] Wrote tests


crypto/multisig/wire.go
```diff
@@ -10,15 +10,15 @@ import (
 // TODO: Figure out API for others to either add their own pubkey types, or
 // to make verify / marshal accept a cdc.
 const (
-	ThresholdPubkeyAminoRoute = "tendermint/PubkeyMultisigThreshold"
+	PubKeyMultisigThresholdAminoRoute = "tendermint/PubKeyMultisigThreshold"
```
ValarDragon: Should we omit the multisig from the varname since it's in the package? Also should we omit the pubkey from the var name since there's no privkey

crypto/multisig/multisignature.go
```diff
@@ -4,20 +4,21 @@ import (
 	"errors"
 
 	"github.com/tendermint/tendermint/crypto"
+	bA "github.com/tendermint/tendermint/crypto/multisig/bitarray"
```
xla: This alias does more harm than good, obscuring what is imported here.

crypto/multisig/wire.go
xla: :+1: to both.

ebuchman: Mmmm .... I was hoping to reach parity between the var name and the Amino route. We can't remove `Multisig` from the amino route because there's no pkg information there, it needs to be the full `tendermint/PubKeyMultisigThreshold`.

We could consider removing `Multisig` from the var name, but I think we should keep PubKey for consistency/clarity

crypto/multisig/multisignature.go
ebuchman: Ok!

Issue #2259, opened 2018-08-22
cs.wal growing without bound on gaia-8001


zmanian:
https://github.com/cosmos/cosmos-sdk/issues/2123

cs.wal is 15GB on my node

kevlubkcm:
collating from riot chat:
seems like seed nodes are affected more >50GB usage
some nodes are relatively unaffected ~2GB usage
but most nodes are sitting around 15-20GB at the moment

jackzampolin:
```
$ du -h ~/.gaiad/
138M	/home/johnzampolin/.gaiad/data/gaia.db
52K	/home/johnzampolin/.gaiad/data/tx_index.db
20K	/home/johnzampolin/.gaiad/data/mempool.wal
20K	/home/johnzampolin/.gaiad/data/evidence.db
20G	/home/johnzampolin/.gaiad/data/cs.wal
4.4M	/home/johnzampolin/.gaiad/data/state.db
147M	/home/johnzampolin/.gaiad/data/blockstore.db
20G	/home/johnzampolin/.gaiad/data
8.0K	/home/johnzampolin/.gaiad/config/gentx
704K	/home/johnzampolin/.gaiad/config
20G	/home/johnzampolin/.gaiad/
```

fkbenjamin:
From one sentry // used ncdu:
~/.gaiad/data ---------------------------------------------------
                         /..
   18.0 GiB [##########] /cs.wal
  146.5 MiB [          ] /blockstore.db
  138.5 MiB [          ] /gaia.db
    3.3 MiB [          ] /state.db
   52.0 KiB [          ] /tx_index.db
   20.0 KiB [          ] /mempool.wal
   20.0 KiB [          ] /evidence.db

nckrtl:
@jackzampolin 
https://gist.github.com/nckrtl/6c036452db93b4b81cf5a126fc1719cd

BaryonNetwork:
10 min size checked every min.  (MB)
20180822-17:24:40  20519 /home/gaia/.gaiad/data/cs.wal
20180822-17:25:40  20551 /home/gaia/.gaiad/data/cs.wal
20180822-17:26:40  20574 /home/gaia/.gaiad/data/cs.wal
20180822-17:27:40  20605 /home/gaia/.gaiad/data/cs.wal
20180822-17:28:40  20631 /home/gaia/.gaiad/data/cs.wal
20180822-17:29:40  20657 /home/gaia/.gaiad/data/cs.wal
[...]

PR #2244, opened 2018-08-18
autofile processTicks

bradyjoestar:
In the `autofile.go`&`group.go` of `libs/autofile` ,

`processTicks` only run once .

So I add a loop to continue to listen the time.Ticker:

```
for {
		select {
		case <-af.ticker.C:
			af.closeFile()
		case <-af.tickerStopped:
			return
		}
	}
```

PR #2261, opened 2018-08-22
libs/autofile: bring back loops

ebuchman:
Trying to understand why https://github.com/tendermint/tendermint/issues/2259 is happening.

Looks like could be due to the changes in https://github.com/tendermint/tendermint/pull/2135, where we may have accidentally removed these loops.

* [ ] Wrote tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 0.23.1
+
+*August 22nd, 2018*
+
+BUG FIXES:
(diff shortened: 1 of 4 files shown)


ebuchman:
Seems to solve the issue from local manual testing.

ValarDragon:
Should we have autofile rotation tests to cover this case?

melekes:
@bradyjoestar wrote two tests which check rotation here https://github.com/tendermint/tendermint/pull/2244

PR #2264, opened 2018-08-23
update secret connection to use a little endian encoded nonce

### LINKED ISSUE #2263: crypto: Endianess of nonce in secret connection
Currently, the nonces used in secret connection are 12 bytes / 96 bits and basically a big endian encoded counter:
https://github.com/tendermint/tendermint/blob/013b9cef642f875634c614019ab13b17570778ad/p2p/conn/secret_connection.go#L335-L344

For the sake of consistency (ChaCha20 is all little endian) it would be great to change the nonce encoding to little endian, too.

Additionally to that, a counter using a simple 64-bit (unsigned) integer should be sufficiently large. It will not overflow in practice and we should have rekeyed long before it happens.

We should then also encode the chunk size [here] using little endian, too.

Related discussion: https://github.com/tendermint/kms/pull/38#discussion_r212392950

[here]: https://github.com/tendermint/tendermint/blob/013b9cef642f875634c614019ab13b17570778ad/p2p/conn/secret_connection.go#L125-L127

cc @tarcieri @ValarDragon

Liamsi:
Minor change that implements #2263.

Looks like circelci fails because this comes from a fork?

### DIFF
--- a/p2p/conn/secret_connection.go
+++ b/p2p/conn/secret_connection.go
@@ -334,11 +334,7 @@ func shareAuthSignature(sc *SecretConnection, pubKey crypto.PubKey, signature []
 
 // increment nonce big-endian by 1 with wraparound.
 func incrNonce(nonce *[aeadNonceSize]byte) {
-	for i := aeadNonceSize - 1; 0 <= i; i-- {
-		nonce[i]++
-		// if this byte wrapped around to zero, we need to increment the next byte
-		if nonce[i] != 0 {
-			return
-		}
-	}
+	counter := binary.LittleEndian.Uint64(nonce[4:])
+	counter++
+	binary.LittleEndian.PutUint64(nonce[4:], counter)
 }
--- a/p2p/conn/secret_connection.go
+++ b/p2p/conn/secret_connection.go
@@ -332,7 +332,10 @@ func shareAuthSignature(sc *SecretConnection, pubKey crypto.PubKey, signature []
 
 //--------------------------------------------------------------------------------
 
-// increment nonce big-endian by 1 with wraparound.
+// Increment nonce little-endian by 1 with wraparound.
+// Due to chacha20poly1305 expecting a 12 byte nonce we do not use the first four
+// bytes and only increment a 64 bit unsigned int in the remaining 8 bytes
+// (little-endian in nonce[4:]).
 func incrNonce(nonce *[aeadNonceSize]byte) {
 	counter := binary.LittleEndian.Uint64(nonce[4:])
 	counter++
--- a/p2p/conn/secret_connection.go
+++ b/p2p/conn/secret_connection.go
@@ -123,7 +123,7 @@ func (sc *SecretConnection) Write(data []byte) (n int, err error) {
 			data = nil
 		}
 		chunkLength := len(chunk)
-		binary.BigEndian.PutUint32(frame, uint32(chunkLength))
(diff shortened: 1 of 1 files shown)


p2p/conn/secret_connection.go
```diff
@@ -334,11 +334,7 @@ func shareAuthSignature(sc *SecretConnection, pubKey crypto.PubKey, signature []
...
 // increment nonce big-endian by 1 with wraparound.
 func incrNonce(nonce *[aeadNonceSize]byte) {
-	for i := aeadNonceSize - 1; 0 <= i; i-- {
-		nonce[i]++
-		// if this byte wrapped around to zero, we need to increment the next byte
-		if nonce[i] != 0 {
-			return
-		}
-	}
+	counter := binary.LittleEndian.Uint64(nonce[4:])
```
melekes: what about the first 4 bytes?

p2p/conn/secret_connection.go
```diff
@@ -334,11 +334,7 @@ func shareAuthSignature(sc *SecretConnection, pubKey crypto.PubKey, signature []
 
 // increment nonce big-endian by 1 with wraparound.
```
melekes: needs to be updated

p2p/conn/secret_connection.go
Liamsi: We do not use them / they will be all zero. We use chacha20poly1305 which expects a [12 byte nonce] though.

[12 byte nonce]:  https://github.com/golang/crypto/blob/614d502a4dac94afa3a6ce146bd1736da82514c6/chacha20poly1305/chacha20poly1305.go#L24

p2p/conn/secret_connection.go
Liamsi: Updated. Thanks!

p2p/conn/secret_connection.go
ValarDragon: It's kind of weird, but only using the last 8 bytes has become the standard for cha cha nonces in stream mode. (This is what is done in noise)

This will be sufficient for our purposes tho, since 2^64 messages will be more than well have. (We can't use random nonces, since that only gave a 2^48 bound, which is in the reachable zone)

melekes:
@Liamsi  can you commit a dummy change (add whitespace or something) to restart CircleCI? It looks like they have fixed the above problem.

Liamsi:
Looks like this needs to land first? https://github.com/tendermint/tendermint/pull/2279
Usually, you can also trigger a rebuild by closing and reopening a PR. Let me try.

melekes:
@Liamsi you're right! I've just merged #2279. 

btw. closing/reopening does not force CircleCI to rebuild.

Liamsi:
Looks like there are still some issues with circelci? e.g:
```
#!/bin/bash -eo pipefail
bash test/app/test.sh
bash: test/app/test.sh: No such file or directory
Exited with code 127
```

melekes:
You need to rebase your branch against latest develop.

https://stackoverflow.com/questions/7244321/how-do-i-update-a-github-forked-repository#7244456

Liamsi:
Meh, should have done that using the command line :-D One second.

Liamsi:
Thanks @melekes! Should I squash the commit messages?

ebuchman:
This needs to be reflected in changelog (breaking)

melekes:
Added.

PR #2270, opened 2018-08-24
wire2amino wasn't actually loading the private/public key bytes

enlight:
The key bytes are stored in the `value` field when serialized to JSON, not the `data` field this tool was attempting to load them from. And the key bytes in the JSON files are base64 encoded, not hex-encoded.

### DIFF
--- a/scripts/wire2amino.go
+++ b/scripts/wire2amino.go
@@ -49,8 +49,8 @@ type PrivVal struct {
 }
 
 type Data struct {
-	Type string       `json:"type"`
-	Data cmn.HexBytes `json:"data"`
+	Type string `json:"type"`
+	Data []byte `json:"value"`
 }
 
 func convertNodeKey(cdc *amino.Codec, jsonBytes []byte) ([]byte, error) {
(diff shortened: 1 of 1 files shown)


melekes:
@enlight can you commit a dummy change (add a whitespace or something) in order to restart CircleCI? Looks like they have fixed the above problem.

ebuchman:
I believe this script is to facilitate converting from the old go-wire JSON format to the new go-amino JSON format. 

The old go-wire used a hex-based `data` field, eg: https://github.com/tendermint/tendermint/blob/v0.16.0/test/p2p/data/mach1/core/config/genesis.json#L27

Were you trying to use this script for something else? If anything, maybe it's time to just remove it. Would you like to do the honours?

enlight:
Interesting, I was trying to use `wire2amino` to upgrade a chain from Tendermint 0.20 to 0.23, but if it was only intended to help people upgrade to Tendermint 0.19 as you say, then that would explain why I had to hack it to work. I'm guessing the amino upgrade in Tendermint 0.22 is the most likely reason why TM 0.23 fails to load `chaindata/config/*.json` created by TM 0.20.

Thanks for clearing that up.

ebuchman:
Sorry about that we haven't released corresponding tools as Amino has made some changes since but it should be stable now.

PR #2271, opened 2018-08-24
close ws connection on stop

### LINKED ISSUE #2258: Close connection when stopping WsClient.
Currently we are using a WsClient in this manner, to monitor tendermint and collect metrics:
```
func(c *rpcClient.WSClient) error {
		if c.IsRunning() {
			c.Stop() // panics on close c.ResponsesCh when not started (=nil)
		}
```
But `c.Stop()` does not close the underlying connection, which results in tendermint logging these errors:
```
E[08-22|09:35:58.903] Error closing connection                     module=rpc-server protocol=websocket remote=127.0.0.1:52428 err="close tcp 127.0.0.1:46657->127.0.0.1:52428: use of closed network connection"
```

My proposal would be to add `c.conn.Close()` to `func (c *WSClient) OnStop() {}` or `Stop()` .
I can contribute this myself too.
Let me know what you think.

FYI @ethanfrey

ruseinov:
Closes #2258 

* [ ] Wrote tests

### DIFF
--- a/rpc/lib/client/ws_client.go
+++ b/rpc/lib/client/ws_client.go
@@ -175,7 +175,9 @@ func (c *WSClient) OnStart() error {
 }
 
 // OnStop implements cmn.Service.
-func (c *WSClient) OnStop() {}
+func (c *WSClient) OnStop() {
+	_ = c.conn.Close()
+}
 
 // Stop overrides cmn.Service#Stop. There is no other way to wait until Quit
 // channel is closed.


ruseinov:
@ValarDragon There you go, not sure what I should write to changelog, as 0.22 is released, should I do 
`## Pending`?

ValarDragon:
You should aim this PR at develop, and update the `CHANGELOG_PENDING.md` file there

rpc/lib/client/ws_client.go
```diff
@@ -175,7 +175,9 @@ func (c *WSClient) OnStart() error {
 }
 
 // OnStop implements cmn.Service.
-func (c *WSClient) OnStop() {}
+func (c *WSClient) OnStop() {
+	_ = c.conn.Close()
```
melekes: We close connection inside `defer` in readRoutine/writeRoutine. If we don't execute that code, we might have a bigger problem at hand (read/write routines do not exit).

ruseinov: @melekes I will take a look

ruseinov: @melekes  the problem is likely here https://github.com/tendermint/tendermint/blob/14e7de15dbd01320c7ad146bd02b568e1794fc93/rpc/lib/client/ws_client.go#L477, where it should return, but it does not.

I'm going to close this PR, let me know if that makes sense and I will contribute another one.

ruseinov: It looks like it actually should be closed, not sure yet why this is not working.

ruseinov: I'm going to test this manually to try and reproduce the issue, as I can see tests there that check this and they seem to pass.

melekes: Please do. Thank you

Issue #2273, opened 2018-08-25
Reduce code in common/math.go


ValarDragon:
Currently the only functions in there that are used elsewhere are:

`MaxInt, MaxInt64, MinInt, MinInt64`

I think we should remove all other functions in that file, so its less code / API we have to support.

swbsin:
Maybe it's better to add a first good issue tag? :smile:

melekes:
Merged to develop.

PR #2274, opened 2018-08-25
Reduce code in common/math

Ahmah2009:
Remove all unused functions in common/math.go,
the only used functions are:` MaxInt`, `MaxInt64`, `MinInt`, `MinInt64`

requested in #2273.

### DIFF
--- a/libs/common/math.go
+++ b/libs/common/math.go
@@ -1,55 +1,6 @@
 package common
 
-func MaxInt8(a, b int8) int8 {
-	if a > b {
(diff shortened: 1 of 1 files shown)


melekes:
@Ahmah2009 can you commit a dummy change (add a whitespace or something) in order to restart CircleCI? Looks like they have fixed the above problem.

melekes:
Still broken... 😢  I think we'll need to turn off caching temporary then.

PR #2281, opened 2018-08-26
test make localnet in CI

zramsay:
starts #1532 and is useful as a stand alone additional test. Re-wiring the p2p test to use localnet will come as a seperate PR.

* [ ] Wrote tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -282,7 +282,7 @@ build-docker-localnode:
 # Run a 4-node testnet locally
 localnet-start: localnet-stop
 	@if ! [ -f build/node0/config/genesis.json ]; then docker run --rm -v $(CURDIR)/build:/tendermint:Z tendermint/localnode testnet --v 4 --o . --populate-persistent-peers --starting-ip-address 192.167.10.2 ; fi
-	docker-compose up
+	docker-compose up -d
 
 # Stop testnet
 localnet-stop:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -282,7 +282,7 @@ build-docker-localnode:
 # Run a 4-node testnet locally
 localnet-start: localnet-stop
 	@if ! [ -f build/node0/config/genesis.json ]; then docker run --rm -v $(CURDIR)/build:/tendermint:Z tendermint/localnode testnet --v 4 --o . --populate-persistent-peers --starting-ip-address 192.167.10.2 ; fi
-	docker-compose up
+	docker-compose up -d
```
melekes: I'd like to leave this attached to `STDOUT` by default. If I am developing something and want to test Tendermint, I'd prefer to see the output instantly and not search some place for logs.

scripts/localnet-blocks-test.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+CNT=0
```
melekes: let's add a newline here. otherwise it's easy to mistake `CTN` for yet another ENV variable (`0` vs `$0`)

melekes:
```
grouped write of manifest, lock and vendor: error while writing out vendor tree: failed to write dep tree: failed to export golang.org/x/sys: unable to deduce repository and source type for "golang.org/x/sys": unable to read metadata: go-import metadata not found
Makefile:84: recipe for target 'get_vendor_deps' failed
```

weird

PR #2286, opened 2018-08-27
Enforce file permissions

### LINKED ISSUE #2285: Handling changed file permissions
In autofile and in fsdb files are created with read-write permissions for the file owner (`0600`). If they are opened again, we do not check if the same restrictve file permissions still hold:

https://github.com/tendermint/tendermint/blob/013b9cef642f875634c614019ab13b17570778ad/libs/autofile/autofile.go#L116-L123

and:

https://github.com/tendermint/tendermint/blob/013b9cef642f875634c614019ab13b17570778ad/libs/db/fsdb.go#L202-L206

We should at least check that the permissions did not change (e.g. by accident) and enforce the intended perms (via chmod). Another approach would be to error in this case and tell the user to change the file perms (to make the user aware that sth. funky has happend here).

Will submit a small PR for both cases which silently fixes the permission in case they were changed.

(see https://github.com/tendermint/security/issues/5 and https://github.com/tendermint/security/issues/1)

Liamsi:
Resolves #2285

### DIFF
--- a/libs/autofile/autofile.go
+++ b/libs/autofile/autofile.go
@@ -1,12 +1,12 @@
 package autofile
 
 import (
-	"fmt"
 	"os"
 	"sync"
 	"time"
 
 	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/libs/errors"
 )
 
 /* AutoFile usage
@@ -35,21 +35,6 @@ const (
 	autoFileOpenDuration = 1000 * time.Millisecond
 	autoFilePerms        = os.FileMode(0600)
 )
-
-// ErrPermissionsChanged occurs if the file permission have changed since the file was created.
-type ErrPermissionsChanged struct {
-	name      string
-	got, want os.FileMode
-}
-
-func (e ErrPermissionsChanged) Error() string {
-	return fmt.Sprintf(
-		"file: [%v]\nexpected file permissions: %v, got: %v",
-		e.name,
-		e.want,
-		e.got,
-	)
-}
 
 // Automatically closes and re-opens file for writing.
 // This is useful for using a log file with the logrotate tool.
@@ -142,7 +127,7 @@ func (af *AutoFile) openFile() error {
 		return err
 	}
 	if fileInfo.Mode() != autoFilePerms {
-		return ErrPermissionsChanged{file.Name(), fileInfo.Mode(), autoFilePerms}
+		return errors.NewErrPermissionsChanged(file.Name(), fileInfo.Mode(), autoFilePerms)
 	}
 	af.file = file
 	return nil
--- a/libs/autofile/autofile.go
+++ b/libs/autofile/autofile.go
@@ -1,9 +1,12 @@
 package autofile
 
 import (
+	"fmt"
 	"os"
 	"sync"
 	"time"
+
+	"github.com/pkg/errors"
 
 	cmn "github.com/tendermint/tendermint/libs/common"
 )
@@ -34,6 +37,8 @@ const (
 	autoFileOpenDuration = 1000 * time.Millisecond
 	autoFilePerms        = os.FileMode(0600)
 )
+
+var PermissionsChangedErr = errors.New("file permissions changed")
 
 // Automatically closes and re-opens file for writing.
 // This is useful for using a log file with the logrotate tool.
@@ -126,10 +131,8 @@ func (af *AutoFile) openFile() error {
 		return err
 	}
 	if fileInfo.Mode() != autoFilePerms {
-		// reset file permissions:
-		if err := file.Chmod(autoFilePerms); err != nil {
-			return err
-		}
+		return errors.WithMessage(PermissionsChangedErr,
+			fmt.Sprintf("expected file permissions: %v, got: %v", autoFilePerms, fileInfo.Mode()))
 	}
 	af.file = file
 	return nil
--- a/libs/autofile/autofile.go
+++ b/libs/autofile/autofile.go
@@ -121,11 +121,11 @@ func (af *AutoFile) openFile() error {
 	if err != nil {
 		return err
 	}
-	fInf, err := file.Stat()
+	fileInfo, err := file.Stat()
 	if err != nil {
 		return err
 	}
-	if fInf.Mode() != autoFilePerms {
+	if fileInfo.Mode() != autoFilePerms {
 		// reset file permissions:
 		if err := file.Chmod(autoFilePerms); err != nil {
 			return err
--- a/libs/autofile/autofile.go
+++ b/libs/autofile/autofile.go
@@ -36,11 +36,19 @@ const (
 	autoFilePerms        = os.FileMode(0600)
 )
 
-// PermissionsChangedErr occurs if the file permission have changed since the file was created.
-type PermissionsChangedErr string
+// ErrPermissionsChanged occurs if the file permission have changed since the file was created.
+type ErrPermissionsChanged struct {
+	name      string
+	got, want os.FileMode
+}
 
-func (e PermissionsChangedErr) Error() string {
-	return string(e)
+func (e ErrPermissionsChanged) Error() string {
+	return fmt.Sprintf(
+		"file: [%v]\nexpected file permissions: %v, got: %v",
+		e.name,
+		e.want,
+		e.got,
+	)
 }
 
 // Automatically closes and re-opens file for writing.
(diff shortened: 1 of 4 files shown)


melekes:
Do you think it make sense to test new behaviour?

Liamsi:
> Do you think it make sense to test new behaviour?

I think it basically always makes sense to test any new behaviour. I'm not sure in this case though: The methods in question are private and it would be like testing the correctness of the standard go-library (and if we can properly write to the filesystem where the test is running etc). 

But let me add a few lines to the existing `autofile_test.go`. I would not introduce a new test-file that only has one test (for `write`) for fsdb.go.

melekes:
On the second thought, **I don't think we should try to fix errors one level above Tendermint**. If someone hacked the OS and changed file's permissions, then owners are screwed in any case. If it's due to some glitch, again it's not the Tendermint task to log that permissions got changed (although it may be nice).

Liamsi:
An alternative would be to error if the file permissions changed (as suggested in the related issue #2285). Could be a sentinel error so the caller can handle it and give the user a proper error message on the command-line / the UI. I do not think we should just ignore it if the permissions changed (either due some glitch or a hack).

Liamsi:
> This is kind of minor, so I'm fine with the overall change being merged. (Hence the approve)

@ValarDragon did you request changes by accident then?

melekes:
> if the permissions changed

keep in mind that if Tendermint loses permission to write to that file, it will panic. so we're only talking about group rw/other rw permissions.

melekes:
> keep in mind that if Tendermint loses permission to write to that file, it will panic. so we're only talking about group rw/other rw permissions.

that said, the reasonable thing to do is to log a warning (e.g. "%s permissions were modified (was: %v, become: %v)")

libs/autofile/autofile_test.go
```diff
@@ -78,19 +79,12 @@ func TestOpenAutoFilePerms(t *testing.T) {
...
 	require.NoError(t, err)
 
-	// reopen
+	// reopen and expect an PermissionsChangedErr as Cause
 	af, err = OpenAutoFile(name)
-	require.NoError(t, err)
-	err = af.Close()
-	require.NoError(t, err)
+	require.Error(t, err)
+	t.Log(err)
```
Liamsi: Will print:
`expected file permissions: -rw-------, got: -rwxr-xr-x: file permission changed`

libs/autofile/autofile.go
```diff
@@ -126,10 +131,8 @@ func (af *AutoFile) openFile() error {
 		return err
 	}
 	if fileInfo.Mode() != autoFilePerms {
-		// reset file permissions:
-		if err := file.Chmod(autoFilePerms); err != nil {
-			return err
-		}
+		return errors.WithMessage(PermissionsChangedErr,
+			fmt.Sprintf("expected file permissions: %v, got: %v", autoFilePerms, fileInfo.Mode()))
```
melekes: does it make sense to print the filename too?

libs/autofile/autofile.go
```diff
@@ -35,6 +38,8 @@ const (
 	autoFilePerms        = os.FileMode(0600)
 )
 
+var PermissionsChangedErr = errors.New("file permissions changed")
```
melekes: godoc is missing

libs/autofile/autofile.go
```diff
@@ -1,10 +1,13 @@
 package autofile
 
 import (
+	"fmt"
 	"os"
 	"sync"
 	"time"
 
+	"github.com/pkg/errors"
```
melekes: we should use https://github.com/tendermint/tendermint/blob/master/libs/common/errors.go

Liamsi: I started with that but it was annoyingly difficult to sth simple as just annotating an error with additional info (the formatted string). Maybe, I was just using it wrong.

Liamsi: Is there a simple example somewhere? 

I want a sentinel error as above `PermissionsChangedErr` (such that the caller can check against it) and simply annotate it with more info. I can use `ErrorWrap` but I do not see a way to get the annotated content without the whole stack trace a la:
```
=== RUN   TestOpenAutoFilePerms
--- PASS: TestOpenAutoFilePerms (0.00s)
	autofile_test.go:85: --= Error =--
		Data: "file permissions changed"
		Msg Traces:
		    0  /Users/ismail/go/src/github.com/tendermint/tendermint/libs/autofile/autofile.go:133 - [/var/folders/qf/7nq2274n50v5x9s2y4jcmfgm0000gn/T/permission_test853068086]: expected file permissions : -rw-------, got: -rwxr-xr-x
		Stack Trace:
		    0  /Users/ismail/go/src/github.com/tendermint/tendermint/libs/common/errors.go:14
		    1  /Users/ismail/go/src/github.com/tendermint/tendermint/libs/autofile/autofile.go:133
		    2  /Users/ismail/go/src/github.com/tendermint/tendermint/libs/autofile/autofile.go:60
		    3  /Users/ismail/go/src/github.com/tendermint/tendermint/libs/autofile/autofile_test.go:82
		    4  /usr/local/Cellar/go/1.10.3/libexec/src/testing/testing.go:777
		    5  /usr/local/Cellar/go/1.10.3/libexec/src/runtime/asm_amd64.s:2362
		--= /Error =--
PASS
```
That's way to verbose for this simple case. Basically, we just want the first msg trace (as possible with `github.com/pkg/errors` via `errors.WithMessage`). `cmn.NewErrorWithData` doesn't provide sth. similar either. `NewError` can not keep the orig. error/Error. 

Am I missing sth. obvious here?

libs/autofile/autofile.go
Liamsi: Yes, now it prints:
```
file: [/var/folders/qf/7nq2274n50v5x9s2y4jcmfgm0000gn/T/permission_test343097720]
		expected file permissions: -rw-------, got: -rwxr-xr-x
```

libs/autofile/autofile.go
Liamsi: added

libs/autofile/autofile.go
Liamsi: I just got rid of `github.com/pkg/errors` and type aliased a string. I think that is the simplest way to handle this.

libs/autofile/autofile.go
```diff
@@ -30,7 +31,17 @@ if err != nil {
...
 */
 
-const autoFileOpenDuration = 1000 * time.Millisecond
+const (
+	autoFileOpenDuration = 1000 * time.Millisecond
+	autoFilePerms        = os.FileMode(0600)
+)
+
+// PermissionsChangedErr occurs if the file permission have changed since the file was created.
+type PermissionsChangedErr string
```
xla: Prefixing of `Err` is the de facto standard, compared to suffxing: `PermissionsChangedErr` -> `ErrPermissionsChanged`.

libs/autofile/autofile.go
```diff
@@ -114,10 +125,18 @@ func (af *AutoFile) Sync() error {
...
+	file, err := os.OpenFile(af.Path, os.O_RDWR|os.O_CREATE|os.O_APPEND, autoFilePerms)
 	if err != nil {
 		return err
 	}
+	fileInfo, err := file.Stat()
+	if err != nil {
+		return err
+	}
+	if fileInfo.Mode() != autoFilePerms {
+		return PermissionsChangedErr(fmt.Sprintf("file: [%v]\nexpected file permissions: %v, got: %v",
```
xla: This error string construction might be better placed in the `Error` method of the error type and carry the relevant info in a struct:

``` go
type ErrPermissionsChanged struct {
	name          string
	expected, got os.FileMode
}

func (e ErrPermissionsChanged) Error() string {
	return fmt.Sprintf(
		"file: [%v]\nexpected file permissions: %v, got: %v",
		e.name,
		e.expected,
		e.got,
	)
}
```

xla: Additionally this way of formatting, I know brought it up before, is confusing. There is no clear rule to when to break a line and recognising where one construct starts and another requires a lot of effort. More consistently it should look like this:

``` go
	return PermissionsChangedErr(
		fmt.Sprintf(
			"file: [%v]\nexpected file permissions: %v, got: %v",
			file.Name(),
			autoFilePerms,
			fileInfo.Mode(),
		),
	)
```

This is allows for faster reading and faster editing of the code.

libs/db/fsdb.go
```diff
@@ -18,6 +18,14 @@ const (
 	dirPerm = os.FileMode(0700)
 )
 
+// PermissionsChangedErr occurs if the file permission have changed since the file was created.
+type PermissionsChangedErr string
```
xla: Prefix over suffix as above.

Is there a reason that both packages need to export their own version of the same error, can those be consolidated?

xla: `libs/common` does have errors, but comes with a lot of other things. I'd say a central `errors` package would be good, maybe for now under libs.

Liamsi: good point. done.

Issue #2232, opened 2018-08-15
introduce ValidateBasic methods to config


ebuchman:
Config data should be validated as soon as possible for internal inconsistencies like negative numbers and badly formatted addresses.

Eg. see https://github.com/tendermint/tendermint/pull/2169#discussion_r209539331

melekes:
FIXME:
- [ ] If `priv_validator_file` does not exist, the node does not produce any blocks

melekes:
We now check negative numbers - https://github.com/tendermint/tendermint/pull/2485

As for badly formatted addresses, if we add checks to ValidateBasic methods, we introduce duplication because right now addresses are checked right before they are used.

Btw, I went through all the config options and tried to mess with them. I haven't been able to crash or deviate Tendermint's behaviour from a normal one.

ebuchman:
Shall we close this then?

melekes:
Yes. Let's close this for now.

PR #2289, opened 2018-08-27
Check for int overflow in clist

### LINKED ISSUE #2288: libs: clist should explictly fail if max list len is reached
In the unlikely case that someone plans to use the linked-list implementation in libs/clist in a way which would lead to overflow max `int` it will behave funky as the underlying `len` field (an `int`) would wrap around. It is used in the mempool and the mempool can be configured to have a certain max. TX len. The input isn't validated yet and the config could theoretically impose a maximum which is larger than a clist can hold: https://github.com/tendermint/tendermint/issues/2232 

IMHO, even if the config is validated, it is  reasonable to be explicit about how many elements a clist can hold (and fail if someone tries to exceed that). This can be achieved by checking if `len` < `MaxInt ` before pushing new elements. (Even though we might run out of memory before seeing this ever happen).

Also, it could be reasonable to change the `len`type from `int` to `uint`.

https://github.com/tendermint/tendermint/blob/013b9cef642f875634c614019ab13b17570778ad/libs/clist/clist.go#L298-L319

See also: https://github.com/tendermint/security/issues/2

Liamsi:
resolves #2288 

@ValarDragon suggested to benchmark insertions first, which makes sense. If we can see a difference it might be more reasonable to let the caller (the mempool) enforce / handle an overflow logic. The mempool currently has this logic anyways. The inputs from the config are not validated yet, though: 
https://github.com/tendermint/tendermint/issues/2232


ValarDragon:
You can use the existing `BenchmarkPushBack` to achieve that benchmark

Liamsi:
No noticeable performance difference between with and without the check. With and without the check `BenchmarkPushBack` fluctuates between 440 and 510 ns/op on my machine.

libs/clist/clist.go
```diff
@@ -315,6 +322,9 @@ func (l *CList) PushBack(v interface{}) *CElement {
 		l.wg.Done()
 		close(l.waitCh)
 	}
+	if l.len == MaxLength {
```
melekes: Don't you think it's better to use `>=`?

libs/clist/clist.go
```diff
@@ -315,6 +322,9 @@ func (l *CList) PushBack(v interface{}) *CElement {
...
+		panic("clist: maximum length list reached")
```
melekes: let's print max here too "clist: reached maximum possible length %d"

libs/clist/clist.go
ValarDragon: Does it make a difference? Maxint is the largest possible value for an int to take on

ebuchman:
Addressed my comments - will merge when passes

Issue #2292, opened 2018-08-28
Support read-only operations on GoLevelDB


alessio:
Not all CLI commands in `gaiacli` require to acquire locks on the storage backends. Though presently the only way to create a `github.com/cosmos/cosmos-sdk/vendor/github.com/tendermint/tendermint/libs/db.GoLevelDB` instance is calling `NewGoLevelDB()`, which in turn calls `github.com/syndtr/goleveldb/leveldb.OpenFile()` with `nil` flags.

I'll propose a patch shortly.

jackzampolin:
This is needed to address key database locking issues in the SDK. I think this approach is a good one.

alessio:
Sure, e.g.:
* https://github.com/cosmos/cosmos-sdk/blob/develop/client/keys/list.go#L39

It should be required to acquire a lock to just listing keys.

melekes:
Merged to develop.

PR #2293, opened 2018-08-28
New NewGoLevelDBWithOpts() to pass opts down to goleveldb

alessio:
This is meant to allow Cosmos SDK to create and handle read-only leveldb instances.

Closes: #2292

* [x] Wrote tests

### DIFF
--- a/libs/db/go_level_db_test.go
+++ b/libs/db/go_level_db_test.go
@@ -8,7 +8,7 @@ import (
 
 	"github.com/syndtr/goleveldb/leveldb/opt"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	cmn "github.com/tendermint/tendermint/libs/common"
 )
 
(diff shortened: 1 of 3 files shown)


libs/db/go_level_db_test.go
```diff
@@ -6,9 +6,30 @@ import (
...
+	// Test write locks
+	db, err := NewGoLevelDB(name, "")
+	assert.Nil(t, err)
+	_, err = NewGoLevelDB(name, "")
+	assert.NotNil(t, err)
+	db.Close() // Close the db to release the lock
+
+	// Open the db twice in a row to test read-only locks
+	ro1, err := NewGoLevelDBWithOpts(name, "", &opt.Options{ReadOnly: true})
+	defer func() { ro1.Close() }()
```
melekes: `defer ro1.Close()`

libs/db/go_level_db_test.go
```diff
@@ -6,9 +6,30 @@ import (
...
+	assert.Nil(t, err)
+	ro2, err := NewGoLevelDBWithOpts(name, "", &opt.Options{ReadOnly: true})
+	defer func() { ro2.Close() }()
```
melekes: `defer ro2.Close()`

libs/db/go_level_db_test.go
```diff
@@ -6,9 +6,30 @@ import (
...
+
+	"github.com/stretchr/testify/assert"
 	cmn "github.com/tendermint/tendermint/libs/common"
 )
 
+func TestNewGoLevelDB(t *testing.T) {
+	name := fmt.Sprintf("test_%x", cmn.RandStr(12))
+	// Test write locks
+	db, err := NewGoLevelDB(name, "")
+	assert.Nil(t, err)
```
melekes: I'd prefer `require` here which terminates the test (i.e. no need to continue it if we failed to open db)

libs/db/go_level_db_test.go
```diff
@@ -6,9 +6,30 @@ import (
...
+	_, err = NewGoLevelDB(name, "")
+	assert.NotNil(t, err)
```
melekes: I'd prefer `require` here which terminates the test (i.e. no need to continue it if we failed to open db)

libs/db/go_level_db_test.go
alessio: ACKd and incorporated.

libs/db/go_level_db_test.go
alessio: ACKd and incorporated.

libs/db/go_level_db_test.go
alessio: ACKd and incorporated.

libs/db/go_level_db_test.go
alessio: ACKd and incorporated.

ValarDragon:
Just want to double check, does CLevelDB offer similar read only features? (I presume it does) I don't want us to be tied down to GoLevelDB features when we are discussing switching

alessio:
> Just want to double check, does CLevelDB offer similar read only features?

Allegedly, it does: https://github.com/amraboelela/CLevelDB/blob/62ac2e2c16b0355937de7e320f9e32d0802fa9ff/util/env_posix.cc

PR #2295, opened 2018-08-29
peng/fix docs base directory

nylira:

This is a mirror of this PR https://github.com/cosmos/cosmos-sdk/pull/2174, which fixes the root directory and rendering of the documentation website (tendermint.com/docs).


docs/config.js
```diff
@@ -2,23 +2,21 @@ module.exports = {
   title: "Tendermint Core",
   description: "Documentation for Tendermint Core",
   dest: "./dist/docs",
-  base: "/",
+  base: "/docs/",
```
melekes: Should we also remove this prefix from links in .md files?

CHANGELOG_PENDING.md
```diff
@@ -26,6 +26,7 @@ FEATURES:
 
 IMPROVEMENTS:
 - [docs] Lint documentation with `write-good` and `stop-words`.
+- [docs] Fix base directory of documentation website.
```
melekes: no need in changelog entry

docs/config.js
nylira: @faboweb I don't think that's necessary, right?

faboweb: can you point me to an example on the website?

nylira: @melekes this `base: "/docs"` refers to the documentation website being run under a subdirectory of the Tendermint website. I do believe there are links that need to be fixed in the Tendermint docs to ensure they work on the docs website. But that probably belongs in a different PR (@jb @zramsay are the ones who would know more). This PR is a high priority merge for develop and master. @greg-szabo

melekes: Then we should ask @ebuchman to include it into https://github.com/tendermint/tendermint/pull/2261

ebuchman: Seems theres a few docs related things on develop that aren't on master. Can we get a separate PR against master that just copies these over, or at least whatever parts we need?

ebuchman: Is https://github.com/tendermint/tendermint/pull/2336 enough?

PR #2300, opened 2018-08-30
config: reduce default mempool size

ValarDragon:
This reduces the mempool size from 100k to 5k. Note that each secp256k1 sig
takes .5ms to compute. Therefore an adversary could previously delay every
node on the network's computation time upon receiving a block by 50 seconds.

This now reduces that ability to being able to only delay each node by 2.5
seconds. This change should be reverted once ABCI recheck is implemented.

ref #2127 

I think this should be included next release, to improve network resiliency. 

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -421,8 +421,10 @@ func DefaultMempoolConfig() *MempoolConfig {
 		RecheckEmpty: true,
 		Broadcast:    true,
 		WalPath:      filepath.Join(defaultDataDir, "mempool.wal"),
-		Size:         100000,
-		CacheSize:    100000,
+		// Each signature verification takes .5ms, size reduced until we implement
+		// ABCI Recheck
+		Size:      5000,
+		CacheSize: 10000,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -421,8 +421,10 @@ func DefaultMempoolConfig() *MempoolConfig {
 		RecheckEmpty: true,
 		Broadcast:    true,
 		WalPath:      filepath.Join(defaultDataDir, "mempool.wal"),
-		Size:         100000,
-		CacheSize:    100000,
+		// Each signature verification takes .5ms, size reduced until we implement
+		// ABCI Recheck
+		Size:      5000,
+		CacheSize: 10000,
```
melekes: why reduce cache size?

ValarDragon: Seemed kinda weird to have a 20x ratio between Size and CacheSize, though I guess there isn't a need to reduce it

melekes: It's not about the ratio. It's about memory (total cache size) and being the first layer of protection from replay

ebuchman: I think more limited defaults is fine.

Issue #1859, opened 2018-07-02
ABCI:  Switch to list of lists of tags for transaction results, begin block, and end block


zramsay:
original issue: https://github.com/tendermint/abci/issues/273

cwgoes:
(thanks @ajcronk for the suggestion!)

ValarDragon:
Is this needed prelaunch? AFAICT this is to increase indexing speed. (Which I think can be punted to postlaunch, unless we don't want to break ABCI soon after launch. If this is true, I think we should make stubs for more ABCI changes). Per Anton's comment in the original issue, I agree that we shouldn't re-invent the wheel here, and should aim to be compatible with existing databases. e.g. https://github.com/tendermint/tendermint/issues/1161

cwgoes:
> Is this needed prelaunch? AFAICT this is to increase indexing speed. (Which I think can be punted to postlaunch, unless we don't want to break ABCI soon after launch. If this is true, I think we should make stubs for more ABCI changes). Per Anton's comment in the original issue, I agree that we shouldn't re-invent the wheel here, and should aim to be compatible with existing databases. e.g. #1161

Why would this increase indexing speed?

The reason on the SDK side is to allow prefixed error codes for particular modules and not need to invent a complex codespacing system to deal with potential module conflicts.

ValarDragon:
Oh woops your right, I completely misunderstood the comments on the prev issue.

If this is the goal though, why don't we make the results a list of structs consisting of a kv pair, k being byte slice and v being list of byte slices. Then we could also sort the results by key to get log n searching.

cwgoes:
> If this is the goal though, why don't we make the results a list of structs consisting of a kv pair, k being byte slice and v being list of byte slices. Then we could also sort the results by key to get log n searching.

How would that preserve the connection between tags associated with a particular event?

ebuchman:
@cwgoes @faboweb thoughts on prioritizing this pre/post launch?

cwgoes:
> @cwgoes @faboweb thoughts on prioritizing this pre/post launch?

Probably the most important use case for this at launch is slashing events, multiple of which can happen per `abci.RequestBeginBlock`. Without this, I don't see an easy way to report any number of slashing events in an easy-to-query manner.

melekes:
> Didn't we explicitly get rid of types here at one point?

yes, but I'd vouch to rethink if the current design does not satisfy our needs

[...]

Issue #1858, opened 2018-07-02
ABCI: add codespace string to error code


zramsay:
original issue: https://github.com/tendermint/abci/issues/272

ValarDragon:
Is this blocked on having the ADR written, per the original issue?

melekes:
What about the Log/Info fields? Aren't they enough?

cwgoes:
> What about the Log/Info fields? Aren't they enough?

They're non-deterministic, right? (at least, that's what's marked in the `.proto` files).

ebuchman:
Closing for #2557

Issue #1421, opened 2018-04-05
[pubsub/query] jsquery query language


melekes:
See https://github.com/tendermint/tendermint/pull/835#issuecomment-347034138

https://github.com/postgrespro/jsquery

jackzampolin:
I think with the postgres indexing issue and @alexanderbez 's work on the GraphQL server we can close this.

melekes:
This is independent from indexing.

tessr:
@melekes is this something you still think we should do?

melekes:
probably not

PR #2314, opened 2018-09-02
[ADR] ABCI errors and events

ebuchman:
Wrote up ADRs for the ABCI codespace (#1858) and events (#1859).

We should move forward with adding the codespace (very simple), but we may want to take our time with the events given the downstream impact on clients - not entirely sure it's launch critical, and the ADR has some TODOs around how indexing will work. Original PR for the events is https://github.com/tendermint/tendermint/pull/1803/ but it would need rebasing, review, testing, updates for clients, etc. @cwgoes do you think we can forgo this for launch?

NOTE https://github.com/tendermint/tendermint/issues/1859#issuecomment-418081011 (suggested priority for list of events is to track multiple slashing events in a single block ...)


docs/architecture/adr-021-abci-events.md
```diff
@@ -0,0 +1,47 @@
...
+
+ABCI tags were first described in [ADR 002](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-002-event-subscription.md).
+They are key-value pairs that can be used to index transactions.
+
+Currently, ABCI messages return a list of tags to describe an
+"event" that took place during the Check/DeliverTx/Begin/EndBlock,
+where each tag refers to a different property of the event, like the sending and receiving account addresses.
+
+Since there is only one list of tags, recording data for multiple such events in
+a single Check/DeliverTx/Begin/EndBlock must be done using prefixes in the key
```
ValarDragon: Maybe I'm missing something fundamentally, but why is there a need for prefixing. Shouldn't the order for trying to search for things by tag be: Merkle tree of tags, `key: tag bytes, value: list of events`, then you get a merkle proof along with the entire tag set for that block? 

i.e. suppose my key is "slashing", then the value for that tag could be "[`Validator 0 slashed for downtime, other relevant details`, `Validator 2 slashed for not voting on proposal X`]"

docs/architecture/adr-022-abci-errors.md
```diff
@@ -0,0 +1,60 @@
...
+single integer so that each module could have its own space to express its
+errors.
+
+## Decision
+
+Include a `string code_space` in all ABCI messages that have a `code`.
+This allows applications to namespace the codes so they can experiment with
+their own code schemes.
+
+It is the responsibility of applications to limit the size of the `code_space`
```
ValarDragon: Why not use amino's method for going from code_space -> byte prefix. (While maybe removing the null byte restriction)

docs/architecture/adr-021-abci-events.md
cwgoes: If you do it that way, you can't structure the value data. I might want to search for "all events where action = slashing and reason = downtime and validator = myValidator", for example, which isn't possible with this format (as I understand TM's current indexing model).

docs/architecture/adr-021-abci-events.md
```diff
@@ -0,0 +1,47 @@
...
+space.
+
+TODO: brief description of how the indexing works
```
cwgoes: It should work no differently than current indexing - just where multiple events can be indexed per transaction, so the transaction hash can be looked up from different events.

docs/architecture/adr-021-abci-events.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Decision
+
+Instead of returning a list of tags, return a list of events, where
+each event is a list of tags. This way we naturally capture the concept of
+multiple events happening during a single ABCI message.
```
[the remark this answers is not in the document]
cwgoes: I don't follow, can you further expand on your complexity calculation?

As I understand this, a multi-Msg transactions with three events would be indexed almost the same as three separate transactions with one event each (except for three separate transaction hashes in the latter case), and the search complexity would be equivalent.

docs/architecture/adr-021-abci-events.md
```diff
@@ -0,0 +1,47 @@
...
+## Consequences
+
+### Positive
+
+- Ability to track distinct events separate from ABCI calls (DeliverTx/BeginBlock/EndBlock)
+- More powerful query abilities
+
+### Negative
+
+- More complex query syntax
```
cwgoes: I don't think the query syntax needs to change. We don't need to allow searching for transactions by combinations of *multiple* events - "slashed AND jailed" - they can just be treated independently, like multiple separate transactions would be at the moment.

docs/architecture/adr-022-abci-errors.md
```diff
@@ -0,0 +1,60 @@
...
+In a [previous conversation](https://github.com/tendermint/abci/issues/165#issuecomment-353704015),
+it was suggested that not all non-zero codes need to be errors, hence why it's called `code` and not `error code`.
+It is unclear exactly how the semantics of the `code` field will evolve, though
+better lite-client proofs (like discussed for tags
+[here](https://github.com/tendermint/tendermint/issues/1007#issuecomment-413917763))
+may play a role.
+
+For now, note that having all type information in a single integer
+places an unfortunate requirement on all users of the `code` system to
+co-ordinate up-front on which `code` means what, hence restricting their ability experiment
```
cwgoes: *ability to experiment*

That's one problem, but the other problem is that it precludes an easy coordination method between "module implementers" and "client implementers". With an unbounded error domain (such as a string), module implementers can pick a globally unique prefix & error code set, so client implementers could easily implement support for "module A" regardless of which Hub/zone it was running in and which other modules were running with it. With only error codes, globally unique codes are difficult/impossible, as the space is finite and collisions are likely without an easy way to coordinate.

docs/architecture/adr-022-abci-errors.md
```diff
@@ -0,0 +1,60 @@
...
+with codes.
+
+For instance, while trying to build an ecosystem of modules that can be composed into a single
+ABCI application, the Cosmos-SDK had to hack a higher level "codespace" into the
+single integer so that each module could have its own space to express its
+errors.
+
+## Decision
+
+Include a `string code_space` in all ABCI messages that have a `code`.
```
cwgoes: Should it be a `codespace` string, or just an `error` string, and `code` changes to essentially a boolean (pass/fail) or a small set of error codes likely to be meaningful for many modules (a la HTTP XYZ)?

docs/architecture/adr-022-abci-errors.md
```diff
@@ -0,0 +1,60 @@
...
+## Positive
+
+- No need for complex codespacing on a single integer
+- More expressive type system for errors
+
+## Negative
+
+- Another field in the response needs to be accounted for
+- Some redundancy with `code` field
+- May encourage more error/code type info to move to the `codespace` string, which
```
cwgoes: Can light clients not verify the `codespace` (are we not putting it in state)?

docs/architecture/adr-021-abci-events.md
faboweb: Why can't you just query for all txs `validator = myValidator` and `slashing` filtering anything else client side?

cwgoes: Because multiple validators can be slashed per `abci.RequestBeginBlock` - so you need to put them under different keys (or put them as a list under one key and somehow associate other parameters such as slash amount and reason).

faboweb: Is is possible to define mutliple tags with different values: `['validator=myValidator','validator=yourValidator']`?

docs/architecture/adr-021-abci-events.md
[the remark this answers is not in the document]
ValarDragon: Actually just realized, having a list of events makes us lose proofs of exclusion as well. (As you can't prove that no other event has a given tag, efficiently)

docs/architecture/adr-021-abci-events.md
ValarDragon: > If you do it that way, you can't structure the value data. I might want to search for "all events where action = slashing and reason = downtime and validator = myValidator", for example, which isn't possible with this format (as I understand TM's current indexing model).

With my proposal here: https://github.com/tendermint/tendermint/pull/2314#discussion_r214748281
that seems do-able. You can only proofs of exclude on one sub-field of a tag (e.g. only validator address, or only reason), but thats still better than no proofs of exclusion on an entire tag.

ValarDragon: Wait also, the flow as I understood it to previously occur doesn't requiring prefixing as well. You would simply require all events that are tagged some way meet a provided structure, and you could then do the same compound searching you described.

docs/architecture/adr-021-abci-events.md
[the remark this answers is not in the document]
cwgoes: I wonder what the expected query workload is like here - maybe it is as you suggest, querying many events with the same tag, but it also could be more random-access, or searching by other tags like account address. I would also expect a lot of combined queries - "slashing events for myValidator".

Why does multiple-events-per-transaction change whether or not you get proofs of exclusion? Seems like that should be possible in either case, but only with one query key/value - prove no events under the key/value path queried, vs. no transaction hashes.

docs/architecture/adr-021-abci-events.md
cwgoes: See comment above. How do you suggest disambiguating which tags are associated with which others? Say two slashings happen in an `EndBlock` - if the resulting tags are `validator=validatorA, validator=validatorB, amount=2, amount=3`, we need a way to associate `validatorA` with `2` and `validatorB` with `3` - that's what I'm trying to get at.

docs/architecture/adr-022-abci-errors.md
ebuchman: Not following - can you expand ?

docs/architecture/adr-021-abci-events.md
```diff
@@ -0,0 +1,47 @@
...
+- *2018-09-02* Remove ABCI errors component. Update description for events
+- *2018-07-12* Initial version
+
+## Context
+
+ABCI tags were first described in [ADR 002](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-002-event-subscription.md).
+They are key-value pairs that can be used to index transactions.
+
+Currently, ABCI messages return a list of tags to describe an
+"event" that took place during the Check/DeliverTx/Begin/EndBlock,
```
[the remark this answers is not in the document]
ebuchman: do we currently support a sequence of events?

docs/architecture/adr-021-abci-events.md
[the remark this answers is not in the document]
ebuchman: requires extra convention in the keys. not necessarily a bad thing, but could be nicer to have native support for the concept

docs/architecture/adr-021-abci-events.md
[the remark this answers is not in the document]
ValarDragon: I don't think we need to optimize for the case where people just care about themselves. That has low bandwidth anyway (e.g. for `slashing events for myValidator` its a question of whether we want 1 or 2 paths. However I do think that there is a better solution where each tag also declares its relevant parties so we could get 1 path in my proposal as well) It feels more natural to me to optimize for the case where the optimization can cause a large performance difference. (e.g. `all events tagged governance`)

> Why does multiple-events-per-transaction change whether or not you get proofs of exclusion? 

I'm confused by this sentence. If this is in regards to `Actually just realized, having a list of events makes us lose proofs of exclusion as well. (As you can't prove that no other event has a given tag, efficiently)`, what I mean by this is if I search "give me everything with the tag slashing" I don't have evidence that you gave me everything, since there isn't a value that contains all slashing elements, or a sequence of values in the tree that are all slashing to do a proof of exclusion on.

docs/architecture/adr-021-abci-events.md
ValarDragon: My understanding is the goal is to find all events that meet some criterion. The way I proposed doing this is to create a merkle tree where all the leafs are basically `TAG_NAME, TAG_TREE_HASH`. The tag `TAG_TREE_HASH` is then the root of a second merkle tree for all events meeting that tag. This enables us to do a proof of exclusion for an entire tag, and we can even compound this by sorting the values in a second manner to get the proof of exclusion on a single other tag (e.g. validator address, or amount). 

Suppose the search query is `slashing`. You would have to return the entire value tree of the `slashing` tag. This only has 1 merkle proof required (as your given all leaves of the slashing tag's tree)

Suppose the search query is `all matches for slashing AND amount=2 AND validator=validatorA`, and you want a proof that you got everything. Then to prove inclusion and no exclusion, I would return the smallest of the 3 tag trees. (i.e. either return the the entire `slashing` tree, `amount` tree, or `validator` tree. Any one of these would convince you that all events under that tag were included). Since the tags are sorted by name, if one of tags doesn't exist, its a rather simple proof of exclusion. This again has only one merkle proof required. (Though requires an entire sub-tree, however this is not at all a problem given how specific the trees are)

Suppose the search query is `all matches for slashing AND validator=validatorA`, and `validator` is the special second field which we sort by, that I mentioned above. Then we would return `merkle path to slashing` + `all slashing tree's merkle paths to correct answers` + `boundary elements to prove no exclusion`. 

I guess after describing this, I do think this scheme I'm proposing is better, but is sufficeintly complicated that it should be punted to postlaunch. (I also think we can use some more advanced cryptography to get even better proofs here, there are authenticated dictionaries) 

I still don't understand how we can do proofs of exclusion in the currently proposed scheme within the ADR, if someone could walk me through how you do a proof that no event was excluded for the search "All events under slashing" I would appreciate that.

docs/architecture/adr-022-abci-errors.md
ValarDragon: amino takes the hash of the prefix you provide it, truncates it and removes null bytes, in order to get the prefix it uses for the registered concrete struct. I think we could do something similar for going from code_space string to prefix, just to remain more consistent. (Though this may just be even more confusing, I'm unsure)

ebuchman: not sure what prefix you're talking about here for the code space, maybe that's why I'm confused

ValarDragon: >  Include a `string code_space` in all ABCI messages that have a `code`. 

I'm talking about that `code_space`. My understanding is were deriving domain seperation from this code_space string.

docs/architecture/adr-021-abci-events.md
ebuchman: The current ADR doesnt say anything about how to do Merkle proofs - that's left for another issue: https://github.com/tendermint/tendermint/issues/1007#issuecomment-413917763. Would love it if you could help write up a proposal with the lite client requirements and some solutions.

My understanding is to some extent that the format in ABCI is independent of how we ultimately merklize everything.

ValarDragon: Oh woops, I didn't respond there mainly because I didn't understand the issue. This is starting to make sense to me now. I'd be glad to write up a proposal for that. 

I guess we could do a conversion layer between ABCI message type, and merkelization format, but that seems like complexity that might be annoying. Its also an O(total number of tags) conversion. (Not distinct tags), the conversion layer probably isn't a complexity concern given all of the other hashes that will be taking place.

docs/architecture/adr-022-abci-errors.md
ebuchman: right - what does it have to do with amino ?

ValarDragon: I was just suggesting to use the same algorithm to go from string -> prefix for domain separation.

ebuchman: Ok maybe I'm still missing something but seems confusing to make this field contain opaque bytes based on amino

ValarDragon: Your right, that probably would be even more confusing.

docs/architecture/adr-021-abci-events.md
ValarDragon: Maybe I'm misunderstanding, the `logical and` would say only return events that fit both slashing and the jailed tags. We still want that right? What we don't need is to support `logical or` imo. (E.g. all queries that are staking or are governance, as these can be independent) we want the former case to mitigate bandwidth for lite clients

docs/architecture/adr-021-abci-events.md
ebuchman: Note we already need to map from ABCI types to merkleable types because the ABCI responses may included non-deterministic data (ie. the log and info fields). Currently we're just merklizing the code+data, but the point of #1007 is to design a scheme for all the fields, ideally in a way that is more intelligent than just putting all the deterministic fields in a single struct (which wouldn't allow proof of exclusion, for instance). So it's worth having some doc to explain what's going on here and what the possible outcomes and tradeoffs are.

ebuchman: Slashing doesn't happen during txs but does happen during BeginBlock, before the txs. BeginBlock returns tags so that clients can be notified of what "events" happened during the BeginBlock. If multiple validators get slashed during a single BeginBlock, we need a way to disentangle those multiple events (ie. each one has a validator address, an amount slashed, perhaps a reason for slashing, etc.)

ebuchman: In theory that's possible, but would require significant change to the state machine, and is contrary to how we designed tendermint (ie. to auto detect and include evidence in blocks as a special type of data, distinct from txs).

docs/architecture/adr-021-abci-events.md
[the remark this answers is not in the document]
melekes: prefixing (`1_address`, `2_address`)

docs/architecture/adr-021-abci-events.md
faboweb: ok, then I retract my proposal

docs/architecture/adr-022-abci-errors.md
ebuchman: added this verbatim :)

ebuchman:
I think it's clear we want to move forward with codespace, but the events is still under consideration. Let's merge this and open a new PR for more discussion about the events

docs/architecture/adr-021-abci-events.md
melekes: refs https://github.com/tendermint/tendermint/issues/1421

docs/architecture/adr-021-abci-events.md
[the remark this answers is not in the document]
melekes: If we can't do with list of `key: value` strings, we should consider bringing in complex types (there was a discussion about this some time ago with Jae where we decided to stick with strings as raw bytes).

If we do so, we can add JSON type (e.g. `key: "Joth", value_type: JSON, value: {"address": "X", "denom": "B"}`). Then it would be super easy to support multiple events per DeliverTx.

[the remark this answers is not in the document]
ebuchman: > there was a discussion about this some time ago with Jae where we decided to stick with strings as raw bytes

Do you know if the reasoning for this was documented anywhere? Didn't we explicitly get rid of types here at one point?

> we can add JSON type

This is cool, but I think it means we'd need to unmarshal the JSON and index the keys in there too if we want to support eg. a query for "validator X slashed". More likely we need to really start thinking about the "indexing outside of tendermint core" story.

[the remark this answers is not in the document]
ebuchman: Let's move this to https://github.com/tendermint/tendermint/issues/1859

PR #2316, opened 2018-09-03
Skip timeout commit fix

bradyjoestar:
currently  we get the +2/3 precommit msg , if we use `SkipTimeoutCommit` in config, it will skip the `timeoutCommit` of next Height.
However , `enterCommit` will fail when we don't have the whole block partset.
```
case types.VoteTypePrecommit:
		precommits := cs.Votes.Precommits(vote.Round)
		cs.Logger.Info("Added to precommit", "vote", vote, "precommits", precommits.StringShort())
		if ok {
			if len(blockID.Hash) == 0 {
				cs.enterNewRound(height, vote.Round+1)
			} else {
				cs.enterCommit(height, vote.Round)

				if cs.config.SkipTimeoutCommit && precommits.HasAll()  {
					// if we have all the votes now,
					// go straight to new round (skip timeout commit)
					// cs.scheduleTimeout(time.Duration(0), cs.Height, 0, cstypes.RoundStepNewHeight)
					cs.enterNewRound(cs.Height, 0)
				}

			}
		} else if cs.Round <= vote.Round && precommits.HasTwoThirdsAny() {
			...
		}
```

I add the following judge case
```
&& cs.Height != height
```

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -1638,7 +1638,7 @@ func (cs *ConsensusState) addVote(vote *types.Vote, peerID p2p.ID) (added bool, 
 				cs.enterPrecommit(height, vote.Round)
 				cs.enterCommit(height, vote.Round)
 
-				if cs.config.SkipTimeoutCommit && precommits.HasAll() {
+				if cs.config.SkipTimeoutCommit && precommits.HasAll() && cs.Height != height {
 					// if we have all the votes now,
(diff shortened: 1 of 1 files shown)


ebuchman:
Not totally sure I follow here.  Can you please open an issue first with some more detail before creating the PR? 

Are you saying it's somehow possible to go to height H+1 before we have the whole block from height H?

PR #2318, opened 2018-09-03
max-bytes PR follow-up

melekes:


types/block_test.go
```diff
@@ -246,11 +247,11 @@ func TestMaxHeaderBytes(t *testing.T) {
...
 		ChainID:            maxChainID,
-		Height:             10,
+		Height:             math.MaxInt64,
 		Time:               time.Now().UTC(),
-		NumTxs:             100,
-		TotalTxs:           200,
-		LastBlockID:        makeBlockID(make([]byte, 20), 300, make([]byte, 20)),
+		NumTxs:             math.MaxInt64,
+		TotalTxs:           math.MaxInt64,
+		LastBlockID:        makeBlockID(make([]byte, 20), math.MaxInt64, make([]byte, 20)),
```
ValarDragon: Can the `20` be changed to tmhash.Size?

mempool/mempool.go
```diff
@@ -142,12 +142,10 @@ func (mem *Mempool) SetLogger(l log.Logger) {
...
 
-// SetFilter sets a filter for mempool to only accept txs for which f(tx)
-// returns true.
-func (mem *Mempool) SetFilter(f func(types.Tx) bool) {
-	mem.proxyMtx.Lock()
-	mem.filter = f
-	mem.proxyMtx.Unlock()
+// Filter sets a filter for mempool to only accept txs for which f(tx) returns
+// true.
+func Filter(f func(types.Tx) bool) MempoolOption {
```
ValarDragon: (Sorry for the bikeshed) Can this still be named `SetFilter`? I feel like that was clearer, since this isn't the actual filter.

ValarDragon: WithFilter sounds better to me

PR #2322, opened 2018-09-04
filter empty string for peers fix

### LINKED ISSUE #2320: Filter out empty addresses in persistent_peers list
```
persistent_peers = "<Member 1 node id>@<Member 1 hostname>:26656,<br />
<Member 2 node id>@<Member 2 hostname>:26656,<br />
<Member N node id>@<Member N hostname>:26656,"
```

gets us

```
Address () does not contain ID
```

we shouldn't even consider "" as an address! when validating persistent_peers and seeds, we should filter out empty strings.

james-ray:
* [ ] Wrote tests

fix #2320


p2p/switch.go
```diff
@@ -384,7 +384,13 @@ func (sw *Switch) IsDialing(id ID) bool {
 // Used to dial peers from config on startup or from unsafe-RPC (trusted sources).
 // TODO: remove addrBook arg since it's now set on the switch
 func (sw *Switch) DialPeersAsync(addrBook AddrBook, peers []string, persistent bool) error {
-	netAddrs, errs := NewNetAddressStrings(peers)
+	validPeers := make([]string, 0, cap(peers))
```
melekes: It may be better to move `cmn.SplitAndTrim` function https://github.com/tendermint/tendermint/blob/4da659d1fd6cec129924d820f6952f5362b095ab/libs/common/string.go#L41 to node/node.go and add the cycle there. That way we also filter private peers, index tags, etc.

Generally it's better to filter out bad input as earlier as you can.

ValarDragon:
SplitAndTrim is only used in node/node.go, so you could just delete `SplitAndTrim` from common.

PR #2323, opened 2018-09-04
filter out empty string

james-ray:
* [x] Wrote tests

fix #2322


node/strings.go
```diff
@@ -0,0 +1,20 @@
+package node
+
+import "strings"
+
+//move from string.go. filter out empty string too.
+func SplitAndTrim(s, sep, cutset string) []string {
```
melekes: let's make it private

node/strings.go
```diff
@@ -0,0 +1,20 @@
+package node
+
+import "strings"
+
+//move from string.go. filter out empty string too.
```
melekes: The proper godoc should be in form of `splitAndTrim ...`

node/strings.go
```diff
@@ -2,7 +2,8 @@ package node
 
 import "strings"
 
-//move from string.go. filter out empty string too.
+// SplitAndTrimEmpty is copied from common/string.go
```
ValarDragon: Whats the point of saying that it is copied from `common/string.go`, we can just delete it from there as its only used here.

ValarDragon:
Please don't keep amending the commits, we can squash it when merging. Constantly amending the commit makes it much harder to keep track of what the change between commits is :)

james-ray:
Sorry for the commits loss thing.

libs/common/string.go
```diff
@@ -58,6 +58,23 @@ func SplitAndTrim(s, sep, cutset string) []string {
 	return spl
 }
 
+// SplitAndTrimEmpty only returns non-empty strings
+func SplitAndTrimEmpty(s, sep, cutset string) []string {
```
melekes: Please move it to `node.go`

node/node.go
```diff
@@ -745,3 +746,21 @@ func saveGenesisDoc(db dbm.DB, genDoc *types.GenesisDoc) {
 	}
 	db.SetSync(genesisDocKey, bytes)
 }
+
+
+// SplitAndTrimEmpty returns only non-empty strings
+func SplitAndTrimEmpty(s, sep, cutset string) []string {
```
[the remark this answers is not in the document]
james-ray: oh,really,can I delete it? I'll check if anywhere used it. I will leave a changelog.

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ IMPROVEMENTS:
     - tweak params
     - only process one block at a time to avoid starving
 - [common] bit array functions which take in another parameter are now thread safe
+- [common] SplitAndTrim was deleted
```
ValarDragon: Its kind of confusing, but we put changelog updates for the next release in `CHANGELOG_PENDING.md`, could you move this there, and put it in the breaking changes section?

ValarDragon:
Looks good to me! Thank you for the PR! Just needs the changelog update moved to `CHANGELOG_PENDING.md`

melekes:
Thanks for your contribution 🖖

PR #2327, opened 2018-09-04
ignore existing peers in DialPeersAsync

### LINKED ISSUE #2253: /dial_peers should not drop existing connections
gaia v0.24.0

**What happened**:
Calling /dial_peers RPC with an existing peer leads to a disconnect and re-dial of said peer.

Since a connection is already established, I'd expect nothing to happen and the dial request to be silently ignored. For bonus points the RPC could return a status indicating it was ignored.

**Why is this useful?!**
It is well suited for [home-baked peer discovery](https://forum.cosmos.network/t/sentry-node-architecture-overview/454/44?u=mdyring), for instance in a sentry node setup.

melekes:
Fixes #2253

* [ ] Wrote tests

### DIFF
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -417,10 +417,10 @@ func (sw *Switch) DialPeersAsync(addrBook AddrBook, peers []string, persistent b
 	for i := 0; i < len(perm); i++ {
 		go func(i int) {
 			j := perm[i]
-
 			addr := netAddrs[j]
-			// do not dial ourselves
-			if addr.Same(ourAddr) {
+
+			if addr.Same(ourAddr) || sw.HasPeerWithAddress(addr) {
+				sw.Logger.Debug("Ignore attempt to connect to an existing peer", "addr", addr)
 				return
 			}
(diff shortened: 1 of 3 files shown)


p2p/switch.go
```diff
@@ -417,10 +417,10 @@ func (sw *Switch) DialPeersAsync(addrBook AddrBook, peers []string, persistent b
 	for i := 0; i < len(perm); i++ {
 		go func(i int) {
 			j := perm[i]
-
 			addr := netAddrs[j]
-			// do not dial ourselves
-			if addr.Same(ourAddr) {
+
+			if addr.Same(ourAddr) || sw.HasPeerWithAddress(addr) {
+				sw.Logger.Debug("Ignore attempt to connect to an existing peer", "addr", addr)
```
ValarDragon: Do we want a log message if you try to connect to yourself? If so, would that warrant its own message?

melekes: Not a biggie since it's `Debug`

PR #2337, opened 2018-09-05
BlockIntervalSeconds metric & DB tests cleanup & benchmark improv.

melekes:
**DO NOT SQUASH**

### DIFF
--- a/docs/introduction/install.md
+++ b/docs/introduction/install.md
@@ -48,6 +48,15 @@ to put the binary in `./build`.
 
 The latest `tendermint version` is now installed.
 
+## Run
+
+To start a one-node blockchain with a simple in-process application:
+
+```
+tendermint init
+tendermint node --proxy_app=kvstore
+```
+
 ## Reinstall
 
 If you already have Tendermint installed, and you make updates, simply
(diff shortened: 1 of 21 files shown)


melekes:
> Do we need to add libstdc++ as a dependency in the installation instructions?

We don't have installation instructions for TM with cleveldb, right? Should we write one?

ValarDragon:
I'm kind of confused why we have `syndtr/goleveldb` and `jmhodges/levigo`, can we remove the `syndtr/goleveldb` dependency as well?

melekes:
> can we remove the syndtr/goleveldb dependency as well?

I think we still want to have a key-value DB written in Go so we can cross-compile easy. But I am less convinced now that it should be default since a) https://github.com/syndtr/goleveldb/issues/226 b) they are magnitude slower than C _I will try to verify this claim_

ValarDragon:
I'm super in favor of just using CLevelDB. I don't think its reasonable to support cross-compilation long term, if we want faster crypto speeds / or any of the newer cryptography ideas. (e.g. BLS)

ebuchman:
Go-ethereum still uses `syndtr/goleveldb` exclusively so it could be something about the way we are configuring it? Also can you point to any popular/production products using this fork of leveldb? Is something of DataDog's using it ?

ValarDragon:
We currently compact every single block. We need to significantly reduce the frequency of compactions imo.

melekes:
> We currently compact every single block. We need to significantly reduce the frequency of compactions imo.

yes, but this is out of scope of this PR. Here I only update outdated cleveldb library.

> Also can you point to any popular/production products using this fork of leveldb? Is something of DataDog's using it ?

Datadog is using it for sure.

melekes:
- [ ] verify https://github.com/DataDog/leveldb/tree/master/deps/leveldb

melekes:
https://github.com/DataDog/leveldb/commit/12a0b6e10a5bf779330ed8b7c0f1f39f212d34e2

```
643805ecb6a4f9aaddc0652cb57ba0963162abe2c2a40b8c12aa31bbc80e645f  deps.leveldb.tar.gz
541e9d781f23b64933c8df7db7e206f047be7091c8c29e8a34a6e025a7ab960e  leveldb.v1.19.tar.gz

Files l2/include/leveldb/c.h and leveldb/deps/leveldb/include/leveldb/c.h differ
Files l2/include/leveldb/options.h and leveldb/deps/leveldb/include/leveldb/options.h differ
Files l2/port/atomic_pointer.h and leveldb/deps/leveldb/port/atomic_pointer.h differ
Files l2/table/format.cc and leveldb/deps/leveldb/table/format.cc differ
Files l2/table/table_builder.cc and leveldb/deps/leveldb/table/table_builder.cc differ
Files l2/util/options.cc and leveldb/deps/leveldb/util/options.cc differ
```

> ask DataDog guys what are they using leveldb for

just saw there is no "Issues" tag in https://github.com/DataDog/leveldb. maybe it is a bad idea after all

docs/introduction/install.md
```diff
@@ -66,11 +75,37 @@ make get_vendor_deps
...
+
+sudo apt-get install libsnappy-dev
+
+wget https://github.com/google/leveldb/archive/v1.20.tar.gz && \
+  tar -zxvf v1.20.tar.gz && \
+  cd leveldb-1.20/ && \
+  make && \
+  sudo scp -r out-static/lib* out-shared/lib* /usr/local/lib/ && \
+  cd include/ && \
+  sudo scp -r leveldb /usr/local/include/ && \
```
ebuchman: can we do this all without using `sudo`?

melekes: think so, yeah

Issue #2051, opened 2018-07-24
tx_search doesn't seem to respect "tx.height>21"


ethanfrey:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):

BUG_REPORT

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):

v0.21.0

**ABCI app** (name for built-in, URL for self-written if it's publicly available):

**Environment**:
- **OS** (e.g. from /etc/os-release):
- **Install tools**:
- **Others**:

**What happened**:

I had indexing on, and can search for matches on custom keys, but when I search for an obvious hit like "tx.height>21" or "tx.height<500000", it returns no matches. This also happens when I combine this with a tx tag search that returned a result without the tx.height range condition.

**What you expected to happen**:

I expected tx.height less than and greater than to return valid matches. This did work in an earlier version of tendermint. I am pretty sure it worked in 0.17. I don't know if I tested this since 0.19, so I am not sure where regression.

**How to reproduce it** (as minimally and precisely as possible):

Use a height range in tx_search.

**Logs (you can paste a small part showing an error or link a pastebin, gist, etc. containing more of the log file)**:

**Config (you can paste only the changes you've made)**:

**`/dump_consensus_state` output for consensus bugs**

**Anything else do we need to know**:

Looked at https://github.com/tendermint/tendermint/blob/dfa9a9a30a666132425b29454e90a472aa579a48/state/txindex/kv/kv.go#L156-L177 and a brief look at the code seemed to show it only handling tx.height= not < or > operators, which I am sure were intended design at one point.

ethanfrey:
Okay, here is a simple way to reproduce:


Server setup:
```
$ tendermint version
0.23.0-013b9cef
$ mkdir ~/tmsearch
$ tendermint --home ~/tmsearch/ init
$ TM_TX_INDEX_INDEX_ALL_TAGS=true tendermint node --home ~/tmsearch --proxy_app=kvstore --log_level=state:info,rpc:info,*:error
```

Client:

Query on tag returns. expected result 
Query on height return nothing (exact match or range).
Query with tag and height returns 400... ???
```
$ curl http://localhost:26657/broadcasttx_commit?tx=\"foo\"
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    "check_tx": {},
    "deliver_tx": {
      "tags": [
        {
          "key": "YXBwLmNyZWF0b3I=",
[...]

PR #2339, opened 2018-09-05
Bucky/changelog

ebuchman:
Updated the changelog to add PR refs and some things I found to be missing.

Need to add a couple more things but have to run for now.

Also changed the "BREAKING" section to have subsections so we can be more clear about what broke and who should care. Here's a direct link: https://github.com/tendermint/tendermint/blob/bucky/changelog/CHANGELOG_PENDING.md

Thoughts?

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -3,53 +3,59 @@
 BREAKING CHANGES:
 
 * CLI/RPC/Config
-
-    - [config] \#2169 Replace MaxNumPeers with MaxNumInboundPeers and MaxNumOutboundPeers
-    - [config] \#2300 Reduce default mempool size from 100k to 5k, until ABCI rechecking is implemented.
-    - [rpc] \#1815 `/commit` returns a `signed_header` field instead of everything being
-      top-level
+  - [config] \#2169 Replace MaxNumPeers with MaxNumInboundPeers and MaxNumOutboundPeers
+  - [config] \#2300 Reduce default mempool size from 100k to 5k, until ABCI rechecking is implemented.
+  - [rpc] \#1815 `/commit` returns a `signed_header` field instead of everything being top-level
 
 * Apps
-
-    - [abci] Added address of the original proposer of the block to Header
-    - [abci] Change ABCI Header to match Tendermint exactly
-    - [abci] \#2159 Update use of `Validator` (see ADR-018):
-        - Remove PubKey from `Validator` (so it's just Address and Power)
-        - Introduce `ValidatorUpdate` (with just PubKey and Power)
-        - InitChain and EndBlock use ValidatorUpdate
-        - Update field names and types in BeginBlock
-    - [state] \#1815 Validator set changes are now delayed by one block
-        - updates returned in ResponseEndBlock for block H will be included in
-        RequestBeginBlock for block H+2
+  - [abci] Added address of the original proposer of the block to Header
+  - [abci] Change ABCI Header to match Tendermint exactly
+  - [abci] \#2159 Update use of `Validator` (see
+    [ADR-018](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-018-ABCI-Validators.md)):
+    - Remove PubKey from `Validator` (so it's just Address and Power)
+    - Introduce `ValidatorUpdate` (with just PubKey and Power)
+    - InitChain and EndBlock use ValidatorUpdate
+    - Update field names and types in BeginBlock
+  - [state] \#1815 Validator set changes are now delayed by one block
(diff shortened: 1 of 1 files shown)


CHANGELOG_PENDING.md
```diff
@@ -1,45 +1,67 @@
...
+- [mempool] \#2234 Now stores txs by hash inside of the cache, to mitigate memory leakage
 
 BUG FIXES:
-- [mempool] No longer possible to fill up linked list without getting caching
-  benefits [#2180](https://github.com/tendermint/tendermint/issues/2180)
-- [state] kv store index tx.height to support search
-- [rpc] /dial_peers does not try to dial existing peers
+- [config] \#2284 Replace `db_path` with `db_dir` from automatically generated configuration files.
+- [mempool] \#2188 Fix OOM issue from cache map and list getting out of sync
+- [state] #2051 KV store index supports searching by `tx.height`
```
melekes: `\` is missing before `#2051`

CHANGELOG_PENDING.md
```diff
@@ -1,45 +1,67 @@
...
+    - [types] Updates to the Header:
+        - \#1815 NextValidatorsHash
+        - \#2106 ProposerAddress
+    - [blockchain] \#2213 Fix Amino routes for blockchain reactor messages
+    - [state] \#1815 Validator set changes are delayed by one block (!)
+        - Add NextValidatorSet to State, changes on-disk representation of state
+    - [consensus] \#2203 Implement BFT time
+        - Timestamp in block must equal median of timestamps in block's LastCommit
+    - [p2p] \#2263 Update secret connection to use a little endian encoded nonce
+    - [crypto] \#2205 Rename AminoRoute variables to no longer be prefixed by signature type.
```
ValarDragon: This is a Go API thing, the actual routes were unchanged

CHANGELOG_PENDING.md
```diff
@@ -1,45 +1,67 @@
...
+    - [crypto] \#2239 Secp256k1 signature changes (See ADR-014):
```
ValarDragon: Can we link to ADR-014 as well?

CHANGELOG_PENDING.md
```diff
@@ -1,45 +1,67 @@
...
-- [types] allow genesis file to have 0 validators ([#2015](https://github.com/tendermint/tendermint/issues/2015))
-- [libs] allow passing options through when creating instances of leveldb dbs ([#2292](https://github.com/tendermint/tendermint/issues/2292))
+- [types] Allow genesis file to have 0 validators ([#2015](https://github.com/tendermint/tendermint/issues/2015))
+    - Initial validator set can be determined by the app in ResponseInitChain
+- [libs/db] \#2293 Allow passing options through when creating instances of leveldb dbs
 
 IMPROVEMENTS:
 - [docs] Lint documentation with `write-good` and `stop-words`.
 - [scripts] Added json2wal tool, which is supposed to help our users restore
   corrupted WAL files and compose test WAL files (@bradyjoestar)
```
ValarDragon: Is suffixing all community contributions with who wrote something we're doing? I don't think I've seen this in changelogs before, but I'm also not opposed.

CHANGELOG_PENDING.md
```diff
@@ -1,45 +1,67 @@
...
+    - [crypto] \#2205 Rename AminoRoute variables to no longer be prefixed by signature type.
+    - [crypto] \#2239 Secp256k1 signature changes (See ADR-014):
+        - format changed from DER to `r || s`, both little endian encoded as 32 bytes.
+        - malleability removed by requiring `s` to be in canonical form.
+
 
 FEATURES:
-- [types] allow genesis file to have 0 validators ([#2015](https://github.com/tendermint/tendermint/issues/2015))
-- [libs] allow passing options through when creating instances of leveldb dbs ([#2292](https://github.com/tendermint/tendermint/issues/2292))
+- [types] Allow genesis file to have 0 validators ([#2015](https://github.com/tendermint/tendermint/issues/2015))
```
ValarDragon: Should we make #2015 follow the same format as everything else? (i.e. be at the front)

ValarDragon:
In the spirit of p2p having its own version (per your ADR), I think it should have its own changelog section instead of being a part of "Protocol"

CHANGELOG_PENDING.md
ebuchman: Not sure how best to do this but I really would love to give shout outs to open source contributors in the changelog

ebuchman:
Thanks for the review folks - I addressed everything and added a couple more points we forgot (including max bytes stuff!)

ValarDragon:
We probably should make a script that takes all regex matches for `\#<3 or 4 numbers>` and builds the correct PR / issue link

PR #2349, opened 2018-09-07
update UPGRADING.md and minor docs fixes

ebuchman:
Note commits 114c4051209f449dc365ac4e8123ce0a343d5904 and 5106af484f3bfe74eaec51aa38a85ea8cf6d8986 were pushed straight to develop.

This is just more follow up on the docs/changelog/upgrading.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -3,10 +3,11 @@
 This guide provides steps to be followed when you upgrade your applications to
 a newer version of Tendermint Core.
 
-## Upgrading from 0.23.0 to 0.24.0
+## v0.24.0
 
 New 0.24.0 release contains a lot of changes to the state and types. It's not
-compatible to the old versions.
+compatible to the old versions and there is no straight forward way to update
+old data to be compatible with the new version.
 
 To reset the state do:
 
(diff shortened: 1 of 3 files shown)


UPGRADING.md
```diff
@@ -28,5 +31,32 @@ max_num_outbound_peers = 10
...
+fields were changed, except their order.
+
+The `Validator` type was split into two, one containing an `Address` and one
+containing a `PubKey`. When processing `RequestBeginBlock`, use the `Validator`
+type, which contains just the `Address`. When returning `ResponseEndBlock`, use
+the `ValidatorUpdate` type, which contains just the `PubKey`.
+
+### Validator Set Updates
+
+Validator set updates returned in ResponseEndBlock for height `H` used to take
```
melekes: Don't think we should duplicate changelog entries here (except when they lead to clear actions like change this or do that)

ebuchman: Ok - wasn't sure how to think about this. The point is that for val sets they do have to change the way they track them in the app and I wanted to better indicate that. I'll add more text.

Issue #2357, opened 2018-09-07
some readme links are broken


ebuchman:
- `Quick Start -> Single Node` is broken link
- `Resources -> Tendermint Core` points to ReadTheDocs
- `Resources -> Applications -> Ethermint` points to old ethermint

Some other things point to ReadTheDocs, and others point to github pages. We should fix all this and probably point more to the tendermint.com/docs/ website :)

melekes:
Merged to develop.

PR #2359, opened 2018-09-07
update readme

zramsay:
#2357

* [x] Wrote tests


README.md
```diff
@@ -46,18 +46,18 @@ For examples of the kinds of bugs we're looking for, see [SECURITY.md](SECURITY.
 
 Requirement|Notes
 ---|---
-Go version | Go1.9 or higher
+Go version | Go1.10 or higher
```
ebuchman: So we actually require 1.10?

melekes: I thought so but looks like Tendermint works just fine with 1.9.3. 

tm-bench requires 1.10 since it uses new https://golang.org/doc/go1.10#math Round function.

melekes: Should we revert?

ValarDragon: I think we should stick to 1.10, or even consider upgrading to 1.11 around launch time. If so, we can use golang's official xchacha implementation, instead of our own.

zramsay: i bumped it b/c 1.11 was released & figured we should stay only one version behind.

PR #2360, opened 2018-09-08
Make mempool aware of MaxGas requirement

### LINKED ISSUE #2310: Implement MaxGas
As of #2184 we limit the size of the block in bytes.

Additionally, we'd like to limit the size of the block in terms of some application controlled quantifier of resources, called "gas". 

Each DeliverTx returns an amount GasUsed. The total GasUsed by the block must be less than the BlockSize.MaxGas consensus param.

Each CheckTx returns an amount GasWanted. The Application must enforce GasUsed < GasWanted. We must track the GasWanted in the mempool so we can reap txs within the gas limit.

Reap should be limited by Gas & Bytes - whichever limit is reached first.

ValarDragon:
Closes #2310 

~TODO: update spec~

* [x] Updated all code comments where relevant - no more instances of ReapMaxBytes AFAIK
* [ ] Wrote tests


mempool/mempool.go
```diff
@@ -380,10 +381,11 @@ func (mem *Mempool) notifyTxsAvailable() {
...
 }
 
-// ReapMaxBytes reaps transactions from the mempool up to n bytes total.
-// If max is negative, there is no cap on the size of all returned
+// ReapMaxBytesMaxGas reaps transactions from the mempool up to maxBytes bytes total
+// with the condition that the total gasWanted must be less than maxGas.
+// If both maxes are negative, there is no cap on the size of all returned
 // transactions (~ all available transactions).
-func (mem *Mempool) ReapMaxBytes(max int) types.Txs {
+func (mem *Mempool) ReapMaxBytesMaxGas(maxBytes int, maxGas int64) types.Txs {
```
ValarDragon: I don't think this is a good name for it, but the best alternate name I could think of was "ReapBlock" which didn't seem right. (Since the mempool isn't really aware of blocks at the moment)

melekes:
We should add this check:

```
Each DeliverTx returns an amount GasUsed. The total GasUsed by the block must be less than the BlockSize.MaxGas consensus param.
```

Also, would be great to see more tests.

ValarDragon:
Agreed for adding more tests. 

Per the issue, I thought we put it on the app to require that gas used is less than gas wanted. We don't run through the block until h+1 so if we decided that the gas used was actually too high, what would we do? Just drop the last txs?

ValarDragon:
Added all tests I could think of. I set `GasWanted=1` within kvstore's CheckTx to enable this. This seems reasonable to me, but if unreasonable, I can make a new proxy app which does something for GasWanted. This tests all cases which I could think of as valuable to test.

mempool/mempool_test.go
```diff
@@ -71,6 +71,54 @@ func checkTxs(t *testing.T, mempool *Mempool, count int) types.Txs {
 	return txs
 }
 
+func TestReapMaxGasMaxBytes(t *testing.T) {
```
melekes: `TestReapMaxBytesMaxGas`

ValarDragon: Good catch!

ValarDragon:
Hrmm, looks like switching the gas from kv store to persistent kv store caused things to break. Not really sure why, since persistent kv store just wraps kv store. I think I'll switch it back to kv store regardless, since it feels kinda strange only the variant that writes to disk does something for gasWanted.

ValarDragon:
Oh actually this doesn't update the filter, to ensure no tx is greater than max gas. I think that should be done in a second PR though.

Issue #2143, opened 2018-08-03
docs: suggestions/issues from review


ebuchman:
tendermint.com/docs/

- ~all links are broken~
- main page should have a small section for each seection in the sidebar: Getting Started, Core, Tools, Networks, Apps, Research.
- You should be able to link directly to one of these sections, rather than "the first subsection within them"

Getting Started
- first subsection is called Tendermint. Should be called Quick Start
    - should link to places to find more advanced starts !

Using Tendermint
- We should reconciule the structure of this section with the larger "Tendermint Core" section
    - protocol stuff can probably in a separate Specification section
- ~~should link to the slate docs more~~ #2361 
- ~~update Reset section to use unsafe_reset_all~~ https://github.com/tendermint/tendermint/pull/2361/commits/f735d5e16c6cd1520f3dc6b6daa53ef927f4a1ea
- ~link to config spec is broken~
- ~~No Empty Blocks~~
    - ~remove "This much requested feature was implemented in version 0.10.3"~
- Adding a Validator -> should become two sections called:
    - "Adding a Validator at Genesis"
    - "Adding a Validator at Runtime"
        - this needs to be written and explained how to update the validator set using EndBlock in the application
        - needs to link to the EndBlock section of the app developers guide
    - This also needs to be easier to find - currently its doesn't even make it into the sidebar its so burried under TendermintCore->UsingTendermint->TendermintNetworks->AddingAValidator

~~Running in Production~~
- ~~TODO~~ https://github.com/tendermint/tendermint/blob/master/docs/tendermint-core/running-in-production.md

~~Block Structure~~
- ~~deduplicate significantly with the spec~~ #2331

~~SecureP2P~~
- ~~link to spec~~ https://github.com/tendermint/tendermint/commit/13e0c98df0ca93c57a168456c9bc69902917e2ea
- ~~update note about using nacl - we don't anymore!~~

Validators
- document how the priv_validator.json works
- document how an out-of-process validator like the KMS is supposed to work
- include link to how to add validators section of Using Tendermint

Networks
- Instead of "Deploy a Testnet", just make "Local Docker Testnet" and "Manual Deployment" subsections of "Tendermint Networks"
- Fast Sync belongs under "Tendermint Core"

App Dev Guide
- Can be simplified and make better use of the abci README and specification.md file in the repo
- EndBlock
    - needs an update, no longer true about max +1/3
    - need to stress the change must have gone through consensus
[...]

PR #2361, opened 2018-09-09
docs: link to rpc

zramsay:
part of #2143 

Note: the updates will need to be manual (until we move to swagger) but I'll stay on top of it with automated Slack notification.

* [x] Wrote tests


docs/tendermint-core/using-tendermint.md
```diff
@@ -213,7 +217,7 @@ To reset a blockchain, stop the node, remove the `~/.tendermint/data`
 directory and run
 
 ```
-tendermint unsafe_reset_priv_validator
+tendermint unsafe_reset_all
```
melekes: `remove the `~/.tendermint/data`` unsafe_reset_all does that.

PR #2362, opened 2018-09-09
adr-021: note about tag spacers

ebuchman:
Follow up to https://github.com/tendermint/tendermint/pull/2314

Realized we could get "events" from a single list of tags by just including a "special" tag that separates groups of tags constituting an event.

@melekes @cwgoes @ValarDragon faboweb thoughts?


ebuchman:
Not sure we're going to use this approach but I'll merge this so it's at least noted

Issue #2347, opened 2018-09-06
RFC: TxSize.MaxBytes consensus parameter


melekes:
Right now, in consensus params we have TxSize.MaxBytes, which is supposed to limit the size of a single transaction. It's enforced in the mempool (>= 0.24.0), but not in the consensus when we validate a block ("A proposer running different mempool code could include txs larger than TxSize.MaxBytes".)

There are two ways we can go:

1) Enforce TxSize.MaxBytes in the consensus.
2) Get rid of TxSize.MaxBytes and use BlockSize.MaxBytes & TxSize.MaxGas to limit transactions.

### Enforce TxSize.MaxBytes in the consensus.

This gives you more control as you can have different values for tx / block maximums (e.g., TxSize.MaxBytes: 1KB and BlockSize.MaxBytes: 10MB). 

Questions:
- Is it worth it? What if we remove it? Will it make Tendermint-based networks more fragile to DDOS where an attacker floods the network with large transactions (tx size ~ maximum block size)?

### Get rid of TxSize.MaxBytes and use BlockSize.MaxBytes & TxSize.MaxGas to limit transactions.

This reduces the surface area (number of config parameters).

Original discussion: https://github.com/tendermint/tendermint/pull/2184#discussion_r214450932

ValarDragon:
I've thought about this a bit more since my original comment in the linked thread. Having TxSize.MaxBytes as a mempool config feels like a bad idea to me. This is basically a p2p config that everyone has set, so it is in effect blocking these large txs from ever getting through. (To communicate a way for you as a validator to accept these large txs would directly link who your sentries AFAICT)

Suppose now you have a valid reason for having a significant numbers of txs all greater than the current max tx size. (Multisigs + Multimsg txs) Your only recourse is to appeal to governance to please change the default. However its not a consensus parameter, so most full nodes / tx relayers are unlikely to adapt this change from a request to update the defaults. (Their only incentive to do so is if the gas is sufficiently large, which they can't determine since the tx is deleted before any "priority" function is called) However if this came from the application to change a consensus parameter, then our default / expected implementations would be to not reject txs based on size that are over the limit. Its reasonable to expect full nodes to change if it was a consensus parameter, since this is the definition of the tx being valid or not.

vrde:
TL;DR: get rid of `TxSize.MaxBytes`.

The maximum size of a transaction should be managed by the validation logic of the proxy application. 

@ValarDragon made a good point:
[...]

PR #2364, opened 2018-09-10
Remove ConsensusParams.TxSize and ConsensusParams.BlockGossip

melekes:
Refs #2347

* [ ] Wrote tests


abci/types/types.proto
```diff
@@ -200,27 +200,21 @@ message ResponseCommit {
...
 message ConsensusParams {
   BlockSize block_size = 1;
-  TxSize tx_size = 2;
-  BlockGossip block_gossip = 3;
+  EvidenceParams evidence_params = 2;
 }
 
 // BlockSize contains limits on the block size.
 message BlockSize {
+  // Note: must be greater than 0
```
ValarDragon: Couldn't it be -1?

blockchain/reactor_test.go
```diff
@@ -48,7 +48,7 @@ func newBlockchainReactor(logger log.Logger, maxBlockHeight int64) *BlockchainRe
 	for blockHeight := int64(1); blockHeight <= maxBlockHeight; blockHeight++ {
 		firstBlock := makeBlock(blockHeight, state)
 		secondBlock := makeBlock(blockHeight+1, state)
-		firstParts := firstBlock.MakePartSet(state.ConsensusParams.BlockGossip.BlockPartSizeBytes)
+		firstParts := firstBlock.MakePartSet(types.BlockPartSizeBytes)
```
ValarDragon: I still don't really understand why this isn't a consensus parameter. What optimizations are we getting by optimizing around 1MB blocks, it feels like we could still leave this as parameterized to me. (Or perhaps parameterize as "number of block parts")

melekes: NOTE: `BlockPartSizeBytes` really depends on the network bandwidth and network topology. It should be possible to make it adaptive to the current env by analysing the network, but it's #post-launch imho.

Plus we wasn't checking the size when accepting block part (no consensus per se 😄 ).

abci/types/types.proto
melekes: Unlimited? Do you think it's a good idea?

types/params.go
```diff
@@ -46,70 +36,56 @@ type EvidenceParams struct {
...
+			params.BlockSize.MaxBytes)
 	}
-
-	// ensure blocks aren't too big
 	if params.BlockSize.MaxBytes > MaxBlockSizeBytes {
 		return cmn.NewError("BlockSize.MaxBytes is too big. %d > %d",
 			params.BlockSize.MaxBytes, MaxBlockSizeBytes)
 	}
+
+	if params.BlockSize.MaxGas < -1 {
```
ebuchman: Maybe these should all be the same for consistency? ie. `< -1`

types/params.go
```diff
@@ -126,25 +102,11 @@ func (params ConsensusParams) Update(params2 *abci.ConsensusParams) ConsensusPar
...
-		}
-	}
-	if params2.TxSize != nil {
-		if params2.TxSize.MaxBytes > 0 {
-			res.TxSize.MaxBytes = int(params2.TxSize.MaxBytes)
-		}
-		if params2.TxSize.MaxGas > 0 {
-			res.TxSize.MaxGas = params2.TxSize.MaxGas
-		}
+		res.BlockSize.MaxBytes = int(params2.BlockSize.MaxBytes)
```
ebuchman: can we unify these and remove the cast?

types/params_test.go
```diff
@@ -9,10 +9,10 @@ import (
 	abci "github.com/tendermint/tendermint/abci/types"
 )
 
-func newConsensusParams(txsBytes, partSize int) ConsensusParams {
+func newConsensusParams(txsBytes, evidenceAge int) ConsensusParams {
```
ebuchman: - txBytes should be called maxBytes or blockBytes
- we should also pass in maxGas or blockGas

types/params_test.go
```diff
@@ -21,50 +21,45 @@ func TestConsensusParamsValidation(t *testing.T) {
...
+			assert.NoErrorf(t, tc.params.Validate(), "expected no error for valid params (#%d)", i)
 		} else {
-			assert.Error(t, tc.params.Validate(), "expected error for non valid params")
+			assert.Errorf(t, tc.params.Validate(), "expected error for non valid params (#%d)", i)
 		}
 	}
 }
 
-func makeParams(txsBytes, blockGas, txBytes, txGas, partSize int) ConsensusParams {
+func makeParams(txsBytes, blockGas, evidenceAge int) ConsensusParams {
```
ebuchman: deduplicate with newConsensusParams?

abci/types/types.proto
ebuchman: Probably best if we're consistent, so these ints all have the same domain

ebuchman:
Follow up in https://github.com/tendermint/tendermint/issues/2382

PR #2385, opened 2018-09-13
mempool: Filter new txs if they have insufficient gas

ValarDragon:
This also refactors the prior mempool to filter to be known as
"precheck filter" and this new filter is called "postcheck filter"

This PR also fixes a bug where the precheck filter previously didn't
account for the amino overhead, which could a maliciously sized tx to
halt blocks from getting any txs in them.

ref #2310 

* [X] Wrote tests
* [X] Updated CHANGELOG_PENDING.md - changelog covered by previous PR


node/node.go
```diff
@@ -250,7 +250,18 @@ func NewNode(config *cfg.Config,
...
 		state.LastBlockHeight,
 		mempl.WithMetrics(memplMetrics),
-		mempl.WithFilter(func(tx types.Tx) bool { return len(tx) <= maxDataBytes }),
+		mempl.WithPreCheckFilter(func(tx types.Tx) bool {
+			// We have to account for the amino overhead in the tx size as well
+			aminoOverhead := amino.UvarintSize(uint64(len(tx)))
+			return (len(tx) + aminoOverhead) <= maxDataBytes
+		}),
+		mempl.WithPostCheckFilter(func(tx types.Tx, res *abci.ResponseCheckTx) bool {
+			maxGas := state.ConsensusParams.BlockSize.MaxGas
```
melekes: Shouldn't this be outside? Otherwise there can be a DATA RACE. Note when it's outside, it's a copy. When inside, you're accessing state without mutex.

state/execution.go
```diff
@@ -149,8 +150,19 @@ func (blockExec *BlockExecutor) Commit(state State, block *types.Block) ([]byte,
...
 	)
-	filter := func(tx types.Tx) bool { return len(tx) <= maxDataBytes }
-	if err := blockExec.mempool.Update(block.Height, block.Txs, filter); err != nil {
+	preFilter := func(tx types.Tx) bool {
+		// We have to account for the amino overhead in the tx size as well
+		aminoOverhead := amino.UvarintSize(uint64(len(tx)))
+		return (len(tx) + aminoOverhead) <= maxDataBytes
+	}
+	postFilter := func(tx types.Tx, res *abci.ResponseCheckTx) bool {
+		maxGas := state.ConsensusParams.MaxGas
```
melekes: same

node/node.go
ValarDragon: You're right, thanks for catching this :)

mempool/mempool.go
```diff
@@ -81,8 +81,11 @@ type Mempool struct {
 	recheckEnd           *clist.CElement // re-checking stops here
 	notifiedTxsAvailable bool
 	txsAvailable         chan struct{} // fires once for each height, when the mempool is not empty
-	// Filter mempool to only accept txs for which filter(tx) returns true.
-	filter func(types.Tx) bool
+	// Filter mempool to only accept txs for which preCheckFilter(tx) returns true.
+	// This is called before each check tx
+	preCheckFilter func(types.Tx) bool
+	// Filter mempool to only accept txs for which postCheckFilter(tx) returns true.
+	postCheckFilter func(types.Tx, *abci.ResponseCheckTx) bool
```
xla: These should be typed a la: 

``` go
type preCheckFilterFunc func(types.Tx) bool

type postCheckFilterFunc func(types.Tx, *abci.ResponseCheckTx) bool
```

Likely also made public if the option functions are called from outside the mempool with it.

mempool/mempool.go
```diff
@@ -299,7 +308,8 @@ func (mem *Mempool) resCbNormal(req *abci.Request, res *abci.Response) {
 	switch r := res.Value.(type) {
 	case *abci.Response_CheckTx:
 		tx := req.GetCheckTx().Tx
-		if r.CheckTx.Code == abci.CodeTypeOK {
+		postCheckFilterPasses := (mem.postCheckFilter == nil) || mem.postCheckFilter(tx, r.CheckTx)
```
xla: It's repeated, might be worth to refactor into a function which can be called inside of the if statement itself.

node/node.go
```diff
@@ -245,12 +245,23 @@ func NewNode(config *cfg.Config,
...
 	mempool := mempl.NewMempool(
 		config.Mempool,
 		proxyApp.Mempool(),
 		state.LastBlockHeight,
 		mempl.WithMetrics(memplMetrics),
-		mempl.WithFilter(func(tx types.Tx) bool { return len(tx) <= maxDataBytes }),
+		mempl.WithPreCheckFilter(func(tx types.Tx) bool {
+			// We have to account for the amino overhead in the tx size as well
+			aminoOverhead := amino.UvarintSize(uint64(len(tx)))
+			return (len(tx) + aminoOverhead) <= maxDataBytes
```
xla: This seems to big of a core concern to have it linger in the setup code of the system. This function should be implemented in the package itself and harnessed with tests there and then be used in the setup of the mempool. It could look like this:

``` go
// AminoMaxBytes needs to be documented.
func AminoMaxBytes(maxDataBytes int) PreCheckFilterFunc {
	return func(tx types.Tx) bool {
		// We have to account for the amino overhead in the tx size as well
		aminoOverhead := amino.UvarintSize(uint64(len(tx)))
		return (len(tx) + aminoOverhead) <= maxDataBytes
	}
}

// ...
	mempl.WithPreCheckFilter(mempl.AminoMaxBytes(maxDataBytes))
// ...
```

mempool/mempool.go
```diff
@@ -448,7 +459,8 @@ func (mem *Mempool) ReapMaxTxs(max int) types.Txs {
 // Update informs the mempool that the given txs were committed and can be discarded.
 // NOTE: this should be called *after* block is committed by consensus.
 // NOTE: unsafe; Lock/Unlock must be managed by caller
-func (mem *Mempool) Update(height int64, txs types.Txs, filter func(types.Tx) bool) error {
+func (mem *Mempool) Update(height int64, txs types.Txs,
+	preCheckFilter func(types.Tx) bool, postCheckFilter func(types.Tx, *abci.ResponseCheckTx) bool) error {
```
xla: Why are those filters passed to the `Update` method directly? It feels odd to have two places to parametrise.

state/execution.go
```diff
@@ -149,8 +150,19 @@ func (blockExec *BlockExecutor) Commit(state State, block *types.Block) ([]byte,
...
+		aminoOverhead := amino.UvarintSize(uint64(len(tx)))
+		return (len(tx) + aminoOverhead) <= maxDataBytes
+	}
+	postFilter := func(tx types.Tx, res *abci.ResponseCheckTx) bool {
+		if maxGas == -1 {
+			return true
+		}
+		return res.GasWanted <= state.ConsensusParams.MaxGas
+	}
+	if err := blockExec.mempool.Update(block.Height, block.Txs, preFilter, postFilter); err != nil {
```
xla: See comment on the changes of the Update method signature.

mempool/mempool.go
ValarDragon: The idea behind passing these to the Update method directly was that we need a way to live-update the current filters used within the mempool. The reason being, if the consensus param changes, we want the mempool to update immediately. Are you suggesting having a separate `mempool.UpdateFilters(prefilter, postfilter)` method?

ebuchman: An alternative might be for the filters to have access to something that can load the latest ConsensusParams. Then we only need to set them once and they can read from the canonical shared state where the latest ConsensusParams are found

ValarDragon: I think that is a better alternative. I'd opt to keep the update signature as is within this PR, and plan to do what you're proposing soon. (It would likely entail either making consensus params a pointer, or adding getLatestConsensusParams method somewhere)

node/node.go
ValarDragon: Our mempool is an interface, with the interface defined in `state`. It seems a bit weird for `AminoMaxBytes` to be defined in the state package. Is it fine to introduce the actual mempool package as an import here for `AminoMaxBytes`?

xla: The `func AminoMaxBytes` was meant to go into the `mempool` package.

mempool/mempool.go
xla: In favour of @ebuchman approach.

node/node.go
ValarDragon: What I mean is that if `AminoMaxBytes` was in the mempool, `state` would have to import `mempool`. This is kind of weird, since `state` also defines the `mempool` interface.

xla: How does the implementation of `AminoMaxBytes` in the `mempool` package necessitate the import of `mempool` in `state`? It's not part of the mempool interface.

node/node.go
```diff
@@ -241,27 +241,22 @@ func NewNode(config *cfg.Config,
...
-		}),
-		mempl.WithPostCheckFilter(func(tx types.Tx, res *abci.ResponseCheckTx) bool {
-			if maxGas == -1 {
-				return true
-			}
-			return res.GasWanted <= maxGas
-		}),
+		mempl.WithPreCheckFilter(
+			mempool.PreCheckAminoMaxBytes(
+				types.MaxDataBytesUnknownEvidence(
```
ValarDragon: I think this change makes it harder to read than having maxDataBytes defined above

xla: Is this a comment regarding your personal taste or is there strong evidence how one is superior to the other?

ValarDragon: Personal preference, though I believe its founded on strong evidence. 3 function calls is less clear than two function calls on an explicitly defined variable name.

mempool/mempool.go
```diff
@@ -22,6 +22,14 @@ import (
 	"github.com/tendermint/tendermint/types"
 )
 
+// PreCheckFunc is an optional filter to determine if a transaction should be
+// rejected. Invoked before CheckTx.
+type PreCheckFunc func(types.Tx) bool
```
ValarDragon: Perhaps add that an example usage would be to ensure a tx isn't too large to ever fit into a block?

mempool/mempool.go
```diff
@@ -22,6 +22,14 @@ import (
...
+
+// PostCheckFunc is an optional filter executed after CheckTx and rejects
+// transaction if false is returned.
```
ValarDragon: Perhaps add that an example usage would be to ensure a tx doesn't have too much gas to ever fit into a block?

node/node.go
xla: The hierarchical nesting can be followed step-by-step to understand what is invoced, where with a variable one has to backtrack. Addtionally the scope of the NewNode function is already quite large, so avoiding adding more variables is desriable. Either case should be fine, if you have strong aversion towards the current formatting we can change it.

ValarDragon: nah no strong opinion :)

Issue #2294, opened 2018-08-28
The “genesis_time” field is ignored


KrzysiekJ:
When the `genesis_time` field in the `genesis.json` file contains a future date & time, Tendermint ignores that and starts producing blocks right after launch (if there is quorum). An expected behaviour would be to wait until that date.

Tested on 0.21.0.

ValarDragon:
Could you double check if this problem persists on v0.23.0? We've made a couple changes to time, so that may have fixed this problem.

KrzysiekJ:
Yes, it persists.

milosevic:
It should normally be fixed in https://github.com/tendermint/tendermint/pull/2203.

ebuchman:
Don't think this is actually fixed in #2203.

Also not 100% on how we should do this - can we somehow just not start the ConsensusState until genesis time ?

melekes:
> Also not 100% on how we should do this - can we somehow just not start the ConsensusState until genesis time ?

Can't we just `Sleep` in `node#OnStart` somewhere until genesis time is reached?

milosevic:
I was thinking the same.

ebuchman:
> Can't we just Sleep in node#OnStart somewhere until genesis time is reached?

Yes, though I was thinking we'd start up the other services (eg. pex and mempool) ahead of time.

Certainly a simple solution for now is to just sleep.

milosevic:
> Yes, though I was thinking we'd start up the other services (eg. pex and mempool) ahead of time.

This would actually be useful. Lets just check the complexity of doing it. If it's too complex then it might be done postlaunch and we go now with the simple sleep.

melekes:
Simple solution was implemented by @milosevic and merged to develop.

PR #2389, opened 2018-09-13
Delay starting node until Genesis time

milosevic:
Refs https://github.com/tendermint/tendermint/issues/2294

* [x] Wrote tests

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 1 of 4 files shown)


node/node.go
```diff
@@ -427,6 +429,12 @@ func NewNode(config *cfg.Config,
 
 // OnStart starts the Node. It implements cmn.Service.
 func (n *Node) OnStart() error {
+	now := tmtime.Now()
+	genTime := n.genesisDoc.GenesisTime
+	if genTime.After(now) {
+		time.Sleep(genTime.Sub(now))
```
melekes: We should print a message like "Genesis time is in the future. Sleeping until then..."

node/node_test.go
```diff
@@ -75,3 +77,17 @@ func TestSplitAndTrimEmpty(t *testing.T) {
...
 	}
 }
+
+func TestNodeDelayedStop(t *testing.T) {
+	config := cfg.ResetTestRoot("node_delayed_node_test")
+	now := tmtime.Now()
+
+	// create & start node
+	n, err := DefaultNewNode(config, log.TestingLogger())
+	n.GenesisDoc().GenesisTime = now.Add(5 * time.Second)
```
ValarDragon: Does this reliably pass for smaller values?

milosevic: It pass for 10 millisecond.

ebuchman:
With this sleep we can't kill the program with SIGINT - see https://github.com/tendermint/tendermint/issues/2434

PR #2398, opened 2018-09-14
update DOCS_README

zramsay:
TODO: add info about RPC


xla:
@zramsay Is this still wip?

zramsay:
@xla was gonna add info about the RPC but that'll change soon enough, see #2419

PR #2403, opened 2018-09-15
checkBlock when proposalBlock is complete

### LINKED ISSUE #2384: Run CheckTx before voting on blocks

james-ray:
* [ ] Wrote tests

fix #2384


james-ray:
Let me debug this abcd=dcba 0x616263643D64636261 in my local enviroment, please wait for a while

james-ray:
I will try write one adr markdown while debugging the code,

james-ray:
Sorry for the fail commit, I would use my AWS vm to run circleCI before commit the code.

PR #2415, opened 2018-09-17
Added /consensus_params endpoint 

### LINKED ISSUE #2311: expose consensus params over RPC
Like `/validators?height=X`, we should be able to query for `/consensus_params?height=X`

scriptonist:
* [x] Wrote tests

Fixes #2311

### DIFF
--- a/rpc/core/consensus.go
+++ b/rpc/core/consensus.go
@@ -305,5 +305,5 @@ func ConsensusParams(heightPtr *int64) (*ctypes.ResultConsensusParams, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &ctypes.ResultConsensusParams{BlockHeight: height, ConsensusParams: &consensusparams}, nil
+	return &ctypes.ResultConsensusParams{BlockHeight: height, ConsensusParams: consensusparams}, nil
 }
--- a/rpc/core/consensus.go
+++ b/rpc/core/consensus.go
@@ -261,3 +261,49 @@ func ConsensusState() (*ctypes.ResultConsensusState, error) {
 	bz, err := consensusState.GetRoundStateSimpleJSON()
 	return &ctypes.ResultConsensusState{bz}, err
 }
+
+// Get the consensus parameters  at the given block height.
+// If no height is provided, it will fetch the current consensus params.
+//
+// ```shell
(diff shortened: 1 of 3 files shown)


scriptonist:
I would really appreciate some early feedback

rpc/core/types/responses.go
```diff
@@ -125,6 +125,12 @@ type ResultValidators struct {
 	Validators  []*types.Validator `json:"validators"`
 }
 
+// ConsensusParams for given height
+type ResultConsensusParams struct {
+	BlockHeight     int64                  `json:"block_height"`
+	ConsensusParams *types.ConsensusParams `json:"consensus_params"`
```
xla: Minor nit, should not be a pointer as it is non-optional.

melekes:
This also needs a changelog entry (`CHANGELOG_PENDING.md`).

xla:
@zramsay Any docs that need to be updated with this new endpoint being introduced?

rpc/core/consensus.go
```diff
@@ -261,3 +261,49 @@ func ConsensusState() (*ctypes.ResultConsensusState, error) {
...
+// }
+// ```
+func ConsensusParams(heightPtr *int64) (*ctypes.ResultConsensusParams, error) {
+	height := consensusState.GetState().LastBlockHeight + 1
+	height, err := getHeight(height, heightPtr)
+	if err != nil {
+		return nil, err
+	}
+
+	consensusparams, err := sm.LoadConsensusParams(stateDB, height)
```
ebuchman: `consensusParams`

scriptonist: @ebuchman What should I fill in the changelog entry ? can you help me with that?

scriptonist: Should I open a new PR for incorporating these changes ?

ebuchman: no worries we'll cover it! thanks again for the contribution!

for reference, for the next PR, add a note about your change in the CHANGELOG_PENDING.md file: https://github.com/tendermint/tendermint/blob/develop/CHANGELOG_PENDING.md

Be sure to follow the formats used there :)

scriptonist: Sure thanks :smile:

Issue #2391, opened 2018-09-13
Refactor node 'other' info from array to object 


fedekunze:
If you want to get a node's `other` information (such as the [/status](https://tendermint.github.io/slate/?shell#status) endpoint) you have to parse the array items:

```
"other": [
"amino_version=0.12.0",
"p2p_version=0.5.0",
"consensus_version=v1/0.2.2",
"rpc_version=0.7.0/3",
"tx_index=on",
"rpc_addr=tcp://0.0.0.0:26657"
]
```

**Proposal**: Change the array to object and change `tx_index` from `on/off` to boolean value.

```
"other": {
"amino_version": "0.12.0",
"p2p_version": "0.5.0",
"consensus_version": "v1/0.2.2",
"rpc_version": "0.7.0/3",
"tx_index": true,
"rpc_addr": "tcp://0.0.0.0:26657"
]
```

melekes:
Merged to develop.

fedekunze:
thanks @melekes @xla !

PR #2417, opened 2018-09-17
[rpc] transform /status result.node_info.other into map

melekes:
Refs #2391

* [x] Wrote tests

### DIFF
--- a/p2p/node_info.go
+++ b/p2p/node_info.go
@@ -32,8 +32,8 @@ type NodeInfo struct {
 	Channels cmn.HexBytes `json:"channels"` // channels this node knows about
 
 	// ASCIIText fields
-	Moniker string   `json:"moniker"` // arbitrary moniker
-	Other   []string `json:"other"`   // other application specific data
+	Moniker string            `json:"moniker"` // arbitrary moniker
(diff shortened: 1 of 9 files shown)


p2p/node_info.go
```diff
@@ -56,13 +77,28 @@ func (info NodeInfo) Validate() error {
...
-			return fmt.Errorf("info.Other[%v] must be valid non-empty ASCII text without tabs, but got %v.", i, s)
+
+	// Sanitize versions
+	// XXX: Should we be more strict about version and address formats?
+	other := info.Other
+	versions := []string{other.AminoVersion,
+		other.AminoVersion,
+		other.P2PVersion,
+		other.ConsensusVersion,
+		other.RPCVersion}
```
xla: Inconsistent formatting, should be:

``` go
	versions := []string{
		other.AminoVersion,
		other.AminoVersion,
		other.P2PVersion,
		other.ConsensusVersion,
		other.RPCVersion,
	}
```

melekes:
It's pity that amino does not support maps.

> By the looks of it there is a lot of version info in there, so might be worth breaking those out into something like Versions

good idea 👍 but I think it should be easy to extract fields

PR #2424, opened 2018-09-17
blockchain: Register evidence amino message

### LINKED ISSUE #2423: Evidence Deserialization
It looks like we still have an unfixed deserialization bug.

https://github.com/cosmos/cosmos-sdk/issues/2031

xla:
In order for deserialisation of the blockchain the evidence amino
messages need to be registered otherwise we going to crash.

Fixes #2423

### DIFF
--- a/blockchain/wire.go
+++ b/blockchain/wire.go
@@ -2,6 +2,8 @@ package blockchain
 
 import (
 	"github.com/tendermint/go-amino"
+
+	"github.com/tendermint/tendermint/evidence"
 	"github.com/tendermint/tendermint/types"
 )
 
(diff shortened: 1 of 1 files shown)


blockchain/wire.go
```diff
@@ -10,4 +12,5 @@ var cdc = amino.NewCodec()
 func init() {
 	RegisterBlockchainMessages(cdc)
 	types.RegisterBlockAmino(cdc)
+	evidence.RegisterEvidenceMessages(cdc)
```
ebuchman: we actually want `types.RegisterEvidences`. This is registering the evidence reactor messages, which we don't care about here. We need the evidence data type itself, which is what is included in the blockchain

ebuchman: oh what. It looks like types.RegisterBlockAmino already does that. We may need to investigate further

xla:
@ebuchman Thanks for the follow-up and correciton.

PR #2425, opened 2018-09-18
metrics: Add additional metrics to p2p and consensus

mslipper:
Partially addresses https://github.com/cosmos/cosmos-sdk/issues/2169.

* [ ] Wrote tests


mslipper:
Two things I'd like direct feedback on:

1. It seems like the RPC API uses the value of `FastSync` to determine if the node is syncing or not. I've followed that model for the metrics. Let me know if this isn't correct.
2. I'm using a goroutine to poll the `Status()` method on `MConnection` to get the state of the send queue, since it seemed expensive to calculate all the time and I didn't want to inject metrics into the `MConnection` struct.

consensus/reactor.go
```diff
@@ -814,6 +819,16 @@ func (conR *ConsensusReactor) StringIndented(indent string) string {
 	return s
 }
 
+func (conR *ConsensusReactor) setCatchingUp() {
+	var catchingUp float64
+	if conR.fastSync {
```
melekes: "Whether or not a node is synced. 0 if syncing, 1 if synced." **1 and 0 mixed up**

if you're adding this condition, you need to modify metric name and description to be

`"Whether or not a node is fast synced. 1 if fast syncing, 0 if synced or syncing without fast sync."`

consensus/reactor.go
```diff
@@ -42,15 +42,19 @@ type ConsensusReactor struct {
...
 	fastSync bool
 	eventBus *types.EventBus
+
+	metrics *Metrics
 }
 
 // NewConsensusReactor returns a new ConsensusReactor with the given
 // consensusState.
-func NewConsensusReactor(consensusState *ConsensusState, fastSync bool) *ConsensusReactor {
+func NewConsensusReactor(consensusState *ConsensusState, fastSync bool, csMetrics *Metrics) *ConsensusReactor {
```
melekes: metrics should be an option. see consensus/state.go for example 

```

// WithMetrics sets the metrics.
func WithMetrics(metrics *Metrics) CSOption {
	return func(cs *ConsensusState) { cs.metrics = metrics }
}
```

mempool/metrics.go
```diff
@@ -4,8 +4,9 @@ import (
 	"github.com/go-kit/kit/metrics"
 	"github.com/go-kit/kit/metrics/discard"
 
-	prometheus "github.com/go-kit/kit/metrics/prometheus"
+	"github.com/go-kit/kit/metrics/prometheus"
 	stdprometheus "github.com/prometheus/client_golang/prometheus"
+	"github.com/tendermint/tendermint/config"
```
melekes: tmcfg to be consistent or better just `cfg`

p2p/peer.go
```diff
@@ -225,12 +236,36 @@ func (p *peer) OnStart() error {
...
+			select {
+			case <-p.metricsTicker.C:
+				status := p.mconn.Status()
+				var sendQueueSize float64
+				for _, chStatus := range status.Channels {
+					sendQueueSize += float64(chStatus.SendQueueSize)
+				}
+
+				p.metrics.PeerPendingSendBytes.With("peer-id", string(p.ID())).Set(sendQueueSize)
+			case <-p.quitChan:
```
melekes: peer already has a quit channel: `p.Quit()`

p2p/peer.go
```diff
@@ -225,12 +236,36 @@ func (p *peer) OnStart() error {
...
 		return err
 	}
+
 	err := p.mconn.Start()
-	return err
+	if err != nil {
+		return err
+	}
+
+	go func() {
```
melekes: please extract to a separate function

p2p/peer.go
```diff
@@ -109,12 +116,16 @@ func newPeer(
 	reactorsByCh map[byte]Reactor,
 	chDescs []*tmconn.ChannelDescriptor,
 	onPeerError func(Peer, interface{}),
+	metrics *Metrics,
```
melekes: again, should be an option, not a requirement

config/metrics.go
```diff
@@ -0,0 +1,3 @@
+package config
+
+const MetricsNamespace = "tendermint"
```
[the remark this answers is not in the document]
mslipper: Was thinking I'd make it configurable in a separate PR that would add support for passed-in namespaces.

consensus/reactor.go
mslipper: Ah, nice catch.

mslipper:
@xla @melekes updates in.

@xla - the data race is related to the `Status()` method. It already has a TODO indicating that `atomic` should be used: https://github.com/tendermint/tendermint/blob/master/p2p/conn/connection.go#L588. Would you like me to refactor that code to use atomic? 

Re: separating the metrics code into an `instrumenting` interface - I'm happy to do this as part of this PR, but I'm not sure that the overhead of adding metrics directly to the peer is very high. From the discussion that spawned this issue (https://github.com/cosmos/cosmos-sdk/issues/2169), my understanding is that Prometheus metrics are very lightweight and there isn't a common pattern in the codebase yet to expose metrics. The only 'heavy' piece of this implementation is the goroutine that polls a peer's status, but I don't think running a fairly tight loop once every 10 seconds will be a huge performance hit.

config/config.go
```diff
@@ -634,6 +634,9 @@ type InstrumentationConfig struct {
 	// you increase your OS limits.
 	// 0 - unlimited.
 	MaxOpenConnections int `mapstructure:"max_open_connections"`
+
+	// Tendermint instrumentation namespace.
+	Namespace string `mapstructure:"namespace"`
```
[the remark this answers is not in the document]
mslipper: Done.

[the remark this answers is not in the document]
melekes: Hmmm.. I do not see any changes to those files

```
# Instrumentation namespace
namespace = {{ .Instrumentation.Namespace }}
```

mslipper:
Cool. I have another PR with a bunch more metrics once this guy is merged :).

config/config.go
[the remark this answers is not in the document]
mslipper: Weird. Let me redo.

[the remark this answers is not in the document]
mslipper: I accidentally reset those files. Just pushed.

consensus/metrics.go
```diff
@@ -3,11 +3,12 @@ package consensus
 import (
 	"github.com/go-kit/kit/metrics"
 	"github.com/go-kit/kit/metrics/discard"
-
```
xla: This newline is meaningful, we group by: `stdlib`, `external packages`, `internal packages`.

config/toml.go
```diff
@@ -284,6 +284,9 @@ prometheus_listen_addr = "{{ .Instrumentation.PrometheusListenAddr }}"
 # you increase your OS limits.
 # 0 - unlimited.
 max_open_connections = {{ .Instrumentation.MaxOpenConnections }}
+
+# Instrumentation namespace
+namespace = {{ .Instrumentation.Namespace }}
```
xla: Seems to break the integration tests, check circle.

p2p/peer.go
```diff
@@ -143,12 +157,19 @@ func (p *peer) OnStart() error {
 	if err := p.BaseService.OnStart(); err != nil {
 		return err
 	}
+
 	err := p.mconn.Start()
-	return err
+	if err != nil {
```
xla: Can be inlined: `if err := p.mconn.Start(); err != nil {`

p2p/peer.go
```diff
@@ -258,8 +287,8 @@ func (pc *peerConn) CloseConn() {
 // the peer.
 // NOTE: blocking
 func (pc *peerConn) HandshakeTimeout(
-	ourNodeInfo NodeInfo,
-	timeout time.Duration,
+		ourNodeInfo NodeInfo,
+		timeout time.Duration,
```
xla: Non meaningful whitespace.

p2p/peer.go
```diff
@@ -314,16 +343,39 @@ func (p *peer) String() string {
...
-	reactorsByCh map[byte]Reactor,
-	chDescs []*tmconn.ChannelDescriptor,
-	onPeerError func(Peer, interface{}),
-	config tmconn.MConnConfig,
+		conn net.Conn,
+		p *peer,
+		reactorsByCh map[byte]Reactor,
+		chDescs []*tmconn.ChannelDescriptor,
+		onPeerError func(Peer, interface{}),
+		config tmconn.MConnConfig,
```
xla: Non meaningful whitespace.

p2p/peer.go
```diff
@@ -99,21 +101,30 @@ type peer struct {
...
-	reactorsByCh map[byte]Reactor,
-	chDescs []*tmconn.ChannelDescriptor,
-	onPeerError func(Peer, interface{}),
+		pc peerConn,
+		mConfig tmconn.MConnConfig,
+		nodeInfo NodeInfo,
+		reactorsByCh map[byte]Reactor,
+		chDescs []*tmconn.ChannelDescriptor,
+		onPeerError func(Peer, interface{}),
+		options ...PeerOption,
```
xla: Non meaningful whitespace.

p2p/peer.go
```diff
@@ -99,21 +101,30 @@ type peer struct {
 
 	// User data
 	Data *cmn.CMap
+
+	metrics *Metrics
+
```
xla: Unnecessary newline.

xla:
@mslipper Thanks for all the ammendments. Could you resolve the CHANGELOG conflict? After that the PR shouldbe good to go.

Issue #2382, opened 2018-09-12
Follow up to removing some consensus params


ebuchman:
Follow up to https://github.com/tendermint/tendermint/pull/2364

- [x] fix the json name (https://github.com/tendermint/tendermint/pull/2364#discussion_r216708356)
- [x] expose `State.FilterTx` (https://github.com/tendermint/tendermint/pull/2364#discussion_r216706708)
- [x] docs!
  - include magic numbers in godocs, like https://github.com/tendermint/tendermint/pull/2364#discussion_r216707065
  - update `docs/` content and explain how maximums and consensus params work ...
- [ ] make int domains consistent (https://github.com/tendermint/tendermint/pull/2364#discussion_r217159079)
- [x] minor test cleanup https://github.com/tendermint/tendermint/pull/2364#discussion_r217159699

PR #2427, opened 2018-09-18
[WIPfollow up to removing some consensus params

melekes:
Refs #2382

* [ ] Wrote tests


state/tx_filter.go
```diff
@@ -0,0 +1,13 @@
+package state
+
+import "github.com/tendermint/tendermint/types"
+
+// TxFilter returns a function to filter transactions. The function limits the
+// size of a transaction to the maximum block's data size.
+func TxFilter(state State) func(tx types.Tx) bool {
```
melekes: test

types/block.go
```diff
@@ -205,22 +205,22 @@ func (b *Block) StringShort() string {
 //-----------------------------------------------------------------------------
 
 // MaxDataBytes returns the maximum size of block's data.
-func MaxDataBytes(maxBytes, valsCount, evidenceCount int) int {
+func MaxDataBytes(maxBytes, valsCount, evidenceCount int64) int64 {
```
melekes: valsCount and evidenceCount should probably remain ints

docs/spec/consensus/creating-proposal.md
```diff
@@ -0,0 +1,42 @@
...
+
+When we reap transactions from the mempool, we calculate maximum data
+size by subtracting maximum header size (`MaxHeaderBytes`), the maximum
+amino overhead for a block (`MaxAminoOverheadForBlock`), the size of
+the last commit (if present) and evidence (if present). While reaping
+we account for amino overhead for each transaction.
+
+```go
+func MaxDataBytes(maxBytes int64, valsCount, evidenceCount int) int64 {
+	return maxBytes -
```
melekes: What should we do if maxBytes is less than the sum of `MaxAminoOverheadForBlock+ MaxHeaderBytes+int64(valsCount)*MaxVoteBytes+int64(evidenceCount)*MaxEvidenceBytes`? 
MaxDataBytes will return a negative number. This in turn will lead to all txs being rejected by TxFilter (`int64(len(tx)) <= maxDataBytes`).

melekes: BlockSize.MaxBytes should account for header size + last commit size + possible evidence.

Right now, the minimum for BlockSize.MaxBytes is `1`.

melekes: I think it's ok to panic.

docs/spec/consensus/creating-proposal.md
```diff
@@ -0,0 +1,42 @@
+# Creating a proposal
```
ebuchman: Awesome thanks for writing this. We should link to it from https://github.com/tendermint/tendermint/tree/develop/docs/spec#consensus-protocol.

mempool/mempool.go
```diff
@@ -406,11 +406,11 @@ func (mem *Mempool) ReapMaxBytesMaxGas(maxBytes int, maxGas int64) types.Txs {
 		memTx := e.Value.(*mempoolTx)
 		// Check total size requirement
 		// amino.UvarintSize is not used here because it won't be possible to reuse buf
-		aminoOverhead := binary.PutUvarint(buf[:], uint64(len(memTx.tx)))
-		if maxBytes > -1 && totalBytes+len(memTx.tx)+aminoOverhead > maxBytes {
+		aminoOverhead := int64(binary.PutUvarint(buf[:], uint64(len(memTx.tx))))
```
ebuchman: Is having to do this everywhere a code smell for our code, or for Go itself, or for neither?

state/tx_filter_test.go
```diff
@@ -0,0 +1,46 @@
...
+		stateDB := dbm.NewDB("state", "memdb", os.TempDir())
+		state, err := LoadStateFromDBOrGenesisDoc(stateDB, genDoc)
+		require.NoError(t, err)
+
+		f := TxFilter(state)
+		assert.Equal(t, tc.isTxValid, f(tc.tx), "#%v", i)
+	}
+}
+
+func randomGenesisDoc() *types.GenesisDoc {
```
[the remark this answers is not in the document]
ebuchman: Doesn't being in the `_test.go` suffice?

types/block.go
```diff
@@ -15,7 +15,7 @@ import (
 
 const (
 	// MaxHeaderBytes is a maximum header size (including amino overhead).
-	MaxHeaderBytes = 511
+	MaxHeaderBytes = int64(511)
```
ebuchman: Would `MaxHeaderBytes int64 = 511` be better for these?

state/tx_filter_test.go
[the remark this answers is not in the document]
xla: They share the same namespace during testing, maybe I'm overly paranoid but usually like to flag it explicitely to not get tempted to use it in other contexts.

types/block.go
xla: 👍

mempool/mempool.go
melekes: for our code

melekes:
Thanks @xla amd @ebuchman for the reviews! PTAL

xla:
@melekes Looks like we have a cyclic dependency: https://circleci.com/gh/tendermint/tendermint/24992#tests/containers/1

ebuchman:
Still a cyclic dep issue?

types/testdouble/genesis.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"github.com/tendermint/tendermint/crypto/ed25519"
+	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/types"
+	tmtime "github.com/tendermint/tendermint/types/time"
+)
+
+// RandomGenesisDoc returns a random genesis document with one validator.
+// Genesis time is set to now().
+func RandomGenesisDoc() *types.GenesisDoc {
```
melekes: do folks think this is a good idea? maybe we should just export the same method from `types` package?

ebuchman: I didn't totally understand why this happened.

melekes: Wanted to reuse RandomGenesisDoc() function.

PR #2435, opened 2018-09-19
update adr-016

ebuchman:
Changes to the ADR taken from #2160

### DIFF
--- a/docs/architecture/adr-016-protocol-versions.md
+++ b/docs/architecture/adr-016-protocol-versions.md
@@ -4,9 +4,17 @@
 
 - How to / should we version the authenticated encryption handshake itself (ie.
   upfront protocol negotiation for the P2PVersion)
+- How to / should we version ABCI itself? Should it just be absorbed by the
+  BlockVersion?
 
 ## Changelog
 
+- 18-09-2018: Updates after working a bit on implementation
+    - ABCI Handshake needs to happen independently of starting the app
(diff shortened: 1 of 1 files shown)


ebuchman:
Another question here is how, if at all, we inform peers that our AppVersion has upgraded (this could happen without a Tendermint restart according to current design)

xla:
> Another question here is how, if at all, we inform peers that our AppVersion has upgraded (this could happen without a Tendermint restart according to current design)

What behaviour would result from this? Are we disconnecting from peers with incompatible app versions? Are we foresee incompatible app versions? If it is a p2p concern it might be worth introduce that as part of the messages exchanged by the PEXs and consequently deciding if the peer should be kept around.

ebuchman:
> Are we disconnecting from peers with incompatible app versions?

Current thinking is no, under the assumption that we want to be able to sync a peer across multiple app versions. But this also means we don't know if we're actually compatible with a peer when we connect to them, since they could be on an old height/version, and we won't find out until we actually disagree on blocks somewhere.

More prudent might be to consider tracking the correct version for each height, and enforcing that on peers, and having peers update each other when their version changes, but this sounds a bit complected, so maybe our initial assumption needs to be challenged.

xla:
@ebuchman Tried to capture the follow-up questions in #2448, please ammend/correct if necessary.

Issue #2436, opened 2018-09-19
Separate Handshaker from proxy.AppConns


ebuchman:
Currently we pass the Handshaker into the proxy.AppConns and let it handle the handshake with the app on `proxyApp.OnStart`. 

This isn't necessary and just hides the result of the Handshake inside the proxy app. Since the app will return its version during the handshake (see ADR-016), we need access to the result of the Handshake.

This was realized while working in #2160 and was added to the ADR-016 in #2435

Ultimately, the Handshake should probably return a full `State`. This would eliminate the need to reload the state after the handshake like we currently do in eg. https://github.com/tendermint/tendermint/blob/v0.24.0/node/node.go#L201

This will require some refactoring of the handshake logic, and will probably intersect with #1029.

We might also consider moving the handshake logic from `consensus` to `state` as it's not immediately clear it needs to be in `consensus`.

PR #2437, opened 2018-09-19
proxy: remove Handshaker from proxy pkg

ebuchman:
Handshaker was removed from proxy package so it can be called
independently of starting the abci app connections and can return a
result to the caller.

Ref https://github.com/tendermint/tendermint/issues/2436

* [ ] Wrote tests

### DIFF
--- a/consensus/replay_file.go
+++ b/consensus/replay_file.go
@@ -298,11 +298,16 @@ func newConsensusStateForReplay(config cfg.BaseConfig, csConfig *cfg.ConsensusCo
 
 	// Create proxyAppConn connection (consensus, mempool, query)
 	clientCreator := proxy.DefaultClientCreator(config.ProxyApp, config.ABCI, config.DBDir())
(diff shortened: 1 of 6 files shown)


consensus/replay_file.go
```diff
@@ -298,13 +298,18 @@ func newConsensusStateForReplay(config cfg.BaseConfig, csConfig *cfg.ConsensusCo
...
-		NewHandshaker(stateDB, state, blockStore, gdoc))
+	proxyApp := proxy.NewAppConns(clientCreator)
 	err = proxyApp.Start()
 	if err != nil {
 		cmn.Exit(fmt.Sprintf("Error starting proxy app conns: %v", err))
 	}
 
+	handshaker := NewHandshaker(stateDB, state, blockStore, gdoc)
+	err = handshaker.Handshake(proxyApp)
+	if err != nil {
```
xla: Can be collapsed into: `if err := handhsaker.Handshake(proxyApp); err != nil {`

Issue #2428, opened 2018-09-18
Group.RotateFile need call Flush() before rename.


goolAdapter:
Group.RotateFile need call Flush() before rename, Group.headBuf may have residual content, which may cause DataCorruptionError(WALDecoder presume whole message)，and can't recovery from a crash.

goolAdapter:
```
func MakeNBlocksWhitWAL(wal *baseWAL, numBlocks int) (err error) {
	config := getConfig()

	app := kvstore.NewPersistentKVStoreApplication(filepath.Join(config.DBDir(), "wal_generator"))

	logger := log.TestingLogger().With("wal_generator", "wal_generator")
	logger.Info("generating WAL (last height msg excluded)", "numBlocks", numBlocks)

	/////////////////////////////////////////////////////////////////////////////
	// COPY PASTE FROM node.go WITH A FEW MODIFICATIONS
	// NOTE: we can't import node package because of circular dependency
	privValidatorFile := config.PrivValidatorFile()
	privValidator := privval.LoadOrGenFilePV(privValidatorFile)
	genDoc, err := tmtypes.GenesisDocFromFile(config.GenesisFile())
	if err != nil {
		return errors.Wrap(err, "failed to read genesis file")
	}
	stateDB := db.NewMemDB()
	blockStoreDB := db.NewMemDB()
	state, err := sm.MakeGenesisState(genDoc)
	if err != nil {
		return errors.Wrap(err, "failed to make genesis state")
	}
	blockStore := bc.NewBlockStore(blockStoreDB)
	handshaker := NewHandshaker(stateDB, state, blockStore, genDoc)
	proxyApp := proxy.NewAppConns(proxy.NewLocalClientCreator(app), handshaker)
	proxyApp.SetLogger(logger.With("module", "proxy"))
	if err := proxyApp.Start(); err != nil {
		return errors.Wrap(err, "failed to start proxy app connections")
	}
	defer proxyApp.Stop()
	eventBus := tmtypes.NewEventBus()
	eventBus.SetLogger(logger.With("module", "events"))
	if err := eventBus.Start(); err != nil {
		return errors.Wrap(err, "failed to start event bus")
	}
	defer eventBus.Stop()
	mempool := sm.MockMempool{}
	evpool := sm.MockEvidencePool{}
	blockExec := sm.NewBlockExecutor(stateDB, log.TestingLogger(), proxyApp.Consensus(), mempool, evpool)
[...]

PR #2439, opened 2018-09-19
fix Group.RotateFile need call Flush() before rename. #2428

goolAdapter:
fix Group.RotateFile need call Flush() before rename. #2428

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -25,3 +25,4 @@ IMPROVEMENTS:
 
 BUG FIXES:
 - [node] \#2294 Delay starting node until Genesis time
+- [autofile] \#2428 Group.RotateFile need call Flush() before rename.
--- a/consensus/wal_test.go
+++ b/consensus/wal_test.go
@@ -4,6 +4,9 @@ import (
 	"bytes"
 	"crypto/rand"
 	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
 	// "sync"
 	"testing"
 	"time"
@@ -15,6 +18,46 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+func TestWALTruncate(t *testing.T) {
+	walDir, err := ioutil.TempDir("", "wal")
+	if err != nil {
+		panic(fmt.Errorf("failed to create temp WAL file: %v", err))
+	}
+
+	walFile := filepath.Join(walDir, "wal")
+
+	wal, err := NewWAL(walFile)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	wal.group.SetHeadSizeLimit(4096) //cause fragment
+	wal.Start()
+
+	err = WALGenerateNBlocks(wal.group, 60) //60 block's size nearly 70K
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	time.Sleep(5 * time.Second) //wait groupCheckDuration, make sure RotateFile run
+
+	wal.group.Flush()
+
+	h := int64(50)
+	gr, found, err := wal.SearchForEndHeight(h, &WALSearchOptions{})
+	assert.NoError(t, err, fmt.Sprintf("expected not to err on height %d", h))
+	assert.True(t, found, fmt.Sprintf("expected to find end height for %d", h))
+	assert.NotNil(t, gr, "expected group not to be nil")
+	defer gr.Close()
+
+	dec := NewWALDecoder(gr)
+	msg, err := dec.Decode()
+	assert.NoError(t, err, "expected to decode a message")
+	rs, ok := msg.Msg.(tmtypes.EventDataRoundState)
+	assert.True(t, ok, "expected message of type EventDataRoundState")
+	assert.Equal(t, rs.Height, h+1, fmt.Sprintf("wrong height"))
+}
 
 func TestWALEncoderDecoder(t *testing.T) {
 	now := tmtime.Now()
--- a/libs/autofile/group.go
+++ b/libs/autofile/group.go
@@ -266,6 +266,14 @@ func (g *Group) RotateFile() {
 
 	headPath := g.Head.Path
 
+	if err := g.headBuf.Flush(); err != nil {
+		panic(err) //panic is used for consistent with below
+	}
+
+	if err := g.Head.Sync(); err != nil {
+		panic(err)
+	}
+
 	if err := g.Head.closeFile(); err != nil {
 		panic(err)
 	}
--- a/consensus/wal_generator.go
+++ b/consensus/wal_generator.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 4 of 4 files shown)


consensus/wal_test.go
```diff
@@ -16,6 +19,46 @@ import (
...
+
+	wal.group.SetHeadSizeLimit(4096) //cause fragment
+	wal.Start()
+
+	err = WALGenerateNBlocks(wal.group, 60) //60 block's size nearly 70K
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	time.Sleep(5 * time.Second) //wait groupCheckDuration, make sure RotateFile run
```
melekes: We're trying to avoid long sleep calls in our tests. Do you think we can do something with this call? Maybe make groupCheckDuration a configuration variable

```go
# libs/autofile/group.go
type Config struct {
    GroupCheckDuration time.Duration
}

func OpenGroup(headPath string, cfg Config) (g *Group, err error) {

# consensus/wal.go
type Config struct {
    GroupConfig
}

func NewWAL(walFile string, cfg Config) (*baseWAL, error) {
```

consensus/wal_test.go
```diff
@@ -16,6 +19,46 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestWALTruncate(t *testing.T) {
+	walDir, err := ioutil.TempDir("", "wal")
+	if err != nil {
+		panic(fmt.Errorf("failed to create temp WAL file: %v", err))
+	}
+
```
melekes: `defer os.RemoveAll(walDir)` ?

consensus/wal_test.go
```diff
@@ -16,6 +19,46 @@ import (
...
+	walFile := filepath.Join(walDir, "wal")
+
+	wal, err := NewWAL(walFile)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	wal.group.SetHeadSizeLimit(4096) //cause fragment
```
melekes: again, better to pass as a config option (although this requires some work)

consensus/wal_test.go
```diff
@@ -16,6 +19,46 @@ import (
...
+	wal.Start()
+
+	err = WALGenerateNBlocks(wal.group, 60) //60 block's size nearly 70K
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	time.Sleep(5 * time.Second) //wait groupCheckDuration, make sure RotateFile run
+
+	wal.group.Flush()
```
melekes: `wal.Group().Flush()`

generally you should not access private fields / methods (even though you can) - sign that there's something wrong with your API

CHANGELOG_PENDING.md
```diff
@@ -25,3 +25,4 @@ IMPROVEMENTS:
 
 BUG FIXES:
 - [node] \#2294 Delay starting node until Genesis time
+- [autofile] \#2428 Group.RotateFile need call Flush() before rename.
```
melekes: Could you please add your github username in the end? `before rename (@goolAdapter)`

consensus/wal_test.go
goolAdapter: Not only the GroupCheckDuration, but HeadSizeLimit also need to config, this config filed seems only need changed by test. I added some member func to Group, left NewWAL concise.

libs/autofile/group.go
```diff
@@ -122,6 +124,20 @@ func (g *Group) Close() {
 	g.mtx.Unlock()
 }
 
+// SetGroupCheckDuration allows you to overwrite default groupCheckDuration, this must call before Start().
+func (g *Group) SetGroupCheckDuration(duration time.Duration) {
```
melekes: please don't. let's use config for this

consensus/wal_test.go
```diff
@@ -32,17 +34,23 @@ func TestWALTruncate(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	wal.group.SetHeadSizeLimit(4096) //cause fragment
+	//this magic number 4K can truncate the content when RotateFile. defaultHeadSizeLimit(10M) is hard to simulate.
+	wal.Group().SetHeadSizeLimit(4096)
```
melekes: same here. we shouldn't expose SetXXX methods for something which is not supposed to change during the WAL group lifetime.

melekes: We would appreciate if you refactor this part! Technical debt tends to accumulate over time, so we need to refactor to maintain Tendermint in good shape continually.

libs/autofile/cmd/logjack.go
```diff
@@ -39,13 +39,12 @@ func main() {
 	}
 
 	// Open Group
-	group, err := auto.OpenGroup(headPath)
+	group, err := auto.OpenGroup(headPath, auto.SetHeadSizeLimit(chopSize), auto.SetTotalSizeLimit(limitSize))
```
melekes: if these are options now, they must not have a `Set` prefix

consensus/wal.go
```diff
@@ -73,13 +73,13 @@ type baseWAL struct {
 	enc *WALEncoder
 }
 
-func NewWAL(walFile string) (*baseWAL, error) {
+func NewWAL(walFile string, groupOptions ...func(*auto.Group)) (*baseWAL, error) {
```
melekes: what are the advantages of options over config struct?

libs/autofile/group.go
```diff
@@ -122,11 +150,11 @@ func (g *Group) Close() {
 	g.mtx.Unlock()
 }
 
-// SetHeadSizeLimit allows you to overwrite default head size limit - 10MB.
-func (g *Group) SetHeadSizeLimit(limit int64) {
+// GroupCheckDuration returns the current groupCheckDuration.
+func (g *Group) GroupCheckDuration() time.Duration {
```
melekes: unused?

libs/autofile/cmd/logjack.go
goolAdapter: yes， Set* -> With* looks better?

consensus/wal.go
goolAdapter: if use config struct, all the places use NewWAL need modify to init and add a param.

libs/autofile/group.go
goolAdapter: yes,  as HeadSizeLimit, expose only for check the current config.

libs/autofile/cmd/logjack.go
melekes: GroupHeadSizeLimit()
GroupTotalSizeLimit()

consensus/wal.go
melekes: > NewWAL need modify to init

that's fine.

> add a param

we can say that if param == 0, then default value will be applied

but let's leave options

libs/autofile/group.go
melekes: can you remove it then?

libs/autofile/cmd/logjack.go
goolAdapter: done!

libs/autofile/group.go
goolAdapter: done!

xla:
@goolAdapter Could you resolve the conflicts with `develop` then this should be good to go.

Issue #2313, opened 2018-09-02
ADR Assignment


ebuchman:
All architectural designs and changes should first be documented in an Architectural Design Record. The current ones can be found in the [ADR directory](https://github.com/tendermint/tendermint/tree/master/docs/architecture).

This issue is a source of truth for assigning the next ADR number. Before giving a number to a new ADR, make sure you update this comment.

021: ABCI events
022: ABCI errors
023: ABCI propose tx
024: Vote Sign Bytes
025: Commit Deduplication
026: General Merkle Proofs
027: Block compression
028: LibP2P integration
029: CheckTx for block validity
030: Consensus Refactor
031: Changelog Structure
032: Result Tag Proofs 
033: PubSub
034: PrivValidator file structure and encryption
035: Documentation
036: Empty Blocks via ABCI
037: DeliverTxs 
038: Non-zero start height
039: PeerBehaviour
040: Refactor blockchain reactor spec #3506 
041: Proposer Selection via ABCI
042: State sync
043: Blockchain riri-org 
044: Lite Client with Weak Subjectivity
045: ABCI Evidence Handling
046: Language agnostic testing for light client
047: InvalidHeaderEvidence to punish lunatic validators
050: Improved trusted peering
051: Double Signing Protection
052: Tendermint Mode 
053: State sync prototype

xla:
*Added 030: Consenus Refactor*

marbar3778:
I propose we move this to the readme.md of the architecture folder in docs. or is there a specific reason it is done this way? @ebuchman @zramsay

ebuchman:
It's basically so we can easily update it without creating PRs.

marbar3778:
I think it should be part of the process of making an adr. Will add it to README but we can keep it here as well

marbar3778:
should create empty ADR files for the ADR's that have not been written yet. Its a bit odd that we skip numbers currently

cc @tessr @melekes @erikgrinaker

erikgrinaker:
>should create empty ADR files for the ADR's that have not been written yet. Its a bit odd that we skip numbers currently

Yes please. Can we make that the standard ADR allocation mechanism? This issue is not very discoverable.

melekes:
Not even sure why we need this issue. If someone creates an ADR with the same number (probability of that is already quite low given the size of the team AND awareness of each other's work), git will just complain and we'll handle collision by hand.

PR #2441, opened 2018-09-19
[ADR][DRAFT] 021: Check block txs before prevote.md

james-ray:
* [ ] Wrote tests

discuss #2384

### DIFF
--- a/docs/architecture/adr-021-check block txs before prevote.md
+++ b/docs/architecture/adr-021-check block txs before prevote.md
@@ -1,0 +1,119 @@
+# ADR 021: check block txs before prevote
+
+## Changelog
+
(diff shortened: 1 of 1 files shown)


ebuchman:
This is a really great start, thanks for doing this @james-ray. I think there are still a few open questions here we need to address, for instance do we really want a new message (CheckBlock) or just to allow CheckTx to be called on the ConsensusConnection? Also I think this ADR will need some more explanation of the problem space and the solution, but I'd be happy to take over from here.

Oh, it should also be called ADR-029 according to https://github.com/tendermint/tendermint/issues/2313.

Thanks!

ebuchman:
Updated in https://github.com/tendermint/tendermint/pull/2544

Issue #1622, opened 2018-05-24
Why `Vote.SignBytes` is a JSON string?


nnkken:
I'm trying to implement a custom blockchain using Tendermint, and I want to have a Relay contract on Ethereum so that my ERC-20 token can transfer between Ethereum and the Tendermint blockchain.

To do so, I need to write a smart contract on Ethereum verifying the block header from Tendermint. While other parts go well, the format of Precommits annoys me.

The signature are signed on the SHA-256 hash of the SignBytes, which are JSON strings, so to extract block header hash from the message, the smart contract needs to parse the JSON string, which is quite expensive.

I'm now finding work around on this problem, and my question is, why Tendermint is designed to sign a JSON string instead of a binary one, say a structure encoded by Amino, or the Merkle root of the information in the precommits (just like the block header hash)?

jaekwon:
Amino would work... do you think that would save on EVM gas costs?
The Merkle root won't work because then we're left with the problem of information withholding... and besides, the Round field only exists in the vote.

nnkken:
Amino will be muching better than JSON string in EVM.

Are the SignBytes directly passed to other validators?
I thought the content of the vote are serialized into another format, packed together with the signature, then other validators receiving this could check the signature by deserialize -> generate SignBytes -> check the signature with SignBytes.
In this way, the content of the vote can be different from the SignBytes

Besides, when I look at the format of current SignBytes:

`{"@chain_id":"test-chain-dvwvaa","@type":"vote","block_id":{"hash":"044B1440465176F74042ACF3493A6BEFB6616E1F","parts":{"hash":"7C563049D57AB3D16817CB9E739303A02D085E9F","total":1}},"height":2,"round":0,"timestamp":"2018-05-24T03:52:21.310Z","type":2}`

I think most of the parts (all except timestamp) are fixed for a successful commit, so if timestamp is not part of the SignBytes, then when validating votes for a block header, the parser will only have to parse the vote once and then reuse the same hash of SignBytes to validate all signatures.
Since timestamp is important, maybe the signature can sign `sha256(sha256(chain_id, type, ...), timestamp)`, such that for each vote, only the timestamps have to be parsed instead of the whole vote.

ebuchman:
> In this way, the content of the vote can be different from the SignBytes

[...]

PR #2445, opened 2018-09-19
[ADR][DRAFT] 024: SignBytes and validator types in privval

Liamsi:
Work in progress: Quick fledged draft summarizing https://github.com/tendermint/tendermint/issues/1622

If time allows, please review for overall sanity (not typos/nitpicks):
https://github.com/tendermint/tendermint/blob/5d46a2d49dba4da726463ba43058c6b230e3d24a/docs/architecture/adr-02x-sign-bytes.md

* [ ] Wrote tests

### DIFF
--- a/docs/architecture/adr-024-sign-bytes.md
+++ b/docs/architecture/adr-024-sign-bytes.md
@@ -3,7 +3,7 @@
 ## Context
 
 Currently, the messages exchanged between tendermint and a (potentially remote) signer/validator, 
-namely votes, proposals, and hearbeats, are encoded as a JSON string 
+namely votes, proposals, and heartbeats, are encoded as a JSON string 
 (e.g., via `Vote.SignBytes(...)`) and then 
 signed . JSON encoding is sub-optimal for both, hardware wallets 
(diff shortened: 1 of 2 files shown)


Liamsi:
@tarcieri also suggested as while ago to include the public-key / fingerprint thereof into each "SignRequest" sent by tendermint. E.g. the KMS holds a bunch of key-pairs and needs to know which priv-key of those to use to sign Votes, Proposals, or Hearbeats. This is probably quite important. I still left it away here as I assume the implied changes to be quite involved already.

zmanian:
IMO there is a lot of scope for improving the validator socket protocol lets focus on getting the MVP working with KMS before we add important features like key ids and and error handling.

Liamsi:
@zmanian: OK, I'll open an issue and a PR that just moves `SignBytes(chainId string)` to be amino encoded and leave everything else as it is now (including leaving the fields variable sized). We can change this incrementally instead. It is still worthwhile to have this discussion earlier than later. As I assume we will change the structs before launch (and the kms needs to catch up with these changes then).

xla:
Error handling should be considered part of an MVP.

PR #2446, opened 2018-09-19
add missing document

WALL-E:


xla:
Thanks for the PR, but I don't see what value the comment would add. Could you explain why it should be there?

WALL-E:
Just keep the style consistent with the context.There really isn't any value

PR #2449, opened 2018-09-19
Minor changelog fixes

ebuchman:
We should do a better job of enforcing changelog standards, of course it should be documented how it works and how to think about it in the first place. Maybe I'll write an ADR for it.

Seems like we may have missed a bunch of stuff though, so we should catch up in preparation for v0.25. It's a relatively minor release, despite at least a few breaking changes.


CHANGELOG_PENDING.md
```diff
@@ -5,25 +5,26 @@ Special thanks to external contributors with PRs included in this release:
...
 
 
 FEATURES:
-  * \#2310 Mempool is now aware of the MaxGas requirement
 
 IMPROVEMENTS:
 - [mempool] [\#2399](https://github.com/tendermint/tendermint/issues/2399) Make mempool cache a proper LRU (@bradyjoestar)
-- [types] add Address to GenesisValidator [\#1714](https://github.com/tendermint/tendermint/issues/1714)
+- [types] [\#1714](https://github.com/tendermint/tendermint/issues/1714) Add Address to GenesisValidator
 - [metrics] `consensus.block_interval_metrics` is now gauge, not histogram (you will be able to see spikes, if any)
```
ebuchman: is this actually a breaking change? technically the metrics are an API exposed to other processes and hence need to be properly covered by SemVer unless we add an exemption in the readme

CHANGELOG_PENDING.md
```diff
@@ -5,25 +5,26 @@ Special thanks to external contributors with PRs included in this release:
...
 
 * CLI/RPC/Config
+  * [rpc] [\#2391](https://github.com/tendermint/tendermint/issues/2391) /status `result.node_info.other` became a map
 
 * Apps
-  [rpc] /status `result.node_info.other` became a map #[2391](https://github.com/tendermint/tendermint/issues/2391)
+  * [mempool] \#2310 Mempool tracks the `ResponseCheckTx.GasWanted` and enforces `ConsensusParams.BlockSize.MaxGas` on proposals.
 
 * Go API
-  * \#2310 Mempool.ReapMaxBytes -> Mempool.ReapMaxBytesMaxGas
```
ValarDragon: Not sure why this was deleted? This is a breaking API change, as its changed the mempool interface. The new line in apps does not note this API change.

CHANGELOG_PENDING.md
melekes: > is this actually a breaking change?

don't think so. prometheus won't crash because metric's type got changed

CHANGELOG_PENDING.md
ebuchman: Fortunately for us maintainers, the `mempool` Go API is not covered by our SemVer, so we can break it as much as we want without worrying about the changelog. See https://github.com/tendermint/tendermint/#semver : 

```
To provide some stability to Tendermint users in these 0.X.X days, the MINOR version is used to signal breaking changes across a subset of the total public API. This subset includes all interfaces exposed to other processes (cli, rpc, p2p, etc.), but does not include the in-process Go APIs.
```

CHANGELOG_PENDING.md
ebuchman: will the code of any clients need to change ?

Issue #2460, opened 2018-09-21
StartHTTPAndTLSServer() gets stuck if http.ServeTLS() fails


alessio:
**Tendermint version**: v0.23.1-rc0

**ABCI app**: gaia

**Environment**:
- **OS**: Ubuntu 18.04.1 LTS
- **Install tools**:
- **Others**:

**What happened**:
When passing a non-existing keyFile to `StartHTTPAndTLSServer()`, the process logs the error and hangs.

**What you expected to happen**:
`StartHTTPAndTLSServer()` pass on the client errors returned by `http.ServeTLS()` - maybe via an error chan?

**Have you tried the latest version**: no

**How to reproduce it** (as minimally and precisely as possible):

1. `git clone -b alessio/tendermint-starttls-bug https://github.com/cosmos/cosmos-sdk`
2. `make install`
3. Setup `gaiad`, run `gaiacli rest-server` and pass a non existing key file through `--ssl-keyfile`.

**Logs (paste a small part showing an error (< 10 lines) or link a pastebin, gist, etc. containing more of the log file)**:

An example follows:

```bash
$ gaiacli rest-server --home=./gaiacli --chain-id=test --ssl-hosts=127.0.0.1 --ssl-certfile=/tmp/cert_849474202 --ssl-keyfile=/tmp/key_315202865x
I[09-21|13:29:27.065] Starting RPC HTTPS server on tcp://localhost:1317 (cert: "/tmp/cert_849474202", key: "/tmp/key_315202865x") module=rest-server 
I[09-21|13:29:27.066] SHA256 Fingerprint=0B:ED:9A:AA:A2:D1:7E:B2:53:56:F6:FC:C0:E6:1A:69:70:21:A2:B0:90:FC:AF:BB:EF:AE:2C:78:52:AB:68:40 module=rest-server 
I[09-21|13:29:27.066] REST server started                          module=rest-server 
E[09-21|13:29:27.066] RPC HTTPS server stopped                     module=rest-server err="open /tmp/key_315202865x: no such file or directory"
```

PR #2461, opened 2018-09-21
Pass http.ServeTLS() errors back to the caller

alessio:
When http.ServeTLS() fails, StartHTTPAndTLSServer() logs the error then hangs for ever.
It should instead pass the error back to the caller.

Closes: #2460

* [x] Wrote tests

### DIFF
--- a/rpc/lib/server/http_server.go
+++ b/rpc/lib/server/http_server.go
@@ -56,14 +56,13 @@ func StartHTTPServer(
 		listener = netutil.LimitListener(listener, config.MaxOpenConnections)
 	}
 
-	err = http.Serve(
-		listener,
-		RecoverAndLogHandler(maxBytesHandler{h: handler, n: maxBodyBytes}, logger),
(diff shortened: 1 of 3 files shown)


rpc/lib/server/http_server.go
```diff
@@ -102,15 +102,16 @@ func StartHTTPAndTLSServer(
 		listener = netutil.LimitListener(listener, config.MaxOpenConnections)
 	}
 
-	go func() {
```
melekes: I will ask you to apply the same changes to `StartHTTPServer` func. We need to keep them consistent, so we can swap HTTP server with HTTPS.

alessio: Will do shortly

alessio: @melekes Done

alessio: mmm tests failed, I'm on it

Issue #1692, opened 2018-06-05
consensus: replace synchronous events system with a function call


ebuchman:
Following up on https://github.com/tendermint/tendermint/issues/951 and https://github.com/tendermint/tendermint/issues/1495

Currently the ConsensusState notifies the ConsensusReactor of messages to broadcast using a synchronous event system.

Instead, we should use an `update` method in the ConsensusState (ala https://github.com/tendermint/tendermint/issues/1495#issuecomment-393140116) to process incoming messages and return any messages to broadcast to peers.

jackzampolin:
Closing this in favor of the currently in progress refactor of the reactors

PR #2462, opened 2018-09-21
[ADR][DRAFT] 030: Consensus refactor

xla:
Refs #1692

### DIFF
--- a/docs/architecture/adr-030-consensus-refactor.md
+++ b/docs/architecture/adr-030-consensus-refactor.md
@@ -1,0 +1,152 @@
+# ADR 030: Consensus Refactor
+
+## Context
+
+One of the biggest challenges this project faces is to proof that the
+implementations of the specifications are correct, much like we strive to
+formaly verify our alogrithms and protocols we should work towards high
+confidence about the correctness of our program code. One of those is the core
+of Tendermint - Consensus - which currently resides in the `consensus` package.
+Over time there has been high friction making changes to the package due to the
+algorithm being scattered in a side-effectful container (the current
+`ConsensusState`). In order to test the algorithm a large object-graph needs to
+be set up and even than the non-deterministic parts of the container makes will
+prevent high certainty. Where ideally we have a 1-to-1 representation of the
+[spec](https://github.com/tendermint/spec), ready and easy to test for domain
+experts.
+
+Addresses:
+
+- [#1495](https://github.com/tendermint/tendermint/issues/1495)
+- [#1692](https://github.com/tendermint/tendermint/issues/1692)
+
+## Decision
+
+To remedy these issues we plan a gradual, non-invasive refactoring of the
+`consensus` package. Starting of by isolating the consensus alogrithm into
+a pure function and a finite state machine to address the most pressuring issue
+of lack of confidence. Doing so while leaving the rest of the package in tact
+and have follow-up optional changes to improve the sepration of concerns.
+
+### Implementation changes
+
+The core of Consensus can be modelled as a function with clear defined inputs:
+
+* `State` - data container for current round, height, etc.
+* `Event`- significant events in the network
+
+producing clear outputs;
+
+* `State` - updated input
+* `Message` - signal what actions to perform
+
+```go
+type Event int
+
+const (
+	EventUnknown Event = iota
+	EventProposal
+	Majority23PrevotesBlock
+	Majority23PrecommitBlock
+	Majority23PrevotesAny
+	Majority23PrecommitAny
+	TimeoutNewRound
+	TimeoutPropose
+	TimeoutPrevotes
+	TimeoutPrecommit
+)
+
+type Message int
+
+const (
+	MeesageUnknown Message = iota
+	MessageProposal
+	MessageVotes
+	MessageDecision
+)
+
+type State struct {
+	height      uint64
+	round       uint64
+	step        uint64
+	lockedValue interface{} // TODO: Define proper type.
+	lockedRound interface{} // TODO: Define proper type.
+	validValue  interface{} // TODO: Define proper type.
+	validRound  interface{} // TODO: Define proper type.
+	// From the original notes: valid(v)
+	valid       interface{} // TODO: Define proper type.
+	// From the original notes: proposer(h, r)
+	proposer    interface{} // TODO: Define proper type.
+}
+
+func Consensus(Event, State) (State, Message) {
+	// Consolidate implementation.
+}
+```
+
+Tracking of relevant information to feed `Event` into the function and act on
+the output is left to the `ConsensusExecutor` (formerly `ConsensusState`). 
+
+Benefits for testing surfacing nicely as testing for a sequence of events
+against algorithm could be as simple as the following example:
+
+``` go
+func TestConsensusXXX(t *testing.T) {
+	type expected struct {
+		message Message
+		state   State
+	}
+
+	// Setup order of events, initial state and expectation.
+	var (
+		events = []struct {
+			event Event
+			want  expected
+		}{
+		// ...
+		}
+		state = State{
+		// ...
+		}
+	)
+
+	for _, e := range events {
+		sate, msg = Consensus(e.event, state)
+
+		// Test message expectation.
+		if msg != e.want.message {
+			t.Fatalf("have %v, want %v", msg, e.want.message)
+		}
+
+		// Test state expectation.
+		if !reflect.DeepEqual(state, e.want.state) {
+			t.Fatalf("have %v, want %v", state, e.want.state)
+		}
+	}
+}
+```
+
+### Implementation roadmap
+
+* implement proposed implementation
+* replace currently scattered calls in `ConsensusState` with calls to the new
+  `Consensus` function
+* rename `ConsensusState` to `ConsensusExecutor` to avoid confusion
+* propose design for improved separation and clear information flow between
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-030-consensus-refactor.md
```diff
@@ -0,0 +1,152 @@
...
+	lockedRound interface{} // TODO: Define proper type.
+	validValue  interface{} // TODO: Define proper type.
+	validRound  interface{} // TODO: Define proper type.
+	// From the original notes: valid(v)
+	valid       interface{} // TODO: Define proper type.
+	// From the original notes: proposer(h, r)
+	proposer    interface{} // TODO: Define proper type.
+}
+
+func Consensus(Event, State) (State, Message) {
```
jaekwon: I think it's a bit weird to call this "Consensus".  It's a very small part of consensus... and "step" is overloaded... maybe "ConsensusIota". 

While agree that ConsensusState can be split into ConsensusExecutor and ConsensusState, I'd still want to keep the exact same method signature:

```go
func (cs *ConsensusState) enterNewRound(height int64, round int) {}
```

Reason: We can't remove things like .Votes and .ProposalBlockParts from ConsensusState and call it ConsensusState -- they're vital parts of "state" just as necessary as those other fields.  And those structures aren't cheap to copy, esp for chains with larger validator sets, and the best way to mutate those things are via a pointer, w/ the intuitive interfaces for VoteSet, PartSet.  Ultimately there's a map, and maps shouldn't be copied around so lightly.

However, you can still set up the test the way you want, by making the test harness (inefficiently) copy the ConsensusState deeply before each iota operation, so you can still get the effect that you want, which I think is great.

So basically, I would prefer:

```go
func (cs *ConsensusState) RunEvent(event) ([]Message) {}
```

"Message" should be a slice.  Returning messages from `ConsensusState.***()` instead of calling eventbus methods in the ConsensusExecutor does make sense.

jaekwon:
While agree that ConsensusState can be split into ConsensusExecutor and ConsensusState, I'd still want to keep most of the exact same method signature:

```go
func (cs *ConsensusState) enterNewRound(height int64, round int) {}
```

We can't remove things like .Votes and .ProposalBlockParts from ConsensusState and call it ConsensusState -- they're vital parts of "state" just as necessary as those other fields.  And those structures aren't cheap to copy, esp for chains with larger validator sets, and the best way to mutate those things are via a pointer, w/ the intuitive interfaces for VoteSet, PartSet.  Ultimately there's a map, and maps shouldn't be copied around so lightly.  So, we should avoid actually implementing "pure" functions, one because of efficiency.

Also, two more reasons why pure functions are imo worse than impure functions for simple mutation operations:

```go
package main

import (
	"fmt"
)

//----------------------------------------
// Pure functions create opportunities to mask variables accidentally.

func runEvent(state int, event int) int {
	return state + event
}

func pureFunctionBug() {
	state := 1

	if true {
		event := 1
		state := runEvent(state, event)
		event = 2
		state = runEvent(state, event)
	}

	fmt.Println(state) // This prints 1
}

//----------------------------------------
// Pure functions require more confusing code for complex types.

type complexType struct {
	children []child
	fieldA   int
}

type child struct {
	fieldB int
	fieldC int
}

func copyComplexChildren(x []child) (r []child) {
	return r
}

// This is confusing and difficult to simplify.
func pureMutation(x complexType) complexType {
	y := x // shallow copy
	y.children = make([]child, len(x.children))
	copy(y.children, x.children)
	y.children[0].fieldB = -1
	return y
}

// This is much easier to understand
func (x *complexType) impureMutation() {
	x.children[0].fieldB = -2
}

func complexTypeNonergonomics() {

	x := complexType{
		fieldA: 0,
		children: []child{
			{1, 2},
			{2, 3},
		},
	}

	y := pureMutation(x)
	fmt.Println(y)

	x.impureMutation()
	fmt.Println(x)

}

func main() {
	pureFunctionBug()
	complexTypeNonergonomics()
}
```

However, you can still set up the test the way you want, by making the test harness (inefficiently) copy the ConsensusState deeply before each iota operation, so you can still get the effect that you want, which I think is great.

So basically, I would prefer:

```go
func (cs *ConsensusState) RunEvent(event) ([]Message) {}
func (cs ConsensusState) DeepCopy() (ConsensusState) {}
```

Another note: "Message" should be a slice.  Returning messages from `ConsensusState.***()` instead of calling eventbus methods in the ConsensusExecutor does make sense.

milosevic:
I really like the proposal!Regarding @jaekwon comment, I don't think that we need to copy data structures around. For example, after ConsensusExecutor detects Majority23PrevotesBlock, it will pass to Consensus event Majority23PrevotesBlock that can contain height, round and BlockID. Consensus can then transition to a new step and output some messages. Once we reach consensus Consensus will signal Decision(height and BlockID) to ConsensusExecutor that can then proceed by executing the corresponding block. At the Consensus level it is sufficient to work with BlockID, we don't need to know what is the corresponding block; we only need information if we have received ProposalBlock. I agree that Vote is essential part of Consensus, but in this proposal it is modelled as a Message that is being sent. ProposalBlockPart is implementation detail of sending Proposal Block efficiently over gossip network, but at the core Consensus level we don't care about this. We only care about the fact that all parts are received so we have received complete Proposal. Note that in case when Proposal contains re-proposed block, we can wait for all Prevotes at the ConsensusExecutor level and then just signal it to Consensus that Proposal message for height, round and valid value and valid round is received.

docs/architecture/adr-030-consensus-refactor.md
```diff
@@ -0,0 +1,152 @@
...
+* `State` - data container for current round, height, etc.
+* `Event`- significant events in the network
+
+producing clear outputs;
+
+* `State` - updated input
+* `Message` - signal what actions to perform
+
+```go
+type Event int
```
melekes: Could be `uint8`, I think.

docs/architecture/adr-030-consensus-refactor.md
```diff
@@ -0,0 +1,152 @@
...
+	Majority23PrecommitBlock
+	Majority23PrevotesAny
+	Majority23PrecommitAny
+	TimeoutNewRound
+	TimeoutPropose
+	TimeoutPrevotes
+	TimeoutPrecommit
+)
+
+type Message int
```
melekes: same

docs/architecture/adr-030-consensus-refactor.md
```diff
@@ -0,0 +1,152 @@
...
+
+const (
+	MeesageUnknown Message = iota
+	MessageProposal
+	MessageVotes
+	MessageDecision
+)
+
+type State struct {
+	height      uint64
```
melekes: I thought we've settled on `int64` https://blog.cosmos.network/choosing-a-type-for-blockchain-height-beware-of-unsigned-integers-714804dddf1d. @xla want to explain the choice? (note: I am not entirely opposite to uint)

docs/architecture/adr-030-consensus-refactor.md
```diff
@@ -0,0 +1,152 @@
...
+	round       uint64
+	step        uint64
```
melekes: same. uint8 is more than enough

docs/architecture/adr-030-consensus-refactor.md
```diff
@@ -0,0 +1,152 @@
...
+		}{
+		// ...
+		}
+		state = State{
+		// ...
+		}
+	)
+
+	for _, e := range events {
+		sate, msg = Consensus(e.event, state)
```
melekes: `sate` -> `state`

docs/architecture/adr-030-consensus-refactor.md
```diff
@@ -0,0 +1,152 @@
...
+			t.Fatalf("have %v, want %v", state, e.want.state)
+		}
+	}
+}
+```
+
+### Implementation roadmap
+
+* implement proposed implementation
+* replace currently scattered calls in `ConsensusState` with calls to the new
```
melekes: Does it make sense to benchmark (both CPU and memory) current and new implementation and compare them? Doing this might rectify the "pure functions are slow" argument or confirm it 😄

milosevic:
The proposed separation of concerns do not constrain us from implementing it less efficiently then the current one. I even think that it will open more space for performance improvements as the roles will be more clear. But having perf results and checking for regression are definitely great idea.

milosevic:
In the current version of the code, important algorithmic control flows are present a in form of comments. See for example: https://github.com/tendermint/tendermint/blob/5c6999cf8f0fc09bec6da16d7038a0f1b6ca4d71/consensus/state.go#L832 or https://github.com/tendermint/tendermint/blob/5c6999cf8f0fc09bec6da16d7038a0f1b6ca4d71/consensus/state.go#L984. The proposed design can normally help us collocate key control logic in Consensus function which will make reading and modifying core algorithm much easier.

docs/architecture/adr-030-consensus-refactor.md
ValarDragon: I think uint makes sense since we only increment by one.

docs/architecture/adr-030-consensus-refactor.md
ValarDragon: For round, I think we can use a uint32. (at `.5` seconds timeout increase per round, we have timeout support for over a millenium :))

ebuchman:
Follow up in https://github.com/tendermint/tendermint/issues/2547

PR #2465, opened 2018-09-21
add wiki

zramsay:
* [ ] Wrote tests


ebuchman:
Maybe we can make a branch called `ref/wiki` and leave it there?

ebuchman:
> Branch based storage seems dangerous, if someone down the road does some drive-by cleanup and stumbles upon a stale branch.

What about `do-not-delete/wiki`? Just feels like it's not worth merging it in.

ebuchman:
https://github.com/tendermint/tendermint/tree/do-not-delete/wiki

PR #2473, opened 2018-09-23
Release/v0.25.0

ebuchman:
Circle failed non-deterministically while installing dependencies the first time we merged v0.25.0 to master, so the binaries didn't get built. Lucky us, since some of the new docs links were broken. So we take the opportunity here to try merging a v0.25.0 release branch to master again, with fixed links.

@greg-szabo how should we handle non-deterministic failures like this? I re-ran the test and it passed the second time - should we make that trigger the release build ?

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -106,14 +106,6 @@ module.exports = {
           "/spec/abci/apps",
           "/spec/abci/client-server"
         ]
-      },
-      {
-        title: "Research",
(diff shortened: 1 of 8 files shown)


greg-szabo:
I was under the assumption that the CircleCI tests run when the PR is started. In that case there's nothing to worry about. The release process only starts when the PR is _committed_ to master. Hence you have all the time in the world to fix the issues in the PR.

Issue #2434, opened 2018-09-19
Unresponsive to signal interupts while sleeping for genesis time


ebuchman:
As of https://github.com/tendermint/tendermint/pull/2389 we sleep until the genesis time is reached, but we don't set up our signal catching until we start the node, so we don't respond to CTRL-C while sleeping. We should probably use a select statement over a ticker and signal channel instead of the sleep.

xla:
Alternatively can't we setup our signal trapping earlier?

melekes:
Merged to develop.

PR #2479, opened 2018-09-24
Make node respond to os interrupts while sleeping until genesis time

melekes:
Refs #2434 

* [ ] ~~Wrote tests~~

### DIFF
--- a/p2p/transport.go
+++ b/p2p/transport.go
@@ -207,7 +207,11 @@ func (mt *MultiplexTransport) Dial(
 func (mt *MultiplexTransport) Close() error {
 	close(mt.closec)
 
-	return mt.listener.Close()
+	if mt.listener != nil {
+		return mt.listener.Close()
+	}
+
+	return nil
 }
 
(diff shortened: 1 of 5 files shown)


p2p/transport.go
```diff
@@ -207,7 +207,11 @@ func (mt *MultiplexTransport) Dial(
 func (mt *MultiplexTransport) Close() error {
 	close(mt.closec)
 
-	return mt.listener.Close()
+	if mt.listener != nil {
```
xla: Is this to not panic when `Close` is called before `Listen`? When does that happen?

melekes: Yes. There's a problem with the current common#Service implementation where it considers a service started before OnStart() finishes. https://github.com/tendermint/tendermint/blob/0c9c3292c918617624f6f3fbcd95eceade18bcd5/libs/common/service.go#L131-L139

Therefore, calling IsRunning() can return true even though the service is not yet started (or partially started)

Issue #2260, opened 2018-08-22
Register autofile signal listeners explicitly rather than in `init()`


silasdavis:
Recent change to:

https://github.com/tendermint/tendermint/blob/master/libs/autofile/sighup_watcher.go#L19-L25

To capture interrupt signal (and so disable default golang runtime behaviour) stop unrelated things we have in our binary from exiting on Ctrl-C (e.g. grpcServer.Serve). It would be better hygiene if `NewNode` could activate these explicitly rather than doing it globally like this in init.

PR #2480, opened 2018-09-24
libs: Handle SIGHUP explicitly inside autofile

melekes:
Refs #2260

* [x] Wrote tests


libs/autofile/autofile.go
```diff
@@ -32,50 +34,70 @@ if err != nil {
...
-	ticker        *time.Ticker
-	tickerStopped chan struct{} // closed when ticker is stopped
-	mtx           sync.Mutex
-	file          *os.File
+	ID   string
+	Path string
+
+	closeTicker       *time.Ticker
+	closeTickerStopCh chan struct{} // closed when closeTicker is stopped
+	hup               chan os.Signal
```
melekes: `hupc` or `hupCh`?

libs/autofile/autofile_test.go
```diff
@@ -87,7 +83,4 @@ func TestOpenAutoFilePerms(t *testing.T) {
 	} else {
 		t.Errorf("unexpected error %v", e)
 	}
-
-	err = af.Close()
```
melekes: note to reviewers: OpenAutoGroup now returns nil if there's an error

libs/autofile/autofile.go
xla: `ch` and `Ch` have a very distinct meaning in our codebase as they represent the reactor communication channels. So I'd lean towards `hupc` and try to use it consistently in the codebase. Attempted that with the recent Transport addition in `p2p`.

libs/autofile/autofile.go
```diff
@@ -32,50 +34,70 @@ if err != nil {
...
-	Path          string
-	ticker        *time.Ticker
-	tickerStopped chan struct{} // closed when ticker is stopped
-	mtx           sync.Mutex
-	file          *os.File
+	ID   string
+	Path string
+
+	closeTicker       *time.Ticker
+	closeTickerStopCh chan struct{} // closed when closeTicker is stopped
```
xla: Probably best to rename this one as well.

PR #2485, opened 2018-09-25
config: Add ValidateBasic

melekes:
**DO NOT SQUASH BEFORE MERGING**

* [ ] Wrote tests


CHANGELOG_PENDING.md
```diff
@@ -10,12 +10,14 @@ BREAKING CHANGES:
 
 * Go API
 - [node] Remove node.RunForever
+- [config] \#2232 timeouts as time.Duration, not ints
```
ebuchman: Also a breaking change to the `CLI/RPC/Config` section since this now means the timeout params in the config.toml are strings like `"100ms"` or `"3s"`

Issue #2457, opened 2018-09-21
fastsync may stuck by a wrong block


goolAdapter:
When VerifyCommit failed, the logic only redo the first block request,  but the wrong block may was the second one. In this situation, didProcess will always empty loop in the next moment .
https://github.com/tendermint/tendermint/blob/d419fffe18531317c28c29a292ad7d253f6cafdf/blockchain/reactor.go#L300-L312

ebuchman:
Merged!

PR #2489, opened 2018-09-26
Fix #2456 and #2457 more clarify

### LINKED ISSUE #2456: fastsync's peerTimeout is too small
Block's default maxSize is 21M
https://github.com/tendermint/tendermint/blob/d419fffe18531317c28c29a292ad7d253f6cafdf/types/params.go#L55-L61

but in fastsync mode, peerTimeout only 15 Second, this need network speed > 1M/s， which is strict in product circumstance, which may lead fastsync to failed.
https://github.com/tendermint/tendermint/blob/d419fffe18531317c28c29a292ad7d253f6cafdf/blockchain/pool.go#L49

I suggest peerTimeout should simply make longer，or adjust dynamically with current fetching block‘s size.

goolAdapter:
Last PR I pushed all modify in one commit，make it hard to review，and since my origin develop is currently outdated, this time I update the latest develop branch，separate it to two step，It looks more clarify.

here are some main point:
1、Add a new protocol bcBlockResponseHeaderMessage，so that we can calculate timeout dynamically.
2、Remove bpPeer's recvMonitor, because it was not useable for large block, timeout is enough.
3、Add a peerID to redoCh chan, because there is a time window, cause redo multitime, and result in sync state error.
4、Rewrite some test case, more realistic.

I'll close the previous PR.

### DIFF
--- a/blockchain/pool.go
+++ b/blockchain/pool.go
@@ -3,13 +3,11 @@ package blockchain
 import (
 	"errors"
 	"fmt"
-	"math"
 	"sync"
 	"sync/atomic"
 	"time"
 
 	cmn "github.com/tendermint/tendermint/libs/common"
-	flow "github.com/tendermint/tendermint/libs/flowrate"
 	"github.com/tendermint/tendermint/libs/log"
 
 	"github.com/tendermint/tendermint/p2p"
@@ -46,7 +44,7 @@ const (
 	maxDiffBetweenCurrentAndReceivedBlockHeight = 100
 )
 
-var peerTimeout = 15 * time.Second // not const so we can override with tests
+var minimumTimeout = 15 * time.Second // fetch header timeout and minimum content timeout, not const so we can override with tests
 
 /*
 	Peers self report their heights when we join the block pool.
@@ -131,19 +129,6 @@ func (pool *BlockPool) removeTimedoutPeers() {
 	defer pool.mtx.Unlock()
 
 	for _, peer := range pool.peers {
-		if !peer.didTimeout && peer.numPending > 0 {
-			curRate := peer.recvMonitor.Status().CurRate
-			// curRate can be 0 on start
-			if curRate != 0 && curRate < minRecvRate {
-				err := errors.New("peer is not sending us data fast enough")
-				pool.sendError(err, peer.id)
-				pool.Logger.Error("SendTimeout", "peer", peer.id,
-					"reason", err,
-					"curRate", fmt.Sprintf("%d KB/s", curRate/1024),
-					"minRate", fmt.Sprintf("%d KB/s", minRecvRate/1024))
-				peer.didTimeout = true
-			}
-		}
 		if peer.didTimeout {
 			pool.removePeer(peer.id)
 		}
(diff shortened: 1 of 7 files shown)


goolAdapter:
@xla could you help to add some label and specific reviewer again? I'm sorry to trouble with you, thank you.

ebuchman:
Also maybe we want to separate the fixes for #2456 and #2457 into separate PRs? Seems #2457 may be a bit easier to deal with first.

blockchain/pool.go
```diff
@@ -565,9 +570,13 @@ OUTER_LOOP:
...
 				return
-			case <-bpr.redoCh:
-				bpr.reset()
-				continue OUTER_LOOP
+			case peerID := <-bpr.redoCh:
+				if peerID == bpr.peerID {
+					bpr.reset()
+					continue OUTER_LOOP
+				} else {
+					continue WAIT_LOOP
```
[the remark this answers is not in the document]
goolAdapter: it has a time window，when we process a new peer，an old peer reset come.

blockchain/reactor.go
```diff
@@ -258,8 +258,10 @@ FOR_LOOP:
 				bcR.Logger.Info("Time to switch to consensus reactor!", "height", height)
 				bcR.pool.Stop()
 
-				conR := bcR.Switch.Reactor("CONSENSUS").(consensusReactor)
-				conR.SwitchToConsensus(state, blocksSynced)
+				conR, ok := bcR.Switch.Reactor("CONSENSUS").(consensusReactor)
+				if ok {
```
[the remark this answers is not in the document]
goolAdapter: yes，just for testing

goolAdapter:
yes，#2457 is easy to fix，but hard to write test case. this PR is mostly to modify test case.

blockchain/reactor.go
```diff
@@ -308,6 +314,13 @@ FOR_LOOP:
 					// still need to clean up the rest.
 					bcR.Switch.StopPeerForError(peer, fmt.Errorf("BlockchainReactor validation error: %v", err))
 				}
+				peerID2 := bcR.pool.RedoRequest(second.Height)
+				peer2 := bcR.Switch.Peers().Get(peerID2)
+				if peer2 != nil && peer2 != peer {
+					// NOTE: we've already removed the peer's request, but we
+					// still need to clean up the rest.
+					bcR.Switch.StopPeerForError(peer2, fmt.Errorf("BlockchainReactor validation error: %v", err))
```
milosevic: In case VerifyCommit fails, it seems that we don't know which block is wrong. Does it make sense stopping both peers for error since we are not sure which one is wrong?

ebuchman:
@goolAdapter can we get this split into two separate PRs please? It's a bit difficult to review it all at once since there's lots of changes and we're addressing two distinct issues. #2457 is definitely the higher priority (and seemingly easier) one that we should treat first. Also seems like it's already relatively isolated in the first commit. Perhaps we can just review and merge that first commit, and then separate the rest into a new PR to solve #2456?

goolAdapter:
ok，I'll do it tomorrow

blockchain/reactor.go
goolAdapter: currently stopping both peers is ok, then it'll fetch block from other peers.

goolAdapter: I have open a new issue to think about this #2622

PR #2493, opened 2018-09-26
consensus: Wait timeout precommit before starting new round

milosevic:
Pull in ensureVote test function from https://github.com/tendermint/tendermint/pull/2132

* [x] Wrote tests

Fix https://github.com/tendermint/tendermint/issues/1690.


consensus/state.go
```diff
@@ -1642,22 +1642,19 @@ func (cs *ConsensusState) addVote(vote *types.Vote, peerID p2p.ID) (added bool,
...
+		if ok && len(blockID.Hash) != 0 {
+			// Executed as TwoThirdsMajority could be from a higher round
+			cs.enterNewRound(height, vote.Round)
+			cs.enterPrecommit(height, vote.Round)
+			cs.enterCommit(height, vote.Round)
 
+			if cs.config.SkipTimeoutCommit && precommits.HasAll() {
+				// if we have all the votes now,
+				// go straight to new round (skip timeout commit)
+				// cs.scheduleTimeout(time.Duration(0), cs.Height, 0, cstypes.RoundStepNewHeight)
```
xla: Do we need to keep this commented code line around?

consensus/state.go
```diff
@@ -1642,22 +1642,19 @@ func (cs *ConsensusState) addVote(vote *types.Vote, peerID p2p.ID) (added bool,
...
+				cs.enterNewRound(cs.Height, 0)
 			}
+
```
xla: Insignificant newline.

consensus/common_test.go
```diff
@@ -326,6 +327,30 @@ func ensureNewStep(stepCh <-chan interface{}) {
...
+		break
+	case v := <-voteCh:
+		edv, ok := v.(types.EventDataVote)
+		if !ok {
+			panic(fmt.Sprintf("expected a *types.Vote, got %v. wrong subscription channel?",
+				reflect.TypeOf(v)))
+		}
+		vote := edv.Vote
+		if vote.Height != height {
+			panic(fmt.Sprintf("expected height %v, got %v", height, vote.Height))
```
xla: Instead of panics this function should return errors, which makes it easier to report back in the tests.

consensus/state_test.go
```diff
@@ -936,6 +942,26 @@ func TestStateLockPOLSafety2(t *testing.T) {
...
+	newRoundCh := subscribe(cs1.eventBus, types.EventQueryNewRound)
+
+	// start round
+	startTestRound(cs1, cs1.Height, 0)
+	<-newRoundCh
+
+	signAddVotes(cs1, types.VoteTypePrecommit, nil, types.PartSetHeader{}, vs2, vs3, vs4)
+
+	<-timeoutWaitCh
+	<-newRoundCh
```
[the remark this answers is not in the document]
milosevic: Good point. I will use the same pattern as in ensureX methods to read from round and timeout channels.

consensus/common_test.go
```diff
@@ -307,46 +307,95 @@ func randConsensusState(nValidators int) (*ConsensusState, []*validatorStub) {
 
 //-------------------------------------------------------------------------------
 
-func ensureNoNewStep(stepCh <-chan interface{}) {
-	timer := time.NewTimer(ensureTimeout)
+func ensureNoNewEvent(t *testing.T, ch <-chan interface{}, timeout time.Duration,
+	errorMessage string) {
+	timer := time.NewTimer(timeout)
 	select {
 	case <-timer.C:
```
melekes: `<-time.After(timeout)`

consensus/common_test.go
```diff
@@ -307,46 +307,95 @@ func randConsensusState(nValidators int) (*ConsensusState, []*validatorStub) {
...
+func ensureNewUnlock(t *testing.T, unlockCh <-chan interface{}) {
+	ensureNewEvent(t, unlockCh, ensureTimeout,
+		"Timeout expired while waiting for NewUnlock event")
+}
+
+func ensureVote(t *testing.T, voteCh chan interface{}, height int64, round int,
+	voteType byte) {
 	timer := time.NewTimer(ensureTimeout)
 	select {
 	case <-timer.C:
```
melekes: `<-time.After(ensureTimeout)`

consensus/common_test.go
```diff
@@ -307,46 +307,95 @@ func randConsensusState(nValidators int) (*ConsensusState, []*validatorStub) {
...
 		break
 	case v := <-voteCh:
 		edv, ok := v.(types.EventDataVote)
 		if !ok {
-			panic(fmt.Sprintf("expected a *types.Vote, got %v. wrong subscription channel?",
+			t.Error(fmt.Sprintf("expected a *types.Vote, got %v. wrong subscription channel?",
```
melekes: `t.Errorf(...)`

consensus/state.go
ebuchman: Don't think so

consensus/state.go
```diff
@@ -1642,21 +1642,17 @@ func (cs *ConsensusState) addVote(vote *types.Vote, peerID p2p.ID) (added bool,
...
-				cs.enterPrecommit(height, vote.Round)
-				cs.enterCommit(height, vote.Round)
-
-				if cs.config.SkipTimeoutCommit && precommits.HasAll() {
-					// if we have all the votes now,
-					// go straight to new round (skip timeout commit)
-					// cs.scheduleTimeout(time.Duration(0), cs.Height, 0, cstypes.RoundStepNewHeight)
-					cs.enterNewRound(cs.Height, 0)
-				}
+		if ok && len(blockID.Hash) != 0 {
```
ebuchman: If `len(blockID.Hash) == 0`, don't we need to explicitly `enterPrecommitWait`?

consensus/common_test.go
```diff
@@ -306,26 +307,97 @@ func randConsensusState(nValidators int) (*ConsensusState, []*validatorStub) {
...
+func ensureNoNewEvent(t *testing.T, ch <-chan interface{}, timeout time.Duration,
+	errorMessage string) {
 	select {
-	case <-timer.C:
+	case <-time.After(timeout):
 		break
-	case <-stepCh:
-		panic("We should be stuck waiting, not moving to the next step")
+	case <-ch:
+		t.Error(errorMessage)
```
ebuchman: We probably want these to be panics for now until we build something more intelligent, otherwise we won't know which line in the actual test function caused the failure.

xla: Or alternatively to the error check in the actual test and `t.Fatalf` there. I know some may be concerned about readability but it's a test after all and the execution should be clear.

milosevic: For now, I would also say to leave panic. We will improve it after.

consensus/state.go
milosevic: We should always enterPrecommitWait upon receiving 2f+1 any precommits. We don't need to treat 2f+1 Precommit nil as a special case. We do it here: https://github.com/tendermint/tendermint/blob/bb8411b600eb71489baf008a646f31db30fcd642/consensus/state.go#L1657-L1660. Only in case we have 2f+1 precommits for some block we go straight to commit and we don't need to trigger timeout.

ebuchman: Ok right, it's taken care of by the `else if`. The first condition is only for `+2/3 majority for a block hash`. If there's a +2/3 majority for nil, it will be treated by the second condition. Thanks.

Issue #2491, opened 2018-09-26
Separate IndexerService from Tendermint core


goolAdapter:
The IndexerService is IO intensive，and easy to suffer from DDOS(TxIndex.Search may scan full db). It's content was duplicately stored by blockstore and stateDB，which blockstore have Block for Tx and stateDB have ABCIResponses for ResponseDeliverTx.

So it's safe to remove the index data.  It's better to separate IndexerService to a independent service? this may improve performance.

As a result,  TxSearch not support by the core, IndexerService can implement with mysql.

melekes:
Hi! Thank you for raising this issue. We're aware of extensive IO and have plans to

1) improve existing indexer by changing the underlying DB structure https://github.com/tendermint/tendermint/issues/1835#issuecomment-421062709
2) support external indexer https://github.com/tendermint/tendermint/issues/1161

I've created a separate issue for KV indexer to avoid doing full scans https://github.com/tendermint/tendermint/issues/2492

ebuchman:
> stateDB have ABCIResponses for ResponseDeliverTx.

Note this is only true for the latest block to support recovery from crash failure. The state does not store abci responses for more than one block (so its not duplicating whats stored in the indexer). The indexer is the only place we store the transaction results for all past blocks.

goolAdapter:
@ebuchman thanks，but ApplyBlock save ABCIResponses at each height, was there something I misunderstand? the statedb will prune dynamically?
https://github.com/tendermint/tendermint/blob/0c9c3292c918617624f6f3fbcd95eceade18bcd5/state/execution.go#L66-L87

ebuchman:
Whoops looks like I was wrong about that! Thanks for noticing. We should definitely try to do some deduplication there!

PR #2499, opened 2018-09-27
state: Store ABCIResponses only for the last block

melekes:
Refs https://github.com/tendermint/tendermint/issues/2491#issuecomment-424927721

* [x] Wrote tests
* [ ] Updated UPGRADING.md if necessary


state/execution.go
```diff
@@ -84,6 +84,11 @@ func (blockExec *BlockExecutor) ApplyBlock(state State, blockID types.BlockID, b
 	// Save the results before we commit.
 	saveABCIResponses(blockExec.db, block.Height, abciResponses)
 
+	// Delete the results for the previous height (not needed anymore).
+	if block.Height > 0 {
+		deleteABCIResponses(blockExec.db, block.Height-1)
+	}
```
goolAdapter: Is better to change calcABCIResponsesKey return a constant string? as the State's key stateKey:  
stateKey = []byte("stateKey")

this may reduce some IO.

melekes: Good idea

rpc/core/blocks.go
```diff
@@ -320,6 +325,10 @@ func Commit(heightPtr *int64) (*ctypes.ResultCommit, error) {
 // BlockResults gets ABCIResults at a given height.
 // If no height is provided, it will fetch results for the latest block.
 //
+// Results for the latest height are fetched from the state database, for
+// earlier heights - using the indexer. Note: indexer does not store EndBlock
+// ("end_block": null).
+//
```
melekes: not sure what to do with EndBlock

state/execution_test.go
```diff
@@ -43,7 +43,14 @@ func TestApplyBlock(t *testing.T) {
 	blockID := types.BlockID{block.Hash(), block.MakePartSet(testPartSize).Header()}
 
 	state, err = blockExec.ApplyBlock(state, blockID, block)
-	require.Nil(t, err)
+	require.NoError(t, err)
+
+	// Test we store results for the current height and do not store results for
```
melekes: stale comment

rpc/core/blocks.go
ebuchman: Can we index the EndBlock tags?

rpc/core/blocks.go
```diff
@@ -357,17 +366,63 @@ func BlockResults(heightPtr *int64) (*ctypes.ResultBlockResults, error) {
 		return nil, err
 	}
 
-	// load the results
-	results, err := sm.LoadABCIResponses(stateDB, height)
-	if err != nil {
-		return nil, err
+	// for last height, use state DB
```
ebuchman: Why bother with the special case? Can't we just get everything through the indexer?

melekes: But if the indexer is disabled, you won't get anything. Now, even though it's off, you can still view the results for the last height.

rpc/core/blocks.go
melekes: Mmm.. We'll need a new event, but yes, certainly possible

melekes: Wait.. why do we need tags in the endblock at all? 

we save params updates and validator updates separately and we also have events for them. so everyone can subscribe to them OR use RPC to fetch them at specific height.

ebuchman: We may want to index certain events happening during an EndBlock - like if a validator gets slashed, and we want to query for all places where validator X got slashed

ebuchman:
@melekes Does this mean that if indexing is off we can't look up results?

melekes:
> Does this mean that if indexing is off we can't look up results?

for current height: you can, for past heights: you can't without indexing

melekes:
> Do we think that's ok? Not totally sure what to do here - realizing now that's why the results were being stored in the state db.

we store validator updates and cs params updates separately, so the only thing here is **endblock tags**. These should presumably have the same semantics as DeliverTx/CheckTx tags - we do not store them, we index them.

ValarDragon:
> @melekes Does this mean that if indexing is off we can't look up results?

Its not clear to me who the set of nodes are that want results, but not indexing. Validators shouldn't do either on their core validator or sentries. They may have debug / archival instances, but both should take the indexing overhead. The point of indexing is that you get a more queryable storage at the cost of some space AFAIU. The only efficient mode you have to look up via results AFAIK is if you know what block the tx you care about it is in.

Maybe we want nodes to be able to store the last `n` blocks of results just to enable automation scripts though. (`n` perhaps being configurable?)

xla:
@melekes Checking in on this one, still relevant?

melekes:
yes

rpc/core/blocks.go
```diff
@@ -380,17 +389,63 @@ func BlockResults(ctx *rpctypes.Context, heightPtr *int64) (*ctypes.ResultBlockR
...
+		return &ctypes.ResultBlockResults{
+			Height:  height,
+			Results: results,
+		}, nil
+	}
+
+	// for other heights, use indexer
+	// if index is disabled, return error
+	if _, ok := txIndexer.(*null.TxIndex); ok {
+		return nil, fmt.Errorf("Transaction indexing is disabled")
```
Liamsi:
```suggestion
		return nil, fmt.Errorf("transaction indexing is disabled")
```

rpc/core/blocks.go
```diff
@@ -380,17 +389,63 @@ func BlockResults(ctx *rpctypes.Context, heightPtr *int64) (*ctypes.ResultBlockR
...
-		Results: results,
+	var deliverTxs []*abci.ResponseDeliverTx
+
+	// If Tendermint is configured to index txs by height, use it for faster search.
+	if txIndexer.IsTagIndexed(types.TxHeightKey) {
+		results, err := txIndexer.Search(
+			tmquery.MustParse(fmt.Sprintf("%s = %d", types.TxHeightKey, height)),
+		)
+		if err != nil {
+			return nil, errors.Wrap(err, "Failed to search indexer for txs")
```
Liamsi:
```suggestion
			return nil, errors.Wrap(err, "failed to search indexer for txs")
```

rpc/core/blocks.go
```diff
@@ -380,17 +389,63 @@ func BlockResults(ctx *rpctypes.Context, heightPtr *int64) (*ctypes.ResultBlockR
...
+		}
+	} else { // otherwise, load the whole block and fetch txs one by one
+		block := blockStore.LoadBlock(height)
+		if block != nil {
+			deliverTxs = make([]*abci.ResponseDeliverTx, len(block.Data.Txs))
+
+			for i, tx := range block.Data.Txs {
+				r, err := txIndexer.Get(tx.Hash())
+				if err != nil {
+					return nil, errors.Wrapf(err, "Failed to get Tx (%X) from indexer", tx.Hash())
```
Liamsi:
```suggestion
					return nil, errors.Wrapf(err, "failed to get Tx (%X) from indexer", tx.Hash())
```

state/errors.go
```diff
@@ -39,10 +42,10 @@ type (
 	ErrNoConsensusParamsForHeight struct {
 		Height int64
 	}
+)
 
-	ErrNoABCIResponsesForHeight struct {
-		Height int64
-	}
+var (
+	ErrNoABCIResponses = errors.New("Could not find results")
```
Liamsi:
```suggestion
	ErrNoABCIResponses = errors.New("could not find results")
```

state/store.go
```diff
@@ -9,6 +9,10 @@ import (
 	"github.com/tendermint/tendermint/types"
 )
 
+var (
+	abciResponsesKey = []byte("abciResponsesKey")
```
Liamsi: Is this breaking as the key in the state store changed?

state/execution_test.go
```diff
@@ -45,7 +45,13 @@ func TestApplyBlock(t *testing.T) {
...
 	//nolint:ineffassign
 	state, err = blockExec.ApplyBlock(state, blockID, block)
-	require.Nil(t, err)
+	require.NoError(t, err)
+
+	// test we save results
+	abciResponses, err := LoadABCIResponses(stateDB)
+	if assert.NoError(t, err) {
+		assert.NotNil(t, abciResponses)
+	}
```
Liamsi: 👍

state/store.go
```diff
@@ -137,10 +137,10 @@ func (arz *ABCIResponses) ResultsHash() []byte {
 // LoadABCIResponses loads the ABCIResponses for the given height from the database.
 // This is useful for recovering from crashes where we called app.Commit and before we called
 // s.Save(). It can also be used to produce Merkle proofs of the result of txs.
-func LoadABCIResponses(db dbm.DB, height int64) (*ABCIResponses, error) {
-	buf := db.Get(calcABCIResponsesKey(height))
+func LoadABCIResponses(db dbm.DB) (*ABCIResponses, error) {
```
Liamsi: The comment above is out dated now: "LoadABCIResponses loads the ABCIResponses for the given height"

ebuchman:
Closing for https://github.com/tendermint/tendermint/issues/3533

ebuchman:
We can keep the branch around for now.

PR #2505, opened 2018-09-28
remove some xxx comments and the config.mempool.recheck_empty

ebuchman:
Remove/update a few XXX/TODO found in the code that are stale or can be quickly addressed.

Also realized we can just remove Config.Mempool.RecheckEmpty. It defaults to `true` and was a useless option.

* [ ] Wrote tests


config/toml.go
```diff
@@ -206,7 +206,6 @@ private_peer_ids = "{{ .P2P.PrivatePeerIDs }}"
 [mempool]
 
 recheck = {{ .Mempool.Recheck }}
-recheck_empty = {{ .Mempool.RecheckEmpty }}
```
melekes: also needs to be removed from docs/tendermint-core/configuration/md

privval/priv_validator.go
```diff
@@ -38,14 +38,16 @@ func voteToStep(vote *types.Vote) int8 {
...
+// It includes the LastSignature and LastSignBytes so we don't lose the signature
+// if the process crashes after signing but before the resulting consensus message is processed.
 type FilePV struct {
 	Address       types.Address  `json:"address"`
 	PubKey        crypto.PubKey  `json:"pub_key"`
 	LastHeight    int64          `json:"last_height"`
 	LastRound     int            `json:"last_round"`
 	LastStep      int8           `json:"last_step"`
-	LastSignature []byte         `json:"last_signature,omitempty"` // so we dont lose signatures XXX Why would we lose signatures?
-	LastSignBytes cmn.HexBytes   `json:"last_signbytes,omitempty"` // so we dont lose signatures XXX Why would we lose signatures?
```
ValarDragon: Perhaps change these lines to "so we maintain signatures in event of a crash"

ValarDragon: Oh I missed it, my bad

Issue #2509, opened 2018-09-29
General Merkle Follow Up


ebuchman:
Still needs some follow up from reviews  https://github.com/tendermint/tendermint/pull/2298#pullrequestreview-159125295 and 
https://github.com/tendermint/tendermint/pull/2298#pullrequestreview-154218462. Namely:

- [ ] Address the XXX/TODO in that PR
- [x] Fix tmlibs->libs
- [ ] Add an ADR 
- [x] Update ABCI and other docs
- [x] Update changelog
- [x] Update code comments as noted in PR review
- [ ] Address other questions/comments in the PR review
- [ ] Address [Anton's review](https://github.com/tendermint/tendermint/pull/2298#pullrequestreview-160193807)

Also see the request for clarifications in https://github.com/tendermint/tendermint/issues/2527

PR #2513, opened 2018-09-29
Remove tmlibs import which just got re-introduced

ValarDragon:


ValarDragon:
closing in favor of #2509

PR #2516, opened 2018-09-29
fix evidence db iter leak

### LINKED ISSUE #2515: Evidence store may leak resource
there is a leak, iter need defer iter.Close().
https://github.com/tendermint/tendermint/blob/0c9c3292c918617624f6f3fbcd95eceade18bcd5/evidence/store.go#L100-L101

by the way, https://github.com/tendermint/tendermint/blob/0c9c3292c918617624f6f3fbcd95eceade18bcd5/evidence/store.go#L82-L85
should replace with reversing trick to eliminate temp slice.
```
for i := len(a)/2-1; i >= 0; i-- {
	opp := len(a)-1-i
	a[i], a[opp] = a[opp], a[i]
}
```
I'll make a PR for this issue.

goolAdapter:
fix #2515 
* [ ] Wrote tests

### DIFF
--- a/evidence/store.go
+++ b/evidence/store.go
@@ -98,6 +98,7 @@ func (store *EvidenceStore) PendingEvidence(maxBytes int64) (evidence []types.Ev
 func (store *EvidenceStore) listEvidence(prefixKey string, maxBytes int64) (evidence []types.Evidence) {
 	var bytes int64
 	iter := dbm.IteratePrefix(store.db, []byte(prefixKey))
+	defer iter.Close()
 	for ; iter.Valid(); iter.Next() {
 		val := iter.Value()
 
--- a/evidence/store.go
+++ b/evidence/store.go
@@ -79,9 +79,8 @@ func NewEvidenceStore(db dbm.DB) *EvidenceStore {
 func (store *EvidenceStore) PriorityEvidence() (evidence []types.Evidence) {
 	// reverse the order so highest priority is first
 	l := store.listEvidence(baseKeyOutqueue, -1)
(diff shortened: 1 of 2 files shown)


evidence/store.go
```diff
@@ -79,11 +79,12 @@ func NewEvidenceStore(db dbm.DB) *EvidenceStore {
 func (store *EvidenceStore) PriorityEvidence() (evidence []types.Evidence) {
 	// reverse the order so highest priority is first
 	l := store.listEvidence(baseKeyOutqueue, -1)
-	l2 := make([]types.Evidence, len(l))
-	for i := range l {
-		l2[i] = l[len(l)-1-i]
+	for i := len(l)/2 - 1; i >= 0; i-- {
```
melekes:
```go
i, j := 0, len(l)-1; i < j; i, j = i+1, j-1 {
  l[i], l[j] = l[j], l[i]
}
```

How about ^^^?

evidence/store.go
```diff
@@ -98,6 +99,7 @@ func (store *EvidenceStore) PendingEvidence(maxBytes int64) (evidence []types.Ev
 func (store *EvidenceStore) listEvidence(prefixKey string, maxBytes int64) (evidence []types.Evidence) {
 	var bytes int64
 	iter := dbm.IteratePrefix(store.db, []byte(prefixKey))
+	defer iter.Close()
```
melekes: Whoops. How did we miss this? Thanks for spotting.

evidence/store.go
goolAdapter: I think both  is ok，respect your prefrence.

Issue #2518, opened 2018-09-29
lib's event implement have some flaw 


goolAdapter:
**This issue doesn't exist in the current runtime, but as an base lib, I think it's better to more robust.**

Lib event designed for multi routine, but AddListenerForEvent and RemoveListener manipulate listener and eventCell in different order.
https://github.com/tendermint/tendermint/blob/0c9c3292c918617624f6f3fbcd95eceade18bcd5/libs/events/events.go#L76-L78
https://github.com/tendermint/tendermint/blob/0c9c3292c918617624f6f3fbcd95eceade18bcd5/libs/events/events.go#L94-L98

consider two routine，A call AddListenerForEvent and B call RemoveListener  simultaneous.
```
B -> listener.SetRemoved()
A -> eventCell.AddListener(listenerID, cb)
A -> listener.AddEvent(event)
B -> for _, event := range listener.GetEvents() {
		evsw.RemoveListenerForEvent(event, listenerID)
}
```
At this time, Listener added by A will not be clean up. the fix is to try AddEvent first, only when AddEvent succeeded, eventCell can AddListener.

by the way，eventCell.FireEvent is easy to deadlock, as lock's range is uncontrollable when run callback.
https://github.com/tendermint/tendermint/blob/0c9c3292c918617624f6f3fbcd95eceade18bcd5/libs/events/events.go#L169-L175

melekes:
note: we'll get rid of the `events` package in https://github.com/tendermint/tendermint/issues/1692

xla:
Merged to develop.

PR #2519, opened 2018-09-29
fix event concurrency flaw

goolAdapter:
ref #2518
* [x] Wrote tests


libs/events/events_test.go
```diff
@@ -162,6 +166,52 @@ func TestAddDifferentListenerForDifferentEvents(t *testing.T) {
 	}
 }
 
+var stopInputEvent = false
```
melekes: can we not introduce global variables?

libs/events/events.go
```diff
@@ -194,14 +203,15 @@ func newEventListener(id string) *eventListener {
...
 }
 
-func (evl *eventListener) AddEvent(event string) {
+func (evl *eventListener) AddEvent(event string) error {
 	evl.mtx.Lock()
 	defer evl.mtx.Unlock()
 
 	if evl.removed {
-		return
+		return fmt.Errorf("listener was removed.")
```
melekes: `var ErrListenerRemoved = errors.New("listener was removed")`

libs/events/events.go
```diff
@@ -168,10 +172,15 @@ func (cell *eventCell) RemoveListener(listenerID string) int {
 
 func (cell *eventCell) FireEvent(data EventData) {
 	cell.mtx.RLock()
+	var listenerCopy []EventCallback
```
melekes: why copy?

melekes: although, copy should be fine too

libs/events/events_test.go
goolAdapter: I'm sorry for dalayed reply，I'm taking a holiday, so the next relay may also delay.
Yes，this global variables  will eliminated.

libs/events/events.go
goolAdapter: that is ok， I'll replace other error message with named error too.

libs/events/events.go
goolAdapter: yes， this previously can exist a deadlock， when we call add/remove event in callback. but it can work under carefully control，copy is a way to discard the burden.

I prefer the copy, it free the event user.

libs/events/events.go
```diff
@@ -58,7 +59,7 @@ func (evsw *eventSwitch) OnStart() error {
 
 func (evsw *eventSwitch) OnStop() {}
 
-func (evsw *eventSwitch) AddListenerForEvent(listenerID, event string, cb EventCallback) {
+func (evsw *eventSwitch) AddListenerForEvent(listenerID, event string, cb EventCallback) (err error) {
```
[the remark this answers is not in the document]
goolAdapter: no strong reason, I'll change to the repository's convention.

libs/events/events.go
```diff
@@ -10,10 +10,17 @@ import (
 	cmn "github.com/tendermint/tendermint/libs/common"
 )
 
+type ErrListenerWasRemoved struct {
+	listener string
+}
+
+func (e ErrListenerWasRemoved) Error() string {
+	return fmt.Sprintf("listener %s was removed.", e.listener)
```
melekes: no dot (`.`) at the end please

Use https://github.com/golang/lint

libs/events/events.go
```diff
@@ -208,8 +216,9 @@ func (evl *eventListener) AddEvent(event string) error {
 	defer evl.mtx.Unlock()
 
 	if evl.removed {
-		return fmt.Errorf("listener was removed.")
+		return &ErrListenerWasRemoved{listener: evl.id}
```
melekes: No need for error to be a pointer (`&`)

libs/events/events_test.go
```diff
@@ -166,29 +166,29 @@ func TestAddDifferentListenerForDifferentEvents(t *testing.T) {
...
-func inputRemoveListener(t *testing.T, evsw EventSwitch, done chan uint64) {
+func inputRemoveListener(t *testing.T, evsw EventSwitch, done chan uint64, roundCount int) {
 	for i := 0; i < roundCount; i++ {
 		evsw.RemoveListener("listener")
 	}
 	done <- 1
 }
 
-func inputAddListenerForEvent(t *testing.T, evsw EventSwitch, done chan uint64) {
+func inputAddListenerForEvent(t *testing.T, evsw EventSwitch, done chan uint64, roundCount int, stopInputEvent *bool) {
```
melekes: can we use a channel instead of `*bool` for stopInputEvent

```go
func inputAddListenerForEvent(t *testing.T, evsw EventSwitch, done chan uint64, roundCount int, stopInputEvent chan<- bool) {
....
stopInputEvent <- true
```

goolAdapter: chan is ok， but bool flag looks more simple，it's only used for reduce fail log.

libs/events/events_test.go
```diff
@@ -162,6 +166,52 @@ func TestAddDifferentListenerForDifferentEvents(t *testing.T) {
 	}
 }
 
+func inputRemoveListener(t *testing.T, evsw EventSwitch, done chan uint64, roundCount int) {
```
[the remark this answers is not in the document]
goolAdapter: yes，done chan uint64 is inspired by other test case，here ```done chan struct{} ```is enough

libs/events/events_test.go
```diff
@@ -162,6 +166,52 @@ func TestAddDifferentListenerForDifferentEvents(t *testing.T) {
...
+	for i := 0; i < roundCount; i++ {
+		evsw.RemoveListener("listener")
+	}
+	done <- 1
+}
+
+func inputAddListenerForEvent(t *testing.T, evsw EventSwitch, done chan uint64, roundCount int, stopInputEvent *bool) {
+	for i := 0; i < roundCount; i++ {
+		index := i
```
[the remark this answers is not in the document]
goolAdapter: it need, the next event callback use index for log.

libs/events/events_test.go
```diff
@@ -162,6 +166,52 @@ func TestAddDifferentListenerForDifferentEvents(t *testing.T) {
...
+
+	go inputRemoveListener(t, evsw, done1, roundCount)
+	go inputAddListenerForEvent(t, evsw, done2, roundCount, &stopInputEvent)
+
+	<-done1
+	<-done2
+
+	evsw.RemoveListener("listener") // make sure remove last
+
+	for i := 0; i < roundCount && !stopInputEvent; i++ {
```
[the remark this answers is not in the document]
goolAdapter: yes, this test is used to uncovers the DATA race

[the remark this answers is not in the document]
goolAdapter: I'm back from a long vacation，sorry for the delayed response，I will fix the review issue.

CHANGELOG_PENDING.md
```diff
@@ -42,4 +42,10 @@ IMPROVEMENTS:
 BUG FIXES:
 - [autofile] \#2428 Group.RotateFile need call Flush() before rename (@goolAdapter)
 - [node] \#2434 Make node respond to signal interrupts while sleeping for genesis time
+- [consensus] [\#1690](https://github.com/tendermint/tendermint/issues/1690) wait for
```
melekes: where are these coming from? improper merge?

libs/events/events_test.go
```diff
@@ -196,11 +179,27 @@ func TestAddAndRemoveListenerConcurrency(t *testing.T) {
...
-	done1 := make(chan uint64)
-	done2 := make(chan uint64)
+	done1 := make(chan struct{})
+	done2 := make(chan struct{})
+
+	go func() {
+		for i := 0; i < roundCount; i++ {
+			evsw.RemoveListener("listener")
+		}
+		done1 <- struct{}{}
```
melekes: `close(done1)`

libs/events/events_test.go
```diff
@@ -196,11 +179,27 @@ func TestAddAndRemoveListenerConcurrency(t *testing.T) {
...
+	go func() {
+		for i := 0; i < roundCount; i++ {
+			index := i //it necessary for closure
+			evsw.AddListenerForEvent("listener", fmt.Sprintf("event%d", index),
+				func(data EventData) {
+					t.Errorf("should not run callback for %d.\n", index)
+					stopInputEvent = true
+				})
+		}
+		done2 <- struct{}{}
```
melekes: `close(done2)`

goolAdapter: here chan can recycled automatically by the garbage collector， but I'll close it explicit

libs/events/events_test.go
goolAdapter: here chan can recycled automatically by the garbage collector， but I'll close it explicit

libs/events/events.go
goolAdapter: I'll create a new test to show the deadlock.

PR #2524, opened 2018-09-30
Bucky/abci docs

ebuchman:
Document the General Merkle Proofs in ABCI spec as follow up from https://github.com/tendermint/tendermint/issues/2509

Also add some missing/extra ConsensusParams docs

### DIFF
--- a/docs/DOCS_README.md
+++ b/docs/DOCS_README.md
@@ -35,6 +35,8 @@ of the sidebar.
 **NOTE:** Strongly consider the existing links - both within this directory
 and to the website docs - when moving or deleting files.
 
+Links to directories *MUST* end in a `/`.
+
 Relative links should be used nearly everywhere, having discovered and weighed the following:
 
 ### Relative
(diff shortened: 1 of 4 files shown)


ebuchman:
@jaekwon @mossid not 100% sure where the KeyPath stuff interfaces with the user. Are we expecting that the KeyPath passed into RequestQuery.Path corresponds to the keys in the ResponseQuery.Proofs? In which case this KeyPath becomes the sort of standard for RequestQuery.Path and should also be documented here? 

Noting that it's a bit weird for the general Merkle proof types to live in `crypto/merkle`. They're effectively part of ABCI and should live in `abci/types`, just like the ABCI `PubKey` type...

We would also benefit from separation in `crypto/merkle` between the SimpleTree code and the GeneralMerkle code. Since we want most of the `crypto/merkle` to be used by external consumers, it really helps to have separation to simplify looking at the godoc.org.

ebuchman:
Please review, but merging for staging deployment. Opened my comment in https://github.com/tendermint/tendermint/issues/2527

PR #2533, opened 2018-10-02
bit_array: Simplify subtraction

ValarDragon:
This significantly simplifies subtraction logic. I suspect it also significantly speeds it up due to defer removal.

* [ ] Updated all relevant documentation in docs - n/a
* [x] Wrote tests
* [ ] Updated CHANGELOG_PENDING.md - seems like a fairly internal change


ValarDragon:
I've been doing too much algebra over finite field extensions lol. I got the subtraction confused with over there where it is an xor. It should actually be an `and not`. Changed that in the latest commit, tests should now pass.

PR #2534, opened 2018-10-03
Fix random distribution in bitArray.PickRandom

ValarDragon:
Previously it was very biased. 63 "_" followed by a single "x" had
much greater odds of being chosen. Additionally, the last element was
skewed. This fixes that by first preproccessing the set of all true
indices, and then randomly selecting a single element from there.

This commit also makes the code here significantly simpler, and
improves test cases.

* [x] Updated all relevant documentation in docs - n/a
* [x] Wrote tests

### DIFF
--- a/libs/common/bit_array.go
+++ b/libs/common/bit_array.go
@@ -234,49 +234,54 @@ func (bA *BitArray) IsFull() bool {
 	return (lastElem+1)&((uint64(1)<<uint(lastElemBits))-1) == 0
 }
 
-// PickRandom returns a random index in the bit array, and its value.
+// PickRandom returns a random index for a set bit in the bit array.
+// If there is no such value, it returns 0, false.
 // It uses the global randomness in `random.go` to get this index.
 func (bA *BitArray) PickRandom() (int, bool) {
 	if bA == nil {
 		return 0, false
 	}
 	bA.mtx.Lock()
-	defer bA.mtx.Unlock()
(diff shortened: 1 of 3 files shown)


libs/common/bit_array.go
```diff
@@ -238,49 +238,54 @@ func (bA *BitArray) IsFull() bool {
...
 	}
 	bA.mtx.Lock()
-	defer bA.mtx.Unlock()
 
-	length := len(bA.Elems)
-	if length == 0 {
+	trueIndices := bA.getTrueIndices()
+	// no bits set to true
+	if len(trueIndices) == 0 {
 		return 0, false
```
melekes: `bA.mtx.Unlock()` ?

ValarDragon: It's unlocked at the bottom

melekes: Huh? It won't be called if there's no `defer` because you `return 0, false` here

ValarDragon: Oh your right, good catch.

melekes: If I am not wrong, we can unlock mtx even sooner https://github.com/tendermint/tendermint/pull/2534/commits/5268d68dbe3ad6f2bdb856de25f0b8af88dadd02

ValarDragon: You're right!

Issue #2027, opened 2018-07-23
error occurred when using SocketPV


WeibingChen:
1. first, I start cmd/priv_val_server/main.go, it reports connection refused:

```
$ go run ./main.go --chain-id='1234' --priv="/Users/donnie/.travis/config/remote/priv_validator.json" --addr=":26659"

I[07-23|10:25:53.860] Starting private validator                   module=priv_val addr=:26659 chainID=1234 privPath=/Users/donnie/.travis/config/remote/priv_validator.json
I[07-23|10:25:53.863] Starting RemoteSigner                        module=priv_val impl=RemoteSigner
E[07-23|10:25:53.863] connect                                      module=priv_val addr=:26659 err="Error{dial tcp :26659: connect: connection refused}"
...
E[07-23|10:26:20.901] OnStart                                      module=priv_val err="Error{dialed maximum retries}"
panic: Error{dialed maximum retries}

goroutine 1 [running]:
main.main()
	$GOPATH/src/github.com/tendermint/tendermint/cmd/priv_val_server/main.go:44 +0x7d6
exit status 2
```

2. second, i change ~/.tendermint/config/config.toml and add `priv_validator_laddr = ":26669"`. Then I start tendermint with nilapp, I got error as below:
```
$ tendermint node start --home ~/.tendermint --proxy_app="nilapp"
...
I[07-23|10:25:57.428] Completed ABCI Handshake - Tendermint and App are synced module=consensus appHeight=0 appHash=
I[07-23|10:25:57.428] Starting SocketPV                            module=privval impl=SocketPV
E[07-23|10:25:58.431] OnStart                                      module=privval err="Error{waited for remote signer for too long}"
ERROR: Failed to create node: Error starting private validator client: Error{waited for remote signer for too long}

```

melekes:
Have you tried starting Tendermint first and only then starting `priv_val_server`?

melekes:
Also, I can't find any docs on running remote private validator.

WeibingChen:
@melekes yes, I've tried starting Tendermint first. It reported the same error.
I also found that error comes from   `conn, err := sc.listener.Accept()`， the base error is `accept tcp [::]:26669: i/o timeout`. I checked the acceptDeadline value that is `1.000000003s`, very strange. The value is defined  by `acceptDeadline = time.Second + defaultAcceptDeadlineSeconds`. I think there is a spell mistake, it should be '*' instead.

melekes:
[...]

PR #2548, opened 2018-10-05
privval: set deadline in readMsg

ebuchman:
This is a quick-fix for https://github.com/tendermint/tendermint/issues/2027.

The problem is that we set the deadline once, but need to reset each time we want to read, else the first deadline will expire at some random later read.

Also note this is using the default. It seems the ability to use the options has been eliminated? We probably want to bring that back, and actually use an internal, configured value for the read deadline rather than the default.

### DIFF
--- a/privval/socket.go
+++ b/privval/socket.go
@@ -585,7 +585,7 @@ func readMsg(r io.Reader) (msg SocketPVMsg, err error) {
 	// set deadline before trying to read
 	conn := r.(net.Conn)
 	if err := conn.SetDeadline(time.Now().Add(time.Second * defaultConnDeadlineSeconds)); err != nil {
-		err = cmn.ErrorWrap(err, "setting connection timeout failedin readMsg")
+		err = cmn.ErrorWrap(err, "setting connection timeout failed in readMsg")
 		return msg, err
 	}
 
--- a/privval/socket.go
+++ b/privval/socket.go
@@ -18,11 +18,11 @@ import (
 )
 
 const (
-	defaultAcceptDeadlineSeconds = 3
-	defaultConnDeadlineSeconds   = 3
-	defaultConnHeartBeatSeconds  = 30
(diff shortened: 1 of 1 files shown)


privval/socket.go
```diff
@@ -587,6 +581,14 @@ type RemoteSignerError struct {
 
 func readMsg(r io.Reader) (msg SocketPVMsg, err error) {
 	const maxSocketPVMsgSize = 1024 * 10
+
+	// set deadline before trying to read
+	conn := r.(net.Conn)
+	if err := conn.SetDeadline(time.Now().Add(time.Second * defaultConnDeadlineSeconds)); err != nil {
+		err = cmn.ErrorWrap(err, "setting connection timeout failedin readMsg")
```
melekes: `failed in`

privval/socket.go
```diff
@@ -18,11 +18,11 @@ import (
 )
 
 const (
-	defaultAcceptDeadlineSeconds = 3
-	defaultConnDeadlineSeconds   = 3
-	defaultConnHeartBeatSeconds  = 30
-	defaultConnWaitSeconds       = 60
-	defaultDialRetries           = 10
+	defaultAcceptDeadlineSeconds = 30 // tendermint waits this long for remote val to connect
+	defaultConnDeadlineSeconds   = 3  // must be set before each read
```
melekes: `ConnReadDeadline`?

ebuchman: Should we set it on writes too?

melekes: Oh, if we have writes, then please ignore my comment. I haven't seen one...

xla:
Looks like it's superseed by #2568.

ValarDragon:
Shouldn't we merge this small focused PR, and split up concerns in #2568?

ebuchman:
> Shouldn't we merge this small focused PR, and split up concerns in #2568?

Yes I think so.

PR #2556, opened 2018-10-05
p2p: NodeInfo is an interface

ebuchman:
Based on review from https://github.com/tendermint/tendermint/pull/2528#pullrequestreview-160688094, changed NodeInfo to an interface so we can start to disentangle concerns and reduce the impact of making changes to the NodeInfo

* [ ] Wrote tests


node/node.go
```diff
@@ -753,8 +753,8 @@ func makeNodeInfo(
 	if _, ok := txIndexer.(*null.TxIndex); ok {
 		txIndexerStatus = "off"
 	}
-	nodeInfo := p2p.NodeInfo{
-		ID:      nodeID,
+	nodeInfo := p2p.DefaultNodeInfo{
+		ID_:     nodeID,
```
ebuchman: Is there a better alternative? I really don't like `GetID()` for this.

p2p/node_info.go
```diff
@@ -17,12 +18,20 @@ func MaxNodeInfoSize() int {
 	return maxNodeInfoSize
 }
 
-// NodeInfo is the basic node information exchanged
+type NodeInfo interface {
+	ID() ID
+	Validate() error
```
ebuchman: ValidateBasic

p2p/transport_test.go
```diff
@@ -534,13 +514,13 @@ func TestTransportHandshake(t *testing.T) {
 		}
 
 		go func(c net.Conn) {
-			_, err := cdc.MarshalBinaryWriter(c, peerNodeInfo)
+			_, err := cdc.MarshalBinaryWriter(c, peerNodeInfo.(DefaultNodeInfo))
```
ebuchman: If we registered the NodeInfo interface with amino, we wouldn't have to do this, and we'd be able to adopt new NodeInfo types. But we probably don't want amino baked in there, so we'll need a better strategy for getting the concrete NodeInfo type than a type assertion in the depths of the code like this.

p2p/node_info.go
```diff
@@ -17,12 +18,20 @@ func MaxNodeInfoSize() int {
 	return maxNodeInfoSize
 }
 
-// NodeInfo is the basic node information exchanged
+type NodeInfo interface {
```
melekes: Shouldn't interface have a comment too?

p2p/test_util.go
```diff
@@ -198,9 +199,11 @@ func MakeSwitch(
 	sw.SetLogger(log.TestingLogger())
 	sw.SetNodeKey(&nodeKey)
 
+	ni_ := ni.(DefaultNodeInfo)
```
melekes: golang's linter does not like underscores `_`

node/node.go
melekes: I saw `Num` suffix somewhere in our codebase

p2p/transport_test.go
```diff
@@ -207,15 +207,10 @@ func TestTransportMultiplexAcceptNonBlocking(t *testing.T) {
...
-		fastNodeInfo = NodeInfo{
-			ID:         PubKeyToID(fastNodePV.PubKey()),
-			ListenAddr: "127.0.0.1:0",
-			Moniker:    "fastNode",
-			Version:    "1.0.0",
-		}
-		errc  = make(chan error)
-		fastc = make(chan struct{})
-		slowc = make(chan struct{})
+		fastNodeInfo = testNodeInfoFromIDAndName(PubKeyToID(fastNodePV.PubKey()), "fastnode")
```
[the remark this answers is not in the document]
ebuchman: Fair enough. But we have a bunch of them that take different args so we need to distinguish at the name level as well. Recommendation?

ebuchman:
> Hot stuff! I'm a bit surprised by the DefaultNodeInfo and that we have to type assert it a lot in many places, if we go all the way with the interface idea wouldn't it mean we only call methods and never access fields directly?

Yes. But we need to do more work to get there.

For one, the NodeInfo goes over the wire, so at the least we have to type assert there. 

Though maybe actually we can just expose `MarshalBinary` method - we might want that anyways to deal with the `ID_` field (ie. just make it private).

ebuchman:
Ok, I addressed a lot of the feedback, but after spending a bunch of time with it, I'm not entirely sure how to improve this much more right now.

I like the many interfaces idea, but our components are rather complected right now, so it will take some time to tease things apart to be able to do that well. Thought I started trying to simplify the interfaces a bit here. 

As for ID_, it will go away when we add the sub-structs.

So I think we should merge this and follow up with trying to improve the interfaces

PR #2557, opened 2018-10-06
abci: codespace

ebuchman:
Closes https://github.com/tendermint/tendermint/issues/1858

* [ ] Wrote tests


ValarDragon:
I'm not really a fan of codespace as a string on every Check/Deliver Tx response. We probably don't have time prelaunch, but I think it would be worth benchmarking the performance overhead vs the int based codespacing. I know in the ADR it said that it was complex and hacky, but I completely disagree. It seemed like a very natural and simple thing to me. (If the performance overhead is negligible, then my concern is gone)

Issue #1959, opened 2018-07-11
Change ed25519 and bcrypt to be forked off golang/x/crypto


ValarDragon:
Ref: https://github.com/tendermint/go-crypto/issues/60

/cc @ebuchman

ValarDragon:
I can see two ways of doing this. One is we fork the entire `x/crypto`, redo our changes, and see the upstream updates directly.

The other is we keep our current fork of ed25519 (agl's library, x/crypto is basically the same + 2 updates which I've already made PR's for). We also then fork bcrypt and maintain that as a repo. 

I think having the seperate ed25519 and bcrypt forks is fine. The issue we ran into with the tendermint repos is that we were developing actively on all of them at once. With ed25519/bcrypt, they should be extremely stable. (Any updates they have should not be breaking)

ebuchman:
Why would we keep separate forks?

ValarDragon:
We could have a single `x/crypto` fork, I'm not at all opposed to that.

ebuchman:
Ok I made one: https://github.com/tendermint/crypto

It has the update for bcrypt and I updated  #2090 to use it. 

Also opened a fork of bip39 rather than copying it in. So #2090 has both pkgs removed :)

ebuchman:
Closing for https://github.com/tendermint/tendermint/pull/2558

PR #2558, opened 2018-10-06
ed25519: use golang/x/crypto fork

ebuchman:
Closes https://github.com/tendermint/tendermint/issues/1959.

- Remove some dead code (Curve25519 methods)
- Assume ed25519 key always comes with pubkey (avoid need for MakePublicKey method)
- Use golang.org/x/crypto

* [ ] Wrote tests


crypto/ed25519/ed25519.go
```diff
@@ -65,14 +64,14 @@ func (privKey PrivKeyEd25519) PubKey() crypto.PubKey {
 			break
 		}
 	}
-	if initialized {
-		var pubkeyBytes [PubKeyEd25519Size]byte
-		copy(pubkeyBytes[:], privKeyBytes[32:])
-		return PubKeyEd25519(pubkeyBytes)
+
+	if !initialized {
+		panic("pubkey bytes should always be initialized!")
```
ValarDragon: This may not be true with multi-language support for generating private keys. I'm in favor of keeping the old code here.

crypto/ed25519/ed25519.go
```diff
@@ -85,17 +84,6 @@ func (privKey PrivKeyEd25519) Equals(other crypto.PrivKey) bool {
 	}
 }
 
-// ToCurve25519 takes a private key and returns its representation on
-// Curve25519. Curve25519 is birationally equivalent to Edwards25519,
-// which Ed25519 uses internally. This method is intended for use in
-// an X25519 Diffie Hellman key exchange.
-func (privKey PrivKeyEd25519) ToCurve25519() *[PubKeyEd25519Size]byte {
```
ValarDragon: Don't we need this for secret connection?

ebuchman: Apparently not? It doesn't seem to be used anywhere

crypto/ed25519/ed25519.go
ebuchman: So the only place we read in a priv key is the `priv_validator.json` and `node_key.json`. I think we can just make it a requirement that these come in the format that includes the pubkey? Then we don't have to worry here.

Think that's legit?

ValarDragon: Having the pubkey inside of the encoded private key for ed25519 is a design quirk they had to do due to the limited interface of cryptosystems. I think we should support using the normal format as well.

Gopkg.toml
```diff
@@ -72,6 +72,11 @@
 ## Some repos dont have releases.
 ## Pin to revision
 
+[[constraint]]
+  name = "golang.org/x/crypto"
```
melekes: not a fan of doing this. if I am a developer and I see an import "github.com/.../...", I usually go to github.com to check out the source code for that import. But here **you get what you see** concept is broken, so I had to be aware of the alias ??somehow??

ebuchman: Ok - would a comment everywhere we import this that it's actually an alias be sufficient?

melekes: Yep. Or we could indicate that it's a fork by appending `fork` or `tendermint` to name:`tendermint.com/golang.org/x/crypto` ||`golang.tendermint.com/x/crypto`

ebuchman: Hmm - I like this idea but `dep` doesn't want to let me do it so long as these URLs don't exist. Any ideas? Or do we need to create the valid endpoint first?

melekes: > I like this idea but dep doesn't want to let me do it so long as these URLs don't exist

did not know that! probably not worth the efforts short-term

ebuchman: Golang forking voodoo. It breaks things like `internal` packages

ValarDragon: ohh, ok this lgtm then

Issue #2563, opened 2018-10-06
Implement IPC as an alternative option to SecureConnection


Slamper:
Ref: https://github.com/tendermint/kms/issues/57

An IPC channel (UNIX socket) between KMS and a tendermint validator node should serve the use case where validators deploy the KMS next to a validator node on the same host.

This could allow for a connection with less overhead complexity thereby increasing reliability.

I could take care of the issue and in the course of this also fix the SecureConnection implementation which is currently broken.

Since the underlying TCP socket is already built in (and can be shared with SecureConnection) I will not need to introduce a lot of unnecessary overhead.

ValarDragon:
Its not clear to me that we want to support unencrypted IPC. Are there benchmarks for the overhead complexity you refer to? Chacha20 is extremely quick, it doesn't feel like this should be an issue. Whereas the alternative, unencrypted IPC, has caused many security issues in the past across many different projects.

Also what do you mean by `SecureConnection implementation which is currently broken.`, whats broken about it?

Slamper:
@ValarDragon

The complexity was referring to complexity of the implementation / reliability rather than computational complexity.

Having a network socket + self implemented encrypted connection is too much added complexity over a simple Unix socket for the reward. Especially if the KMS is running on the same host.

For the broken part please take a look at my PR. Connection deadlines were implemented in the wrong way and keepalive was not working.

ValarDragon:
I do still feel like we should authenticate the data. The overhead here would be an HMAC and an initial key exchange. I guess the system is probably unsafe anyway against any sort of physical tampering, so this isn't _needed_, but this does leave open a bit of low-hanging fruit for such an adversary.

amrali:
Isn't it also a low hanging fruit to protect it? To properly secure the keys used to authenticate and encrypt messages between the validator and the KMS is more complex than protecting that simple UNIX socket.

Compromise of any node (validator or KMS) means the connection is compromised, regardless to what we use.

Compromise of the medium/connection is easier for a UNIX socket compared to SecureConnection, but easier is a poor term to describe this, to give an example, if we compare compromising seccomp or any other kernel security to cryptographic security, sure a cryptosystem is more secure than this. But what we are talking about is compromise of several systems and software layers before you can attack the KMS host. I think the cryptosystem adds incremental security there and doesn't act as a primary security barrier for preventing comm channel/medium compromise.

ValarDragon:
```
[...]

PR #2568, opened 2018-10-07
privval: Add IPCPV and fix SocketPV

hendrikhofstadt:
Ref: #2563 

I added IPC as an unencrypted alternative to SocketPV.

Besides I fixed the following aspects of SocketPV:
- Added locking since we are operating on a single socket
- The connection deadline is extended every time a successful packet exchange happens; otherwise the connection would always die permanently x seconds after the connection was established.
- Added a ping/heartbeat mechanism to keep the connection alive; native TCP keepalives do not work in this use-case

* [X] Wrote tests

### DIFF
--- a/privval/ipc_test.go
+++ b/privval/ipc_test.go
@@ -20,7 +20,7 @@ func TestIPCPVVote(t *testing.T) {
 		sc, rs  = testSetupIPCSocketPair(t, chainID, types.NewMockPV())
 
 		ts    = time.Now()
-		vType = types.VoteTypePrecommit
+		vType = types.PrecommitType
 		want  = &types.Vote{Timestamp: ts, Type: vType}
 		have  = &types.Vote{Timestamp: ts, Type: vType}
 	)
@@ -38,7 +38,7 @@ func TestIPCPVVoteResetDeadline(t *testing.T) {
 		sc, rs  = testSetupIPCSocketPair(t, chainID, types.NewMockPV())
 
 		ts    = time.Now()
-		vType = types.VoteTypePrecommit
+		vType = types.PrecommitType
 		want  = &types.Vote{Timestamp: ts, Type: vType}
 		have  = &types.Vote{Timestamp: ts, Type: vType}
 	)
@@ -65,7 +65,7 @@ func TestIPCPVVoteKeepalive(t *testing.T) {
 		sc, rs  = testSetupIPCSocketPair(t, chainID, types.NewMockPV())
 
 		ts    = time.Now()
-		vType = types.VoteTypePrecommit
+		vType = types.PrecommitType
 		want  = &types.Vote{Timestamp: ts, Type: vType}
 		have  = &types.Vote{Timestamp: ts, Type: vType}
 	)
--- a/privval/tcp_test.go
+++ b/privval/tcp_test.go
@@ -96,7 +96,7 @@ func TestSocketPVVoteResetDeadline(t *testing.T) {
 		sc, rs  = testSetupSocketPair(t, chainID, types.NewMockPV())
 
 		ts    = time.Now()
-		vType = types.VoteTypePrecommit
+		vType = types.PrecommitType
 		want  = &types.Vote{Timestamp: ts, Type: vType}
 		have  = &types.Vote{Timestamp: ts, Type: vType}
 	)
@@ -123,7 +123,7 @@ func TestSocketPVVoteKeepalive(t *testing.T) {
 		sc, rs  = testSetupSocketPair(t, chainID, types.NewMockPV())
 
 		ts    = time.Now()
-		vType = types.VoteTypePrecommit
+		vType = types.PrecommitType
 		want  = &types.Vote{Timestamp: ts, Type: vType}
 		have  = &types.Vote{Timestamp: ts, Type: vType}
 	)


privval/ipc.go
```diff
@@ -0,0 +1,218 @@
...
+type IPCPV struct {
+	cmn.BaseService
+	*RemoteSignerClient
+
+	addr string
+
+	connTimeout time.Duration
+
+	lock       sync.Mutex
+	cancelPing chan bool
```
xla: Not used.

privval/ipc.go
```diff
@@ -0,0 +1,218 @@
...
+	"time"
+
+	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/types"
+)
+
+// IPCPV implements PrivValidator, it uses a unix socket to request signatures
+// from an external process.
+type IPCPV struct {
```
xla: It makes sense to encode the protocol in with the advent of alternative implementations, but we should have all implemenetations consistent. Means that `SocketPV` should likely become `TCPPV`. Which in turn reads horribly, maybe we can consider `IPCVal` and `TCPVal`.

privval/ipc.go
```diff
@@ -0,0 +1,218 @@
...
+	if err != nil {
+		return err
+	}
+
+	conn, err := net.DialUnix("unix", nil, la)
+	if err != nil {
+		return err
+	}
+
+	// Wrap in a timeoutConn
```
xla: This comment is not adding information beyond what the line below already communicates, can be omitted or rephrased.

privval/ipc.go
```diff
@@ -0,0 +1,218 @@
...
+	sc.conn = newTimeoutConn(conn, sc.connTimeout)
+
+	return nil
+}
+
+//---------------------------------------------------------
```
xla: Let's not continue to have meaningless ascii art in our codebase, unless it entails unicorns, axolotls or nyancats.

privval/ipc_test.go
```diff
@@ -0,0 +1,134 @@
...
+import (
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/types"
+	"io/ioutil"
+	"os"
+	"testing"
+	"time"
```
xla: Imports should go in the following order seperated by a blank line: `stdlib, external packages, internal packages`.

privval/ipc.go
```diff
@@ -0,0 +1,218 @@
...
+
+// IPCPV implements PrivValidator, it uses a unix socket to request signatures
+// from an external process.
+type IPCPV struct {
+	cmn.BaseService
+	*RemoteSignerClient
+
+	addr string
+
+	connTimeout time.Duration
```
xla: This needs to be controlled from outside the package boundaries. I'd suggest we introduce functional options like the ones you removed for `SocketPV`.

privval/ipc.go
```diff
@@ -0,0 +1,218 @@
...
+	logger log.Logger,
+	socketAddr string,
+) *IPCPV {
+	sc := &IPCPV{
+		addr:        socketAddr,
+		connTimeout: connTimeout,
+	}
+
+	sc.BaseService = *cmn.NewBaseService(logger, "IPCPV", sc)
+	sc.RemoteSignerClient = NewRemoteSignerClient(logger, nil)
```
xla: Either we accept it as a dependency of the constructor or the construction of the remote signer client and it being public needs to be challenged.

privval/ipc.go
```diff
@@ -0,0 +1,218 @@
...
+// from an external process.
+type IPCPV struct {
+	cmn.BaseService
+	*RemoteSignerClient
+
+	addr string
+
+	connTimeout time.Duration
+
+	lock       sync.Mutex
```
xla: Appears that `lock` is not used anywhere, let's remove it.

privval/ipc.go
```diff
@@ -0,0 +1,218 @@
...
+		err = cmn.ErrorWrap(err, "failed to connect")
+		sc.Logger.Error(
+			"OnStart",
+			"err", err,
+		)
+
+		return err
+	}
+
+	err = sc.RemoteSignerClient.Start()
```
xla: Nested lifecycle management seems undesirable, see general comment about the abstraction of the rs client.

privval/ipc.go
```diff
@@ -0,0 +1,218 @@
...
+
+// OnStart implements cmn.Service.
+func (sc *IPCPV) OnStart() error {
+	err := sc.connect()
+	if err != nil {
+		err = cmn.ErrorWrap(err, "failed to connect")
+		sc.Logger.Error(
+			"OnStart",
+			"err", err,
+		)
```
xla: This easily fits in a single line.

privval/ipc.go
```diff
@@ -0,0 +1,218 @@
...
+		return err
+	}
+
+	err = sc.RemoteSignerClient.Start()
+	if err != nil {
+		err = cmn.ErrorWrap(err, "failed to start RemoteSignerClient")
+		sc.Logger.Error(
+			"OnStart",
+			"err", err,
+		)
```
xla: Should be a one liner.

privval/ipc.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// OnStop implements cmn.Service.
+func (sc *IPCPV) OnStop() {
+	if err := sc.RemoteSignerClient.Stop(); err != nil {
+		err = cmn.ErrorWrap(err, "failed to stop RemoteSignerClient")
+		sc.Logger.Error(
+			"OnStop",
+			"err", err,
+		)
```
xla: One liner.

privval/ipc.go
```diff
@@ -0,0 +1,218 @@
...
+	sc.RemoteSignerClient = NewRemoteSignerClient(logger, nil)
+
+	return sc
+}
+
+// OnStart implements cmn.Service.
+func (sc *IPCPV) OnStart() error {
+	err := sc.connect()
+	if err != nil {
+		err = cmn.ErrorWrap(err, "failed to connect")
```
xla: We don't want to promote this part of the common package and instead have sentinel or proper error types which the caller can check against.

privval/ipc.go
```diff
@@ -0,0 +1,218 @@
...
+			"OnStart",
+			"err", err,
+		)
+
+		return err
+	}
+
+	err = sc.RemoteSignerClient.Start()
+	if err != nil {
+		err = cmn.ErrorWrap(err, "failed to start RemoteSignerClient")
```
xla: Should be sentinel or full error type.

privval/ipc.go
```diff
@@ -0,0 +1,218 @@
...
+		err = cmn.ErrorWrap(err, "failed to stop RemoteSignerClient")
+		sc.Logger.Error(
+			"OnStop",
+			"err", err,
+		)
+	}
+
+	if sc.conn != nil {
+		if err := sc.conn.Close(); err != nil {
+			err = cmn.ErrorWrap(err, "failed to close connection")
```
xla: Proper error type.

privval/socket_tcp.go
```diff
@@ -50,17 +67,24 @@ func (ln tcpTimeoutListener) Accept() (net.Conn, error) {
...
-		return nil, err
-	}
+	return conn, nil
+}
 
-	if err := tc.SetKeepAlivePeriod(ln.period); err != nil {
-		return nil, err
-	}
+// Read implements net.Listener.
+func (c timeoutConn) Read(b []byte) (int, error) {
```
xla: Methods should be close to the definition of the type.

privval/socket.go
```diff
@@ -35,51 +36,25 @@ var (
 
 var (
 	acceptDeadline = time.Second * defaultAcceptDeadlineSeconds
-	connDeadline   = time.Second * defaultConnDeadlineSeconds
+	connTimeout    = time.Second * defaultConnDeadlineSeconds
 	connHeartbeat  = time.Second * defaultConnHeartBeatSeconds
 )
 
-// SocketPVOption sets an optional parameter on the SocketPV.
-type SocketPVOption func(*SocketPV)
```
xla: Why were all the options removed? They provide a way to tune vital aspects of the implementation from outside the package. Reasoning here is appreciated, but ideally those deletions should be reverted.

privval/ipc_test.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func testUnixAddr() string {
+	f, err := ioutil.TempFile("/tmp", "nettest")
+	if err != nil {
+		panic(err)
+	}
+	addr := f.Name()
+	f.Close()
+	os.Remove(addr)
```
xla: Both those return errors, shouldn't be ignored.

privval/rs_client.go
```diff
@@ -0,0 +1,236 @@
...
+
+import (
+	"fmt"
+	"github.com/tendermint/tendermint/crypto"
+	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/types"
+	"net"
+	"sync"
+	"time"
```
xla: Import order.

privval/rs_client.go
```diff
@@ -0,0 +1,236 @@
...
+
+// RemoteSignerClient implements PrivValidator, it uses a socket to request signatures
+// from an external process.
+type RemoteSignerClient struct {
+	cmn.BaseService
+
+	connHeartbeat time.Duration
+
+	conn       net.Conn
+	lock       sync.Mutex
```
xla: What is the reason for the lock? `net.Conn` is threadsafe.

privval/rs_client.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	resp, ok := res.(*SignedVoteResponse)
+	if !ok {
+		return ErrUnexpectedResponse
+	}
+	if resp.Error != nil {
+		return fmt.Errorf("remote error occurred: code: %v, description: %s",
+			resp.Error.Code,
+			resp.Error.Description)
```
xla: Needs proper error type.

Odd formatting, should be in the form of:

``` go
	return fmt.Errorf(
		"remote error occurred: code: %v, description: %s",
		resp.Error.Code,
		resp.Error.Description,
	)
```

privval/rs_client.go
```diff
@@ -0,0 +1,236 @@
...
+// RemoteSignerClient implements PrivValidator, it uses a socket to request signatures
+// from an external process.
+type RemoteSignerClient struct {
+	cmn.BaseService
+
+	connHeartbeat time.Duration
+
+	conn       net.Conn
+	lock       sync.Mutex
+	cancelPing chan bool
```
xla: If this channel is used to signal termination it can well be just chan struct{} and the closing of the channel should suffice.

privval/socket.go
```diff
@@ -35,51 +36,25 @@ var (
...
-	connHeartbeat   time.Duration
+	connTimeout     time.Duration
 	connWaitTimeout time.Duration
 	privKey         ed25519.PrivKeyEd25519
 
-	conn     net.Conn
-	listener net.Listener
+	listener   net.Listener
+	lock       sync.Mutex
+	cancelPing chan bool
```
xla: Not used.

privval/socket.go
```diff
@@ -35,51 +36,25 @@ var (
...
-	connDeadline    time.Duration
-	connHeartbeat   time.Duration
+	connTimeout     time.Duration
 	connWaitTimeout time.Duration
 	privKey         ed25519.PrivKeyEd25519
 
-	conn     net.Conn
-	listener net.Listener
+	listener   net.Listener
+	lock       sync.Mutex
```
xla: Not used.

xla:
@Liamsi Could you review the proposed heartbeat messages in this PR?

privval/rs_client.go
Slamper: The idea was that if the KMS does asynchronous receive the following could happen:

TM sends a sign request.
TM send a getpubkey request.

The KMS handles both but for whichever reason it finishes the getpubkey request first.

Then the sign method would receive the packet for the getpubkey then since we have no package IDs etc. 

As long as the KMS side does proper synchronizing this is not an issue but we should take care of this here.

Or do I have misconceptions about how this works ?

privval/rs_client.go
Slamper: Damn I thought I had run go imports...
Sry. Will fix.

privval/ipc.go
Slamper: Ok. I will also remove it from the existing socketpv

privval/socket.go
Slamper: I thought they were unnecessary to add becuase nonetheless these options can right now not be modified after OnStart is called. But outside modification makes sense so I will reintroduce them.

@xla do you think we should refactor the implementation to allow modifications after Start is called ?

privval/ipc.go
Slamper: @xla How do I then pass the original error alongside the error type ?

privval/rs_client.go
Slamper: I will create a dedicated error type.

Liamsi:
> @Liamsi Could you review the proposed heartbeat messages in this PR?

They are sane in the sense that they do the job: every `connHeartbeat` period a `PingRequest` is sent to keep the connection alive. What I found confusing though is that I would have expected the ´SignHeartbeatRequest´ to be used exactly for this purpose though. Now we have two "hearbeat messages", one which is signed (and holds info about the current round/height etc) and one which is a simple (empty) registered amino message. Am I the only one who finds this confusing?

Slamper:
@Liamsi 
While the pings I implemented are there to maintain and check the health of the connection, the Heartbeat you mention is an integral part of consensus and imho should not be used for connection maintenance. Not only because it is a part of the consensusreactor and not the connection but also because it might not be suited for every connection type.

Also the pings are sent regularly at fixed intervals opposed to the Heartbeat (at least from what I have read).

Liamsi:
Thanks for the prompt reply! Although you are right and we do not have a go-routine or alike which periodically sends `Heartbeat` messages, I thought the name alone was already an indicator that it is meant to be send periodically (fixed intervals). I might be wrong though (cc @ebuchman, @milosevic). If the orig. heartbeat message is meant to be sent periodically, would we still need an additional message to keep the connection alive? For which connection types would it be unsuited? 

Although the reactor deals with a `ProposalHeartbeatMessage`, the core consensus doesn't know about the heartbeat either (at least the [spec] of the core consensus doesn't explicitly mention a heartbeat message). 

[spec]: https://github.com/tendermint/spec/releases

Slamper:
@Liamsi 
Thanks for explaining more in-depth what you meant.

What I saw was that the heartbeats are only sent if `waitForTxs` is true.
https://github.com/tendermint/tendermint/blob/cd172acee82aa2aa9c34735eb8f65827c943b623/consensus/state.go#L783-L787

A connection relevant heartbeat/keepalive should not rely on a consensus reactor setting but have to sole purpose to maintain the connection also the period should be deterministic/fixed which the proposeHeartbeat is not from what I could see.

ebuchman:
@Slamper do you mind if we merge https://github.com/tendermint/tendermint/pull/2548 first to get the small fix out of the way?

Slamper:
@ebuchman
No. I think it makes sense to merge your fix first 👍 I did not seen it when I created this PR.

ebuchman:
Cool thanks - merged

privval/ipc.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	sc.RemoteSignerClient = NewRemoteSignerClient(sc.conn)
+
+	// Start a routine to keep the connection alive
+	sc.cancelPing = make(chan struct{}, 1)
+	go func() {
+		for {
+			select {
+			case <-time.Tick(sc.connHeartbeat):
```
melekes: `While Tick is useful for clients that have no need to shut down the Ticker, be aware that without a way to shut it down the underlying Ticker cannot be recovered by the garbage collector; it "leaks".`

won't we leak memory here? and why not have a regular ticker and stop it when `<-sc.cancelPing`?

privval/ipc.go
```diff
@@ -0,0 +1,117 @@
...
+var _ types.PrivValidator = (*IPCVal)(nil)
+
+// NewIPCVal returns an instance of IPCVal.
+func NewIPCVal(
+	logger log.Logger,
+	socketAddr string,
+) *IPCVal {
+	sc := &IPCVal{
+		addr:          socketAddr,
+		connTimeout:   connTimeout,
```
melekes: not obvious where it's coming from

privval/ipc_test.go
```diff
@@ -0,0 +1,146 @@
+package privval
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
melekes: can we add a newline here to separate our libs from external ones?

privval/remote_signer.go
```diff
@@ -0,0 +1,313 @@
...
+) *RemoteSignerClient {
+	sc := &RemoteSignerClient{
+		conn: conn,
+	}
+	return sc
+}
+
+// GetAddress implements PrivValidator.
+func (sc *RemoteSignerClient) GetAddress() types.Address {
+	addr, err := sc.getAddress()
```
melekes: Can't we just do `return sc.GetPubKey().Address()`?

privval/tcp.go
```diff
@@ -0,0 +1,223 @@
...
+	sc.conn = conn
+
+	sc.RemoteSignerClient = NewRemoteSignerClient(sc.conn)
+
+	// Start a routine to keep the connection alive
+	sc.cancelPing = make(chan struct{}, 1)
+	go func() {
+		for {
+			select {
+			case <-time.Tick(sc.connHeartbeat):
```
melekes: same question

privval/tcp.go
```diff
@@ -0,0 +1,223 @@
...
+// waitConnection uses the configured wait timeout to error if no external
+// process connects in the time period.
+func (sc *TCPVal) waitConnection() (net.Conn, error) {
+	var (
+		connc = make(chan net.Conn, 1)
+		errc  = make(chan error, 1)
+	)
+
+	go func(connc chan<- net.Conn, errc chan<- error) {
+		conn, err := sc.acceptConnection()
```
melekes: will this goroutine unblock if we exit with `ErrConnWaitTimeout` or it will be running forever?

privval/ipc.go
Slamper: Absolutely right. I will refactor it to use a regular ticker

privval/tcp.go
Slamper: See above.

privval/remote_signer.go
Slamper: I stuck to how it was implemented before in SocketPV. There the public method implements the PV interface and the private methods impllement it.

However it just adds overhead code here I agree. I would also go with simplification over "style consistency" here.

privval/tcp.go
Slamper: @melekes It will not in the normal case since the acceptConnection function will timeout first (if the timeout is shorter which is normally the cases).

But what you point out is very important because this means the original implementation was broken because the function would almost always fail long before the timer is called because acceptConnection errors (because of the acceptDeadline) making the connWaitTimeout useless.

I think I will just get rid of the timer and adjust the default acceptDeadline accordingly. Also some tests would make sense then.

Slamper:
Btw this is still WIP since we decided to implement a simple MAC auth in IPCVal #2563

privval/ipc.go
Slamper: @melekes Do you mean because of the name clash of the constant and field ? It was like this in the original implementation but I'm happy to change that if that's what you mean.

melekes: No, it's just the const is not in this file, right? can't find it. please ignore

ValarDragon:
You could add the MAC in a separate PR :)

melekes:
@Slamper will you have time to address my comments? After these are fixed, we can merge this PR.

Slamper:
@melekes I will do so today and then we can merge :+1:

PR #2571, opened 2018-10-08
separate mock evidence from real evidence

### LINKED ISSUE #2525: RegisterEvidences(cdc) currently registers mock evidence types
Currently register evidences is as follows:
```
func RegisterEvidences(cdc *amino.Codec) {
	cdc.RegisterInterface((*Evidence)(nil), nil)
	cdc.RegisterConcrete(&DuplicateVoteEvidence{}, "tendermint/DuplicateVoteEvidence", nil)

	// mocks
	cdc.RegisterConcrete(MockGoodEvidence{}, "tendermint/MockGoodEvidence", nil)
	cdc.RegisterConcrete(MockBadEvidence{}, "tendermint/MockBadEvidence", nil)
}
```
I think this will lead to unexpected behavior happening if people start submitting MockGoodEvidence, since we're currently supporting that. We should separate this into `RegisterEvidence` and `RegisterMockEvidence`.

melekes:
Closes #2525

* [ ] Wrote tests

### DIFF
--- a/evidence/pool_test.go
+++ b/evidence/pool_test.go
@@ -1,6 +1,7 @@
 package evidence
 
 import (
+	"os"
 	"sync"
 	"testing"
 
@@ -13,6 +14,13 @@ import (
 )
 
 var mockState = sm.State{}
+
+func TestMain(m *testing.M) {
+	types.RegisterMockEvidences(cdc)
+
+	code := m.Run()
+	os.Exit(code)
+}
 
 func initializeValidatorState(valAddr []byte, height int64) dbm.DB {
 	stateDB := dbm.NewMemDB()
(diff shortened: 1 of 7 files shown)


evidence/pool_test.go
```diff
@@ -14,6 +15,13 @@ import (
 
 var mockState = sm.State{}
 
+func TestMain(m *testing.M) {
+	types.RegisterMockEvidences(cdc)
+
+	code := m.Run()
```
xla: Why the extra assignment instead of passing `m.Run` straight to `os.Exit`?

melekes: in my case it's easier to follow the execution path. in the case of `os.Exit(m.Run())`, the accent is shifted towards os.Exit, which should not be the case

Issue #2414, opened 2018-09-16
Validators can run with Secp256k1 pubkeys


ValarDragon:
**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):
23.1

**ABCI app** (name for built-in, URL for self-written if it's publicly available):
Cosmos SDK develop

**What happened**:
Blocks were built with a validator with a Secp256k1Pubkey

**What you expected to happen**:
No blocks to be built / there to be an error or panic

**How to reproduce it** (as minimally and precisely as possible):
Take a gaia genesis file, replace the validator pubkey with a secp one, e.g:
```
"pub_key": {
        "type": "tendermint/PubKeySecp256k1",
        "value": "AmmUPwalYvC/zycpMeUy5X8pPU6oE7DFbu5fo3y8nAoY"
      }
```

We probably should check the key type on `types.NewValidator`, make this a consensus param, or make the consensus codec only know about ed25519 and not secp. I'm leaning towards making this a consensus param, since there is interest in using other signature schemes.

ValarDragon:
If we're making this a consensus parameter, it probably shouldn't be enforced in `types.NewValidator`. Where / when should it be checked? On every validator set update?

ebuchman:
- [ ] Add ValidatorPubkeyTypes to ConsensusParams - #2636 
- [ ] Enforce ValidatorPubkeyTypes in the State

ebuchman:
Completed in https://github.com/tendermint/tendermint/pull/2739

PR #2574, opened 2018-10-08
crypto: Add a way to go from pubkey to route

ValarDragon:
This is intended for use in a future PR for #2414 

* [x] Wrote tests
* [ ] Updated CHANGELOG_PENDING.md - n/a imo

### DIFF
--- a/crypto/encoding/amino/amino.go
+++ b/crypto/encoding/amino/amino.go
@@ -1,8 +1,11 @@
 package cryptoAmino
 
 import (
-	amino "github.com/tendermint/go-amino"
+	"errors"
 
+	"reflect"
+
+	amino "github.com/tendermint/go-amino"
 	"github.com/tendermint/tendermint/crypto"
 	"github.com/tendermint/tendermint/crypto/ed25519"
 	"github.com/tendermint/tendermint/crypto/multisig"
@@ -10,6 +13,7 @@ import (
 )
 
 var cdc = amino.NewCodec()
+var routeTable = make(map[reflect.Type]string, 3)
 
 func init() {
 	// NOTE: It's important that there be no conflicts here,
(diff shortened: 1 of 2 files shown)


crypto/encoding/amino/amino.go
```diff
@@ -19,6 +23,23 @@ func init() {
 	// https://github.com/tendermint/go-amino/issues/9
 	// is resolved
 	RegisterAmino(cdc)
+
+	// TODO: Have amino provide a way to go from concrete struct to route directly.
+	// Its currently a private API
```
Liamsi: If not happened already: please open an issue in amino too and link all relevant issues / PRs.

crypto/encoding/amino/amino.go
```diff
@@ -1,15 +1,19 @@
...
 
+	amino "github.com/tendermint/go-amino"
 	"github.com/tendermint/tendermint/crypto"
 	"github.com/tendermint/tendermint/crypto/ed25519"
 	"github.com/tendermint/tendermint/crypto/multisig"
 	"github.com/tendermint/tendermint/crypto/secp256k1"
 )
 
 var cdc = amino.NewCodec()
+var routeTable = make(map[reflect.Type]string, 3)
```
melekes: don't u think this deserve a comment? with an example entry `ed25519 -> tendermint/XXX/ed25519`

crypto/encoding/amino/amino.go
```diff
@@ -19,6 +23,23 @@ func init() {
...
+	routeTable[reflect.TypeOf(&multisig.PubKeyMultisigThreshold{})] = multisig.PubKeyMultisigThresholdAminoRoute
+}
+
+// PubkeyAminoRoute returns the amino route of a pubkey
+// cdc is currently passed in, as eventually this will not be using
+// a package level codec.
+func PubkeyAminoRoute(cdc *amino.Codec, key crypto.PubKey) (string, error) {
+	route, ok := routeTable[reflect.TypeOf(key)]
+	if !ok {
+		return "", errors.New("Pubkey type not known")
```
melekes: can we return a bool instead of error? it will simplify the code drastically

```
return routeTable[reflect.TypeOf(key)]
```

ValarDragon: Sure!

sos34211: Ff

PR #2576, opened 2018-10-08
libs: Refactor & document events code

melekes:


libs/events/events_test.go
```diff
@@ -174,40 +172,38 @@ func TestAddAndRemoveListenerConcurrency(t *testing.T) {
...
 	}()
 
 	<-done1
 	<-done2
 
-	close(done1)
-	close(done2)
-
-	evsw.RemoveListener("listener") // make sure remove last
+	evsw.RemoveListener("listener") // remove the last listener
```
melekes: @goolAdapter why do we treat last listener specially?

goolAdapter: because AddListenerForEvent routine may finish after RemoveListener routine，at this time， the evsw have some "listener" callback as except.
our test is to demonstration RemoveListener can't clear up the cb because race time window.

PR #2578, opened 2018-10-09
Locked blockID

### LINKED ISSUE #2567: defaultSetProposal will set cs.ProposalBlockParts, this may lead to interrupt the block receiving process of PoLC's BlockID

**Environment**:
- **Install tools**:
- **Others**:

**What happened**:
I think there are 2 situations when a node needs to receive block parts,
1. it receives a proposal of the round, so it needs to receive the proposalBlock from the proposer
2. it receives a PoLc, so it has got a BlockID which contains a BlockPartsHeader. This situation will also set cs.ProposalBlockParts ,but it can go without receiving the proposal.

So how about the situation 2 when latter the proposal comes, it will override cs.ProposalBlockParts . It will lead to interrupt the block receiving of PoLC's BlockID.

I suggest we can add a cs.LockedBlockID to indicate we are now receiving blockParts because we have a lockedBlockID.
Now in defaultSetProposal, we can judge 
if cs.LockedBlockID !=nil {
   return
}

so the cs.ProposalBlockParts won't be overwritten.

james-ray:
* [ ] Wrote tests

fix #2567

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -1162,6 +1162,7 @@ func (cs *ConsensusState) enterPrecommit(height int64, round int) {
 	cs.LockedBlockParts = nil
 	if !cs.ProposalBlockParts.HasHeader(blockID.PartsHeader) {
 		cs.ProposalBlock = nil
+		cs.LockedRound = cs.Round
 		cs.LockedBlockID = &types.BlockID{blockID.Hash,blockID.PartsHeader}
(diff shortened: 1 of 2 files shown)


james-ray:
I'll try to write some tests to imitate that complicated exception scene

james-ray:
as @ebuchman said, this isn't a bug. So it's not necessary to do this.
I will close this PR.

ebuchman:
Thanks!

Issue #2108, opened 2018-07-30
libs/db: panic triggered by large FSBD key


Liamsi:
> the FSDB.SetNoLock() method panics when passed a large key with a length of 256 bytes on a Linux machine. This is due to the method creating a filename which includes the key in the file name. Linux has a maximum filename length of 256 bytes. On line 92 below, the library panics:

code:

https://github.com/tendermint/tendermint/blob/dfa9a9a30a666132425b29454e90a472aa579a48/libs/db/fsdb.go#L85-L94

PoC (slightly modified imports to reflect merged repos):

```go
package main
import "github.com/tendermint/tendermint/libs/db"
import "strings"
func main() {
	db := db.NewDB("mydb", "fsdb","/tmp/")
	db.Set([]byte(strings.Repeat("A",(256))),[]byte("bar"))
}
```
Result: `panic: Setting key`

Possible mitigation:

`SetNoLock` should handle the not allowed input (here the key) by returning a (ideally sentinel) error instead of panicing.

ValarDragon:
To be clear, this issue entails changing the `atomicSetDeleter` interface to return an error on all the set methods.

overbool:
>Resolving this issue entails changing the atomicSetDeleter interface to return an error on all the set methods.

@ValarDragon I have 2 question about your this.

1. `FSDB` does not implement the `atomicSetDeleter` interface 
2. Should we changing the `DB` interface to return an error on all the set methods.

ebuchman:
> Should we changing the DB interface to return an error on all the set methods?

Dealing with this issue properly would require returning errors in all the Set methods, as suggested, but I'm not sure we want to do that, as these methods don't return errors by design (since an error here is supposed to imply some kind of critical issue, and hence warrants a panic. whether or not that's appropriate design choice is a larger and separate discussion). So it might be better to just explicitly document this in the godoc and leave the current behaviour, and leave it to the caller not to pass an key that's too long. See comment to same effect in https://github.com/tendermint/tendermint/pull/2586#pullrequestreview-171882176

Thoughts?

PR #2586, opened 2018-10-09
[WIP] libs: Modify db atomicSetDeleter interface

overbool:
Fixes: https://github.com/tendermint/tendermint/issues/2108


ebuchman:
Given the above comment, I'm going to close this for now. Feel free to submit another PR updating the comments to make this panic more explicit, or comment on https://github.com/tendermint/tendermint/issues/2108 about why we should do something else. Thanks!

Issue #2315, opened 2018-09-03
Let the proposal with the highest priority to be committed with 2/3+ Prevote instead of 2/3+ Precommit


james-ray:
We know Tendermint generate a predictable proposer sequence. For each height or each round, every validator checks the sequence to see if it is the proposer of that round. That means there is a legal proposer of each round. How about we add a priority table that the round 0 proposer with the highest priority, round 1 proposer be the second priority ...

From my perspective of view, I understand Tendermint Consensus has 2 confinements:
1. If a node enters the next round, it cound not vote for previous round anymore.
2. If a node has a PoLc of previous round and has sent out a Precommit, it has to lock that block until it receives another PoLc of another block in a higher round.

So, how about we add a third rule:
3. If a node doesn't have a PoLc of any block, it should change its votes of next round based on the proposal of the highest priority that it already received. That means if a node has received a round 0 proposal, it would not change its votes in the next rounds except a proposal with low priority has reached a PoLc.

Safety Proof:
If there is a 2/3+ PoLc of a highest priority proposal. Then there will be at least 1/3+ honest nodes prevote that proposal, that means these nodes would not change their vote except there is a PoLc of the lower priority proposal. However the PoLc of the lower priority proposal cannot be reached because there is less than 2/3 nodes left that are available to vote another proposal.

a proposal which is not the highest priority of that height still needs to collect 2/3+ precommits to ensure the safety.

How to do this:
right now,the block format in Tendermint is like this:
```go
type Block struct {
	*Header    `json:"header"`
	*Data      `json:"data"`
	Evidence   EvidenceData `json:"evidence"`
	LastCommit *Commit      `json:"last_commit"`
}
```
and the judge process is as below:
```go
func (commit *Commit) ValidateBasic() error {
	if commit.BlockID.IsZero() {
		return errors.New("Commit cannot be for nil block")
	}
	if len(commit.Precommits) == 0 {
		return errors.New("No precommits in commit")
	}
	height, round := commit.Height(), commit.Round()

	// validate the precommits
	for _, precommit := range commit.Precommits {
		// It's OK for precommits to be missing.
		if precommit == nil {
			continue
		}
		// Ensure that all votes are precommits
		if precommit.Type != VoteTypePrecommit {
			return fmt.Errorf("Invalid commit vote. Expected precommit, got %v",
				precommit.Type)
		}
```

So we could change the judge process, to the below logic:
```go
if lastBlock.priority == 0 then
[...]

PR #2591, opened 2018-10-10
Fix add proposal block part

### LINKED ISSUE #2567: defaultSetProposal will set cs.ProposalBlockParts, this may lead to interrupt the block receiving process of PoLC's BlockID

james-ray:
* [ ] Wrote tests
fix #2567 

don't add a LockedBlockID to check.   Just modify addProposalBlockPart to enable it to receive LockedBlockPart.  So the 2 don't interfere with each other


ebuchman:
Is this addressed by https://github.com/tendermint/tendermint/issues/2567#issuecomment-429463006 ?

james-ray:
> Is this addressed by [#2567 (comment)](https://github.com/tendermint/tendermint/issues/2567#issuecomment-429463006) ?

Yes, it is.
Based on my current code. enterCommit's code has considered not having the proposalBlock.
I will see ##2540 to see the recent updates.

ebuchman:
@james-ray assuming my understanding in https://github.com/tendermint/tendermint/issues/2567#issuecomment-429995700 is correct, can we close this PR for now? We need to focus on Cosmos launch requirements and bug fixes in the meantime.

Thanks!

james-ray:
The code
if cs.Round < vote.Round && prevotes.HasTwoThirdsAny() {
// Round-skip if there is any 2/3+ of votes ahead of us
} else if cs.Round == vote.Round && cstypes.RoundStepPrevote <= cs.Step { // current round
if prevotes.HasTwoThirdsMajority() {
} else if prevotes.HasTwoThirdsAny() {
cs.enterPrevoteWait(height, vote.Round)
}
}

this code cstypes.RoundStepPrevote <= cs.Step could make sure we won't enterPrecommit before enterPrevote. We cound enterPrevote without receiving proposal if timeout after enterPropose. But this case is rare.

So actually it's indeed a rare case.

My proposal may speed up some rounds in LockedBlock nodes is >1/3 but <2/3 and the proposer is not the lockedBlock node.

But as you said , it may go a little different from the offical spec.
I can understand.

james-ray:
Hi,@ebuchman, I have briefly see the spec, in the algorithm psycho-code:
22: upon hPROPOSAL, h p , round p , v, −1i from proposer(h p , round p ) while step p = propose do
23:
if valid(v) ∧ (lockedRound p = −1 ∨ lockedV alue p = v) then
24:
broadcast hPREVOTE, h p , round p , id(v)i
25:
else
26:
broadcast hPREVOTE, h p , round p , nili
27:
step p ← prevote

I think this part is a little different from what we do in defaultDoPrevote, right?
I mean this part,  lockedV alue p = v

Right now, we just prevote for lockedBlock if we have, no matter it equals or not current proposal.

I don't mean any offence , just raise a question.   Thanks for reading and clarification.

milosevic:
> Right now, we just prevote for lockedBlock if we have, no matter it equals or not equals to current proposal.

That's right. Prevoting for locked value instead of nil can be seen as an optimisation as it might lead to committing a block in case where Proposal is not accepted but all correct processes have locked the same value for example. We don't have this in spec as it will complicate algorithm. We are still discussing if it makes sense to keep this optimisation or to prevote nil in case proposal is not accepted.

james-ray:
@milosevic Thanks for your time reading my question and reply.  Actually The TM spec I've read before is not the offical version. This is my first time to read the official version.  It's a little different from what I've read before.  I think the core safety proof of TM is that 2/3+ precommits means 1/3+ honest nodes have locked one same block, this can ensure that there can't be another block with 2/3+ prevotes. (Because honest node will vote lockedBlock if he has, not for any other block).
As long as this safety proof exists, I think they vote for LockedBlock when it equals to or not equals to ProposalBlock are both ok, even ok with any block I think, any block with 2/3+ precommits can ensure the safety.

Thanks for you and the TM team to bring a so excellent open source project to us. I really gained a lot through reading your paper and the codes.  I think it's really hard for anyone to think out a so rigorous consensus algorithm and write a so nice architecture of the code.  I think anyone here hopes Tendermint becomes better and better.   I would be happy and honored to contribute any of my effort to the ADR or the code and approved by you.

milosevic:
@james-ray Thanks a lot for kind words and your help and contribution. It is super appreciated.
In general I agree with your last comment with a small remark that we need to ensure that blocks committed across different rounds are the same. Therefore, we can't prevote for any block, and rely only on precommit phase to ensure agreement.  

I would recommend to take a look at Launch project board (https://github.com/orgs/tendermint/projects/7) and to see if there is something that looks interesting for you to contribute. We are focused on Launch so everything outside the scope (unless it is a new bug) has a smaller chance to have our attention.

james-ray:
Sorry for the any block comment, I realize the mistake. I edit it to "his lockedBlock" ( To other peers, they may think it is "any block" to them)

Thanks for the recommendation, but I found this page  (https://github.com/orgs/tendermint/projects/7)  404 not found.   May I submit a ADR or a paper draft for this proposal?

james-ray:
Along with, I want to propose that we add a Round0ProposalBlockPart to receive the round 0 proposer's block parts.  When it's time to prevote, the priority is: LockedBlock > Round0Proposal > CurrentRound Proposal.    We can refuse to receive current round's proposal unless we confirm it is a potential LockedBlock which means we have received a 2/3+ prevotes have that blockID.
The advantage is the round0proposal can be committed with 2/3+ prevotes instead of 2/3+ precommits, as mentioned in #2315 .
Because 2/3+ prevotes can ensure 1/3+ honest node have received round 0 proposal. It won't change its prevote unless it received a LockedBlock. However 1/3+ honest nodes prevote round 0 proposal can ensure they don't have any lockedBlock, so it's impossible for any block in this round or previous rounds exist. Because if it exists there won't be a 1/3+ honest nodes not vote for it.  Also, after that round, it's impossible for any other block to be get locked either. Because 1/3+ honest nodes won't change their mind and there isn't any block before.

james-ray:
Actually , I've planned to further study in the US. I have finished a GRE test and is going to take TOFEL soon. So any paper work assigned to me would be sincerely appreciated.

james-ray:
Sorry, I have realized my proposal has flaw. Let me explain.

james-ray:
Thanks for the time reading my proposal, and sorry for the carelessness previously.  Would you please give me a chance to summurize my point?

My proposal contains 2 parts:
1.Give a chance to the node to receive LockedBlock of previous Round. 
   When it confirms there is a PoLc (2/3+ prevotes) of previous round and the round number is > his LockedRound, he can unlock his previous block and set LockedRound to that previous round.   And later he must prevote for this block,and if any round this block has got a PoLc, he relocks this block, It means he updates the LockedRound to current round. This is exactly the same as we currently do.

2.Let Round 0 proposal directly enterPrecommit instead of enterPrevote phase.
 The round 0 proposal's precommit is like this commitment:" I don't have any lockedBlock right now, and I will not change my prevote to any other block unless it's a LockedBlock with a PoLc"

But as I found above, These 2 parts combine together will cause a fatal error, below are my thoughts of solution.

1.Disable Part 1, that means disable the node's ability to receive previous round's LockedBlock. That ensures if any round any node has seen a 2/3+ precommit of round 0 proposal, which means 1/3+ honest nodes don't have LockedBlock at that time and since we disable its ability to receive the LockedBlock of previous rounds and ,               LockedBlock of  future round is impossible to formed ( If a node has Locked a block in previous round, it can only propose it in the form of ProposalBlock in future round, however that 1/3+ honest nodes will ignore it because they have precommit round 0 proposal)          So the 2/3+ precommits of round 0 proposal in any round would be strictly secured and safe.

2.Retain the ablity of receiving previous rounds' LockedBlock.    I think in this case we can precommit round 0 proposal without collect a PoLc of it only in round 0 and round 1.  This is because:
In round 0, we can certainly ensure that there can't be a LockedBlock in previous rounds.
And in round 1, if there is another LockedBlock other than round 1's proposal, that must be round 0's proposal.

So the precommit of round 0 proposal ( without collecting 2/3+ prevotes of round 0 or round 1)  combines with this assertion:  there is no previous LockedBlock before.   That means the round 0 proposal has no previous rival,  and the future rival is impossible to form.   So it's secured.

But in round 2 and later rounds,  the round 0 proposal and its block cannot be accepted anymore, that is currently what we do: we can only accept the proposal of current round's proposer.


Thanks for the time reading my proposal again.   I know maybe it's not the right time to propose to the optimization to the consensus core.    But if I got a chance to communicate and discuss with you, the creator of TM, I will get encouraged and inspired, and your remark will help me to improve.  Certainly it's not current major task of us. But I think if I got a chance to discuss about it I will feel greatly honored.   Frankly speaking, now my company is using Tendermint as the consensus core and Ethermint as above application.  We made some optimization of them, such as importing BLS to generate a random number(learned from Dfinity)  we add a LastSign below LastCommits in block.go.  We can use this random number to generate different validator set of each block.   We also implement the award of the block( award to the proposer and the LastCommit validators ).    Now my leader wants me to see if the consensus process can be optimized in anyway.   So your confirm or suggestion to me is very precious, and I feel it's a very precious opportunity to communicate with you too.

ebuchman:
Thanks @james-ray. Appreciate your passion here and glad you've enjoyed working on the project and code. Let's close this PR for now. If you'd like, please open a new issue describing where and how you think we can optimize the consensus.

Note that changes to the consensus code will require corresponding updates to the [spec and proofs](https://github.com/tendermint/spec) first. Our priority right now is ensuring correctness of the consensus through alignment with the spec. There is a lot of refactoring we'd like to do to this package to make it easier to reason about and better align with the spec, so that implementing optimizations later will become much easier. See [ADR-30](https://github.com/tendermint/tendermint/blob/v0.26.0-dev0/docs/architecture/adr-030-consensus-refactor.md) for initial thinking there. I'm afraid we probably don't want to implement any optimizations until then, but happy to have you think through them and work on spec/proofs in the meantime. If you have thoughts on how we can improve the architecture of the code itself, that would also be much appreciated.

Thanks!

Issue #2587, opened 2018-10-09
Timestamp of first block is not validated


ebuchman:
We only start validating `block.Time` when `Height > 1`, because there is no commit to compute the median timestamp from for the first block. This means a faulty proposer could make the first block with whatever time they want.

Instead, we should require the timestamp of block 1 to match the genesis time.

I discovered this while refactoring the ValidateBlock tests to be table-driven while working on tests for #2560.

ebuchman:
Merged in #2594

PR #2594, opened 2018-10-10
state: require block.Time of the fist block to be genesis time

melekes:
Refs #2587 

* [x] Wrote tests

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -118,10 +118,7 @@ func (state State) MakeBlock(
 
 	// Set time
 	if height == 1 {
-		block.Time = tmtime.Now()
-		if block.Time.Before(state.LastBlockTime) {
-			block.Time = state.LastBlockTime // state.LastBlockTime for height == 1 is genesis time
-		}
+		block.Time = state.LastBlockTime // genesis time
 	} else {
 		block.Time = MedianTime(commit, state.LastValidators)
 	}
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -27,6 +27,7 @@ BREAKING CHANGES:
 * Blockchain Protocol
   * [types] \#2459 `Vote`/`Proposal`/`Heartbeat` use amino encoding instead of JSON in `SignBytes`.
   * [types] \#2512 Remove the pubkey field from the validator hash
+  * [state] \#2587 require block.Time of the fist block to be genesis time
 
 * P2P Protocol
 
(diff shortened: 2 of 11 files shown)


melekes:
```
--- FAIL: TestStateLockNoPOL (0.29s)
	state_test.go:381: #### ONTO ROUND 1
	state_test.go:426: #### ONTO ROUND 2
	state_test.go:467: #### ONTO ROUND 3
panic: Timeout expired while waiting for NewTimeout event [recovered]
	panic: Timeout expired while waiting for NewTimeout event

goroutine 37416 [running]:
testing.tRunner.func1(0xc424828870)
panic(0xef3b80, 0xc42896f0e0)
github.com/tendermint/tendermint/consensus.ensureNewEvent(0xc4272e4720, 0x2faf080, 0x1047f16, 0x32)
	/go/src/github.com/tendermint/tendermint/consensus/common_test.go:334 +0x159
github.com/tendermint/tendermint/consensus.ensureNewTimeout(0xc4272e4720, 0x989680)
	/go/src/github.com/tendermint/tendermint/consensus/common_test.go:352 +0x5c
github.com/tendermint/tendermint/consensus.TestStateLockNoPOL(0xc424828870)
	/go/src/github.com/tendermint/tendermint/consensus/state_test.go:487 +0x19ec
testing.tRunner(0xc424828870, 0x117d4d8)
created by testing.(*T).Run
FAIL	github.com/tendermint/tendermint/consensus	26.267s
Exited with code 1
```

state/state.go
```diff
@@ -118,10 +118,7 @@ func (state State) MakeBlock(
 
 	// Set time
 	if height == 1 {
-		block.Time = tmtime.Now()
-		if block.Time.Before(state.LastBlockTime) {
-			block.Time = state.LastBlockTime // state.LastBlockTime for height == 1 is genesis time
-		}
+		block.Time = state.LastBlockTime // genesis time
```
milosevic: This modification lead to generation of the same blocks for height 1, which leads to some consensus tests failing as they assumed that whenever MakeBlock is created, it will return new block. As for height 1 we don't have a way to constrain time, probably it is probably acceptable behaviour to have idempotent behaviour for MakeBlock. We just need to fix failing tests to modify returned block in case new block is needed.

melekes: Thank you 🍏 

Why don't add `round` to block's header? Or we could start from `height=2` as an alternative solution.

milosevic: Maybe better option is to modify failing test. For example https://github.com/tendermint/tendermint/pull/2607.

PR #2598, opened 2018-10-10
[R4R] Fixed sized and reordered fields for Vote/Proposal/Heartbeat SignBytes

### LINKED ISSUE #2545: Use fixed lengths in and finalize sign bytes
In #1622 we resolved to change sign bytes to amino, and to use fixed size fields so HSMs can easily read height and round from the data, and thus prevent double signing.

We switched to amino in #2459. We still need to make the values fixed length.

See the draft ADR-024 from PR #2445 for much more details

Liamsi:
resolves #2545

* [x] Wrote tests


types/vote_test.go
```diff
@@ -53,6 +53,37 @@ func TestVoteSignable(t *testing.T) {
...
+			&Vote{Height: 1, Round: 1},
+			"",
+			[]byte{
+				0x1d,                                   // total length
+				0x9,                                    // (field_number << 3) | wire_type
+				0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, // height
+				0x11,                                   // (field_number << 3) | wire_type
+				0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, // round
+				// remaining fields:
+				0x22, 0x9, 0x9, 0x0, 0x9, 0x6e, 0x88, 0xf1, 0xff, 0xff, 0xff},
```
Liamsi: @jleni do this kind of tests tests suit your need for test-cases (as mentioned in https://github.com/tendermint/tendermint/issues/2545#issuecomment-427678845)?

jleni: That is great. What about version number? 
It would be good that it is there too. So breaking changes can be signaled and the HSM can reject signing.

jleni: btw, what checks would you recommend in HSM code? I am planning to check at least:

- total length == blob size
- version == whatever we agree on
- field_number == 0x09 (or the corresponding value)
- block number >= 0
- field_number == 0x11 (or the corresponding value)
- round number >= 0

Liamsi: Thanks! Let me sync up with @ebuchman or someone from the tm-core team about the version field.
The checks looks reasonable! I'll think about it and let you know if there should be more checks.

Liamsi: So, the story around the version field is that the version isn't fully integrated yet. I'll add the field to `CanocialVote` and could add it manually for a few test cases so it would unblock your work on the HSM parsing side. 
@ebuchman does it make sense to change the `Signable` interface to take a version before we integrating the "Block Protocol Version"?

jleni: That is fine, what we have now is enough for me to continue. I can now use these offsets and layout, complete things and later adjust.
We will eventually need this field anyway.
Maybe it makes sense that you add it and set to zero for now, until the other code starts using it but I don't actually need this anyway.

Liamsi: Did you notice that fields are skipped whenever they are the default value? Does that pose a problem for you? I think you checks would look slightly different, e.g. if the first field is missing field_number == 0x09 would be false but field_number == 0x11 would be true (in case the second field isn't zero too etc).

ebuchman: Does that behaviour (skipping defaults) have an analog in proto3? Or is it amino specific?

jleni: @Liamsi Yes, I am aware of that. No problem, I was planning to consider this case
@ebuchman AFAIK, it is the same in protobuf3

Liamsi: > Does that behaviour (skipping defaults) have an analog in proto3? Or is it amino specific?

@jleni is right. It is the default behaviour in proto3, too. In amino this ought to be configurable via the `write_empty` tag, though: https://github.com/tendermint/go-amino/blob/faa6e731944e2b7b6a46ad202902851e8ce85bee/binary-encode.go#L404-L408
Surprisingly, this doesn't work here. If this isn't a problem while parsing, let's keep it that way. I'll investigate and open an issue in amino if necessary.

Liamsi:
@jleni @ebuchman I think this is ready for review. Let me know if we need to add, or reorder anything. The `Version` field currently isn't set anywhere. It is merely a placeholder. As soon as we fully integrated the "Block Version", we can add it as a param to `SignBytes`.

types/canonical.go
```diff
@@ -23,32 +23,32 @@ type CanonicalPartSetHeader struct {
 }
 
 type CanonicalProposal struct {
+	Version          uint64                 `json:"version" binary:"fixed64"`
+	Height           int64                  `json:"height" binary:"fixed64"`
+	Round            int64                  `json:"round" binary:"fixed64"`
+	POLRound         int64                  `json:"pol_round" binary:"fixed64"`
+	Timestamp        time.Time              `json:"timestamp"`
 	ChainID          string                 `json:"@chain_id"`
```
ebuchman: Should we be consistent with putting the ChainID at the end?

Also, do we even JSON marshal this ever? If not, should we drop the JSON tags too?

types/canonical.go
```diff
@@ -23,32 +23,32 @@ type CanonicalPartSetHeader struct {
...
-	Type             string                 `json:"@type"`
```
ebuchman: I guess one problem with removing `type` is that we have to be careful that eg. the encoding of a CanonicalProposal could not be mistaken for that of a CanonicalVote. 

Are we certain that `marhsalBinary(x) != marshalBinary(y)` for `var x CanonicalVote` and `var y CanonicalProposal` ?

types/canonical.go
Liamsi: Good point. Moved to the end. 

> Also, do we even JSON marshal this ever? If not, should we drop the JSON tags too?

No, I kept them in case someone wants to marshal them to JSON though. But yeah, it makes sense to delete those JSON tags and re-add them if we see a need.

types/canonical.go
```diff
@@ -72,36 +72,33 @@ func CanonicalizePartSetHeader(psh PartSetHeader) CanonicalPartSetHeader {
 
 func CanonicalizeProposal(chainID string, proposal *Proposal) CanonicalProposal {
 	return CanonicalProposal{
-		ChainID:          chainID,
-		Type:             "proposal",
-		BlockPartsHeader: CanonicalizePartSetHeader(proposal.BlockPartsHeader),
 		Height:           proposal.Height,
+		Round:            int64(proposal.Round), // cast int->int64 to make amino encode it fixed64 (does not work for int)
```
melekes: Should we standardize Round's type? make it `uint16` or `int32`?

types/vote_test.go
```diff
@@ -53,6 +53,70 @@ func TestVoteSignable(t *testing.T) {
...
+	voteWithVersion.Version = 123
+
+	tests := []struct {
+		canonicalVote CanonicalVote
+		want          []byte
+	}{
+		{
+			CanonicalizeVote("", &Vote{}),
+			// NOTE: Height and Round are skipped here. This case needs to be considered while parsing.
+			[]byte{0xb, 0x2a, 0x9, 0x9, 0x0, 0x9, 0x6e, 0x88, 0xf1, 0xff, 0xff, 0xff},
```
melekes: how do I get this something changes?

types/canonical.go
Liamsi: You are right. They would actually be the same if for instance the first 3 fields are equal (because of the same order and nothing else to differentiate between the two). I'll add a test-case and add back the type.

types/vote_test.go
Liamsi: It's just the output of `MarshalBinary`. The test serves two purposes: 1) mainly documentation for people wanting to parse this without having to fully implement amino/proto3 (like @jleni) 2) get notified if the encoding changes for whatever reason.

types/canonical.go
```diff
@@ -27,6 +27,7 @@ type CanonicalProposal struct {
 	Height           int64  `binary:"fixed64"`
 	Round            int64  `binary:"fixed64"`
 	POLRound         int64  `binary:"fixed64"`
+	Type             string
```
ebuchman: https://github.com/tendermint/tendermint/pull/2598#discussion_r224906032

types/canonical.go
ebuchman: Can you open a new issue with the case for it?

types/canonical.go
Liamsi: Thats much better. Thanks for the thorough review and the great suggestions ❤️  
This change will touch quite a lot of files though. I'll rebase (and resolve conflicts) and this should be ready for another review.

types/canonical.go
```diff
@@ -13,44 +13,45 @@ import (
...
-	BlockPartsHeader CanonicalPartSetHeader `json:"block_parts_header"`
-	Height           int64                  `json:"height"`
-	POLBlockID       CanonicalBlockID       `json:"pol_block_id"`
-	POLRound         int                    `json:"pol_round"`
-	Round            int                    `json:"round"`
-	Timestamp        time.Time              `json:"timestamp"`
+	Version          uint64 `binary:"fixed64"`
+	Height           int64  `binary:"fixed64"`
+	Round            int64  `binary:"fixed64"`
+	POLRound         int64  `binary:"fixed64"`
```
ebuchman: Shouldn't it be

```
Round
Type
POLRound
```

types/canonical.go
```diff
@@ -13,44 +13,45 @@ import (
...
-	ChainID          string  `json:"@chain_id"`
-	Type             string  `json:"@type"`
-	Height           int64   `json:"height"`
-	Round            int     `json:"round"`
-	Sequence         int     `json:"sequence"`
-	ValidatorAddress Address `json:"validator_address"`
-	ValidatorIndex   int     `json:"validator_index"`
+	Version          uint64 `binary:"fixed64"`
+	Height           int64  `binary:"fixed64"`
+	Round            int    `binary:"fixed64"`
```
ebuchman: add `Type`

types/signed_msg_type.go
```diff
@@ -0,0 +1,22 @@
+package types
+
+type SignedMsgType byte
+
+const (
+	// Types of votes
+	PrevoteType   SignedMsgType = 0x01
+	PrecommitType SignedMsgType = 0x02
+	// To separate (canonicalized) Vote and Proposal messages:
+	ProposalType SignedMsgType = 0x03
```
ebuchman: Let's distinguish categories of msg type like "vote-associated", "proposal-associated", "heartbeat-associated". So eg.

```
        // Vote types
	PrevoteType   SignedMsgType = 0x01
	PrecommitType SignedMsgType = 0x02

	// Proposal Types
	ProposalType SignedMsgType = 0x10

        // Heartbeat Types
        HeartbeatType SignedMsgType = 0x20
        ExampleNewHeartbeatType SignedMsgType = 0x21

```

ebuchman:
FYI: https://github.com/tendermint/tendermint/issues/2626

Sorry I didn't open that sooner

Liamsi:
> Thanks Ismail. I'll make the final adjustments.

Just saw this. I've added a commit that uses `SignedMsgType` in `Vote` and `CanonicalVote` to avoid casting to `byte` everywhere. I'm going to bed now. So no further commits from my side until tomorrow.

ebuchman:
Oh, I changed the order of some fields in that commit and the tests didn't break. Could that be an issue?

Also, is `byte` fixed length?

Liamsi:
> Oh, I changed the order of some fields in that commit and the tests didn't break. Could that be an issue?

That's probably because those where not included in the tests (because they are skipped if they are empty).  The tests include only those fields which are relevant for the HSM (and currently only for Votes as requested per @jleni) and those that are not skipped by default (`Timestamp`).

> Also, is byte fixed length?

Yes, amino treats `byte` as `uint8` (calling `EncodeUint8`).

Issue #2603, opened 2018-10-11
Make types.Txs [][]byte instead of []Tx


ValarDragon:
I believe Tx and `[]byte` can be used interchangeably, but the same is not true for `[][]byte` and Txs. This causes some efficiency losses for copying between the two.

ebuchman:
This is a shame - I like having stronger types than []byte. @xla any thoughts on how this might be handled better? The challenge is to avoid having to do eg. https://github.com/tendermint/tendermint/pull/2635/files#diff-a9cbec9a4f9f0903f5e9d965be0e3d76R34

ebuchman:
Also fix:
- [ ] https://github.com/tendermint/tendermint/pull/2635/files#diff-a9cbec9a4f9f0903f5e9d965be0e3d76R34

marbar3778:
If preprocess Block is accepted into Tendermint this issue may not be applicable anymore as the key of the transaction will need to be provided by the application.

PR #2604, opened 2018-10-11
Reduce memory allocations within merkle.SimpleHashFromByteSlices

ValarDragon:
Do we want a benchmark for this? Seems like a clear improvement so I opted to not make one. 

* [ ] Updated all relevant documentation in docs - n/a ?
* [ ] Wrote tests - current tests cover it

### DIFF
--- a/crypto/merkle/simple_tree.go
+++ b/crypto/merkle/simple_tree.go
@@ -45,15 +45,20 @@ func SimpleHashFromMap(m map[string][]byte) []byte {
 
 // Expects hashes!
 func simpleHashFromHashes(hashes [][]byte) []byte {
-	// Recursive impl.
-	switch len(hashes) {
+	return simpleHashFromHashesInPlace(hashes, 0, len(hashes))
+}
+
+func simpleHashFromHashesInPlace(hashes [][]byte, start, end int) []byte {
+	// In place recursive impl.
+	numHashes := end - start
+	switch numHashes {
 	case 0:
 		return nil
 	case 1:
-		return hashes[0]
+		return hashes[start]
 	default:
-		left := simpleHashFromHashes(hashes[:(len(hashes)+1)/2])
-		right := simpleHashFromHashes(hashes[(len(hashes)+1)/2:])
+		left := simpleHashFromHashesInPlace(hashes, start, start+(numHashes+1)/2)
+		right := simpleHashFromHashesInPlace(hashes, start+(numHashes+1)/2, end)
 		return SimpleHashFromTwoHashes(left, right)
 	}
 }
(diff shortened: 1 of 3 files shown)


types/tx.go
```diff
@@ -31,18 +31,11 @@ type Txs []Tx
...
-	switch len(txs) {
-	case 0:
-		return nil
-	case 1:
-		return txs[0].Hash()
-	default:
-		left := Txs(txs[:(len(txs)+1)/2]).Hash()
-		right := Txs(txs[(len(txs)+1)/2:]).Hash()
-		return merkle.SimpleHashFromTwoHashes(left, right)
+	txBz := make([][]byte, len(txs))
```
melekes: can't we pass txs to merkle.SimpleHashFromByteSlices()? + cast before `merkle.SimpleHashFromByteSlices(([][]byte)(txs))`

ValarDragon: When I try that, i get the following `cannot convert txs (type Txs) to type [][]byte`. I'm hoping to fix this having to be the case with #2603

crypto/merkle/simple_tree.go
```diff
@@ -45,15 +45,20 @@ func SimpleHashFromMap(m map[string][]byte) []byte {
...
+	switch numHashes {
 	case 0:
 		return nil
 	case 1:
-		return hashes[0]
+		return hashes[start]
 	default:
-		left := simpleHashFromHashes(hashes[:(len(hashes)+1)/2])
-		right := simpleHashFromHashes(hashes[(len(hashes)+1)/2:])
+		left := simpleHashFromHashesInPlace(hashes, start, start+(numHashes+1)/2)
```
ebuchman: Unless I'm mistaken, this should have no real impact on memory, since slices don't copy, they just create new references. Was there some reason you thought this was important?

types/tx.go
ebuchman: isn't this adding an allocation?

crypto/merkle/simple_tree.go
ValarDragon: This is reducing the number of slice objects, which do incur memory overhead. Previously we were creating `n-1` such objects.

types/tx.go
ValarDragon: Yes it is. We can actually add a fairly straight forward api in merkle to remove this allocation as long there is a bytes method

crypto/merkle/simple_tree.go
ebuchman: The overhead being garbage collection? There's no new allocations, right?

ValarDragon:
Actually I think we can probably close this, and I can just implement an iterative / non-recursive method which would solve these concerns. @ebuchman is right, whether or not this is actually faster isn't that clear. More arguments in a recursive function call translates to more stack allocations per call, and its not clear whether or not the number of stack allocations for the new slice compensates for this.

ValarDragon:
Closed in favor of #2611

PR #2605, opened 2018-10-11
refactor nop_event_bus.go into event_bus.go

ValarDragon:
This is just to match the style of the rest of the codebase, and to
reduce the number of files in types.


ebuchman:
> IMHO there's nothing wrong with many files

Generally agree, but there's definitely too many files in `types` :P

Issue #2565, opened 2018-10-06
ConsensusParams.Evidence.MaxAge should be time, not height


ebuchman:
Currently, `ConsensusParams.Evidence.MaxAge` is the maximum age, in heights, of a valid piece of evidence. Ie. if its 1000, and we're at block 5000, only evidence since block 4000 will be considered valid.

Now that we have BFT time, we should change this to be time based rather than height based.

ebuchman:
It should be the same Timestamp format used in the header (ie. google's "WellKnownType.Timestamp" thing)

ValarDragon:
Oh theres a well known type for duration. I think we should use that instead. (Fundamentally we want a time duration, not a single point in time which is what timestamps are for)

https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/duration

melekes:
I made a PR earlier https://github.com/tendermint/tendermint/pull/2606, but closing it because of the https://github.com/tendermint/tendermint/issues/2653.

ebuchman:
It sounds like addressing #2653 properly will happy post launch and require larger discussion.

Meanwhile, we should probably adopt the hybrid approach - support both MaxAgeDuration and MaxAgeHeight and only reject evidence if it is both too old in height and duration. That way if validators lie about time, at least evidence will still be valid for some number of heights.

melekes:
> support both MaxAgeDuration and MaxAgeHeight and only reject evidence if it is both too old in height and duration.

What could happen is people will be setting MaxAgeDuration because it's more natural and leave MaxAgeHeight as is, which will lead to bad consequences if validators manipulate with time.

> That said, actually, we can probably leave this as is for now. Thanks.

Should we close this then?

melekes:
> Now that we have BFT time, we should change this to be time based rather than height based.

I'd vote for closing this since we're planning to switch back to "proposer's timestamp".

ebuchman:
We'd still want to change this I think, no?

marbar3778:
In the last core dev session, it seemed https://github.com/tendermint/tendermint/issues/2840, was in the discussion phase. So it may be beneficial to take the hybrid approach for this release if we do want to include this. Otherwise seems this and https://github.com/tendermint/tendermint/issues/2840 need to be discussed further before implementation.

ebuchman:
Yup, we were also discussing using a hybrid approach involving both time and height

marbar3778:
[...]

PR #2606, opened 2018-10-11
make `ConsensusParams.EvidenceParams.MaxAge` time.Duration

melekes:
Refs #2565

* [x] Wrote tests

TODO
* [x] Ensure we record LastBlockTime in PeerState
* [x] Make MaxAge `nonnullable`


consensus/reactor.go
```diff
@@ -948,14 +948,22 @@ func (ps *PeerState) ToJSON() ([]byte, error) {
...
+// the mempool to ensure peers are caught up before broadcasting new txs.
 func (ps *PeerState) GetHeight() int64 {
 	ps.mtx.Lock()
 	defer ps.mtx.Unlock()
 	return ps.PRS.Height
 }
 
+// GetLastBlockTime returns an atomic snapshot of the PeerRoundState's last
+// block time used by the evidence reactor when sending evidence.
+func (ps *PeerState) GetLastBlockTime() time.Time {
```
ValarDragon: Its not clear to me why they need the last block time? Why doesn't knowledge of the previous block hash suffice, if you already have the block?

CHANGELOG_PENDING.md
```diff
@@ -12,10 +12,17 @@ BREAKING CHANGES:
   * [rpc] \#2298 `/abci_query` takes `prove` argument instead of `trusted` and switches the default
     behaviour to `prove=false`
   * [privval] \#2459 Split `SocketPVMsg`s implementations into Request and Response, where the Response may contain a error message (returned by the remote signer)
+  * [genesis] \#2565 `consensus_params.evidence_params.max_age` is now `time.Duration` (nanosecond count)
+        ```json
+        "evidence_params": {
+          "max_age": "172800000000000"
```
ebuchman: oof. Is there not some way to make this more human readable?

CHANGELOG_PENDING.md
```diff
@@ -12,10 +12,17 @@ BREAKING CHANGES:
...
+        }
+        ```
 
 * Apps
   * [abci] \#2298 ResponseQuery.Proof is now a structured merkle.Proof, not just
     arbitrary bytes
+  * [abci] \#2565 InitChain `ConsensusParams.EvidenceParams.MaxAge` is now `google.protobuf.Duration` (see https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/duration)
```
ebuchman: ++ for marking the change in the multiple relevant sections :)

consensus/reactor.go
```diff
@@ -438,6 +437,7 @@ func makeRoundStepMessages(rs *cstypes.RoundState) (nrsMsg *NewRoundStepMessage,
 	if rs.Step == cstypes.RoundStepCommit {
 		csMsg = &CommitStepMessage{
 			Height:           rs.Height,
+			Time:             rs.CommitTime,
```
ebuchman: Why did we need to add this?

evidence/pool.go
```diff
@@ -128,19 +129,27 @@ func (evpool *EvidencePool) MarkEvidenceAsCommitted(height int64, evidence []typ
 
 	// remove committed evidence from the clist
 	maxAge := evpool.State().ConsensusParams.EvidenceParams.MaxAge
-	evpool.removeEvidence(height, maxAge, blockEvidenceMap)
-
+	evpool.removeEvidence(
+		height,
```
ebuchman: I think if we pass in the Header to MarkEvidenceAsCommitted instead of just the height (or maybe the timestamp, along with the height, without the full header), we can pass it in here too and avoid the need to get the LastBlockTime from the state ... then we don't need the Time field in those consensus msgs.

types/protobuf.go
```diff
@@ -209,13 +209,18 @@ func (pb2tm) ValidatorUpdates(vals []abci.ValidatorUpdate) ([]*Validator, error)
 }
 
 func (pb2tm) ConsensusParams(csp *abci.ConsensusParams) ConsensusParams {
-	return ConsensusParams{
-		BlockSize: BlockSize{
+	params := ConsensusParams{}
+
+	// we must defensively consider any structs may be nil
```
ebuchman: ++

consensus/reactor.go
melekes: > if you already have the block?

what if we don't?

ValarDragon: But if they don't have the latest block, shouldn't I be tracking by block height which blocks to send to them?

evidence/pool.go
melekes: True, but what about `checkSendEvidenceMessage` in evidenceReactor? where we decide whenever we should send an evidence to a peer based on its state.

melekes: another alternative is to pass blockstore and load block header?

```go
maxAge := evR.evpool.State().ConsensusParams.EvidenceParams.MaxAge
	// peerLastBlockTime := blockStore.LoadBlockMeta(peerHeight).Header.Time
	if peerLastBlockTime.Sub(ev.Time()) > maxAge {
		// evidence is too old, skip
		// NOTE: if evidence is too old for an honest peer,
		// then we're behind and either it already got committed or it never will!
		evR.Logger.Info("Not sending peer old evidence", "peerHeight", peerHeight, "evHeight", evHeight, "maxAge", maxAge, "peer", peer)
		return nil, false
	}
```

CHANGELOG_PENDING.md
melekes: https://stackoverflow.com/a/48051946/820520

melekes:
Maybe we should hold off merging this until https://github.com/tendermint/tendermint/issues/2653 is resolved.

Issue #1841, opened 2018-07-01
types: header.Hash() should use ordered list instead of SimpleHashFromMap


ebuchman:
The hash of the header is a merkle root of all its fields.

The fields are loaded into a map, using the field names as their key.

Though actually the key names don't 100% match. We should fix this so its a perfect match.

See https://github.com/tendermint/tendermint/blob/master/types/block.go#L207

Eg. key `Validators` for field `ValidatorsHash`

ebuchman:
We should reconsider whether we need to amino encode things all over the types package, like the header fields. Note we no longer amino encode tx byte arrays

ebuchman:
Also worth considering if we really need a merkle tree of the header and document either way the why

ValarDragon:
Its probably valuable to keep this merkelized, since as a lite client you may not want to store all the headers, but you may need to query state at a header whose hash you stored. Until we get BLS, the lite client probably wouldn't want to have to download the LastCommit with all of the signatures on it. The only downside of merkelization here is basically increased computation. Thats not really a big deal though for full nodes. For lite clients, if they wanted to be able to do anything interesting, they'd already need support for this level of hashing. Hence I think it makes sense to merkelize it.

ebuchman:
> the lite client probably wouldn't want to have to download the LastCommit with all of the signatures on it.

How does merklizing the header or not impact the lite client needing to download these sigs?

ValarDragon:
The point of merkilization would be that they can just have a single hash stored locally. (Either before or after verifying the header) then the question is how much overhead is there in going from hash to getting the part of the header you want. If you don't want the signatures on the previous commit, then merkelizing does save space. (That's half a kilobyte you don't have to download)

ebuchman:
So @mossid fixed the naming in https://github.com/tendermint/tendermint/pull/2610 but it actually shouldn't matter because the SimpleHashFromMap sorts the fields by key before building the merkle tree, so adding `Hash` to field names doesn't change anything. So the only thing here that would cause a change is `Results -> LastResults`. So we need to document that as a breaking change.

That said, I'm not sure this key-based sorting is the best approach, because it locks us to these field names rather than their order. Thoughts?

ValarDragon:
Agreed, I dislike the key names as determining sorting other vs just having an implicit order. (e.g. using `HashFromByteSlices`)

PR #2610, opened 2018-10-11
[R4R] Standardize header.Hash() for Merklizing

mossid:
* [ ] Wrote tests

Closes: #1841

### DIFF
--- a/types/block.go
+++ b/types/block.go
@@ -290,21 +290,21 @@ func (h *Header) Hash() cmn.HexBytes {
 		return nil
 	}
 	return merkle.SimpleHashFromMap(map[string][]byte{
-		"ChainID":        cdcEncode(h.ChainID),
-		"Height":         cdcEncode(h.Height),
-		"Time":           cdcEncode(h.Time),
-		"NumTxs":         cdcEncode(h.NumTxs),
-		"TotalTxs":       cdcEncode(h.TotalTxs),
-		"LastBlockID":    cdcEncode(h.LastBlockID),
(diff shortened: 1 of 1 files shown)


ValarDragon:
I don't think we've really finalized on the way to do this in the issue.

types/block.go
```diff
@@ -290,21 +290,21 @@ func (h *Header) Hash() cmn.HexBytes {
...
-		"Results":        cdcEncode(h.LastResultsHash),
-		"Evidence":       cdcEncode(h.EvidenceHash),
-		"Proposer":       cdcEncode(h.ProposerAddress),
+		"ChainID":            cdcEncode(h.ChainID),
+		"Height":             cdcEncode(h.Height),
+		"Time":               cdcEncode(h.Time),
+		"NumTxs":             cdcEncode(h.NumTxs),
+		"TotalTxs":           cdcEncode(h.TotalTxs),
+		"LastBlockID":        cdcEncode(h.LastBlockID),
+		"LastCommitHash":     cdcEncode(h.LastCommitHash),
```
ebuchman: Note this actually shouldn't matter because the SimpleHashFromMap sorts the fields by key before building the merkle tree. So the only thing here that would cause a change is `Results -> LastResults`. So we need to document that as a breaking change as well.

That said, I'm not sure this key-based sorting is the best approach, because it locks us to these field names rather than their order. Thoughts?

ebuchman: Copied to https://github.com/tendermint/tendermint/issues/1841#issuecomment-430001613

PR #2621, opened 2018-10-12
fix fastsync may stuck by a wrong block

goolAdapter:
ref #2457 

* [x] Wrote tests


blockchain/pool.go
```diff
@@ -292,7 +293,7 @@ func (pool *BlockPool) RemovePeer(peerID p2p.ID) {
 func (pool *BlockPool) removePeer(peerID p2p.ID) {
 	for _, requester := range pool.requesters {
 		if requester.getPeerID() == peerID {
-			requester.redo()
+			requester.redo(peerID)
```
milosevic: As we are removing peer at this point, it seems like there is no risk of requesting a block from the same peer. Do we really need to pass peerID on a channel?

blockchain/pool.go
```diff
@@ -453,7 +458,7 @@ type bpRequester struct {
 	pool       *BlockPool
 	height     int64
 	gotBlockCh chan struct{}
-	redoCh     chan struct{}
+	redoCh     chan p2p.ID //redo may send multitime, add peerId to identify repeat
```
milosevic: As we only remember last requester, I don't see how this helps. Maybe it would be useful to remember what are peers we already asked block for in case bad peer reconnect, but I am not sure that this suggestion address that case.

blockchain/pool.go
milosevic: I noticed that we don't update pool.maxPeerHeight after we remove peer. Furthermore, if we connect to a peer that reports very high height, we will update pool.MaxPeerHeight so isCaughtUp will never evaluates to true. @ebuchman Have I overlooked something?

blockchain/reactor_test.go
```diff
@@ -1,72 +1,149 @@
...
+	sort.Sort(types.PrivValidatorsByAddress(privValidators))
+
+	return &types.GenesisDoc{
+		GenesisTime: tmtime.Now(),
+		ChainID:     config.ChainID(),
+		Validators:  validators,
+	}, privValidators
+}
+
+func makeVote(header *types.Header, blockID types.BlockID, valset *types.ValidatorSet, privVal types.PrivValidator) *types.Vote {
```
milosevic: Can you use signVote from consensus/common_test instead?

blockchain/pool.go
```diff
@@ -252,7 +252,8 @@ func (pool *BlockPool) AddBlock(peerID p2p.ID, block *types.Block, blockSize int
 			peer.decrPending(blockSize)
 		}
 	} else {
-		// Bad peer?
```
goolAdapter: @milosevic , peerID is use to make sure here is an Bad peer. 
previously， normal logic can meet here， because redo was send message to chan，and deal it delayed，so  redo may apply to an good peer.

blockchain/pool.go
```diff
@@ -252,7 +252,8 @@ func (pool *BlockPool) AddBlock(peerID p2p.ID, block *types.Block, blockSize int
...
+		pool.Logger.Info("invalid peer", "peer", peerID, "blockHeight", block.Height)
```
milosevic: I am still not sure this is always sign of a bad peer. Could it be that this happens due to message retransmission as sending peer had transient crash failure?We probably need better error handling in setBlock to detect the nature of the error.

ebuchman: If the peer had a crash failure and restarted, it should have cleared all reactor state and started again, so it shouldn't be sending us things unless we explicitly asked for it.

ebuchman:
Replaced with https://github.com/tendermint/tendermint/pull/2731 so I could make some minor fixes.

Thanks!

Issue #2468, opened 2018-09-21
Implement Protocol Versions [ADR 016]


ebuchman:
[ADR-016](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-016-protocol-versions.md)

Subsumes #1627 and #1135 

- [x] Add Version to Header and State - #2644
- [x] Add Version to NodeInfo - https://github.com/tendermint/tendermint/pull/2654
- [x] Remove Version from CanonicalXxx -  #2666
- [ ] Add Version to ConsensusParams so apps can update the App version
- [x] Add Version to ResponseInfo so apps can set initial App version during handshake - #2662
- [x] Update ADR to match - #2667

ValarDragon:
`Add Version to ResponseInfo so apps can set initial App version during handshake` 

We need a way for also upgrading app version mid-execution, right? We earlier proposed making clients have a conditional for switching the app version logic on a certain block height. But at that blockheight their app version should change as well. I think we need another way to communicate that update, since it wouldn't be repeating the node <-> app handshake.

ebuchman:
>  Add Version to ConsensusParams so apps can update the App version

^^

PR #2627, opened 2018-10-12
[WIP] p2p: node info sub-structs

ebuchman:
Copies the NodeInfo over from https://github.com/tendermint/tendermint/pull/2528 and updates it to implement the NodeInfo interface from #2556.

Part of the work from https://github.com/tendermint/tendermint/issues/2468.

Though might be better for the new NodeInfo to be less disruptive, and just add a version struct field for now?

Still need to wire it up and consider how we might support both the DefaultNodeInfo and the new NodeInfoV5

* [ ] Updated all relevant documentation in doc.s
* [ ] Wrote tests

### DIFF
--- a/p2p/node_info.go
+++ b/p2p/node_info.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 
 	cmn "github.com/tendermint/tendermint/libs/common"
+
+	"github.com/tendermint/tendermint/version"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


ebuchman:
I'm wondering if it might be better to, instead of going with the design here right away, just:

- add a Version struct to the top of the NodeInfo containing P2P, Block, and App protocol versions
- change the existing `Version string` to `SoftwareVersion string` but otherwise leave as is
- remove all current validation on versions
- remove the version info from DefaultNodeInfoOther (ie. so it just contains TxIndex and RPCAddress)

Thoughts?

ebuchman:
Closing for a new PR which implements the above comment. We can perhaps make a larger breaking change like this later

ebuchman:
Replaced by https://github.com/tendermint/tendermint/pull/2654, for now

PR #2635, opened 2018-10-14
Make txs and evidencelist use merkle.SimpleHashFromBytes to create hash

### LINKED ISSUE #2612: merkle: Create a `SimpleHashFromByters`
Currently txs and evidence re-implement merkelization to avoid extra allocations. We can DRY this by making another method for merkelization which accepts an array of byters. This simplifies implementation of RFC 6962. (ref #1892)

ValarDragon:
This is a performance regression, but will also spare the types directory
from knowing about RFC 6962, which is a more correct abstraction. For txs
this performance hit will be fixed soon with #2603. For evidence, the
performance impact is negligible due to it being capped at a small number.

Closes #2612 

* [x] Wrote tests - current tests cover this

### DIFF
--- a/types/block.go
+++ b/types/block.go
@@ -576,7 +576,6 @@ func (sh SignedHeader) StringIndented(indent string) string {
 		indent, sh.Header.StringIndented(indent+"  "),
 		indent, sh.Commit.StringIndented(indent+"  "),
 		indent)
-	return ""
 }
 
 //-----------------------------------------------------------------------------
@@ -660,7 +659,6 @@ func (data *EvidenceData) StringIndented(indent string) string {
 %s}#%v`,
 		indent, strings.Join(evStrings, "\n"+indent+"  "),
 		indent, data.hash)
-	return ""
 }
 
 //--------------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


types/block.go
```diff
@@ -660,7 +659,6 @@ func (data *EvidenceData) StringIndented(indent string) string {
 %s}#%v`,
 		indent, strings.Join(evStrings, "\n"+indent+"  "),
 		indent, data.hash)
-	return ""
```
ValarDragon: My local lint was failing with this being here. Figured it wasn't worth making a separate PR to fix this.

types/evidence.go
```diff
@@ -55,6 +55,7 @@ func (err *ErrEvidenceOverflow) Error() string {
 type Evidence interface {
 	Height() int64                                     // height of the equivocation
 	Address() []byte                                   // address of the equivocating validator
+	Bytes() []byte                                     // bytes which compromise the evidence
```
ebuchman: `compromise -> comprise` ... I hope this method doesn't compromise the evidence :P

ValarDragon: LOL

PR #2636, opened 2018-10-14
Add ValidatorPubkeyTypes as a consensus param

ValarDragon:
Ref #2414

* [x] Updated all relevant documentation in docs - updated all I'm aware of. 
* [x] Wrote tests


ValarDragon:
Should `ValidatorPubkeyTypes` instead be `PubkeyTypes` as a sub-component of `ValidatorParams`?

types/params.go
```diff
@@ -17,8 +18,9 @@ const (
 // ConsensusParams contains consensus critical parameters that determine the
 // validity of blocks.
 type ConsensusParams struct {
-	BlockSize      `json:"block_size_params"`
-	EvidenceParams `json:"evidence_params"`
+	BlockSize            `json:"block_size_params"`
```
ebuchman: Thinking about this, we can probably drop the `Params` and `_params` suffixes from these fields.

Also I wonder if the ConsensusParams is a good place for `Version` to live? Then we don't need to add anything to EndBlock for the app to update the versions, its just part of the ConsensusParams.

As for this PR, maybe we should have a new struct to hold the pubkey types? Could be called `Signatures` or `Cryptography` or something ?

types/params.go
```diff
@@ -55,6 +58,12 @@ func DefaultEvidenceParams() EvidenceParams {
 	}
 }
 
+// DefaultValidatorPubkeyTypes Params returns a default set of pubkey types to
+// be accepted for validators.
```
ebuchman: Comment should tell us what the default is

types/params.go
```diff
@@ -106,5 +140,8 @@ func (params ConsensusParams) Update(params2 *abci.ConsensusParams) ConsensusPar
 	if params2.EvidenceParams != nil {
 		res.EvidenceParams.MaxAge = params2.EvidenceParams.MaxAge
 	}
+	if params2.ValidatorPubkeyTypes != nil {
```
ebuchman: Works for structs but not sure if we might just end up with a non-nil empty list here?

ValarDragon:
My plan was to add enforcement in a second PR. (Primarily because I was unsure what function should be aware of the consensus params when getting the validator set updates)

types/params.go
ValarDragon: This makes a lot of sense for app version imo.

Re new struct, sure I think Validators makes sense! I do think having the params suffix is nice for disambiguation when your looking at types.

ebuchman: Cool. Let's go with `Validators`.

Yeh, `Params` makes sense for the types, but it's redundant for the field names. I guess here it's all embedded structs, but maybe it shouldn't be, and then we can have proper field names and struct type name distinctions.

ValarDragon:
Is this good to merge? I'd like to get it in soon before we make another .proto change and I have to fix the merge conflict

types/params.go
```diff
@@ -77,6 +90,10 @@ func (params *ConsensusParams) Validate() error {
 			params.EvidenceParams.MaxAge)
 	}
 
+	if len(params.Validator.ValidatorPubkeyTypes) == 0 {
```
ebuchman: Should we also validate that the types given are known?

types/params.go
```diff
@@ -94,6 +111,27 @@ func (params *ConsensusParams) Hash() []byte {
 	return hasher.Sum(nil)
 }
 
+func (params *ConsensusParams) Equals(params2 *ConsensusParams) bool {
+	if params.BlockSize == params2.BlockSize &&
```
ebuchman: We can just return this conjunction instead of using the conditional

types/protobuf.go
```diff
@@ -24,6 +24,12 @@ const (
 	ABCIPubKeyTypeSecp256k1 = "secp256k1"
 )
 
+// TODO: Make non-global by allowing for registration of more pubkey types
```
melekes: is there an associated issue?

types/params.go
```diff
@@ -94,6 +121,24 @@ func (params *ConsensusParams) Hash() []byte {
 	return hasher.Sum(nil)
 }
 
+func (params *ConsensusParams) Equals(params2 *ConsensusParams) bool {
+	return params.BlockSize == params2.BlockSize &&
+		params.Evidence == params2.Evidence &&
+		stringSliceEqual(params.Validator.PubKeyTypes, params2.Validator.PubKeyTypes)
+}
+
+func stringSliceEqual(a, b []string) bool {
```
melekes: move to cmn#string.go?

types/params.go
```diff
@@ -108,8 +153,11 @@ func (params ConsensusParams) Update(params2 *abci.ConsensusParams) ConsensusPar
 		res.BlockSize.MaxBytes = params2.BlockSize.MaxBytes
 		res.BlockSize.MaxGas = params2.BlockSize.MaxGas
 	}
-	if params2.EvidenceParams != nil {
-		res.EvidenceParams.MaxAge = params2.EvidenceParams.MaxAge
+	if params2.Evidence != nil {
+		res.Evidence.MaxAge = params2.Evidence.MaxAge
+	}
+	if params2.Validator != nil {
+		res.Validator.PubKeyTypes = params2.Validator.PubKeyTypes
```
ebuchman: Should we copy here? 

Also I wonder if it makes sense to use Amino naming in the types#ConsensusParams but ABCI naming in the abci#ConsensusParams. Is that crazy? Or maybe we put an empty PubKey of the respective type in the types#ConsensusParams instead of strings.

ValarDragon: Good idea, it probably is best to copy.

Issue #1811, opened 2018-06-24
[Pubsub] failure to drain channel after unsubsribe can halt entire pubsub


silasdavis:
Because of the blocking send to every subscriber channel: https://github.com/tendermint/tendermint/blob/master/libs/pubsub/pubsub.go#L347:

```go
func (state *state) send(msg interface{}, tags TagMap) {
	for q, clientToChannelMap := range state.queries {
		if q.Matches(tags) {
			for _, ch := range clientToChannelMap {
				ch <- msg
			}
		}
	}
}
```

If one subscriber blocks on receiving a message then the entire pubsub server stops. In some settings this might seem like bad behaviour for a pub sub system (i.e. some subscribers starving others). In Tendermint maybe this is needed in some places because a non-blocking send might mean some component misses vital messages. A non-blocking send would fix this:

```go
func (state *state) send(msg interface{}, tags TagMap) {
	for q, clientToChannelMap := range state.queries {
		if q.Matches(tags) {
			for _, ch := range clientToChannelMap {
				select {
				case ch <- msg:
				default:
					continue
				}
			}
		}
	}
}
```

Currently a particular bad consequence of this behaviour is as in the following:

``go
	s := pubsub.Server{}
	s.Subscribe(ctx, sub, qry, out)
	defer s.UnsubscribeAll(ctx, sub)
	for msg := range out {
		// Do stuff
		if err != nil {
			// Return from error doing stuff hoping the defer unsubscribe cleans things up for you
			return err
		}
	}
```
In this scenario we exit from our subscribing loop in an error and unsubscribing alone is not enough to clean up. In fact if we get any more messages we will freeze up the entire pubsub server. Instead cleanup looks like:
```go
	s.Subscribe(ctx, sub, qry, out)
	defer func() {
		s.UnsubscribeAll(ctx, sub)
		for range out {
			// drain out to make sure we don't block
		}
	}() 
	for msg := range out {
		// Do stuff
		if err != nil {
			// Return from error doing stuff hoping the defer unsubscribe cleans things up for you
			return err
		}
	}
```

This feels like having to know too about the internals of pubsub and is easy to forget since it may not immediately cause a problem until some unrelated subscriptions stop working.

[...]

PR #2641, opened 2018-10-15
[pubsub] document design shortcomings

melekes:
Refs https://github.com/tendermint/tendermint/issues/1811#issuecomment-427825250


melekes:
> This seems like something we should absolutely fix, but I believe were planning on changing most of the current implementation anyways?

not sure it can be fixed without switching to non-blocking send.

Issue #2625, opened 2018-10-12
#2540 - +2/3-precommit-any liveness follow up


ebuchman:
#2540 fixes #1745 and improves alignment between code and spec.

There are some minor suggestions and confirmations of logic in the reviews there that we should address:

- update godoc for enterPrevote and enterPrecommit that explains when these functions are called
- introduce ensurePrevote and ensurePrecommit test methods instead of ensureVote

ebuchman:
Fixed in #2642

PR #2642, opened 2018-10-15
Align propose phase with spec

milosevic:
Fix https://github.com/tendermint/tendermint/issues/2529 and https://github.com/tendermint/tendermint/issues/2625.

### DIFF
--- a/consensus/state_test.go
+++ b/consensus/state_test.go
@@ -214,16 +214,16 @@ func TestStateBadProposal(t *testing.T) {
 	ensureNewProposal(proposalCh, height, round)
 
 	// wait for prevote
-	ensureVote(voteCh, height, round, types.PrevoteType)
+	ensurePrevote(voteCh, height, round)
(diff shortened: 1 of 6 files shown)


consensus/state_test.go
```diff
@@ -873,11 +873,97 @@ func TestStateLockPOLSafety2(t *testing.T) {
...
+	round = round + 2 // moving to the next round
+
+	ensureNewRound(newRoundCh, height, round)
+	ensureNewTimeout(timeoutWaitCh, height, round, cs1.config.TimeoutPrecommit.Nanoseconds())
+
+	round = round + 1 // moving to the next round
+
+	ensureNewRound(newRoundCh, height, round)
+
+	t.Log("### ONTO ROUND 4")
```
melekes: where's the third round? (`ONTO ROUND 3`)

milosevic: It happens at line 951; I forgot to add log statement.

PR #2644, opened 2018-10-16
ADR-016: Add versions to Block and State

ebuchman:
Implements part of ADR-016 - #2468.

- Version structs to block.Header and State.
- the change to the block.Header bumps all proto3 fields forward by one and increases the max size of blocks.
- this does not yet include the ability for apps to signal or upgrade their protocol version.
- does not include changes to P2P. See https://github.com/tendermint/tendermint/pull/2627 for first draft there

* [x] Wrote tests


types/block.go
```diff
@@ -538,6 +543,9 @@ func (sh SignedHeader) ValidateBasic(chainID string) error {
 	if sh.Commit == nil {
 		return errors.New("SignedHeader missing commit (precommit votes).")
 	}
+
+	// TODO: Check Version?
```
Liamsi: Will this be addressed in a followup issue?

ebuchman: Whoops this is actually checked in the state.ValidateBlock so nothing to check here, can remove this comment

docs/spec/abci/abci.md
```diff
@@ -363,6 +364,15 @@ Commit are included in the header of the next block.
   - Provides the proposer of the current block, for use in proposer-based
     reward mechanisms.
 
+### Version
+
+- **Fields**:
+  - `Block (uint64)`: Protocol version of the blockchain data structures.
+  - `App (uint64)`: Protocol version of the application.
+- **Usage**:
+  - Block version should be static in the life of a blockchain.
```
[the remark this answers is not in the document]
ebuchman: That's the idea for now. We can probably relax it somewhat later but needs more work.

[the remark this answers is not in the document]
ValarDragon: I definitely think this can be relaxed #postlaunch. My intuition is that only changes to the merkle tree structure absolutely require going back to genesis files, from a block version pov. Header formats and Block Structures can be changed in a live-upgrade, but probably shouldn't be due to how it would break a ton of other things.

docs/spec/blockchain/blockchain.md
```diff
@@ -65,6 +67,19 @@ type Header struct {
 
 Further details on each of these fields is described below.
 
+## Version
+
+The `Version` contains the protocol version for the blockchain and the
+application as two `uint64` values:
```
melekes: no need to duplicate type information imho

```suggestion
application:
```

Issue #2372, opened 2018-09-12
Ubuntu bash installation root permission requirement.


Uzair1995:
While installing on Ubuntu it requires permission for the root access which is not required in my experience. You just need to replace GOPATH=/root/goApps with GOPATH =/home/{username}/goApps. This would remove the requirement of install it in the root, and the program will function as it is suppose to be.

Uzair1995:
Forgot to mention I was referencing this file:
https://github.com/tendermint/tendermint/blob/master/scripts/install/install_tendermint_ubuntu.sh

ValarDragon:
The reason for the root requirement was more so for installing make. That could be worked around though, e.g. by only installing it to a single user. Would you like to submit a PR?

Uzair1995:
Very well. I will create a pull request... Its just a small change btw.

melekes:
Merged to develop. Thanks!

PR #2647, opened 2018-10-16
 Ubuntu bash installation root permission requirement.

Uzair1995:
Install script changes for installing in ubuntu. See #2372

### DIFF
--- a/scripts/install/install_tendermint_ubuntu.sh
+++ b/scripts/install/install_tendermint_ubuntu.sh
@@ -22,12 +22,12 @@ curl -O https://storage.googleapis.com/golang/go1.10.linux-amd64.tar.gz
 tar -xvf go1.10.linux-amd64.tar.gz
 
 # move go binary and add to path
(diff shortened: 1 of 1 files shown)


scripts/install/install_tendermint_ubuntu.sh
```diff
@@ -22,12 +22,12 @@ curl -O https://storage.googleapis.com/golang/go1.10.linux-amd64.tar.gz
 tar -xvf go1.10.linux-amd64.tar.gz
 
 # move go binary and add to path
-mv go /usr/local
+sudo mv go /usr/local
```
melekes: this should not require sudo.

melekes: also, the above comment seems to be wrong. we're moving go folder, not binary to /usr/local, no?

melekes: I'll address in a separate PR

PR #2654, opened 2018-10-17
ADR-016: Add protocol Version to NodeInfo

ebuchman:
More from https://github.com/tendermint/tendermint/issues/2468

* [x] Wrote tests


p2p/node_info.go
```diff
@@ -124,36 +137,40 @@ func (info DefaultNodeInfo) ValidateBasic() error {
...
+
+	// Validate Other.
+	other := info.Other
+	txIndex := other.TxIndex
+	switch txIndex {
+	case "", "on", "off":
+	default:
+		return fmt.Errorf("info.Other.TxIndex should be either 'on' or 'off', got '%v'", txIndex)
+	}
+	// XXX: Should we be more strict about address formats?
```
Liamsi: Are they always IP4 addresses?

ebuchman: No actually, in theory they could be IPv6. We don't actually do anything with this RPCAddress though, just display it over RPC, so it doesn't matter too much here ...

p2p/node_info.go
```diff
@@ -124,36 +137,40 @@ func (info DefaultNodeInfo) ValidateBasic() error {
...
+		return fmt.Errorf("info.Moniker must be valid non-empty ASCII text without tabs, but got %v", info.Moniker)
+	}
+
+	// Validate Other.
+	other := info.Other
+	txIndex := other.TxIndex
+	switch txIndex {
+	case "", "on", "off":
+	default:
+		return fmt.Errorf("info.Other.TxIndex should be either 'on' or 'off', got '%v'", txIndex)
```
melekes: 'on' or 'off' or empty string

ebuchman: Thanks - bb173bcae29e1aa76cff0bf842822318d87fe77b

Issue #2099, opened 2018-07-28
crypto/random.go: Review MixEntropy


ValarDragon:
There are ways to improve our current implementation of crypto.MixEntropy(). 

- [x] We should switch the Stream cipher primitive we are using. I'd recommend using Chacha20. Theres less to fear about non-constant time implementations. (Though the golang implementation is constant time I believe) #postlaunch we may even want to consider increasing the number of internal rounds used here.
- [ ]  We should use a slower hash function than Sha2 here. (Or use iterated Sha2) The reason is not due to a fear that SHA2 will be broken. Rather its that it provides defense in the case where an attacker-controlled input is being mixed into the entropy pool, and this input can read the other inputs, but can't directly broadcast its readings to the rest of the world. So instead it grinds the hash (or whatever is used for combining) to have leading zeroes or some biasing. 
- [ ] We should increase the entropy gathering period for initially seeding the CSPRNG _significantly_. We're generating peoples private keys out of this, they can wait for the key to take a bit to generate. We can have a seperate method for generating random privkeys for testing. 
- [ ] We should add methods for people to easily increase the entropy gathering periods.

ValarDragon:
Theres another issue here. Currently we don't update the key in between encryptions. This means that system access would reveal all previous private keys / OS randomness under our current API usage. (The only mechanism to thwart this would be to call MixEntropy under the current API)

We should hash the stream cipher private key after each read.

ebuchman:
Do we actually need this at all ? Seems the only place it's currently used is:

```
func init() {
	gRandInfo = &randInfo{}
	gRandInfo.MixEntropy(randBytes(32)) // Init
}
```

But I'm not sure why we bother when we already are reading 32 random bytes?

Since MixEntropy is otherwise not used anywhere, I'd think we could move this out of pre-launch

ValarDragon:
MixEntropy seeds a custom csprng, which is then used to derive all private keys. (We use this instead of crypto/rand) Hence I think it should be launch. (At least increasing the entropy gathering periods should be)

ebuchman:
Why is it better than just using crypto/rand alone?

ValarDragon:
I'd be happy to just use crypto/rand alone. I kinda think this custom thing were doing is a little silly. It would make sense if we had API to actually incorporate user randomness into the key generation process. Iirc that is the eventual plan. At the moment it does feel like feature creep.

ValarDragon:
[...]

PR #2669, opened 2018-10-18
crypto: use stdlib crypto/rand. ref #2099

ebuchman:
Ref #2099 

* [ ] Wrote tests


crypto/random.go
```diff
@@ -26,6 +27,7 @@ func init() {
 	gRandInfo.MixEntropy(randBytes(32)) // Init
```
ValarDragon: I'd prefer if we commented out this as well, to mitigate confusion.

PR #2673, opened 2018-10-18
Bucky/changelog

ebuchman:
Updates to the changelog.

Also added a script for checking who has contributed to the release. This is still a relatively manual process, but we can probably almost completely automate it (except some email addresses return "null" and we just need to google for them). The idea is to include all contributors who we aren't actively paying for their contributions :)

Also we should try to maintain alphabetical order as we go. It's not a big deal but I'm possessed around release time to re-order entries alphabetically ...

TODO:
- linkify the issues
- add external contributor names to the changelog items they contributed

Question:
- Some contribs are super minor and don't get included in the changelog (like some docs fixes). Should the contributor still be mentioned? Maybe we should add these items to changelog anyways?

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -59,6 +59,8 @@ BREAKING CHANGES:
       `SignedMsgType` to enumerate.
   * [types] \#2512 Remove the pubkey field from the validator hash
   * [types] \#2644 Add Version struct to Header
+  * [types] \#2609 ConsensusParams.Hash() is the hash of the amino encoded
+    struct instead of the Merkle tree of the fields
   * [state] \#2587 Require block.Time of the fist block to be genesis time
   * [state] \#2644 Require block.Version to match state.Version
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,7 +1,23 @@
 # Pending
 
+## v0.26.0
+
+*October 19, 2018*
+
 Special thanks to external contributors on this release:
-@goolAdapter, @bradyjoestar
+@goolAdapter, @bradyjoestar, @HaoyangLiu, @Slamper, @james-ray,
+@overbool, @connorwstein, @phymbert, @Uzair1995
+
+This release is primarily about adding Version fields to various data structures,
+optimizing consensus messages for signing and verification in
+restricted environments (like HSMs and the Ethereum Virtual Machine), and
+aligning the consensus code with the [specification](https://arxiv.org/abs/1807.04938).
+It also includes our first take at a generalized merkle proof system.
+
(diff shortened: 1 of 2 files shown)


CHANGELOG_PENDING.md
```diff
@@ -1,19 +1,35 @@
 # Pending
 
+## v0.26.0
+
+*October 19, 2018*
+
 Special thanks to external contributors on this release:
-@goolAdapter, @bradyjoestar
+@goolAdapter, @bradyjoestar, @HaoyangLiu, @Slamper, @james-ray,
```
ValarDragon: in the spirit of defining sort order, should we also sort contributor names?

CHANGELOG_PENDING.md
```diff
@@ -63,24 +80,27 @@ IMPROVEMENTS:
 - [config] \#2232 Added ValidateBasic method, which performs basic checks
 - [crypto/ed25519] \#2558 Switch to use latest `golang.org/x/crypto` through our fork at
   github.com/tendermint/crypto
+- [crypto] \#2099 make random use chacha, and have forward secrecy of generated randomness
```
ValarDragon: This line should now be deleted per deprecating the old mix entropy.

CHANGELOG_PENDING.md
```diff
@@ -1,19 +1,35 @@
...
 Special thanks to external contributors on this release:
-@goolAdapter, @bradyjoestar
+@goolAdapter, @bradyjoestar, @HaoyangLiu, @Slamper, @james-ray,
+@overbool, @connorwstein, @phymbert, @Uzair1995
+
+This release is primarily about adding Version fields to various data structures,
+optimizing consensus messages for signing and verification in
+restricted environments (like HSMs and the Ethereum Virtual Machine), and
+aligning the consensus code with the [specification](https://arxiv.org/abs/1807.04938).
+It also includes our first take at a generalized merkle proof system.
```
ValarDragon: May be worth mentioning limiting total number of evidences per block? (Thats in the changelog, though I don't remember a PR switching to number instead of proportion)

ValarDragon:
> Also we should try to maintain alphabetical order as we go. It's not a big deal but I'm possessed around release time to re-order entries alphabetically ...

Sounds like another script to add to our changelog arsenal, lol

melekes:
> Some contribs are super minor and don't get included in the changelog (like some docs fixes). Should the contributor still be mentioned? Maybe we should add these items to changelog anyways?

We can mention the contributor, but we shouldn't add a changelog entry. If we want people to read it, we must keep its size under control.

CHANGELOG_PENDING.md
ebuchman: We haven't made that switch yet. Not sure it's a high level concern worth pointing out here.

Issue #2665, opened 2018-10-18
KMS handshake with Version


liamsi:
@ValarDragon suggested to remove the `Version` field from the Canonical(Vote/Proposal) altogether and let the KMS (instead of the HSM) make sure (via an initial handshake) that it doesn't sign any message while "talking to a different version". 

This would have the following advantages:
- less communication overhead when sending Votes/Proposals
- less bytes to sign inside the HSM 
- no need to change Vote/Proposal structs (or the Signable interface)

This would change the offsets though (cc @jleni) 
(related: https://github.com/tendermint/tendermint/issues/2545)

ebuchman:
Added "Remove Version from CanonicalXxx" to https://github.com/tendermint/tendermint/issues/2468. This issue can track performing a handshake between KMS and Tendermint and passing a version through there

Liamsi:
Can't this be closed? We probably do not need a handshake anymore but rather make sure that the messages passed to the kms have a message type the kms knows of (it should check if they are valid when decoding).

Liamsi:
Actually, from the messages exchanged between tendermint and the kms, only votes contain the `SignMsgType` as a field. The HSM module (or whatever provider behind signatory) which processes / signs the canoncialized version of all messages. Not sure if we should fail earlier if the schema changed or not. An easy solution to this would be to add a `Type` field to proposals (and heartbeats) too. Then, we basically "versioned" everything that is sent between tm <-> kms.

ebuchman:
So we send the KMS the non-canonicalized structs, which contain signer info (address, index), but not the ChainID.

We should align the Vote/Proposal fields with the Canonicalized fields so eg. Proposal will have SignMsgType, but I think we still want a handshake. I don't know how multi tenacy will work with the KMS yet (both across keys and across chains), but I would expect to authorize some kind of session up front for a particular key and a particular chain (and version of the protocol), and then be able to send messages that should continue to include all of that information and be checked each time.

In summary, we should consider

- aligning Vote/Proposal fields with canonical order and fields
- send chain-id to the kms along with vote
- introduce a handshake the sets a validator and a chain-id and verify it with every message before signing

Liamsi:
> send chain-id to the kms along with vote

Would this be another breaking change as this would change the messages that get exchanged between remote signer / kms and tendermint?

jleni:
@Liamsi ping me in case we need to adjust the validator app specs/implementation.
[...]

PR #2675, opened 2018-10-19
types: first field in Canonical structs is Type

ebuchman:
Despite deciding in #1622 and ADR-024 to include a version field in the sign bytes,  https://github.com/tendermint/tendermint/issues/2665 suggested it was unnecessary, so it was removed. Further discussion has re-iterated the desire for the HSMs to be able to independently reject schemas they do not know about, without relying on something like the KMS to do it for them. 

Rather than bring back the Version field, we can use the existing `Type byte` field, which currently has only 4 defined types (prevote, precommit, proposal, heartbeat). If and when we need to upgrade the sign bytes format, we can introduce new types here. 256 options should be more than enough.

* [x] Wrote tests

### DIFF
--- a/docs/spec/blockchain/encoding.md
+++ b/docs/spec/blockchain/encoding.md
@@ -307,9 +307,9 @@ We call this encoding the SignBytes. For instance, SignBytes for a vote is the A
 
 ```go
 type CanonicalVote struct {
+	Type      byte
 	Height    int64            `binary:"fixed64"`
 	Round     int64            `binary:"fixed64"`
-	Type      byte
 	Timestamp time.Time
 	BlockID   CanonicalBlockID
 	ChainID   string


docs/spec/blockchain/encoding.md
```diff
@@ -301,15 +301,15 @@ Where the `"value"` is the base64 encoding of the raw pubkey bytes, and the
...
+are first, making it easy to quickly check the type, height, and round.
 The `ChainID` is also appended to the end.
 We call this encoding the SignBytes. For instance, SignBytes for a vote is the Amino encoding of the following struct:
 
 ```go
 type CanonicalVote struct {
 	Height    int64            `binary:"fixed64"`
 	Round     int64            `binary:"fixed64"`
-	VoteType  byte
+	Type      byte
```
Liamsi: This would need to be the first field in the documentation too.

ebuchman: Weird I thought I fixed that - thanks!

PR #2681, opened 2018-10-19
abci: update Header to match amino encoding of types.Header

ebuchman:
Turns out the amino encoding of the types.Header didn't match the proto3 encoding of the abci.Header - ref #2682.

The problem was that the ABCI header wasn't using signed integers properly - `int64` in proto files maps to `uint64` in Go. Need to use `sint64` in the proto files instead.

We also needed to update to amino v0.13.0 so that the Timestamp encoding would be compatible.

This PR fixes all of that - now the aminoEncoding(types.Header) and proto3Encoding(abci.Header) match.

* [x] Wrote tests


abci/types/types.proto
```diff
@@ -238,10 +238,10 @@ message Header {
   // basic block info
   Version version = 1 [(gogoproto.nullable)=false];
   string chain_id = 2 [(gogoproto.customname)="ChainID"];
-  int64 height = 3;
```
ValarDragon: I hope its not too late to revert this decision prelaunch, but I think height being signed is the wrong decision. Feel free to close this comment if the ship has sailed.

abci/types/types.proto
```diff
@@ -215,19 +215,19 @@ message ConsensusParams {
...
+  sint64 max_bytes = 1;
   // Note: must be greater or equal to -1
-  int64 max_gas = 2;
+  sint64 max_gas = 2;
 }
 
 // EvidenceParams contains limits on the evidence.
 message EvidenceParams {
   // Note: must be greater than 0
-  int64 max_age = 1;
```
ValarDragon: IMO, This really should be a uint, since we even say must be greater than 0.

abci/types/types.proto
ebuchman: Did you read https://blog.cosmos.network/choosing-a-type-for-blockchain-height-beware-of-unsigned-integers-714804dddf1d ? Can you argue against it ? Having to add an `s` here made me doubt our decision for the first time, but I still think it's probably right.

ValarDragon: I have read it, will respond more later. I do disagree with it, and think that not choosing uint does increase chance of error. (I'll have to double check later, but I think we even have evidence of it doing so in our project)

state/state.go
```diff
@@ -140,29 +140,22 @@ func (state State) MakeBlock(
...
-	block.LastBlockID = state.LastBlockID
-	block.TotalTxs = state.LastBlockTotalTx + block.NumTxs
-
-	block.ValidatorsHash = state.Validators.Hash()
-	block.NextValidatorsHash = state.NextValidators.Hash()
-	block.ConsensusHash = state.ConsensusParams.Hash()
-	block.AppHash = state.AppHash
-	block.LastResultsHash = state.LastResultsHash
-
-	// NOTE: we can't use the state.Validators because we don't
```
melekes: this is important

ebuchman:
Not thrilled about having to change these all to `sint64`. See summary in https://github.com/tendermint/tendermint/issues/2682#issuecomment-431591436

state/state.go
melekes: >  it's commented in the functions godoc.

where? 👀

ebuchman:
I think we should close this for https://github.com/tendermint/tendermint/pull/2687.

I don't think it makes sense to switch ABCI ints to `sint`. We should probably go with #2684  but I'm not sure that's in scope for this release.

PR #2696, opened 2018-10-24
[WIP] Fill in consensus core details in ADR 030

milosevic:
* [ ] Wrote tests

### DIFF
--- a/docs/architecture/adr-030-consensus-refactor.md
+++ b/docs/architecture/adr-030-consensus-refactor.md
@@ -136,14 +136,10 @@ type Event interface{}
 
 type EventNewHeight struct {
     Height           int64
-    ValidatorSetSize int
     ValidatorId      int
 }
 
-type EventNewRound struct {
-    Height           int64
-    Round            int
-}
+type EventNewRound HeightAndRound
 
 type EventProposal struct {
     Height           int64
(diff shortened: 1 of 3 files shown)


docs/architecture/adr-030-consensus-refactor.md
```diff
@@ -126,6 +126,184 @@ func TestConsensusXXX(t *testing.T) {
...
+    Height           int64
+    Round            int
+}
+
+type Majority23PrecommitAny struct {
+    Height           int64
+    Round            int
+}
+
+type TimeoutPropose struct {
```
ebuchman: We should just define a struct with `Height` and `Round` once and let all these alias it, eg. `type TimeoutPropose HeightRoundMsg`

docs/architecture/adr-030-consensus-refactor.md
```diff
@@ -126,6 +126,184 @@ func TestConsensusXXX(t *testing.T) {
...
+	Step             RoundStep
+	LockedValue      BlockID
+	LockedRound      int
+	ValidValue       BlockID
+	ValidRound       int
+	ValidatorSetSize int
+	ValidatorId      int
+}
+
+func proposer(height int64, round int) int {}
```
ebuchman: I'm not sure how to structure this function given that the validator set is changing over time ...

PR #2697, opened 2018-10-24
Update consensus.md

TraceBundy:
* [ ] Wrote tests


melekes:
On the second look, the current spec seems to be correct. The corresponding line is https://github.com/tendermint/tendermint/blob/6643c5dd1151fdee803eb1a27e54b5dd81c65be5/consensus/state.go#L1645.

TraceBundy:
After receiving proposal block and all prevotes at PoLC-Round

Issue #2702, opened 2018-10-25
consensus: proposal#POLRound checking code might be incorrect


melekes:
https://github.com/tendermint/tendermint/blob/bbf15b3d09ac05ff2cfd58b53313ec612e1a9b07/consensus/state.go#L1428-L1432

Refs https://github.com/tendermint/tendermint/issues/2392#issuecomment-427775713
In `log-node2.txt` I found 

```
Okt 01 15:50:24 ip-172-31-9-150.us-east-2.compute.internal thorchaind[20170]: E[10-01|15:50:24.490] Error with msg                               module=consensus height=366511 round=0 type=*consensus.ProposalMessage peer= err="Error invalid proposal POL round" msg="[Proposal Proposal{366511/0 1:707EC619D399 (0,:0:000000000000) /EA1BAEEF4B7D.../ @ 2018-10-01T15:50:24.485Z}]"
```

Can't `POLRound` be 0? If not, what could've led the honest node to create such a proposal?

In any case, I'd prefer the above code to be rewritten:

```go
	if !(proposal.POLRound == -1 ||
		(proposal.POLRound > 0 && proposal.POLRound < proposal.Round)) {
		return ErrInvalidProposalPOLRound
	}
```

melekes:
Note here

```
Okt 01 15:50:24 ip-172-31-9-150.us-east-2.compute.internal thorchaind[20170]: E[10-01|15:50:24.490] Error with msg                               module=consensus height=366511 round=0 type=*consensus.ProposalMessage peer= err="Error invalid proposal POL round" msg="[Proposal Proposal{366511/0 1:707EC619D399 (0,:0:000000000000) /EA1BAEEF4B7D.../ @ 2018-10-01T15:50:24.485Z}]"
```

proposal.Round == proposal.POLRound

ebuchman:
```go
	if !(proposal.POLRound < -1 ||
		(proposal.POLRound >= 0 && proposal.POLRound >= proposal.Round)) {
		return ErrInvalidProposalPOLRound
	}
```

This doesn't look right - wouldn't this return an error if POLRound == -1 ?

I think perhaps what's happening is that the proposer is receiving +2/3 prevotes for nil for the round before they even get a chance to make a proposal. So then POLInfo returns a 0 for POLRound since it's seen a +2/3 prevotes

milosevic:
[...]

PR #2704, opened 2018-10-25
if some process locks a block in round 0, then 0 is valid proposal.PO…

melekes:
…LRound in rounds > 0

This condition is really hard to get. Initially, lockedRound and
validRound are set to -1 as we start with round 0.

Refs #2702

* [ ] Wrote tests


ebuchman:
Closing for https://github.com/tendermint/tendermint/pull/2735/

@milosevic please confirm

melekes:
I think changes here make the code better&more readable (NOTE they do not fix/change anything)

ebuchman:
Ah right, happy to bring that in, but it  looks like it changes the logic to allow POLRound to equal the current round which I don't think we want.

melekes:
No, it will return err if `proposal.POLRound >= proposal.Round`

PR #2660, opened 2018-10-18
Added the marshal,unmarshal and size methods to make compatible with grpc

nagarajmanjunath:
Added the marshal,unmarshal and size methods to make compatible with grpc protobuf
         *Added the above function in blockmeta.go and block.go

PR #2705, opened 2018-10-26
Added Protobuf marshal and unmarshal  compatiablity function in block.g…

nagarajmanjunath:
Added Protbuf marshal and unmarshal  compatiablity function in block.go,blockmeta.go,peerround and roundstate.go

As you have mentioned the methods using [(https://github.com/gogo/protobuf)](https://github.com/gogo/protobuf) for marshal and unmarshal here #2660 , I am using same in my proto file.
         
       import "github.com/gogo/protobuf/gogoproto/gogo.proto";
             option (gogoproto.marshaler_all) = true;
             option (gogoproto.unmarshaler_all) = true;
             option (gogoproto.sizer_all) = true;
             option (gogoproto.goproto_registration) = true;

       // Generate tests
              option (gogoproto.populate_all) = true;
              option (gogoproto.equal_all) = true;
              option (gogoproto.testgen_all) = true;,
 
but when i create protobuf using above code in my proto, Its generate MarshalTo(),Size(),Unmarshal ()and marshal() methods in protobuf. And gives error from protobuf file saying that its unable to detect MarshalTo() Size() function for particular types, which is not present in tendermint block,blockmeta.go

* [ ] Wrote tests

### DIFF
--- a/consensus/types/peer_round_state.go
+++ b/consensus/types/peer_round_state.go
@@ -55,3 +55,26 @@ func (prs PeerRoundState) StringIndented(indent string) string {
 		indent, prs.CatchupCommit, prs.CatchupCommitRound,
 		indent)
 }
+
+
+// Protobuf Compatiablity
+func (ps *PeerRoundState) Unmarshal(bs []byte) error {
+	return cdc.UnmarshalBinaryBare(bs, ps)
+}
+
+func (ps *PeerRoundState) Marshal() ([]byte, error) {
+	return cdc.MarshalBinaryBare(ps)
+}
+
+func (ps *PeerRoundState) MarshalTo(data []byte) (int, error) {
+	bs, err := ps.Marshal()
+	if err != nil {
+		return -1, err
+	}
+	return copy(data, bs), nil
+}
+
+func (ps *PeerRoundState) Size() int {
+	bs, _ := ps.Marshal()
+	return len(bs)
+}
--- a/types/block.go
+++ b/types/block.go
@@ -723,3 +723,20 @@ func (blockID BlockID) Key() string {
 func (blockID BlockID) String() string {
 	return fmt.Sprintf(`%v:%v`, blockID.Hash, blockID.PartsHeader)
 }
+// Protobuf Compatiablity
+
+func (bc *Block) Unmarshal(bs []byte) error {
+	return cdc.UnmarshalBinaryBare(bs, bc)
+}
+
+func (bc *Block) Marshal() ([]byte, error) {
+	return cdc.MarshalBinaryBare(bc)
+}
+
+func (bc *Block) MarshalTo(data []byte) (int, error) {
+	bs, err := bc.Marshal()
+	if err != nil {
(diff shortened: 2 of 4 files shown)


b00f:
@melekes Could you please check this PR?

melekes:
> And gives error from protobuf file saying that its unable to detect MarshalTo() Size() function for particular types, which is not present in tendermint block,blockmeta.go

That's weird. Could you share your `.proto` file? It should look similar to https://github.com/tendermint/tendermint/blob/4f61b97bbe69a08c3406e3b04148ef8fa866641d/types/proto3/block.proto.

b00f:
my colleague is working on grpc and as he explained to me he is unable to get block information due to some functions are missed (and it looks they are necessary for marshaling and unmarshaling)
the rpc method is:
 
```grpc
rpc GetBlock(BlockRequest) returns(BlockResponse);

message BlockRequest {
	uint64 height = 1;
}

message BlockResponse {
  bytes  BlockMeta  = 1  [(gogoproto.customtype) = "github.com/tendermint/tendermint/types.BlockMeta"];
  bytes Block  = 2  [(gogoproto.customtype) = "github.com/tendermint/tendermint/types.Block"];
}
```

melekes:
You're using custom types. Now everything makes sense. Thank you

consensus/types/peer_round_state.go
```diff
@@ -55,3 +55,26 @@ func (prs PeerRoundState) StringIndented(indent string) string {
...
+// Protobuf Compatiablity
+func (ps *PeerRoundState) Unmarshal(bs []byte) error {
+	return cdc.UnmarshalBinaryBare(bs, ps)
+}
+
+func (ps *PeerRoundState) Marshal() ([]byte, error) {
+	return cdc.MarshalBinaryBare(ps)
+}
+
+func (ps *PeerRoundState) MarshalTo(data []byte) (int, error) {
```
ebuchman: Is the idea that `Size()` is called before calling `MarshalTo` to ensure the `data` buffer is correctly sized?

types/block.go
```diff
@@ -723,3 +723,20 @@ func (blockID BlockID) Key() string {
 func (blockID BlockID) String() string {
 	return fmt.Sprintf(`%v:%v`, blockID.Hash, blockID.PartsHeader)
 }
+// Protobuf Compatiablity
+
+func (bc *Block) Unmarshal(bs []byte) error {
```
ebuchman: These should be moved up to where the rest of the `Block` methods live

consensus/types/peer_round_state.go
nagarajmanjunath: @ebuchman, yes its used to check the size/len of the buffer.

types/block.go
nagarajmanjunath: @ebuchman, I have update the methods #2918, Sorry for inconvenience, I created new pull request due to changes occurred in Nodeinfo.go. Kindly request you verify it.

Issue #2706, opened 2018-10-26
Add year to log format


yutianwu:
Our time format is `[05-02|11:06:44.322]` now.

But we found it inconvenient to collect logs to ELK stack for we don't have a complete timestamp and timestamp depends on the collector's time.

IMHO, if we add year to log format, it will do no harm to our format.

Issue #1212, opened 2018-02-14
Improve logging context information


xla:
Inspired by a [conversation](https://github.com/tendermint/tendermint/pull/1204#discussion_r167392392) in #1204  is this proposal to improve the information carried in every log line. In essence each log line should provide enough context so that an operator can infer origin and meainng, even when looked at in isolation in a remote place. From the original conversation:

> Usually thinking about in what remote ends logs can end up in typical infrastructure I try to imagine what vital info is needed to make sense of it when consumed in some log infrastructure remote from the origin. Additionally how each log line can interpreted by itself without context of the lines before or after. As the logger is passed from the outside to the inside in the program we can enrich the keyvals with information as well which subsystems don't need to care about. A standard set of fields I try to surface, most of them which can easily be achieved with Dynamic Contextual Values from the go-kit log package:

Propsed set of standard fields:

* **caller**: https://godoc.org/github.com/go-kit/kit/log#Caller
* **timestamp**: https://godoc.org/github.com/go-kit/kit/log#pkg-variables
* **revision**: compile flag: https://github.com/tendermint/tendermint/blob/master/Makefile#L10
* **job**: `tendermint` in our case
* **task**: `node`for the node command
* **service**: what we pass to `cmn.NewBaseService(...)` e.g.: `privValidatorSocketServer`
* **err**: set if error present

***

*All field names are up for discussion and can be adjusted according to some sane strategy.*
*Proper error propagation and messages are important for the overall goal but exceeds the scope of this issue*

tessr:
The team discussed this today, and we'd like to move forward on it (although it hasn't been put on the roadmap yet) so I'm removing the S:proposal label. 

I think the proposed keys are a good starting point. It would also be interesting to solicit feedback from validators and other operators. What might be most helpful to people running Tendermint?

marbar3778:
`tmLogger.Error` is currently treated the same as `Error` and `Errorf` from the std lib. It would be nice to have `tmLogger.Error` for strings and `tmlogger.Errorf` when formatting is needed.

PR #2707, opened 2018-10-26
[R4R] libs/log: add year to log format

yutianwu:
According to https://github.com/tendermint/tendermint/issues/2706, I encountered a problem about log format collecting logs to ELK stack for lacking year in timestamp.

Our log format before is:
```
NOTE[04-25|14:45:08] ABCI Replay Blocks                       module=consensus appHeight=0 storeHeight=0 stateHeight=0
```
I want to change it to 
```
NOTE[2017-04-25|14:45:08] ABCI Replay Blocks                       module=consensus appHeight=0 storeHeight=0 stateHeight=0
```

It will help a lot to analyze logs for us.


melekes:
Refs https://github.com/tendermint/tendermint/issues/1212

yutianwu:
I see, but it seems there is no conclusion. When can we sit down to reach an agreement and set out to do that. I can help.

Issue #2683, opened 2018-10-20
Validate all reactor messages


ebuchman:
We received another bug bounty submission where a negative integer in a reactor message crashed the process.

All reactor messages should have a `ValidateBasic()` method which checks for internally consistency before handling them, eg. negative ints, field sizes, etc.

ebuchman:
Closed in #2711.

Follow up to write more tests in https://github.com/tendermint/tendermint/issues/2740

PR #2711, opened 2018-10-26
validate reactor messages

melekes:
Refs #2683

* [ ] ~~Updated all relevant documentation in docs~~ no need(?)
* [ ] Wrote tests

TODO

* [x] validate netAddr in pex reactor
* [x] check signatures size


types/heartbeat.go
```diff
@@ -50,3 +52,27 @@ func (heartbeat *Heartbeat) String() string {
...
+	if heartbeat.Height < 0 {
+		return errors.New("Negative Height")
+	}
+	if heartbeat.Round < 0 {
+		return errors.New("Negative Round")
+	}
+	if heartbeat.Sequence < 0 {
+		return errors.New("Negative Sequence")
+	}
+	// should we check signature size? > 0 or exact size?
```
melekes: feel like we should check 0 < sig < MaxSize(128 bytes?)

melekes:
Also we should try offload this task to compiler :) if possible

blockchain/reactor.go
```diff
@@ -352,7 +358,9 @@ func (bcR *BlockchainReactor) BroadcastStatusRequest() error {
 // Messages
 
 // BlockchainMessage is a generic message for this reactor.
-type BlockchainMessage interface{}
+type BlockchainMessage interface {
+	ValidateBasic() error
```
melekes: Is it OK for this interface to have `ValidateBasic` method? Or it must be completely generic?

ebuchman: No I think it's good to have this method

types/heartbeat.go
```diff
@@ -50,3 +52,32 @@ func (heartbeat *Heartbeat) String() string {
 		heartbeat.Height, heartbeat.Round, heartbeat.Sequence,
 		fmt.Sprintf("/%X.../", cmn.Fingerprint(heartbeat.Signature[:])))
 }
+
+// ValidateBasic performs basic validation.
+func (heartbeat *Heartbeat) ValidateBasic() error {
+	if len(heartbeat.ValidatorAddress) != tmhash.Size {
```
ebuchman: Note Address size will diverge from tmhash.Size soon

types/part_set.go
```diff
@@ -36,6 +36,14 @@ func (part *Part) Hash() []byte {
 	return part.hash
 }
 
+// ValidateBasic performs basic validation.
+func (part *Part) ValidateBasic() error {
+	if part.Index < 0 {
+		return errors.New("Negative Index")
+	}
+	return nil
```
ebuchman: Should we check part.Bytes size here now that it's hardcoded to max 64kb?

types/proposal.go
```diff
@@ -42,6 +42,32 @@ func NewProposal(height int64, round int, blockPartsHeader PartSetHeader, polRou
...
 }
 
+// ValidateBasic performs basic validation.
+func (p *Proposal) ValidateBasic() error {
+	if p.Height < 0 {
+		return errors.New("Negative Height")
+	}
+	if p.Round < 0 {
+		return errors.New("Negative Round")
+	}
```
ebuchman: timestamp?

types/signable.go
```diff
@@ -1,5 +1,11 @@
 package types
 
+const (
+	// MaxSignatureSize is a maximum allowed signature size for the Heartbeat,
+	// Proposal and Vote.
+	MaxSignatureSize = 64
```
ebuchman: I think we need this to be the `math.Max` of values read from the crypto package incase we introduce bigger signatures

types/vote.go
```diff
@@ -107,3 +109,32 @@ func (vote *Vote) Verify(chainID string, pubKey crypto.PubKey) error {
 	}
 	return nil
 }
+
+// ValidateBasic performs basic validation.
+func (vote *Vote) ValidateBasic() error {
+	if len(vote.ValidatorAddress) != tmhash.Size {
```
ebuchman: Will diverge from tmhash.Size.  Think we need AddressSize param

types/vote.go
```diff
@@ -107,3 +109,32 @@ func (vote *Vote) Verify(chainID string, pubKey crypto.PubKey) error {
...
+	}
+	if vote.Height < 0 {
+		return errors.New("Negative Height")
+	}
+	if vote.Round < 0 {
+		return errors.New("Negative Round")
+	}
+	if err := vote.BlockID.ValidateBasic(); err != nil {
+		return fmt.Errorf("Wrong BlockID: %v", err)
+	}
```
ebuchman: timestamp and Type

types/evidence.go
```diff
@@ -172,6 +174,20 @@ func (dve *DuplicateVoteEvidence) Equal(ev Evidence) bool {
 	return bytes.Equal(dveHash, evHash)
 }
 
+// ValidateBasic performs basic validation.
+func (dve *DuplicateVoteEvidence) ValidateBasic() error {
+	if len(dve.PubKey.Bytes()) == 0 {
+		return errors.New("Empty PubKey")
+	}
+	if err := dve.VoteA.ValidateBasic(); err != nil {
```
ebuchman: These votes are pointers. We either need to make them non-pointers, or check that they're not nil before dereferencing to validate.

blockchain/reactor.go
ValarDragon: I agree, I think this method being required is a good idea

blockchain/reactor.go
```diff
@@ -377,6 +385,14 @@ type bcBlockRequestMessage struct {
 	Height int64
 }
 
+// ValidateBasic performs basic validation.
+func (m *bcBlockRequestMessage) ValidateBasic() error {
+	if m.Height < 0 {
+		return errors.New("Negative Height")
```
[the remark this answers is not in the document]
melekes: And what exactly will we gain by doing that? Should we do this for every reactor? I guess this is only fair for common fields like height & round, but wouldn't it be weird to have errors defined just for those two fields? I am sceptical.

consensus/reactor.go
```diff
@@ -1488,6 +1595,21 @@ type VoteSetBitsMessage struct {
...
+	if m.Height < 0 {
+		return errors.New("Negative Height")
+	}
+	if m.Round < 0 {
+		return errors.New("Negative Round")
+	}
+	if err := m.BlockID.ValidateBasic(); err != nil {
+		return fmt.Errorf("Wrong BlockID: %v", err)
+	}
+	// NOTE: Votes.Size() can be zero if the node does not have any
```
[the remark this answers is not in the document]
melekes: It's a bit hard to read the code since we have these `if nil return nil` bits. Here I think we can return nil https://github.com/tendermint/tendermint/blob/c8cf78039f45a08c7f8fec7fd7acc12d47090832/types/vote_set.go#L345 if we don't have prevotes/precommits for that blockID

p2p/pex/pex_reactor.go
```diff
@@ -294,15 +294,24 @@ func (r *PEXReactor) ReceiveAddrs(addrs []*p2p.NetAddress, src Peer) error {
 
 	srcAddr := src.NodeInfo().NetAddress()
 	for _, netAddr := range addrs {
-		// NOTE: GetSelection methods should never return nil addrs
-		if netAddr == nil {
-			return cmn.NewError("received nil addr")
+		// Validate netAddr.
+		// TODO: extract validating logic from NewNetAddressStringWithOptionalID
+		// and put it in netAddr#Valid (#2722)
+		na, err := p2p.NewNetAddressString(netAddr.String())
```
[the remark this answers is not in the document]
melekes: Good catch! Thank you

types/proposal.go
melekes: how should validate it? compare it to the local time with some delta?

types/evidence.go
melekes: Awesome find

consensus/reactor.go
```diff
@@ -1375,6 +1381,27 @@ type NewRoundStepMessage struct {
 	LastCommitRound       int
 }
 
+// ValidateBasic performs basic validation.
+func (m *NewRoundStepMessage) ValidateBasic() error {
+	if m.Height < 1 {
+		return errors.New("Negative or zero Height")
```
ebuchman: I meant this for everything. Instead of `Height < 0` we might want `Height < 1` everywhere.

consensus/reactor.go
```diff
@@ -1375,6 +1381,27 @@ type NewRoundStepMessage struct {
...
+		return errors.New("Negative Round")
+	}
+	if !m.Step.IsValid() {
+		return errors.New("Invalid Step")
+	}
+	if m.SecondsSinceStartTime < 0 {
+		return errors.New("Negative SecondsSinceStartTime")
+	}
+	if (m.Height == 1 && m.LastCommitRound < -1) ||
+		(m.Height > 1 && m.LastCommitRound < 0) {
```
ebuchman:
```go
if (m.Height == 1 && m.LastCommitRound != -1) ||
		(m.Height > 1 && m.LastCommitRound < 0) {
```

types/evidence.go
ebuchman: Thank @Slamper . He's killing it on the bug bounty

types/proposal.go
ebuchman: Yeh, maybe plus or minus a year? Just to eliminate the crazy outliers. The decoder will have done alot of work for us here already actually. We can punt on it for now.

consensus/reactor.go
ebuchman: Let's leave it at `< 0` everywhere for now and revisit.

types/evidence.go
Slamper: Thanks @ebuchman :+1:  Always happy to contribute!

types/proposal.go
```diff
@@ -50,6 +50,11 @@ func (p *Proposal) ValidateBasic() error {
 	if p.Round < 0 {
 		return errors.New("Negative Round")
 	}
+	now := tmtime.Now()
+	oneYear := 8766 * time.Hour
```
ebuchman: let's make a `ValidateTime(timestamp time.Time) error` function for this?

types/block.go
```diff
@@ -65,46 +66,46 @@ func (b *Block) ValidateBasic() error {
...
+
 	if !bytes.Equal(b.EvidenceHash, b.Evidence.Hash()) {
 		return fmt.Errorf(
 			"Wrong Block.Header.EvidenceHash.  Expected %v, got %v",
 			b.EvidenceHash,
 			b.Evidence.Hash(),
 		)
 	}
+
 	return nil
```
Slamper: We should also do `VerifyBasic` on the evidence here since otherwise the votes in the `DuplicateVoteEvidence` could still be `nil` when the block is further processed.

Slamper:
```suggestion
	for _, ev := range b.Evidence.Evidence {
		if err := ev.ValidateBasic(); err != nil {
			return err
		}
	}
```

Slamper: @ebuchman @melekes This would be a quick fix. However `evidence` is of the type `EvidenceData` which should also have a `ValidateBasic` function. That would also have to check whether `EvidenceData.Evidence != nil` since a nil slice could be passed which could cause another `Consensus Failure`.

We have to check this before doing `b.Evidence.Hash()` in the lines above since it could otherwise crash there!

This sounds like another bounty or do we count this as included in the first finding ? ;)

blockchain/reactor.go
```diff
@@ -395,6 +419,11 @@ type bcBlockResponseMessage struct {
 	Block *types.Block
 }
 
+// ValidateBasic performs basic validation.
+func (m *bcBlockResponseMessage) ValidateBasic() error {
+	return m.Block.ValidateBasic()
```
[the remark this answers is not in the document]
melekes: note: If we validate time +/- 1 year, users won't be able to sync chains older than 1 year (from the start).

consensus/reactor.go
```diff
@@ -1375,6 +1383,23 @@ type NewRoundStepMessage struct {
...
+	if m.Height < 0 {
+		return errors.New("Negative Height")
+	}
+	if m.Round < 0 {
+		return errors.New("Negative Round")
+	}
+	if m.SecondsSinceStartTime < 0 {
+		return errors.New("Negative SecondsSinceStartTime")
+	}
+	if m.LastCommitRound < -1 {
```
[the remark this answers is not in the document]
melekes:
```
2018-10-31T10:59:12+00:00 172.17.0.1 local_testnet_2[1256]: E[31106-10-31|10:59:12.176] Peer sent us invalid msg                     module=consensus peer="Peer{MConn{172.5
7.0.101:38716} 55854e179f195424fbf969016e1a2ab47e1a0aea in}" msg="[NewRoundStep H:22 R:0 S:RoundStepPrevote LCR:-1]" err="Invalid LastCommitRound (for 1st block: -1, for ot
hers: >= 0)"
2018-10-31T10:59:12+00:00 172.17.0.1 local_testnet_2[1256]: E[31106-10-31|10:59:12.176] Stopping peer for error                      module=p2p peer="Peer{MConn{172.57.0.10
1:38716} 55854e179f195424fbf969016e1a2ab47e1a0aea in}" err="Invalid LastCommitRound (for 1st block: -1, for others: >= 0)"
```

I am observing a different picture. p2p tests are failing because of this

ebuchman:
Why were the test files deleted? p2p tests fail:

```
Successfully initialized 4 node directories
ERROR: open /go/src/github.com/tendermint/tendermint/test/p2p/data/mach0/config/node_key.json: no such file or directory
Exited with code 1
```

melekes:
> Why were the test files deleted? p2p tests fail:

we now use `tendermint testnet` cmd to generate them (that way we won't need to update them every time we change genesis/node_key/priv_validator files). Tests pass for me locally. I am not sure why they're failing on circle.

test/p2p/peer.sh
```diff
@@ -21,7 +21,7 @@ if [[ "$ID" == "x" ]]; then # Set "x" to "1" to print to console.
 		--ip=$(test/p2p/ip.sh "$ID") \
 		--name "local_testnet_$ID" \
 		--entrypoint tendermint \
-		-e TMHOME="/go/src/github.com/tendermint/tendermint/test/p2p/data/mach$ID/core" \
+		-e TMHOME="/go/src/github.com/tendermint/tendermint/test/p2p/data/mach$((ID-1))" \
```
melekes: write a comment about where's mach data is coming from (because it's not in the repo anymore)

melekes:
Done here.

types/block.go
ebuchman: `b.Evidence` can't be nil since it's not a pointer. `b.Evidence.Evidence` could be nil because it's a slice,  but here we just loop over it, so if its nil its no problem, the loop just won't run. The methods defined on EvidenceList similarly only loop over it, so if it's nil there should be no issue. Am I missing something? Or you're saying we should always ensure we have a zero-length array here and not nil ?

Issue #2671, opened 2018-10-18
Add Version to ConsensusParams


ebuchman:
Follow up to #2468, this is the final piece of [ADR-016](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-016-protocol-versions.md), to allow apps to update their version without a restart. 

Originally ADR-016 proposed adding a new field to EndBlock for updating versions, but it was realized in https://github.com/tendermint/tendermint/pull/2636#discussion_r225602628 that it might make more sense to just add it directly to the ConsensusParams.

Note we forgot to update the changelog in ADR-016 when it was updated for this change in #2667

ebuchman:
#3152 makes this non protocol breaking. Bumping it back

b00f:
I have an idea. Out of box idea, Instead of changing protocol. What we need is an external application, lets call it `AppSelector`.
Changing protocol to support possible fork or updates will make codebase nasty. This messy code currently is applied Ethereum. Look [here](https://github.com/ethereum/go-ethereum/blob/f4094d09cd3bf4c1420f358d77b803d5d84d0994/params/config.go#L35)

AppSelector will run the application but it can manage versioning. AppSelector:
- Monitor the blockchain. Anytime the AppVersion changes it will automatically find the new version (possibly from GitHub release page) and run the required verision
- It can be programmed to start new version at a certain height.

Issue #1007, opened 2017-12-27
Add events to results hash


ebuchman:
Should be in: https://github.com/tendermint/tendermint/pull/999

But waiting on possible changes to abci responses: https://github.com/tendermint/abci/issues/165

ebuchman:
Instead, we should keep the tags in their own merkle tree so we can do proofs of inclusion/exclusion on the tags directly for a block

ebuchman:
Noting that right now the results hash only includes the code and data, so we also need to accommodate the other response fields, like gas used/wanted.

ValarDragon:
I have an idea for this described in https://github.com/tendermint/tendermint/pull/2314#discussion_r215092568, along with some ways to further compress it using accumulators.

I'll work on writing this up.

ebuchman:
Should we consider just not merklizing the tags for now? At least until we sort out #1859 and clean up the lite client? This issue would then move from launch to first-hard-fork.

melekes:
To give people more context (and to validate my assumptions), I will try to summarize what do we want and why:

Currently, Header#LastResultsHash only includes ResponseDeliverTx.Code and ResponseDeliverTx.Data.

We want to incorporate:

1) GasWanted and GasUsed fields from ResponseDeliverTx.
2) events (from ResponseBeginBlock, ResponseDeliverTx, and ResponseEndBlock) into LastResultsHash so that the clients were able to verify event X indeed happened in block Y. It's not clear whenever we should 

a) hash them directly OR 
b) have a separate Merkle tree hash (which gives clients proofs) OR 
c) have a separate Merkle tree hash w/ RSA accumulator.

alexanderbez:
How does this currently work @melekes? Are the `Code` and `Data` values from each tx in a block simply concatenated?  Doesn't seem to useful atm, does it?

Also, why do options (b) and (c) add an additional 32 bytes? Would the merkle root simply replace the hash we have now yielding no effective change in size? Or am I missing something?

I'd opt for (b) as it's much simpler, and AFAIK, doesn't rely on a trusted setup. In addition, the key/value space in each of these trees should never be that too large either (unless you have hundreds of txs in a block?).

melekes:
> How does this currently work @melekes?

we build a simple Merkle tree https://github.com/tendermint/tendermint/blob/master/types/results.go#L57

> Would the merkle root simply replace the hash we have now yielding no effective change in size? Or am I missing something?

[...]

PR #2713, opened 2018-10-27
Simple merkle rfc compatibility

### LINKED ISSUE #1892: crypto/merkle: SimpleMerkle should differentiate between interior & leaf nodes
See 
https://github.com/tendermint/tmlibs/pull/211#issuecomment-391102658
https://github.com/tendermint/tmlibs/pull/211#issuecomment-391044446

Moved over from https://github.com/tendermint/go-crypto/issues/107 (contains more pointers).

ValarDragon:
Closes #1892 

What remains:
* [x] Fix current error with indexer in the lite client
* [x] Update docs
* [x] Add Trillian test vectors
* [x] Change the split point per RFC 6962 for efficiency of leaf additions

* [x] Wrote tests


crypto/merkle/rfc6962_test.go
```diff
@@ -0,0 +1,94 @@
...
+	leaf1, leaf2 := []byte("Hello"), []byte("World")
+	_, leafHashTrail := trailsFromByteSlices([][]byte{leaf1})
+	hash1 := leafHashTrail.Hash
+	_, leafHashTrail = trailsFromByteSlices([][]byte{leaf2})
+	hash2 := leafHashTrail.Hash
+	if bytes.Equal(hash1, hash2) {
+		t.Errorf("Leaf hashes should differ, but both are %x", hash1)
+	}
+	// Compute an intermediate subtree hash.
+	_, subHash1Trail := trailsFromByteSlices([][]byte{hash1, hash2})
```
Liamsi: I wonder if shouldn't also split up `trailsFromByteSlices` into `hashChildren` and `hashLeaf` or so. Orthogonal to this PR though.

crypto/merkle/simple_tree_test.go
```diff
@@ -43,30 +42,53 @@ func TestSimpleProof(t *testing.T) {
 		require.Equal(t, proof.Total, total, "Unmatched totals: %d vs %d", proof.Total, total)
 
 		// Verify success
-		err := proof.Verify(rootHash, itemHash)
-		require.NoError(t, err, "Verificatior failed: %v.", err)
+		err := proof.Verify(rootHash, item)
+		require.NoError(t, err, "Verificatiom failed: %v.", err)
```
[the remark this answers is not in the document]
ValarDragon: LOL one typo changed to another typo

crypto/merkle/rfc6962_test.go
```diff
@@ -0,0 +1,94 @@
...
+	_, leafHashTrail := trailsFromByteSlices([][]byte{[]byte("L123456")})
+	leafHash := leafHashTrail.Hash
+	_, leafHashTrail = trailsFromByteSlices([][]byte{[]byte{}})
+	emptyLeafHash := leafHashTrail.Hash
+	for _, tc := range []struct {
+		desc string
+		got  []byte
+		want string
+	}{
+		// Check that the empty hash is not the same as the hash of an empty leaf.
```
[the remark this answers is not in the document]
ValarDragon: Not sure I follow, how would that ever happen. Thats also not a trillian test vector

crypto/merkle/rfc6962_test.go
ValarDragon: Sounds like a good idea to me, but lets open up in a separate issue

crypto/merkle/rfc6962_test.go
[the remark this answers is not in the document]
Liamsi: Sorry, I meant the first test-vector (empty root, not empty inner leaf; which indeed doesn't make any sense :-D)

crypto/merkle/hash.go
```diff
@@ -0,0 +1,21 @@
+package merkle
+
+import (
+	"github.com/tendermint/tendermint/crypto/tmhash"
+)
+
+// TODO: make these have a large predefined capacity
```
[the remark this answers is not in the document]
ValarDragon: Made one! #2833

types/tx.go
```diff
@@ -69,7 +70,7 @@ func (txs Txs) Proof(i int) TxProof {
 	l := len(txs)
 	bzs := make([][]byte, l)
 	for i := 0; i < l; i++ {
-		bzs[i] = txs[i]
+		bzs[i] = txs[i].Hash()
```
ebuchman: Think we need to update this in the spec. It seems txs are treated differently from everything else by being hashed before passed to the MerkleTree. Should the other data structures be treated similarly, so they're hashed before being passed in (and eg. concatenated with 0) ?

types/tx.go
```diff
@@ -87,8 +88,8 @@ type TxProof struct {
 	Proof    merkle.SimpleProof
 }
 
-// LeadHash returns the hash of the this proof refers to.
-func (tp TxProof) LeafHash() []byte {
+// Leaf returns the hash(tx), which is the leaf in the merkle tree which this proof refers to.
+func (tp TxProof) Leaf() []byte {
```
ebuchman: We should track these in the GoAPI breaking changes

crypto/merkle/simple_proof.go
```diff
@@ -67,7 +66,8 @@ func SimpleProofsFromMap(m map[string][]byte) (rootHash []byte, proofs map[strin
 
 // Verify that the SimpleProof proves the root hash.
 // Check sp.Index/sp.Total manually if needed
-func (sp *SimpleProof) Verify(rootHash []byte, leafHash []byte) error {
+func (sp *SimpleProof) Verify(rootHash []byte, leaf []byte) error {
```
ebuchman: Did we change the semantics of Verify here? It used to take the hash of the thing passed to SimpleMerkle, and now it takes the pre-image bytes?

crypto/merkle/simple_tree_test.go
```diff
@@ -43,30 +42,53 @@ func TestSimpleProof(t *testing.T) {
...
 		require.Error(t, err, "Expected verification to fail for mutated leaf hash")
 
 		// Mutating the rootHash should make it fail.
-		err = proof.Verify(MutateByteSlice(rootHash), itemHash)
+		err = proof.Verify(MutateByteSlice(rootHash), item)
 		require.Error(t, err, "Expected verification to fail for mutated root hash")
 	}
 }
+
+func Test_getSplitPoint(t *testing.T) {
```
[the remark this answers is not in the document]
ValarDragon: I use https://github.com/cweill/gotests to automate alot of the boiler plate of table driven tests. I forgot to change the template to remove the underscore though!

crypto/merkle/simple_proof.go
ValarDragon: Yeah, we had to so that everything that uses it doesn't have to be aware of the 0 byte prefix.

ebuchman:
@ValarDragon will this require updates on the SDK? Let's make sure those can get in if we're going to merge this for v0.27

ValarDragon:
It would break the store package test cases, but other than that I don't think it breaks anything. I'm not sure if anyone will have time to update the SDK though.

ebuchman:
> It would break the store package test cases, but other than that I don't think it breaks anything. I'm not sure if anyone will have time to update the SDK though.

Ok - I don't think we can bring it in if we're not going to take the time to fix the tests. It might also break lite client code if Verify is expecting something different. So I think this is justification to hold off for v0.28. We should look to include all final breaking changes for release in v0.28 if possible, ie. this, #1007 , and possibly #2671

ebuchman:
Let's merge this and follow up in https://github.com/tendermint/tendermint/issues/3120

PR #2714, opened 2018-10-27
Enforce that validators use pubkey types from validator params

ValarDragon:
* [ ] Wrote tests


state/execution.go
```diff
@@ -373,7 +391,7 @@ func updateState(
 	// Update the validator set with the latest abciResponses.
 	lastHeightValsChanged := state.LastHeightValidatorsChanged
 	if len(abciResponses.EndBlock.ValidatorUpdates) > 0 {
-		err := updateValidators(nValSet, abciResponses.EndBlock.ValidatorUpdates)
+		err := updateValidators(nValSet, abciResponses.EndBlock.ValidatorUpdates, state.ConsensusParams.Validator)
```
melekes: can we not mix validation and update logic? maybe extract validation into a separate function

```go
if err := validateUpdates(abciResponses.EndBlock.ValidatorUpdates, state.ConsensusParams.Validator.ValidatorPubkeyTypes); err != nil {
    return state, fmt.Errorf("Error in validator updates: %v", err)
}
```

ebuchman:
- [ ] Address review from https://github.com/tendermint/tendermint/pull/2636

state/execution.go
ValarDragon: Great point! I was originally trying to avoid an extra iteration over the validator set, but I think separating concerns is better for code clarity.

ValarDragon:
Sorry the commit history kinda became a mess, since this was forked off another branch which was then squash merged.

ValarDragon:
Actually theres a weird protobuf change here as well. I'll just close this PR and open a new one.

PR #2724, opened 2018-10-29
mempool: ErrPreCheck and more log info

ebuchman:
We weren't returning error when failing PreCheck before.

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -66,6 +66,9 @@ var (
 
 	// ErrMempoolIsFull means Tendermint & an application can't handle that much load
 	ErrMempoolIsFull = errors.New("Mempool is full")
+
+	// ErrPreCheck means the tx bytes are invalid - ie. too big.
+	ErrPreCheck = errors.New("Tx failed precheck")
 )
 
 // PreCheckAminoMaxBytes checks that the size of the transaction plus the amino
@@ -286,7 +289,7 @@ func (mem *Mempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
 	}
 
 	if mem.preCheck != nil && !mem.preCheck(tx) {
-		return
+		return ErrPreCheck
 	}
 
 	// CACHE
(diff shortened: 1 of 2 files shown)


mempool/mempool.go
```diff
@@ -66,6 +66,9 @@ var (
 
 	// ErrMempoolIsFull means Tendermint & an application can't handle that much load
 	ErrMempoolIsFull = errors.New("Mempool is full")
+
+	// ErrPreCheck means the tx bytes are invalid - ie. too big.
+	ErrPreCheck = errors.New("Tx failed precheck")
```
melekes: Should we be explicit about the reason in the msg / logs? Otherwise devs will have to dig into the code in order to figure out what precheck is.

ebuchman: Yeh that would be nice - maybe those funcs should return errors instead of bool?

melekes: Good idea

melekes:
```
--- FAIL: TestMempoolFilters (0.00s)
	mempool_test.go:69: Error after CheckTx: Tx failed precheck
```

mempool/mempool.go
```diff
@@ -566,7 +598,13 @@ func (mem *Mempool) recheckTxs(goodTxs []types.Tx) {
 }
 
 func (mem *Mempool) isPostCheckPass(tx types.Tx, r *abci.ResponseCheckTx) bool {
-	return mem.postCheck == nil || mem.postCheck(tx, r)
+	if mem.postCheck == nil {
+		return true
+	}
+	if err := mem.postCheck(tx, r); err != nil {
```
ebuchman: I think we probably want to report this through the logger. Looks like right now we just swallow it, so it won't be obvious why a tx failed

mempool/mempool_test.go
```diff
@@ -66,7 +65,10 @@ func checkTxs(t *testing.T, mempool *Mempool, count int) types.Txs {
 			t.Error(err)
 		}
 		if err := mempool.CheckTx(txBytes, nil); err != nil {
-			t.Fatalf("Error after CheckTx: %v", err)
+			if IsPreCheckError(err) {
```
ebuchman: why are these happening here?

mempool/mempool_test.go
```diff
@@ -126,47 +128,29 @@ func TestMempoolFilters(t *testing.T) {
 	mempool := newMempoolWithApp(cc)
 	emptyTxArr := []types.Tx{[]byte{}}
 
-	nopPreFilter := func(tx types.Tx) bool { return true }
-	nopPostFilter := func(tx types.Tx, res *abci.ResponseCheckTx) bool { return true }
-
-	// This is the same filter we expect to be used within node/node.go and state/execution.go
```
ebuchman: Nice!

mempool/mempool_test.go
melekes: TestMempoolFilters will fail otherwise (it asserts a number of txs created, not panic)

PR #2728, opened 2018-10-29
one less dependency

### LINKED ISSUE #2638: Remove ebuchman/fail-test dependency
We currently import ebuchman/fail-test in `consensus/state.go`, and `state/execution.go`. This repo is one small file, we should move it into tendermint.

zramsay:
closes #2638 

* [ ] Wrote tests

### DIFF
--- a/libs/fail/fail.go
+++ b/libs/fail/fail.go
@@ -1,0 +1,78 @@
+package fail
+
+import (
+	"fmt"
+	"math/rand"
+	"os"
+	"strconv"
+)
+
+var callIndexToFail int
+
+func init() {
+	callIndexToFailS := os.Getenv("FAIL_TEST_INDEX")
+
+	if callIndexToFailS == "" {
+		callIndexToFail = -1
+	} else {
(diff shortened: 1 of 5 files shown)


libs/fail/fail.go
```diff
@@ -0,0 +1,78 @@
+package fail
+
+import (
+	"fmt"
+	"math/rand"
+	"os"
+	"strconv"
+)
+
+var callIndexToFail int
```
melekes: note: we should probably rewrite this to avoid `callIndexToFail` being included in the runtime

libs/fail/fail.go
```diff
@@ -0,0 +1,78 @@
...
+		Exit()
+	}
+
+	callIndex += 1
+}
+
+// FailRand should be called n successive times.
+// It will fail on a random one of those calls
+// n must be greater than 0
+func FailRand(n int) {
```
ebuchman: We can just get rid of this func, it's never used.

PR #2730, opened 2018-10-30
Align Vote/Proposal fields with canonical order and fields

### LINKED ISSUE #2727: Align Vote/Proposal fields with canonical order and fields
We should align the Vote/Proposal fields with the Canonicalized fields so eg. Proposal will have SignMsgType and the orders will be the same.

Now that they both are encoded Amino and the order is determined by convenience for HSMs, these should be aligned for consistency so we don't have random orders in the spec.

Ref https://github.com/tendermint/tendermint/issues/2665#issuecomment-433632585

Liamsi:
resolves #2727 

* [ ] Wrote tests

### DIFF
--- a/types/proposal.go
+++ b/types/proposal.go
@@ -46,7 +46,7 @@ func NewProposal(height int64, round int, blockPartsHeader PartSetHeader, polRou
 
 // String returns a string representation of the Proposal.
 func (p *Proposal) String() string {
-	return fmt.Sprintf("Proposal{%v %v/%v %v (%v,%v) %X @ %s}",
+	return fmt.Sprintf("Proposal{%X(Proposal) %v/%v %v (%v,%v) %X @ %s}",
 		p.Type,
 		p.Height,
 		p.Round,
--- a/types/proposal.go
+++ b/types/proposal.go
@@ -20,11 +20,12 @@ var (
 // to be considered valid. It may depend on votes from a previous round,
 // a so-called Proof-of-Lock (POL) round, as noted in the POLRound and POLBlockID.
 type Proposal struct {
+	Type             SignedMsgType
 	Height           int64         `json:"height"`
 	Round            int           `json:"round"`
+	POLRound         int           `json:"pol_round"` // -1 if null.
 	Timestamp        time.Time     `json:"timestamp"`
 	BlockPartsHeader PartSetHeader `json:"block_parts_header"`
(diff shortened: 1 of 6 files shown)


types/proposal.go
```diff
@@ -33,21 +34,27 @@ type Proposal struct {
...
-		POLRound:         polRound,
 		POLBlockID:       polBlockID,
 	}
 }
 
 // String returns a string representation of the Proposal.
 func (p *Proposal) String() string {
-	return fmt.Sprintf("Proposal{%v/%v %v (%v,%v) %X @ %s}",
-		p.Height, p.Round, p.BlockPartsHeader, p.POLRound,
+	return fmt.Sprintf("Proposal{%X(Proposal) %v/%v %v (%v,%v) %X @ %s}",
```
ebuchman: I don't think we need the `%X(Proposal)` since we already have `Proposal{`

Liamsi: Thought I make it consistent with Vote but you are right. Removed.

Liamsi:
Updated but there seem to be some problems with circleci pulling the docker image:
```
Error pulling image circleci/golang:1.10.3: Error response from daemon: received unexpected HTTP status: 503 Service Unavailable... retrying
  image is cached as circleci/golang:1.10.3, but refreshing...
```

ebuchman:
Still WIP?

ebuchman:
Oh we need to update spec

Liamsi:
updated the spec. Should I add an entry to the changelog, or changelog pending?

ebuchman:
>  Should I add an entry to the changelog, or changelog pending?

We're in a twilight zone re changelog right now so don't worry about it, I'll handle it on the release front.

Thanks!

types/vote.go
```diff
@@ -85,16 +85,17 @@ func (vote *Vote) String() string {
 		cmn.PanicSanity("Unknown vote type")
 	}
 
-	return fmt.Sprintf("Vote{%v:%X %v/%02d/%v(%v) %X %X @ %s}",
-		vote.ValidatorIndex,
-		cmn.Fingerprint(vote.ValidatorAddress),
-		vote.Height,
-		vote.Round,
+	return fmt.Sprintf("Vote{%v(%v) %v/%02d @ %s %X %v:%X   %X}",
```
ebuchman: Oh I don't think we should change the order in the print statement. It's super convenient for humans to have the validator info at the front

Issue #2646, opened 2018-10-16
Remove POLInfo and simplify Proposal message


milosevic:
According to the spec, Proposal message has the following format: 
⟨PROPOSAL, height , round , proposal, validRound⟩ where proposal corresponds to the id of the block proposal (BlockID). Currently, Proposal message has the following format:
https://github.com/tendermint/tendermint/blob/80562669bfc0ad56d9d37662407c60661fdf476b/types/proposal.go#L22-L30

We could simplify Proposal message as done in spec by removing BlockPartsHeader and only having a single field that is BlockID of the proposed block. Furthermore, we currently rely on function POLInfo to decide what is POLRound and POLBlockID that goes into Proposal message. See https://github.com/tendermint/tendermint/blob/80562669bfc0ad56d9d37662407c60661fdf476b/consensus/types/height_vote_set.go#L148-L159

This is not needed as ValidRound, ValidBlock and ValidBlockParts keep equivalent of those information. Therefore, we could completely remove POLInfo function.

ebuchman:
> removing BlockPartsHeader and only having a single field that is BlockID

Do you mean instead of having BlockPartsHeader and POLBlockID, just have one BlockID?

I think the reason they're separate is because you don't need a BlockID.Hash for proposals in the happy path, which is why we just have the BlockPartsHeader, but you do need them for verifying POLs, since votes sign BlockIDs (BlockID.Hash is more convenient for lite-clients than BlockID.PartsHeader). 

Was it always the case that if POLRound is set, `POLBlockID.PartsHeader == BlockPartsHeader` ?

If we do this, it would add ~33 bytes per proposal on the happy path, but it's probably worth the alignment.

Note the spec doesn't treat the block gossipping mechanism, so it has no concept of the Hash vs PartsHeader yet.

> Therefore, we could completely remove POLInfo function.

Maybe we can keep it for now but run sanity checks to ensure when we make a proposal that our valid round/value match what we get from POLInfo.

milosevic:
> Do you mean instead of having BlockPartsHeader and POLBlockID, just have one BlockID?

Yes. 

> Was it always the case that if POLRound is set, POLBlockID.PartsHeader == BlockPartsHeader ?

Yes. 

Could we avoid setting Hash in case POLRound is not set? Would this help?

> Maybe we can keep it for now but run sanity checks to ensure when we make a proposal that our valid round/value match what we get from POLInfo.

[...]

PR #2735, opened 2018-10-31
Simplify proposal msg

milosevic:
Fix https://github.com/tendermint/tendermint/issues/2646.

Note that we could easily add nice Proposal related tests after https://github.com/tendermint/tendermint/pull/2652 is merged. 
 

* [x] Wrote tests

### DIFF
--- a/types/canonical.go
+++ b/types/canonical.go
@@ -23,14 +23,13 @@ type CanonicalPartSetHeader struct {
 }
 
 type CanonicalProposal struct {
-	Type             SignedMsgType // type alias for byte
-	Height           int64         `binary:"fixed64"`
-	Round            int64         `binary:"fixed64"`
-	POLRound         int64         `binary:"fixed64"`
-	Timestamp        time.Time
-	BlockPartsHeader CanonicalPartSetHeader
-	POLBlockID       CanonicalBlockID
-	ChainID          string
+	Type      SignedMsgType // type alias for byte
+	Height    int64         `binary:"fixed64"`
+	Round     int64         `binary:"fixed64"`
+	POLRound  int64         `binary:"fixed64"`
+	BlockID   CanonicalBlockID
+	Timestamp time.Time
+	ChainID   string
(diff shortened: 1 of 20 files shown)


types/canonical.go
```diff
@@ -23,14 +23,13 @@ type CanonicalPartSetHeader struct {
...
-	POLRound         int64         `binary:"fixed64"`
-	Timestamp        time.Time
-	BlockPartsHeader CanonicalPartSetHeader
-	POLBlockID       CanonicalBlockID
-	ChainID          string
+	Type      SignedMsgType // type alias for byte
+	Height    int64         `binary:"fixed64"`
+	Round     int64         `binary:"fixed64"`
+	BlockID   CanonicalBlockID
+	POLRound  int64 `binary:"fixed64"`
```
Liamsi: As far as I understand, the `BlockID` is not of fixed size (?). For HSMs to be able to parse the proposals, it is important that we keep the fixed sized fields in front (Type, Height, Round, POLRound, and then, BlockID, Timestamp, ChainID).

types/proposal.go
```diff
@@ -15,43 +15,41 @@ var (
...
-	POLRound         int           `json:"pol_round"` // -1 if null.
-	Timestamp        time.Time     `json:"timestamp"`
-	BlockPartsHeader PartSetHeader `json:"block_parts_header"`
-	POLBlockID       BlockID       `json:"pol_block_id"` // zero if null.
-	Signature        []byte        `json:"signature"`
+	Type      SignedMsgType
+	Height    int64     `json:"height"`
+	Round     int       `json:"round"`
+	BlockID   BlockID   `json:"block_id"`
+	POLRound  int       `json:"pol_round"` // -1 if null.
```
ebuchman: Let's keep the order here aligned with the Canonical order, so POLRound should come before BlockID

ebuchman:
Looks good, also need to update the spec!

PR #2738, opened 2018-10-31
some ValidateBasic updates

ebuchman:
Just addressing my own ongoing review of #2711. Still going through it.

### DIFF
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -1402,9 +1402,9 @@ func (m *NewRoundStepMessage) ValidateBasic() error {
 	if !m.Step.IsValid() {
 		return errors.New("Invalid Step")
 	}
-	if m.SecondsSinceStartTime < 0 {
-		return errors.New("Negative SecondsSinceStartTime")
-	}
+
+	// NOTE: SecondsSinceStartTime may be negative
+
 	if (m.Height == 1 && m.LastCommitRound != -1) ||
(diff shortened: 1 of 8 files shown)


consensus/reactor.go
```diff
@@ -1402,9 +1402,9 @@ func (m *NewRoundStepMessage) ValidateBasic() error {
 	if !m.Step.IsValid() {
 		return errors.New("Invalid Step")
 	}
-	if m.SecondsSinceStartTime < 0 {
-		return errors.New("Negative SecondsSinceStartTime")
-	}
+
+	// NOTE: SecondsSinceStartTime may be negative
```
melekes: really? huh

types/block.go
```diff
@@ -78,55 +87,74 @@ func (b *Block) ValidateBasic() error {
 		)
 	}
 
-	// Validate evidence.
-	for i, ev := range b.Evidence.Evidence {
-		if err := ev.ValidateBasic(); err != nil {
-			return fmt.Errorf("Invalid evidence (#%d): %v", i, err)
-		}
+	if b.TotalTxs < b.NumTxs {
```
melekes: that failed for me since we do not set `TotalTxs` in many tests. Maybe we should though

consensus/reactor.go
ebuchman: It probably shouldn't be but I can't guarantee it easily

PR #2739, opened 2018-11-01
Enforce validators can only use the correct pubkey type

ValarDragon:
The first commit on this branch mirrors #2714 before the commit chaos happened.

What remains to be done: 
* [x] Address anton's comment of separating validation and update logic (https://github.com/tendermint/tendermint/pull/2714#discussion_r228706478)
* [x] Address review from #2636 

* [x] Wrote tests


ValarDragon:
I'm kind of confused about the pre-existing situation about how errors are handled in execution.go. If one validator update fails, won't all subsequent validator updates fail? Really we just want that one update to fail. I think that issue is distinct from this one, though I am still confused by it.

state/execution_test.go
```diff
@@ -218,10 +239,15 @@ func TestUpdateValidators(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			err := updateValidators(tc.currentSet, tc.abciUpdates)
+			err1 := validateValidatorUpdates(tc.abciUpdates, tc.validatorParams)
```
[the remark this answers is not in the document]
ValarDragon: I'm not entirely sure. I think these two should only be called in conjunction. That was my reasoning behind testing them together, but I'm happy to change it

types/params.go
```diff
@@ -157,7 +145,7 @@ func (params ConsensusParams) Update(params2 *abci.ConsensusParams) ConsensusPar
 		res.Evidence.MaxAge = params2.Evidence.MaxAge
 	}
 	if params2.Validator != nil {
-		res.Validator.PubKeyTypes = params2.Validator.PubKeyTypes
+		res.Validator.PubKeyTypes = append(params2.Validator.PubKeyTypes[0:0], params2.Validator.PubKeyTypes...)
```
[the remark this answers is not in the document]
ValarDragon: good point! I was trying to avoid an allocation which sets everything to 0's and then a write. I'll add a comment that this is efficiently copying it.

docs/spec/blockchain/state.md
```diff
@@ -85,6 +85,7 @@ type ConsensusParams struct {
 	TxSize
```
ebuchman: Oh damn, this needs to be updated -  TxSize and BlockGossip were removed!

libs/common/string_test.go
```diff
@@ -35,3 +37,20 @@ func TestASCIITrim(t *testing.T) {
...
+func TestStringSliceEqual(t *testing.T) {
+	tests := []struct {
+		a    []string
+		b    []string
+		want bool
+	}{
+		{[]string{"hello", "world"}, []string{"hello", "world"}, true},
+		{[]string{"test"}, []string{"test"}, true},
+		{[]string{"test1"}, []string{"test2"}, false},
+		{[]string{"hello", "world."}, []string{"hello", "world!"}, false},
```
ebuchman: If we're going to go so far as to write these great tests, we should add cases where len(a) != len(b) :)

types/params.go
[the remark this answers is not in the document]
ebuchman: why aren't we just using `copy` ? What effect does this even have? Isn't it copying right back into the same location ?

types/protobuf.go
```diff
@@ -187,20 +187,19 @@ var PB2TM = pb2tm{}
...
+				len(pubKey.Data), ed25519.PubKeyEd25519Size)
 		}
 		var pk ed25519.PubKeyEd25519
 		copy(pk[:], pubKey.Data)
 		return pk, nil
 	case ABCIPubKeyTypeSecp256k1:
-		if len(pubKey.Data) != sizeSecp {
-			return nil, fmt.Errorf("Invalid size for PubKeyEd25519. Got %d, expected %d", len(pubKey.Data), sizeSecp)
+		if len(pubKey.Data) != secp256k1.PubKeySecp256k1Size {
+			return nil, fmt.Errorf("Invalid size for PubKeyEd25519. Got %d, expected %d",
```
ebuchman: Secp

state/execution.go
```diff
@@ -329,19 +355,19 @@ func updateValidators(currentSet *types.ValidatorSet, abciUpdates []abci.Validat
...
 		if valUpdate.VotingPower == 0 {
 			// remove val
 			_, removed := currentSet.Remove(address)
 			if !removed {
 				return fmt.Errorf("Failed to remove validator %X", address)
 			}
-		} else if val == nil {
+			continue
+		}
+		// add or update validator.
```
ebuchman: Why change this from an `else if` ?

state/execution.go
```diff
@@ -318,6 +318,32 @@ func getBeginBlockValidatorInfo(block *types.Block, lastValSet *types.ValidatorS
 
 }
 
+func validateValidatorUpdates(abciUpdates []abci.ValidatorUpdate,
+	consensusValidatorParams types.ValidatorParams) error {
```
ebuchman: Excessive argument name. Why not just `params` ?

state/execution.go
```diff
@@ -318,6 +318,32 @@ func getBeginBlockValidatorInfo(block *types.Block, lastValSet *types.ValidatorS
...
+	for _, valUpdate := range abciUpdates {
+		if valUpdate.GetPower() < 0 {
+			return fmt.Errorf("Voting power can't be negative %v", valUpdate)
+		} else if valUpdate.GetPower() == 0 {
+			continue
```
ebuchman: Shouldn't we still assert a valid key type ?

state/execution.go
```diff
@@ -318,6 +318,32 @@ func getBeginBlockValidatorInfo(block *types.Block, lastValSet *types.ValidatorS
...
+		}
+
+		// Check if validator's pubkey matches an ABCI type in the consensus params
+		thisKeyType := valUpdate.PubKey.Type
+		validKeyType := false
+		for i := 0; i < len(consensusValidatorParams.PubKeyTypes); i++ {
+			if consensusValidatorParams.PubKeyTypes[i] == thisKeyType {
```
ebuchman: So I think it might be nicer to have this as a method on ValidatorParams, like `ValidatorParams.IsValidPubKeyType`

state/execution_test.go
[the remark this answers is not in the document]
ebuchman: probably makes sense to test them separately so we can disentangle what's causing the errors. otherwise it's easy for the tests to accidentally pass for the wrong reason.

types/params.go
```diff
@@ -157,7 +154,9 @@ func (params ConsensusParams) Update(params2 *abci.ConsensusParams) ConsensusPar
 		res.Evidence.MaxAge = params2.Evidence.MaxAge
 	}
 	if params2.Validator != nil {
-		res.Validator.PubKeyTypes = params2.Validator.PubKeyTypes
+		// Copy params2.Validator.PubkeyTypes, and set result's value to the copy.
+		// This avoids having to initialize the slice to 0 values, and then write to it again.
+		res.Validator.PubKeyTypes = append([]string{}, params2.Validator.PubKeyTypes...)
```
melekes: so the simple assignment will result in copy, right? premature optimization?

ValarDragon: The simple assignment will both reference the same underlying data. Your right, this append may be a premature optimization, but only because golang doesn't have a better copy method! This is actually semi-recommended: https://github.com/go101/go101/wiki/How-to-efficiently-clone-a-slice%3F

melekes: No need. Comment is enough

PR #2746, opened 2018-11-02
evidence: Fix iterator leak

kostko:
This fixes a potential iterator leak in `listEvidence`.


kostko:
Ah after rebasing it seems this is already fixed in the `develop` branch (commit 5c6999cf8f0fc09bec6da16d7038a0f1b6ca4d71), so we can close it ;-)

Issue #1571, opened 2018-05-14
rpc: subscribe to begin/endblock tags


ebuchman:
https://github.com/tendermint/abci/pull/237 added tags to begin/end block.

Now we need to allow folks to subscribe to them. This was requested here: https://github.com/tendermint/tendermint/issues/915

Will probably involve a new event type. It should wait until after https://github.com/tendermint/tendermint/issues/847

ebuchman:
Done in #2747 :)

PR #2747, opened 2018-11-02
types: Emit tags from BeginBlock/EndBlock with EventNewBlock

kostko:
See #1571.

This may need some discussion if the design seems sound. Note that this breaks the on-disk format due to including the `BeginBlock` field in the `ABCIResponses` structure.

* [x] Wrote tests


state/execution.go
```diff
@@ -425,7 +426,11 @@ func updateState(
 // Fire TxEvent for every tx.
 // NOTE: if Tendermint crashes before commit, some or all of these events may be published again.
 func fireEvents(logger log.Logger, eventBus types.BlockEventPublisher, block *types.Block, abciResponses *ABCIResponses) {
-	eventBus.PublishEventNewBlock(types.EventDataNewBlock{block})
+	eventBus.PublishEventNewBlock(types.EventDataNewBlock{
+		Block:            block,
+		ResultBeginBlock: *abciResponses.BeginBlock,
```
melekes: Not sure we want to bundle begin/end block results together with the block. The original issue was more about notifying clients about tags in begin block / commit. Blocks can be quite large.

@t-bast @ebuchman thoughts?

kostko: Yeah, that is why I mentioned we need to discuss the design :-)

So the other solution would be a new event type for these.

t-bast: I agree with @melekes. In my scenario I really don't need the whole result, including tags (that I can somehow control from my ABCI app) are enough.

t-bast: It looks like this is going in the right direction, it's just a matter of finding a lightweight way of passing the tags without the block right?

kostko: Yes, if clients don't want to subscribe to full blocks we could just emit an additional (new event type or we could use `EventNewBlockHeader`?) event that doesn't contain the block, but still contains the tags and/or other metadata.

kostko: I think the most correct way would be to emit tags in both `EventNewBlock` and `EventNewBlockHeader` and then clients can decide which to listen on.

kostko:
I've addressed some design comments and rebased on the latest develop branch. PTAL.

state/execution.go
t-bast: I agree with your latest comment. I'll let the Tendermint team comment on the design choices.

kostko:
It seems that `test_apps` test is failing due to an unrelated issue with Debian repositories and should be rerun.

state/execution.go
```diff
@@ -425,8 +426,9 @@ func updateState(
 // Fire TxEvent for every tx.
 // NOTE: if Tendermint crashes before commit, some or all of these events may be published again.
 func fireEvents(logger log.Logger, eventBus types.BlockEventPublisher, block *types.Block, abciResponses *ABCIResponses) {
-	eventBus.PublishEventNewBlock(types.EventDataNewBlock{block})
-	eventBus.PublishEventNewBlockHeader(types.EventDataNewBlockHeader{block.Header})
+	resultTags := append(abciResponses.BeginBlock.Tags, abciResponses.EndBlock.Tags...)
+	eventBus.PublishEventNewBlock(types.EventDataNewBlock{block}, resultTags)
```
melekes: Also, should we make `abciResponses.BeginBlock` and `abciResponses.EndBlock` a part of `EventDataNewBlock` and `EventDataNewBlockHeader`?

melekes: pros: we won't need a separate event for EventDataValidatorSetUpdates
cons: bigger size of NewBlock & NewBlockHeader structs

kostko: Yeah I had this implemented previously, but removed it due to your previous comments. So it can be either way, I think having it in `EventDataNewBlock(Header)` is more consistent with `EventDataTx`.

state/execution.go
```diff
@@ -425,8 +426,9 @@ func updateState(
 // Fire TxEvent for every tx.
 // NOTE: if Tendermint crashes before commit, some or all of these events may be published again.
 func fireEvents(logger log.Logger, eventBus types.BlockEventPublisher, block *types.Block, abciResponses *ABCIResponses) {
-	eventBus.PublishEventNewBlock(types.EventDataNewBlock{block})
-	eventBus.PublishEventNewBlockHeader(types.EventDataNewBlockHeader{block.Header})
+	resultTags := append(abciResponses.BeginBlock.Tags, abciResponses.EndBlock.Tags...)
```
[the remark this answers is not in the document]
kostko: Currently the tag in the end block results would overwrite the tag in the begin block results. Separating these would likely require separate events.

types/event_bus.go
```diff
@@ -185,11 +209,11 @@ func (NopEventBus) UnsubscribeAll(ctx context.Context, subscriber string) error
 	return nil
 }
 
-func (NopEventBus) PublishEventNewBlock(data EventDataNewBlock) error {
+func (NopEventBus) PublishEventNewBlock(data EventDataNewBlock, tags []cmn.KVPair) error {
```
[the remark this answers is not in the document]
kostko: This also depends on the question of whether to include begin/end results in the event data. Because in case we can include them, this extra argument is not required.

state/execution.go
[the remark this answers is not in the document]
melekes: We should probably articulate it in https://github.com/tendermint/tendermint/blob/develop/docs/spec/abci/abci.md

types/event_bus.go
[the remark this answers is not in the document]
melekes: Exactly. Let's me chew on this idea for a day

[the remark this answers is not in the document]
kostko: Any updates?

state/execution.go
melekes: Agree. Let's do that. `Also, should we make abciResponses.BeginBlock and abciResponses.EndBlock a part of EventDataNewBlock and EventDataNewBlockHeader`

kostko: :+1:

state/execution.go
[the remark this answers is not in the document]
kostko: Not sure where I should put this as the "Tags" section doesn't seem to talk about events and I want to double check if documenting it there is appropriate?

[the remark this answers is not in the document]
melekes: "Tags" is the right place. You can just write something like "note begin and end block tags are merged together"

kostko:
Rebased on latest develop, updated docs and pending changelog. PTAL.

CHANGELOG_PENDING.md
```diff
@@ -11,6 +11,7 @@ Friendly reminder, we have a [bug bounty program](https://hackerone.com/tendermi
 ### BREAKING CHANGES:
 
 * CLI/RPC/Config
+  * [state] [\#2747](https://github.com/tendermint/tendermint/pull/2747) Add `BeginBlock` response, breaking the format of `ABCIResponses` structure as encoded on disk.
```
```suggestion
  * [state] [\#2747](https://github.com/tendermint/tendermint/pull/2747) Add BeginBlock & EndBlock responses to ABCIResponses (BREAKING).
```

types/event_bus.go
```diff
@@ -71,12 +71,47 @@ func (b *EventBus) Publish(eventType string, eventData TMEventData) error {
 	return nil
 }
 
+func (b *EventBus) validateAndStringifyTags(tags []cmn.KVPair, logger log.Logger) map[string]string {
+	result := make(map[string]string)
+	for _, tag := range tags {
+		// basic validation
+		if len(tag.Key) == 0 {
+			logger.Info("Got tag with an empty key (skipping)", "tag", tag)
```
melekes:
```suggestion
			logger.Debug("Got tag with an empty key (skipping)", "tag", tag)
```

types/event_bus.go
```diff
@@ -71,12 +71,47 @@ func (b *EventBus) Publish(eventType string, eventData TMEventData) error {
...
+	return result
+}
+
 func (b *EventBus) PublishEventNewBlock(data EventDataNewBlock) error {
-	return b.Publish(EventNewBlock, data)
+	// no explicit deadline for publishing events
+	ctx := context.Background()
+
+	resultTags := append(data.ResultBeginBlock.Tags, data.ResultEndBlock.Tags...)
+	tags := b.validateAndStringifyTags(resultTags, b.Logger.With("block", data.Block))
```
melekes:
```suggestion
	tags := b.validateAndStringifyTags(resultTags, b.Logger.With("block", data.Block.StringShort()))
```

types/event_bus.go
```diff
@@ -71,12 +71,47 @@ func (b *EventBus) Publish(eventType string, eventData TMEventData) error {
...
+	return nil
 }
 
 func (b *EventBus) PublishEventNewBlockHeader(data EventDataNewBlockHeader) error {
-	return b.Publish(EventNewBlockHeader, data)
+	// no explicit deadline for publishing events
+	ctx := context.Background()
+
+	resultTags := append(data.ResultBeginBlock.Tags, data.ResultEndBlock.Tags...)
+	tags := b.validateAndStringifyTags(resultTags, b.Logger.With("header", data.Header))
```
melekes:
```suggestion
	// TODO: create StringShort method for Header and use it in logger
	tags := b.validateAndStringifyTags(resultTags, b.Logger.With("header", data.Header))
```

kostko:
All comments should be addressed.

types/event_bus.go
```diff
@@ -71,12 +71,48 @@ func (b *EventBus) Publish(eventType string, eventData TMEventData) error {
...
+		if len(tag.Key) == 0 {
+			logger.Debug("Got tag with an empty key (skipping)", "tag", tag)
+			continue
+		}
+		result[string(tag.Key)] = string(tag.Value)
+	}
+	return result
+}
+
 func (b *EventBus) PublishEventNewBlock(data EventDataNewBlock) error {
```
ebuchman: Would it maybe be better to have independent subscription events for BeginBlock and EndBlock, rather than stuffing them in with NewBlock/NewBlockHeader like this?

types/events.go
```diff
@@ -54,11 +55,17 @@ func RegisterEventDatas(cdc *amino.Codec) {
 
 type EventDataNewBlock struct {
 	Block *Block `json:"block"`
+
+	ResultBeginBlock abci.ResponseBeginBlock `json:"result_begin_block"`
```
ebuchman: Should we treat these like we treat transactions, as independent events, with EventDataBeginBlock and EventDataEndBlock?

What are the use cases for getting just this data - is it more useful with the block itself or with other results tags from the block?

kostko:
> Not sure if this is the design we should go with - it feels a little adhoc to just add these tags to the NewBlock events. From looking at the fire event sequence for a block:

Well, transaction tags are added to a transaction event, not to a separate event?

What would the hypothetical BeginBlock/EndBlock events contain? Only the ABCI responses? This seems strange, what if you (as a subscriber) want the block or the block header, then you need to maintain additional state?

ebuchman:
> Well, transaction tags are added to a transaction event, not to a separate event?

Right, they're part of the ResponseDeliverTx - I think it makes sense to have a single event per abci response.

> What would the hypothetical BeginBlock/EndBlock events contain? Only the ABCI responses? This seems strange, what if you (as a subscriber) want the block or the block header, then you need to maintain additional state?

Note we have a validator update event - I think we should fold that into an EndBlock event, and we can include the tags there.  If you really think it's necessary, we can put the header in the EndBlock event too, but that's probably over kill.

We can avoid the BeginBlock event for now by putting it's response in the NewBlock and NewBlockHeader events, like you've done. I think that makes sense.

My main gripe here is having the EndBlock event stuff distinct from the existing UpdateValidators event. I think we need to consolidate those.

ebuchman:
> My main gripe here is having the EndBlock event stuff distinct from the existing EventValidatorSetUpdates. I think we need to consolidate those.

Is that right? I'm not sure actually. Will some folks only want to subscribe to the ValidatorUpdates? Or is it fair to say that if you want to know about updates, subscribe to the EndBlock and check every time? I think it would be kind of nice for the Event system to as much as possible match the ABCI structure, unless there are really strong reasons otherwise. 

We should probably write an ADR about all this ...

ebuchman:
Last point: thanks for noting this is a breaking upgrade on the state. Appreciate it. 

We can minimize the impact of that by writing a small upgrade tool that will allow people to upgrade their existing State. It can even be an automatic part of the node, now that we have the software version built into the state! Do you want to take this on?

Thanks!

kostko:
> We can minimize the impact of that by writing a small upgrade tool that will allow people to upgrade their existing State. It can even be an automatic part of the node, now that we have the software version built into the state! Do you want to take this on?

So I switched the field order as you suggested. Not sure how Amino serializes things, does it assign field tags based on field order? In this case the change may already be non-breaking now?

kostko:
> We can avoid the BeginBlock event for now by putting its response in the NewBlock and NewBlockHeader events, like you've done. I think that makes sense.
>
> My main gripe here is having the EndBlock event stuff distinct from the existing EventValidatorSetUpdates. I think we need to consolidate those.

So the NewBlock/NewBlockHeader events would only contain responses and tags from BeginBlock and EventValidatorSetUpdates would only contain responses and tags from EndBlock?

For me, either would be fine actually as long as we can get the block height corresponding to the emitted tags together with the event(s).

@t-bast would this work for your use cases?

t-bast:
Same for me, either would be fine. Thanks guys!

Issue #2379, opened 2018-09-12
docs: improve AppDev guide / organize ABCI documentation


zramsay:
- Can be simplified and make better use of the abci README and specification.md file in the repo
- EndBlock
   - needs an update, no longer true about max +1/3
   - need to stress the change must have gone through consensus
   - explain that you have to return the full pubkey and link to somewhere that specifies what that looks like

original issue: #2143

zramsay:
currently we have:

```
abci/README.md
docs/app-dev/*
docs/spec/abci/*
```

@ebuchman has been moving stuff to the latter location and there is also lots of duplicate information. My proposal is to:

1. slim down `abci/README.md` to a quick overview and links to `docs/abci/spec/`, i.e., remove all "content"
2. continue deprecation/deduplication of a few files in `docs/app-dev/` and focus on practical stuff while linking to `spec/abci` for everything else

zramsay:
#3000

tessr:
@marbar3778 shall we close this one too? a lot of the links are out of date, and I think it's captured by your planned docs refactor

marbar3778:
Yea, lots of the docs have moved/edited and will be edited.

PR #2749, opened 2018-11-02
[docs] improve organization of ABCI docs & fix links

zramsay:
starts #2379, see linked PRs in my fork for the separate pieces

* [ ] Wrote tests


docs/spec/abci/implementations.md
```diff
@@ -0,0 +1,35 @@
+# ABCI Implementations
```
ebuchman: This is great but seems its still largely a duplicate of https://github.com/tendermint/tendermint/blob/develop/docs/spec/abci/client-server.md

Can we just make sure that file contains anything from here it might be missing?

zramsay: fixed

docs/spec/abci/client-server.md
```diff
@@ -24,26 +20,34 @@ For each request, a server should respond with the corresponding
...
-  known as TSP or Teaspoon)
+- In-process (Golang only)
+- ABCI-socket
 - GRPC
 
-Both can be tested using the `abci-cli` by setting the `--abci` flag
+The latter two can be tested using the `abci-cli` by setting the `--abci` flag
 appropriately (ie. to `socket` or `grpc`).
 
+Note that the GRPC version is maintained primarily to simplify onboarding and prototyping and is not receiving the same
```
melekes: Not sure the spec is the right place for such things.

zramsay: removed

Issue #2530, opened 2018-10-02
Remove many fields from the block header


ValarDragon:
Note, we probably no longer have time to do this prelaunch, but (unless I'm missing something), I think it should be done next BlockVersion upgrade. (Hopefully soon)

Currently the BlockHeader contains many Hashes. (See https://github.com/tendermint/tendermint/blob/master/types/block.go#L258) Its unclear to me why most of these are in here, as it gets merkelized anyway. The importance for lite clients comes in _only_ if there is a data availability problem or a knowledge of change problem AFAICT. Otherwise, your just requesting a single additional node internal node in a merkle proof only on the blocks you wish to query. (Saving bandwidth for everyone, including the consensus engine)

The hashes in the blockheader which I want to comment on are:
```
	DataHash       cmn.HexBytes `json:"data_hash"`        // transactions
	ValidatorsHash     cmn.HexBytes `json:"validators_hash"`      // validators for the current block
	NextValidatorsHash cmn.HexBytes `json:"next_validators_hash"` // validators for the next block
	ConsensusHash      cmn.HexBytes `json:"consensus_hash"`       // consensus params for current block
	AppHash            cmn.HexBytes `json:"app_hash"`             // state after txs from the previous block
	LastResultsHash    cmn.HexBytes `json:"last_results_hash"`    // root hash of all results from the txs 
	EvidenceHash    cmn.HexBytes `json:"evidence_hash"`    // evidence included in the block
	ProposerAddress Address      `json:"proposer_address"` // original proposer of the block
```
Below is what I think it should be:
```
	BlockHash       cmn.HexBytes `json:"data_hash"`        // Merkle tree of DataHash, Evidence Hash, ConsensusHash, LastResultsHash
	AppHash            cmn.HexBytes `json:"app_hash"`             // state after txs from the previous block
	NextValidatorsHash cmn.HexBytes `json:"next_validators_hash"` // validators for the next block
	ProposerAddress Address      `json:"proposer_address"` // original proposer of the block
        NumEvidence      int8   // Number of evidence in this block
```
* The reason for removal of ValidatorsHash is due to it being the previous block's NextValidatorsHash. (Ref #2483)
* Lite Clients will most often not be querying the inclusion of a tx by tx hash. Hence the data hash does not need to be in the top layer, and was thus relegated to BlockHash. (We want them querying by tag)
[...]

PR #2752, opened 2018-11-03
Create adr-036-app-data.md

b00f:
* [ ] Wrote tests

### DIFF
--- a/docs/architecture/adr-036-app-data.md
+++ b/docs/architecture/adr-036-app-data.md
@@ -1,0 +1,21 @@
+# ADR 035: Application custom data inside the block header
+
+## Changelog
+
+## Context
+At the moment block header stores application hash (AppHash). It will be very useful if block header can store a custom data (AppData) related to a specific block next to the application hash.
+For example AppData can be the block fee (rewards) or even it can be some flags related to that specific block. This information should be deterministic.
+
+There is an old Pull request related to this ADR [here](https://github.com/tendermint/tendermint/pull/1953).
+
+## Decision
+
+## Status
+
+## Consequences
+
+### Positive
+
+### Negative
+
+### Neutral


ebuchman:
I don't think we should do this, at least not in the header. The Header isn't the right place to store arbitrary information like this. If anything, we want to reduce the header size, and make it more deterministic, for instance https://github.com/tendermint/tendermint/issues/2530.

We could consider putting some app-returned data into the block itself, but we'd have to allocate some space in the block for this, and probably not much. For instance, the block.Data might become:

```
type Data struct {
  Proposer []ProposerData // proposer data included in block by the proposer
  Txs Txs // user data included in block by proposer
  App []AppData // app data included in block by the app
}
```

and we'd set some limit on the size of []AppData and []ProposerData. But all of this is more parameterization and complexity, and especially the AppData part can effectively be done already by the App, so I'm not sure it's worth it right now. The ProposerData is a separate issue.

The app can also return Data in every DeliverTx, but note we don't provide a Data field in EndBlock, just Tags and updates. It would be much simpler to just consider returning the data in EndBlock, and making it queryable (like with tags and other result data), rather than putting it in the block. Can we start with that?

ebuchman:
I'm going to close this for now - happy to see a new one with more details around the various considerations as mentioned above.

b00f:
Thanks for your comment. 
I think we both agree that Tendermint should give some space for the application to save some data inside the block. But where and how?
I do agree that header might not be a good place. But I believe it should pass though EndBlock method (Before committing phase).

I have a question. Can application use Tags for saving data inside the block? For example `tag.Fee = 100`.
I remember when I send the original PR there was no `Tags` in `ResponseEndBlock`. If yes, then we have already what we need!

ebuchman:
Tags aren't stored inside the block itself, but they are indexed in the node and queryable, just like DeliverTx response data (and tags). Is that sufficient, or is there some strong reason why the data must be directly in the block itself?

b00f:
Tendermint as a service should provide more and request less ! I think there will be strong requests by applications for saving some deterministic data (not only txs) inside the block. 
Please keep this silent for the moment. If anyone else has same request later please let me know, then we can discuss more and decide the best for lovey Tendermint project. ;) cheers and thanks

ebuchman:
Ok, thanks @b00f 

>  I think there will be strong requests by applications for saving some deterministic data (not only txs) inside the block.

If you want to provide some of your own rationale for why you think this is the case, would be happy to hear. Maybe this is a good topic for a forum post. We can gather feedback/sentiment there, before it turns into a github issue again.

Thanks!

PR #2760, opened 2018-11-05
autofile/fsdb: comment out permissions check until someone decides to tackle #2285

melekes:
current code results in panic and we certainly don't want that.
https://github.com/tendermint/tendermint/pull/2286#issuecomment-418281846

Refs #2285


Liamsi:
I started working on this a while ago here: https://github.com/tendermint/tendermint/commit/3de3969f7fda8e5d3f65a9c7c11c79a2cf12c2d1
This only fixes the calling code in the mem-pool. The other case (fsdb) doesn't come with a logger in the caller (and I wasn't sure if we need to introduce one there or use some standard way to print to standard out).

PR #2764, opened 2018-11-06
libs/autofile/group: don't panic when checking Head size

melekes:
* [x] Wrote tests (partially)

### DIFF
--- a/consensus/wal.go
+++ b/consensus/wal.go
@@ -13,6 +13,7 @@ import (
 	amino "github.com/tendermint/go-amino"
 	auto "github.com/tendermint/tendermint/libs/autofile"
 	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/libs/log"
 	"github.com/tendermint/tendermint/types"
 	tmtime "github.com/tendermint/tendermint/types/time"
 )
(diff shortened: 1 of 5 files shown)


consensus/wal.go
```diff
@@ -95,6 +96,11 @@ func (wal *baseWAL) Group() *auto.Group {
 	return wal.group
 }
 
+func (wal *baseWAL) SetLogger(l log.Logger) {
```
ebuchman: Do we need to call this from somewhere?

libs/autofile/autofile.go
```diff
@@ -158,23 +168,22 @@ func (af *AutoFile) openFile() error {
...
+// get stats or open file.
+// Opens AutoFile if needed.
 func (af *AutoFile) Size() (int64, error) {
 	af.mtx.Lock()
 	defer af.mtx.Unlock()
 
 	if af.file == nil {
-		err := af.openFile()
-		if err != nil {
-			if err == os.ErrNotExist {
```
ebuchman: Why not keep this?

melekes: see commit msg here  https://github.com/tendermint/tendermint/pull/2764/commits/dca010d8dfed2fa5f533a01361fb935eab410141

consensus/wal.go
melekes: We do call it, already. This method overrides a default one.

PR #2767, opened 2018-11-06
[R4R] Add proposer to NewRound event and proposal info to CompleteProposal event

kevlubkcm:
* [ ] Wrote tests


kevlubkcm:
I was thinking that maybe we should also add the `ProposerAddress` and `ProposerIndex` to the `NewRound` event as well, since we may want to know who the proposer should be before actually receiving the proposal. wdyt? should I include that in this pull request or open a separate one?

ebuchman:
Hmm - can you briefly describe the use case here? We should try to think about the event msgs a bit more holistically before we start adding fields here and there.

kevlubkcm:
Sure. I am trying to build a live tendermint consensus visualizer with all of the critical steps for consensus

essentially a live version of this flow chart https://tendermint.com/docs/assets/img/consensus_logic.e9f4ca6f.png
(with a visual for each validator and each message being passed around)

Really the only thing I am missing is being able to highlight the proposer. Originally I was thinking it makes sense to include that in the `CompleteProposal` event, but that is at the wrong point. Makes more sense to do it at `New Round` and/or `New Height`. (though also it is nice to get the `BlockID` in the `CompleteProposal` event)

ebuchman:
Right of course. I think a NewRound msg makes sense. Currently we just use EventDataRoundState but I don't think we should add the proposer stuff to that. Instead we should probably have a new type, EventDataNewRound, that has the basic H/R/S info as well as a ProposerInfo field that contains the proposer's address and index. How's that sound?

kevlubkcm:
sounds good to me. will update

types/events.go
```diff
@@ -81,12 +82,21 @@ type EventDataRoundState struct {
 	RoundState interface{} `json:"-"`
 }
 
+type EventDataNewRound struct {
+	Height int64  `json:"height"`
+	Round  int    `json:"round"`
+	Step   string `json:"step"`
+	ProposerAddress Address `json:"proposer_address"`
```
ebuchman: What about packaging these in a new `ProposerInfo` type containing `Address` and `Index`. Just thinking it might be more future proof if we change stuff about proposers

kevlubkcm: sounds good. should we make it slightly more generic and call it `ValidatorInfo`? For now it would contain a pair of the existing `Validator` object and an `ValidatorIndex int`?

kevlubkcm: kept it simple for now. let me know if we want to expand

types/events.go
```diff
@@ -80,6 +82,27 @@ type EventDataRoundState struct {
...
+
+	Proposer ValidatorInfo `json:"proposer"`
+}
+
+type EventDataCompleteProposal struct {
+	Height int64  `json:"height"`
+	Round  int    `json:"round"`
+	Step   string `json:"step"`
+
+	BlockID         BlockID `json:"block_id"`
```
ebuchman: Do we want to add the ValidatorInfo to this event as well? Seems it would make sense

consensus/types/round_state.go
```diff
@@ -112,6 +112,41 @@ func (rs *RoundState) RoundStateSimple() RoundStateSimple {
...
+// CompleteProposalEvent returns information about a proposed block as an event.
+func (rs *RoundState) CompleteProposalEvent() types.EventDataCompleteProposal {
+	// We must construct BlockID from ProposalBlock and ProposalBlockParts
+	// cs.Proposal is not guaranteed to be set when this function is called
+	blockId := types.BlockID{
+		Hash:        rs.ProposalBlock.Hash(),
+		PartsHeader: rs.ProposalBlockParts.Header(),
+	}
+
+	edcp := types.EventDataCompleteProposal{
```
ebuchman: No need for the variable here, just return the struct literal directly

consensus/types/round_state.go
```diff
@@ -112,6 +112,41 @@ func (rs *RoundState) RoundStateSimple() RoundStateSimple {
...
+		Height:     rs.Height,
+		Round:      rs.Round,
+		Step:       rs.Step.String(),
+		BlockID:         blockId,
```
ebuchman: Indent seems off.

consensus/common_test.go
```diff
@@ -492,6 +524,31 @@ func ensureVote(voteCh <-chan interface{}, height int64, round int,
 	}
 }
 
+func ensureProposal(proposalCh <-chan interface{}, height int64, round int, propId types.BlockID) {
```
ebuchman: What prompted the need for this if we already have ensureNewProposal?

consensus/common_test.go
```diff
@@ -405,8 +405,24 @@ func ensureNewVote(voteCh <-chan interface{}, height int64, round int) {
...
+		panic("Timeout expired while waiting for NewRound event")
+	case ev := <-roundCh:
+		rs, ok := ev.(types.EventDataNewRound)
+		if !ok {
+			panic(
+				fmt.Sprintf(
+					"expected a EventDataNewRound, got %v.Wrong subscription channel?",
+					reflect.TypeOf(rs)))
+		}
+		if rs.Height != height {
```
ebuchman: We could probably benefit from a `ensureHeightRound(rs, height, round)` function

CHANGELOG_PENDING.md
```diff
@@ -22,6 +22,8 @@ Friendly reminder, we have a [bug bounty program](https://hackerone.com/tendermi
 
 ### FEATURES:
 
+- [eventbus] create new event types for NewRound and CompleteProposal. add proposed block info to CompleteProposal event and proposer info to NewRound event
```
ebuchman: `[types] \#2767 Create ...`

consensus/common_test.go
ebuchman: Oh I see we can't do that because `rs` is a different type each time

consensus/common_test.go
kevlubkcm: allows us to also assert details about the proposal (eg we want to assert *this* specific block is proposed vs *a* block is proposed). I took the pattern from `ensureNewVote` (though it isnt actually used at the moment) vs `ensureVote`

types/events.go
kevlubkcm: I had it there at first, but `EventDataNewRound` seems like the *correct* place to put it. Can add it back if you like.

kevlubkcm:
thanks! responded to leftover comments

consensus/common_test.go
ebuchman: Yep - we probably want to replace ensureNewProposal with this eventually so we're always asserting the correct value. Fine for now.

types/events.go
ebuchman: No worries. Let's leave it as is for now - can always add it later.

Issue #2187, opened 2018-08-08
mempool cache: use a fast hash


ValarDragon:
Tx hashes aren't consensus critical. AFAIK, they are only used to know if your transaction appeared on chain. A second use case being considered is within the mempool to save on memory stored. 

For the latter use-case, it definitely makes sense to use a faster hash. We could use https://en.wikipedia.org/wiki/HighwayHash, which is 30x faster than the standard SHA2. (Source: https://blog.minio.io/highwayhash-fast-hashing-at-over-10-gb-s-per-core-in-golang-fee938b5218a?gi=34fcf777b99e). Its not a cryptographic hash (as it hasn't received enough review / analysis yet), however I believe that the authors are currently analyzing it, and have proven security against differential analysis. Using something fast for the mempool is likely safe, since the worst that can happen is one tx gets marked as a duplicate (and which one gets marked as dupe depends on order received, so its likely that one proposer will eventually get one correct one).

Its significantly harder to find collisions between txs than it is to find collisions in random messages, since signatures bottleneck the amount of txs you can create. Because of this, it may be worth considering even using HighwayHash for normal TxHashes. This is a much more minute point, so we should use the faster hash for the mempool cache at least.

Highway hash is 10x faster than blake2b on a single core. For language support, highway hash has bindings for `C, Golang, Rust, Python3, JS`.

ValarDragon:
Oh good point, your right. Then nevermind, we should just use this suggestion for the mempool cache. But #postlaunch ofc :)

JekaMas:
I have a prototype of this idea about cache generator - https://github.com/JekaMas/nicecache . The example can be seen https://github.com/JekaMas/nicecache/blob/master/example/testvalue_nicecache.go

The main storage is simply preallocated array (not slice) with a bunch of mutexes to reduce lock time. Any key is mapped into an index in the array and simple GC is implemented.

For hashing I used as simple and fast as possible github.com/dgryski/go-farm

ValarDragon:
So sorry for the late response! This looks incredible, and really well designed! I like how theres separate bucket sizes and multiple indices per bucket. I think we should definitely use this for a lot of our caches. For most of the stack its preferrable for its speed properties. Here in particular, it looks way better due to how it handles locking. Having our cache lock on each addition is quite bad atm :)

This probably isn't that significant, but just out of curiosity what is the space overhead of the mutex per bucket? I also don't think we have a need for the `expiredTime` field. Is that something we can not generate?

[...]

PR #2778, opened 2018-11-07
mempool no gossip back

### LINKED ISSUE #1798: mempool: wasteful tx broadcasting
Currently, every valid tx received by a node is sent to each of its peers, regardless of whether that peer has it or not, even if that peer was the original sender of the tx.

This results in a lot of wasted bandwidth and spamming the log message:

``` 
I[06-22|22:07:31.670] Could not check tx                           module=mempool tx=25CF190AB603E698BFFD8FD72F35E28FBE13057E err="Tx already exists in cache"   
```

At the least, we should be able to record which peer originally sent us a tx and not send it back to them.

We should also consider adding a new message type so peers can indicate which txs they already have, though this is more tricky.

ValarDragon:
Closes #1798 

This is done by making every mempool tx maintain a list of peers who its received the tx from. Instead of using the 20byte peer ID, it instead uses a local map from peerID to uint16 counter, so every peer adds 2 bytes. (Word aligned to probably make it 8 bytes)

This also required resetting the callback function on every CheckTx. This likely has performance ramifications for instruction caching. The actual setting operation isn't costly with the removal of defers in this PR. 

* [x] Wrote tests


mempool/mempool.go
```diff
@@ -296,6 +298,10 @@ func (mem *Mempool) TxsWaitChan() <-chan struct{} {
 //     It gets called from another goroutine.
 // CONTRACT: Either cb will get called, or err returned.
 func (mem *Mempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
+	return mem.checkTxFromPeer(tx, cb, 0)
```
melekes: `const unknownPeerID = 0`

mempool/mempool.go
```diff
@@ -177,7 +179,7 @@ func NewMempool(
 	} else {
 		mempool.cache = nopTxCache{}
 	}
-	proxyAppConn.SetResponseCallback(mempool.resCb)
+	proxyAppConn.SetResponseCallback(mempool.resCb(0))
```
```suggestion
	proxyAppConn.SetResponseCallback(mempool.resCb(unknownPeerID))
```

mempool/bench_test.go
```diff
@@ -34,7 +46,7 @@ func BenchmarkCacheInsertTime(b *testing.B) {
 	}
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		cache.Push(txs[i])
+		cache.Push(txs[i], 0)
```
melekes:
```suggestion
		cache.Push(txs[i])
```

mempool/bench_test.go
```diff
@@ -46,7 +58,7 @@ func BenchmarkCacheRemoveTime(b *testing.B) {
 	for i := 0; i < b.N; i++ {
 		txs[i] = make([]byte, 8)
 		binary.BigEndian.PutUint64(txs[i], uint64(i))
-		cache.Push(txs[i])
+		cache.Push(txs[i], 0)
```
melekes:
```suggestion
		cache.Push(txs[i])
```

mempool/mempool.go
```diff
@@ -296,6 +298,10 @@ func (mem *Mempool) TxsWaitChan() <-chan struct{} {
 //     It gets called from another goroutine.
 // CONTRACT: Either cb will get called, or err returned.
 func (mem *Mempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
+	return mem.checkTxFromPeer(tx, cb, 0)
+}
+
+func (mem *Mempool) checkTxFromPeer(tx types.Tx, cb func(*abci.Response), peerID uint16) (err error) {
```
melekes: Would it make sense to create a struct

```go
type txInfo struct {
  PeerID uint16
}
```

and use it instead of raw uint16?

mempool/mempool.go
```diff
@@ -619,7 +631,7 @@ func (memTx *mempoolTx) Height() int64 {
 
 type txCache interface {
 	Reset()
-	Push(tx types.Tx) bool
+	Push(tx types.Tx, peerID uint16) bool
```
melekes: I'd advocate for having 2 methods here

```go
Push(tx types.Tx) bool
PushWithInfo(tx types.Tx, info txInfo) bool
```

or 

```go
Push(tx types.Tx) bool
PushTxFromPeer(tx types.Tx, peerID uint16) bool
```

mempool/reactor.go
```diff
@@ -23,17 +24,27 @@ const (
...
 // MempoolReactor handles mempool tx broadcasting amongst peers.
+// It maintains a map from peer ID to counter, to prevent gossiping txs to the
+// peers you received it from.
 type MempoolReactor struct {
 	p2p.BaseReactor
 	config  *cfg.MempoolConfig
 	Mempool *Mempool
+
+	idMtx   *sync.RWMutex
+	nextID  uint16 // assumes that a node will never have over 65536 peers
```
melekes: `type peerID uint16`?

mempool/reactor.go
```diff
@@ -67,11 +78,18 @@ func (memR *MempoolReactor) GetChannels() []*p2p.ChannelDescriptor {
 // AddPeer implements Reactor.
 // It starts a broadcast routine ensuring all txs are forwarded to the given peer.
 func (memR *MempoolReactor) AddPeer(peer p2p.Peer) {
+	memR.idMtx.Lock()
+	memR.peerMap[peer.ID()] = memR.nextID
+	memR.nextID++
```
melekes: what about peers who constantly drop / reconnect? what will happen when we go beyond 65k?

mempool/mempool.go
melekes: I think this method should be public. Like CheckTx it's a part of the public interface for all the consumers (reactor or anybody else)

mempool/mempool.go
```diff
@@ -366,7 +375,9 @@ func (mem *Mempool) resCbNormal(req *abci.Request, res *abci.Response) {
 				counter:   mem.counter,
 				height:    mem.height,
 				gasWanted: r.CheckTx.GasWanted,
-				tx:        tx,
+				// senders:   []uint16{mem.cache.GetPeerID(tx)},
```
```suggestion
```

mempool/mempool_test.go
```diff
@@ -83,7 +83,7 @@ func TestReapMaxBytesMaxGas(t *testing.T) {
 	mempool := newMempoolWithApp(cc)
 
 	// Ensure gas calculation behaves as expected
-	checkTxs(t, mempool, 1)
+	checkTxs(t, mempool, 1, 0)
```
melekes:
```suggestion
	checkTxs(t, mempool, 1, unknownPeerID)
```

mempool/mempool_test.go
```diff
@@ -117,7 +117,7 @@ func TestReapMaxBytesMaxGas(t *testing.T) {
 		{20, 20000, 30, 20},
 	}
 	for tcIndex, tt := range tests {
-		checkTxs(t, mempool, tt.numTxsToCreate)
+		checkTxs(t, mempool, tt.numTxsToCreate, 0)
```
melekes:
```suggestion
		checkTxs(t, mempool, tt.numTxsToCreate, unknownPeerID)
```

mempool/mempool_test.go
```diff
@@ -157,7 +157,7 @@ func TestMempoolFilters(t *testing.T) {
 	}
 	for tcIndex, tt := range tests {
 		mempool.Update(1, emptyTxArr, tt.preFilter, tt.postFilter)
-		checkTxs(t, mempool, tt.numTxsToCreate)
+		checkTxs(t, mempool, tt.numTxsToCreate, 0)
```
melekes:
```suggestion
		checkTxs(t, mempool, tt.numTxsToCreate, unknownPeerID)
```

mempool/mempool_test.go
```diff
@@ -175,7 +175,7 @@ func TestTxsAvailable(t *testing.T) {
 	ensureNoFire(t, mempool.TxsAvailable(), timeoutMS)
 
 	// send a bunch of txs, it should only fire once
-	txs := checkTxs(t, mempool, 100)
+	txs := checkTxs(t, mempool, 100, 0)
```
melekes:
```suggestion
	txs := checkTxs(t, mempool, 100, unknownPeerID)
```

mempool/mempool_test.go
```diff
@@ -190,7 +190,7 @@ func TestTxsAvailable(t *testing.T) {
 	ensureNoFire(t, mempool.TxsAvailable(), timeoutMS)
 
 	// send a bunch more txs. we already fired for this height so it shouldnt fire again
-	moreTxs := checkTxs(t, mempool, 50)
+	moreTxs := checkTxs(t, mempool, 50, 0)
```
melekes:
```suggestion
	moreTxs := checkTxs(t, mempool, 50, unknownPeerID)
```

mempool/mempool_test.go
```diff
@@ -201,7 +201,7 @@ func TestTxsAvailable(t *testing.T) {
 	ensureNoFire(t, mempool.TxsAvailable(), timeoutMS)
 
 	// send a bunch more txs, it should only fire once
-	checkTxs(t, mempool, 100)
+	checkTxs(t, mempool, 100, 0)
```
melekes:
```suggestion
	checkTxs(t, mempool, 100, unknownPeerID)
```

mempool/mempool_test.go
```diff
@@ -324,7 +324,7 @@ func TestCacheRemove(t *testing.T) {
 		txBytes := make([]byte, 32)
 		rand.Read(txBytes)
 		txs[i] = txBytes
-		cache.Push(txBytes)
+		cache.Push(txBytes, 0)
```
melekes:
```suggestion
		cache.PushTxFromPeer(txBytes, unknownPeerID)
```

mempool/reactor.go
```diff
@@ -23,17 +24,27 @@ const (
...
 
 // MempoolReactor handles mempool tx broadcasting amongst peers.
+// It maintains a map from peer ID to counter, to prevent gossiping txs to the
+// peers you received it from.
 type MempoolReactor struct {
 	p2p.BaseReactor
 	config  *cfg.MempoolConfig
 	Mempool *Mempool
+
+	idMtx   *sync.RWMutex
```
melekes: does it has to be a pointer?

mempool/reactor.go
```diff
@@ -23,17 +24,27 @@ const (
...
 
 // NewMempoolReactor returns a new MempoolReactor with the given config and mempool.
 func NewMempoolReactor(config *cfg.MempoolConfig, mempool *Mempool) *MempoolReactor {
 	memR := &MempoolReactor{
 		config:  config,
 		Mempool: mempool,
+
+		idMtx:   &sync.RWMutex{},
+		peerMap: make(map[p2p.ID]uint16),
+		nextID:  1, // reserve 0 for mempoolReactor.BroadcastTx
```
melekes:
```suggestion
		nextID:  1, // reserve unknownPeerID(0) for mempoolReactor.BroadcastTx
```

mempool/reactor.go
```diff
@@ -88,7 +106,10 @@ func (memR *MempoolReactor) Receive(chID byte, src p2p.Peer, msgBytes []byte) {
 
 	switch msg := msg.(type) {
 	case *TxMessage:
-		err := memR.Mempool.CheckTx(msg.Tx, nil)
+		memR.idMtx.RLock()
+		peerID := memR.peerMap[src.ID()]
+		memR.idMtx.RUnlock()
```
melekes: we have to use `defer` because CheckTx might panic and mempool will be frozen forever.

attach_messages_flow_-_new_page

mempool/mempool.go
ValarDragon: I'd prefer if we spec'd out some more details about what else would go here before making more abstraction. Its not clear to me why we would want time to always live in the cache for example.

mempool/reactor.go
ValarDragon: Thats moreso an argument to just not remove them from the map, which seems like a good idea to me!

mempool/mempool.go
ValarDragon: Good point. I do find it weird to have both CheckTx and CheckTxFromPeer in a public interface. Should we then just have "CheckTxFromPeer" and have UnknownPeerID be exported as well?

mempool/reactor.go
ValarDragon: I thought mutexes being pointers is standard? Its also whats done in: https://gobyexample.com/mutexes

mempool/mempool.go
melekes: > Should we then just have "CheckTxFromPeer" and have UnknownPeerID be exported as well?

Yes!

mempool/reactor.go
melekes: Huh. You're probably right. We need to examine the current code because I am pretty sure I saw mutexes without pointers.

mempool/reactor.go
```diff
@@ -143,6 +167,18 @@ func (memR *MempoolReactor) broadcastTxRoutine(peer p2p.Peer) {
 				continue
 			}
 		}
+		// ensure peer hasn't already sent us this tx
+		skip := false
+		for i := 0; i < len(memTx.senders); i++ {
+			if peerID == memTx.senders[i] {
+				next = next.Next()
```
ebuchman: I think it would be better to not do this here, since we're losing the NextWaitChain logic.

Instead, let's just figure out if skip is true (we should extract this to a hasSender function) and use:

```
if !skip{
  // send memTx
  ...
}

select {
 ...
} 
```

mempool/reactor.go
melekes: Oh, you're right. I am becoming a bit paranoid :)

mempool/reactor.go
melekes: > But what if we're spammed by 65k peers? Won't we overflow?

That was my question basically.

ValarDragon: A reverse map sounds good to me

mempool/reactor.go
ValarDragon: Actually, I'm not sure that we should make a custom type for this if we go through with https://github.com/tendermint/tendermint/pull/2778#discussion_r232207251

mempool/mempool.go
ValarDragon: Not having CheckTx at all looked a bit odd to me, do you still think it should be removed?

mempool/reactor.go
ValarDragon: I believe the same desire is achieved by having TxInfo?

mempool/mempool.go
```diff
@@ -337,6 +349,7 @@ func (mem *Mempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
 	if err = mem.proxyAppConn.Error(); err != nil {
 		return err
 	}
+	mem.proxyAppConn.SetResponseCallback(mem.resCb(txInfo.PeerID))
```
ebuchman: so this works for a the `localClient` where `Async` calls are actually synchronous, but for the `socketClient`, isn't it possible that this could overwrite the response callback for a call that hasn't returned yet? ie. we call `CheckTxWithInfo(tx1, cb, txInfo1)` and then call `CheckTxWithInfo(tx2, cb, txInfo2)` before the response comes back from tx1 - then the callback will have been overwritten and when resCb executes it will use `txInfo2` for processing the response of both `tx1` and `tx2`! Unless I'm missing something, we might need to rethink the approach here...

ValarDragon: My understanding was that every `CheckTxAsync` creates a new ReqRes with the currently known callback. So that ReqRes object knows its own callback, there isn't a global per-connection callback. Looking at socket_client, I have no idea where the callback gets set on that created ReqRes, so your probably right. https://github.com/tendermint/tendermint/blob/dev/mempool_no_gossip_back/abci/client/socket_client.go#L246

ebuchman: there's two callbacks, a global one (set on the proxy app) and a per-call one (set on the reqres). this is setting the global one, but maybe we need to fix the mempool architecture to set it on the reqres

mempool/reactor_test.go
```diff
@@ -123,10 +130,26 @@ func TestReactorBroadcastTxMessage(t *testing.T) {
 
 	// send a bunch of txs to the first reactor's mempool
 	// and wait for them all to be received in the others
-	txs := checkTxs(t, reactors[0].Mempool, NUM_TXS)
+	txs := checkTxs(t, reactors[0].Mempool, NUM_TXS, 0)
```
melekes:
```suggestion
	txs := checkTxs(t, reactors[0].Mempool, NUM_TXS, UnknownPeerID)
```

mempool/mempool.go
```diff
@@ -299,6 +307,10 @@ func (mem *Mempool) TxsWaitChan() <-chan struct{} {
 //     It gets called from another goroutine.
 // CONTRACT: Either cb will get called, or err returned.
 func (mem *Mempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
```
melekes: Should we remove this function?

melekes: Ok, let's leave it then

mempool/mempool.go
ValarDragon: Makes sense, though I still find the way that global one is handled kind of odd. This shouldn't be hard regardless. We already set the reqres callback if a callback is supplied. I can just always set on request and compose the two if a callback is supplied.

ValarDragon: resolved in latest commit

mempool/mempool.go
```diff
@@ -654,17 +684,30 @@ func (cache *mapTxCache) Reset() {
...
+			return false
+		}
+		for i := 0; i < len(memTx.senders); i++ {
+			if txInfo.PeerID == memTx.senders[i] {
+				// TODO: consider punishing peer for dups,
+				// its non-trivial since invalid txs can become valid,
+				// but they can spam the same tx with little cost to them atm.
+				return false
+			}
+		}
```
srmo: shouldn't we return false here too? The TX was in cache, after all. Otherwise we would have duplicates in the cache because the key is now "hash(tx), peerId", or am I reading this wrong?

ValarDragon: Good catch! We should return false here just for it to be cleaner. It won't actually cause a problem since we'd then have two linked list entries and just 1 hashmap entry (the key didn't change for the cache), and when going to delete that second linked list entry, the delete call would just be a no-op. (Deleting non-existant hashmap keys doesn't panic)

mempool/reactor.go
```diff
@@ -20,20 +21,35 @@ const (
...
+// peers you received it from.
 type MempoolReactor struct {
 	p2p.BaseReactor
 	config  *cfg.MempoolConfig
 	Mempool *Mempool
+
+	idMtx     sync.RWMutex
+	nextID    uint16 // assumes that a node will never have over 65536 active peers
+	peerMap   map[p2p.ID]uint16
+	activeIDs map[uint16]bool // used to check if a given peerID key is used, the value doesn't matter
```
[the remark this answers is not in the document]
ValarDragon: Thats a cool trick! I hadn't realized that `struct{}` took up 0 bytes

mempool/mempool.go
```diff
@@ -654,17 +684,30 @@ func (cache *mapTxCache) Reset() {
...
 	cache.mtx.Lock()
 	defer cache.mtx.Unlock()
 
 	// Use the tx hash in the cache
 	txHash := sha256.Sum256(tx)
 	if moved, exists := cache.map_[txHash]; exists {
 		cache.list.MoveToFront(moved)
-		return false
+		memTx, succ := moved.Value.(*mempoolTx)
+		// succ would be false if tx was reaped
```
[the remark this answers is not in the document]
ValarDragon: Your right, I'm not sure why I thought it would do that lol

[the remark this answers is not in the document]
ValarDragon: I realized what happened. In my head, the hashmap was returning a pointer to the main mempool's node for it, not the linked list. So none of this logic works for adding subsequent peers, we have to update the cache to know about the main mempool's tx index in order for that to happen.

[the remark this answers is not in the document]
ValarDragon: For now would it be best to just comment out this section, and merge the rest? (It currently works for not gossiping it back to the first peer who sent it to you, which is still a significant improvement imo)

mempool/mempool.go
```diff
@@ -675,8 +720,8 @@ func (cache *mapTxCache) Push(tx types.Tx) bool {
 			cache.list.Remove(popped)
 		}
 	}
-	cache.list.PushBack(txHash)
-	cache.map_[txHash] = cache.list.Back()
+	e := cache.list.PushFront(txHash)
```
ebuchman: Ok I think something is wrong above. If the tx is already in the cache, we move it to the front, and if its new, it should be added to the front (thanks for the fix here). But if the cache is full, it seems we pop the element from the front - shouldn't we take that from the back?

mempool/reactor.go
```diff
@@ -20,20 +21,35 @@ const (
...
 
 // MempoolReactor handles mempool tx broadcasting amongst peers.
+// It maintains a map from peer ID to counter, to prevent gossiping txs to the
+// peers you received it from.
 type MempoolReactor struct {
 	p2p.BaseReactor
 	config  *cfg.MempoolConfig
 	Mempool *Mempool
+
+	idMtx     sync.RWMutex
```
ebuchman: Might be nicer if this block had its own struct with some methods

mempool/mempool.go
```diff
@@ -337,27 +352,43 @@ func (mem *Mempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
...
 }
 
 // ABCI callback function
-func (mem *Mempool) resCb(req *abci.Request, res *abci.Response) {
-	if mem.recheckCursor == nil {
-		mem.resCbNormal(req, res)
-	} else {
-		mem.metrics.RecheckTimes.Add(1)
-		mem.resCbRecheck(req, res)
+func (mem *Mempool) resCb() abcicli.Callback {
```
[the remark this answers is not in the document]
ebuchman: Also, does it even make sense to have this function do anything? Since we're always setting the callback on the reqRes now, why not just handle the recheck login in there too?

mempool/mempool.go
```diff
@@ -337,27 +349,43 @@ func (mem *Mempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
 	}
 	reqRes := mem.proxyAppConn.CheckTxAsync(tx)
 	if cb != nil {
-		reqRes.SetCallback(cb)
+		composedCallback := func(res *abci.Response) {
+			cb(res)
+			mem.reqResCb(tx, txInfo.PeerID)(res)
```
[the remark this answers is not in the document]
ValarDragon: I'll switch it, just in case it matters somewhere!

mempool/mempool.go
ValarDragon: Oh you're right, great catch! I'll also add a cache test for the LRU property.

mempool/reactor.go
ValarDragon: Good idea!

mempool/mempool.go
melekes: Noticed that too https://github.com/tendermint/tendermint/pull/3036 today. Issue https://github.com/tendermint/tendermint/issues/3035

mempool/mempool.go
```diff
@@ -351,27 +364,43 @@ func (mem *Mempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
...
 
-func (mem *Mempool) resCbNormal(req *abci.Request, res *abci.Response) {
+// ABCI request result callback function incorporates local information, like
+// the peer that sent us this tx, so we can avoid sending it back to the same peer.
+func (mem *Mempool) reqResCb(tx []byte, peerID uint16) func(res *abci.Response) {
+	return func(res *abci.Response) {
+		if mem.recheckCursor != nil {
+			return
+		}
+		mem.resCbNormal(tx, peerID, res)
```
melekes: what about updating the mempool size metric here?

```
mem.metrics.Size.Set(float64(mem.Size()))
```

mempool/reactor.go
```diff
@@ -21,20 +22,35 @@ const (
 	maxTxSize  = maxMsgSize - 8 // account for amino overhead of TxMessage
 
 	peerCatchupSleepIntervalMS = 100 // If peer is behind, sleep this amount
+  
+  // UnknownPeerID is the peer ID to use when running CheckTx when there is
```
melekes: formatting is off

mempool/reactor.go
```diff
@@ -145,13 +190,23 @@ func (memR *MempoolReactor) broadcastTxRoutine(peer p2p.Peer) {
...
 
-		// send memTx
-		msg := &TxMessage{Tx: memTx.tx}
-		success := peer.Send(MempoolChannel, cdc.MustMarshalBinaryBare(msg))
-		if !success {
-			time.Sleep(peerCatchupSleepIntervalMS * time.Millisecond)
-			continue
+		if sendTxToPeer {
+			// send memTx
+			msg := &TxMessage{Tx: memTx.tx}
```
melekes: What if we send an information about the sender of a transaction along with it? (rather then remembering it?)

blank diagram

Just an idea.

ValarDragon: I feel like this makes eclipsing another node too easy. (Just have every tx you gossip include the node to eclipse in its list)

mempool/mempool.go
```diff
@@ -677,16 +708,28 @@ func (cache *mapTxCache) Reset() {
...
-func (cache *mapTxCache) Push(tx types.Tx) bool {
+func (cache *mapTxCache) PushTxWithInfo(tx types.Tx, txInfo TxInfo) bool {
 	cache.mtx.Lock()
 	defer cache.mtx.Unlock()
 
 	// Use the tx hash in the cache
 	txHash := sha256.Sum256(tx)
-	if moved, exists := cache.map_[txHash]; exists {
-		cache.list.MoveToBack(moved)
+	if listEntry, exists := cache.map_[txHash]; exists {
```
milosevic: The commented code is essential for this PR as without this functionality we only remember single sender per tx. So in case we receive tx from N peers we will forward it to N-1 peer. in case it's too complex adding senders list to main mempool list, what if we would keep senders as part of cache? As senders list is relevant only for tx that are in main mempool list we can empty senders list when when we remove tx from the main mempool list.

mempool/reactor.go
```diff
@@ -21,20 +22,35 @@ const (
...
+// It maintains a map from peer ID to counter, to prevent gossiping txs to the
+// peers you received it from.
 type MempoolReactor struct {
 	p2p.BaseReactor
 	config  *cfg.MempoolConfig
 	Mempool *Mempool
+
+	idMtx     sync.RWMutex
+	nextID    uint16 // assumes that a node will never have over 65536 active peers
+	peerMap   map[p2p.ID]uint16
```
milosevic: What is size of p2p.ID, i.e., what is the gain of this indirection?

mempool/reactor.go
```diff
@@ -145,13 +190,23 @@ func (memR *MempoolReactor) broadcastTxRoutine(peer p2p.Peer) {
 			time.Sleep(peerCatchupSleepIntervalMS * time.Millisecond)
 			continue
 		}
+		// ensure peer hasn't already sent us this tx
+		sendTxToPeer := true
+		for _, sender := range memTx.senders {
```
milosevic: senders can maybe be map?

mempool/reactor.go
```diff
@@ -145,13 +190,23 @@ func (memR *MempoolReactor) broadcastTxRoutine(peer p2p.Peer) {
...
-		// send memTx
-		msg := &TxMessage{Tx: memTx.tx}
-		success := peer.Send(MempoolChannel, cdc.MustMarshalBinaryBare(msg))
-		if !success {
-			time.Sleep(peerCatchupSleepIntervalMS * time.Millisecond)
-			continue
+		if sendTxToPeer {
+			// send memTx
+			msg := &TxMessage{Tx: memTx.tx}
+			success := peer.Send(MempoolChannel, cdc.MustMarshalBinaryBare(msg))
```
milosevic: In case we send fails we don't retry to send the tx?

mempool/reactor.go
ebuchman: It's 20 bytes. The gain is to store max 2 bytes with each tx to identify the sender, rather than 20

melekes:
```
WARNING: DATA RACE
Read at 0x00c0013fcd3a by goroutine 183:
  github.com/tendermint/tendermint/mempool.(*MempoolReactor).broadcastTxRoutine()
      /go/src/github.com/tendermint/tendermint/mempool/reactor.go:195 +0x3c7

Previous write at 0x00c0013fcd3a by D[2019-02-27|10:10:49.058] Read PacketMsg                               switch=3 peer=35bc1e3558c182927b31987eeff3feb3d58a0fc5@127.0.0.1
:46552 conn=MConn{pipe} packet="PacketMsg{30:2B06579D0A143EB78F3D3299DE8213A51D4E11FB05ACE4D6A14F T:1}"
goroutine 190:
  github.com/tendermint/tendermint/mempool.(*Mempool).CheckTxWithInfo()
      /go/src/github.com/tendermint/tendermint/mempool/mempool.go:387 +0xdc1
  github.com/tendermint/tendermint/mempool.(*MempoolReactor).Receive()
      /go/src/github.com/tendermint/tendermint/mempool/reactor.go:134 +0xb04
  github.com/tendermint/tendermint/p2p.createMConnection.func1()
      /go/src/github.com/tendermint/tendermint/p2p/peer.go:374 +0x25b
  github.com/tendermint/tendermint/p2p/conn.(*MConnection).recvRoutine()
      /go/src/github.com/tendermint/tendermint/p2p/conn/connection.go:599 +0xcce

Goroutine 183 (running) created at:
D[2019-02-27|10:10:49.058] Send                                         switch=2 peer=1efafad5443abeea4b7a8155218e4369525d987e@127.0.0.1:46193 channel=48 conn=MConn{pipe} m
sgBytes=2B06579D0A146194480ADAE00C2836ED7125FEE65C1D9DD51049
  github.com/tendermint/tendermint/mempool.(*MempoolReactor).AddPeer()
      /go/src/github.com/tendermint/tendermint/mempool/reactor.go:105 +0x1b1
  github.com/tendermint/tendermint/p2p.(*Switch).startInitPeer()
      /go/src/github.com/tendermint/tendermint/p2p/switch.go:683 +0x13b
  github.com/tendermint/tendermint/p2p.(*Switch).addPeer()
      /go/src/github.com/tendermint/tendermint/p2p/switch.go:650 +0x585
  github.com/tendermint/tendermint/p2p.(*Switch).addPeerWithConnection()
      /go/src/github.com/tendermint/tendermint/p2p/test_util.go:145 +0x939
  github.com/tendermint/tendermint/p2p.Connect2Switches.func2()
      /go/src/github.com/tendermint/tendermint/p2p/test_util.go:109 +0x50

I[2019-02-27|10:10:49.058] Added good transaction                       validator=0 tx=43B4D1F0F03460BD262835C4AA560DB860CFBBE85BD02386D83DAC38C67B3AD7 res="&{CheckTx:gas_w
anted:1 }" height=0 total=375
Goroutine 190 (running) created at:
  github.com/tendermint/tendermint/p2p/conn.(*MConnection).OnStart()
      /go/src/github.com/tendermint/tendermint/p2p/conn/connection.go:210 +0x313
  github.com/tendermint/tendermint/libs/common.(*BaseService).Start()
      /go/src/github.com/tendermint/tendermint/libs/common/service.go:139 +0x4df
  github.com/tendermint/tendermint/p2p.(*peer).OnStart()
      /go/src/github.com/tendermint/tendermint/p2p/peer.go:179 +0x56
  github.com/tendermint/tendermint/libs/common.(*BaseService).Start()
      /go/src/github.com/tendermint/tendermint/libs/common/service.go:139 +0x4df
  github.com/tendermint/tendermint/p2p.(*peer).Start()
      <autogenerated>:1 +0x43
  github.com/tendermint/tendermint/p2p.(*Switch).startInitPeer()
```

mempool/mempool.go
```diff
@@ -357,6 +373,17 @@ func (mem *Mempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
...
 	// CACHE
 	if !mem.cache.Push(tx) {
+		// record the sender
+		e, ok := mem.txsMap[sha256.Sum256(tx)]
+		if ok { // tx may be in cache, but not in the mempool
+			memTx := e.Value.(*mempoolTx)
+			if _, loaded := memTx.senders.LoadOrStore(txInfo.PeerID, true); loaded {
+				// TODO: consider punishing peer for dups,
+				// its non-trivial since invalid txs can become valid,
+				// but they can spam the same tx with little cost to them atm.
```
Liamsi: Should we add an issue for this TODO?

melekes: https://github.com/tendermint/tendermint/issues/3445

mempool/cache_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+		cache := mempool.cache.(*mapTxCache)
+		node := cache.list.Front()
+		counter := 0
+		for node != nil {
+			require.NotEqual(t, len(tc.txsInCache), counter,
+				"cache larger than expected on testcase %d", tcIndex)
+
+			nodeVal := node.Value.([sha256.Size]byte)
+			expectedBz := sha256.Sum256([]byte{byte(tc.txsInCache[len(tc.txsInCache)-counter-1])})
```
Liamsi: change to use tmhash?

mempool/cache_test.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+
+		cache := mempool.cache.(*mapTxCache)
+		node := cache.list.Front()
+		counter := 0
+		for node != nil {
+			require.NotEqual(t, len(tc.txsInCache), counter,
+				"cache larger than expected on testcase %d", tcIndex)
+
+			nodeVal := node.Value.([sha256.Size]byte)
```
Liamsi: Now that we launched: #2187 🚀

mempool/mempool.go
```diff
@@ -31,6 +31,12 @@ type PreCheckFunc func(types.Tx) error
 // transaction doesn't require more gas than available for the block.
 type PostCheckFunc func(types.Tx, *abci.ResponseCheckTx) error
 
+// TxInfo are parameters that get passed when attempting to add a tx to the
+// mempool.
+type TxInfo struct {
+	PeerID uint16
```
Liamsi: Would it make sense to use the type `p2p.ID`? If not should we add a comment why this is necessary / better?

mempool/mempool.go
```diff
@@ -150,7 +156,8 @@ type Mempool struct {
 
 	proxyMtx             sync.Mutex
 	proxyAppConn         proxy.AppConnMempool
-	txs                  *clist.CList    // concurrent linked-list of good txs
+	txs                  *clist.CList // concurrent linked-list of good txs
+	txsMap               map[[sha256.Size]byte]*clist.CElement
```
Liamsi: Suggestion: add a comment why this map is useful?

mempool/reactor_test.go
```diff
@@ -102,6 +100,13 @@ func _waitForTxs(t *testing.T, wg *sync.WaitGroup, txs types.Txs, reactorIdx int
 	wg.Done()
 }
 
+// ensure no txs on reactor after some timeout
+func ensureNoTxs(t *testing.T, reactor *MempoolReactor, timeout time.Duration) {
+	// wait for the txs in all mempools
+	time.Sleep(timeout)
+	assert.Equal(t, 0, reactor.Mempool.Size(), "mempool had non-zero size")
```
Liamsi: assert.Zero

mempool/mempool.go
```diff
@@ -381,27 +408,58 @@ func (mem *Mempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
...
+}
+
+func (mem *Mempool) addTx(memTx *mempoolTx) {
+	e := mem.txs.PushBack(memTx)
+	mem.txsMap[sha256.Sum256(memTx.tx)] = e
+	atomic.AddInt64(&mem.txsBytes, int64(len(memTx.tx)))
+	mem.metrics.TxSizeBytes.Observe(float64(len(memTx.tx)))
+}
+
+func (mem *Mempool) resCbNormal(tx []byte, peerID uint16, res *abci.Response) {
```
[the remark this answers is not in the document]
melekes: `resCbFirstTime`?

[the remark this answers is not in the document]
Liamsi: Yes, I think that would be a bit better. I think it's OK to merge without renaming even (as these methods are not exported).

melekes:
250bytes txs for 120 sec. at 500 txs/sec. rate, so total 15MB

total bytes received from 1st node in 4 node localnet

before: 42793967 (43MB)
after:    30003256 (30MB)

total bytes sent to 1st node in 4 node localnet

before: 30569339 (30MB)
after: 19304964 (19MB)

AFTER -> BEFORE
Screenshot 2019-03-25 at 20 47 56
Screenshot 2019-03-25 at 20 48 06

mempool/mempool.go
```diff
@@ -357,6 +380,17 @@ func (mem *Mempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
 
 	// CACHE
 	if !mem.cache.Push(tx) {
+		// record the sender
+		e, ok := mem.txsMap[sha256.Sum256(tx)]
+		if ok { // tx may be in cache, but not in the mempool
+			memTx := e.Value.(*mempoolTx)
```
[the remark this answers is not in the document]
melekes: I guess the comment is confusing... if `ok` here means tx resides in the mempool (txsMap and txs hold transactions). But tx can be only in the cache and not present in the mempool (txsMap and txs). When we've committed a block, txs are removed from the mempool, but not from cache.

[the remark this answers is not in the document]
melekes: > Is this guaranteed to be non nil ?

yes

mempool/reactor.go
```diff
@@ -21,20 +22,82 @@ const (
...
+
+type mempoolIDs struct {
+	mtx       sync.RWMutex
+	peerMap   map[p2p.ID]uint16
+	nextID    uint16              // assumes that a node will never have over 65536 active peers
+	activeIDs map[uint16]struct{} // used to check if a given peerID key is used, the value doesn't matter
+}
+
+// Reserve searches for the next unused ID and assignes it to the peer.
+func (ids *mempoolIDs) ReserveForPeer(peer p2p.Peer) {
```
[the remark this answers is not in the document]
melekes: More expressive?

mempool/reactor.go
```diff
@@ -21,20 +22,82 @@ const (
...
+		ids.nextID++
+		_, idExists = ids.activeIDs[ids.nextID]
+	}
+	curID := ids.nextID
+	ids.nextID++
+	return curID
+}
+
+// Reclaim returns the ID reserved for the peer back to unused pool.
+func (ids *mempoolIDs) Reclaim(peer p2p.Peer) {
```
melekes: we should also remove removedID from all the transactions in the mempool!!!

melekes: ok, because nextID is always increasing until it overflows the chance of that is very low. can be ignored for now

melekes:
```go 	
        // XXX (melekes): commented out for performance reasons
	//
	// Remove the sender if present to avoid a new node joining and not receiving
	// a tx because we think it already has it. For that two things should happen:
	//   1) node has to connect to over 65536 peers during its lifetime
	//   2) tx must reside in the mempool for the above duration
	//
	// for e := memR.Mempool.txs.Front(); e != nil; e = e.Next() {
	// 	memTx := e.Value.(*mempoolTx)
	// 	memTx.senders.Delete(id)
	// }
```

Liamsi: Some context: The above code could  become necessary once 1) and 2) show to happen. It is unclear if this will ever be the case or not.

mempool/mempool.go
```diff
@@ -650,7 +730,7 @@ func (mem *Mempool) recheckTxs(txs []types.Tx) {
 	mem.recheckEnd = mem.txs.Back()
 
 	// Push txs to proxyAppConn
-	// NOTE: resCb() may be called concurrently.
+	// NOTE: reqResCb may be called concurrently.
```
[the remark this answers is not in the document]
melekes: Thanks for clarification 👍

Issue #2773, opened 2018-11-07
p2p: fix logging peer address


ebuchman:
Pointed out in https://github.com/tendermint/tendermint/issues/2721#issuecomment-435326581

> module=p2p peer=0xb78f00

It's because we're passing a function pointer rather than calling the function in: https://github.com/tendermint/tendermint/blob/6e9aee546061423864fef3f9fa332ef37c3dca96/p2p/switch.go#L619

We should just pass the NetAddress() there, no need to call String().

But note even if we do that, we still lose the ID. That's because in https://github.com/tendermint/tendermint/blob/6e9aee546061423864fef3f9fa332ef37c3dca96/p2p/netaddress.go#L316-L321 we expect the protocol to be at the beginning, but it could come after the `@`, eg. `76e35576b8dd1cf8049fea382fda8ac621a9c6ee@tcp://0.0.0.0:26656`. So we need fix this too

ebuchman:
Fixed in #2797

PR #2786, opened 2018-11-08
fix logformat peer address #2773

u5surf:
#2773 
i fix logformat in peer address.
prealse check this PR.

### DIFF
--- a/p2p/netaddress.go
+++ b/p2p/netaddress.go
@@ -176,7 +176,7 @@ func (na *NetAddress) String() string {
 	if na.str == "" {
 		addrStr := na.DialString()
 		if na.ID != "" {
-			addrStr = IDAddressString(na.ID, addrStr)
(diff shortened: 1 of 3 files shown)


p2p/netaddress.go
```diff
@@ -176,7 +176,7 @@ func (na *NetAddress) String() string {
 	if na.str == "" {
 		addrStr := na.DialString()
 		if na.ID != "" {
-			addrStr = IDAddressString(na.ID, addrStr)
+			addrStr = removeProtocolIfDefined(IDAddressString(na.ID, addrStr))
```
melekes: I think what was proposed in the issue is that we fix `removeProtocolIfDefined ` to remove the protocol from any place in the string, not just from the beginning.

ebuchman:
Thanks for this - fixed in https://github.com/tendermint/tendermint/pull/2797/commits/1f832891548d74b1844c6cf3ec11e4d40587b8c0

PR #2792, opened 2018-11-09
fix amino overhead computation for Tx

### LINKED ISSUE #2789: Method of deciding amino overhead of tx in ReapMaxBytesMaxGas is wrong
I wrote a test for amino overhead of tx:
```go
func TestAmino(t *testing.T) {
	txs := []types.Tx{
		[]byte{1, 2, 3},
		[]byte{1, 2, 3},
	}
	var totalBytes int64 = 0
	for _, tx := range txs {
		aminoOverhead := int64(amino.UvarintSize(uint64(len(tx))))

		totalBytes += aminoOverhead + int64(len(tx))
	}

	bz, _ := cdc.MarshalBinary(txs)
	println("len: ", len(txs), "\nexpected length: ", totalBytes, "\nactual length: ", len(bz), "\ndiff: ", (int64(len(bz)) - totalBytes))
}
/*
=== RUN   TestAmino
len:  2 
expected length:  8 
actual length:  11 
diff:  3
--- PASS: TestAmino (0.00s)
PASS
*/
```

The underlying reason is `amino.UvarintSize` forget `encodeFieldNumberAndTyp3` which will be called in marshal list slice. So we add one extra byte in every tx when marshaling real txs and that's why we have a diff in the test.

It will cause consensus failure because the actual block size exceeds the max block size for the error deciding amino overhead of txs.

Liamsi:
adds a method to compute overhead per Tx which also counts the fieldnum / typ3
fixes #2789 

* [x] Wrote tests


mempool/mempool.go
```diff
@@ -88,8 +87,8 @@ func IsPreCheckError(err error) bool {
 func PreCheckAminoMaxBytes(maxBytes int64) PreCheckFunc {
 	return func(tx types.Tx) error {
 		// We have to account for the amino overhead in the tx size as well
-		aminoOverhead := amino.UvarintSize(uint64(len(tx)))
-		txSize := int64(len(tx) + aminoOverhead)
+		aminoOverhead := types.ComputeAminoOverhead(tx, 1)
```
ValarDragon: Can there be a short comment about why the field number is 1 here?

mempool/mempool.go
```diff
@@ -482,7 +481,7 @@ func (mem *Mempool) ReapMaxBytesMaxGas(maxBytes, maxGas int64) types.Txs {
 	for e := mem.txs.Front(); e != nil; e = e.Next() {
 		memTx := e.Value.(*mempoolTx)
 		// Check total size requirement
-		aminoOverhead := int64(amino.UvarintSize(uint64(len(memTx.tx))))
+		aminoOverhead := types.ComputeAminoOverhead(memTx.tx, 1)
```
melekes: https://github.com/tendermint/tendermint/pull/2792/files#r232387921

types/block.go
```diff
@@ -21,6 +21,8 @@ const (
 
 	// MaxAminoOverheadForBlock - maximum amino overhead to encode a block (up to
 	// MaxBlockSizeBytes in size) not including it's parts except Data.
+	// This means it also excludes the overhead for individual transactions.
+	// To compute individual transactions' overhead use types.ComputeAminoOverhead(tx types.Tx, fieldNum int).
```
melekes: ❤️

types/tx.go
```diff
@@ -118,3 +120,9 @@ type TxResult struct {
 	Tx     Tx                     `json:"tx"`
 	Result abci.ResponseDeliverTx `json:"result"`
 }
+
+func ComputeAminoOverhead(tx Tx, fieldNum int) int64 {
```
melekes: I thought you're going to write a godoc comment?)

types/tx.go
```diff
@@ -118,3 +120,18 @@ type TxResult struct {
...
+// ComputeAminoOverhead calculates the overhead for amino encoding a transaction.
+// The overhead consists of varint encoding the field number and the wire type
+// (= length-delimited = 2), and another varint encoding the length of the
+// transaction.
+// The field number can be the field number of the particular transaction, or
+// the field number of the parenting struct that contains the transactions []Tx
+// as a field (this field number is repeated for each contained Tx).
+// If some []Tx are encoded directly (without a parenting struct), the default
+// fieldNum is also 1 (see BinFieldNum in amino.MarshalBinaryBare).
+func ComputeAminoOverhead(tx Tx, fieldNum int) int64 {
```
ValarDragon: Protobuf is so silly :/, the type is getting prepended to every single element of the slice. Why wouldn't they just make a native concept of slices with a length prefix at the beginning :L.

Liamsi: Protobuf can actually behave as you'd expect it. But only for primitive types. `bytes` doesn't seem to fall in this category (as it is length-delimited) ¯\_(ツ)_/¯ 

Otherwise, you could do sth along the lines of:
```proto
message Data {
  repeated bytes Txs = 1 [packed = true];
}
```

ValarDragon: :(, protobuf why you no treat bytes as primitives :(

Good to know though, thank you!

mempool/mempool.go
```diff
@@ -88,8 +87,12 @@ func IsPreCheckError(err error) bool {
 func PreCheckAminoMaxBytes(maxBytes int64) PreCheckFunc {
 	return func(tx types.Tx) error {
 		// We have to account for the amino overhead in the tx size as well
-		aminoOverhead := amino.UvarintSize(uint64(len(tx)))
-		txSize := int64(len(tx) + aminoOverhead)
+		// NOTE: fieldNum = 1 as types.Block.Data contains Txs []Tx as first field.
```
ebuchman: Is the `[]Tx` relevant here? The mempool only ever deals in `Tx`, not `[]Tx`

types/tx.go
```diff
@@ -118,3 +120,18 @@ type TxResult struct {
 	Tx     Tx                     `json:"tx"`
 	Result abci.ResponseDeliverTx `json:"result"`
 }
+
+// ComputeAminoOverhead calculates the overhead for amino encoding a transaction.
+// The overhead consists of varint encoding the field number and the wire type
+// (= length-delimited = 2), and another varint encoding the length of the
```
ebuchman: What does `(= length-delimited = 2)` mean?

mempool/mempool.go
ebuchman: nevermind - we want to compute the overhead for when it appears in the block, not what it might look like in the mempool

ebuchman:
Oops I need to fix the test

ebuchman:
Opened follow-up issue to write better test: https://github.com/tendermint/tendermint/issues/2806

types/tx.go
Liamsi: The wire-type is what we call `Typ3_ByteLength` in amino, or, length-delimited in proto3, or, equal to `2` in both cases. Maybe that comment is too verbose?

ebuchman: Ah I see. Yeh I would have been more explicit, like `(2, in this case, for a length-delimited type)` but that's fine

PR #2799, opened 2018-11-10
2798 fix clieng golang dep

ackratos:
* [ ] Wrote tests


Gopkg.lock
```diff
@@ -348,7 +350,7 @@
     "leveldb/util",
   ]
   pruneopts = "UT"
-  revision = "6b91fda63f2e36186f1c9d0e48578defb69c5d43"
+  revision = "f9080354173f192dfc8821931eacf9cfd6819253"
```
melekes: As we're approaching 1.0 release, we should be mindful about upgrading dependencies. This includes listing and auditing new changes and testing them. I've checked leveldb (minor changes).

Gopkg.toml
```diff
@@ -90,11 +90,6 @@
   name = "github.com/tendermint/btcd"
   revision = "e5840949ff4fff0c56f9b6a541e22b63581ea9df"
 
-# Haven't made a release since 2016.
```
ebuchman: Ideally the Gopkg.toml lists explicit versions of all dependencies.

Nice to see that this project cut new releases recently. Instead of removing this entry outright, can we move it to the section above (for repos that have releases) and pin it to the latest release version? Looks like that would be v0.9.1

ebuchman: > The issue here is we pinned to an outdated version of a dependencies dependency

Oh? `github.com/prometheus/client_golang` is a direct dependency of Tendermint code.

ValarDragon: oh, I didn't know that dep would let you use transitive dependencies correctly! I wonder if theres a way to just block that directly.

Issue #2582, opened 2018-10-09
Enable CORS on RPC API


arturalbov:
Is there any way to enable CORS on Tendermint node RPC API?
I want to have an ability to sign and send transactions to node directly from browser.
I really don't want to start some proxy server
Thanks

jackzampolin:
@hleb-albau That should work! Can you open a PR?

melekes:
PR with a fix/feature was merged to develop. Will be shipped with 0.26.2 release (check the changelog).

Issue #2798, opened 2018-11-10
client_golang dependency is out of date


ackratos:
our cosmos + tendermint project depends on https://github.com/deathowl/go-metrics-prometheus which relies on latest https://github.com/prometheus/client_golang. This caused conflict with tendermint dependency.

As the comment in Gopkg.toml file looks wrong, we should remove client_golang constrain.
```
# Haven't made a release since 2016.
[[constraint]]
  name = "github.com/prometheus/client_golang"
  revision = "ae27198cdd90bf12cd134ad79d1366a6cf49f632"
```

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):
1944d8534b6048666bb73fa3142c666acddb098d

**ABCI app** (name for built-in, URL for self-written if it's publicly available):

**Environment**:
- **OS** (e.g. from /etc/os-release):
- **Install tools**:
- **Others**:

**What happened**:

**What you expected to happen**:

**Have you tried the latest version**: yes/no

**How to reproduce it** (as minimally and precisely as possible):

**Logs (paste a small part showing an error (< 10 lines) or link a pastebin, gist, etc. containing more of the log file)**:

**Config (you can paste only the changes you've made)**:

**node command runtime flags**:

**`/dump_consensus_state` output for consensus bugs**

**Anything else we need to know**:

ackratos:
PR https://github.com/tendermint/tendermint/pull/2799 is raised for fixing this issue

PR #2800, opened 2018-11-10
2582 Enable CORS on RPC API 

hleb-albau:
ref https://github.com/tendermint/tendermint/issues/2582

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -41,6 +41,10 @@
   version = "=1.2.0"
 
 [[constraint]]
+  name = "github.com/rs/cors"
+  version = "=1.6.0"
+
+[[constraint]]
   name = "github.com/pkg/errors"
   version = "=0.8.0"
 
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -42,7 +42,7 @@
 
 [[constraint]]
   name = "github.com/rs/cors"
-  version = "=1.6.0"
+  version = "1.6.0"
 
 [[constraint]]
   name = "github.com/pkg/errors"
--- a/config/config.go
+++ b/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"github.com/rs/cors"
 	"os"
 	"path/filepath"
 	"time"
@@ -242,6 +243,18 @@ type RPCConfig struct {
 	// TCP or UNIX socket address for the RPC server to listen on
 	ListenAddress string `mapstructure:"laddr"`
 
+	// A list of origins a cross-domain request can be executed from.
+	// If the special '*' value is present in the list, all origins will be allowed.
+	// An origin may contain a wildcard (*) to replace 0 or more characters (i.e.: http://*.domain.com).
+	// Only one wildcard can be used per origin.
+	AllowedOrigins []string `mapstructure:"allowed_origin"`
+
+	// A list of methods the client is allowed to use with cross-domain requests.
+	AllowedMethods []string `mapstructure:"allowed_methods"`
+
+	// A list of non simple headers the client is allowed to use with cross-domain requests.
+	AllowedHeaders []string `mapstructure:"allowed_headers"`
+
 	// TCP or UNIX socket address for the gRPC server to listen on
 	// NOTE: This server only supports /broadcast_tx_commit
 	GRPCListenAddress string `mapstructure:"grpc_laddr"`
@@ -269,8 +282,10 @@ type RPCConfig struct {
 // DefaultRPCConfig returns a default configuration for the RPC server
 func DefaultRPCConfig() *RPCConfig {
 	return &RPCConfig{
-		ListenAddress: "tcp://0.0.0.0:26657",
-
+		ListenAddress:          "tcp://0.0.0.0:26657",
+		AllowedOrigins:         []string{},
+		AllowedMethods:         []string{"HEAD", "GET", "POST"},
+		AllowedHeaders:         []string{"Origin", "Accept", "Content-Type", "X-Requested-With", "X-Server-Time"},
 		GRPCListenAddress:      "",
 		GRPCMaxOpenConnections: 900,
 
@@ -298,6 +313,20 @@ func (cfg *RPCConfig) ValidateBasic() error {
 		return errors.New("max_open_connections can't be negative")
 	}
 	return nil
+}
+
+// returns true if cors is enabled.
+func (cfg *RPCConfig) IsCorsEnabled() bool {
+	return len(cfg.AllowedOrigins) != 0
+}
+
+// returns cors options for RPC endpoint
+func (cfg *RPCConfig) CorsOptions() cors.Options {
+	return cors.Options{
+		AllowedOrigins: cfg.AllowedOrigins,
+		AllowedMethods: cfg.AllowedMethods,
+		AllowedHeaders: cfg.AllowedHeaders,
+	}
 }
 
 //-----------------------------------------------------------------------------
--- a/config/config.go
+++ b/config/config.go
@@ -247,7 +247,7 @@ type RPCConfig struct {
 	// If the special '*' value is present in the list, all origins will be allowed.
 	// An origin may contain a wildcard (*) to replace 0 or more characters (i.e.: http://*.domain.com).
 	// Only one wildcard can be used per origin.
-	CORSAllowedOrigins []string `mapstructure:"cors_allowed_origin"`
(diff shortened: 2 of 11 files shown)


Gopkg.toml
```diff
@@ -40,6 +40,10 @@
   name = "github.com/gorilla/websocket"
   version = "=1.2.0"
 
+[[constraint]]
+  name = "github.com/rs/cors"
+  version = "=1.6.0"
```
melekes: I think we can lessen the version requirement here. 

@ebuchman What is the reason for tendermint requiring precise versions again? This results in great pain when integrating TM (see https://github.com/tendermint/tendermint/issues/2798). Versions are already frozen in `.lock` file. Plus it's possible to update one dependency without updating the other one.

config/config.go
```diff
@@ -242,6 +243,18 @@ type RPCConfig struct {
 	// TCP or UNIX socket address for the RPC server to listen on
 	ListenAddress string `mapstructure:"laddr"`
 
+	// A list of origins a cross-domain request can be executed from.
+	// If the special '*' value is present in the list, all origins will be allowed.
+	// An origin may contain a wildcard (*) to replace 0 or more characters (i.e.: http://*.domain.com).
+	// Only one wildcard can be used per origin.
+	AllowedOrigins []string `mapstructure:"allowed_origin"`
```
melekes: Should we prefix these options with `CORS`?

config/config_test.go
```diff
@@ -36,3 +36,8 @@ func TestConfigValidateBasic(t *testing.T) {
 	cfg.Consensus.TimeoutPropose = -10 * time.Second
 	assert.Error(t, cfg.ValidateBasic())
 }
+
+func TestRpcConfigCorsDisabledByDefault(t *testing.T) {
```
melekes: Don't think we need this.

rpc/client/rpc_test.go
```diff
@@ -32,6 +33,19 @@ func GetClients() []client.Client {
 	}
 }
 
+// Make sure status is correct (we connect properly)
```
melekes: Shouldn't we instead test if CORS is enabled, headers are present and everything works (i.e. HAPPY PATH)?

Gopkg.toml
hleb-albau: I think so too. Only one reason I did precise version - all versions in toml are precise. I will change to **^1.6.0**. Also, do you plan to use go modules?

config/config.go
hleb-albau: Both in go and in config?

rpc/client/rpc_test.go
hleb-albau: To test Happy path, we need to run another tendermint instance with cors enabled. Is it okey to do in single test?

config/config.go
melekes: Yes. They only make sense for cors, right?

Gopkg.toml
ebuchman: It's to prevent dependencies from being upgraded uncontrollably/accidentally.

The "great pain" should be overcomeable by using [override], no?

hleb-albau: Finished with plain **1.6.0**. I think, it's better to revise dependencies versions in separate issue.

hleb-albau:
@melekes done!

ebuchman:
Can we also merge in the latest develop and add a `FEATURE` entry in the `CHANEGLOG_PENDING.md` please (should be under v0.26.2)?

hleb-albau:
> v0.26.2

Should I create v0.26.2 version by myself? Right now there is only v0.26.1

Gopkg.toml
melekes: > uncontrollably/accidentally

I think this can easily be prevented by a) using `dep ensure -update XXX` to update single dependency b) review process.

> The "great pain" should be overcomeable by using [override], no?

I want for our users to have **no pain by default** We shouldn't put our problems on their shoulders.

ebuchman: I somehow didn't know about `dep ensure -update XXX`. Is that new? 

What do you recommend we change this too instead of `=` ?

melekes: > I somehow didn't know about dep ensure -update XXX. Is that new?

no. it existed as long as I remember dep

> What do you recommend we change this too instead of = ?

Depending on a library. I think sticking with `^` major version is fine for most of the libs. Some require locking onto minor version '~`

config/config.go
```diff
@@ -300,6 +315,20 @@ func (cfg *RPCConfig) ValidateBasic() error {
 	return nil
 }
 
+// returns true if cors is enabled.
```
melekes:
```suggestion
// IsCorsEnabled returns true if cross-origin resource sharing is enabled.
```

rpc/core/doc.go
```diff
@@ -12,7 +12,10 @@ See it here: https://github.com/tendermint/tendermint/tree/master/rpc/lib
 
 ## Configuration
 
-Set the `laddr` config parameter under `[rpc]` table in the `$TMHOME/config/config.toml` file or the `--rpc.laddr` command-line flag to the desired protocol://host:port setting.  Default: `tcp://0.0.0.0:26657`.
+RPC can be configured by tuning parameters under `[rpc]` table in the `$TMHOME/config/config.toml` file or by using the `--rpc.X` command-line flags.
+
+Default rpc listen address is `tcp://0.0.0.0:26657`. To set another address,  set the `laddr` config parameter to desired value.
+Cors can be enabled by setting `cors_allowed_origins`, `cors_allowed_methods`, `cors_allowed_headers` config parameters.
```
melekes:
```suggestion
CORS (Cross-Origin Resource Sharing) can be enabled by setting `cors_allowed_origins`, `cors_allowed_methods`, `cors_allowed_headers` config parameters.
```

config/config.go
```diff
@@ -300,6 +315,20 @@ func (cfg *RPCConfig) ValidateBasic() error {
...
+func (cfg *RPCConfig) IsCorsEnabled() bool {
+	return len(cfg.CORSAllowedOrigins) != 0
+}
+
+// returns cors options for RPC endpoint
+func (cfg *RPCConfig) CorsOptions() cors.Options {
```
melekes: I'd prefer not to create this method because it ties our config API with a third-party library API.

node/node.go
```diff
@@ -651,9 +652,16 @@ func (n *Node) startRPC() ([]net.Listener, error) {
 		wm.SetLogger(rpcLogger.With("protocol", "websocket"))
 		mux.HandleFunc("/websocket", wm.WebsocketHandler)
 		rpcserver.RegisterRPCFuncs(mux, rpccore.Routes, coreCodec, rpcLogger)
+
+		var rootHandler http.Handler = mux
+		if n.config.RPC.IsCorsEnabled() {
+			corsMiddleware := cors.New(n.config.RPC.CorsOptions())
```
```suggestion
			corsMiddleware := cors.New(cors.Options{
				AllowedOrigins: n.config.RPC.CORSAllowedOrigins,
				AllowedMethods: n.config.RPC.CORSAllowedMethods,
				AllowedHeaders: n.config.RPC.CORSAllowedHeaders,
			})
```

rpc/test/helpers.go
```diff
@@ -84,6 +84,7 @@ func GetConfig() *cfg.Config {
 		tm, rpc, grpc := makeAddrs()
 		globalConfig.P2P.ListenAddress = tm
 		globalConfig.RPC.ListenAddress = rpc
+		globalConfig.RPC.CORSAllowedOrigins = []string{"*"}
```
melekes: let's limit it to only tendermint.com

melekes:
Awesome work @hleb-albau! Thanks for contributing to Tendermint 🚀

config/toml.go
```diff
@@ -119,6 +119,17 @@ filter_peers = {{ .BaseConfig.FilterPeers }}
...
+# A list of origins a cross-domain request can be executed from
+# Default value '[]' disables cors support
+# Use '["*"]' to allow any origin
+cors_allowed_origins = "{{ .RPC.CORSAllowedOrigins }}"
+
+# A list of methods the client is allowed to use with cross-domain requests
+cors_allowed_methods = "{{ .RPC.CORSAllowedMethods }}"
+
+# A list of non simple headers the client is allowed to use with cross-domain requests
+cors_allowed_headers = "{{ .RPC.CORSAllowedHeaders }}"
```
zramsay: this file should be updated https://github.com/tendermint/tendermint/blob/master/docs/tendermint-core/configuration.md

PR #2811, opened 2018-11-12
Handling integer IDs in JSON-RPC requests -- fixes #2366

### LINKED ISSUE #2366: tendermint rpc server does not accept integer id for request
JSONRPC2 specification allows to specify request ID as a number: https://www.jsonrpc.org/specification#request_object

> An identifier established by the Client that MUST contain a String, Number, or NULL value if included.

However tendermint rpc server will return an error if ID is anything but a string.

tomtau:
Addresses https://github.com/tendermint/tendermint/issues/2366

* [x] Wrote tests

Hi everyone, we're using one JSON-RPC client library that only does integer IDs (incrementally), so rather than having a workaround for that, I guess it'd be better to fix this issue in Tendermint.
This is the first Go code I ever wrote, so apologies if it's too bad.
I decided to use this "interface wrapper with a single method" workaround for the lack of sum types in Go, as it seems to be an OK tradeoff between type safety (e.g. "empty interface" wrapper would work as well here, but passing in booleans / objects / arrays as IDs wouldn't result in an error) and boilerplate.


rpc/lib/types/types.go
```diff
@@ -13,17 +13,63 @@ import (
 	tmpubsub "github.com/tendermint/tendermint/libs/pubsub"
 )
 
+// a wrapper to emulate a sum type: jsonrpcid = string | int
+// TODO: refactor when Go 2.0 arrives https://github.com/golang/go/issues/19412
```
ebuchman: LOL

rpc/lib/types/types.go
```diff
@@ -13,17 +13,63 @@ import (
...
+
+func (JSONRPCIntID) isJSONRPCID() {}
+
 //----------------------------------------
 // REQUEST
 
 type RPCRequest struct {
 	JSONRPC string          `json:"jsonrpc"`
-	ID      string          `json:"id"`
+	ID      jsonrpcid       `json:"id"`
```
ebuchman: One interesting thing to note is that the `json` pkg can natively unmarshal into empty interfaces, but not into interfaces with methods. Crazy, I know, it's one of the main reasons we invented Amino.

In any case, I like the type strength here, but when we actually unmarshal it might be nicer to unmarshal into an anonymous struct like:

```
struct {
	JSONRPC string          `json:"jsonrpc"`
	ID      interface{}       `json:"id"`
	Method  string          `json:"method"`
	Params  json.RawMessage `json:"params"`
}
```

Then we can populate a proper RPCRequest with these fields, switching on the ID to see if we should use JSONRPCIntID or JSONRPCStringID.

I like this better than using the errors because we don't know if the error was from a real failure to unmarshal, or just from using a different ID type, and then we blindly just try again. So marshalling into the interface{} and checking the type feels more correct.

What do you think?

rpc/lib/types/types_test.go
```diff
@@ -19,20 +19,41 @@ func TestResponses(t *testing.T) {
...
+	l := `{"jsonrpc":"2.0","id":1,"result":{"Value":"hello"}}`
+	assert.Equal(string(l), string(k))
+}
+
+func TestUnmarshallResponses(t *testing.T) {
+	assert := assert.New(t)
+	cdc := amino.NewCodec()
+	var err error
+	response := &RPCResponse{}
+	err = json.Unmarshal([]byte(`{"jsonrpc":"2.0","id":"1","result":{"Value":"hello"}}`), response)
```
ebuchman: Mind writing a [table driven test](https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go) for this? We can try multiple strings ("", "1", "alphabet", something with non-ascii, etc.), multiple ints (-1, 0, 1, 100, etc.), and ensure we get failures for things that aren't strings or ints. Technically it should allow negative ints as IDs as well (according to the JSON spec for "Numbers").

Can we also add tests to the client and/or server ensuring that we get Responses with the same ID as sent in the Request, both for strings and ints?

CHANGELOG_PENDING.md
```diff
@@ -26,3 +26,4 @@ Friendly reminder, we have a [bug bounty program](https://hackerone.com/tendermi
 
 ### BUG FIXES:
 
+* Handling integer IDs in JSON-RPC requests
```
ebuchman:
```
- [rpc] \#2811 Allow integer IDs in JSON-RPC requests
```

rpc/lib/types/types.go
tomtau: Yeah, it's a bit unfortunate accidental complexity

CHANGELOG_PENDING.md
tomtau: Done: https://github.com/tendermint/tendermint/pull/2811/commits/71fe74dbc763a08a832ac936c26ed18e7950bd58#diff-2df96dd4ec5f206ab10b01eae20b3b21

rpc/lib/types/types.go
tomtau: I checked `Allow edits from maintainers`, so feel free to push to the `fix/rpc-integer-id` branch on my fork.

tomtau: From the Go style guides I've seen, they usually recommend avoiding empty interfaces, but since this is the first Go code I ever wrote, I don't have any strong opinion for or against that.

rpc/lib/types/types.go
```diff
@@ -13,17 +13,63 @@ import (
...
 	Method  string          `json:"method"`
 	Params  json.RawMessage `json:"params"` // must be map[string]interface{} or []interface{}
 }
 
-func NewRPCRequest(id string, method string, params json.RawMessage) RPCRequest {
+// UnmarshalJSON custom JSON unmarshalling due to jsonrpcid being string or int
+func (request *RPCRequest) UnmarshalJSON(data []byte) error {
+	type AliasStr RPCRequest
```
melekes: We should Unmarshal interface{} and then use switch for different possible types

```go
switch ID.(type) {
  case string:
           XXX
  case int:
           YYY
  default:
    return fmt.Errorf("Unknown ID type. Got %T, expected string or int", ID)
```

tomtau: @melekes https://github.com/tendermint/tendermint/pull/2811#discussion_r232934518 ?

melekes: yes! also you can do without intermediate struct

```go
unsafeResp := &struct {
            JSONRPC string          `json:"jsonrpc"`
            ID      interface{}     `json:"id"`
            Method  string          `json:"method"`
            Params  json.RawMessage `json:"params"` // must be map[string]interface{} or []interface{}
        }{}
```

you just use the original one

rpc/lib/types/types.go
tomtau: changed here: https://github.com/tendermint/tendermint/pull/2811/commits/f2a71d7b9dd636816c0da35384779390ff1a3338#diff-f595cd0be03b00ec17ec43d65afb7780 -- needed nil check, since ID may not be provided (e.g. in ws_client_test)

rpc/lib/types/types.go
tomtau: @melekes @ebuchman I've changed it here: https://github.com/tendermint/tendermint/pull/2811/commits/f2a71d7b9dd636816c0da35384779390ff1a3338#diff-f595cd0be03b00ec17ec43d65afb7780

I'm not that experienced with Go, so not sure how to do it without the intermediate struct -- or did you mean that `jsonrpcid` will be an empty interface and it should get rid of the extra ID type wrappers (`JSONRPCStringID`/`JSONRPCIntID`)?

Issue #2815, opened 2018-11-12
Tendermint starts with inconsistent genesis.json


srmo:
**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source): 0.26.1

**ABCI app** (name for built-in, URL for self-written if it's publicly available): private

**Environment**:
- **OS** (e.g. from /etc/os-release): linux / `Linux test2 3.16.0-4-amd64 #1 SMP Debian 3.16.39-1+deb8u2 (2017-03-07) x86_64 GNU/Linux`
- **Install tools**: ansible
- **Others**: 3 nodes, 2 validators. `create_empty_blocks=false`

**What happened**:
Starting a fresh chain, using a genesis file coming from 0.22.8. Migrated the "consensus_param" section.
Tendermint starts and builds first empty block. Then logs errors
```
2018-11-12_15:37:28.75853 E[12116-11-12|15:37:28.758] Peer sent us invalid msg                     module=consensus peer="Peer{MConn{192.168.55.12:38432} ec325039be4a8997c5e1c740870d7f0c29b16cca in}" msg="[HEARTBEAT Heartbeat{-1:7C31F891F553 2/00 (293) /215B9D2937A8.../}]" err="Negative ValidatorIndex"
```

Looking at the tendermint code, it is assumed that this happens because the genesis.json has no "address" field for any validator.

**What you expected to happen**:
Tendermint should either not start because the address field is missing or not throw errors.

**Have you tried the latest version**: yes/no
latest release

**How to reproduce it** (as minimally and precisely as possible):
init 0.26.1 and remove address field from genesis

**Logs (paste a small part showing an error (< 10 lines) or link a pastebin, gist, etc. containing more of the log file)**:

srmo:
Hint: If I read types/genesis.go:86 correctly, the could would add the pubKeys address as validator key, if it is not present in the genesis.
So the error I'm seeing must have another cause, just reproduced with the address added.
See attached genesis
[genesis.txt](https://github.com/tendermint/tendermint/files/2572847/genesis.txt)

ebuchman:
It looks like the non-validator peer thinks he's a validator for some reason? The message is the result of him thinking it's his turn to propose but having nothing to propose yet ...

srmo:
I'll tripple check its configuration. Will come back to it tomorrow. Thanks!

srmo:
[...]

PR #2819, opened 2018-11-12
#2815 do not broadcast heartbeat proposal when we are non-validator

srmo:
Refs #2815

* [x] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -34,7 +34,10 @@ program](https://hackerone.com/tendermint).
 
 ### IMPROVEMENTS:
 
+- [rpc] \#2780 Add read and write timeouts to HTTP servers
+- [state] \#2765 Make "Update to validators" msg value pretty (@danil-lashin)
 - [p2p] \#2857 "Send failed" is logged at debug level instead of error.
 
 ### BUG FIXES:
-2518 - fixes a bug where non-validators will send proposalHearbeats when `create_empty_blocks=false` or `create_empty_blocks_interval > 0`
+- [consensus] \#2819 Don't send proposalHearbeat if not a validator
+- [p2p] \#2869 Set connection config properly instead of always using default
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -65,7 +65,7 @@
 
 [[constraint]]
   name = "google.golang.org/grpc"
-  version = "~1.13.0"
+  version = "^1.13.0"
 
 [[constraint]]
   name = "github.com/fortytw2/leaktest"
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -53,6 +53,6 @@ Vagrant.configure("2") do |config|
 
     # get all deps and tools, ready to install/test
     su - vagrant  -c 'source /home/vagrant/.bash_profile'
-    su - vagrant -c 'cd /home/vagrant/go/src/github.com/tendermint/tendermint && make get_tools && make get_vendor_deps'
+    su - vagrant -c 'cd /home/vagrant/go/src/github.com/tendermint/tendermint && make get_tools && make get_dev_tools && make get_vendor_deps'
   SHELL
 end
--- a/abci/client/grpc_client.go
+++ b/abci/client/grpc_client.go
@@ -54,7 +54,7 @@ RETRY_LOOP:
 			if cli.mustConnect {
 				return err
 			}
-			cli.Logger.Error(fmt.Sprintf("abci.grpcClient failed to connect to %v.  Retrying...\n", cli.addr))
+			cli.Logger.Error(fmt.Sprintf("abci.grpcClient failed to connect to %v.  Retrying...\n", cli.addr), "err", err)
 			time.Sleep(time.Second * dialRetryIntervalSeconds)
 			continue RETRY_LOOP
 		}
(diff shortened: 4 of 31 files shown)


srmo:
Sorry for the noisy cosmetic commits

srmo:
Seems as if I'm not able to understand how to write correct tests in your framework.
I have a vague idea how to write the test that waits for the Event, but I would have no idea on how to write the negative test with your `ensure` methods.
Help appreciated.

srmo:
OK, fixed the positive case. The negative case still looks off.

consensus/common_test.go
```diff
@@ -409,6 +409,20 @@ func ensureNewRound(roundCh <-chan interface{}, height int64, round int) {
 		"Timeout expired while waiting for NewRound event")
 }
 
+func ensureProposalHeartbeat(heartBeatCh <-chan interface{}) {
+	select {
+	case <-time.After(ensureTimeout):
+		panic("Timeout expired while waiting for ProposalHearbeat event")
+	case ev := <-heartBeatCh:
+		block, ok := ev.(types.EventDataProposalHeartbeat)
```
melekes: `block` -> `heartbeat`

consensus/common_test.go
```diff
@@ -409,6 +409,20 @@ func ensureNewRound(roundCh <-chan interface{}, height int64, round int) {
 		"Timeout expired while waiting for NewRound event")
 }
 
+func ensureProposalHeartbeat(heartBeatCh <-chan interface{}) {
```
melekes: `heartBeatCh` -> `heartbeatCh`

consensus/state.go
```diff
@@ -806,8 +806,14 @@ func (cs *ConsensusState) needProofBlock(height int64) bool {
 }
 
 func (cs *ConsensusState) proposalHeartbeat(height int64, round int) {
-	counter := 0
+	logger := cs.Logger.With("height", height, "round", round)
 	addr := cs.privValidator.GetAddress()
+
+	if !cs.Validators.HasAddress(addr) {
+		logger.Debug("Not sending proposalHearbeat. This node is not a validator", "addr", addr, "vals", cs.Validators)
```
melekes: not necessary, I think

```suggestion
```

consensus/state_test.go
```diff
@@ -1028,6 +1030,31 @@ func TestSetValidBlockOnDelayedPrevote(t *testing.T) {
...
+// regression for #2518
+func TestNoHearbeatWhenNotValidator(t *testing.T) {
+	cs, _ := randConsensusState(4)
+	cs.Validators = types.NewValidatorSet(nil) // make sure we are not in the validator set
+
+	cs.evsw.AddListenerForEvent("testing", types.EventProposalHeartbeat,
+		func(data tmevents.EventData) {
+			t.Errorf("Should not have broadcasted heartbeat")
+		})
+	go cs.proposalHeartbeat(10, 1)
```
melekes: do we need to run this in a goroutine?

consensus/state_test.go
```diff
@@ -1028,6 +1030,31 @@ func TestSetValidBlockOnDelayedPrevote(t *testing.T) {
...
+
+	cs.Stop()
+	time.Sleep(proposalHeartbeatIntervalSeconds * time.Second)
```
melekes: why do we need this timeout?

consensus/state.go
srmo: As "proposalHeartbeat" is a debugging mechanism, the idea here is to also aid you in understanding why a broadcast has not been sent.

consensus/state_test.go
srmo: I have no idea. I did this for two reasons:
1) calling `proposalHearbeat(..)` blocks because of the for-loop in the implementation and I have not enough golang experience to have another solution at hand
2) it's the same as it is called in the implementation

consensus/state_test.go
srmo: Though process behind this defensive:
If we don't wait at least `proposalHeartbeatIntervalSeconds`, the test method will terminate immediately after `go cs.proposalHeartbeat(10, 1)` and depending on performance,  a EventProposalHearbeat` would have been sent just after leaving the test method. And that's exactly what we want to ensure it doesn't happen.

Tried to clarify with a comment and added an extra second.

consensus/state_test.go
```diff
@@ -1027,6 +1029,33 @@ func TestSetValidBlockOnDelayedPrevote(t *testing.T) {
 	assert.True(t, rs.ValidRound == round)
 }
 
+// regression for #2518
```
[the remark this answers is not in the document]
srmo: oops :)

consensus/state_test.go
```diff
@@ -1027,6 +1029,33 @@ func TestSetValidBlockOnDelayedPrevote(t *testing.T) {
...
+	cs.evsw.AddListenerForEvent("testing", types.EventProposalHeartbeat,
+		func(data tmevents.EventData) {
+			t.Errorf("Should not have broadcasted heartbeat")
+		})
+	go cs.proposalHeartbeat(10, 1)
+
+	cs.Stop()
+
+	// if a faulty implementation sends an event, we should wait here a little bit to make sure we don't miss it by prematurely leaving the test method
+	time.Sleep((proposalHeartbeatIntervalSeconds  + 1) * time.Second)
```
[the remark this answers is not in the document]
srmo: Hmm...I could try myself on removing this proposalHeartbeat infrastructure

PR #2822, opened 2018-11-12
update to amino 0.14.1

zramsay:
fixes int overflow that prevent `tendermint` from building on RPi

* [ ] Wrote tests

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -58,7 +58,7 @@
 
 [[constraint]]
   name = "github.com/tendermint/go-amino"
-  version = "v0.14.0"
+  version = "v0.14.1"
 
 [[constraint]]
   name = "google.golang.org/grpc"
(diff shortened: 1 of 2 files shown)


Gopkg.toml
```diff
@@ -58,7 +58,7 @@
 
 [[constraint]]
   name = "github.com/tendermint/go-amino"
-  version = "v0.14.0"
```
ValarDragon: I find it kind of weird that this now break dependencies for any downstream application who was still using `v0.14.0`. perhaps we should set it to allow `v0.14.0` and `v0.14.1`, and then the downstream rasp-pi app would just select `v0.14.1`?

ValarDragon: Doesn't block merge, but is a general note about dependencies

ebuchman:
Can you redo this using `dep ensure -update github.com/tendermint/go-amino` so only one entry in the .lock file changes? Please and thank you :)

Gopkg.toml
melekes: Yeah, I think we should switch to `^X.Y.Z` or `~X.Y.Z` for most of the dependencies.

PR #2825, opened 2018-11-13
Add Composite() to PubKey interface

alessio:
ref cosmos/cosmos-sdk#2019

* [ ] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -16,6 +16,8 @@ Friendly reminder, we have a [bug bounty program](https://hackerone.com/tendermi
 
 * Go API
 
+  * [\#2825](https://github.com/tendermint/tendermint/pull/2825) Add Composite() field to PubKey interface
+
 * Blockchain Protocol
 
 * P2P Protocol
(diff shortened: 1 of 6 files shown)


CHANGELOG_PENDING.md
```diff
@@ -16,6 +16,8 @@ Friendly reminder, we have a [bug bounty program](https://hackerone.com/tendermi
 
 * Go API
 
+  * [\#2825](https://github.com/tendermint/tendermint/pull/2825) Add Composite() field to PubKey interface
```
```suggestion
  * \#2825 Add Composite() func to PubKey interface, which returns true if the pubkey is multisig
```

crypto/crypto.go
```diff
@@ -24,6 +24,7 @@ type PubKey interface {
 	Bytes() []byte
 	VerifyBytes(msg []byte, sig []byte) bool
 	Equals(PubKey) bool
+	Composite() bool
```
melekes: godoc?

alessio: Done, thanks

cwgoes:
This could be useful anyways, but I don't think we can solve the SDK-side problem this way - see https://github.com/cosmos/cosmos-sdk/issues/2019#issuecomment-438252878.

alessio:
This PR is the only way I can distinguish a multisig from a single-sig key.
Required to solve cosmos/cosmos-sdk#2019

Please @ValarDragon unblock.

PR #2831, opened 2018-11-13
docs: update genesis info

### LINKED ISSUE #2814: Documentation on genesis is outdated
Documentation on "genesis" is outdated and shows a version that would be compatible with 0.22.8 for example. https://tendermint.com/docs/tendermint-core/using-tendermint.html#initialize

It's missing the current consensus_param description as well as "address" field on validator.
Also, there are broken links like the "definition" link in section "Genesis"

zramsay:
closes #2814

* [ ] Wrote tests

### DIFF
--- a/docs/tendermint-core/using-tendermint.md
+++ b/docs/tendermint-core/using-tendermint.md
@@ -60,42 +60,34 @@ definition](https://github.com/tendermint/tendermint/blob/master/types/genesis.g
 
 ```
 {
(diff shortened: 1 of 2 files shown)


melekes:
We also need to document ConsensusParams in https://tendermint.com/docs/tendermint-core/using-tendermint.html#fields.

zramsay:
#2836

PR #2832, opened 2018-11-14
Config option for structured Logging

zmanian:
Introduce a config option for structured logging as JSON.

### DIFF
--- a/cmd/tendermint/commands/init.go
+++ b/cmd/tendermint/commands/init.go
@@ -2,11 +2,13 @@ package commands
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
 
 	cfg "github.com/tendermint/tendermint/config"
 	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/libs/log"
 	"github.com/tendermint/tendermint/p2p"
 	"github.com/tendermint/tendermint/privval"
 	"github.com/tendermint/tendermint/types"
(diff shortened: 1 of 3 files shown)


cmd/tendermint/commands/init.go
```diff
@@ -20,6 +22,8 @@ var InitFilesCmd = &cobra.Command{
 	RunE:  initFiles,
 }
 
+var logger = log.NewTMLogger(log.NewSyncWriter(os.Stdout))
```
melekes: why do we need another instance of logger if we already have one in `root.go` https://github.com/tendermint/tendermint/blob/5a6822c8acbd7353729cbb5d8393de2f6da39425/cmd/tendermint/commands/root.go#L18

cmd/tendermint/commands/root.go
```diff
@@ -54,6 +53,12 @@ var RootCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
+		var logger log.Logger
```
melekes:
```suggestion
if config.StructuredLog {
		logger = log.NewTMJSONLogger(log.NewSyncWriter(os.Stdout))
}
```

config/config.go
```diff
@@ -145,6 +145,9 @@ type BaseConfig struct {
 	// Output level for logging
 	LogLevel string `mapstructure:"log_level"`
 
+	//Structured Logging as JSON or colored text
+	StructuredLog bool `mapstructure:"structured_log"`
```
melekes: I'd call it `LogFormat string` ("plain", "json"). Log lines we output to stdout also have structure, so it's not fair to call them unstructured.

melekes:
Closing for https://github.com/tendermint/tendermint/pull/2843

PR #2844, opened 2018-11-15
do not pin deps to exact versions

### LINKED ISSUE #2798: client_golang dependency is out of date

melekes:
(at least those with releases)

because
- they are locked in .lock file already
- individual dependencies can be updated with `dep ensure -update XXX`
- review process (and ^^^) should help us prevent accidental updates

Closes #2798

* [ ] Wrote tests

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -20,57 +20,60 @@
 #   unused-packages = true
 #
 ###########################################################
-# NOTE: All packages should be pinned to specific versions.
-# Packages without releases must pin to a commit.
 
-
+# Allow only patch releases for serialization libraries
 [[constraint]]
-  name = "github.com/go-kit/kit"
-  version = "=0.6.0"
+  name = "github.com/tendermint/go-amino"
+  version = "~0.14.1"
 
 [[constraint]]
   name = "github.com/gogo/protobuf"
-  version = "=1.1.1"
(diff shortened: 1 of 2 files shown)


Gopkg.toml
```diff
@@ -20,57 +20,60 @@
...
-
-[[constraint]]
-  name = "github.com/tendermint/go-amino"
-  version = "v0.14.1"
+  version = "^1.2.1"
 
 [[constraint]]
   name = "google.golang.org/grpc"
-  version = "=1.13.0"
+  version = "~1.13.0"
```
ebuchman: Is this being considered a serialization library? We should move it to the other section then.

Issue #2827, opened 2018-11-13
Implement config flag for different PrivVal implementations


Slamper:
Since the implementation of IPCVal we have 3 different implementations of `PrivVal`.

For TCPVal we have a config flag that overrides the privValidator that is passed to the node constructor:

https://github.com/tendermint/tendermint/blob/80d0a362500fea2dd089258319075a54e5d40a2d/node/node.go#L224-L241

On the one hand this might cause confusions for people implementing an App on Tendermint as you would not expect that the privValidator you are passing could be overwritten and on the other hand this implementation makes it harder or at least ugly to implement the third (IPCVal) and potentially even more implementations.

The questions are:
* Do we want to leave the choice of the PrivVal implementation to the App built on tendermint and use the passed PV ?
* If not how can we beautifully implement the third interface IPCVal ?

joe-bowman:
It'd probably be sensible to add a config field for privValidatorType, and instantiate the correct class based upon that - then pass in privValidatorLocation (whether it be a file, socket or URI), effectively replacing both of the original fields; that should be pretty future proof too, should we add future validator types.

Would probably be sensible to contiue to support PrivValidatorFile field for backwards compatibility (and possibly PrivValidatorListenAddr too, if it is used elsewhere).

Liamsi:
In the KMS, we do not use command line arguments but a config file instead. I think this is slightly cleaner (but this wouldn't be backwards compatible and might have other downsides which I'm not seeing currently):
https://github.com/tendermint/kms/blob/c80135c8dd86d47df034eb07fde80922053223e8/tmkms.toml.example#L9-L19

Either way (config flag only, or config files), a validator-type flag plus a a validator-path/-location as @joe-bowman suggested makes the most sense. (viper gives us config file support with cobra / config flags anyways?)

Liamsi:
ref #2651

Liamsi:
@ebuchman and @melekes suggested to just keep `priv_validator_laddr` and include a `tcp://` or `unix://` prefix. This would also be consistent with other listen addresses, e.g.,:
https://github.com/tendermint/tendermint/blob/80d0a362500fea2dd089258319075a54e5d40a2d/cmd/tendermint/commands/lite.go#L39

joe-bowman:
This was also my first thought. I like the consistency but it's a little
unclean and doesn't have the extensibility for future use cases. That said
it is a lower risk change which isn't a bad thing right now!

melekes:
[...]

PR #2847, opened 2018-11-15
WIP: Enable both privval types (tcp, unix) per prefix

Liamsi:
ref #2827

Some quick changes done to test the IPC / unix communication with the kms. If the general direction is OK, I would add a few tests and clean up this PR.

* [ ] Wrote tests


Liamsi:
closing in favour of #2866. It does the same thing but has some tests and documentation.

PR #2851, opened 2018-11-15
Release/v0.26.2

ebuchman:
[Changelog](https://github.com/tendermint/tendermint/blob/9973decff9d6abb23350351f59dc8409611b5683/CHANGELOG.md#v0262)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+## v0.26.2
+
+*November 15th, 2018*
+
+Special thanks to external contributors on this release: @hleb-albau, @zhuzeyu
+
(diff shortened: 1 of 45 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Changelog
 
+## v0.26.2
+
+*November 15th, 2018*
+
+Special thanks to external contributors on this release: @hleb-albau, @zhuzeyu
```
melekes: hmm.. i don't see anything from zhuzeyu below

ebuchman: It's the one liner that replaced stateKey with key - I thought we decided some things aren't worth an entry but worth mentioning the contributor?

PR #2866, opened 2018-11-16
allow IPCVal to be instantiated from config

joe-bowman:
Ref #2827 

(I have since seen #2847 which is a fix for the same issue; this PR has tests and docs too ;) )

* [x] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -2,8 +2,6 @@
 
 ## v0.26.4
 
-*TBD*
-
 Special thanks to external contributors on this release:
 
 Friendly reminder, we have a [bug bounty
@@ -27,5 +25,6 @@ program](https://hackerone.com/tendermint).
 
 - [config] \#2877 add blocktime_iota to the config.toml (@ackratos)
 - [mempool] \#2855 add txs from Update to cache
+- [node] \#2827 add ability to instantiate IPCVal (@joe-bowman)
 
 ### BUG FIXES:
--- a/docs/architecture/adr-008-priv-validator.md
+++ b/docs/architecture/adr-008-priv-validator.md
@@ -5,14 +5,17 @@ implementations:
 
 - FilePV uses an unencrypted private key in a "priv_validator.json" file - no
   configuration required (just `tendermint init`).
-- SocketPV uses a socket to send signing requests to another process - user is
-  responsible for starting that process themselves.
+- TCPVal and IPCVal use TCP and Unix sockets respectively to send signing requests
+  to another process - the user is responsible for starting that process themselves.
 
-The SocketPV address can be provided via flags at the command line - doing so
-will cause Tendermint to ignore any "priv_validator.json" file and to listen on
-the given address for incoming connections from an external priv_validator
-process. It will halt any operation until at least one external process
-succesfully connected.
+Both TCPVal and IPCVal addresses can be provided via flags at the command line
+or in the configuration file; TCPVal addresses must be of the form
+`tcp://<ip_address>:<port>` and IPCVal addresses `unix:///path/to/file.sock` -
+doing so will cause Tendermint to ignore any private validator files.
+
+TCPVal will listen on the given address for incoming connections from an external
+private validator process. It will halt any operation until at least one external
+process successfully connected.
 
 The external priv_validator process will dial the address to connect to
 Tendermint, and then Tendermint will send requests on the ensuing connection to
@@ -20,6 +23,9 @@ sign votes and proposals. Thus the external process initiates the connection,
 but the Tendermint process makes all requests. In a later stage we're going to
 support multiple validators for fault tolerance. To prevent double signing they
 need to be synced, which is deferred to an external solution (see #1185).
+
+Conversely, IPCVal will make an outbound connection to an existing socket opened
+by the external validator process.
 
 In addition, Tendermint will provide implementations that can be run in that
 external process. These include:
--- a/node/node.go
+++ b/node/node.go
@@ -148,6 +148,44 @@ type Node struct {
 	prometheusSrv    *http.Server
 }
 
+func createExternalPrivValidator(listenAddr string, logger log.Logger) (types.PrivValidator, error) {
+	protocol, address := cmn.ProtocolAndAddress(listenAddr)
+
+	var pvsc types.PrivValidator
+
+	switch (protocol) {
+	case "unix":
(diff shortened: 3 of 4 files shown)


CHANGELOG_PENDING.md
```diff
@@ -1,5 +1,7 @@
 # Pending
 
+* Add ability to instantiate IPCVal (ref #2827).
```
```suggestion
- [node] \#2827 add ability to instantiate IPCVal (@joe-bowman)
```

this should go under "IMPROVEMENTS"

docs/architecture/adr-008-priv-validator.md
```diff
@@ -5,14 +5,17 @@ implementations:
...
 
-The SocketPV address can be provided via flags at the command line - doing so
-will cause Tendermint to ignore any "priv_validator.json" file and to listen on
-the given address for incoming connections from an external priv_validator
-process. It will halt any operation until at least one external process
-succesfully connected.
+Both TCPVal and IPCVal addresses can be provided via flags at the command line
+or in the configuration file; TCPVal addresses must be of the form
+`tcp://<ip_address>:<port>` and IPCVal addresses `unix:///path/to/file.sock` -
+doing so will cause Tendermint to ignore any "priv_validator.json" file.
```
melekes:
```suggestion
doing so will cause Tendermint to ignore any private validator files.
```

let's not focus on the concrete file extension (`.json`) and use "private validator file" or "node key file" instead.

node/node.go
```diff
@@ -219,25 +219,50 @@ func NewNode(config *cfg.Config,
...
 
-		privValidator = pvsc
+			if err := pvsc.Start(); err != nil {
+				return nil, fmt.Errorf("Error starting private validator client: %v", err)
+			}
+
+			privValidator = pvsc
+
+		} else {
+			return nil, fmt.Errorf(
```
```suggestion
			return nil, fmt.Errorf(
				"Error creating private validator: expected either tcp or unix protocols, got %s", 
				protocol)
```

node/node.go
```diff
@@ -219,25 +219,50 @@ func NewNode(config *cfg.Config,
...
-			privKey = ed25519.GenPrivKey()
-			pvsc    = privval.NewTCPVal(
-				logger.With("module", "privval"),
-				config.PrivValidatorListenAddr,
-				privKey,
+		// If an address is provided, listen on the socket for a
+		// connection from an external signing process.
+		protocol, address := cmn.ProtocolAndAddress(config.PrivValidatorListenAddr)
+
+		if protocol == "unix" {
```
melekes: better to use a `switch` here

node/node.go
```diff
@@ -219,25 +219,50 @@ func NewNode(config *cfg.Config,
...
 	if config.PrivValidatorListenAddr != "" {
-		var (
-			// TODO: persist this key so external signer
-			// can actually authenticate us
-			privKey = ed25519.GenPrivKey()
-			pvsc    = privval.NewTCPVal(
-				logger.With("module", "privval"),
-				config.PrivValidatorListenAddr,
-				privKey,
+		// If an address is provided, listen on the socket for a
```
melekes: can we extract this into a function?

node/node_test.go
```diff
@@ -113,3 +118,111 @@ func TestNodeSetAppVersion(t *testing.T) {
...
+		ed25519.GenPrivKey(),
+	)
+	privval.RemoteSignerConnDeadline(5 * time.Millisecond)(rs)
+	privval.RemoteSignerConnRetries(1e6)(rs)
+
+	config := cfg.ResetTestRoot("node_priv_val_tcp_test")
+
+	config.SetPrivValidatorListenAddr(addr)
+
+  // kick off remote signer routine, and then start TM.
```
melekes: formatting is off

node/node_test.go
```diff
@@ -113,3 +118,111 @@ func TestNodeSetAppVersion(t *testing.T) {
...
```
melekes: formatting is off

node/node_test.go
```diff
@@ -113,3 +118,111 @@ func TestNodeSetAppVersion(t *testing.T) {
...
+	go func(rs *privval.RemoteSigner) {
+		defer rs.Stop()
```
melekes: defer usually goes after `Start()`, not before

node/node_test.go
```diff
@@ -113,3 +118,111 @@ func TestNodeSetAppVersion(t *testing.T) {
...
+
+	n, err := DefaultNewNode(config, log.TestingLogger())
+
+	assert.NoError(t, err, "expected no err on DefaultNewNode")
+
+	assert.IsType(t, &privval.TCPVal{}, n.PrivValidator())
+}
+
+func TestNodeSetPrivValIPC(t *testing.T) {
+  tmpfile := "/tmp/kms." + cmn.RandStr(6) + ".sock"
```
melekes: formatting is off

config/config.go
```diff
@@ -96,6 +96,11 @@ func (cfg *Config) SetRoot(root string) *Config {
 	return cfg
 }
 
+func (cfg *Config) SetPrivValidatorListenAddr(addr string) *Config {
```
melekes: I'd prefer not to have this function. It does not bring any value and encourages mutation.

node/node.go
```diff
@@ -148,6 +148,48 @@ type Node struct {
...
+	default:
+		return nil, fmt.Errorf(
+			"Error creating private validator: expected either tcp or unix "+
+			"protocols, got %s",
+			protocol,
+		)
+	}
+
+	pvServ, ok := pvsc.(cmn.Service)
+	if !ok {
```
melekes: let's remove this check. It's clearly a developer's error and it's OK to panic if priv validator client does not implement Service.

config/config.go
joe-bowman: Any pointers as to how I might override the priv_val_laddr config when testing option without it?

node/node.go
joe-bowman: Sure thing.

joe-bowman:
If 0.26.4 is planned to be release in preparation for GoS, is there any chance to get this merged in too? It doesn't touch any existing code, so should be low risk but will enable some participants to run their planned production configurations for GoS.

CHANGELOG_PENDING.md
```diff
@@ -2,8 +2,6 @@
 
 ## v0.26.4
 
-*TBD*
```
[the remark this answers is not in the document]
melekes:
```suggestion
*TBD*
```

PR #2873, opened 2018-11-17
Release/v0.26.3

ebuchman:
[Changelog](https://github.com/tendermint/tendermint/blob/ccf6b2b51286193b9b9c582c202589facc4d053a/CHANGELOG.md#v0263)


CONTRIBUTING.md
```diff
@@ -127,3 +151,13 @@ master constitutes a tagged release.
 - merge hotfix-vX.X.X to master
 - merge hotfix-vX.X.X to develop
 - delete the hotfix-vX.X.X branch
+
+
+## Testing
+
+All repos should be hooked up to [CircleCI](https://circleci.com/).
```
ValarDragon: This seems odd for contributing.md. (It reads to me as your fork should have circle CI testing)

I think it was intended to mean that all of tendermint organization's repos must have commits passing circle CI tests before merging.

ebuchman: That's right. That's why I moved it to the bottom and will probably break it out all together

PR #2876, opened 2018-11-18
Close and retry a RemoteSigner on err

zmanian:
This allows recovery if the RemoteSigner service is restarted etc without restarting the process.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -25,4 +25,6 @@ program](https://hackerone.com/tendermint).
 
 ### IMPROVEMENTS:
 
+Retry RemoteSigner connections on Error
+
 ### BUG FIXES:
--- a/privval/tcp.go
+++ b/privval/tcp.go
@@ -123,6 +123,9 @@ func (sc *TCPVal) OnStart() error {
 						"Ping",
 						"err", err,
 					)
+					sc.conn.Close()
+					sc.RemoteSignerClient = NewRemoteSignerClient(sc.conn)
+
 				}
 			case <-sc.cancelPing:
 				sc.pingTicker.Stop()


CHANGELOG_PENDING.md
```diff
@@ -25,4 +25,6 @@ program](https://hackerone.com/tendermint).
 
 ### IMPROVEMENTS:
 
+Retry RemoteSigner connections on Error
```
```suggestion
- [privval] retry RemoteSigner connections on Error
```

privval/tcp.go
```diff
@@ -123,6 +123,9 @@ func (sc *TCPVal) OnStart() error {
 						"Ping",
 						"err", err,
 					)
+					sc.conn.Close()
+					sc.RemoteSignerClient = NewRemoteSignerClient(sc.conn)
```
melekes: If we're closing the `sc.conn`, why do we create a new remote signer client with it? Shouldn't we call OnStop(), then OnStart() to wait for conn and recreate all necessary structures and goroutines?

Shouldn't we assert the err's type? We may want to shut everything down in case of `ErrUnexpectedResponse`, no?

Liamsi: > If we're closing the sc.conn, why do we create a new remote signer client with it? Shouldn't we call OnStop(), then OnStart() to wait for conn and recreate all necessary structures and goroutines?

I think that makes sense! 

> Shouldn't we assert the err's type? We may want to shut everything down in case of ErrUnexpectedResponse, no?

Probably yes. `ErrUnexpectedResponse` means the other side responded but with another message than expected (not `PingResponse`).

Liamsi: I'll continue here: https://github.com/tendermint/tendermint/pull/2923

Liamsi:
closing in favour of: https://github.com/tendermint/tendermint/pull/2923

privval/tcp.go
Liamsi: BTW calling OnStart won't be enough as this might time out (again) if the remote signer isn't up yet. We might need sth like retrying with exponential backoff. 
Also, the whole ConsensusState shuts down if the remote signer drops (restarting the remote signer won't help with this): https://github.com/tendermint/tendermint/issues/2926

Issue #2877, opened 2018-11-18
include BlockTimeIota should be included in default config toml


ackratos:

melekes:
PR with a fix/feature was merged to develop. Will be shipped with 0.26.4 or 0.27.0 release (check the changelog).

ackratos:
> PR with a fix/feature was merged to develop. Will be shipped with 0.26.4 or 0.27.0 release (check the changelog).

thanks for merging! :)

PR #2878, opened 2018-11-18
#2877 include BlockTimeIota should be included in default config toml

ackratos:
* [ ] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -25,6 +25,6 @@ program](https://hackerone.com/tendermint).
 
 ### IMPROVEMENTS:
 
-- [config] \#2877 add blocktime_iota to the config.toml
+- [config] \#2877 add blocktime_iota to the config.toml (@ackratos)
 
 ### BUG FIXES:
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -25,4 +25,6 @@ program](https://hackerone.com/tendermint).
 
 ### IMPROVEMENTS:
 
+- [config] \#2877 add blocktime_iota to the config.toml
+
 ### BUG FIXES:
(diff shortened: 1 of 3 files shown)


config/toml.go
```diff
@@ -260,6 +260,9 @@ create_empty_blocks_interval = "{{ .Consensus.CreateEmptyBlocksInterval }}"
 peer_gossip_sleep_duration = "{{ .Consensus.PeerGossipSleepDuration }}"
 peer_query_maj23_sleep_duration = "{{ .Consensus.PeerQueryMaj23SleepDuration }}"
 
+# Block time parameters. Corresponds to the minimum time increment between consecutive blocks.
```
melekes: this needs to be reflected in https://github.com/tendermint/tendermint/blob/develop/docs/tendermint-core/configuration.md

CHANGELOG_PENDING.md
```diff
@@ -25,4 +25,6 @@ program](https://hackerone.com/tendermint).
 
 ### IMPROVEMENTS:
 
+- [config] [\#2877](https://github.com/tendermint/tendermint/issues/2877) include BlockTimeIota should be included in default config toml
```
```suggestion
- [config] \#2877 add blocktime_iota to the config.toml
```

melekes:
@milosevic is there a reason `BlockTimeIota` wasn't included in the config.toml file in the first place?

milosevic:
There is no reason why it hasn't been included; we most probably forgot to do it.

config/toml.go
```diff
@@ -260,6 +260,9 @@ create_empty_blocks_interval = "{{ .Consensus.CreateEmptyBlocksInterval }}"
...
+blocktime_iota = "{{ .Consensus.BlockTimeIota }}"
```
ebuchman: we should probably change this to `block_time_iota`. Also the description doesn't help distinguish it from TimeoutCommit - should be clear that this refers to the timestamp.

Also, isn't it a ConsensusParam, rather than just a config option, @milosevic ?

milosevic: I agree that description should clearly say this is only about timestamp. Yep it's a ConsensusParam (definition: ConsensusParams contains consensus critical parameters that determine the validity of blocks.).

PR #2881, opened 2018-11-19
Modify the txs number in ReapMaxTxs

zhuzeyu:
if mem.txs.Len() is less than max, we should collect the mem.txs.Len()
txs in ReapMaxTxs, So it's better to use minimum among  mem.txs.Len() and
max when we colletc txs

Signed-off-by: Zeyu Zhu <zhuzeyu0409@gmail.com>

* [ ] Wrote tests

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -517,8 +517,9 @@ func (mem *Mempool) ReapMaxTxs(max int) types.Txs {
 		time.Sleep(time.Millisecond * 10)
 	}
 
-	txs := make([]types.Tx, 0, cmn.MinInt(mem.txs.Len(), max))
-	for e := mem.txs.Front(); e != nil && len(txs) <= max; e = e.Next() {
+	numTxsToReap := cmn.MinInt(mem.txs.Len(), max)
+	txs := make([]types.Tx, 0, numTxsToReap)
+	for e := mem.txs.Front(); len(txs) < numTxsToReap; e = e.Next() {
 		memTx := e.Value.(*mempoolTx)
 		txs = append(txs, memTx.tx)
 	}


mempool/mempool.go
```diff
@@ -517,8 +517,9 @@ func (mem *Mempool) ReapMaxTxs(max int) types.Txs {
 		time.Sleep(time.Millisecond * 10)
 	}
 
-	txs := make([]types.Tx, 0, cmn.MinInt(mem.txs.Len(), max))
-	for e := mem.txs.Front(); e != nil && len(txs) <= max; e = e.Next() {
+	txsNum := cmn.MinInt(mem.txs.Len(), max)
```
```suggestion
	numTxsToReap := cmn.MinInt(mem.txs.Len(), max)
```

mempool/mempool.go
```diff
@@ -517,8 +517,9 @@ func (mem *Mempool) ReapMaxTxs(max int) types.Txs {
 		time.Sleep(time.Millisecond * 10)
 	}
 
-	txs := make([]types.Tx, 0, cmn.MinInt(mem.txs.Len(), max))
-	for e := mem.txs.Front(); e != nil && len(txs) <= max; e = e.Next() {
```
melekes: note: `e != nil` prevents us from going over mem.txs.Len()

zhuzeyu: I know，I think "len(txs) <= max" is not precise, and It's a little redundant to judge twice for "e != nil && len(txs) <= max",
if we modify it to numTxsToReap , "e != nil" is without need, what do you think of

melekes: Agree, if we switch to `numTxsToReap`, we can remove `e != nil`

zhuzeyu: thanks for review :)

melekes: note for myself: need to check if mem.txs can be mutated while we iterate over it in ReapMaxTxs

melekes:
> So it's better to use minimum among mem.txs.Len() and max when we colletc txs

why it's better?

zhuzeyu:
> On the second thought: I don't think we should limit to `numTxsToReap`. If new txs are added to the list while we are iterating over them in `ReapMaxTxs`, it's completely fine to return them too (up to max txs).
> 
> I'd vote for closing this.

The lock is called at the beginning of the function, so it is impossible that new txs are added to the list when function executesWe should use actual txs numbers to judge, not the max, and the actual txs numbers is the minimum among mem.txs.Len() and max
defer mem.proxyMtx.Unlock()

> On the second thought: I don't think we should limit to `numTxsToReap`. If new txs are added to the list while we are iterating over them in `ReapMaxTxs`, it's completely fine to return them too (up to max txs).
> 
> I'd vote for closing this.

The lock is called at the beginning of the function, so it is impossible that new txs are added to the list when function executes.  this can also be turned off,  because It's just optimized, doesn't affect functionality	
mem.proxyMtx.Lock()
defer mem.proxyMtx.Unlock()

zhuzeyu:
> > So it's better to use minimum among mem.txs.Len() and max when we colletc txs
> 
> why it's better?

i think we should use actual txs numbers to judge, not the max, and the actual txs numbers is the minimum among mem.txs.Len() and max

melekes:
> The lock is called at the beginning of the function, so it is impossible that new txs are added to the list when function executes. this can also be turned off, because It's just optimized, doesn't affect functionality mem.proxyMtx.Lock() defer mem.proxyMtx.Unlock()

Not sure about that. For example, I don't see a lock in `resCbNormal` https://github.com/tendermint/tendermint/blob/b487feba4269660b18252695df9e2e79ca635724/mempool/mempool.go#L371

zhuzeyu:
> > The lock is called at the beginning of the function, so it is impossible that new txs are added to the list when function executes. this can also be turned off, because It's just optimized, doesn't affect functionality mem.proxyMtx.Lock() defer mem.proxyMtx.Unlock()
> 
> Not sure about that. For example, I don't see a lock in `resCbNormal`
> 
> [tendermint/mempool/mempool.go](https://github.com/tendermint/tendermint/blob/b487feba4269660b18252695df9e2e79ca635724/mempool/mempool.go#L371)
> 
> Line 371 in [b487feb](/tendermint/tendermint/commit/b487feba4269660b18252695df9e2e79ca635724)
> 
>  mem.txs.PushBack(memTx)

there is a lock in PushBack function, here
https://github.com/tendermint/tendermint/blob/b487feba4269660b18252695df9e2e79ca635724/libs/clist/clist.go#L317

melekes:
We should bring back the `e != nil` check.

melekes:
Closing because of https://github.com/tendermint/tendermint/pull/2881#issuecomment-440950775.

PR #2882, opened 2018-11-19
mempool: add txs from Update to cache

melekes:
* [x] Wrote tests

### DIFF
--- a/mempool/mempool_test.go
+++ b/mempool/mempool_test.go
@@ -163,6 +163,17 @@ func TestMempoolFilters(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


mempool/mempool_test.go
```diff
@@ -163,6 +163,18 @@ func TestMempoolFilters(t *testing.T) {
 	}
 }
 
+func TestMempoolUpdate(t *testing.T) {
```
ValarDragon: Perhaps rename to `TestMempoolUpdate_cache`, or something to that extent. From the name, I thought it was going to test that update removed the tx from the mempool list itself.

PR #2895, opened 2018-11-21
node: refactor privValidator ext client code & tests

melekes:


node/node.go
```diff
@@ -3,19 +3,19 @@ package node
...
 	"strings"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/rs/cors"
 
-	"github.com/tendermint/go-amino"
+	amino "github.com/tendermint/go-amino"
```
Liamsi: I never understood why we alias here.

melekes: `go-imports` does this automatically. in short: because `amino` != `go-amino`

Issue #2916, opened 2018-11-26
Stuck in fast sync with no peers and validator set > 1


ebuchman:
If the validator set is greater than 1, but we have no peers, we will be stuck in fast sync, even if we have more than 2/3 of the voting power.

This is easy to replicate. Run a node with `tendermint node --proxy_app=persistent_kvstore`. Add a new validator with 1/11 of the voting power: `curl localhost:26657/broadcast_tx_sync?tx=\"val:0e783029302e8e4713f2c3085ee35c3335bbfd698f2bba09d3196aca3872d651/1\"`. Now restart the node and it will be stuck in fast sync.

We treat the special case of being the only validator in https://github.com/tendermint/tendermint/blob/develop/node/node.go#L236. We should generalize this - instead of just checking if we're the sole validator, we should check if we have > 2/3 of the voting power - if so, we can skip fast sync. If we have less than 2/3 of the voting power, we'll need at least one peer before we can switch out of fast sync.

Discovered this while trying to test that https://github.com/tendermint/tendermint/pull/2747 is backwards compatible by making validator set changes on current version of the software and then trying to run that branch.

ebuchman:
Never mind - we don't necessarily want this. It's possible a node with +2/3 is still behind and wants to sync up. It's even possible if theres just a single validator. But we'll leave the heuristic for now and others can just turn off fast sync if they find themselves in this position.

PR #2917, opened 2018-11-26
dont fast sync if we have +2/3 of the voting power

ebuchman:
Fix https://github.com/tendermint/tendermint/issues/2916

* [ ] Wrote tests

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -1,7 +1,6 @@
 package node
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"net"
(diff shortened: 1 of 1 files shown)


node/node.go
```diff
@@ -230,14 +229,13 @@ func NewNode(config *cfg.Config,
...
+	// We don't fast-sync when we have +2/3 of the validator set.
 	fastSync := config.FastSync
-	if state.Validators.Size() == 1 {
-		addr, _ := state.Validators.GetByIndex(0)
-		if bytes.Equal(privValidator.GetAddress(), addr) {
-			fastSync = false
-		}
+	ourAddr := privValidator.GetAddress()
+	_, val := state.Validators.GetByAddress(ourAddr)
+	if val != nil && val.VotingPower > state.Validators.TotalVotingPower()*2/3 {
```
ValarDragon: This is another case where we require total voting power to be smaller than its max value. (In this case less than 2**62, previous mention of need for a requirement: https://github.com/tendermint/tendermint/pull/2785#issuecomment-440357821)

ValarDragon:
Actually on second thought, I'm unsure we want this. A node with 2/3rds of stake may want to spawn a replica and therefore still need to fast sync. Now they would have to pick another privkey, sync, and then switch to their main node id, which seems like a lot of overhead.

ebuchman:
I think you're right.

Issue #2784, opened 2018-11-08
Data race in p2p tests (related to secret connection)


Liamsi:
`TestRemoteSignProposalErrors` fails because of a data race (non-deterministically).
See https://circleci.com/gh/tendermint/tendermint/33036 for details.

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):

**ABCI app** (name for built-in, URL for self-written if it's publicly available):

**Environment**:
- **OS** (e.g. from /etc/os-release):
- **Install tools**:
- **Others**:

**What happened**:

**What you expected to happen**:

**Have you tried the latest version**: yes/no

**How to reproduce it** (as minimally and precisely as possible):

**Logs (paste a small part showing an error (< 10 lines) or link a pastebin, gist, etc. containing more of the log file)**:

**Config (you can paste only the changes you've made)**:

**node command runtime flags**:

**`/dump_consensus_state` output for consensus bugs**

**Anything else we need to know**:

ebuchman:
There's a number of non-deterministic bugs in these tests. See #2677 for more!

ebuchman:
Seen again:

Seen in https://circleci.com/gh/tendermint/tendermint/36761?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

```
==================
WARNING: DATA RACE
Read at 0x00c42084d18b by goroutine 103:
  github.com/tendermint/tendermint/vendor/golang.org/x/crypto/chacha20poly1305.setupState()
      /usr/local/go/src/encoding/binary/binary.go:63 +0x367
  github.com/tendermint/tendermint/vendor/golang.org/x/crypto/chacha20poly1305.(*chacha20poly1305).seal()
      /go/src/github.com/tendermint/tendermint/vendor/golang.org/x/crypto/chacha20poly1305/chacha20poly1305_amd64.go:56 +0xcb
  github.com/tendermint/tendermint/vendor/golang.org/x/crypto/chacha20poly1305.(*chacha20poly1305).Seal()
      /go/src/github.com/tendermint/tendermint/vendor/golang.org/x/crypto/chacha20poly1305/chacha20poly1305.go:69 +0x109
  github.com/tendermint/tendermint/p2p/conn.(*SecretConnection).Write()
      /go/src/github.com/tendermint/tendermint/p2p/conn/secret_connection.go:136 +0x3e5
[...]

PR #2923, opened 2018-11-27
Close and retry a RemoteSigner on err

Liamsi:
Supersedes #2876 (I can't push to Zaki's fork)
Address https://github.com/tendermint/tendermint/pull/2876#discussion_r234509650
ref https://github.com/tendermint/kms/issues/116

depends on #2948 (after that gets merged I'll try a proper war to reanimate TCPVal/IPCVal).

Edit: This PR was tested manually against the KMS.~~Probably adding a unit-test that drops the conn from the server and reconnects the client is still necessary though.~~ Simple tests added.

* [x] Wrote tests


amrali:
@Liamsi I found the same issue exists for ipcval as well. I don't think there should be any panic, it's a missed vote or proposal, the node can continue running just fine.

I think even a dumb retrial for ipcval should do. Should abstract this across both methods or perhaps do the same thing for ipcval?

Liamsi:
> I found the same issue exists for ipcval as well. I don't think there should be any panic, it's a missed vote or proposal, the node can continue running just fine.

Are you referring to #2948? I think so too.

> Should abstract this across both methods or perhaps do the same thing for ipcval?

Good question. Probably we can abstract this across both methods. I'd still start with handling (and testing) both cases separately and then find a common way to handle both cases if they are actually the same. Note that the above changes do not properly solve the issue yet. I'll get back to this after #2948 is merged. Feel free to experiment with the ipcval case if you want!

joe-bowman:
> Note that the above changes do not properly solve the issue yet. I'll get back to this after #2948 is merged. Feel free to experiment with the ipcval case if you want!

I've been running this change (mirrored to IPCVal), and #2948 in a test env and the IPC side seems to work as expected (although I appreciate there are subtle differences between the TCP and IPC implementations).

Liamsi:
That's pretty cool! Thanks @joe-bowman 👍 Yeah, I also experimented with this change in combination with #2948 (and ICP changes). ICP seemed fine while TCP didn't always work as expected.

yulidai:
I also have this problem under the version "0.29.0-0-g2b3842c".

Whether restart kms or signature failure, gaiad will disconnect immediately until i restart gaiad

yulidai:
Will it be repaired before the main network launch?

So we can use KMS to build validator with high availability.  Thanks :)

Liamsi:
> Will it be repaired before the main network launch?

Sorry for the late reply @yulidai! Yes, that is certainly planned to be fixed before main net launch. 

@joe-bowman I tested this further and even with IPC this change doesn't work reliably (at least if the conn drops severals time in short period of time). 

Instead of calling OnStart (recursively from OnStart and in a go-routine which recursively spawns more go-routines then), I tried to establish a new connection and create a `NewRemoteSignerClient` with that fresh connection (I'll do more tests and push sth. later in the day). This works fine if restarting the kms happens "quick enough". This means if the kms is restarted before the `ConsensusState` handles a timeout, which also happens to default to 3 sec. If this time passes tendermint doesn't seem to continue sending consensus messages to the remote signer. This param is configurable though (search for `timeout_propose` in your config.toml). 
(Note: this timeout issue is independent from how exactly tendermint tries to reconnect to the remote signer).

p2p/conn/secret_connection.go
```diff
@@ -112,6 +114,9 @@ func (sc *SecretConnection) RemotePubKey() crypto.PubKey {
 // Writes encrypted frames of `sealedFrameSize`
 // CONTRACT: data smaller than dataMaxSize is read atomically.
 func (sc *SecretConnection) Write(data []byte) (n int, err error) {
+	// TODO(ismail): is this safe to be called in go-routines?
```
ebuchman: I think this should stay unsafe. It's the callers responsibility to serialize access to this connection. See explanations in https://github.com/tendermint/tendermint/issues/3064#issuecomment-453368403 and https://github.com/tendermint/tendermint/issues/2784#issuecomment-453361640

privval/ipc_client.go
```diff
@@ -35,7 +37,10 @@ type IPCVal struct {
 	connTimeout   time.Duration
 	connHeartbeat time.Duration
 
-	conn       net.Conn
+	conn net.Conn
+	// connMtx guards writing and reading the field (methods on net.Conn itself are gorountine safe though)
+	connMtx sync.RWMutex
```
ebuchman: mtx should always go *above* the field it guards

p2p/conn/secret_connection.go
```diff
@@ -31,8 +31,10 @@ const aeadNonceSize = chacha20poly1305.NonceSize
 // It is an implementation of the STS protocol.
 // Note we do not (yet) assume that a remote peer's pubkey
 // is known ahead of time, and thus we are technically
-// still vulnerable to MITM. (TODO!)
+// still vulnerable to MITM.
+// (TODO(ismail): see also https://github.com/tendermint/tendermint/issues/3010)
 // See docs/sts-final.pdf for more info
+// (TODO: this does not exist anymore; probably refers to docs/tendermint-core/secure-p2p.md now?)
```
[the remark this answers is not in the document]
Liamsi: Thanks! Updated comment.

ebuchman:
Oh, this would probably benefit from the unit tests testing both the TCP and IPC implementations. 

This is tested at an integrations level in the `node` package, but should probably happen in `privval` as well. Currently we're only using the TCPListener, but we should loop over both options, like we do in the rpc package when testing both URI and JSONRPC styles.

cmd/priv_val_server/main.go
```diff
@@ -34,13 +35,20 @@ func main() {
...
+	protocol, address := cmn.ProtocolAndAddress(*addr)
+	switch protocol {
+	case "unix":
+		dialer = privval.DialUnixFn(address)
+	case "tcp":
+		connTimeout := 3 * time.Second // TODO
+		dialer = privval.DialTCPFn(address, connTimeout, ed25519.GenPrivKey())
+	default:
+		logger.Error("Unknown protocol", "protocol", protocol)
+		return
```
melekes: shouldn't we exit with `os.Exit(1)`?

node/node.go
```diff
@@ -878,27 +878,30 @@ func createAndStartPrivValidatorSocketClient(
...
 		)
 	}
 
-	if pvsc, ok := pvsc.(cmn.Service); ok {
-		if err := pvsc.Start(); err != nil {
-			return nil, errors.Wrap(err, "failed to start")
-		}
+	pvsc := privval.NewSocketVal(logger.With("module", "privval"), listener)
+	if err := pvsc.Start(); err != nil {
+		return nil, errors.Wrap(err, "failed to start")
```
```suggestion
		return nil, errors.Wrap(err, "failed to start private validator")
```

privval/client.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+//--------------------------------------------------------
+// Connection and signer management
+
+// waits to accept and sets a new connection.
+// connection is closed in OnStop.
+// returns true if the listener is closed
+// (ie. it returns a nil conn).
+func (sc *SocketVal) reset() (bool, error) {
```
melekes:
```suggestion
func (sc *SocketVal) reset() (closed bool, err error) {
```

privval/client.go
```diff
@@ -0,0 +1,263 @@
...
+		if !sc.IsRunning() {
+			return nil, nil // Ignore error from listener closing.
+		}
+		return nil, err
+
+	}
+	return conn, nil
+}
+
+// waitConnection uses the configured wait timeout to error if no external
```
melekes: The comment seems off. I don't see any wait timeout (1) and `errc` only being populated in a goroutine. Do we even need a goroutine in such case?

ebuchman: I think you're right we don't need this go-routine, it can just be a straight forward function. The waiting happens implicitly in the implementation of `.Accept()`

PR #2924, opened 2018-11-27
don't ignore key when executing CONTAINS

### LINKED ISSUE #2912: CONTAINS in queries ignores key
*Tendermint version:* 0.22.8

The following patch adds a test that fails:

```
--- a/state/txindex/kv/kv_test.go
+++ b/state/txindex/kv/kv_test.go
@@ -89,6 +89,8 @@ func TestTxSearch(t *testing.T) {
                {"account.owner CONTAINS 'an'", 1},
                // search using CONTAINS
                {"account.owner CONTAINS 'Vlad'", 0},
+               // search for the wrong key using CONTAINS
+               {"account.number CONTAINS 'Iv'", 1},
        }

```

This happens because `match()` in `state/txindex/kv/kv.go` performs a full scan and tests the value, but neglects to test the key.

The best way to fix is to do a prefix lookup on `key/` instead of a full scan. It should also work faster.

melekes:
Fixes #2912

* [x] Wrote tests

### DIFF
--- a/state/txindex/kv/kv.go
+++ b/state/txindex/kv/kv.go
@@ -340,10 +340,10 @@ func (txi *TxIndex) match(c query.Condition, startKey []byte) (hashes [][]byte) 
 			hashes = append(hashes, it.Value())
 		}
 	} else if c.Op == query.OpContains {
-		// XXX: doing full scan because startKey does not apply here
-		// For example, if startKey = "account.owner=an" and search query = "accoutn.owner CONSISTS an"
-		// we can't iterate with prefix "account.owner=an" because we might miss keys like "account.owner=Ulan"
-		it := txi.store.Iterator(nil, nil)
+		// XXX: startKey does not apply here.
+		// For example, if startKey = "account.owner/an/" and search query = "accoutn.owner CONTAINS an"
+		// we can't iterate with prefix "account.owner/an/" because we might miss keys like "account.owner/Ulan/"
+		it := dbm.IteratePrefix(txi.store, []byte(fmt.Sprintf("%s/", c.Tag)))
 		defer it.Close()
 		for ; it.Valid(); it.Next() {
 			if !isTagKey(it.Key()) {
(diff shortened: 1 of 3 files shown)


state/txindex/kv/kv.go
```diff
@@ -340,10 +340,10 @@ func (txi *TxIndex) match(c query.Condition, startKey []byte) (hashes [][]byte)
 			hashes = append(hashes, it.Value())
 		}
 	} else if c.Op == query.OpContains {
-		// XXX: doing full scan because startKey does not apply here
-		// For example, if startKey = "account.owner=an" and search query = "accoutn.owner CONSISTS an"
-		// we can't iterate with prefix "account.owner=an" because we might miss keys like "account.owner=Ulan"
-		it := txi.store.Iterator(nil, nil)
+		// XXX: startKey does not apply here.
+		// For example, if startKey = "account.owner/an/" and search query = "accoutn.owner CONTAINS an"
```
ebuchman:
```suggestion
		// For example, if startKey = "account.owner/an/" and search query = "account.owner CONTAINS an"
```

state/txindex/kv/kv.go
```diff
@@ -340,10 +340,10 @@ func (txi *TxIndex) match(c query.Condition, startKey []byte) (hashes [][]byte)
...
 		}
 	} else if c.Op == query.OpContains {
-		// XXX: doing full scan because startKey does not apply here
-		// For example, if startKey = "account.owner=an" and search query = "accoutn.owner CONSISTS an"
-		// we can't iterate with prefix "account.owner=an" because we might miss keys like "account.owner=Ulan"
-		it := txi.store.Iterator(nil, nil)
+		// XXX: startKey does not apply here.
+		// For example, if startKey = "account.owner/an/" and search query = "account.owner CONTAINS an"
+		// we can't iterate with prefix "account.owner/an/" because we might miss keys like "account.owner/Ulan/"
+		it := dbm.IteratePrefix(txi.store, []byte(fmt.Sprintf("%s/", c.Tag)))
```
ebuchman: Should we use the new `startKey` for this from #2925 ?

melekes: yes

Issue #2908, opened 2018-11-24
False positive when searching for a prefix of a tag value


feldgendler:
*Tendermint version:* 0.22.8

The patch below adds a unit test that fails:

```
--- a/state/txindex/kv/kv_test.go
+++ b/state/txindex/kv/kv_test.go
@@ -89,6 +89,8 @@ func TestTxSearch(t *testing.T) {
                {"account.owner CONTAINS 'an'", 1},
                // search using CONTAINS
                {"account.owner CONTAINS 'Vlad'", 0},
+               // search using a prefix of the stored value
+               {"account.owner = 'Iv'", 0},
        }

        for _, tc := range testCases {
```

The stored value for `account.owner` is `Ivan`, but searching for just `Iv` produces a false positive.

The reason this happens is that `startKey` in `state/txindex/kv/kv.go` returns a prefix that matches too much: `account.owner/Iv` in this case. To avoid false positives, it should rather return `account.owner/Iv/`. 

There is _probably_ a similar bug where `matchRange` looks up `key` when it really should look up `key/`.

ebuchman:
Merged to develop. To be released in v0.26.4. Thanks!

feldgendler:
Thank you!

What about range matching? I haven't had the time to test that, but I suspect a similar bug.

feldgendler:
I took another look at the code. `Search` passes `[]byte(r.key)` as `prefix` into `matchRange`, which calls `dbm.IteratePrefix` for it. This looks prone to the same problem: it will find any longer keys of which the one specified is a prerfix. Either `Search` or `matchRange` has to append a slash to the key.

melekes:
Missed it. Thanks for spotting!

PR #2925, opened 2018-11-27
kv indexer: add separator to start key when matching ranges

melekes:
to avoid including false positives

Refs #2908

* [ ] Wrote tests

### DIFF
--- a/state/txindex/kv/kv.go
+++ b/state/txindex/kv/kv.go
@@ -172,10 +172,10 @@ func (txi *TxIndex) Search(q *query.Query) ([]*types.TxResult, error) {
 
 		for _, r := range ranges {
 			if !hashesInitialized {
-				hashes = txi.matchRange(r, []byte(fmt.Sprintf("%s/", r.key)))
+				hashes = txi.matchRange(r, startKey(r.key))
 				hashesInitialized = true
 			} else {
-				hashes = intersect(hashes, txi.matchRange(r, []byte(fmt.Sprintf("%s/", r.key))))
+				hashes = intersect(hashes, txi.matchRange(r, startKey(r.key)))
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


state/txindex/kv/kv.go
```diff
@@ -429,11 +419,36 @@ func extractValueFromKey(key []byte) string {
...
+		result.Index,
+	))
 }
 
 func keyForHeight(result *types.TxResult) []byte {
-	return []byte(fmt.Sprintf("%s/%d/%d/%d", types.TxHeightKey, result.Height, result.Height, result.Index))
+	return []byte(fmt.Sprintf("%s/%d/%d/%d",
+		types.TxHeightKey,
+		result.Height,
+		result.Height,
```
ebuchman: Is height supposed to be here twice?

state/txindex/kv/kv.go
```diff
@@ -429,11 +419,36 @@ func extractValueFromKey(key []byte) string {
...
+	if height > 0 {
+		return startKey(c.Tag, c.Operand, height)
+	}
+	return startKey(c.Tag, c.Operand)
+}
+
+func startKey(fields ...interface{}) []byte {
+	var b bytes.Buffer
+	for _, f := range fields {
+		b.Write([]byte(fmt.Sprintf("%v", f) + tagKeySeparator))
```
ebuchman: Why the `+` when we're already using Sprintf?

state/txindex/kv/kv.go
melekes: Yes. It's sort of a hack.

state/txindex/kv/kv.go
melekes: You mean?

```suggestion
		b.Write([]byte(fmt.Sprintf("%v%v", f, tagKeySeparator)))
```

state/txindex/kv/kv.go
Liamsi: Should we add a comment here about the rationale of this hack?

melekes:
@Liamsi thanks! added a test case

PR #2929, opened 2018-11-28
Refactor updateState #2865

### LINKED ISSUE #2865: state: don't pass logger to updateState
See https://github.com/tendermint/tendermint/pull/2848#discussion_r234007969

In that PR, we fix the way we're logging the validator updates. That's great, but it involved passing in the logger to the updateState function. Instead, we should process the ABCI responses to the Tendermint form (ie. call PB2TM) before calling updateState so we can log them before calling updateState. Then we don't need to pass the logger in there and we can pass the TM version of things instead of ABCI responses to the updateState method

danil-lashin:
Fixes #2865

* [x] Wrote tests

### DIFF
--- a/state/execution_test.go
+++ b/state/execution_test.go
@@ -218,7 +218,11 @@ func TestUpdateValidators(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			_, err := updateValidators(tc.currentSet, tc.abciUpdates)
+			updates, err := types.PB2TM.ValidatorUpdates(tc.abciUpdates)
+			if err != nil {
+				assert.NoError(t, err)
+			}
+			err = updateValidators(tc.currentSet, updates)
 			if tc.shouldErr {
 				assert.Error(t, err)
 			} else {
--- a/state/execution_test.go
+++ b/state/execution_test.go
@@ -219,9 +219,7 @@ func TestUpdateValidators(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			updates, err := types.PB2TM.ValidatorUpdates(tc.abciUpdates)
-			if err != nil {
-				assert.NoError(t, err)
-			}
+			assert.NoError(t, err)
 			err = updateValidators(tc.currentSet, updates)
 			if tc.shouldErr {
 				assert.Error(t, err)
--- a/state/state_test.go
+++ b/state/state_test.go
@@ -5,12 +5,10 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/tendermint/tendermint/libs/log"
-
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	abci "github.com/tendermint/tendermint/abci/types"
-	crypto "github.com/tendermint/tendermint/crypto"
+	"github.com/tendermint/tendermint/crypto"
 	"github.com/tendermint/tendermint/crypto/ed25519"
 	cmn "github.com/tendermint/tendermint/libs/common"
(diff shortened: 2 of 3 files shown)


state/execution_test.go
```diff
@@ -218,7 +218,11 @@ func TestUpdateValidators(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			_, err := updateValidators(tc.currentSet, tc.abciUpdates)
+			updates, err := types.PB2TM.ValidatorUpdates(tc.abciUpdates)
+			if err != nil {
```
melekes:
```suggestion
assert.NoError(t, err)
```

`if err != nil` can be dropped.

state/state_test.go
```diff
@@ -303,7 +304,10 @@ func TestManyValidatorChangesSaveLoad(t *testing.T) {
 
 	// Save state etc.
 	var err error
-	state, err = updateState(log.TestingLogger(), state, blockID, &header, responses)
+	var validatorUpdates []*types.Validator
+	validatorUpdates, err = types.PB2TM.ValidatorUpdates(responses.EndBlock.ValidatorUpdates)
+	assert.Nil(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

state/state_test.go
```diff
@@ -223,14 +221,17 @@ func TestOneValidatorChangesSaveLoad(t *testing.T) {
...
 	for i := int64(1); i < highestHeight; i++ {
 		// When we get to a change height, use the next pubkey.
 		if changeIndex < len(changeHeights) && i == changeHeights[changeIndex] {
 			changeIndex++
 			power++
 		}
 		header, blockID, responses := makeHeaderPartsResponsesValPowerChange(state, i, power)
-		state, err = updateState(log.TestingLogger(), state, blockID, &header, responses)
+		validatorUpdates, err = types.PB2TM.ValidatorUpdates(responses.EndBlock.ValidatorUpdates)
+		assert.Nil(t, err)
```
melekes:
```suggestion
		require.NoError(t, err)
```

state/state_test.go
```diff
@@ -223,14 +221,17 @@ func TestOneValidatorChangesSaveLoad(t *testing.T) {
...
+		state, err = updateState(state, blockID, &header, responses, validatorUpdates)
 		assert.Nil(t, err)
```
melekes:
```suggestion
		require.NoError(t, err)
```

state/state_test.go
```diff
@@ -375,14 +379,17 @@ func TestConsensusParamsChangesSaveLoad(t *testing.T) {
...
 	for i := int64(1); i < highestHeight; i++ {
 		// When we get to a change height, use the next params.
 		if changeIndex < len(changeHeights) && i == changeHeights[changeIndex] {
 			changeIndex++
 			cp = params[changeIndex]
 		}
 		header, blockID, responses := makeHeaderPartsResponsesParams(state, i, cp)
-		state, err = updateState(log.TestingLogger(), state, blockID, &header, responses)
+		validatorUpdates, err = types.PB2TM.ValidatorUpdates(responses.EndBlock.ValidatorUpdates)
+		assert.Nil(t, err)
```
melekes:
```suggestion
		require.NoError(t, err)
```

state/state_test.go
```diff
@@ -5,12 +5,10 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/tendermint/tendermint/libs/log"
-
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
```
melekes:
```suggestion
	"github.com/stretchr/testify/require"

```

PR #2936, opened 2018-11-28
Add client.Start() to RPC WS docs

danil-lashin:
If we trying to use examples from [WS RPC docs](https://tendermint.com/rpc/#subscribe), we getting such panic: 

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x1643589]

goroutine 1 [running]:
github.com/daniildulin/rpc-test/vendor/github.com/tendermint/tendermint/rpc/lib/client.(*WSClient).Call(0x0, 0x1a1e480, 0xc000012f00, 0x183227a, 0x9, 0xc000817c30, 0x20, 0x18)
```

This happens because we need to start client service before using WS methods like `Subscribe`/`Unsubscribe`.

### DIFF
--- a/rpc/core/events.go
+++ b/rpc/core/events.go
@@ -54,11 +54,12 @@ import (
 // import "github.com/tendermint/tendermint/types"
 //
 // client := client.NewHTTP("tcp://0.0.0.0:26657", "/websocket")
+// err := client.Start()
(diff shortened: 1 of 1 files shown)


melekes:
I'll do the same for other methods.

PR #2941, opened 2018-11-29
types: ValidatorSet.Update preserves Accum

ebuchman:
This solves the other issue discovered as part of #2718,
where Accum is reset to 0 every time a validator is updated.

* [ ] Wrote tests


ValarDragon:
I know we need more comprehensive tests of the proposer selection algorithm anyway, in this PR can we at least amend existing valUpdate tests to ensure accum is preserved?

ebuchman:
Done. Thanks for the quality check. Also opened https://github.com/tendermint/tendermint/issues/2942

types/validator_set.go
```diff
@@ -272,13 +272,20 @@ func (vals *ValidatorSet) Add(val *Validator) (added bool) {
...
+// If the val is not found, it returns false; otherwise,
+// it returns true. The val.ProposerPriority field is ignored
+// and unchanged by this method.
 func (vals *ValidatorSet) Update(val *Validator) (updated bool) {
 	index, sameVal := vals.GetByAddress(val.Address)
 	if sameVal == nil {
 		return false
 	}
+	// Overwrite the ProposerPriority so it doesn't change.
+	// During block execution, val.ProposerPriority
```
ValarDragon: should this comment indicate that this happens since the incoming val comes from ABCI? (Maybe too excessive of a comment but should get resolved by our disentanglement in the long term)

ebuchman: yep, done.

types/validator_set.go
```diff
@@ -272,13 +272,22 @@ func (vals *ValidatorSet) Add(val *Validator) (added bool) {
...
 	if sameVal == nil {
 		return false
 	}
+	// Overwrite the ProposerPriority so it doesn't change.
+	// During block execution, the val passed in here comes
+	// from ABCI via PB2TM.ValidatorUpdates. Since ABCI
+	// doesn't know about ProposerPriority, PB2TM.ValidatorUpdates
+	// uses the default value of 0, which would cause issues for
+	// proposer selection every time a validator's voting power changes.
+	val.ProposerPriority = sameVal.ProposerPriority
```
Liamsi: Oh my, after all the hassle in IncrementAccum we didn't even fix the initial issue because it was overwritten again here.. Great find! Thanks for fixing this!

types/validator_set_test.go
```diff
@@ -62,8 +62,12 @@ func TestValidatorSetBasic(t *testing.T) {
 
 	// update
 	assert.False(t, vset.Update(randValidator_()))
-	val.VotingPower = 100
+	_, val = vset.GetByAddress(val.Address)
+	val.VotingPower += 100
+	proposerPriority := val.ProposerPriority
 	assert.True(t, vset.Update(val))
+	_, val = vset.GetByAddress(val.Address)
+	assert.Equal(t, proposerPriority, val.ProposerPriority)
```
Liamsi: Interestingly, this assertion doesn't fail even if we didn't add the line that aims to preserve the validator's ProposerPriority: https://github.com/tendermint/tendermint/pull/2941/files#diff-bd488e139d955daba658159f507010f7R290
Probably, due to the update not coming from `types.PB2TM.ValidatorUpdates` where the ProposerPriority would be `0`. To make this test, test the change made, we'd need to add a line `val.ProposerPriority = 0` right before the `Update`.

types/validator_set_test.go
```diff
@@ -62,8 +62,12 @@ func TestValidatorSetBasic(t *testing.T) {
 
 	// update
 	assert.False(t, vset.Update(randValidator_()))
-	val.VotingPower = 100
+	_, val = vset.GetByAddress(val.Address)
+	val.VotingPower += 100
+	proposerPriority := val.ProposerPriority
```
Liamsi: We still need proper tests as suggested in #2942 but this change would test the changed behaviour:
```suggestion
	proposerPriority := val.ProposerPriority
	// Mimic update from types.PB2TM.ValidatorUpdates which does not know about ProposerPriority
	// and hence chooses the default value 0.
	val.ProposerPriority = 0
```

PR #2945, opened 2018-11-29
docs: update ecosystem.json: add Rust ABCI

melekes:


docs/app-dev/ecosystem.json
```diff
@@ -133,6 +133,12 @@
       "language": "Javascript",
       "author": "Tendermint"
     },
+    {
+      "name": "rust-tsp",
```
Liamsi: Shouldn't this be renamed to rust-abci (the repo, too)?

melekes: Probably, but I don't have rights to do so.

ebuchman: Done!

PR #2947, opened 2018-11-29
rpc: broadcastTxCommit: check if deliverTxResCh is still open, return an err otherwise

### LINKED ISSUE #2408: [rpc] interface conversion: interface {} in nil, not EventDataTx
Reported via Riot.

rpc-error

melekes:
`deliverTxResCh`, like any other eventBus (pubsub) channel, is closed when eventBus is stopped. We must check if the channel is still open. The alternative approach is to not close any channels, which seems a bit odd.

Fixes #2408

* [ ] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -46,3 +46,4 @@ program](https://hackerone.com/tendermint).
     - keep accums averaged near 0
 - [types] \#2941 Preserve val.Accum during ValidatorSet.Update to avoid it being
   reset to 0 every time a validator is updated
+- [rpc] \#2408 `/broadcast_tx_commit`: Fix interface conversion: interface {} in nil, not EventDataTx error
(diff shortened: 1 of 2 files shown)


CHANGELOG_PENDING.md
```diff
@@ -46,3 +46,4 @@ program](https://hackerone.com/tendermint).
     - keep accums averaged near 0
 - [types] \#2941 Preserve val.Accum during ValidatorSet.Update to avoid it being
   reset to 0 every time a validator is updated
+- [rpc] \#2408 `/broadcast_tx_commit`: Fix interface conversion: interface {} in nil, not EventDataTx error
```
Liamsi: This might be slightly too cryptic. Maybe:
```suggestion
- [rpc] \#2408 `/broadcast_tx_commit`: Check if (`deliverTxRes`) channel is still open before casting to expected type (`types.EventDataTx`)
```

PR #2870, opened 2018-11-17
R4R: Split immutable and mutable parts of priv_validator.json

yutianwu:
refs: 
https://github.com/tendermint/tendermint/issues/1181
https://github.com/tendermint/tendermint/pull/2751

* [x] Updated all relevant documentation in docs
* [x] Updated all code comments where relevant
* [x] Wrote tests
* [x] Updated CHANGELOG_PENDING.md

Issue #2677, opened 2018-10-19
privval: non-deterministic test failure


ebuchman:
https://circleci.com/gh/tendermint/tendermint/30540#tests/containers/3

```
panic: test timed out after 5m0s

goroutine 134 [running]:
testing.(*M).startAlarm.func1()
	/usr/local/go/src/testing/testing.go:1240 +0x146
created by time.goFunc
	/usr/local/go/src/time/sleep.go:172 +0x52

goroutine 1 [chan receive, 5 minutes]:
testing.(*T).Run(0xc420139ef0, 0xe03db6, 0x15, 0xf46918, 0xc4200adc01)
	/usr/local/go/src/testing/testing.go:825 +0x597
testing.runTests.func1(0xc420139ef0)
	/usr/local/go/src/testing/testing.go:1063 +0xa5
testing.tRunner(0xc420139ef0, 0xc4200add48)
	/usr/local/go/src/testing/testing.go:777 +0x16e
testing.runTests(0xc4204e07a0, 0x1384140, 0x15, 0x15, 0xc42049c080)
	/usr/local/go/src/testing/testing.go:1061 +0x4e2
testing.(*M).Run(0xc42049c080, 0x0)
	/usr/local/go/src/testing/testing.go:978 +0x2ce
main.main()
	_testmain.go:134 +0x325

goroutine 5 [syscall, 5 minutes]:
os/signal.signal_recv(0x488741)
	/usr/local/go/src/runtime/sigqueue.go:139 +0xa6
os/signal.loop()
	/usr/local/go/src/os/signal/signal_unix.go:22 +0x30
created by os/signal.init.0
	/usr/local/go/src/os/signal/signal_unix.go:28 +0x4f

goroutine 132 [runnable]:
time.Sleep(0xf4240)
	/usr/local/go/src/runtime/time.go:102 +0x146
github.com/tendermint/tendermint/privval.(*RemoteSigner).connect(0xc4209981a0, 0xc400000001, 0x5, 0x7f36f8145d90, 0x0)
	/go/src/github.com/tendermint/tendermint/privval/socket.go:423 +0x9ed
github.com/tendermint/tendermint/privval.(*RemoteSigner).OnStart(0xc4209981a0, 0xc420900900, 0x15)
	/go/src/github.com/tendermint/tendermint/privval/socket.go:396 +0x5e
[...]

PR #2948, opened 2018-11-29
privval: remove GetAddress(), memoize pubkey

### LINKED ISSUE #2926: Consensus Failure when remote signer connection drops
**Tendermint version** 
develop branch

**What happened**:

When the remote signer drops the connection:
`CONSENSUS FAILURE!!!                         module=consensus err=EOF` or
or
`CONSENSUS FAILURE!!!                         module=consensus err=remote signer timed out`

Additionally,  `ConsensusState/receiveRoutine` shuts down.

A clearer error message and tendermint should probably retry / continue receiving. 

Probably, these methods should not panic but return an error instead:
https://github.com/tendermint/tendermint/blob/b771798d48c27630fb37b74c8e229da5a8b2042c/privval/remote_signer.go#L36-L53

Depending on the error (timeout/eof vs unknown other), we should continue or exit.

`./tendermint node --priv_validator_laddr=tcp://127.0.0.1:26659 --proxy_app=kvstore `
and start e.g. start `priv_val_server` (or the kms) as a remote signer and and shut it down after a few rounds of signing.

<details>
<summary> log (click to expand)</summary>

```
I[27116-11-27|19:20:13.952] Executed block                               module=state height=492 validTxs=0 invalidTxs=0
I[27116-11-27|19:20:13.952] Committed state                              module=state height=492 txs=0 appHash=0000000000000000
I[27116-11-27|19:20:13.953] Indexed block                                module=txindex height=492
I[27116-11-27|19:20:14.954] Timed out                                    module=consensus dur=998.016427ms height=493 round=0 step=RoundStepNewHeight
I[27116-11-27|19:20:14.954] enterNewRound(493/0). Current: 493/0/RoundStepNewHeight module=consensus height=493 round=0
I[27116-11-27|19:20:14.954] enterPropose(493/0). Current: 493/0/RoundStepNewRound module=consensus height=493 round=0c

E[27116-11-27|19:20:14.955] CONSENSUS FAILURE!!!                         module=consensus err=EOF stack="goroutine 93 [running]:\nruntime/debug.Stack(0xc420ca5550, 0x1, 0x1)\n\t/usr/local/Cellar/go/1.10.3/libexec/src/runtime/debug/stack.go:24 +0xa7\ngithub.com/tendermint/tendermint/consensus.(*ConsensusState).receiveRoutine.func2(0xc4200b6a80, 0x1b74350)\n\t/Users/ismail/go/src/github.com/tendermint/tendermint/consensus/state.go:583 +0xf9\npanic(0x18dc060, 0xc420074040)\n\t/usr/local/Cellar/go/1.10.3/libexec/src/runtime/panic.go:502 +0x229\ngithub.com/tendermint/tendermint/privval.(*RemoteSignerClient).GetAddress(0xc42019e560, 0x0, 0x0, 0x0)\n\t/Users/ismail/go/src/github.com/tendermint/tendermint/privval/remote_signer.go:39 +0x91\ngithub.com/tendermint/tendermint/consensus.(*ConsensusState).enterPropose(0xc4200b6a80, 0x1ed, 0x0)\n\t/Users/ismail/go/src/github.com/tendermint/tendermint/consensus/state.go:872 +0x613\ngithub.com/tendermint/tendermint/consensus.(*ConsensusState).enterNewRound(0xc4200b6a80, 0x1ed, 0x0)\n\t/Users/ismail/go/src/github.com/tendermint/tendermint/consensus/state.go:790 +0x81a\ngithub.com/tendermint/tendermint/consensus.(*ConsensusState).handleTimeout(0xc4200b6a80, 0x3b7c85ab, 0x1ed, 0x0, 0x1, 0x1ed, 0x0, 0x1, 0x38b16ab2, 0xed38f81de, ...)\n\t/Users/ismail/go/src/github.com/tendermint/tendermint/consensus/state.go:701 +0x551\ngithub.com/tendermint/tendermint/consensus.(*ConsensusState).receiveRoutine(0xc4200b6a80, 0x0)\n\t/Users/ismail/go/src/github.com/tendermint/tendermint/consensus/state.go:623 +0x430\ncreated by github.com/tendermint/tendermint/consensus.(*ConsensusState).OnStart\n\t/Users/ismail/go/src/github.com/tendermint/tendermint/consensus/state.go:307 +0x140\n"
I[27116-11-27|19:20:14.955] Stopping baseWAL                             module=consensus wal=/Users/ismail/.tendermint/data/cs.wal/wal impl=baseWAL
I[27116-11-27|19:20:14.955] Stopping Group                               module=consensus wal=/Users/ismail/.tendermint/data/cs.wal/wal impl=Group
E[27116-11-27|19:20:15.853] Ping                                         module=privval err="remote signer timed out"
```

</details>

------
ref https://github.com/tendermint/kms/issues/116
ref https://github.com/tendermint/tendermint/pull/2923

### LINKED ISSUE #2981: Remove PrivValidator.GetAddress
Remove `PrivValidator.GetAddress` from `PrivValidator` interface. 
Rationale: In some places we do `PrivValidator.GetPublicKey` and then `PrivValidator.GetAddress` which is just a wrapper around the former for convenience. In the remote signer / kms case this causes another (unnecessary) network request. Removing `GetAddress` makes this more obvious. 

ref: https://github.com/tendermint/tendermint/pull/2948#discussion_r239737037

Liamsi:
Change the remote signer such that `GetPublicKey()` returns the public key it memoized on initialization. If retrieving the public key on init, the constructor will return an error (which can be  used to retry). As a side-effect this PR removes `GetAddress()` from the `PrivValidator` interface (as the address can easily be retrieved from the public key (resolves #2981). 
Also, this PR splits up `PubKeyMsg` into a request and response message. The latter change is consistent with the other privval messages.

This (partly) resolves #2926: Instead of the Consensus failure their, which shuts down the ConsensusState, the error message now reads (for example):
```
I[29116-11-29|18:53:11.047] Committed state                              module=state height=528 txs=0 appHash=0000000000000000
E[29116-11-29|18:53:11.867] Ping                                         module=privval err=EOF
E[29116-11-29|18:53:12.047] Could not retrieve potenial validator address module=consensus height=529 round=0 TCPVal=(MISSING)
E[29116-11-29|18:53:13.867] Ping                                         module=privval err="remote signer timed out"
E[29116-11-29|18:53:15.047] Error signing vote. Could not retrieve privValidator's address. module=consensus privValidator=TCPVal height=529 round=0
E[29116-11-29|18:53:15.869] Ping                                         module=privval err="remote signer timed out"
E[29116-11-29|18:53:17.867] Ping                                         module=privval err="remote signer timed out"
E[29116-11-29|18:53:19.865] Ping                                         module=privval err="remote signer timed out"
```

This PR does not restart the remote signer yet (which will be handled in #2923). But as this is a rather big refactoring, I thought it might be worth keeping this in a separate PR. This is a breaking change.

* [ ] Wrote tests


privval/tcp_test.go
```diff
@@ -25,15 +25,13 @@ func TestSocketPVAddress(t *testing.T) {
...
+	assert.NoError(t, err)
 
-	clientAddr := sc.GetAddress()
+	clientAddr, err := sc.GetAddress()
+	assert.NoError(t, err)
 
 	assert.Equal(t, serverAddr, clientAddr)
-
-	// TODO(xla): Remove when PrivValidator2 replaced PrivValidator.
-	assert.Equal(t, serverAddr, sc.GetAddress())
```
Liamsi: Not sure what this comment means. The above assertion `assert.Equal(t, serverAddr, clientAddr)` should test the same, or not? `sc.GetAddress()` does not change between the two asserts. Hence, removed.

cmd/tendermint/commands/init.go
```diff
@@ -57,9 +57,13 @@ func initFilesWithConfig(config *cfg.Config) error {
 			GenesisTime:     tmtime.Now(),
 			ConsensusParams: types.DefaultConsensusParams(),
 		}
+		key, err := pv.GetPubKey()
+		if err != nil {
+			return err
```
melekes: `errrors.Wrap(err, "failed to get pubkey from private validator")` ?

cmd/tendermint/commands/show_validator.go
```diff
@@ -17,6 +18,11 @@ var ShowValidatorCmd = &cobra.Command{
 
 func showValidator(cmd *cobra.Command, args []string) {
 	privValidator := privval.LoadOrGenFilePV(config.PrivValidatorFile())
-	pubKeyJSONBytes, _ := cdc.MarshalJSON(privValidator.GetPubKey())
+	key, err := privValidator.GetPubKey()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Could not read public key from FilePV %v", privValidator)
```
melekes:
```suggestion
		fmt.Fprintf(os.Stderr, "Failed to read pubkey from private validator file (%s): %v", config.PrivValidatorFile(), err)
```

cmd/tendermint/commands/testnet.go
```diff
@@ -90,9 +90,13 @@ func testnetFiles(cmd *cobra.Command, args []string) error {
 
 		pvFile := filepath.Join(nodeDir, config.BaseConfig.PrivValidator)
 		pv := privval.LoadFilePV(pvFile)
+		pubKey, err := pv.GetPubKey()
+		if err != nil {
+			return err
```
melekes: errors.Wrap?

consensus/reactor_test.go
```diff
@@ -143,7 +143,9 @@ func TestReactorWithEvidence(t *testing.T) {
 		// mock the evidence pool
 		// everyone includes evidence of another double signing
 		vIdx := (i + 1) % nValidators
-		evpool := newMockEvidencePool(privVals[vIdx].GetAddress())
+		addr, err := privVals[vIdx].GetAddress()
+		assert.NoError(t, err)
```
melekes:
```suggestion
		require.NoError(t, err)
```

consensus/reactor_test.go
```diff
@@ -268,7 +270,9 @@ func TestReactorVotingPowerChange(t *testing.T) {
 	// map of active validators
 	activeVals := make(map[string]struct{})
 	for i := 0; i < nVals; i++ {
-		activeVals[string(css[i].privValidator.GetAddress())] = struct{}{}
+		addr, err := css[i].privValidator.GetAddress()
+		assert.NoError(t, err)
```
melekes:
```suggestion
		require.NoError(t, err)
```

consensus/reactor_test.go
```diff
@@ -279,7 +283,9 @@ func TestReactorVotingPowerChange(t *testing.T) {
 	//---------------------------------------------------------------------------
 	logger.Debug("---------------------------- Testing changing the voting power of one validator a few times")
 
-	val1PubKey := css[0].privValidator.GetPubKey()
+	pubKey, err := css[0].privValidator.GetPubKey()
+	assert.NoError(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

consensus/reactor_test.go
```diff
@@ -331,7 +337,9 @@ func TestReactorValidatorSetChanges(t *testing.T) {
 	// map of active validators
 	activeVals := make(map[string]struct{})
 	for i := 0; i < nVals; i++ {
-		activeVals[string(css[i].privValidator.GetAddress())] = struct{}{}
+		addr, err := css[i].privValidator.GetAddress()
+		assert.NoError(t, err)
```
melekes:
```suggestion
		require.NoError(t, err)
```

consensus/reactor_test.go
```diff
@@ -342,7 +350,9 @@ func TestReactorValidatorSetChanges(t *testing.T) {
 	//---------------------------------------------------------------------------
 	logger.Info("---------------------------- Testing adding one validator")
 
-	newValidatorPubKey1 := css[nVals].privValidator.GetPubKey()
+	pubKey, err := css[nVals].privValidator.GetPubKey()
+	assert.NoError(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

consensus/reactor_test.go
```diff
@@ -369,7 +379,9 @@ func TestReactorValidatorSetChanges(t *testing.T) {
 	//---------------------------------------------------------------------------
 	logger.Info("---------------------------- Testing changing the voting power of one validator")
 
-	updateValidatorPubKey1 := css[nVals].privValidator.GetPubKey()
+	pubKey, err = css[nVals].privValidator.GetPubKey()
+	assert.NoError(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

consensus/reactor_test.go
```diff
@@ -386,11 +398,15 @@ func TestReactorValidatorSetChanges(t *testing.T) {
 	//---------------------------------------------------------------------------
 	logger.Info("---------------------------- Testing adding two validators at once")
 
-	newValidatorPubKey2 := css[nVals+1].privValidator.GetPubKey()
+	pubKey, err = css[nVals+1].privValidator.GetPubKey()
+	assert.NoError(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

consensus/reactor_test.go
```diff
@@ -386,11 +398,15 @@ func TestReactorValidatorSetChanges(t *testing.T) {
...
+	newValidatorPubKey2 := pubKey
 	newVal2ABCI := types.TM2PB.PubKey(newValidatorPubKey2)
 	newValidatorTx2 := kvstore.MakeValSetChangeTx(newVal2ABCI, testMinPower)
 
-	newValidatorPubKey3 := css[nVals+2].privValidator.GetPubKey()
+	pubKey, err = css[nVals+2].privValidator.GetPubKey()
+	assert.NoError(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

consensus/replay_test.go
```diff
@@ -331,7 +331,9 @@ func testHandshakeReplay(t *testing.T, nBlocks int, mode uint) {
 	chain, commits, err := makeBlockchainFromWAL(wal)
 	require.NoError(t, err)
 
-	stateDB, state, store := stateAndStore(config, privVal.GetPubKey(), kvstore.ProtocolVersion)
+	pubKey, err := privVal.GetPubKey()
+	assert.NoError(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

consensus/replay_test.go
```diff
@@ -634,7 +636,9 @@ func TestInitChainUpdateValidators(t *testing.T) {
 
 	config := ResetConfig("proxy_test_")
 	privVal := privval.LoadFilePV(config.PrivValidatorFile())
-	stateDB, state, store := stateAndStore(config, privVal.GetPubKey(), 0x0)
+	pubKey, err := privVal.GetPubKey()
+	assert.NoError(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

consensus/state.go
```diff
@@ -829,22 +829,27 @@ func (cs *ConsensusState) enterPropose(height int64, round int) {
 	}
 
 	// if not a validator, we're done
-	if !cs.Validators.HasAddress(cs.privValidator.GetAddress()) {
-		logger.Debug("This node is not a validator", "addr", cs.privValidator.GetAddress(), "vals", cs.Validators)
+	address, err := cs.privValidator.GetAddress()
+	if err != nil {
+		logger.Error("Could not retrieve potenial validator address", cs.privValidator)
```
```suggestion
		logger.Error("Failed to get private validator address", "err", err)
```

consensus/state.go
```diff
@@ -929,7 +934,11 @@ func (cs *ConsensusState) createProposalBlock() (block *types.Block, blockParts
 		cs.state.Validators.Size(),
 		len(evidence),
 	), maxGas)
-	proposerAddr := cs.privValidator.GetAddress()
+	proposerAddr, err := cs.privValidator.GetAddress()
+	if err != nil {
+		cs.Logger.Error("could not retrieve proposer's address from privValidator", cs.privValidator)
```
```suggestion
		cs.Logger.Error("Failed to get private validator address", "err", err)
```

consensus/state.go
```diff
@@ -1474,7 +1483,9 @@ func (cs *ConsensusState) tryAddVote(vote *types.Vote, peerID p2p.ID) (bool, err
 		if err == ErrVoteHeightMismatch {
 			return added, err
 		} else if voteErr, ok := err.(*types.ErrVoteConflictingVotes); ok {
-			if bytes.Equal(vote.ValidatorAddress, cs.privValidator.GetAddress()) {
+			addr, err := cs.privValidator.GetAddress()
+			cs.Logger.Error("Can not retrieve privValidator's address", "err", err)
```
melekes: where's if?

melekes:
```suggestion
			if err != nil {
				cs.Logger.Error("Failed to get private validator address", "err", err)
				return
			}

File: consensus/state.go [hunk 18]
```diff
@@ -1639,7 +1650,11 @@ func (cs *ConsensusState) addVote(vote *types.Vote, peerID p2p.ID) (added bool,
 }
 
 func (cs *ConsensusState) signVote(type_ types.SignedMsgType, hash []byte, header types.PartSetHeader) (*types.Vote, error) {
-	addr := cs.privValidator.GetAddress()
+	addr, err := cs.privValidator.GetAddress()
+	if err != nil {
+		cs.Logger.Error("Could not retrieve privValidator's address. The remote signer's connection might have dropped?", "err", err)
```
```suggestion
		cs.Logger.Error("Failed to get private validator address", "err", err)
```

consensus/state.go
```diff
@@ -1639,7 +1650,11 @@ func (cs *ConsensusState) addVote(vote *types.Vote, peerID p2p.ID) (added bool,
...
+		return nil, err
```
melekes: `errors.Wrap(err, "Failed to get private validator address")`

consensus/state.go
```diff
@@ -1675,7 +1690,12 @@ func (cs *ConsensusState) voteTime() time.Time {
 // sign the vote and publish on internalMsgQueue
 func (cs *ConsensusState) signAddVote(type_ types.SignedMsgType, hash []byte, header types.PartSetHeader) *types.Vote {
 	// if we don't have a key or we're not in the validator set, do nothing
-	if cs.privValidator == nil || !cs.Validators.HasAddress(cs.privValidator.GetAddress()) {
+	privValAddr, err := cs.privValidator.GetAddress()
+	if err != nil {
+		cs.Logger.Error("Error signing vote. Could not retrieve privValidator's address.", "privValidator", cs.privValidator, "height", cs.Height, "round", cs.Round)
```
```suggestion
		cs.Logger.Error("Failed to get private validator address", "err", err, "height", cs.Height, "round", cs.Round)
```

consensus/state_test.go
```diff
@@ -73,7 +73,9 @@ func TestStateProposerSelection0(t *testing.T) {
 
 	// Commit a block and ensure proposer for the next height is correct.
 	prop := cs1.GetRoundState().Validators.GetProposer()
-	if !bytes.Equal(prop.Address, cs1.privValidator.GetAddress()) {
+	address, err := cs1.privValidator.GetAddress()
+	assert.NoError(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

consensus/state_test.go
```diff
@@ -87,7 +89,9 @@ func TestStateProposerSelection0(t *testing.T) {
 	ensureNewRound(newRoundCh, height+1, 0)
 
 	prop = cs1.GetRoundState().Validators.GetProposer()
-	if !bytes.Equal(prop.Address, vss[1].GetAddress()) {
+	addr, err := vss[1].GetAddress()
+	assert.NoError(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

consensus/state_test.go
```diff
@@ -110,7 +114,9 @@ func TestStateProposerSelection2(t *testing.T) {
 	// everyone just votes nil. we get a new proposer each round
 	for i := 0; i < len(vss); i++ {
 		prop := cs1.GetRoundState().Validators.GetProposer()
-		correctProposer := vss[(i+round)%len(vss)].GetAddress()
+		addr, err := vss[(i+round)%len(vss)].GetAddress()
+		assert.NoError(t, err)
```
melekes:
```suggestion
		require.NoError(t, err)
```

consensus/state_test.go
```diff
@@ -505,7 +511,9 @@ func TestStateLockPOLRelock(t *testing.T) {
 
 	timeoutWaitCh := subscribe(cs1.eventBus, types.EventQueryTimeoutWait)
 	proposalCh := subscribe(cs1.eventBus, types.EventQueryCompleteProposal)
-	voteCh := subscribeToVoter(cs1, cs1.privValidator.GetAddress())
+	addr, err := cs1.privValidator.GetAddress()
+	assert.NoError(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

consensus/state_test.go
```diff
@@ -596,7 +604,9 @@ func TestStateLockPOLUnlock(t *testing.T) {
 	timeoutWaitCh := subscribe(cs1.eventBus, types.EventQueryTimeoutWait)
 	newRoundCh := subscribe(cs1.eventBus, types.EventQueryNewRound)
 	unlockCh := subscribe(cs1.eventBus, types.EventQueryUnlock)
-	voteCh := subscribeToVoter(cs1, cs1.privValidator.GetAddress())
+	addr, err := cs1.privValidator.GetAddress()
+	assert.NoError(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

consensus/state_test.go
```diff
@@ -689,7 +699,9 @@ func TestStateLockPOLSafety1(t *testing.T) {
 	timeoutProposeCh := subscribe(cs1.eventBus, types.EventQueryTimeoutPropose)
 	timeoutWaitCh := subscribe(cs1.eventBus, types.EventQueryTimeoutWait)
 	newRoundCh := subscribe(cs1.eventBus, types.EventQueryNewRound)
-	voteCh := subscribeToVoter(cs1, cs1.privValidator.GetAddress())
+	addr, err := cs1.privValidator.GetAddress()
+	assert.NoError(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

cmd/tendermint/commands/init.go
Liamsi: Thanks! Done. I do not really understand where we use `cmn.ErrorWrap()` over `errors.Wrap` and vice versa. Did we establish some best-practice/guideline here I'm not aware of? (I prefer using `github.com/pkg/errors` over the common package anyways)

consensus/state.go
Liamsi: Thanks for catching this! (must be `return added, err` though)

consensus/common_test.go
```diff
@@ -72,16 +72,20 @@ func NewValidatorStub(privValidator types.PrivValidator, valIndex int) *validato
 }
 
 func (vs *validatorStub) signVote(voteType types.SignedMsgType, hash []byte, header types.PartSetHeader) (*types.Vote, error) {
+	addr, err := vs.PrivValidator.GetAddress()
+	if err != nil {
+		return nil, err
```
melekes: errors.Wrap?

privval/priv_validator_test.go
```diff
@@ -25,10 +25,13 @@ func TestGenLoadValidator(t *testing.T) {
 	height := int64(100)
 	privVal.LastHeight = height
 	privVal.Save()
-	addr := privVal.GetAddress()
+	addr, err := privVal.GetAddress() // FilePV doesn't err here
+	assert.NoError(err)
```
melekes:
```suggestion
	require.NoError(err)
```

privval/priv_validator_test.go
```diff
@@ -25,10 +25,13 @@ func TestGenLoadValidator(t *testing.T) {
 	height := int64(100)
 	privVal.LastHeight = height
 	privVal.Save()
-	addr := privVal.GetAddress()
+	addr, err := privVal.GetAddress() // FilePV doesn't err here
```
melekes: don't think comment is necessary if we're handling the error

also, it's not quite true. FilePV can error (in 0.1% cases due to file corruption or some other rare event)

privval/remote_signer.go
```diff
@@ -229,7 +229,10 @@ func handleRequest(req RemoteSignerMsg, chainID string, privVal types.PrivValida
 	switch r := req.(type) {
 	case *PubKeyMsg:
 		var p crypto.PubKey
-		p = privVal.GetPubKey()
+		p, err = privVal.GetPubKey()
+		if err != nil {
+
```
melekes: ??? can we add at least a comment

privval/priv_validator_test.go
```diff
@@ -25,10 +25,13 @@ func TestGenLoadValidator(t *testing.T) {
...
+	assert.NoError(err)
 
 	privVal = LoadFilePV(tempFile.Name())
-	assert.Equal(addr, privVal.GetAddress(), "expected privval addr to be the same")
+	loadedAddr, err := privVal.GetAddress()
+	assert.NoError(err)
```
melekes:
```suggestion
	require.NoError(err)
```

privval/priv_validator_test.go
```diff
@@ -42,9 +45,11 @@ func TestLoadOrGenValidator(t *testing.T) {
 		t.Error(err)
 	}
 	privVal := LoadOrGenFilePV(tempFilePath)
-	addr := privVal.GetAddress()
+	addr, err := privVal.GetAddress()
+	assert.NoError(err)
```
melekes:
```suggestion
	require.NoError(err)
```

types/vote_test.go
```diff
@@ -190,12 +191,13 @@ func TestIsVoteTypeValid(t *testing.T) {
 
 func TestVoteVerify(t *testing.T) {
 	privVal := NewMockPV()
-	pubkey := privVal.GetPubKey()
+	pubkey, err := privVal.GetPubKey()
+	assert.NoError(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

types/vote_test.go
```diff
@@ -140,13 +140,14 @@ func TestVoteProposalNotEq(t *testing.T) {
 
 func TestVoteVerifySignature(t *testing.T) {
 	privVal := NewMockPV()
-	pubkey := privVal.GetPubKey()
+	pubkey, err := privVal.GetPubKey()
+	assert.NoError(t, err)
```
melekes:
```suggestion
	require.NoError(t, err)
```

types/vote_set_test.go
```diff
@@ -474,20 +515,24 @@ func TestMakeCommit(t *testing.T) {
...
 			t.Error(err)
 		}
 	}
 
 	// The 8th voted like everyone else.
 	{
-		vote := withValidator(voteProto, privValidators[7].GetAddress(), 7)
-		_, err := signAddVote(privValidators[7], vote, voteSet)
+		addr, err := privValidators[7].GetAddress()
+		assert.NoError(t, err)
```
melekes:
```suggestion
		require.NoError(t, err)
```

privval/remote_signer.go
Liamsi: Sorry, I missed that one. I'll return the error. In a follow-ip PR we should split up `PubKeyMsg` into `PubKeyRequest` and `PubKeyResponse`. The latter should contain the error as with the other requests below. 
Thanks a lot for the through review!

consensus/common_test.go
Liamsi: In this case we will run into a panic shortly after. But yeah, it's more consistent and the panic would give a better indication what is going on without going through the stack trace.

cmd/tendermint/commands/show_validator.go
```diff
@@ -17,6 +18,11 @@ var ShowValidatorCmd = &cobra.Command{
 
 func showValidator(cmd *cobra.Command, args []string) {
 	privValidator := privval.LoadOrGenFilePV(config.PrivValidatorFile())
```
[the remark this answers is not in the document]
Liamsi: You are correct. 
ref: https://github.com/cosmos/cosmos-sdk/pull/2975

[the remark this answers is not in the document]
joe-bowman: Excellent, ty.

Liamsi:
Is there anything left to be done here @melekes?

consensus/common_test.go
```diff
@@ -72,16 +74,20 @@ func NewValidatorStub(privValidator types.PrivValidator, valIndex int) *validato
 }
 
 func (vs *validatorStub) signVote(voteType types.SignedMsgType, hash []byte, header types.PartSetHeader) (*types.Vote, error) {
+	addr, err := vs.PrivValidator.GetAddress()
+	if err != nil {
+		return nil, errors.Wrap(err, "Error while retrieving private validator's address")
```
melekes: better to use single style everywhere

```suggestion
		return nil, errors.Wrap(err, "failed to get address from private validator")
```

consensus/state.go
```diff
@@ -9,6 +9,8 @@ import (
 	"sync"
 	"time"
 
+	errs "github.com/pkg/errors"
```
melekes: again, this prefix is not common in this repo. sorry for being too picky, but single style reduces cognitive load on developer when reading code

node/node.go
```diff
@@ -235,16 +235,25 @@ func NewNode(config *cfg.Config,
 	fastSync := config.FastSync
 	if state.Validators.Size() == 1 {
 		addr, _ := state.Validators.GetByIndex(0)
-		if bytes.Equal(privValidator.GetAddress(), addr) {
+		privValAddr, err := privValidator.GetAddress()
+		if err != nil {
+			return nil, errors.Wrap(err, "Error while retrieving private validator's address")
```
```suggestion
			return nil, errors.Wrap(err, "failed to get address from private validator")
```

node/node.go
```diff
@@ -235,16 +235,25 @@ func NewNode(config *cfg.Config,
...
+		}
+		if bytes.Equal(privValAddr, addr) {
 			fastSync = false
 		}
 	}
 
+	pubKey, err := privValidator.GetPubKey()
+	if err != nil {
+		return nil, errors.Wrap(err, "Error while retrieving private validator's public key")
```
```suggestion
		return nil, errors.Wrap(err, "failed to get pubkey from private validator")
```

node/node.go
```diff
@@ -611,7 +620,11 @@ func (n *Node) ConfigureRPC() {
 	rpccore.SetEvidencePool(n.evidencePool)
 	rpccore.SetP2PPeers(n.sw)
 	rpccore.SetP2PTransport(n)
-	rpccore.SetPubKey(n.privValidator.GetPubKey())
+	pubKey, err := n.privValidator.GetPubKey()
+	if err != nil {
+		n.Logger.Error("Error configuring RPC. Can not retrieve privValidator's public key", "err", err)
```
melekes: same here. too many different variations of a single thing

privval/priv_validator_test.go
```diff
@@ -25,10 +25,13 @@ func TestGenLoadValidator(t *testing.T) {
...
 	privVal.Save()
-	addr := privVal.GetAddress()
+	addr, err := privVal.GetAddress()
+	require.NoError(t, err)
 
 	privVal = LoadFilePV(tempFile.Name())
-	assert.Equal(addr, privVal.GetAddress(), "expected privval addr to be the same")
+	loadedAddr, err := privVal.GetAddress()
+	require.NoError(t, err)
+	assert.Equal(addr, loadedAddr, "expected privval addr to be the same")
```
melekes:
```suggestion
	assert.Equal(addr, loadedAddr)
```

privval/priv_validator_test.go
```diff
@@ -42,9 +45,11 @@ func TestLoadOrGenValidator(t *testing.T) {
 		t.Error(err)
 	}
 	privVal := LoadOrGenFilePV(tempFilePath)
-	addr := privVal.GetAddress()
+	addr, err := privVal.GetAddress()
+	require.NoError(t, err)
 	privVal = LoadOrGenFilePV(tempFilePath)
-	assert.Equal(addr, privVal.GetAddress(), "expected privval addr to be the same")
+	loadedAddr, err := privVal.GetAddress()
+	assert.Equal(addr, loadedAddr, "expected privval addr to be the same")
```
melekes:
```suggestion
	assert.Equal(addr, loadedAddr)
```

types/priv_validator.go
```diff
@@ -12,8 +12,10 @@ import (
 // PrivValidator defines the functionality of a local Tendermint validator
 // that signs votes and proposals, and never double signs.
 type PrivValidator interface {
-	GetAddress() Address // redundant since .PubKey().Address()
-	GetPubKey() crypto.PubKey
+	// TODO: shouldn't we remove GetAddress? In case of a remote signer this will trigger
```
melekes: I'd vote for removing. Is there an issue?

melekes:
Also, I believe this needs a changelog entry

consensus/state.go
Liamsi: You are not too picky. It just seems there is no prefix/alias for this particular package in the project though. There is no other place where this is aliased. In other places we use `lerr` for `"github.com/tendermint/tendermint/lite/errors"` and `stderr` for `"errors"`. 

Would `pkgerr` be more suitable?

melekes: why not `errors`?

Liamsi: Because of https://github.com/tendermint/tendermint/pull/2948/files/df20323cd2f261ec8ec83462c866ce602371741a#diff-80343e80650fe43a4394fc26afd5cb3bR5

privval/priv_validator_test.go
Liamsi: BTW, while it's true that due to file corruption FilePv might error, it won't error on that specific function call (but it will fail earlier and exit in `LoadFilePV`). Once a FilePV object is loaded it won't err on `GetAddress`.

consensus/state.go
melekes: you can remove standard `errors` package. `pkg/errors` implements standard interface

node/node.go
melekes: that sounds even better!

types/priv_validator.go
melekes: awesome. thanks

consensus/state.go
Liamsi: Done, thanks!

node/node.go
Liamsi: Done in https://github.com/tendermint/tendermint/pull/2948/commits/5c09275691b2190d8747f4f1b1817f63114b1c6b
Besides the wrapping in remote signer only the logger messages are left to give more context.

cmd/tendermint/commands/testnet.go
Liamsi: Changed as per: https://github.com/tendermint/tendermint/pull/2948#discussion_r239760333

cmd/tendermint/commands/init.go
melekes: I'd propose we use `errors.Wrap` and if/when we decide to switch to cmn.Error, we'll replace errors.Wrap with cmn.ErrorWrap

consensus/state.go
```diff
@@ -929,7 +935,11 @@ func (cs *ConsensusState) createProposalBlock() (block *types.Block, blockParts
 		cs.state.Validators.Size(),
 		len(evidence),
 	), maxGas)
-	proposerAddr := cs.privValidator.GetAddress()
+	proposerAddr, err := cs.privValidator.GetAddress()
+	if err != nil {
+		cs.Logger.Error("enterPropose: Cannot create block without private validator's address:", "err", err)
```
```suggestion
		cs.Logger.Error("enterPropose: Cannot create block without private validator's address", "err", err)
```

node/node.go
```diff
@@ -616,7 +625,11 @@ func (n *Node) ConfigureRPC() {
 	rpccore.SetEvidencePool(n.evidencePool)
 	rpccore.SetP2PPeers(n.sw)
 	rpccore.SetP2PTransport(n)
-	rpccore.SetPubKey(n.privValidator.GetPubKey())
+	pubKey, err := n.privValidator.GetPubKey()
+	if err != nil {
+		n.Logger.Error("Error configuring RPC.", "err", err)
```
```suggestion
		n.Logger.Error("Error configuring RPC", "err", err)
```

melekes:
So after this change, if remote signer is unavailable, Tendermint will just refuse to propose & votes nil. Right? Won't it be better to retry N times instead and fail if no success? 

Do we memoize the pubkey/address the first time we get them?

Liamsi:
> So after this change, if remote signer is unavailable, Tendermint will just refuse to propose & votes nil. Right? Won't it be better to retry N times instead and fail if no success?

Yes, that's correct. Now that we do not panic, and we return an error, we can handle this case gracefully. I experimented a bit with #2923 in combination with this PR and it was at least possible to retry now.

> Do we memoize the pubkey/address the first time we get them?

No, we don't. I think we should.

joe-bowman:
I agree. Caching this locally makes sense, especially with an external
validator where there is a real life cost to these calls (e.g. Yubihsm
being comparatively slow). I guess it was just less costly before pulling
from a local file.

privval/remote_signer.go
```diff
@@ -152,7 +154,8 @@ type RemoteSignerMsg interface{}
 
 func RegisterRemoteSignerMsg(cdc *amino.Codec) {
 	cdc.RegisterInterface((*RemoteSignerMsg)(nil), nil)
-	cdc.RegisterConcrete(&PubKeyMsg{}, "tendermint/remotesigner/PubKeyMsg", nil)
+	cdc.RegisterConcrete(&PubKeyRequest{}, "tendermint/remotesigner/PubKeyRequest", nil)
+	cdc.RegisterConcrete(&PubKeyResponse{}, "tendermint/remotesigner/PubKeyResponse", nil)
```
Liamsi: Splitting up `PubKeyMsg` into  `PubKeyRequest ` and `PubKeyResponse` is the reason this is marked as breaking. If that blocks merging this PR I can revert this change and follow up in a separate PR.

Liamsi:
@melekes @ebuchman this is ready for another review.

consensus/reactor_test.go
```diff
@@ -279,7 +281,9 @@ func TestReactorVotingPowerChange(t *testing.T) {
 	//---------------------------------------------------------------------------
 	logger.Debug("---------------------------- Testing changing the voting power of one validator a few times")
 
-	val1PubKey := css[0].privValidator.GetPubKey()
+	pubKey := css[0].privValidator.GetPubKey()
```
melekes: why create extra variable?

cmd/tendermint/commands/show_validator.go
```diff
@@ -16,7 +16,10 @@ var ShowValidatorCmd = &cobra.Command{
 }
 
 func showValidator(cmd *cobra.Command, args []string) {
+	// TODO(ismail): add a flag and check if we actually want to see the pub key
+	//  of the remote signer instead of the FilePV
 	privValidator := privval.LoadOrGenFilePV(config.PrivValidatorFile())
-	pubKeyJSONBytes, _ := cdc.MarshalJSON(privValidator.GetPubKey())
+	key := privValidator.GetPubKey()
```
melekes: same question

consensus/reactor_test.go
```diff
@@ -342,7 +347,8 @@ func TestReactorValidatorSetChanges(t *testing.T) {
 	//---------------------------------------------------------------------------
 	logger.Info("---------------------------- Testing adding one validator")
 
-	newValidatorPubKey1 := css[nVals].privValidator.GetPubKey()
+	pubKey := css[nVals].privValidator.GetPubKey()
```
melekes: same question

consensus/reactor_test.go
```diff
@@ -369,7 +375,8 @@ func TestReactorValidatorSetChanges(t *testing.T) {
 	//---------------------------------------------------------------------------
 	logger.Info("---------------------------- Testing changing the voting power of one validator")
 
-	updateValidatorPubKey1 := css[nVals].privValidator.GetPubKey()
+	pubKey = css[nVals].privValidator.GetPubKey()
```
melekes: same

consensus/reactor_test.go
```diff
@@ -386,11 +393,13 @@ func TestReactorValidatorSetChanges(t *testing.T) {
 	//---------------------------------------------------------------------------
 	logger.Info("---------------------------- Testing adding two validators at once")
 
-	newValidatorPubKey2 := css[nVals+1].privValidator.GetPubKey()
+	pubKey = css[nVals+1].privValidator.GetPubKey()
```
melekes: same

consensus/reactor_test.go
```diff
@@ -386,11 +393,13 @@ func TestReactorValidatorSetChanges(t *testing.T) {
...
+	newValidatorPubKey2 := pubKey
 	newVal2ABCI := types.TM2PB.PubKey(newValidatorPubKey2)
 	newValidatorTx2 := kvstore.MakeValSetChangeTx(newVal2ABCI, testMinPower)
 
-	newValidatorPubKey3 := css[nVals+2].privValidator.GetPubKey()
+	pubKey = css[nVals+2].privValidator.GetPubKey()
```
melekes: same

consensus/replay_test.go
```diff
@@ -331,7 +331,9 @@ func testHandshakeReplay(t *testing.T, nBlocks int, mode uint) {
 	chain, commits, err := makeBlockchainFromWAL(wal)
 	require.NoError(t, err)
 
-	stateDB, state, store := stateAndStore(config, privVal.GetPubKey(), kvstore.ProtocolVersion)
+	pubKey := privVal.GetPubKey()
+	require.NoError(t, err)
```
melekes: what error?

consensus/state.go
```diff
@@ -1675,7 +1678,8 @@ func (cs *ConsensusState) voteTime() time.Time {
 // sign the vote and publish on internalMsgQueue
 func (cs *ConsensusState) signAddVote(type_ types.SignedMsgType, hash []byte, header types.PartSetHeader) *types.Vote {
 	// if we don't have a key or we're not in the validator set, do nothing
-	if cs.privValidator == nil || !cs.Validators.HasAddress(cs.privValidator.GetAddress()) {
+	privValAddr := cs.privValidator.GetPubKey().Address()
```
melekes: if `cs.privValidator` can be nil (see the check later), this will panic!

privval/remote_signer.go
```diff
@@ -15,8 +17,9 @@ import (
 // RemoteSignerClient implements PrivValidator, it uses a socket to request signatures
 // from an external process.
 type RemoteSignerClient struct {
-	conn net.Conn
-	lock sync.Mutex
+	conn            net.Conn
+	consensusPubKey crypto.PubKey
+	lock            sync.Mutex
```
ebuchman: `mtx` for consistency

privval/remote_signer.go
ebuchman: this is fine. we're prepping a breaking release now

privval/remote_signer.go
ebuchman: do we need it anyways? removed it from the FilePV in https://github.com/tendermint/tendermint/pull/2870 - afaik, it's not actually necessary since we don't access it concurrently

privval/tcp_test.go
```diff
@@ -47,12 +42,10 @@ func TestSocketPVPubKey(t *testing.T) {
 	clientKey, err := sc.getPubKey()
 	require.NoError(t, err)
 
-	privKey := rs.privVal.GetPubKey()
-
-	assert.Equal(t, privKey, clientKey)
+	privvalPubKey := rs.privVal.GetPubKey()
+	require.NoError(t, err)
```
ebuchman: Can remove

privval/tcp_test.go
```diff
@@ -153,9 +146,9 @@ func TestSocketPVDeadline(t *testing.T) {
 	go func(sc *TCPVal) {
 		defer close(listenc)
 
-		require.NoError(t, sc.Start())
+		assert.Equal(t, sc.Start().(cmn.Error).Data(), ErrConnTimeout)
```
ebuchman: Why is this supposed to error now?

Liamsi: Because `OnStart` of `TCPVal` and `ICPVal` will call `NewRemoteSignerClient` which will request the public key:
https://github.com/tendermint/tendermint/blob/f9a564cfc37f7a7f6c2d2a19de72a73a71523ac5/privval/remote_signer.go#L29-L35

Liamsi: This will time out (err) when calling Start(). Previously this error was supposed to happen after start: https://github.com/tendermint/tendermint/pull/2948/files#diff-f22985f984c7cb5c7d4f0702ea6325beL179

Liamsi:
> This looks good, but I'm not sure removing the GetAddress is really worth it anymore. It's true that it's redundant, but it's a great convenience function since without it we have to add all these extra lines of code to get the address variable. 

Actually no add any extra lines of code were added (but some code got deleted, namely all GetAddress() methods). Instead of `pv.GetAddress()` you call `pv.GetPubKey().Address()`. But you are right the main justification (network request) is obsolete now. I still think having a method in an interface (`PrivValidator.GetPubKey()`) which is basically a wrapper around another interface method `PrivKey.PubKey()` a slightly weird abstraction. I can undo the change (for the sake of making the diff smaller) and we can discuss about removing the GetAddress method in #2981.

privval/remote_signer.go
Liamsi: The justification seems to be this: https://github.com/tendermint/tendermint/pull/2568#discussion_r223306805
and AFAIU this can't happen; the sign method would never be called with the pub key package as the amino prefix identifies the right message types. Removed.

Liamsi: Hmm, maybe I'm wrong and the mtx is still necessary: https://circleci.com/gh/tendermint/tendermint/39026?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
The prefix there (A6160A51) is not any of the registered messages (RemoteSignerMsg).

Maybe this is a new variant of https://github.com/tendermint/tendermint/issues/2677#issuecomment-433938289 which doesn't deadlock & timeout anymore.

melekes:
> I still think having a method in an interface (PrivValidator.GetPubKey()) which is basically a wrapper around another interface method PrivKey.PubKey() a slightly weird abstraction (unless we have implementations planned where the address can't be derived from the pub key).

Another argument may be to make the interface as small as possible https://go-talks.appspot.com/github.com/ChrisHines/talks/structured-logging/structured-logging.slide#21

Liamsi:
Yeah, that's another good argument. Either way, we could cherry-pick this commit to revert all changes related to removing `GetAddress` in case we do not want this in this PR: https://github.com/tendermint/tendermint/commit/5d6f468b2a144d2fdef569db8af806c9668ada80

Currently investigating what is going on with the failing test (probably related to removing the mutex).

ebuchman:
Ok removing the GetAddress seems fine. Smaller interfaces are better :)

privval/remote_signer.go
Liamsi: It actually looks like with the mutex in place, we sometimes deadlock (in tests, haven't seen this in happening in the wild and I can't reproduce it locally) and without it we seem to get messages out of order or mixed up as Hendrik described in above comment: (https://github.com/tendermint/tendermint/pull/2568#discussion_r223306805)
Keeping the mutex seems more robust but I think we further need to investigate what is wrong and we probably do not need it (and need to update the tests instead).

Liamsi:
Cool! Merged in the changes from #2870 as requested.

cmd/tendermint/commands/testnet.go
```diff
@@ -134,32 +134,14 @@ func testnetFiles(cmd *cobra.Command, args []string) error {
...
 	}
 
-	// Gather persistent peer addresses.
-	var (
-		persistentPeers string
-		err             error
-	)
 	if populatePersistentPeers {
-		persistentPeers, err = persistentPeersString(config)
+		err := populatePersistentPeersInConfigAndWriteIt(config)
```
ebuchman: why do we only write the config file if populatePersistentPeers?

ebuchman: ah - I think this accidentally undid the fix from https://github.com/tendermint/tendermint/pull/2983/files

PR #2950, opened 2018-11-29
Bucky/v0.27.0

ebuchman:
* [ ] Wrote tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,69 @@
 # Changelog
 
+## v0.27.0
+
+*November 29th, 2018*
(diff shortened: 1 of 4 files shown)


melekes:
I think we should also get https://github.com/tendermint/tendermint/pull/2870 in (it's breaking too).

ebuchman:
> I think we should also get #2870 in (it's breaking too).

Doesn't need to be since we can provide a simple upgrade path, but will consider it tmrw.

ebuchman:
Will merge this and continue updating if we merge a few more PRs

PR #2964, opened 2018-12-04
Add missing endpoints to HTTPClient and LocalClient

danil-lashin:
* [x] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -19,6 +19,7 @@ Special thanks to external contributors on this release:
 ### FEATURES:
 
 ### IMPROVEMENTS:
+- [rpc] Add `/unconfirmed_txs` and `/num_unconfirmed_txs` endpoints to HTTPClient and LocalClient
 
 ### BUG FIXES:
 - [rpc] \#2408 `/broadcast_tx_commit`: Fix "interface conversion: interface {} in nil, not EventDataTx" panic (could happen if somebody sent a tx using /broadcast_tx_commit while Tendermint was being stopped)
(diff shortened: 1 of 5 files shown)


danil-lashin:
Also, we need to update the `Client` interface. Where should we put this methods? May be introduce new `MempoolClient`?

melekes:
> Also, we need to update the Client interface. Where should we put this methods? May be introduce new MempoolClient?

`MempoolClient` sounds OK

Can we also add tests for these methods? Thanks 👍

melekes:
```
--- FAIL: TestTx (0.06s)
	rpc_test.go:344: client 0, case 0
	Error Trace:	rpc_test.go:356
	Error:      	Should be zero, but was 1
	Test:       	TestTx
	rpc_test.go:344: client 0, case 1
	Error Trace:	rpc_test.go:356
	Error:      	Should be zero, but was 1
	Test:       	TestTx
	rpc_test.go:344: client 0, case 2
	rpc_test.go:344: client 0, case 3
	rpc_test.go:344: client 0, case 4
	rpc_test.go:344: client 0, case 5
	rpc_test.go:344: client 1, case 0
	Error Trace:	rpc_test.go:356
	Error:      	Should be zero, but was 1
	Test:       	TestTx
	rpc_test.go:344: client 1, case 1
	Error Trace:	rpc_test.go:356
	Error:      	Should be zero, but was 1
	Test:       	TestTx
	rpc_test.go:344: client 1, case 2
	rpc_test.go:344: client 1, case 3
	rpc_test.go:344: client 1, case 4
	rpc_test.go:344: client 1, case 5
FAIL
```

not sure why TestTx is failing. seems not relevant to the changes made in this PR

danil-lashin:
@melekes, this is because of new tests. Seems like TestTx expects empty mempool, but previous test leaves it with 1 tx. I will try to fix that.

Issue #2960, opened 2018-12-02
Problems related to ProposerPriority


Slamper:
The new ProposerPriority mechanism caused problems in the latest Gaia-9002 network.

Since the default priority is -totalPrio * C I assume that in the block in which a validator is added the shifting effectively adds `AvgProposerPriority` (subtraction of negative prio) to everyone. So the ones bonding earlier will experience more of these "bumps"

We have some strange developments of proposer_priority.

![image](https://user-images.githubusercontent.com/1405758/49338613-b4a80a80-f623-11e8-83cd-1b2e73f7724b.png)

Table of nuevax priority data: https://docs.google.com/spreadsheets/d/1_GC_HRO4iDTLDfH7OBkPw83p5oFHFA_BQL68RoIBOVw/edit?usp=sharing

The small spikes happen every 10 blocks.
Every time the sudden spike up happened a new validator was added to the active set.
The delta of the sudden increases is `AvgProposerPriority` due to the shifting.

When then the genesis validators unbonded, some validators were left with close to MaxInt64 Priorities which in this case lead to one validator proposing the following 200 blocks after the unbonding of the high-weight validators.

That validator proposing the following 200 blocks was the one, which was the first to bond after the genesis validators. He had the highest proposer_priority because he experienced the most validator added events (n) and subsequently the most sudden spikes as seen in the graphs. 

After these 200 blocks the only ones to propose were the said validator and the validator that bonded after him as that one experienced (n-1) spikes.

Rel: #2942

cc: @ebuchman @Liamsi @melekes @ValarDragon

cwgoes:
Note: graph is of `ProposerPriority` of [nuevax](https://hubble.figment.network/chains/gaia-9002/validators/9D93AD02FE1BD7E76B9E58D7EB461088EAD011E5).

cwgoes:
Another interesting edge case caused by a very large validator suddenly unbonding: validators which had recently proposed were reset to `-total_power` accum (large and negative), while a validator who bonded just after was set to the new `-total_power` accum (small and negative), so ended up proposing hundreds of blocks in a row while the other validators' accums "caught up".

cwgoes:
Haven't though through too much, but what about, when a validator unbonds (or also decreases in power, I suppose), just increasing everyone's accum by the power difference? That seems like it would solve the imbalance between previously-bonded validators and ones which bond just after the large one unbonds.

ebuchman:
This should be fixed by #3049

PR #2966, opened 2018-12-04
More ProposerPriority tests 

Liamsi:
* [ ] Wrote tests

### DIFF
--- a/state/state_test.go
+++ b/state/state_test.go
@@ -331,9 +331,6 @@ func TestProposerPriorityProposerAlternates(t *testing.T) {
 	state.Validators = types.NewValidatorSet([]*types.Validator{val1})
 	state.NextValidators = state.Validators
 
-	// NewValidatorSet calls IncrementProposerPriority but uses on a copy of val1
-	assert.EqualValues(t, 0, val1.ProposerPriority)
-
 	block := makeBlock(state, state.LastBlockHeight+1)
 	blockID := types.BlockID{block.Hash(), block.MakePartSet(testPartSize).Header()}
 	// no updates:
@@ -342,9 +339,10 @@ func TestProposerPriorityProposerAlternates(t *testing.T) {
 	}
 	validatorUpdates, err := types.PB2TM.ValidatorUpdates(abciResponses.EndBlock.ValidatorUpdates)
 	require.NoError(t, err)
-	updatedState, err := updateState(state, blockID, &block.Header, abciResponses, validatorUpdates)
 
+	updatedState, err := updateState(state, blockID, &block.Header, abciResponses, validatorUpdates)
 	assert.NoError(t, err)
+
 	// 0 + 10 (initial prio) - 10 (avg) - 10 (mostest - total) = -10
 	assert.Equal(t, -origVotinPower, updatedState.NextValidators.Validators[0].ProposerPriority)
 
@@ -353,6 +351,7 @@ func TestProposerPriorityProposerAlternates(t *testing.T) {
 	updateAddVal := abci.ValidatorUpdate{PubKey: types.TM2PB.PubKey(val2PubKey), Power: 10}
 	validatorUpdates, err = types.PB2TM.ValidatorUpdates([]abci.ValidatorUpdate{updateAddVal})
 	assert.NoError(t, err)
+
 	updatedState2, err := updateState(updatedState, blockID, &block.Header, abciResponses, validatorUpdates)
 	assert.NoError(t, err)
 
@@ -374,6 +373,7 @@ func TestProposerPriorityProposerAlternates(t *testing.T) {
 
 	validatorUpdates, err = types.PB2TM.ValidatorUpdates(abciResponses.EndBlock.ValidatorUpdates)
 	require.NoError(t, err)
+
 	updatedState3, err := updateState(updatedState2, blockID, &block.Header, abciResponses, validatorUpdates)
 	assert.NoError(t, err)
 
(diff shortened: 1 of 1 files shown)


state/state_test.go
```diff
@@ -263,6 +263,294 @@ func TestOneValidatorChangesSaveLoad(t *testing.T) {
...
+	validatorUpdates, err = types.PB2TM.ValidatorUpdates([]abci.ValidatorUpdate{updateAddVal})
+	assert.NoError(t, err)
+	updatedState2, err := updateState(updatedState, blockID, &block.Header, abciResponses, validatorUpdates)
+	assert.NoError(t, err)
+
+	require.Equal(t, len(updatedState2.NextValidators.Validators), 2)
+	_, addedVal2 := updatedState2.NextValidators.GetByAddress(val2PubKey.Address())
+	// adding a validator should not lead to a ProposerPriority equal to zero (unless the combination of averaging and
+	// incrementing would cause so; which is not the case here)
+	assert.NotZero(t, addedVal2.ProposerPriority, "Added validator's ProposerPriority should not be zero")
```
cwgoes: Shouldn't we know what it would be exactly - can we just assert that?

state/state_test.go
```diff
@@ -263,6 +263,294 @@ func TestOneValidatorChangesSaveLoad(t *testing.T) {
...
+	// Updating a validator does not reset the ProposerPriority to zero:
+	updateVal := abci.ValidatorUpdate{PubKey: types.TM2PB.PubKey(val2PubKey), Power: 1}
+	validatorUpdates, err = types.PB2TM.ValidatorUpdates([]abci.ValidatorUpdate{updateVal})
+	assert.NoError(t, err)
+	updatedState3, err := updateState(updatedState2, blockID, &block.Header, abciResponses, validatorUpdates)
+	assert.NoError(t, err)
+
+	require.Equal(t, len(updatedState3.NextValidators.Validators), 2)
+	_, updatedVal2 := updatedState3.NextValidators.GetByAddress(val2PubKey.Address())
+	assert.NotZero(t, updatedVal2.ProposerPriority)
```
cwgoes: Ditto, why not assert exact?

state/state_test.go
```diff
@@ -263,6 +263,294 @@ func TestOneValidatorChangesSaveLoad(t *testing.T) {
...
+
+	// val1 will still be proposer as val2 just got added:
+	assert.Equal(t, updatedState2.Validators.Proposer.Address, updatedState2.NextValidators.Proposer.Address)
+	assert.Equal(t, updatedState2.Validators.Proposer.Address, val1PubKey.Address())
+	assert.Equal(t, updatedState2.NextValidators.Proposer.Address, val1PubKey.Address())
+
+	_, updatedVal1 := updatedState2.NextValidators.GetByAddress(val1PubKey.Address())
+	_, updatedVal2 := updatedState2.NextValidators.GetByAddress(val2PubKey.Address())
+	// val1's ProposerPriority story: -10 (see above) + 10 (voting pow) - (-1) (avg) - 20 (total) == -19
+	assert.EqualValues(t, -19, updatedVal1.ProposerPriority)
```
cwgoes: Can we calculate `-19` in terms of the initial priorities (inline)?

state/state_test.go
```diff
@@ -263,6 +263,294 @@ func TestOneValidatorChangesSaveLoad(t *testing.T) {
...
+	// val2 prio when added: -(totalVotingPower + (totalVotingPower >> 3)) == -11
+	// -> -11 + 10  (voting power) - (-1) (avg) == 0
+	assert.EqualValues(t, 0, updatedVal2.ProposerPriority, "unexpected proposer priority for validator: %v", updatedVal2)
```
cwgoes: Can we calculate `0` in terms of the initial priorities (inline)?

state/state_test.go
```diff
@@ -263,6 +263,294 @@ func TestOneValidatorChangesSaveLoad(t *testing.T) {
...
+
+	block := makeBlock(state, state.LastBlockHeight+1)
+	blockID := types.BlockID{block.Hash(), block.MakePartSet(testPartSize).Header()}
+	// no updates:
+	abciResponses := &ABCIResponses{
+		EndBlock: &abci.ResponseEndBlock{ValidatorUpdates: nil},
+	}
+	validatorUpdates, err := types.PB2TM.ValidatorUpdates(abciResponses.EndBlock.ValidatorUpdates)
+	require.NoError(t, err)
+
```
cwgoes: Can we also assert which validator is the proposer (in all of these tests)?

state/state_test.go
```diff
@@ -263,6 +263,294 @@ func TestOneValidatorChangesSaveLoad(t *testing.T) {
...
+	assert.NotEqual(t, updatedState3.Validators.Proposer.Address, updatedState3.NextValidators.Proposer.Address)
+
+	assert.Equal(t, updatedState3.Validators, updatedState2.NextValidators)
+	_, updatedVal1 = updatedState3.NextValidators.GetByAddress(val1PubKey.Address())
+	_, updatedVal2 = updatedState3.NextValidators.GetByAddress(val2PubKey.Address())
+
+	// check if expected proposer prio is matched:
+
+	// -19 + 10 - 0 (avg) == -9
+	assert.EqualValues(t, -9, updatedVal1.ProposerPriority, "unexpected proposer priority for validator: %v", updatedVal2)
```
cwgoes: Ditto on above (clculate inline)

state/state_test.go
```diff
@@ -263,6 +263,294 @@ func TestOneValidatorChangesSaveLoad(t *testing.T) {
...
+		// no validator updates:
+		abciResponses := &ABCIResponses{
+			EndBlock: &abci.ResponseEndBlock{ValidatorUpdates: nil},
+		}
+		validatorUpdates, err = types.PB2TM.ValidatorUpdates(abciResponses.EndBlock.ValidatorUpdates)
+		require.NoError(t, err)
+
+		updatedState, err := updateState(oldState, blockID, &block.Header, abciResponses, validatorUpdates)
+		assert.NoError(t, err)
+		// alternate:
```
cwgoes: cool! can we also check that the powers change cyclically (as they ought to)?

state/state_test.go
```diff
@@ -263,6 +263,294 @@ func TestOneValidatorChangesSaveLoad(t *testing.T) {
...
+		block = makeBlock(curState, curState.LastBlockHeight+1)
+		blockID = types.BlockID{block.Hash(), block.MakePartSet(testPartSize).Header()}
+		curState, err = updateState(curState, blockID, &block.Header, abciResponses, validatorUpdates)
+		if !bytes.Equal(curState.Validators.Proposer.Address, curState.NextValidators.Proposer.Address) {
+			isProposerUnchanged = false
+		}
+		count++
+	}
+	// first proposer change happens after this many iters; we probably want to lower this number:
+	// TODO: change with https://github.com/tendermint/tendermint/issues/2960
```
cwgoes: cool test, ideally we could calculate what this value ought to be in terms of the power ratio, and then of course try to fix it

Liamsi: I think, I've addressed all your comments besides this one. I'll use this to see if we can change the behaviour instead. Please take another look if you like.

cwgoes:
The testcase additions look good to me.

After reading and reasoning through the proposer election mechanism in detail with @Liamsi, I think it would greatly benefit in clarity (and in lowering the likelihood of mistakes in implementation) from rearchitecting for clarity. A lot of the code does not make too much sense - e.g. we create a heap each round to only calculate a maximum, which is much slower than just iterating through the validators once, pass around by reference or by copy in inconsistent ways, shift by the average more often than necessary (still, I think, even after the once-every-n - I think we only need to shift by the average once).

Liamsi:
> we create a heap each round to only calculate a maximum, which is much slower than just iterating through the validators once,

I'm removing the heap in the follow up PR for #2960. 

Thanks a lot @cwgoes!!!

Issue #2823, opened 2018-11-12
p2p: acceptRoutine may exit and never restart without stopping the node


ebuchman:
The switch runs a go routine called `acceptRoutine` responsible for accepting new peer connections.

This routine exits when the transport is closed, but also if we receive an error on Accept. In this case, the node still runs normally, but it will no longer accept incoming connections. It does log an error, but otherwise there is no indication that the node no longer accepts connections, nor is there a way to get it to start again, other than by restarting it.

An example causing this error looks like:

```
E[10116-11-10|14:37:02.305] Accept on transport errored                  module=p2p err="accept tcp [::]:26656: accept4: too many open files" numPeers=50
```

See https://github.com/cosmos/cosmos-sdk/issues/2787 for an issue about too many open files - we should probably have a corresponding issue in this repo.

Trying to connect to a peer where the acceptRoutine has exited looks like:

```
I[12116-11-12|14:19:45.468] Error reconnecting to peer. Trying again     module=p2p tries=17 err="auth failure: secrect conn failed: read tcp 192.168.0.92:64943->13.56.78.160:26656: i/o timeout" addr=6a04069a2736f096404df2bbc31d7c2807c080c0@13.56.78.160:26656
```

We should fix this by either:

- killing the node if the acceptRoutine exits
- having a retry loop around the acceptRoutine to restart it some number of times before killing the node

ebuchman:
Closing for now. We may want to revisit the error handling for this at a later date

PR #2968, opened 2018-12-04
p2p: panic on transport error

ebuchman:
Addresses #2823. Currently, the acceptRoutine exits if the transport returns
an error trying to accept a new connection. Once this happens, the node
can't accept any new connections. So here, we panic instead. While we
could potentially be more intelligent by rerunning the acceptRoutine, the
error may indicate something more fundamental (eg. file desriptor limit)
that requires a restart anyways. We can leave it to process managers to
handle that restart, and notify operators about the panic.

* [ ] Wrote tests

### DIFF
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -505,6 +505,12 @@ func (sw *Switch) acceptRoutine() {
 					"err", err,
 					"numPeers", sw.peers.Size(),
 				)
+				// We could instead have a retry loop around the acceptRoutine,
+				// but that would need to stop and let the node shutdown eventually.
+				// So might as well panic and let process managers restart the node.
+				// There's no point in letting the node run without the acceptRoutine,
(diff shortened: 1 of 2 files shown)


p2p/switch.go
```diff
@@ -505,6 +505,12 @@ func (sw *Switch) acceptRoutine() {
 					"err", err,
 					"numPeers", sw.peers.Size(),
 				)
+				// We could instead have a retry loop around the acceptRoutine,
+				// but that would need to stop and let the node shutdown eventually.
+				// So might as well panic and let process managers restart the node.
+				// There's no point in letting the node run without the acceptRoutine,
+				// since it won't be able to accept new connections.
+				panic(fmt.Errorf("accept routine exited: %v", err))
```
melekes: this should not happen normally, right?

PR #2983, opened 2018-12-07
tendermint testnet: always turn off strict routability

melekes:
previously, we're turning it off only if `--populate-persistent-peers`
flag was used, which is obviously incorrect.

Fixes https://github.com/cosmos/cosmos-sdk/issues/2983

* [ ] Wrote tests


yaram:
It might be a good idea to modify the documentation to state that, on Windows, the .tendermint directory is located in the windows user directory. Perhaps using the USERPROFILE environment variable to indicate that?

PR #2984, opened 2018-12-07
Use os/user to retrieve home directory instead of $HOME

yaram:
This change fixes an issue where, on Windows, the .tendermint directory is put in C:\ instead of the user's home directory due to the HOME environment variable not being set on Windows.

* [ ] Wrote tests

### DIFF
--- a/cmd/tendermint/main.go
+++ b/cmd/tendermint/main.go
@@ -1,7 +1,7 @@
 package main
 
 import (
-	"os"
+	"os/user"
 	"path/filepath"
 
 	"github.com/tendermint/tendermint/libs/cli"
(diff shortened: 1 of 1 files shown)


cmd/tendermint/main.go
```diff
@@ -41,7 +41,13 @@ func main() {
...
 	rootCmd.AddCommand(cmd.NewRunNodeCmd(nodeFunc))
 
-	cmd := cli.PrepareBaseCmd(rootCmd, "TM", os.ExpandEnv(filepath.Join("$HOME", cfg.DefaultTendermintDir)))
+	// Retrieve user's home directory
+	usr, err := user.Current()
+	if err != nil {
+		panic(err)
+	}
+
+	cmd := cli.PrepareBaseCmd(rootCmd, "TM", filepath.Join(usr.HomeDir, cfg.DefaultTendermintDir))
```
HaoyangLiu: Have you verified this change on linux?

yaram:
It appears that os/user still relies on cgo and thus cannot be used for cross-compiling. An alternative is to use the go-homedir package. I could also write the logic manually, perhaps using the USERPROFILE environment variable on Windows and the HOME environment variable on unix-likes.

melekes:
Is this Windows machine your personal laptop? Or you want to deploy Tendermint network onto Windows cluster? Why not use Linux?

PR #2988, opened 2018-12-08
fix docs / proxy app

### LINKED ISSUE #2986: sync docs & tmint for nil/counter apps
Hello! 

I tried to follow the instruction from [this page](https://tendermint.com/docs/tendermint-core/using-tendermint.html#run) by using other in-process apps such as 'nil' or 'counter' but it shows errors as follows

image

Please help advice. Thank you.

zramsay:
closes #2986 

* [ ] Wrote tests

### DIFF
--- a/abci/cmd/abci-cli/abci-cli.go
+++ b/abci/cmd/abci-cli/abci-cli.go
@@ -58,7 +58,7 @@ var RootCmd = &cobra.Command{
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 
 		switch cmd.Use {
-		case "counter", "kvstore", "dummy": // for the examples apps, don't pre-run
+		case "counter", "kvstore": // for the examples apps, don't pre-run
 			return nil
 		case "version": // skip running for version command
 			return nil
@@ -127,10 +127,6 @@ func addCounterFlags() {
 	counterCmd.PersistentFlags().BoolVarP(&flagSerial, "serial", "", false, "enforce incrementing (serial) transactions")
 }
 
-func addDummyFlags() {
-	dummyCmd.PersistentFlags().StringVarP(&flagPersist, "persist", "", "", "directory to use for a database")
-}
-
 func addKVStoreFlags() {
 	kvstoreCmd.PersistentFlags().StringVarP(&flagPersist, "persist", "", "", "directory to use for a database")
 }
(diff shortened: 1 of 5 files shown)


proxy/client.go
```diff
@@ -64,13 +65,17 @@ func (r *remoteClientCreator) NewABCIClient() (abcicli.Client, error) {
...
-		return NewLocalClientCreator(kvstore.NewKVStoreApplication())
-	case "persistent_kvstore":
 		fallthrough
+	case "kvstore":
+		return NewLocalClientCreator(kvstore.NewKVStoreApplication())
 	case "persistent_dummy":
+		fallthrough
+	case "persistent_kvstore":
 		return NewLocalClientCreator(kvstore.NewPersistentKVStoreApplication(dbDir))
 	case "nilapp":
```
melekes: I don't like the name to be honest. Maybe rename to `empty` (or `noop`)

proxy/client.go
```diff
@@ -64,13 +65,17 @@ func (r *remoteClientCreator) NewABCIClient() (abcicli.Client, error) {
 
 func DefaultClientCreator(addr, transport, dbDir string) ClientCreator {
 	switch addr {
-	case "kvstore":
-		fallthrough
+	case "counter":
+		return NewLocalClientCreator(counter.NewCounterApplication(false))
+	case "counter_serial":
+		return NewLocalClientCreator(counter.NewCounterApplication(true))
 	case "dummy":
```
melekes: I think it's time to remove `dummy/persistent_dummy`

proxy/client.go
ebuchman: Great idea - let's call it `noop`. And we should deprecate `dummy` and `nillapp` names for v0.28

zramsay:
comments addressed

abci/cmd/abci-cli/abci-cli.go
```diff
@@ -58,7 +58,7 @@ var RootCmd = &cobra.Command{
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 
 		switch cmd.Use {
-		case "counter", "kvstore", "dummy": // for the examples apps, don't pre-run
+		case "counter", "kvstore": // for the examples apps, don't pre-run
```
melekes:
```suggestion
		case "counter", "kvstore", "noop": // for the examples apps, don't pre-run
```

PR #2991, opened 2018-12-10
mempool: notifyTxsAvailable if there're txs left, but recheck=false

### LINKED ISSUE #2961: Set mempool.recheck =  false and create empty block = false may cause some transactions to be stuck.
**Tendermint version** : 0.26.4-b771798d

**ABCI app** : kv-store

**Environment**:
- **OS** (e.g. from /etc/os-release): macOS 10.14.1

**What happened**:
Set `mempool.recheck =  false` and `create empty block = false` in `config.toml`.
When transactions get added right between a new empty block is being proposed and committed,
the proposer won't propose new block for that transactions immediately after.
That transactions are stuck in the mempool until a new transaction is added and trigger the proposer.

If there is a transaction left in the mempool, new block should be proposed immediately.

Fire two transaction using `broadcast_tx_sync` with specific delay between them.
(You may need to do it multiple time before the right delay is found, on my machine the delay is 0.98s)

**Logs (paste a small part showing an error (< 10 lines) or link a pastebin, gist, etc. containing more of the log file)**: https://pastebin.com/0Wt6uhPF

[mempool]
recheck = false
create_empty_block = false

In `mempool.go`, we found that proposer will immediately propose new block if 
1. Last committed block has some transaction (causing AppHash to changed) or
2. `mem.notifyTxsAvailable()` is called.

Our scenario is as followed. 
- A transaction is fired, it will create 1 block with 1 tx (line 1-4 in the log) and 1 empty block.
- After the empty block is proposed but before it is committed, second transaction is fired and added to mempool. (line 8-16)

Now, since the last committed block is empty and `mem.notifyTxsAvailable()` will be called only if `mempool.recheck = true`.
The proposer won't immediately propose new block, causing the second transaction to stuck in mempool until another transaction is added to mempool and trigger `mem.notifyTxsAvailable()`.

melekes:
(left after committing a block)

Fixes #2961

* [ ] Wrote tests


ValarDragon:
I'm confused structurally about this. (I havne't spent much time looking at our notification / tx broadcasting process)
AFAIU, we always set the txs to be un-notified. Then we either recheck or don't recheck. 

* When rechecking, we notify tx availability one by one. This means that we will re-notify / re-broadcast to peers, starting from the beginning, right? We don't actually want to reset our position under Update, instead we just want to stay at the same index in the linked list, unless it was removed. Then we want to use the next element of the linked list. (Which may be hard to get, so we'd use the first element as a "hack" until we get time to prioritize doing it properly)
* When not rechecking, we just turn on notify tx availability again, so we start notifying txs again from the beginning, not from where we left off. Like before, we want to start from where we left off?

Am I understanding this correctly?

melekes:
we notify txs available when:

- we get a new tx and it passes CheckTx
- after committing a block & rechecking all uncommitted txs https://github.com/tendermint/tendermint/blob/41eaf0e31d3e9462b165378d8870ee7366ed7b38/mempool/mempool.go#L427
- (**now**) after committing a block (when recheck is turned off)

ValarDragon:
Oh cool! Thanks for the clarification. I misunderstood the genenral availability vs specific tx availability part! Seems good to merge to me

melekes:
```
--- FAIL: TestNodeSetPrivValTCP (0.51s)
panic: EOF [recovered]
	panic: EOF

goroutine 132 [running]:
testing.tRunner.func1(0xc420b124b0)
panic(0xfd9320, 0xc4200f6030)
github.com/tendermint/tendermint/privval.(*RemoteSignerClient).GetAddress(0xc4204dc1a0, 0x1, 0xd59330, 0xc420a70320)
	/go/src/github.com/tendermint/tendermint/privval/remote_signer.go:39 +0xe5
github.com/tendermint/tendermint/node.NewNode(0xc420b27d40, 0x12f1540, 0xc420b24000, 0xc42097a010, 0x12e5720, 0xc420280f20, 0xc42097a0b0, 0x124ef58, 0xc42097a0c0, 0x12f1400, ...)
	/go/src/github.com/tendermint/tendermint/node/node.go:252 +0x5fca
github.com/tendermint/tendermint/node.DefaultNewNode(0xc420b27d40, 0x12f1400, 0x1814b38, 0xa50f997764672123, 0x8e86a54ce29e3ad6, 0xd11b75fbfbc36040)
	/go/src/github.com/tendermint/tendermint/node/node.go:89 +0x38b
github.com/tendermint/tendermint/node.TestNodeSetPrivValTCP(0xc420b124b0)
	/go/src/github.com/tendermint/tendermint/node/node_test.go:141 +0x5e6
testing.tRunner(0xc420b124b0, 0x124ef90)
created by testing.(*T).Run
FAIL	github.com/tendermint/tendermint/node	5.655s
```

this seems not related

CHANGELOG_PENDING.md
```diff
@@ -22,4 +22,5 @@ Special thanks to external contributors on this release:
 
 ### BUG FIXES:
 - [kv indexer] \#2912 don't ignore key when executing CONTAINS
+- [mempool] \#2961 notifyTxsAvailable if there're txs left after committing a block, but recheck=false
```
ebuchman: I don't think I've ever seen `there're` before :P .

PR #2996, opened 2018-12-10
Fix overflow issue when summing mempool gas wanteds

ValarDragon:
* [ ] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -22,3 +22,5 @@ Special thanks to external contributors on this release:
 
 ### BUG FIXES:
 - [kv indexer] \#2912 don't ignore key when executing CONTAINS
+- [mempool] \#2994 Don't allow txs with negative gas wanteds
+- [mempool] \#2995 Don't allow gas wanted to overflow when reaping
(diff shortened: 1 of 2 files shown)


mempool/mempool.go
```diff
@@ -490,11 +490,14 @@ func (mem *Mempool) ReapMaxBytesMaxGas(maxBytes, maxGas int64) types.Txs {
 			return txs
 		}
 		totalBytes += int64(len(memTx.tx)) + aminoOverhead
-		// Check total gas requirement
-		if maxGas > -1 && totalGas+memTx.gasWanted > maxGas {
+		// Check total gas requirement. We are guaranteed from the
+		// postcheck filter that memTx.gasWanted is non-negative,
+		// it suffices to ensure that the sum is non-negative.
+		newTotalGas := totalGas + memTx.gasWanted
+		if maxGas > -1 && newTotalGas > maxGas && newTotalGas > -1 {
```
ebuchman: isnt' this redundant because we already check `maxGas > -1` ?

CHANGELOG_PENDING.md
```diff
@@ -22,4 +22,5 @@ Special thanks to external contributors on this release:
 
 ### BUG FIXES:
 - [kv indexer] \#2912 don't ignore key when executing CONTAINS
+- [mempool] \#2994 Don't allow txs with negative gas wanted
```
melekes: All this PR does now is adding the comment. Should we remove the changelog entry then?

```suggestion
```

ValarDragon: That was from a previous pr, forgot to add that entry

PR #2997, opened 2018-12-10
Make testing logger that doesn't write to stdout

ValarDragon:
I'm making this PR for @jackzampolin, since he doesn't have TM branching perms.

* [ ] Wrote tests

### DIFF
--- a/libs/log/testing_logger.go
+++ b/libs/log/testing_logger.go
@@ -1,6 +1,7 @@
 package log
 
 import (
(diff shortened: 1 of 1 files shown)


jackzampolin:
This PR adds a logger that you can pass an arbitrary `io.Writer`. This will allow easy writing to files.

jackzampolin:
Good change @ebuchman, done!

PR #3004, opened 2018-12-11
full-text docs search

zramsay:
Woot Woot https://github.com/tendermint/devops/issues/117

it's only a search API so it's not a secret (approved by Jessy). Fixed the "Last Updated" feature which uses git and refers to each individual file

* [ ] Wrote tests

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -8,7 +8,12 @@ module.exports = {
     lineNumbers: true
   },
   themeConfig: {
-    lastUpdated: "Last Updated",
+    lastUpdated: true,
+    algolia: {
+      apiKey: '59f0e2deb984aa9cdf2b3a5fd24ac501',
+      indexName: 'tendermint',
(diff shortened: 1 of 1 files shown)


docs/.vuepress/config.js
```diff
@@ -8,7 +8,12 @@ module.exports = {
     lineNumbers: true
   },
   themeConfig: {
-    lastUpdated: "Last Updated",
+    lastUpdated: true,
+    algolia: {
+      apiKey: '59f0e2deb984aa9cdf2b3a5fd24ac501',
```
melekes: this is search-only API key (https://www.algolia.com/doc/guides/security/api-keys/?language=php#search-only-api-key), right? just want to make sure

zramsay: yes

zramsay:
linking https://github.com/algolia/docsearch-configs/pull/668 which we can make additional PRs to update the config

PR #3006, opened 2018-12-12
During replay, when appHeight==0, only saveState if stateHeight is also 0

james-ray:
* [ ] Wrote tests


james-ray:
This pr has some related to #3007

consensus/replay.go
```diff
@@ -247,6 +247,7 @@ func (h *Handshaker) Handshake(proxyApp proxy.AppConns) error {
 
 	// Set AppVersion on the state.
 	h.initialState.Version.Consensus.App = version.Protocol(res.AppVersion)
+	sm.SaveState(h.stateDB, h.initialState)
```
melekes: Why do we need to save initialState here?

james-ray: in replayBlock, only mutableState will ApplyBlock (the final block), mutable state means the stateHeight not equal to storeHeight. We cannot ensure this.
So either we add code in if mutableState 's else, or just saveState here, otherwise this set appVersion will not appear in state, or in some nodes it may appear but some nodes not, because we cannot make sure every node will apply the final block or not.

melekes: Do we need to store appVersion? if we can get it from the application itself `version.Protocol(res.AppVersion)` every time

james-ray: if mutableState, in the ApplyBlock of the final block, it will SaveState. if not mutable state, we do nothing. So the state is saved or not depends on mutableState. If the app return the appVersion is same as the appVersion in current state, it's no problem. Otherwise if we do some upgrade or someting, the app will return a higher appVersion, thus some nodes will save the state and some not ( We cannot control precisely all nodes stopped with mutableState or not)

james-ray: If some nodes with mutableState ApplyBlock the final block, they saveState the state which appVersion is changed, other nodes with no mutableState will not saveState, thus their state's version remain unchanged. The chain forks.

ebuchman:
Also why are there so many commits?

Issue #2716, opened 2018-10-27
p2p: trying to reconnect despite already being connected


ebuchman:
In DialPeersAsync, we check if we're already dialing the peer, but not if we're already connected to them. Then we sleep a random amount before actually dialing them. 

Since the PEX doesn't know about persistent peers, it's possible for the PEX to be dialing or even to have already connected to a peer during this time. Then the DialPeersAsync will try to dial, and when it fails, it will keep trying to reconnect, even if we're already connected!

I think this might also explain https://github.com/tendermint/tendermint/issues/2332

I think we want to centralize dialing into a single location so that these kind of races aren't possible.

ebuchman:
Note we tried to address this once before but looks like it was insufficient: https://github.com/tendermint/tendermint/pull/2327

melekes:
I moved IsDialingOrExistingAddress check into DialPeerWithAddress [here](https://github.com/tendermint/tendermint/pull/3011)

melekes:
Merged https://github.com/tendermint/tendermint/pull/3011 to develop.

ebuchman:
Are we sure that a race still isn't theoretically possible? 

And shouldn't we still do:

> I think we want to centralize dialing into a single location so that these kind of races aren't possible.

melekes:
> Are we sure that a race still isn't theoretically possible?

The chance is low, but still possible I guess.

ebuchman:
Ok opened https://github.com/tendermint/tendermint/issues/3540

PR #3011, opened 2018-12-13
Seed mode refactoring

melekes:
- ListOfKnownAddresses is removed
- panic if addrbook size is less than zero
- CrawlPeers does not attempt to connect to existing or peers we're currently dialing
- various perf. fixes
- improved tests (though not complete)
- move IsDialingOrExistingAddress check into DialPeerWithAddress (Fixes https://github.com/tendermint/tendermint/issues/2716)

* [x] Wrote tests


p2p/pex/file.go
```diff
@@ -16,16 +16,18 @@ type addrBookJSON struct {
...
-
 	a.mtx.Lock()
 	defer a.mtx.Unlock()
-	// Compile Addrs
-	addrs := []*knownAddress{}
+
+	bookSize := a.size()
+	a.Logger.Info("Saving AddrBook to file", "size", bookSize)
+
+	addrs := make([]*knownAddress, bookSize)
```
ebuchman: Do we know if using `i` here is significantly better than using `make([]*knownAddress, 0, bookSize)` and `append`?

melekes: It's faster

melekes:
> Any ideas here for the cause of the deadlock?

No. I just wasn't satisfied with the existing code.

melekes:
> I don't think we should panic on addrbook issues, because there might be some bugs, but we shouldn't panic the caller due to it. Lets continue to log error messages and try to do sensible things?

We panic in many places. Are you saying we should get rid of all panics completely?

> The getPeersToCrawl() separation was good, why are we removing it? Can you please reintroduce?

I am not removing it. I am replacing it with a call to a standard addrbook#GetSelection method (instead of a hacky GetAll

screenshot 2018-12-17 at 09 28 35

p2p/pex/file.go
jaekwon: The previous is more obviously correct.  With custom counters I need to think harder to know that it's correct.  For one, I need to look back up at the implementation of .size() in order to know that there is sufficient capacity in addrs, otherwise it will panic.  And, as this code mutates, it's easy to add a continue statement in the wrong place (e.g. before the i++) (this is what I mean by increased surface area for bugs).

The previous code was more clean and beautiful, and more obviously panic-free.

There's another way to make it more performant, while retaining the obvious correctness and simplicity/less-surface-area.

```
        addrs := make([]*knownAddress, 0, bookSize)
	for _, ka := range a.addrLookup {
		addrs = append(addrs, ka)
	}
```

jaekwon:
RE: removing panics, no I don't mean we should remove panics everywhere.  I was just referring to the addrbook, where issues seem less critical.  But this is debatable.  OK.

p2p/pex/file.go
melekes: > The previous code was more clean and beautiful, and more obviously panic-free.

Agree

> I also need to look back up at the implementation of .size() in order to know that there is sufficient capacity in addrs, otherwise it will panic.

One might argue that if it panics, then our software is incorrect and it's easy to bring this fact to our attention. Otherwise, software will be running, but it may be incorrect and nobody will notice it (until somebody opens an issue "Why is my addrbook contains only 50% of addresses").

jaekwon: Sure, but it would still be better to avoid the usage of .size() altogether and just use len(addrLookup) instead.  I'm not advocating for covering bugs, but rather making code less brittle to changes.  By not relying on .size() when we're going to iterate over a thing that supports len() anyways, we also get to change the definition of size, etc.

melekes: Will do.

p2p/pex/pex_reactor.go
```diff
@@ -585,81 +591,70 @@ func (r *PEXReactor) crawlPeersRoutine() {
...
-		if now.Sub(pi.LastAttempt) < defaultCrawlPeerInterval {
+		if r.Switch.IsDialingOrExistingAddress(addr) {
 			continue
 		}
-		// Otherwise, attempt to connect with the known address
-		err := r.Switch.DialPeerWithAddress(pi.Addr, false)
+
+		// Record crawling attempt.
+		peerInfo.LastCrawled = now
+		// XXX: grows linearly with a number of peers crawled
```
melekes: I am a bit worried about this one

p2p/switch.go
```diff
@@ -358,10 +356,14 @@ func (sw *Switch) reconnectToPeer(addr *NetAddress) {
 		// sleep an exponentially increasing amount
 		sleepIntervalSeconds := math.Pow(reconnectBackOffBaseSeconds, float64(i))
 		sw.randomSleep(time.Duration(sleepIntervalSeconds) * time.Second)
+
 		err := sw.DialPeerWithAddress(addr, true)
```
melekes: note we did not have IsDialingOrExistingAddress check here

node/node.go
```diff
@@ -459,6 +459,7 @@ func NewNode(config *cfg.Config,
 			&pex.PEXReactorConfig{
 				Seeds:    splitAndTrimEmpty(config.P2P.Seeds, ",", " "),
 				SeedMode: config.P2P.SeedMode,
+				SeedDisconnectWaitPeriod: pex.DefaultSeedDisconnectWaitPeriod,
```
melekes: should probably be moved to config

p2p/pex/pex_reactor.go
melekes: also, we must persist `LastCrawled` info to avoid dialing peers too soon in case the node has crashed

node/node.go
Liamsi: Do you want to open an issue/add a todo? Or is this sth. that should be decided and done here (in this PR)?

melekes: In this PR

p2p/pex/pex_reactor.go
```diff
@@ -585,81 +592,65 @@ func (r *PEXReactor) crawlPeersRoutine() {
...
-func (of oldestFirst) Swap(i, j int)      { of[i], of[j] = of[j], of[i] }
-func (of oldestFirst) Less(i, j int) bool { return of[i].LastAttempt.Before(of[j].LastAttempt) }
-
-// getPeersToCrawl returns addresses of potential peers that we wish to validate.
-// NOTE: The status information is ordered as described above.
-func (r *PEXReactor) getPeersToCrawl() []crawlPeerInfo {
-	var of oldestFirst
+// CrawlPeers will crawl the network looking for new peer addresses.
+// Exposed for testing.
+func (r *PEXReactor) CrawlPeers(addrs []*p2p.NetAddress) {
```
Liamsi: Same. will this be used aóutside of the `pex` package?

p2p/pex/pex_reactor.go
```diff
@@ -585,81 +592,65 @@ func (r *PEXReactor) crawlPeersRoutine() {
...
 			r.RequestAddrs(peer)
 		}
 	}
 }
 
-// attemptDisconnects checks if we've been with each peer long enough to disconnect
-func (r *PEXReactor) attemptDisconnects() {
+// AttemptDisconnects checks if we've been with each peer long enough to disconnect
+// Exposed for testing.
+func (r *PEXReactor) AttemptDisconnects() {
```
Liamsi: Will this be useful (and used) outside of the `pex` package?

p2p/pex/pex_reactor.go
```diff
@@ -585,81 +592,65 @@ func (r *PEXReactor) crawlPeersRoutine() {
...
-	// TODO: be more selective
-	addrs := r.book.ListOfKnownAddresses()
 	for _, addr := range addrs {
-		if len(addr.ID()) == 0 {
-			continue // dont use peers without id
+		peerInfo, ok := r.crawlPeerInfos[addr.ID]
+		if !ok {
+			peerInfo = &crawlPeerInfo{
+				Addr:        addr,
+				LastCrawled: now.Add(-defaultCrawlPeerInterval).Add(-1 * time.Second),
```
Liamsi: This looks a bit random to me. Suggestion: Maybe add a comment why we subtract a second here (additionally to `defaultCrawlPeerInterval` which is 2 mins).

p2p/pex/pex_reactor.go
melekes: No, only in tests.

p2p/pex/pex_reactor.go
melekes: No, only in tests.

Liamsi: Does it make sense to un-export `AttemptDisconnects` and `CrawlPeers` then?

p2p/pex/pex_reactor.go
```diff
@@ -585,81 +592,65 @@ func (r *PEXReactor) crawlPeersRoutine() {
...
+			continue
+		}
 
-// crawlPeers will crawl the network looking for new peer addresses. (once)
-func (r *PEXReactor) crawlPeers() {
-	peerInfos := r.getPeersToCrawl()
+		// Record crawling attempt.
+		peerInfo.LastCrawled = now
+		// XXX: - grows linearly with a number of peers crawled
+		//      - does not get persisted
```
Liamsi: Sounds like this might eventually cause trouble. Maybe open an issue for this too? (I do not know how problematic this might become though)

node/node.go
Liamsi: Is there anything blocking us from adding it to the config? Sounds like a good idea to me.

ebuchman:
@melekes want to merge develop in and bring this up-to-date?

melekes:
> want to merge develop in and bring this up-to-date?

sure

node/node.go
melekes: Okay, I think we should not do that because

a) DefaultSeedDisconnectWaitPeriod depends on a private const blocksToContributeToBecomeGoodPeer
b) and average time between blocks, which may change during the course of the network

That said, I think it should be a function, not a const

p2p/pex/pex_reactor.go
melekes: To ensure `now.Sub(peerInfo.LastCrawled) < defaultCrawlPeerInterval` is false the first time we add the peer

CHANGELOG_PENDING.md
```diff
@@ -16,6 +16,10 @@
 * P2P Protocol
 
 ### FEATURES:
+- [p2p/pex] Add `p2p.seed_crawl_data_filename` config variable, which is used
+  in seed mode for storing crawling data. Set to `data/seed_crawl_data.json` by
+  default. If "", no data will be saved (this may lead to seed crawling a peer
```
```suggestion
  default. If  `p2p.seed_crawl_data_filename` is empty, no data will be saved (this may lead to seed crawling a peer
```

p2p/pex/pex_reactor_test.go
```diff
@@ -289,31 +289,41 @@ func TestConnectionSpeedForPeerReceivedFromSeed(t *testing.T) {
...
 
-	// Create a peer, add it to the peer set and the addrbook.
-	peer := p2p.CreateRandomPeer(false)
-	p2p.AddPeerToSwitch(pexR.Switch, peer)
-	addr1 := peer.NodeInfo().NetAddress()
-	pexR.book.AddAddress(addr1, addr1)
+	assert.Equal(t, 0, sw.Peers().Size())
+
+	peer := testCreateDefaultPeer(dir, 1)
+	require.Nil(t, peer.Start())
```
Liamsi:
```suggestion
	require.NoError(t, peer.Start())
```

p2p/pex/pex_reactor_test.go
```diff
@@ -289,31 +289,41 @@ func TestConnectionSpeedForPeerReceivedFromSeed(t *testing.T) {
...
+	err = sw.Start()
+	require.NoError(t, err)
+	defer sw.Stop()
 
-	// Create a peer, add it to the peer set and the addrbook.
-	peer := p2p.CreateRandomPeer(false)
-	p2p.AddPeerToSwitch(pexR.Switch, peer)
-	addr1 := peer.NodeInfo().NetAddress()
-	pexR.book.AddAddress(addr1, addr1)
+	assert.Equal(t, 0, sw.Peers().Size())
```
Liamsi:
```suggestion
	assert.Zero(t, sw.Peers().Size())
```

p2p/errors.go
```diff
@@ -172,3 +172,13 @@ type ErrNetAddressLookup struct {
 func (e ErrNetAddressLookup) Error() string {
 	return fmt.Sprintf("Error looking up host (%s): %v", e.Addr, e.Err)
 }
+
+// ErrDialingOrExistingAddress indicates that we're currently
+// dialing this address or it belongs to an existing peer.
+type ErrDialingOrExistingAddress struct {
```
Liamsi: Nit: maybe rename to `ErrIsDialingOrExistingAddress` or `ErrCurrentlyDialingOrExistingAddress`? When I first read the code I thought this mean "error while dialing or dialing an existing address".

p2p/pex/addrbook.go
```diff
@@ -550,8 +537,7 @@ func (a *addrBook) getBucket(bucketType byte, bucketIdx int) map[string]*knownAd
 	case bucketTypeOld:
 		return a.bucketsOld[bucketIdx]
 	default:
-		cmn.PanicSanity("Should not happen")
-		return nil
+		panic("Should not happen")
```
Liamsi: Can we change this to: 
```suggestion
		panic("Invalid bucket type")
```

p2p/errors.go
```diff
@@ -172,3 +172,13 @@ type ErrNetAddressLookup struct {
...
+	Addr string
+}
+
+func (e ErrDialingOrExistingAddress) Error() string {
+	return fmt.Sprintf("Connection with %s has been established or dialed", e.Addr)
```
Liamsi: As far as I remember we switched to follow the convention that error messages are non-capitalized?

Liamsi:
```suggestion
	return fmt.Sprintf("connection with %s has been established or dialed", e.Addr)
```

melekes:
I don't like the fact that we are using multiple storage formats and mechanisms:

- addrbook: json file
- seed data: json file
- cs state, evidence, .. - leveldb
- indexer - leveldb (should be something more suitable for range queries)

This also brings a question of using an external DB like Postgresql? Ext. DB will require our users to setup it, but that's a small lose comparing to possible gains (schema, schema upgrades, SQL, performance, stability proven by years). There's also sqllite, which sits in the middle of the spectrum.

melekes:
Maybe we don't need to persist seed data after all (the chance of seed going down is low; even if does goes down, contacting peers sooner than allowed period is OK if this happens rarely).

melekes:
I think this is ready to merge. PTAL @ebuchman @Liamsi

p2p/switch.go
```diff
@@ -365,10 +363,14 @@ func (sw *Switch) reconnectToPeer(addr *NetAddress) {
...
 		if err == nil {
 			return // success
 		}
+		if _, ok := err.(ErrCurrentlyDialingOrExistingAddress); ok {
```
ebuchman: Can we make this an `else if` ? Nice to keep the structure of the code reflect the logic

node/node.go
```diff
@@ -498,6 +498,9 @@ func NewNode(config *cfg.Config,
 			&pex.PEXReactorConfig{
 				Seeds:    splitAndTrimEmpty(config.P2P.Seeds, ",", " "),
 				SeedMode: config.P2P.SeedMode,
+				// see consensus/reactor.go: blocksToContributeToBecomeGoodPeer
+				// 10000 blocks assuming 1s blocks ~ 2.7 hours.
```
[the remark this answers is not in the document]
melekes: > Would be super cool if this could be dynamic based on the actual block latencies from the live network

Thought so too! But it's hard to pull right now since a) `blocksToContributeToBecomeGoodPeer` is private and can be moved/changed in the future b) average block time is not exposed in consensus (prometheus does this, but we can't rely on external services)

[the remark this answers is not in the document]
melekes: https://github.com/tendermint/tendermint/issues/3523

p2p/pex/pex_reactor.go
```diff
@@ -589,80 +594,76 @@ func (r *PEXReactor) crawlPeersRoutine() {
 func (r *PEXReactor) hasPotentialPeers() bool {
 	out, in, dial := r.Switch.NumPeers()
 
-	return out+in+dial > 0 && len(r.book.ListOfKnownAddresses()) > 0
+	return out+in+dial > 0 && r.book.Size() > 0
```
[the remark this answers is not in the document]
melekes: We pick addresses from the addrbook (`try := r.book.PickAddress(newBias)`), so no - it should be &&. 

Plus we save outbound & inbound peers in addrbook, so if a node has any outbound / inbound peers, its addrbook should not be empty.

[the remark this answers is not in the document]
ebuchman: So why is the `r.book.Size() > 0` check even necessary ?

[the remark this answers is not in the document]
melekes: I think it should be just `r.book.Size() > 0`.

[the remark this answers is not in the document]
melekes: You're right! r.book.Size is not necessary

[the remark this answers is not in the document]
melekes: I'll remove

Issue #2980, opened 2018-12-07
There is something wrong with CORSAllowedOrigins


HaoyangLiu:
config.toml file
```
# A list of origins a cross-domain request can be executed from
# Default value '[]' disables cors support
# Use '["*"]' to allow any origin
cors_allowed_origins = "[]"

# A list of methods the client is allowed to use with cross-domain requests
cors_allowed_methods = "[HEAD GET POST]"

# A list of non simple headers the client is allowed to use with cross-domain requests
cors_allowed_headers = "[Origin Accept Content-Type X-Requested-With X-Server-Time]"

```
Load result:
```
*config = {*github.com/irisnet/irishub/vendor/github.com/tendermint/tendermint/config.Config} 
 BaseConfig = {github.com/irisnet/irishub/vendor/github.com/tendermint/tendermint/config.BaseConfig} 
 *RPC = {*github.com/irisnet/irishub/vendor/github.com/tendermint/tendermint/config.RPCConfig} 
  RootDir = "/home/lhy/temp/debug/iris-normal"
  ListenAddress = "tcp://0.0.0.0:26657"
  CORSAllowedOrigins = {[]string} len:2, cap:2
   0 = "91"
   1 = "93"
  CORSAllowedMethods = {[]string} len:15, cap:24
   0 = "91"
   1 = "72"
   2 = "69"
   3 = "65"
   4 = "68"
   5 = "32"
   6 = "71"
   7 = "69"
   8 = "84"
   9 = "32"
   10 = "80"
   11 = "79"
   12 = "83"
   13 = "84"
   14 = "93"
  CORSAllowedHeaders = {[]string} len:59, cap:80
  GRPCListenAddress = ""
  GRPCMaxOpenConnections = 900
  Unsafe = false
  MaxOpenConnections = 900
```
It seems that the string array is not well supported by toml serializer. @hleb-albau 

Besides, if you repeat to execute tendermint init on the same home, these following sections will be very large.
```
CORSAllowedOrigins
CORSAllowedMethods
CORSAllowedHeaders
```

melekes:
What serializer are you using? "[]" are not forbidden in strings https://github.com/toml-lang/toml#user-content-string nor do they require escaping.

hleb-albau:
@HaoyangLiu try without quotes

```
cors_allowed_origins = []
cors_allowed_methods = [HEAD GET POST]
cors_allowed_headers = [Origin Accept Content-Type X-Requested-With X-Server-Time]
[...]

PR #2992, opened 2018-12-10
set allow_duplicate_ip to false

melekes:
Refs https://github.com/tendermint/tendermint/issues/2712

* [x] Updated all relevant documentation in docs
* [x] Updated all code comments where relevant
* [ ] Wrote tests
* [x] Updated CHANGELOG_PENDING.md

PR #3013, opened 2018-12-13
#2980 fix cors doc

hleb-albau:


melekes:
Depends on https://github.com/tendermint/tendermint/pull/2992

PR #3028, opened 2018-12-16
Add metrics for tx in cache errors

mattkanwisher:
Under high load we see a lot of errors with "Tx already exists in cache", we are trying to track down is this is something wrong with the client or the server. This seems like a very useful metric to determine the health of a Tendermint node.

* [ [N/A] Updated all relevant documentation in docs
* [ X] Updated all code comments where relevant
* [N/A ] Wrote tests
* [X ] Updated CHANGELOG_PENDING.md

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -20,4 +20,6 @@ Special thanks to external contributors on this release:
 
 ### IMPROVEMENTS:
 
+Added a Metric to track errors when a Tx is already in the Mempool cache.
+
 ### BUG FIXES:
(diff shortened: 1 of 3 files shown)


ebuchman:
#2778 should alleviate this somewhat, but we can also add the metrics

melekes:
But it's not an error https://github.com/tendermint/tendermint/blob/9a6dd96cba96155cfbb590cc522b943d6deda499/mempool/reactor.go#L93

What made you think it's an error?

> This seems like a very useful metric to determine the health of a Tendermint node.

It's not. It's due to nodes gossiping transactions to each other. I.e. node A can receive the same transaction from node B and node C. Does not mean something is wrong with it.

ValarDragon:
What is the performance overhead of metrics? (I doubt its large, but I think its important to know before adding more work to caches. I think we already have cache benchmarks which could just be ran?)

melekes:
Closing due to no response.

PR #3040, opened 2018-12-17
secret connection: check for low order points

Liamsi:
> Implement a check for the blacklisted low order points, ala the X25519 has_small_order() function in libsodium 

(https://github.com/tendermint/tendermint/issues/3010#issuecomment-447152260)
resolves first half of #3010

* [x] Wrote tests


p2p/conn/secret_connection.go
```diff
@@ -237,6 +243,52 @@ func shareEphPubKey(conn io.ReadWriteCloser, locEphPub *[32]byte) (remEphPub *[3
 	return &_remEphPub, nil
 }
 
+// use the samne blacklist as lib sodium (see https://eprint.iacr.org/2017/806.pdf for reference):
+// https://github.com/jedisct1/libsodium/blob/536ed00d2c5e0c65ac01e29141d69a30455f2038/src/libsodium/crypto_scalarmult/curve25519/ref10/x25519_ref10.c#L11-L17
+var blacklist = [][32]byte{
```
melekes: why not a const?

ValarDragon: I'll check the paper later today, but this list seems incomplete. IIRC, the multiplicative group of Curve25519 is of order `8p`, so there should be 8 points in the subgroup `<g^p>`, whereas only 3 or 4 points from the subgroup are listed here. (The last 3 points are just non-canonical forms, didn't check the 4th to last)

Instead of blacklisting like this, may just be easier to check if a point is low order. (`x^8 == 0`)

Liamsi: []byte can't be const AFAIK

melekes: oh, ok 😌

Liamsi: You are right that there are more points which aren't valid. This method is just a band-aid that detects some funky behaviour. According to djb even these exclusions are not necessary for DH: https://cr.yp.to/ecdh.html#validate
But there are other takes: https://research.kudelskisecurity.com/2017/04/25/should-ecdh-keys-be-validated/

In any case, AFAIU, we should additionally authenticate the transcript as Tony suggested in the discussion of https://github.com/tendermint/tendermint/issues/3010#issuecomment-447043829. This probably makes worrying about invalid / low-order points in this context obsolete.

HarryR:
apologies for late delay.

you can verify the point is of low order by multiplying by the cofactor and verifying the result is infinity, which is more generic than the blacklist

tarcieri:
@HarryR while that's technically true, it's also unnecessary by design in X25519

https://moderncrypto.org/mail-archive/curves/2017/000896.html

> The X25519 DH function (aka "Curve25519" from 2006) handles all input public keys without returning errors, regardless of whether the input is fully reduced mod the field prime, on the correct curve, or generates the correct subgroup.

The reason it's being added is to "stop the bleeding" with regard to the specific design of the current key exchange. Here's a 2016 writeup on the attack described in https://github.com/tendermint/tendermint/issues/3010 in a similarly designed key exchange protocol:

https://vnhacker.blogspot.com/2016/08/the-internet-of-broken-protocols.html

That post suggests blacklisting low order points. It does solve that particular attack. However the core flaw of this key exchange is the attacker is able to undetectably rewrite traffic, and so long as they can do that the [protocol is potentially vulnerable to a wide range of attacks including downgrade attacks on future versions of the protocol (See section 6)](http://www.cs.ox.ac.uk/files/10583/downgrade-taxonomy-18.pdf).

If the current key exchange concluded with a MAC of the transcript, no checks would be strictly necessary, even the blacklist.

HarryR:
Yup you are correct, checking all zeros is the most elegant way

Issue #2920, opened 2018-11-27
Remove BlockTimeIota and other Consensus Params from Config


ebuchman:
It should be a consensus parameter, not a locally configurable option

_Originally posted by @milosevic in https://github.com/tendermint/tendermint/pull/2878_

ebuchman:
Actually, we probably want to remove most of the consensus timeout params from the config since according to the spec now there are formal relations between them. Instead, we should only expose a couple knobs, and derive all the other values from those.

Namely, we have

```
timeoutPropose(r)>2∆+timeoutPrecommit(r−1)
timeoutPrevote(r)>2∆
timeoutPrecommit(r)> 2∆
```

where delta is the partial synchrony bound.
this delta should be increasing with each round (since the bound is actually unknown) according to some other delta. this latter delta can be part of the config.

ebuchman:
We should be able to reduce the number of consensus timeout params in the config down to roughly the following three:

- initial proposer timeout
- round delta
- timeout commit

ebuchman:
Bumping this back a milestone so v0.29 can ship with just the protocol breaking stuff

ValarDragon:
I'm in favor of of removing `time_iota_ms` as a param, and hardcoding it as `1ms` in the code. (As its only role AFAIK is to ensure monotonicity of proposed block time)


I think removing the parameter would be a breaking change to the encoding of the `BlockParams`, but is removing the usage of `time_iota_ms` in the code a breaking change? (And then just leaving the parameter as deadcode?)

If making `time_iota_ms` unused, and instead hardcoded to 1 ms is fine, I'm happy to submit a PR for it.

cmwaters:
Yeah I am for this, especially as it can cause problems if set incorrectly and doesn't really need to be there. 

I guess regardless of whether this gets fixed now or not, it will most likely be removed anyway in the next major release (especially if we implement proposer based timestamps). So I don't think it's too much of a problem if you don't submit a PR, but I see it's only going to be a couple of lines, so if you do then I'm happy to try get it in to 0.34.1

marbar3778:
BlockTimeIota was removed recently, we need to remove the config params in favor of a more global approach (#5911)

williambanfield:
I'd like to close this in favor of 
https://github.com/tendermint/tendermint/issues/2920 since `BlockTimeIota` was removed and this issue is therefore a bit unfocused.

williambanfield:
[...]

PR #3048, opened 2018-12-19
make BlockTimeIota a consensus parameter, not a locally configurable …

melekes:
…option

Refs #2920

* [ ] Wrote tests


types/protobuf.go
```diff
@@ -222,16 +224,28 @@ func (pb2tm) ValidatorUpdates(vals []abci.ValidatorUpdate) ([]*Validator, error)
...
 func (pb2tm) ConsensusParams(csp *abci.ConsensusParams) ConsensusParams {
-	return ConsensusParams{
-		BlockSize: BlockSizeParams{
-			MaxBytes: csp.BlockSize.MaxBytes,
-			MaxGas:   csp.BlockSize.MaxGas,
-		},
-		Evidence: EvidenceParams{
+	params := ConsensusParams{}
+
+	// we must defensively consider any structs may be nil
```
Liamsi: 👍

state/state_test.go
```diff
@@ -419,11 +421,16 @@ func TestConsensusParamsChangesSaveLoad(t *testing.T) {
 	}
 }
 
-func makeParams(blockBytes, blockGas, evidenceAge int64) types.ConsensusParams {
+func makeParams(
+	blockBytes, blockGas int64,
```
Liamsi: Nit: is there a reason those are paired while other params get a separate line?

ebuchman: Same type, similar semantic meaning (ie. block maximums)

types/time/duration.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"time"
+)
+
+// DurationPretty is a wrapper around time.Duration implementing custom
+// marshaller/unmarshaller which make it pretty (e.g. "10s", not
+// "10000000000").
+type DurationPretty struct {
```
ebuchman: Is there some reason this has to be a struct? Doesn't that change how this field would be amino encoded compared to using `type DurationPretty time.Duration`? 

Ideally the Amino encoding of structs in `types` matches the protobuf encoding of the types in `abci/types`

ebuchman: Hmm. I don't necessarily think its bad to have the custom marshalling so it reads nicer, but just want the Amino marshalling to be consistent. So either we'd need to implement the MarshalBinary methods (does Amino even let you do that yet?) or make it not wrapped in a struct.

We could remove the prettifying all-together because it is JSON after all, which is less intended for human modification. But seeing "100ms" will always be nicer than "100000000" ... 

I think ideally even it would just be in milliseconds, so eg. "100" would be fine, but not sure we can control that without wrapping the type anyways!

Liamsi: Amino gives you the possibility of custom marhsalling via implementing `MarshalAmino` / `UnmarshalAmino` (and `MarshalJSON` / `UnmarshalJSON`)

Liamsi: See a simple example here: https://gist.github.com/Liamsi/01729f51259f912b0dd64e686025eaee

melekes: > Ideally the Amino encoding of structs in types matches the protobuf encoding of the types in abci/types

I believe the way protobuf encodes google.protobug.Duration is different from how amino encodes time.Duration.

melekes: Looks like the simplest thing to do is to make TimeIota int64 ms.

abci/types/types.proto
```diff
@@ -207,17 +207,19 @@ message ResponseCommit {
...
 message ConsensusParams {
-  BlockSizeParams block_size = 1;
+  BlockParams block = 1;
   EvidenceParams evidence = 2;
   ValidatorParams validator = 3;
 }
 
-// BlockSize contains limits on the block size.
-message BlockSizeParams {
+// BlockParams contains limits on the block size.
```
ebuchman:
```suggestion
// BlockParams contains limits on the block size and timestamp.
```

config/config.go
```diff
@@ -646,16 +642,9 @@ func TestConsensusConfig() *ConsensusConfig {
 	cfg.SkipTimeoutCommit = true
 	cfg.PeerGossipSleepDuration = 5 * time.Millisecond
 	cfg.PeerQueryMaj23SleepDuration = 250 * time.Millisecond
-	cfg.BlockTimeIota = 10 * time.Millisecond
 	return cfg
 }
 
-// MinValidVoteTime returns the minimum acceptable block time.
-// See the [BFT time spec](https://godoc.org/github.com/tendermint/tendermint/docs/spec/consensus/bft-time.md).
-func (cfg *ConsensusConfig) MinValidVoteTime(lastBlockTime time.Time) time.Time {
```
ebuchman: Maybe we should move this to be a method on the BlockParams instead?

types/params.go
```diff
@@ -81,18 +85,23 @@ func (params *ValidatorParams) IsValidPubkeyType(pubkeyType string) bool {
...
 
-	if params.BlockSize.MaxGas < -1 {
-		return cmn.NewError("BlockSize.MaxGas must be greater or equal to -1. Got %d",
-			params.BlockSize.MaxGas)
+	if params.Block.MaxGas < -1 {
+		return cmn.NewError("Block.MaxGas must be greater or equal to -1. Got %d",
+			params.Block.MaxGas)
+	}
+
+	if params.Block.TimeIotaMs < 0 {
```
ebuchman: Can this really be 0?

milosevic: It should probably be always positive so we ensure time monotonicity.

PR #3049, opened 2018-12-19
Normalize priorities to not exceed total voting power

### LINKED ISSUE #2960: Problems related to ProposerPriority

Liamsi:
Followup PR to #2966 
This should fix #2960

It basically introduces a rescaling/normalization of the priorities to make sure that the prios do not exceed 2*TotalVotingPower. Also, the averaging is done only once (at the end of an IncrementProposerPriority step instead of in each iteration).

@cwgoes and me wrote up a little proof/formalization why this works. It boils down to to showing:

For any sequence of proposer elections with no power changes and n blocks, the difference between n*power/totalPower and the number of proposed blocks (for each validator) is at most 1.

Work in progress as I still want to tex that proof and discuss it with the team first.
 

* [x] Wrote tests


cwgoes:
I've written up a formalization of the fairness proof in Idris [here](https://github.com/cwgoes/tm-proposer-idris) - right now just for the two-validator case, but I believe @ValarDragon had a reduction from the n-validator case to the 2-validator case (which is also worth writing up).

types/validator_set.go
```diff
@@ -78,44 +78,45 @@ func (vals *ValidatorSet) IncrementProposerPriority(times int) {
...
 	// Decrement the validator with most ProposerPriority:
-	mostest := validatorsHeap.Peek().(*Validator)
+	mostest := vals.getValWitMostPriority()
 	// mind underflow
 	mostest.ProposerPriority = safeSubClip(mostest.ProposerPriority, vals.TotalVotingPower())
 
 	return mostest
 }
 
+func (vals *ValidatorSet) dividePrioritiesBy(divisor int64) {
```
Liamsi: note to myself: add a comment that this is not expected to be called with divisor == 0

types/validator_set.go
```diff
@@ -78,44 +78,45 @@ func (vals *ValidatorSet) IncrementProposerPriority(times int) {
 		panic("Cannot call IncrementProposerPriority with non-positive times")
 	}
 
-	const shiftEveryNthIter = 10
+	// cap the difference between priorities to be proportional to 2*totalPower by
+	// re-normalizing priorities, i.e., rescale all priorities by multiplying with:
+	//  2*totalVotingPower/(maxPriority - minPriority)
+	diff := computeMaxMinPriorityDiff(vals)
+	threshold := 2 * vals.TotalVotingPower()
+	if diff > threshold && threshold > 0 {
```
ebuchman: Why would we ever have `threshold <= 0`

Liamsi: As far as I remember, I've observed total voting power being 0 (at least in tests). Then, threshold would be 0, too.

ebuchman: Ok. Shouldn't ever happen. Maybe there's some weird test where it does. I suppose it's fine for a sanity check but lets leave a note that it shouldn't happen and we ought to be able to remove that check

Liamsi: E.g., the test `TestAveragingInIncrementProposerPriority` doesn't init voting powers. Only prios. as it only aims to test the priority averaging logic. But I remember there were other unrelated tests with voting power 0, too.

Liamsi:
> One thing I'm wondering about amidst all these renormalizing is how we're handling ties in priority - it looks like we break them based on an address comparison but could that compromise fairness if the same address always wins?

Hmm, looks you are right on first sight. But being elected proposer changes the priority and hence the next round won't consist of the same tie. Also, there are few tests that test that proposers with the same power alternate, e.g. `TestProposerPriorityProposerAlternates`: https://github.com/tendermint/tendermint/pull/3049/files#diff-a1153b96e079da8da43231ad48de57a0L334

ebuchman:
> But being elected proposer changes the priority and hence the next round won't consist of the same tie.

Yeh but I wonder if there's a situation where you can end up with the same priority again, before the other guy got to be proposer, and ultimately prevent them from ever being proposer, or at least prevent it for a long time.

types/validator_set.go
```diff
@@ -78,44 +78,52 @@ func (vals *ValidatorSet) IncrementProposerPriority(times int) {
 		panic("Cannot call IncrementProposerPriority with non-positive times")
 	}
 
-	const shiftEveryNthIter = 10
+	// cap the difference between priorities to be proportional to 2*totalPower by
```
milosevic: As imbalance happens only upon validator set changes, I would suggest to move this code to `execution.go/updateValidators` and/or `validator_set.go/Add, validator_set.go/Remove`. Idea is to to averaging and normalisation only when val set is changed and then have "plain" proposer selection at the beginning of each round. As this might involve significant changes maybe it's better to address this in a separate issue/PR.

Liamsi: Thanks for the clarification. That makes sense for the normalisation, not sure about the averaging/shifting but probably too.

types/validator_set.go
```diff
@@ -78,44 +78,52 @@ func (vals *ValidatorSet) IncrementProposerPriority(times int) {
...
+	diff := computeMaxMinPriorityDiff(vals)
+	threshold := 2 * vals.TotalVotingPower()
+	// the 2nd check (threshold > 0) is merely a sanity check which could be
+	// removed if all tests would init. voting power appropriately;
+	// i.e. threshold should always be > 0
+	if diff > threshold && threshold > 0 {
+		// div = Ceil((maxPriority - minPriority) / 2*totalVotingPower)
+		// threshold > 0 and diff > threshold guarantees (diff / threshold > 0):
+		div := int64(math.Ceil(float64(diff) / float64(threshold)))
+		vals.dividePrioritiesBy(div)
```
Liamsi: we could probably inline `dividePrioritiesBy` here as it's just a for-loop and we would save one casting to int and back to float64 inside of `dividePrioritiesBy`.

types/validator_set.go
```diff
@@ -78,44 +78,52 @@ func (vals *ValidatorSet) IncrementProposerPriority(times int) {
...
-		shiftByAvgProposerPriority := i%shiftEveryNthIter == 0
-		proposer = vals.incrementProposerPriority(shiftByAvgProposerPriority)
-	}
-	isShiftedAvgOnLastIter := (times-1)%shiftEveryNthIter == 0
-	if !isShiftedAvgOnLastIter {
-		validatorsHeap := cmn.NewHeap()
-		vals.shiftByAvgProposerPriority(validatorsHeap)
+		proposer = vals.incrementProposerPriority()
 	}
+	vals.shiftByAvgProposerPriority()
```
cwgoes: Are we sure averaging only once per call is safe? If we don't average every `n` times, and `times` is really large (maybe when catching up syncing a node), it seems like we might run into the clipping logic and end up with a different final set of priorities (and possibly validators).

types/validator_set.go
cwgoes: How about we panic? Dividing-by-zero is definitely justification for panicking. Will Golang panic anyways?

types/validator_set.go
```diff
@@ -131,11 +139,34 @@ func (vals *ValidatorSet) computeAvgProposerPriority() int64 {
...
+		}
+	}
+	diff := max - min
+	return diff
+}
+
+func (vals *ValidatorSet) getValWitMostPriority() *Validator {
+	var res *Validator
+	for _, val := range vals.Validators {
+		res = res.CompareProposerPriority(val)
```
cwgoes: This is pretty unintuitive to read - do most objects in Tendermint have functions called `CompareXYZ` which return the higher-valued object by some property? Maybe I'm just not used to it. Otherwise I think directly comparing `val.ProposerPriority` or having a function `LargestValidatorByPriority` would be a bit more obvious.

types/validator_set.go
```diff
@@ -131,11 +139,34 @@ func (vals *ValidatorSet) computeAvgProposerPriority() int64 {
...
+		}
+		if v.ProposerPriority > max {
+			max = v.ProposerPriority
+		}
+	}
+	diff := max - min
+	return diff
+}
+
+func (vals *ValidatorSet) getValWitMostPriority() *Validator {
```
cwgoes:
```suggestion
func (vals *ValidatorSet) getValWithMostPriority() *Validator {
```

types/validator_set.go
```diff
@@ -78,44 +78,52 @@ func (vals *ValidatorSet) IncrementProposerPriority(times int) {
...
-		vals.shiftByAvgProposerPriority(validatorsHeap)
-	} else { // just update the heap
-		for _, val := range vals.Validators {
-			validatorsHeap.PushComparable(val, proposerPriorityComparable{val})
-		}
-	}
-
 	// Decrement the validator with most ProposerPriority:
-	mostest := validatorsHeap.Peek().(*Validator)
+	mostest := vals.getValWitMostPriority()
```
cwgoes:
```suggestion
	mostest := vals.getValWithMostPriority()
```

CHANGELOG_PENDING.md
```diff
@@ -18,6 +18,8 @@ Special thanks to external contributors on this release:
   * [merkle] \#2713 Merkle trees now match the RFC 6962 specification
 
 * P2P Protocol
+  - [consensus] \#2960 normalize priorities to not exceed `2*TotalVotingPower` to mitigate unfair proposer selection
```
cwgoes: Note that `2 * TotalVotingPower` doesn't give the strictest possible fairness bound - one validator might still get an extra block after some power changes - but I think the bound it gives is enough in practice.

types/validator_set.go
Liamsi: Good point. Given the speed tendermint creates blocks, being offline for a few weeks (or months) may cause the clipping to kick in. Not sure how realistic that actually is but we should probably account for this.  Previously, we had every 10th iter. Maybe we should just make this number a bit larger then?

types/validator_set.go
Liamsi: It is the same method as used in `GetProposer` (via `findProposer`). I agree that it is weird, though.

types/validator_set.go
ebuchman: Go will panic

ebuchman:
@Liamsi @cwgoes would appreciate a look at the PR I just merged in https://github.com/tendermint/tendermint/pull/3155

Some final points before this is truly complete:
- review #3155 
- avg every n times instead of just once per call to increment
- review/eliminate use of floats (!)

CHANGELOG_PENDING.md
ebuchman: Why are we using 2* ? Doesn't the idris proof just use the 1* ?

Liamsi:
> review/eliminate use of floats (!)

I removed the usage of float64 altogether. To compute the `div`,  I manually ceil (add 1 if `diff` isn't a multiple of threshold) and I removed the 2nd ceil in `dividePrioritiesBy`.

ebuchman:
Merging this. Thanks for the hard work everyone. Let's follow up in a new PR if need be.

CHANGELOG_PENDING.md
cwgoes: I think because we wanted to retain the `-9/8 * TotalPower` initial priority for newly added validators - which would be negated if we immediately renormalized to max-diff `TotalPower` afterwards.

types/validator_set.go
```diff
@@ -78,44 +78,57 @@ func (vals *ValidatorSet) IncrementProposerPriority(times int) {
...
-func (vals *ValidatorSet) incrementProposerPriority(subAvg bool) *Validator {
-	for _, val := range vals.Validators {
-		// Check for overflow for sum.
-		val.ProposerPriority = safeAddClip(val.ProposerPriority, val.VotingPower)
+func (vals *ValidatorSet) RescalePriorities(diffMax int64) {
+	// NOTE: This check is merely a sanity check which could be
+	// removed if all tests would init. voting power appropriately;
+	// i.e. diffMax should always be > 0
+	if diffMax == 0 {
+		return
```
melekes: Is the purpose here to never panic? What if diffMax is 0, will we be able to find it out?

Liamsi: It is probably appropriate to panic here. Some tests do not init. voting power accordingly though. We should fix those (as all those would panic). Outside of tests this (diffMax <= 0= should never happen.

PR #3053, opened 2018-12-20
Hotfix/validating query result length

gianfelipe93:
Ref: https://github.com/tendermint/tendermint/issues/3080

Validating that there are txs in the query results before looping through the array

* [X] Wrote tests


gianfelipe93:
Hey guys, this is the first time I open a PR so let me know if I didn't follow the guideline or code of conduct properly.

rpc/core/tx.go
```diff
@@ -205,7 +205,8 @@ func TxSearch(query string, prove bool, page, perPage int) (*ctypes.ResultTxSear
 
 	apiResults := make([]*ctypes.ResultTx, cmn.MinInt(perPage, totalCount-skipCount))
 	var proof types.TxProof
-	for i := 0; i < len(apiResults); i++ {
+	// if there's no tx in the results array, we don't need to loop through the apiResults array
+	for i := 0; i < len(apiResults) && len(results) > 0; i++ {
```
melekes: Is this really needed? The test case above seems to cover a different use-case.

if `totalCount` is 0, then the `len(apiResults)` will also be 0

gianfelipe93: Not necessarily @melekes, I had this issue myself when I had to query the page 1 returning one item per page.

The app crashes returning `index out of range` because of this line:
`r := results[skipCount+i]`

if results is empty, I get that error.

You can reproduce it by removing the `&& len(results)` and running the test I created.

melekes: oh okay

gianfelipe93: all good ?

Gopkg.lock
```diff
@@ -9,14 +9,6 @@
   pruneopts = "UT"
   revision = "3a771d992973f24aa725d07868b467d1ddfceafb"
 
-[[projects]]
```
ebuchman: why is this stuff being removed?

gianfelipe93:
Hi @ebuchman, thanks for your help.

> Typical process is to first open an issue describing the problem, and then submit a PR referencing it.

How can I do that ? I have opened the issue: https://github.com/tendermint/tendermint/issues/3080

> For the changelog entry, you should include the PR number, so like [rpc] \#3053.

Done! Hope it's right now.

> As for the fix itself, we should probably catch it before the loop - it's incorrect that apiResults is being created with size > 0 if there are no results to return! So we should fix it there. Possibly skipCount just needs to be bounded below at 0 (currently it can get to -1).

Done as well.

> why is this stuff being removed? 

Sorry, I don't know as well, I don't recall changing this file.

Let me know if I have to change anything.

Have a good 2019 guys!

rpc/core/tx.go
melekes: yes

rpc/core/pipe.go
```diff
@@ -154,3 +154,12 @@ func validatePerPage(perPage int) int {
 	}
 	return perPage
 }
+
+func validateSkipCount(page, perPage int) int {
+	skipCount := (page - 1) * perPage
```
melekes: I don't think this is correct. How skipCount can be negative? if page >= 1

ebuchman: You should be right but I think validatePage can return a 0 - ie. `validatePage(1, 1, 0) == 0`

melekes: I will double check.

PR #3056, opened 2018-12-20
Skip missed proposer to current proposer in NextValidators(#3044)

bshang:
1. skip missed proposers if it has two or more
	2. differ its position in next round

* [x] Wrote tests

### DIFF
--- a/state/execution.go
+++ b/state/execution.go
@@ -432,7 +432,7 @@ func updateState(
 	}
 
 	// Update validator proposer priority and set state variables.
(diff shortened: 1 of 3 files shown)


melekes:
Could you please expand on the issue you're trying to solve?

ebuchman:
Thanks @bshang for contributing - can you please open an issue describing the problem/solution before opening a PR? It would also help to situate the change in the context of other work happening around the proposer selection currently.

PR #3059, opened 2018-12-21
node_info : bit -> Byte

needkane:
* [ ] Wrote tests

1024kb  means 1024k bit not Byte, but actually it should be Byte. 
Reference code :https://github.com/tendermint/go-amino/blob/master/amino.go#L276

### DIFF
--- a/p2p/node_info.go
+++ b/p2p/node_info.go
@@ -9,7 +9,7 @@ import (
 )
 
 const (
-	maxNodeInfoSize = 10240 // 10Kb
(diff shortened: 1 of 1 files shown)


ValarDragon:
LGTM, but its aimed at the wrong branch. (Should be against develop, not master)

ebuchman:
heh thanks!

needkane:
> 
> 
> LGTM, but its aimed at the wrong branch. (Should be against develop, not master)

Sorry， I forget

Issue #2820, opened 2018-11-12
Write a proper spec for validator signing rules


ebuchman:
Should include more details on Canonical form and validity rules around double signing prevention.

ebuchman:
https://github.com/tendermint/tendermint/pull/3061

PR #3061, opened 2018-12-21
add signing spec

ebuchman:
Ref https://github.com/tendermint/tendermint/issues/2820

Spec for those implementing validator signers.

* [ ] Wrote tests

### DIFF
--- a/docs/spec/consensus/signing.md
+++ b/docs/spec/consensus/signing.md
@@ -48,8 +48,8 @@ BlockID is the structure used to represent the block:
 
 ```
 type BlockID struct {
-	Hash	[]byte
-	Header	PartSetHeader
+	Hash        []byte
+	PartsHeader PartSetHeader
 }
 
 type PartSetHeader struct {
@@ -66,8 +66,8 @@ are true:
 
 ```
 b.Hash == nil
-b.Header.Total == 0
-b.Header.Hash == nil
+b.PartsHeader.Total == 0
+b.PartsHeader.Hash == nil
 ```
 
 `BlockID.IsComplete()` returns true for BlockID `b` if each of the following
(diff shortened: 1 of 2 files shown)


ebuchman:
I'm noticing now that we somehow managed to mangle the order of BlockID and Timestamp in the Canonical Proposal and Vote :(

I hope we can still change that, but maybe it's too late

docs/spec/consensus/signing.md
```diff
@@ -0,0 +1,205 @@
...
+and to only sign votes and proposals corresponding to that ChainID.
+
+## BlockID
+
+BlockID is the structure used to represent the block:
+
+```
+type BlockID struct {
+	Hash	[]byte
+	Header	PartSetHeader
```
Liamsi: The var in the code is still named `PartsHeader`. Could we name the vars here (in the spec) and in the implementation consistently?

docs/spec/consensus/signing.md
```diff
@@ -0,0 +1,205 @@
...
+}
+
+type PartSetHeader struct {
+	Hash  []byte
+	Total int
+}
+```
+
+To be included in a valid vote or proposal, BlockID must either represent a `nil` block, or a complete one.
+We introduce two methods, `BlockID.IsNil()` and `BlockID.IsComplete()` for these cases, respectively.
```
Liamsi: Introducing these methods in the go-code as well would make the validation more readable.

Liamsi:
Oh no. The question is, if we'd need to be backwards compatible in some way? If not, it should be fairly easy to change that.

ebuchman:
> The question is, if we'd need to be backwards compatible in some way? If not, it should be fairly easy to change that. 

We need to make a non-backwards compatible release soon (hopefully the last one for some time), so let's include it in there. I opened an issue: https://github.com/tendermint/tendermint/issues/3077

PR #3067, opened 2018-12-24
cs/replay: execCommitBlock should not read from state.lastValidators

### LINKED ISSUE #3054: panic: precommit length (X) doesn't match valset length (Y) at height Z
**What happened**:

My pr #3006 would cause a bug if the validatorSet changed after height 0
The call stack is:

```
goroutine 1 [running]:
github.com/tendermint/tendermint/state.getBeginBlockValidatorInfo(0xc4202b2b40, 0xc421e38500, 0x17a1760, 0xc4200bac10, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /root/go/src/github.com/tendermint/tendermint/state/execution.go:313 +0x7cf
github.com/tendermint/tendermint/state.execBlockOnProxyApp(0x1796760, 0xc42226ece0, 0x179c1c0, 0xc42225f040, 0xc4202b2b40, 0xc421e38500, 0x17a1760, 0xc4200bac10, 0x0, 0xc00, ...)
        /root/go/src/github.com/tendermint/tendermint/state/execution.go:250 +0x24c
github.com/tendermint/tendermint/state.ExecCommitBlock(0x179c1c0, 0xc42225f040, 0xc4202b2b40, 0x1796760, 0xc42226ece0, 0xc421e38500, 0x17a1760, 0xc4200bac10, 0xc4225ee060, 0x20, ...)
        /root/go/src/github.com/tendermint/tendermint/state/execution.go:631 +0x9b
github.com/tendermint/tendermint/consensus.(*Handshaker).replayBlocks(0xc438bcc780, 0x7, 0x0, 0xc4219e35c6, 0x6, 0xc4219e35e0, 0xc, 0x14, 0x0, 0xc4200e7980, ...)
        /root/go/src/github.com/tendermint/tendermint/consensus/replay.go:429 +0x250
github.com/tendermint/tendermint/consensus.(*Handshaker).ReplayBlocks(0xc438bcc780, 0x7, 0x0, 0xc4219e35c6, 0x6, 0xc4219e35e0, 0xc, 0x14, 0x0, 0xc4200e7980, ...)
        /root/go/src/github.com/tendermint/tendermint/consensus/replay.go:367 +0xc8a
github.com/tendermint/tendermint/consensus.(*Handshaker).Handshake(0xc438bcc780, 0x17a2160, 0xc420237110, 0x2, 0x1796760)
        /root/go/src/github.com/tendermint/tendermint/consensus/replay.go:256 +0x46f
```

error info is:

```
I[12-20|07:31:10.043] Applying block                               module=consensus height=2
panic: precommit length (12) doesn't match valset length (7) at height 2
```

Recently I was thinking whether the state should be restored from 0 and begin to change again. But what if the appHeight is not 0? For example, if the appHeight==20, stateHeight==40, and 20~40 has some validators change txs. How can we get the state in height 20?

I think there's no need to restore state and replay changes.

What we need is to fetch historical validators in `getBeginBlockValidatorInfo` (see https://github.com/tendermint/tendermint/issues/3054#issuecomment-448920376).

james-ray:
Fixes #3054 

* [x] Wrote tests


melekes:
Would be great if you could write a test case (or multiple test cases) covering the bug from #3054 and earlier changes https://github.com/tendermint/tendermint/pull/3006 you've made.

james-ray:
OK, I'll handle the test cases.

james-ray:
Hello, I manage to add a test, but now circleci failed on lint
--> Downloading linters (this may take awhile)
/go/src/github.com/alecthomas/gometalinter/scripts/install.sh -b /tmp/workspace/bin
alecthomas/gometalinter info checking GitHub for latest tag
alecthomas/gometalinter info found version: 3.0.0 for v3.0.0/linux/amd64
alecthomas/gometalinter info installed /tmp/workspace/bin/gometalinter
alecthomas/gometalinter info installed /tmp/workspace/bin/gocyclo
alecthomas/gometalinter info installed /tmp/workspace/bin/nakedret
alecthomas/gometalinter info installed /tmp/workspace/bin/misspell
alecthomas/gometalinter info installed /tmp/workspace/bin/gosec
alecthomas/gometalinter info installed /tmp/workspace/bin/golint
alecthomas/gometalinter info installed /tmp/workspace/bin/ineffassign
alecthomas/gometalinter info installed /tmp/workspace/bin/goconst
alecthomas/gometalinter info installed /tmp/workspace/bin/errcheck
alecthomas/gometalinter info installed /tmp/workspace/bin/maligned
alecthomas/gometalinter info installed /tmp/workspace/bin/unconvert
alecthomas/gometalinter info installed /tmp/workspace/bin/dupl
alecthomas/gometalinter info installed /tmp/workspace/bin/structcheck
alecthomas/gometalinter info installed /tmp/workspace/bin/varcheck
alecthomas/gometalinter info installed /tmp/workspace/bin/safesql
alecthomas/gometalinter info installed /tmp/workspace/bin/deadcode
alecthomas/gometalinter info installed /tmp/workspace/bin/lll
alecthomas/gometalinter info installed /tmp/workspace/bin/goimports
alecthomas/gometalinter info installed /tmp/workspace/bin/gotype
install: cannot stat '/tmp/tmp.VTD172w5QB/gometalinter-3.0.0-linux-amd64/gosimple': No such file or directory


something wrong with this alecthomas/gometalinter/scripts/install.sh ?

james-ray:
I have tried the linter install on my local pc, it works fine.

james@james-Latitude-5480:~/gopath/src/github.com/alecthomas/gometalinter$ ./scripts/install.sh -b /tmp/workspace/bin
alecthomas/gometalinter info checking GitHub for latest tag
alecthomas/gometalinter info found version: 3.0.0 for v3.0.0/linux/amd64
alecthomas/gometalinter info installed /tmp/workspace/bin/gometalinter
alecthomas/gometalinter info installed /tmp/workspace/bin/gocyclo
alecthomas/gometalinter info installed /tmp/workspace/bin/nakedret
alecthomas/gometalinter info installed /tmp/workspace/bin/misspell
alecthomas/gometalinter info installed /tmp/workspace/bin/gosec
alecthomas/gometalinter info installed /tmp/workspace/bin/golint
alecthomas/gometalinter info installed /tmp/workspace/bin/ineffassign
alecthomas/gometalinter info installed /tmp/workspace/bin/goconst
alecthomas/gometalinter info installed /tmp/workspace/bin/errcheck
alecthomas/gometalinter info installed /tmp/workspace/bin/maligned
alecthomas/gometalinter info installed /tmp/workspace/bin/unconvert
alecthomas/gometalinter info installed /tmp/workspace/bin/dupl
alecthomas/gometalinter info installed /tmp/workspace/bin/structcheck
alecthomas/gometalinter info installed /tmp/workspace/bin/varcheck
alecthomas/gometalinter info installed /tmp/workspace/bin/safesql
alecthomas/gometalinter info installed /tmp/workspace/bin/deadcode
alecthomas/gometalinter info installed /tmp/workspace/bin/lll
alecthomas/gometalinter info installed /tmp/workspace/bin/goimports
alecthomas/gometalinter info installed /tmp/workspace/bin/gotype
alecthomas/gometalinter info installed /tmp/workspace/bin/staticcheck
alecthomas/gometalinter info installed /tmp/workspace/bin/interfacer
alecthomas/gometalinter info installed /tmp/workspace/bin/unparam
alecthomas/gometalinter info installed /tmp/workspace/bin/gochecknoinits
alecthomas/gometalinter info installed /tmp/workspace/bin/gochecknoglobals

melekes:
Oops, looks like I've made an error while merging develop.
```
state/execution.go:351:6: syntax error: unexpected validateValidatorUpdates, expecting (
```
@james-ray can you take a look?

james-ray:
@melekes the goimports errors regarding reactor_test.go and common_test.go, I have goimports them many times, but nothing changed. Is there some goimports version issue?

james-ray:
@melekes  Hey， any review feedback?

james-ray:
@melekes  @ebuchman   Is my test case written not good? Please give some advice?

state/execution.go
```diff
@@ -66,6 +66,10 @@ func NewBlockExecutor(db dbm.DB, logger log.Logger, proxyApp proxy.AppConnConsen
 	return res
 }
 
+func (blockExec *BlockExecutor) Db() dbm.DB {
```
ebuchman: Our convention is to capitalize both letters - `DB()`

state/execution.go
```diff
@@ -296,51 +299,54 @@ func execBlockOnProxyApp(
 	return abciResponses, nil
 }
 
-func getBeginBlockValidatorInfo(block *types.Block, lastValSet *types.ValidatorSet, stateDB dbm.DB) (abci.LastCommitInfo, []abci.Evidence) {
-
-	// Sanity check that commit length matches validator set size -
-	// only applies after first block
+func getBeginBlockValidatorInfo(block *types.Block, stateDB dbm.DB) (abci.LastCommitInfo, []abci.Evidence) {
+	voteInfos := make([]abci.VoteInfo, len(block.LastCommit.Precommits))
```
ebuchman: Oh, we should use `LastCommit.Size()` here and everywhere else instead! Can be done in a separate PR though.

state/execution.go
```diff
@@ -296,51 +299,54 @@ func execBlockOnProxyApp(
...
+				Validator:       types.TM2PB.Validator(val),
+				SignedLastBlock: vote != nil,
+			}
+			voteInfos[i] = voteInfo
 		}
-		voteInfo := abci.VoteInfo{
-			Validator:       types.TM2PB.Validator(val),
-			SignedLastBlock: vote != nil,
+
+		for i, ev := range block.Evidence.Evidence {
```
ebuchman: This should be outside the `if block.Height > 1` as well - can't we include evidence in the first block?

state/execution.go
```diff
@@ -116,7 +120,7 @@ func (blockExec *BlockExecutor) ApplyBlock(state State, blockID types.BlockID, b
 	}
 
 	startTime := time.Now().UnixNano()
-	abciResponses, err := execBlockOnProxyApp(blockExec.logger, blockExec.proxyApp, block, state.LastValidators, blockExec.db)
+	abciResponses, err := execBlockOnProxyApp(blockExec.logger, blockExec.proxyApp, block, blockExec.db)
```
ebuchman: awesome. not sure why we didn't come up with this a long time ago :P

state/execution.go
```diff
@@ -296,51 +299,54 @@ func execBlockOnProxyApp(
...
 		}
-	}
 
-	// Collect the vote info (list of validators and whether or not they signed).
-	voteInfos := make([]abci.VoteInfo, len(lastValSet.Validators))
-	for i, val := range lastValSet.Validators {
-		var vote *types.CommitSig
-		if i < len(block.LastCommit.Precommits) {
-			vote = block.LastCommit.Precommits[i]
+		for i, val := range lastValSet.Validators {
```
ebuchman: Can we move this outside the `if block.Height > 1` scope? We should create `lastValSet` outside that scope and let it be an empty validator set for the first block. It's easier to read the code when less of it is so indented.

consensus/wal_generator.go
```diff
@@ -46,13 +46,14 @@ func WALGenerateNBlocks(t *testing.T, wr io.Writer, numBlocks int) (err error) {
 	if err != nil {
 		return errors.Wrap(err, "failed to read genesis file")
 	}
-	stateDB := db.NewMemDB()
 	blockStoreDB := db.NewMemDB()
+	stateDB := blockStoreDB
```
ebuchman: Why are these the same? Shouldn't they be different?

consensus/common_test.go
```diff
@@ -278,7 +280,8 @@ func newConsensusStateWithConfigAndBlockStore(thisConfig *cfg.Config, state sm.S
 	evpool := sm.MockEvidencePool{}
 
 	// Make ConsensusState
-	stateDB := dbm.NewMemDB()
+	stateDB := blockDB
+	sm.SaveState(stateDB, state) //for save height 1's validators info
```
ebuchman: I don't love how we have to remember to do this everywhere. And again why does the stateDB have to be the same as blockDB?

Maybe we should have a `newStateDB(state)` function in this package that we could use in all these tests instead?

consensus/mempool_test.go
```diff
@@ -106,7 +107,9 @@ func deliverTxsRange(cs *ConsensusState, start, end int) {
 
 func TestMempoolTxConcurrentWithCommit(t *testing.T) {
 	state, privVals := randGenesisState(1, false, 10)
-	cs := newConsensusState(state, privVals[0], NewCounterApplication())
+	blockDB := dbm.NewMemDB()
+	cs := newConsensusStateWithConfigAndBlockStore(config, state, privVals[0], NewCounterApplication(), blockDB)
+	sm.SaveState(blockDB, state)
```
ebuchman: Should this be `stateDB` ?

consensus/mempool_test.go
```diff
@@ -129,7 +132,9 @@ func TestMempoolTxConcurrentWithCommit(t *testing.T) {
 func TestMempoolRmBadTx(t *testing.T) {
 	state, privVals := randGenesisState(1, false, 10)
 	app := NewCounterApplication()
-	cs := newConsensusState(state, privVals[0], app)
+	blockDB := dbm.NewMemDB()
+	cs := newConsensusStateWithConfigAndBlockStore(config, state, privVals[0], app, blockDB)
+	sm.SaveState(blockDB, state)
```
ebuchman: same?

consensus/replay.go
```diff
@@ -257,8 +257,10 @@ func (h *Handshaker) Handshake(proxyApp proxy.AppConns) error {
 	)
 
 	// Set AppVersion on the state.
-	h.initialState.Version.Consensus.App = version.Protocol(res.AppVersion)
-	sm.SaveState(h.stateDB, h.initialState)
+	if h.initialState.Version.Consensus.App != version.Protocol(res.AppVersion) {
```
ebuchman: Do we even need the if statement here?

consensus/replay.go
```diff
@@ -417,7 +419,7 @@ func (h *Handshaker) replayBlocks(state sm.State, proxyApp proxy.AppConns, appBl
 	for i := appBlockHeight + 1; i <= finalBlock; i++ {
 		h.logger.Info("Applying block", "height", i)
 		block := h.store.LoadBlock(i)
-		appHash, err = sm.ExecCommitBlock(proxyApp.Consensus(), block, h.logger, state.LastValidators, h.stateDB)
+		appHash, err = sm.ExecCommitBlock(proxyApp.Consensus(), block, h.logger, h.stateDB)
```
ebuchman: Huh. I was pretty convinced we weren't going to be able to fix this issue without eliminating `ExecCommitBlock` and using `ApplyBlock` properly.  ExecCommitBlock doesn't mutate the `State` - so how do the updates propagate?

consensus/replay_test.go
```diff
@@ -278,42 +281,194 @@ var (
 // 2 - save block and committed but state is behind
 var modes = []uint{0, 1, 2}
 
+// This actually is not a test, it's for storing validator change tx data for testHandshakeReplay
```
ebuchman: Does it need to look like a test then? Why can't we have a function that returns a struct with all the necessary fields instead of further clutering the global namespace?

```
type testSim struct {
  GenesisState sm.State
  Config *cfg.Config
  Chain []*types.Block
  Commits []*types.Commit
  CleanupFunc cleanupFunc
}
```

consensus/replay_test.go
```diff
@@ -278,42 +281,194 @@ var (
...
+func TestSimulateValidatorsChange(t *testing.T) {
+	nPeers := 7
+	nVals := 4
+	css, genDoc, config, cleanup := randConsensusNetWithPeers(nVals, nPeers, "replay_test", newMockTickerFunc(true), newPersistentKVStoreWithPath)
```
ebuchman: Oy. It's so bad that so much complexity is required just to simulate the creation of a valid blockchain. We'll need to address this. Thanks for working through it though.

consensus/replay_test.go
```diff
@@ -278,42 +281,194 @@ var (
...
+		defer os.RemoveAll(config.RootDir)
+		testHandshakeReplay(t, config, 0, m, true)
 	}
 }
 
 // Sync many, not from scratch
 func TestHandshakeReplaySome(t *testing.T) {
 	for i, m := range modes {
-		config := ResetConfig(fmt.Sprintf("%s_%v", t.Name(), i))
+		config := ResetConfig(fmt.Sprintf("%s_%v_s", t.Name(), i))
```
ebuchman: This set of three lines is repeated so often we should make it its own function? then we can just call it twice for each iteration (once with false, once with true)

consensus/replay.go
james-ray: This judge can avoid we save state every time when do the handshake, if the app version not change, I think it's no need we save state here

consensus/replay_test.go
james-ray: Yeah, I think I could do it. But the orignial test containing some WAL generating chain process makeBlockchainFromWAL, isn't that also unreasonable? It needs to be executed every time call the testHandshakeReplay, totally 4*3=12 times.

consensus/replay_test.go
james-ray: Do you mean I can just hardcode some block data with txs changing validators, what if in the future we change some format of block or tx? then we are going to hardcode those blocks again?

consensus/common_test.go
james-ray: It's because in reactor_test the func startConsensusNet
if css[i].state.LastBlockHeight == 0 { //simulate handle initChain in handshake
			sm.SaveState(css[i].blockExec.Db(), css[i].state)
		}
we should fetch the stateDB out, but I don't get a way, I have to add a public method in blockExecutor,  so that they have to be the same.

consensus/replay.go
james-ray: In replay, we should not update the state, so I think retain ExecCommitBlock is reasonable.  Replay could call or not call ApplyBlock, if call, then state will be saved, then propagate the updates, if not, no updates made to state, so no need to propagate.

consensus/replay_test.go
james-ray: Em, that could be some better way. But hardcode is clearly not feasible.

james-ray:
Reviewer's opinion:
1. Use LastCommit.Size()     Done
2. capitalize both letters - `DB()`   Done
3. if h.initialState.Version.Consensus.App != version.Protocol(res.AppVersion) {         I think should keep it
4. move 2 for loops out of if block.Height > 1      Done
5. define sim_xxx variables into a structure in replay_test      Done
6. move ResetConfig into testHandshakeReplay      Done
7. Why stateDB and blockDB have to be same?        Explained

@ebuchman     Let's see if I drop some point or not

james-ray:
@ebuchman   I have modified that test to use no reactors, only start one consensusState, imitate the style of state_test, now the test time of it is less than 0.25s in my machine.

It still could be further simplified though, use no consensusState, just blocks and commits.

james-ray:
@ebuchman  Aloha, Looking forward to your reply.

james-ray:
@ebuchman  Hi, would you please have a look at my updates and give some advice?

james-ray:
@ebuchman   I'm sorry, what is the situation now?

james-ray:
Some advices? It's been nearly 5 mouths since this pr opened

james-ray:
The test case executing time is 0.25s, even if there is still some opinion to optimize it. At least the function in the code is good.    
I think not able to replay blocks with validators change is a some severe flaw, better patch it sooner.

james-ray:
We'd better merge the main code sooner, if you still have opinion with my test case, I can open another PR separately to refine it.

consensus/replay.go
ebuchman: I see. We already know what the validators are supposed to be in the Tendermint state, we just need to fetch that for the BeginBlock info.

consensus/common_test.go
ebuchman: not sure I follow. why can't we use `stateDB := dbm.NewMemDB()` here? Isn't the BlockExecutor supposed to use the stateDB?

consensus/replay_test.go
ebuchman: No, not hard code, but it should be much easier to build a valid blockchain with the functions. No need to go through the entire consensus process like this!

consensus/reactor.go
```diff
@@ -491,7 +491,7 @@ OUTER_LOOP:
 			if prs.ProposalBlockParts == nil {
 				blockMeta := conR.conS.blockStore.LoadBlockMeta(prs.Height)
 				if blockMeta == nil {
-					panic(fmt.Sprintf("Failed to load block %d when blockStore is at %d",
+					cmn.PanicCrisis(fmt.Sprintf("Failed to load block %d when blockStore is at %d",
```
golangcibot: PanicCrisis not declared by package common (from `typecheck`)

consensus/replay_test.go
```diff
@@ -281,46 +291,285 @@ var (
...
+	height++
+	incrementHeight(vss...)
+	newValidatorPubKey1 := css[nVals].privValidator.GetPubKey()
+	valPubKey1ABCI := types.TM2PB.PubKey(newValidatorPubKey1)
+	newValidatorTx1 := kvstore.MakeValSetChangeTx(valPubKey1ABCI, testMinPower)
+	err := assertMempool(css[0].txNotifier).CheckTx(newValidatorTx1, nil)
+	assert.Nil(t, err)
+	propBlock, _ := css[0].createProposalBlock() //changeProposer(t, cs1, vs2)
+	propBlockParts := propBlock.MakePartSet(partSize)
+	blockID := types.BlockID{propBlock.Hash(), propBlockParts.Header()}
```
golangcibot: composites: `github.com/tendermint/tendermint/types.BlockID` composite literal uses unkeyed fields (from `govet`)

consensus/replay_test.go
```diff
@@ -281,46 +291,285 @@ var (
...
+	height++
+	incrementHeight(vss...)
+	updateValidatorPubKey1 := css[nVals].privValidator.GetPubKey()
+	updatePubKey1ABCI := types.TM2PB.PubKey(updateValidatorPubKey1)
+	updateValidatorTx1 := kvstore.MakeValSetChangeTx(updatePubKey1ABCI, 25)
+	err = assertMempool(css[0].txNotifier).CheckTx(updateValidatorTx1, nil)
+	assert.Nil(t, err)
+	propBlock, _ = css[0].createProposalBlock() //changeProposer(t, cs1, vs2)
+	propBlockParts = propBlock.MakePartSet(partSize)
+	blockID = types.BlockID{propBlock.Hash(), propBlockParts.Header()}
```
golangcibot: composites: `github.com/tendermint/tendermint/types.BlockID` composite literal uses unkeyed fields (from `govet`)

consensus/replay_test.go
```diff
@@ -281,46 +291,285 @@ var (
...
+	err = assertMempool(css[0].txNotifier).CheckTx(newValidatorTx2, nil)
+	assert.Nil(t, err)
+	newValidatorPubKey3 := css[nVals+2].privValidator.GetPubKey()
+	newVal3ABCI := types.TM2PB.PubKey(newValidatorPubKey3)
+	newValidatorTx3 := kvstore.MakeValSetChangeTx(newVal3ABCI, testMinPower)
+	err = assertMempool(css[0].txNotifier).CheckTx(newValidatorTx3, nil)
+	assert.Nil(t, err)
+	propBlock, _ = css[0].createProposalBlock() //changeProposer(t, cs1, vs2)
+	propBlockParts = propBlock.MakePartSet(partSize)
+	blockID = types.BlockID{propBlock.Hash(), propBlockParts.Header()}
```
golangcibot: composites: `github.com/tendermint/tendermint/types.BlockID` composite literal uses unkeyed fields (from `govet`)

consensus/replay_test.go
james-ray: I use a new test case to simulate the chain, just start one consensusState, the others are mocked, like the style in state.go.  The time executing it is about 0.25s now.

consensus/common_test.go
james-ray: Before you suggest, I use startConsensusNet in TestSimulateValidatorsChange, which you said is too complicated and expensive just to simulate some blocks with validator change txs.
In that situation, the state must get saved in startConsensusNet, because later it will be used in replay test.
Now, I modifed TestSimulateValidatorsChange using no reactor, just start one ConsensusState, so I think deleting the SaveState code in startConsensusNet is OK. But think of the consistency of the code, we do saveState if we initChain in handshake, so startConsensusNet is to simulate that handshake, it is supposed to do exactly the same action.

james-ray: If startConsensusNet has to call SaveState, that is the problem I encountered, I have no way to fetch the stateDB in that test method, that is why I set blockExec.Db() and stateDB to be a same one.    Because I can fetch  blockExec.Db()  here.

james-ray: I see you have opened a new issue to optimize the simulate test case. Um, I think without consensus process is feasible. As you said, the test case is quite large now.   Maybe call several makeBlock() with some precommits and validator change txs will work.

james-ray: > I don't love how we have to remember to do this everywhere. And again why does the stateDB have to be the same as blockDB?
> 
> Maybe we should have a `newStateDB(state)` function in this package that we could use in all these tests instead?

I think this could work, we should ensure this function returns a singleton stateDB, that is if the state is passed in at the first time, it returns a new stateDB, else it returns a existed one.

james-ray:
I see now there are 2 suggestions,
1. Simplify that simulate test case to calling some MakeBlock() with some precommits and validatorChange Txs.
2. A lot of places need to call saveState() in order to save height 1's validator set.  About how to fetch the stateDB,   reviewer suggests 

> Maybe we should have a newStateDB(state) function in this package that we could use in all these tests instead?

> 

I think they are both good suggestions.   Would you mind approve this first?  I could work on them at another pr.

libs/pubsub/pubsub.go
```diff
@@ -404,7 +406,7 @@ func (state *state) send(msg interface{}, tags map[string]string) {
 					// don't block on buffered channels
 					select {
 					case subscription.out <- Message{msg, tags}:
-					default:
+					case <-time.After(2 * time.Second): //give some time for the subscriber to handle
```
ebuchman: Why was this necessary?

I don't think we want this, especially not with a hard coded time. We just refactored this module to not block here.

ebuchman:
Thank you for updating. I opened those two issues so we can track them without blocking this PR. This PR is basically good to merge, except for the new change to the pubsub. Is it possible to revert that? cc @melekes 

Also, would be great if you could clarify why the changes were necessary in the mempool_test.go

james-ray:
> blockDB := dbm.NewMemDB()
	cs := newConsensusStateWithConfigAndBlockStore(config, state, privVals[0], NewCounterApplication(), blockDB)
	sm.SaveState(blockDB, state)

> 
These code serve a similar purpose, to call the SaveState, otherwise I couldn't get stateDB.   We could adopt your suggestion, add a function newStateDB(state)  to do that


Why I modify pubsub is because TestStateLockPOLRelock in state_test has some probability to fail, you can try it.   I debug and found 

> state.remove(clientID, qStr, ErrOutOfCapacity)

> 
this code is the reason.  It will delete the subscriber of vote event, the ensurePrecommit or ensure Prevote will fail.
I can revert that, but the CI test may fail.   Otherwise I wouldn't have modified it.
As in the pubsub_test, I see 3 publish and 2 receive, I think the 3rd publish won't cause the  state.remove(clientID, qStr, ErrOutOfCapacity) code above, so I add one more publish.

james-ray:
I found an interesting phenomenon:

> func TestStateFullRound1(t *testing.T) {
	cs, vss := randConsensusState(1)
	height, round := cs.Height, cs.Round

	// NOTE: buffer capacity of 0 ensures we can validate prevote and last commit
	// before consensus can move to the next height (and cause a race condition)
	cs.eventBus.Stop()
	eventBus := types.NewEventBusWithBufferCapacity(0)
	eventBus.SetLogger(log.TestingLogger().With("module", "events"))
	cs.SetEventBus(eventBus)
	eventBus.Start()

	voteCh := subscribeUnBuffered(cs.eventBus, types.EventQueryVote)
	propCh := subscribe(cs.eventBus, types.EventQueryCompleteProposal)
	newRoundCh := subscribe(cs.eventBus, types.EventQueryNewRound)

	// Maybe it would be better to call explicitly startRoutines(4)
	startTestRound(cs, height, round)

	ensureNewRound(newRoundCh, height, round)

	ensureNewProposal(propCh, height, round)
	propBlockHash := cs.GetRoundState().ProposalBlock.Hash()

	ensurePrevote(voteCh, height, round) // wait for prevote
	validatePrevote(t, cs, round, vss[0], propBlockHash)

	ensurePrecommit(voteCh, height, round) // wait for precommit

	// we're going to roll right into new height
	ensureNewRound(newRoundCh, height+1, 0)

	validateLastPrecommit(t, cs, vss[0], propBlockHash)
}

> 

This case,  if we set voteCh := subscribeUnBuffered(..)  , then there is a small probability that the test case will halt there because when the publish is executed, the ensurePrevote hasn't been executed.

But if we set voteCh := subscribe(...), then there is also a small probability that the last code validateLastPrecommit(t, cs, vss[0], propBlockHash) could fail,  because it may go through several heights, once a time I ran it, it went to height 4, so this validation failed.

james-ray:
--- FAIL: TestBadBlockStopsPeer (19.83s)
    testing.go:809: race detected during execution of test

Also 2 times the CI error on this.   No hint of where or which variable the race is.

consensus/common_test.go
```diff
@@ -14,6 +14,8 @@ import (
 
 	"github.com/go-kit/kit/log/term"
 
+	"path"
```
melekes:
```suggestion
	"filepath"
```

path does not work on Windows

consensus/byzantine_test.go
```diff
@@ -81,6 +82,7 @@ func TestByzantine(t *testing.T) {
 		}
 
 		reactors[i] = conRI
+		sm.SaveState(css[i].blockExec.DB(), css[i].state) //for save height 1's validators info
```
melekes:
```suggestion
		sm.SaveState(css[i].blockExec.DB(), css[i].state) // save height=1 validators info
```

consensus/common_test.go
```diff
@@ -228,7 +248,7 @@ func validatePrevoteAndPrecommit(t *testing.T, cs *ConsensusState, thisRound, lo
 }
 
 func subscribeToVoter(cs *ConsensusState, addr []byte) <-chan tmpubsub.Message {
-	votesSub, err := cs.eventBus.Subscribe(context.Background(), testSubscriber, types.EventQueryVote)
+	votesSub, err := cs.eventBus.SubscribeUnbuffered(context.Background(), testSubscriber, types.EventQueryVote)
```
melekes: maybe we should leave a comment why SubscribeUnbuffered must be used here?

consensus/common_test.go
```diff
@@ -278,7 +298,8 @@ func newConsensusStateWithConfigAndBlockStore(thisConfig *cfg.Config, state sm.S
 	evpool := sm.MockEvidencePool{}
 
 	// Make ConsensusState
-	stateDB := dbm.NewMemDB()
+	stateDB := blockDB
```
melekes: this is super confusing. please add a comment

consensus/common_test.go
```diff
@@ -278,7 +298,8 @@ func newConsensusStateWithConfigAndBlockStore(thisConfig *cfg.Config, state sm.S
...
+	sm.SaveState(stateDB, state) //for save height 1's validators info
```
melekes:
```suggestion
	sm.SaveState(stateDB, state) // save height=1 validators info
```

consensus/common_test.go
melekes: can't we add yet another argument to `newConsensusStateWithConfigAndBlockStore` function - `stateDB dbm.DB`

consensus/common_test.go
```diff
@@ -605,15 +628,19 @@ func randConsensusNetWithPeers(nValidators, nPeers int, testName string, tickerF
...
 		}
 
-		app := appFunc()
+		app := appFunc(path.Join(config.DBDir(), fmt.Sprintf("%s_%d", testName, i)))
 		vals := types.TM2PB.ValidatorUpdates(state.Validators)
+		if _, ok := app.(*kvstore.PersistentKVStoreApplication); ok {
+			state.Version.Consensus.App = kvstore.ProtocolVersion //simulate handshake, receive app version. If don't do this, replay test will fail
+		}
 		app.InitChain(abci.RequestInitChain{Validators: vals})
+		//sm.SaveState(stateDB,state)	//height 1's validatorsInfo already saved in LoadStateFromDBOrGenesisDoc above
```
melekes: remove?

consensus/common_test.go
```diff
@@ -719,3 +746,7 @@ func newPersistentKVStore() abci.Application {
 	}
 	return kvstore.NewPersistentKVStoreApplication(dir)
 }
+
+func newPersistentKVStoreWithPath(dbDir string) abci.Application {
```
melekes: why wrap one-liner into a private function??

consensus/mempool_test.go
melekes: this is super confusing. please add a comment at the very least

consensus/mempool_test.go
melekes: this is super confusing. please add a comment at the very least

consensus/wal_generator.go
```diff
@@ -45,13 +45,14 @@ func WALGenerateNBlocks(t *testing.T, wr io.Writer, numBlocks int) (err error) {
...
```
melekes: this is super confusing. please add a comment at the very least

libs/pubsub/pubsub_test.go
```diff
@@ -46,13 +46,16 @@ func TestSubscribe(t *testing.T) {
...
+		err = s.Publish(ctx, "Ivan")
+		assert.NoError(t, err)
 	}()
 
 	select {
 	case <-published:
 		assertReceive(t, "Quicksilver", subscription.Out())
 		assertCancelled(t, subscription, pubsub.ErrOutOfCapacity)
-	case <-time.After(100 * time.Millisecond):
+	case <-time.After(3 * time.Second):
```
melekes: why increase?

libs/pubsub/pubsub_test.go
```diff
@@ -101,7 +104,7 @@ func TestSubscribeUnbuffered(t *testing.T) {
 	select {
 	case <-published:
 		t.Fatal("Expected Publish(Darkhawk) to block")
-	case <-time.After(100 * time.Millisecond):
+	case <-time.After(3 * time.Second):
```
melekes: same question

consensus/replay_test.go
```diff
@@ -140,10 +142,10 @@ LOOP:
 
 		// create consensus state from a clean slate
 		logger := log.NewNopLogger()
-		stateDB := dbm.NewMemDB()
+		blockDB := dbm.NewMemDB()
+		stateDB := blockDB
```
melekes: this is super confusing. please add a comment at the very least

consensus/replay_test.go
```diff
@@ -262,7 +264,13 @@ func (w *crashingWAL) Stop() error  { return w.next.Stop() }
 func (w *crashingWAL) Wait()        { w.next.Wait() }
 
 //------------------------------------------------------------------------------------------
-// Handshake Tests
+type testSim struct {
```
melekes: what Sim stands for?

PR #3078, opened 2019-01-03
fix order of BlockID and Timestamp in Vote and Proposal

### LINKED ISSUE #3077: types: fix order of BlockID and Timestamp in Vote and Proposal
In https://github.com/tendermint/tendermint/pull/3061 it was discovered that the BlockID and Timestamp are not consistently ordered in the Proposal and Vote (both in the main struct and the Canonical signing struct). They should be. BlockID should come first.

Liamsi:
resolves #3077

* [ ] Wrote tests


docs/spec/consensus/signing.md
```diff
@@ -59,9 +59,9 @@ type PartSetHeader struct {
 ```
 
 To be included in a valid vote or proposal, BlockID must either represent a `nil` block, or a complete one.
-We introduce two methods, `BlockID.IsNil()` and `BlockID.IsComplete()` for these cases, respectively.
+We introduce two methods, `BlockID.IsZero()` and `BlockID.IsComplete()` for these cases, respectively.
```
Liamsi: The method `IsZero` already existed. Hence preferred over `IsNil`.

Issue #3102, opened 2019-01-10
Multisig keys have 32-byte addresses


alessio:
Patch to reproduce it:

```patch
--- a/crypto/multisig/threshold_pubkey_test.go
+++ b/crypto/multisig/threshold_pubkey_test.go
@@ -94,6 +94,12 @@ func TestMultiSigPubKeyEquality(t *testing.T) {
        multisigKey2 := NewPubKeyMultisigThreshold(2, pubkeysCpy)
        require.False(t, multisigKey.Equals(multisigKey2))
 }
+func TestAddress(t *testing.T) {
+       msg := []byte{1, 2, 3, 4}
+       pubkeys, _ := generatePubKeysAndSignatures(5, msg)
+       multisigKey := NewPubKeyMultisigThreshold(2, pubkeys)
+       require.Len(t, multisigKey.Address().Bytes(), 20)
+}
 
 func generatePubKeysAndSignatures(n int, msg []byte) (pubkeys []crypto.PubKey, signatures [][]byte) {
        pubkeys = make([]crypto.PubKey, n)
```

ebuchman:
Thanks @alessio

PR #3103, opened 2019-01-10
Ensure multisig keys have 20-byte address

alessio:
Use crypto.AddressHash() to avoid returning 32-byte long address.

Closes: #3102

* [x] Wrote tests


ebuchman:
Ugh I broke it sorry. Let me fix.

Issue #1909, opened 2018-07-05
abci: use EndBlock to control empty block creation


ebuchman:
Since some apps will change the app hash every block, even with empty blocks, setting `create_empty_blocks=false` won't achieve anything.

To solve this, we could add a `Pause bool` field to ResponseEndBlock that tells Tendermint whether to make a new block right away or not. 

We should write a proper ADR before making this change. Note that any state transitions that occur in the last block committed will not have light-client proofs until the next block is committed, so apps need to be careful when setting `Pause=true`

b00f:
It's good idea giving the application this privilege to decide when pausing block creation.

ebuchman:
Another thing we could consider:

We could remove the dependence on AppHash in the meantime, and only make blocks if there are transactions (or evidence). It means we won't have proofs for time-dependent transitions in the empty proofBlock, but perhaps that is worth it to eliminate the confusion in this feature. We'd still need to make an empty proofBlock if the last block has transactions so we can get proofs for the new state resulting from the previous block.

jdkanani:
@ebuchman Can we move "AppHash change" in EndBlock or StateCommit (introduce new one) hook and assign AppHash in current block only before commit (like validator set)? I am not sure any implications it might have.

Or atleast have a configuration flag to disable support block proof?

b00f:
how about processing but not commiting empty block? I mean every node agree that empty node won't change the state... Just an idea...

KrzysiekJ:
Note that an application may want to punish absent validators and therefore application state may potentially change in every block. Also applications may want to use block timestamps to perform some time-dependent transitions, which also implies changing hash with every block. Therefore it may be hard to imagine a non-enterprise (local) application which doesn’t potentially change hash with every block, however light clients don’t necessarily need to be interested in such changes. Not depending on `AppHash` may be a reasonable default then.

Still there may be situations where an application wants to accelerate block creation or create next block in a predictable time, for example when it announces a change in the validator set, so either introducing a `Pause` boolean or at least creating a block if there is a pending change in the validator set may be a good idea.

b00f:
So I think we agree that application can decide to commit an empty block or not. Imagine there is an empty block and application decide to not commit it. Then the previous round will be repeated with same validator set, note that most nodes are in consensus about it. In this case the empty block will be processed but not committed and app hash won't change (either the block height).

ebuchman:
[...]

Issue #2227, opened 2018-08-14
clist: non-deterministic test failure


ebuchman:
Happened in unrelated PR:
- https://github.com/tendermint/tendermint/pull/2205/
- https://circleci.com/gh/tendermint/tendermint/20143#tests/containers/3

```
--- FAIL: TestWaitChan (1.26s)
	clist_test.go:271: number of pushed items (99) not equal to number of seen items (98)
```

ValarDragon:
Same error as #2206. Strangely enough, it was on the same PR as well. Though this PR is certainly unrelated

ebuchman:
Happened again in https://circleci.com/gh/tendermint/tendermint/31639?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

```
?   	github.com/tendermint/tendermint/libs/autofile/cmd	[no test files]
Pushed 0K elements so far...
stopped
stopped
stopped
Scanner 6 restartCounter: 2 counter: 1183
stopped
Scanner 8 restartCounter: 1 counter: 612
stopped
stopped
Scanner 5 restartCounter: 2 counter: 1109
Scanner 0 restartCounter: 2 counter: 1967
stopped
stopped
Scanner 7 restartCounter: 1 counter: 1003
Scanner 3 restartCounter: 1 counter: 1027
Scanner 1 restartCounter: 2 counter: 1556
stopped
stopped
Scanner 2 restartCounter: 2 counter: 1409
Scanner 9 restartCounter: 2 counter: 1275
Scanner 4 restartCounter: 2 counter: 1321
--- FAIL: TestWaitChan (1.33s)
	clist_test.go:284: number of pushed items (99) not equal to number of seen items (98)
FAIL
coverage: 64.6% of statements
FAIL	github.com/tendermint/tendermint/libs/clist	1.378s
Exited with code 1
```

KrzysiekJ:
Apparently again in https://circleci.com/gh/tendermint/tendermint/40785?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link .

```
Pushed 0K elements so far...
stopped
Scanner 6 restartCounter: 1 counter: 419
stopped
Scanner 0 restartCounter: 1 counter: 757
stopped
stopped
Scanner 5 restartCounter: 1 counter: 455
stopped
Scanner 2 restartCounter: 1 counter: 566
stopped
stopped
Scanner 4 restartCounter: 1 counter: 411
stopped
[...]

PR #3118, opened 2019-01-13
Do not force block creation when app hash changes

KrzysiekJ:
For rationale of why this may be a reasonable default, see the following comments in #1909:

* https://github.com/tendermint/tendermint/issues/1909#issuecomment-433934284
* https://github.com/tendermint/tendermint/issues/1909#issuecomment-437585377

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -15,6 +15,7 @@ Special thanks to external contributors on this release:
 * Blockchain Protocol
 
 * P2P Protocol
+- [consensus] Do not force block creation when app hash changes.
 
 ### FEATURES:
 
--- a/consensus/mempool_test.go
+++ b/consensus/mempool_test.go
@@ -37,7 +37,6 @@ func TestMempoolNoProgressUntilTxsAvailable(t *testing.T) {
 	ensureNoNewEventOnChannel(newBlockCh)
 	deliverTxsRange(cs, 0, 1)
 	ensureNewEventOnChannel(newBlockCh) // commit txs
-	ensureNewEventOnChannel(newBlockCh) // commit updated app hash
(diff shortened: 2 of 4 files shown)


KrzysiekJ:
I’ve rebased the PR agains develop, CI checks have now passed (previous failure was non-deterministic and unrelated to this PR, see #2227).

ebuchman:
So the key part from  https://github.com/tendermint/tendermint/issues/1909#issuecomment-433934284 that is missing here is `We'd still need to make an empty proofBlock if the last block has transactions so we can get proofs for the new state resulting from the previous block.`

In any case we should probably do this right by moving the control of block production into EndBlock. Of course that's a bit more involved, and should start with a spec to outline exactly how it should be done. ADR-036 has been reserved for this if you want to take a stab at it.

In the meantime, we should probably close this. Thanks.

KrzysiekJ:
This pull request only removes creating empty blocks when app hash changes, not when there are transactions, so I don’t see anything missing from the comment. Maybe there is an error in the documentation? It says that “[…] blocks will be created when there are new transactions or when the AppHash changes”. According to this, if there is a transaction that doesn’t change app hash, a block still will be created, so there should exist a mechanism that enforces block creation when there are transactions.

ebuchman:
The problem is that if there are txs in block H, we need to create block H+1, even if block H+1 is empty, because block H+1 contains the AppHash for the result of the tx included in block H. If we don't create block H+1, a light client will not be able to verify the result of the tx included in block H.

The simplest way to achieve this "make a block *after* the block with txs" idea was to rely on the changing app hash, but that of course has the side effect that it will keep making blocks if the app hash keeps changing. 

So we could consider a way to only make one empty block after a block with txs. That actually might not be very difficult, but would technically be a breaking change to the current semantics (though it's not clear the current semantics actually benefits anyone!)

Issue #3115, opened 2019-01-13
privval: test should include unix sockets


ebuchman:
Current tests in the `privval` pkg primarily use the tcp listener/dialer. They should be written to also test the unix listener/dialer. This is similar to how we test both URI and JSONRPC behaviour in `rpc` package.

Both tcp and unix sockets are integrations tested in the `node` pkg.

Follow up to #2923

_Originally posted by @ebuchman in https://github.com/tendermint/tendermint/pull/2923#issuecomment-453812717_

thanethomson:
How much detail/integration is required for the socket testing? Right now we only effectively test that we can connect - do we also need to test more detailed messaging/communication over the TCP and IPC sockets?

ebuchman:
The tests in `privval/socket_test.go` are just about connecting and sending a basic message, but we should make sure they're replicated for the unix sockets (shouldn't need to duplicate the test functions themselves, just modify to loop over the two transports).

The tests in `privval/client_test.go` test much more of the functionality, but again only use tcp sockets. We should change all those funcs to be like `testSocketXYZ(t *testing.T, sv *SocketVal, rs *RemoteSigner, xyz ...)` so that they take already setup versions of the SocketVal and RemoteSigner using whichever transport. That way the `TestSocketXYZ` funcs can just be simple loops over the two transports that call `testSocketXYZ` with the appropriate one.

ebuchman:
Eg. the rpc tests loop over the two kinds of rpc connections: https://github.com/tendermint/tendermint/blob/develop/rpc/client/rpc_test.go#L53

Liamsi:
`privval/socket_test.go` were updated via #3121. Should I take care of the tests in `privval/client_test.go` or is did someone already start working on this?

thanethomson:
@Liamsi, I'm having a little trouble at the moment with getting the Unix sockets to work with the `client_test.go` tests. If it's urgent and you've got time, you're welcome to work on this now. Otherwise I'll have to take a look at it in the morning.

Liamsi:
I do not think it is so urgent that I can't wait till tomorrow morning. Also, sounds like you've already started working on this (and I'll probably run into the issues as you). Let me know if I can help with anything though!

thanethomson:
I'm looking into this now 👍

ebuchman:
Thanks @thanethomson and @Liamsi !

PR #3121, opened 2019-01-14
Adds tests for Unix sockets

thanethomson:
As per #3115, adds simple Unix socket connect/accept deadline tests in
pretty much the same way as the TCP connect/accept deadline tests work.

* [x] Wrote tests

### DIFF
--- a/privval/socket_test.go
+++ b/privval/socket_test.go
@@ -26,9 +26,9 @@ type listenerTestCase struct {
 	dialer      Dialer
 }
 
-// getTestUnixAddr will attempt to obtain a platform-independent temporary file
+// testUnixAddr will attempt to obtain a platform-independent temporary file
 // name for a Unix socket
-func getTestUnixAddr() (string, error) {
+func testUnixAddr() (string, error) {
 	f, err := ioutil.TempFile("", "tendermint-privval-test")
 	if err != nil {
 		return "", err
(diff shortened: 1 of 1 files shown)


privval/socket_test.go
```diff
@@ -82,3 +84,89 @@ func TestTCPListenerConnDeadline(t *testing.T) {
...
+	addr, err := testUnixAddr()
+	if err != nil {
+		t.Fatal(err)
+	}
+	ln, err := net.Listen("unix", addr)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	unixLn := NewUnixListener(ln)
```
melekes: both tcp and unix listeners should adhere to `net.Listener` interface. it means we can probably merge two tests into one (same thing we did with rpc clients https://github.com/tendermint/tendermint/blob/ec53ce359bb8f011e4dbb715da098bea08c32ded/rpc/client/rpc_test.go#L53). m?

thanethomson: I've combined the tests now into an iterable set of test cases. Does this work?

privval/socket_test.go
```diff
@@ -18,67 +20,114 @@ func newPrivKey() ed25519.PrivKeyEd25519 {
...
+	UnixListenerConnDeadline(connectDeadline)(unixLn)
+	return listenerTestCase{
+		description: "Unix",
+		listener:    unixLn,
+		dialer:      DialUnixFn(addr),
+	}
+}
 
-		time.Sleep(2 * time.Millisecond)
+func constructListenerTestCases(t *testing.T, acceptDeadline, connectDeadline time.Duration) []listenerTestCase {
```
melekes:
```suggestion
func listenerTestCases(t *testing.T, acceptDeadline, connectDeadline time.Duration) []listenerTestCase {
```

I'd drop `construct` prefix

privval/socket_test.go
melekes: Yep, thanks

Issue #3082, opened 2019-01-04
Add Prometheus "chain_id" label for all metrics


mdyring:
Since people are running multiple chains (game-of-stakes, genki etc.), it would be very helpful to have the chain id as a label to differentiate between metrics of unrelated chains. Ideally this should be set on all metrics.

This could be worked around by operators when registering the targets (adding the chain_id there manually or via automation), but I believe it would be generally useful to have this built in. This also makes upgrades nicer.

ebuchman:
Added in #3123 - should be released in v0.29.0

PR #3123, opened 2019-01-14
add "chain_id" label for all metrics

melekes:
Refs #3082

* [ ] ~~Wrote tests~~ tested manually

### DIFF
--- a/consensus/metrics.go
+++ b/consensus/metrics.go
@@ -8,7 +8,11 @@ import (
 	stdprometheus "github.com/prometheus/client_golang/prometheus"
 )
 
-const MetricsSubsystem = "consensus"
+const (
+	// MetricsSubsystem is a subsystem shared by all metrics exposed by this
+	// package.
+	MetricsSubsystem = "consensus"
+)
 
 // Metrics contains metrics exposed by this package.
 type Metrics struct {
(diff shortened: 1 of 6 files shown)


consensus/metrics.go
```diff
@@ -50,101 +54,107 @@ type Metrics struct {
 }
 
 // PrometheusMetrics returns Metrics build using Prometheus client library.
-func PrometheusMetrics(namespace string) *Metrics {
+// Optionally, labels can be provided along with their values ("foo",
+// "fooValue").
+func PrometheusMetrics(namespace string, labelValues ...string) *Metrics {
+	labels := []string{}
+	for i := 0; i < len(labelValues)/2; i += 2 {
```
ebuchman: I don't think this is right - we're looping over half the length and incrementing i by 2 each time ... this won't catch everything. We need to either loop over the whole length and increment by 2 each time, or loop over half and multiple the counter by 2 when indexing ...

consensus/metrics.go
```diff
@@ -50,101 +54,107 @@ type Metrics struct {
...
+		labels = append(labels, labelValues[i])
+	}
 	return &Metrics{
 		Height: prometheus.NewGaugeFrom(stdprometheus.GaugeOpts{
 			Namespace: namespace,
 			Subsystem: MetricsSubsystem,
 			Name:      "height",
 			Help:      "Height of the chain.",
-		}, []string{}),
+		}, labels).With(labelValues...),
```
ebuchman: How do these labels work? We pass in just the values as an argument to NewGaugeFrom and then the whole list of keys and values into .With?

melekes: Yes. See the example here https://github.com/go-kit/kit/blob/1e404df823eb187378b081f22f00e4e38f05a8bd/metrics/prometheus/prometheus_test.go#L32-L37

consensus/metrics.go
melekes: Duh :( Thanks for catching this

Issue #3099, opened 2019-01-10
privval: add a MAC for unix sockets


ebuchman:
See previous issue for discussion, which should also be more permanently documented: https://github.com/tendermint/tendermint/issues/2563

IPCVal is an unencrypted local unix socket connection. Encryption+key-management is overkill for this connection. Just need to use a MAC

marbar3778:
With the migration to gRPC for all connection types, TLS is now available for all connection types. The user will have to provide a certificate. Are we okay with this or would like to add MAC on top?

cc @hendrikhofstadt @ebuchman

hendrikhofstadt:
TLS sounds solid but probably only acts as defense in depth measure here. I assume that the access to these services should never be exposed. As long as that's documented client certificates / auth are not necessary imo,

marbar3778:
okay will add documentation with the refactor to gRPC, closing this issue. reopen if anyone feels the nessecity for more secruity.

PR #3132, opened 2019-01-16
Expanding tests to cover Unix sockets version of client

thanethomson:
This PR attempts to complete #3115, where these commits add code (where relevant) to test the client over Unix domain sockets. Two tests in particular (now named `TestSocketPVDeadlineTCPOnly` and `TestRemoteSignerRetryTCPOnly`) are only relevant to the TCP version of the client (the OS knows both sides of a Unix domain socket instantaneously, and so these tests are not relevant to such connections).

* [x] Wrote tests

### DIFF
--- a/privval/client_test.go
+++ b/privval/client_test.go
@@ -278,47 +278,51 @@ func TestRemoteSignerRetryTCPOnly(t *testing.T) {
 
 func TestRemoteSignVoteErrors(t *testing.T) {
 	for _, tc := range socketTestCases(t) {
-		var (
-			chainID = cmn.RandStr(12)
-			sc, rs  = testSetupSocketPair(t, chainID, types.NewErroringMockPV(), tc.addr, tc.dialer)
+		func() {
+			var (
+				chainID = cmn.RandStr(12)
+				sc, rs  = testSetupSocketPair(t, chainID, types.NewErroringMockPV(), tc.addr, tc.dialer)
 
-			ts    = time.Now()
-			vType = types.PrecommitType
-			vote  = &types.Vote{Timestamp: ts, Type: vType}
-		)
-		defer sc.Stop()
-		defer rs.Stop()
+				ts    = time.Now()
+				vType = types.PrecommitType
+				vote  = &types.Vote{Timestamp: ts, Type: vType}
+			)
+			defer sc.Stop()
+			defer rs.Stop()
 
-		err := sc.SignVote("", vote)
-		require.Equal(t, err.(*RemoteSignerError).Description, types.ErroringMockPVErr.Error())
+			err := sc.SignVote("", vote)
+			require.Equal(t, err.(*RemoteSignerError).Description, types.ErroringMockPVErr.Error())
 
-		err = rs.privVal.SignVote(chainID, vote)
-		require.Error(t, err)
-		err = sc.SignVote(chainID, vote)
-		require.Error(t, err)
+			err = rs.privVal.SignVote(chainID, vote)
+			require.Error(t, err)
+			err = sc.SignVote(chainID, vote)
+			require.Error(t, err)
+		}()
 	}
 }
 
 func TestRemoteSignProposalErrors(t *testing.T) {
 	for _, tc := range socketTestCases(t) {
-		var (
-			chainID = cmn.RandStr(12)
-			sc, rs  = testSetupSocketPair(t, chainID, types.NewErroringMockPV(), tc.addr, tc.dialer)
+		func() {
+			var (
+				chainID = cmn.RandStr(12)
+				sc, rs  = testSetupSocketPair(t, chainID, types.NewErroringMockPV(), tc.addr, tc.dialer)
 
-			ts       = time.Now()
-			proposal = &types.Proposal{Timestamp: ts}
-		)
-		defer sc.Stop()
-		defer rs.Stop()
+				ts       = time.Now()
+				proposal = &types.Proposal{Timestamp: ts}
+			)
+			defer sc.Stop()
+			defer rs.Stop()
 
-		err := sc.SignProposal("", proposal)
-		require.Equal(t, err.(*RemoteSignerError).Description, types.ErroringMockPVErr.Error())
+			err := sc.SignProposal("", proposal)
+			require.Equal(t, err.(*RemoteSignerError).Description, types.ErroringMockPVErr.Error())
 
-		err = rs.privVal.SignProposal(chainID, proposal)
-		require.Error(t, err)
+			err = rs.privVal.SignProposal(chainID, proposal)
+			require.Error(t, err)
 
-		err = sc.SignProposal(chainID, proposal)
-		require.Error(t, err)
+			err = sc.SignProposal(chainID, proposal)
+			require.Error(t, err)
+		}()
 	}
 }
 
--- a/privval/client_test.go
+++ b/privval/client_test.go
@@ -186,104 +186,93 @@ func TestSocketPVVoteKeepalive(t *testing.T) {
 	}
 }
 
+// This test is not relevant to Unix domain sockets, since the OS knows
+// instantaneously the state of both sides of the connection.
 func TestSocketPVDeadline(t *testing.T) {
-	for _, tc := range socketTestCases(t) {
-		// This test is not relevant to Unix domain sockets, since the OS knows
-		// instantaneously the state of both sides of the connection.
-		proto, _ := cmn.ProtocolAndAddress(tc.addr)
-		if proto != "unix" {
-			func() {
-				var (
-					listenc         = make(chan struct{})
-					thisConnTimeout = 100 * time.Millisecond
-					sc              = newSocketVal(log.TestingLogger(), tc.addr, thisConnTimeout)
-				)
+	var (
+		addr            = testFreeTCPAddr(t)
+		listenc         = make(chan struct{})
+		thisConnTimeout = 100 * time.Millisecond
+		sc              = newSocketVal(log.TestingLogger(), addr, thisConnTimeout)
+	)
 
-				go func(sc *SocketVal) {
-					defer close(listenc)
+	go func(sc *SocketVal) {
+		defer close(listenc)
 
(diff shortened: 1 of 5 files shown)


privval/client_test.go
```diff
@@ -172,7 +222,12 @@ func TestSocketPVDeadline(t *testing.T) {
 	<-listenc
 }
 
-func TestRemoteSignerRetry(t *testing.T) {
+// TestRemoteSignerRetryTCPOnly will test connection retry attempts over TCP. We
+// don't need this for Unix sockets because the OS instantly knows the state of
+// both ends of the socket connection. This basically causes the
+// RemoteSigner.dialer() call inside RemoteSigner.connect() to return
+// successfully immediately, putting an instant stop to any retry attempts.
+func TestRemoteSignerRetryTCPOnly(t *testing.T) {
```
Liamsi: On an unrelated note: This test seems misplaced in `client_test.go`. It doesn't really test any client related behaviour (ok the closing and reconnecting is sth the client might do but it is merely there to count the retries of the server). It basically spins off a server and doesn't test any client related code line.

thanethomson: You're right - this looks more relevant to the `RemoteSigner` class. Would this be better placed in a new file `privval/remote_signer_test.go`?

melekes:
Shouldn't all PRs go to develop first?

ebuchman:
> Two tests in particular (now named TestSocketPVDeadlineTCPOnly and TestRemoteSignerRetryTCPOnly) are only relevant to the TCP version of the client (the OS knows both sides of a Unix domain socket instantaneously, and so these tests are not relevant to such connections).

I don't totally understand this. Couldn't the process on the other end of the socket still be slow or crash, in which case we'd still want to enforce the deadlines and be able to retry?

thanethomson:
> I don't totally understand this. Couldn't the process on the other end of the socket still be slow or crash, in which case we'd still want to enforce the deadlines and be able to retry?

The trouble with Unix domain sockets is that, the moment you try to connect to them, the OS is instantaneously aware of the state of both ends of the connection. That is, the OS knows instantly if something is in an `accept` state on the other end.

For the `TestSocketPVDeadlineTCPOnly` test, from what I can tell, all it does is try to make a connection with a timeout (i.e. waiting for an `accept`) and then make sure that it does, in fact, time out. This works fine for a TCP socket, where something may or may not be `accept`'ing connections on the other side, and the deadline is a parameter passed down the networking stack to give a maximum wait period until giving up. With a Unix domain socket, this returns instantly.

When I did run this test with Unix sockets, all I got was an infinite loop of connection failures.

Also for `TestSocketPVDeadlineTCPOnly`, it makes a call to `p2pconn.MakeSecretConnection`. Is this not TCP-specific?

Then, for `TestRemoteSignerRetryTCPOnly`, @Liamsi and I decided to move it out to its own file (`remote_signer_test.go`) since it doesn't make any use of the client. This particular test, with TCP, the `RemoteSigner.connect()` function attempts to `dial()`, but fails to `dial()` with some small networking-related delay, which triggers a retry.

With Unix domain sockets, the `TestRemoteSignerRetryTCPOnly` test will `dial()` successfully and instantaneously, since the OS knows something's listening on the other side. This kills the [retry loop](https://github.com/tendermint/tendermint/blob/release/v0.28.0/privval/server.go#L122). Specifically, [line 127](https://github.com/tendermint/tendermint/blob/release/v0.28.0/privval/server.go#L127) returns instantaneously due to the test's `ln.Accept()` call [here](https://github.com/tendermint/tendermint/blob/release/v0.28.0/privval/client_test.go#L188). The subsequent `conn.Close()` call kicks the `RemoteSigner` into a different part of its lifecycle, which stops it altogether.

As @Liamsi pointed out to me this morning, [this test](https://github.com/thanethomson/tendermint/blob/a1580f7b2b018964d87c374065053b9125e0a6b7/privval/client_test.go#L391) will then go and test the reconnection efforts made by a Unix domain socket.

The output logs for `TestRemoteSignerRetry` when testing with a TCP connection:
```
=== RUN   TestRemoteSignerRetry
I[16016-01-16|11:18:15.044] Listener started                             proto=tcp addr=127.0.0.1:49646
I[16016-01-16|11:18:15.044] Waiting to accept connection...              attempts=0
I[16016-01-16|11:18:15.044] Starting RemoteSigner                        impl=RemoteSigner
I[16016-01-16|11:18:15.044] Accepted incoming connection
I[16016-01-16|11:18:15.044] Successfully closed incoming connection
I[16016-01-16|11:18:15.044] Waiting to accept connection...              attempts=1
E[16016-01-16|11:18:15.044] dialing                                      err=EOF
I[16016-01-16|11:18:15.046] Accepted incoming connection
I[16016-01-16|11:18:15.046] Successfully closed incoming connection
E[16016-01-16|11:18:15.046] dialing                                      err="read tcp 127.0.0.1:49648->127.0.0.1:49646: read: connection reset by peer"
E[16016-01-16|11:18:15.046] OnStart                                      err="dialed maximum retries"
E[16016-01-16|11:18:15.046] Not stopping RemoteSigner -- have not been started yet impl=RemoteSigner
--- PASS: TestRemoteSignerRetry (0.00s)
```

And the output logs for `TestRemoteSignerRetry` when testing with a Unix connection:
```
=== RUN   TestRemoteSignerRetry
I[16016-01-16|11:17:34.938] Listener started                             proto=unix addr=/var/folders/27/xz26zgdx1w79mqxpy6hhybqr0000gn/T/tendermint-privval-test-102044498
I[16016-01-16|11:17:34.939] Waiting to accept connection...              attempts=0
I[16016-01-16|11:17:34.939] Starting RemoteSigner                        impl=RemoteSigner
I[16016-01-16|11:17:34.939] Accepted incoming connection
I[16016-01-16|11:17:34.939] Successfully closed incoming connection
I[16016-01-16|11:17:34.939] Waiting to accept connection...              attempts=1
I[16016-01-16|11:17:35.041] Stopping RemoteSigner                        impl=RemoteSigner
I[16016-01-16|11:17:35.041] RemoteSigner stopping
--- FAIL: TestRemoteSignerRetry (0.10s)
        Error Trace:    client_test.go:278
                        client_test.go:286
        Error:          Not equal:
                        expected: *errors.fundamental(dialed maximum retries)
                        actual  : <nil>(<nil>)
        Test:           TestRemoteSignerRetry
    client_test.go:284: expected remote to observe connection attempts
```

Note the test log line for the Unix domain socket:
```
I[16016-01-16|11:17:35.041] RemoteSigner stopping
```
implies it was completely started and is entering the Stop phase of its lifecycle, where the same doesn't happen for the TCP connection above:
```
E[16016-01-16|11:18:15.046] Not stopping RemoteSigner -- have not been started yet 
```

Does that perhaps clarify a bit more?

ebuchman:
Thanks for the exquisite detail!

> Also for TestSocketPVDeadlineTCPOnly, it makes a call to p2pconn.MakeSecretConnection. Is this not TCP-specific?

We have decided for the priv validator that we'd encrypt the TCP connections and not the Unix connections. So for this use case, it's TCP specific, but afaik there's no fundamental reason you can't wrap a unix socket in a secret connection. Since unix sockets are on the same machine, the encryption is unecessary, but we plan to at least use a MAC to assert data integrity (eg. we'd wrap it with MakeMACConnection or something) - #3099

As for the unix socket retries - what will happen when a KMS connected by unix socket crashes and has to restart, or suffers a significant delay in responding that causes Tendermint to close the connection? Will it be able to successfully reconnect? How can we test for that?

thanethomson:
> We have decided for the priv validator that we'd encrypt the TCP connections and not the Unix connections. So for this use case, it's TCP specific, but afaik there's no fundamental reason you can't wrap a unix socket in a secret connection. Since unix sockets are on the same machine, the encryption is unecessary, but we plan to at least use a MAC to assert data integrity (eg. we'd wrap it with MakeMACConnection or something) - #3099 

Let me see what I can dig up to find a meaningful test for Unix sockets, along these lines, if there is such a meaningful test (e.g. if there's a delay between the `listen` and the `accept` portion of the Unix domain socket server's operations - I have a feeling though that if nothing's `accept`'ing on the server side, you'll get an instant failure, but I'll test this out).

> As for the unix socket retries - what will happen when a KMS connected by unix socket crashes and has to restart, or suffers a significant delay in responding that causes Tendermint to close the connection? Will it be able to successfully reconnect? How can we test for that?

The same thing will happen as if a TCP connection successfully completes, is operational for some time, but then dies and then later comes back. This is already tested in [`TestRetryConnToRemoteSigner`](https://github.com/tendermint/tendermint/blob/master/privval/client_test.go#L330).

Here are the output logs for this test for the Unix domain sockets (note the `Re-created connection to remote signer` line from the `SocketVal` side, which is triggered by a [`reset`](https://github.com/tendermint/tendermint/blob/master/privval/client.go#L157) call when the heartbeat fails):

```
I[17016-01-17|06:06:32.038] Listening at                                 proto=unix address=/var/folders/27/xz26zgdx1w79mqxpy6hhybqr0000gn/T/tendermint-privval-test-439157311
I[17016-01-17|06:06:32.038] Starting RemoteSigner                        impl=RemoteSigner
I[17016-01-17|06:06:32.038] Starting SocketVal                           impl=SocketVal
I[17016-01-17|06:06:32.059] Stopping RemoteSigner                        impl=RemoteSigner
E[17016-01-17|06:06:32.060] handleConnection writeMsg                    err="remote signer timed out"
E[17016-01-17|06:06:32.060] Ping                                         err=EOF
I[17016-01-17|06:06:32.067] Starting RemoteSigner                        impl=RemoteSigner
I[17016-01-17|06:06:32.067] Re-created connection to remote signer       impl=SocketVal
I[17016-01-17|06:06:32.268] Stopping RemoteSigner                        impl=RemoteSigner
E[17016-01-17|06:06:32.269] handleConnection readMsg                     err="remote signer timed out"
I[17016-01-17|06:06:32.269] Stopping SocketVal                           impl=SocketVal
E[17016-01-17|06:06:32.269] Ping                                         err="remote signer timed out"
--- PASS: TestRetryConnToRemoteSigner (0.47s)
```

Issue #3120, opened 2019-01-13
merkle: follow up to #2713


ebuchman:
See the review for points of follow up: https://github.com/tendermint/tendermint/pull/2713#pullrequestreview-180548151

The main concern is that `Verify` now seems to expect an object before it is hashed, so it can prepend the 0-prefix ala the rfc.

However, the leaves of our Merkle trees should always be hashed anyways, even before the 0-prefix is added. @ValarDragon see https://github.com/tendermint/tendermint/pull/2713#discussion_r238086049

Few other minor points mentioned in the review.

We're also going to need to update the SDK and possible fix some things in the `store` pkg, @jackzampolin

jackzampolin:
@ebuchman thank you for the heads up. Are some of the required `store` changes included in or facilitated by https://github.com/cosmos/cosmos-sdk/pull/2985?

ebuchman:
Wow that looks like it's going to be a great cleanup but I don't think it will really help here.

Consequences of this for the SDK:
- breaks the way the stores are hashed, so any hard coded merkle roots will be invalid
- changes the interface for merkle proofs by now taking in the full leaf object being proven, rather than the hash of the object

@ValarDragon did I miss anything? 

Also, @ValarDragon, can you please address https://github.com/tendermint/tendermint/pull/2713#discussion_r238085473 with a small comment to explain what's going on :) . I can handle the rest of the review comments there.

I think it should be ok to leave the new API as is - ie. the proofs take the leaf object, not the hash. If some consumer wants to pass in the hash instead, it's welcome to, like Tendermint does for txs. We just need to be clear in the documentation/spec where we pass in a hash and where we pass the whole object!

ValarDragon:
I believe thats all the changes that need to be made SDK side. Also agreed for leaving the API as is!

I'll make a PR for that comment

ebuchman:
Fixed in #3175

PR #3136, opened 2019-01-16
Add comment to simple_merkle get_split_point

ValarDragon:
Ref #3120


crypto/merkle/simple_tree.go
```diff
@@ -32,6 +32,7 @@ func SimpleHashFromMap(m map[string][]byte) []byte {
 	return sm.Hash()
 }
 
+// getSplitPoint returns the largest power of 2 less than length
```
Liamsi: Not sure if we should refer to the spec here too? https://tools.ietf.org/html/rfc6962#section-2.1.2

PR #3140, opened 2019-01-17
Formalize proposer election algorithm properties 

ancazamfir:
Ref #1558, #2890


docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,187 @@
...
+#### R1: Determinism
+Given a validator set `V`, and two honest validators `p` and `q`, for each height `h` and each round `r` the following must hold:
+
+  `proposer_p(h,r) = proposer_q(h,r)`
+
+where `proposer_p(h,r)` is the proposer returned by the Proposer Selection Procedure at process `p`, at height `h` and round `r`.
+
+#### R2: Liveness
+In every consecutive sequence of rounds of size K (K is system parameter), at least a
+single round has an honest proposer (??)
```
milosevic: Yep, this definition is fishy. Probably it should expect that during period of stability (with no validator set changes), every process should propose during some sequence of rounds (how big this sequence is).

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,187 @@
...
+
+#### R3: Fairness
+In a stable network (no validator set changes) and for any sequence of K rounds with K > P, a validator v is elected as proposer with a frequency f proportional to its voting power VP(v) divided by the total voting power P.
+f ~ VP(v) / P
+
+#### R4: Starvation Prevention ;)
```
milosevic: Good try :) Let's try to sharpen it a bit. I guess we also want to somehow address the stability aspect, i.e., be able to specify slow start for newly added validators.

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,187 @@
...
+The fairness of the proposer selection is ensured by assigning priorties to each validator based on the voting power and updating them at each round. In addition, everytime a validator is elected as proposer its priority is decreased with the total voting power.
+
+### Basic Algorithm
+
+This section includes an overview of the proposer selection algorithm. 
+
+A model that gives a good intuition on how/ why the selection algorithm works and it is fair is that of a priority queue. The validators move ahead in this queue according to their voting power (the higher the voting power the faster a validator moves towards the head of the queue). At each round the following happens:
+- all validators move "ahead" according to their powers: for each validator, increase the priority by the voting power
+- first in the queue becomes the proposer: select the validator with highest priority 
+- move the proposer to the end of the queue: decrease the proposer's priority by the total voting power
```
milosevic: Does it really always go to the end of the queue? For example, in cases we have consecutive proposals.

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,186 @@
...
+where `proposer_p(h,r)` is the proposer returned by the Proposer Selection Procedure at process `p`, at height `h` and round `r`.
+
+#### R2: Liveliness
+In every consecutive sequence of rounds of size K (K is system parameter), at least a
+single round has an honest proposer (??)
+
+#### R3: Fairness
+In a stable network (no validator set changes) and for any sequence of K rounds with K > P, a validator v is elected as proposer with a frequency f proportional to its voting power VP(v) divided by the total voting power P:
+
+    f ~ VP(v) / P
```
Liamsi: Question regarding the notation: Instead of `f`, should we call the frequency `f ᵥ ` , `f_v`, or `f(v)` to indicate it is the frequency of validator `v`?

ancazamfir: yes, fixed locally.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: But isn't this under Fairness then? I had a feeling it was added there for another reason, you mentioned 2/3 consensus. The consensus aspect is not covered in this spec. Should it? Or add reference?

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,187 @@
...
+
+|Priority Round |-3 | -2 | -1 | 0  | 1   | 2   | 3 | 4 | 5 | Comment
+|---------------|-- | ---|--- |--- |---  |---  |---|---|---|--------
+| R end         |p3 |    |    |    | p1  | p2  |   |   |   |remove p2
+| 1             |   |    |    | p3 |     | p1  |   |   |   |A(i)+=VP(i)
+|               |   | p1 |    | p3 |     |     |   |   |   |A(p1)-= N
+| *new step*    |   |    | p1 |    | p3  |     |   |   |   |A(i) -= avg
+
+The procedure could continue without modifications. However, it is possible that after a sufficiently large number of modifications in validator set, the priority values would group towards maximum or minimum allowed values.
+For this reasons, the selection procedure adds another step (see last row) that shifts the current priority values left or right such that the average remains 0.
```
milosevic: Isn't the sum 0, and not average?

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,187 @@
...
+ 
+
+|Priority Round |-3 | -2 | -1 | 0  | 1   | 2   | 3 | 4 | 5 | Comment
+|---------------|-- | ---|--- |--- |---  |---  |---|---|---|--------
+| R end         |p3 |    |    |    | p1  | p2  |   |   |   |remove p2
+| 1             |   |    |    | p3 |     | p1  |   |   |   |A(i)+=VP(i)
+|               |   | p1 |    | p3 |     |     |   |   |   |A(p1)-= N
+| *new step*    |   |    | p1 |    | p3  |     |   |   |   |A(i) -= avg
+
+The procedure could continue without modifications. However, it is possible that after a sufficiently large number of modifications in validator set, the priority values would group towards maximum or minimum allowed values.
```
milosevic: In case sum of priorities is 0 at the beginning of every height (round), does this ensure fairness during period of stability?

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,187 @@
...
+For this reasons, the selection procedure adds another step (see last row) that shifts the current priority values left or right such that the average remains 0.
+
+The modified selection algorithm is:
+
+    for each validator i in vset:
+      A(i) += VP(i)
+    prop = max(A)
+    A(prop) -= N
+
+    // normalize - shift with average priority
```
milosevic: Shouldn't we normalise upon validator removal so invariant (sum is 0) holds at the beginning of next round, instead of doing it as part of algorithm (every time)?

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: On its own for sure not. But it's a nice invariant (when it holds) and might be useful to simplify a bit a formal proof.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: In the penultimate row is -2

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: Yes the code could be changed. Just documenting what goes in 0.29 for the moment. I believe that there are many ideas for changes (just reading the different comments in related issues). Also @Liamsi may have comments.

docs/spec/reactors/consensus/proposer-selection.md
ebuchman: Yeh I'm not sure this is really important here. The assumptions about +2/3 honest nodes live elsewhere. So I think we can get rid of this. Unless there really is something distinct from fairness that's valuable here.

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,228 @@
...
+  `proposer_p(h,r) = proposer_q(h,r)`
+
+where `proposer_p(h,r)` is the proposer returned by the Proposer Selection Procedure at process `p`, at height `h` and round `r`.
+
+#### R2: Liveliness
+In every consecutive sequence of rounds of size K (K is system parameter), at least a
+single round has an honest proposer (??)
+
+#### R3: Fairness
+In a stable network (no validator set changes) and for any sequence of K rounds with K > P, a validator v is elected as proposer with a frequency f proportional to its voting power VP(v) divided by the total voting power P:
```
ebuchman:
```suggestion
In a stable network (no validator set changes) with total voting power P and for any sequence of K rounds with K > P, a validator v is elected as proposer with a frequency f proportional to its voting power VP(v) divided by P:
```

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,228 @@
...
+
+    f(v) ~ VP(v) / P
+
+#### O4: Starvation Prevention ;)
+In a churning network with many validator set changes, minimal guarantees for a chance in future election should be offered, e.g. for stable lower power validators.
```
ebuchman: This is basically saying we don't want changing the validator set to allow gaming the proposer selection, right? Why is that `Starvation Prevention`  and what's the relevance of `stable lower power validators`?

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,228 @@
...
+
+## Proposer Selection Specification
+### Requirement Fulfillment Claims
+__[R1]__ The proposer algorithm is deterministic giving consistent results across executions with same transactions and validator set modifications. A pseudocode of the alorithm is given below.
```
ebuchman:
```suggestion
__[R1]__ The proposer algorithm is deterministic giving consistent results across executions with same transactions and validator set modifications. A pseudocode of the algorithm is given below. 
```

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,228 @@
...
+
+__[R3]__ Given a set of processes with the total voting power P, during a sequence of rounds of length P, every process is selected as proposer in a number of rounds equal to its voting power. The sequence of the P proposers then repeats.
+If we consider the validator set:
+
+Validator | p1 | p2 
+----------|--- | ---
+VP        | 1  | 3
+
+The proposer selection generates the sequence:
+`p2, p1, p2, p2, p2, p1, p2, p2,...` or [`p2, p1, p2, p2`]*
```
ebuchman: Presumably this isn't the only acceptable sequence though, right? It might be what our algorithm provides, but do we know if a different algorithm could still satisfy the properties and give a different order, eg. `[p2, p2, p1, p2]` , or even `[p2, p2, p2, p1]` ?

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,228 @@
...
+
+Assigning priorities to each validator based on the voting power and updating them at each round ensures the fairness of the proposer selection. In addition, every time a validator is elected as proposer its priority is decreased with the total voting power.
+
+__[O4]__ There is currently no definite guarantee that a low power stable validator will be elected within a large enough round window (??)
+Note: Just a claim here. If new high power validators bound and unbound continuously, the low validator moves back and forward in the queue via shifting. Since there is no "aging" that would cause a steady increase in its voting power, or similar mechanism, it is possible that this validator election is postponed indefinitely.
```
ebuchman: Yikes!

docs/spec/reactors/consensus/proposer-selection.md
ebuchman: Sometimes the same proposer proposes multiple times in a row, which means they're not necessarily pushed to the end of the queue after proposing, even if there's no validator changes.

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,228 @@
...
+    
+    def proposalSelection (vset, times):
+        for in in range(times):
+            singleProposalSelection(vset)
+
+### Validator Set Changes
+At each block height the validator set may change. Some of the changes have implications on the proposer selection.
+
+#### Voting Power Change
+Consider again the earlier example and assume that the voting power of p1 is changed to 6:
```
ebuchman:
```suggestion
Consider again the earlier example and assume that the voting power of p1 is changed to 4:
```

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,228 @@
...
+
+Validator | p1 | p2 
+----------|--- | ---
+VP        | 4  | 3
+
+Let's also assume that the last round R before this change the proposer priorites were as shown in first row (R-1 end). As it can be seen, the procedure can continue with round 1 at next height without changes as before.
```
ebuchman: `R end` or `R-1 end` ? Text conflicts with table

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,228 @@
...
+|              | p1 |    |      |     | p2  |   |   |   |A(p1)-= P
+
+#### Validator Removal
+Consider a new example with set.
+
+Validator | p1 | p2 | p3 |
+--------- |--- |--- |--- |
+VP        | 1  | 2  | 3  |
+
+Let's assume that the last round was R and the proposer priorities were as shown in first row (R end) with their sum being 0. At this point p2 is removed and round 1 of next height runs. At the end of the round (penultimate row) the sum of priorities is -2 (minus the priority of the removed process).
```
ebuchman: Minor point - our rounds start at 0, so the first round of a new height is round 0, not round 1.

(But note our block heights start at 1, so the first block in the chain is block 1 and there is no 0 block)

docs/spec/reactors/consensus/proposer-selection.md
ebuchman: Why not? Isn't that just `simpleProposalSelection`, which is fair ?

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,228 @@
...
+ 
+
+|Priority Round |-3 | -2 | -1 | 0  | 1   | 2   | 3 | 4 | 5 | Comment
+|---------------|-- | ---|--- |--- |---  |---  |---|---|---|--------
+| R end         |p3 |    |    |    | p1  | p2  |   |   |   |remove p2
+| 1             |   |    |    | p3 |     | p1  |   |   |   |A(i)+=VP(i)
+|               |   | p1 |    | p3 |     |     |   |   |   |A(p1)-= P
+| *new step*    |   |    | p1 |    | p3  |     |   |   |   |A(i) -= avg
+
+The procedure could continue without modifications. However, it is possible that after a sufficiently large number of modifications in validator set, the priority values would group towards maximum or minimum allowed values.
```
ebuchman: Can we clarify why this is a problem - ie. it will lead to overflow issues if we're using standard integers. Is there some other reason?

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,228 @@
...
+        avg = sum(A(i) for i in vset)/len(vset)
+        for each validator i in vset:
+            A(i) -= avg
+
+Note that the shifting operation happens after singleProposalSelection() has run times times.
+
+#### New Validator
+When a new validator is added same problem as the one described for removal appears, the sum of priorities in the new set is not zero. This is now fixed with the shift step introduced above.
+
+One other issue that needs to be addressed is the following. A validator V with low voting power that has just been elected is moved to the end of the queue. If the validator set is large and/ or with other validators with significantly higher power, validator V will have to wait a lot of rounds to be elected. If V could remove and read itself to the validator set, it would make a significant (albeit unfair) "jump" ahead in the queue, decreasing the numbers of rounds it has to wait.
```
ebuchman:
```suggestion
One other issue that needs to be addressed is the following. A validator V with low voting power that has just been elected is moved to the end of the queue. If the validator set is large and/ or with other validators with significantly higher power, validator V will have to wait a lot of rounds to be elected. If V could remove and re-add itself to the validator set, it would make a significant (albeit unfair) "jump" ahead in the queue, decreasing the numbers of rounds it has to wait. 
```

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,228 @@
...
+    ...
+    A(V)-=P  
+
+It follows that `A(V)-=sum(VP(i) for i in {v1,..,vn})`
+
+Therefore when `V` is added to the `{v1,..,vn}` set its initial priority will be:
+
+    A(V) = -sum(VP(i) for i in {v1,..,vn})
+
+With a penalty factor of 1.125, the initial priority for a newly added verifier is:
```
ebuchman: Can we explain why we need the penalty factor and why `9/8` is reasonable?

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,228 @@
...
+    A(V) = -1.125 * P
+
+If we consider the validator set where p3 has just been added:
+
+Validator | p1 | p2 | p3
+----------|--- |--- |---
+VP        | 1  | 3  | 8
+
+Let's assume that the last round was R and the proposer priorities were as shown in first row (R end) with their sum being 0. p3 is added with A(p3) = -4
+In the next round, p1 will still be ahead in the queue, elected as proposer and move back in the queue (same position it was just added)
```
ebuchman:
```suggestion
In the next round, p3 will still be ahead in the queue, elected as proposer and move back in the queue (same position it was just added)
```

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,20 +2,228 @@
...
+|--------------- | ---|--- |----|--- |--- |---|---|---|---|--------
+|R end          |    |    | p2 |    |    |   | p1|   |   |add p3
+|                | p3 |    | p2 |    |    |   | p1|   |   |A(p3) = -4
+|1              |    |    |    |    |    | p2|   | p1| p3|A(i)+=VP(i)
+|               | p3 |    |    |    |    | p2|   | p1|   |A(p3)-=P
+
+
+### Wrinkles
+
+#### Validator Power Overflow Conditions
```
ebuchman: I thought we were normalizing the priorities? That would help prevent overflow, but if I recall correctly, it was also necessary for the @cwgoes proof

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: Removed

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: added the 'times' parameter and handling

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: changed

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: changed

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: It is not about distinguishing validator changes that are made to game the selection vs legitimate scenarios. What I am trying to say is that a stable small validator should be eventually be able to propose. Zarko also had issues with this O4 (O - means optional requirement BTW, a nice to have) and suggested maybe to remove.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: True, but at this point in the spec we document what we have. I could clarify.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: done

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: Will try my best, not sure why the 1/8th was chosen though.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: This is for the voting power.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: I think the question was if the fact that sum==0 guarantees fairness. What (we claim) makes simpleProposalSelection fair is this: prop = max(A)...that is we do take into account the priority at election. It would be fair even without ensuring sum==0.

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,268 @@
...
+  `proposer_p(h,r) = proposer_q(h,r)`
 
-Let consider now the following scenario where a total voting power of faulty processes is aggregated in a single process
-p0: (p0,3), (p1, 1), (p2, 1), (p3, 1), (p4, 1), (p5, 1), (p6, 1), (p7, 1).
-In this case the sequence of proposer selections looks like this:
+where `proposer_p(h,r)` is the proposer returned by the Proposer Selection Procedure at process `p`, at height `h` and round `r`.
 
-`p0, p1, p2, p3, p0, p4, p5, p6, p7, p0, p0, p1, p2, p3, p0, p4, p5, p6, p7, p0, etc`
+#### R2: Fairness
+In a stable network (no validator set changes) with total voting power P and for any sequence of K rounds with K > P, a validator v is elected as proposer with a frequency f proportional to its voting power VP(v) divided by P:
```
milosevic: I am not sure this definition is correct as it does not hold for any K. In case you pick K = 1.5P, then frequency is not proportional to voting power. Probably we need to stick with K=P, or we can use sth a bit more complex where K would be P/greatest-common-divisor of voting power of every process.

docs/spec/reactors/consensus/proposer-selection.md
milosevic: Maybe it can be defined as follows: "If a validator is persistently part of the validator set (in a churning network with many validator set changes), he will eventually be selected as a proposer."

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,268 @@
...
 
-`p0, p1, p2, p3, p4, p5, p6, p7, p8, p9, p0, p1, p2, p3, p4, p5, p6, p7, p8, p9, etc`
+#### O3: Starvation Prevention
+In a churning network with many validator set changes, minimal guarantees for a chance in future election should be offered, e.g. for stable lower power validators.
 
-In this case, we have 3 consecutive rounds with a faulty proposer.
-One special case we consider is the case where a single honest process p0 has most of the voting power, for example:
-(p0,100), (p1, 2), (p2, 3), (p3, 4). Then the sequence of proposer selection looks like this:
+## Proposer Selection Specification
+### Requirement Fulfillment Claims
```
milosevic: Maybe this can go after the algorithm is presented.

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,268 @@
...
 
-This basically means that almost all rounds have the same proposer. But in this case, the process p0 has anyway enough
-voting power to decide whatever he wants, so the fact that he coordinates almost all rounds seems correct.
+Validator | p1| p2 
+----------|---|---
+VP        | 1 | 3
+
+The current implementation of proposer selection generates the sequence:
+`p2, p1, p2, p2, p2, p1, p2, p2,...` or [`p2, p1, p2, p2`]*
+A a sequence that starts with any circular permutation of the [`p2, p1, p2, p2`] sub-sequence would also provide the same degree of fairness. In fact these circular permutations show in the sliding window (over the generated sequence) of size equal to the length of the sub-sequence.
```
milosevic: Typo, remove one a.

milosevic: I find this paragraph a bit confusing. Maybe instead: "A sequence with any circular permutation of the [`p2, p1, p2, p2`] sub-sequence would also provide the same degree of fairness."

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: True, not well written, but I was thinking of a stronger req... what about something like:
Given a stable network (no validator set changes) with total voting power P and  a sequence S of K >= P rounds. In any sub-sequence of S with length P a validator v is elected as proposer with a frequency f proportional to its voting power VP(v) divided by P:
    f(v) ~ VP(v) / P

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,268 @@
...
+|2               |   |   |     |   |p1,p2|   |   |   |A(i)+=VP(i)
+|                | p1|   |     |   |   p2|   |   |   |A(p1)-= P
+|3               |   | p1|     |   |     |   |   | p2|A(i)+=VP(i)
+|                |   | p1|     | p2|     |   |   |   |A(p2)-= P
+|4               |   |   |   p1|   |     |   | p2|   |A(i)+=VP(i)
+|                |   |   |p1,p2|   |     |   |   |   |A(p2)-= P
+
+The actual implementation does not create this queue, it only has to store the current accumulated proposer priority. 
+
+There are cases where the proposer selection is executed multiple times.
```
milosevic: I am not sure that this is important for the purpose of this spec. I would suggest to remove the discussion and pseudocode about multiple proposer selection as it trivially follows from the single proposer selection.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: changed

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,268 @@
...
+----------|---|---
+VP        | 1 | 3
+
+The current implementation of proposer selection generates the sequence:
+`p2, p1, p2, p2, p2, p1, p2, p2,...` or [`p2, p1, p2, p2`]*
+A a sequence that starts with any circular permutation of the [`p2, p1, p2, p2`] sub-sequence would also provide the same degree of fairness. In fact these circular permutations show in the sliding window (over the generated sequence) of size equal to the length of the sub-sequence.
+
+Assigning priorities to each validator based on the voting power and updating them at each round ensures the fairness of the proposer selection. In addition, every time a validator is elected as proposer its priority is decreased with the total voting power.
+
+__[O3]__ There is currently no definite guarantee that a low power stable validator will be elected within a large enough round window (??)
```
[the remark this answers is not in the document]
ancazamfir: removed

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,268 @@
...
+
+#### Voting Power Change
+Consider again the earlier example and assume that the voting power of p1 is changed to 4:
+
+
+Validator | p1| p2 
+----------|---| ---
+VP        | 4 | 3
+
+Let's also assume that at the last round R at height H before this change the proposer priorites were as shown in first row (H/R end). As it can be seen, the procedure could continue with round 1 at next height without changes as before.
```
milosevic: Mentioning the state of the last round at some height is misleading as in the current implementation proposer selections within some height H does not affect selection at height H+1. For a given height H, initial validator set is defined and then for each round priorities are modified. For height H+1, the only thing that matters is the initial validator set of the previous height (H) and potential validator set changes at the end of height H. Between heights H and H+1 we apply the same algorithm for proposer selection that modifies priorities and select the initial proposer of round 0.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: Ok, so should i also remove the outer loop that calls 'times' times the simpleProposerSelection? And collapse the pseudocode in a single function?

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,268 @@
...
+
+Let's also assume that at the last round R at height H before this change the proposer priorites were as shown in first row (H/R end). As it can be seen, the procedure could continue with round 1 at next height without changes as before. 
+
+|Round Priority| -2 | -1 | 0    | 1   | 2   | 3 | 4 | 5 | Comment
+|--------------| ---|--- |------|---  |---  |---|---|---|--------
+| H/R end      |    | p2 |      |  p1 |     |   |   |   |update VP(p1)
+| H+1/0        |    |    |      |     | p2  |   |   | p1|A(i)+=VP(i)
+|              | p1 |    |      |     | p2  |   |   |   |A(p1)-= P
+
+However, when a validator changes power from a high to a low value it may be stuck at the end of the queue for a long time if it happened to be there (e.g. it just proposed). This scenario is considered again after the New Validator section.
```
milosevic: Why is this important? He will eventually be selected in case period of stability is long enough?

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,268 @@
...
+
+#### Validator Removal
+Consider a new example with set.
+
+Validator | p1 | p2 | p3 |
+--------- |--- |--- |--- |
+VP        | 1  | 2  | 3  |
+
+Let's assume that the last round at height H was R and the proposer priorities were as shown in first row (H/R end) with their sum being 0. At this point p2 is removed and round 1 of next height H+1 runs. At the end of the round (penultimate row) the sum of priorities is -2 (minus the priority of the removed process).
```
milosevic: The height is started with round 0.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: But the priorities are remembered between blocks, no? We don't reset them AFAIK

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,268 @@
...
+|Round Priority |-3 | -2 | -1 | 0  | 1   | 2   | 3 | 4 | 5 | Comment
+|---------------|-- | ---|--- |--- |---  |---  |---|---|---|--------
+| H/R end       |p3 |    |    |    | p1  | p2  |   |   |   |remove p2
+| H+1/0         |   |    |    | p3 |     | p1  |   |   |   |A(i)+=VP(i)
+|               |   | p1 |    | p3 |     |     |   |   |   |A(p1)-= P
+| *new step*    |   |    | p1 |    | p3  |     |   |   |   |A(i) -= avg
+
+The procedure could continue without modifications. However, it is possible that after a sufficiently large number of modifications in validator set, the priority values would migrate towards maximum or minimum allowed values causing truncations due to overflow detection.
+For this reason, the selection procedure adds another step (see last row) that shifts the current priority values left or right such that the average remains close to 0. Due to integer division the priority sum is in {-1, 0, 1}
+
```
milosevic: . is missing at the end of the previous sentence.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: true but the problem may become severe, it is exactly issue described in https://github.com/tendermint/tendermint/issues/2960

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: fixed

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: fixed.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: fixed

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: removed for now

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,268 @@
...
+    def ProposerSelection (vset, times):
+        for in in range(times):
+            singleProposerSelection(vset)
+
+        // shift priorities with the average
+        avg = sum(A(i) for i in vset)/len(vset)
+        for each validator i in vset:
+            A(i) -= avg
+
+Note that the shifting operation happens after singleProposerSelection() has run times times.
```
milosevic: As already discussed, the shifting operation should probably happen only after validator set is modified so that sum of priorities is always around 0 before round is started. This can be addressed together with the corresponding code modifications is a separate PR.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: ok. I stored this, but now this is removed

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,268 @@
...
+    ...
+    A(V)-=P  
+
+It follows that `A(V)-=sum(VP(i) for i in {v1,..,vn})`
+
+Therefore when `V` is added to the `{v1,..,vn}` set its initial priority will be:
+
+    A(V) = -sum(VP(i) for i in {v1,..,vn})
+
+Curent implementation uses a penalty factor of 1.125, so the initial priority for a newly added verifier is:
```
milosevic: verifier -> validator

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,268 @@
...
+|Round Priority  | -4 | -3 | -2 | -1 | 0  | 1 | 2 | 3 | 4 | Alg step
+|--------------- | ---|--- |----|--- |--- |---|---|---|---|--------
+|H/R end         |    |    | p2 |    |    |   | p1|   |   |add p3
+|                | p3 |    | p2 |    |    |   | p1|   |   |A(p3) = -4
+|H+1/0           |    |    |    |    |    | p2|   | p1| p3|A(i)+=VP(i)
+|                | p3 |    |    |    |    | p2|   | p1|   |A(p3)-=P
+
+### Proposer Priority Range
+With the introduction of shifting, an interesting case occurs. Low power validators that bind early benefit from subsequent addition of high power validators. This is because new validators are added with negative priorities and this causes an increase in priority for all other validators during the shifting operation. As a consequence, low power validators may end up with high priority and election algorithm will fail with respect to fairness.
+
```
milosevic: This case does not invalidate fairness as fairness is defined only as an eventual property (during period of stability). Adding or removing validators will not break the fairness if validator set stabilise; the expected sequence (that corresponds to the voting power of each validator) will just appear after some number of rounds.

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,268 @@
...
+    def singleProposerSelection (vset):
+        for each validator i in vset:
+            A(i) += VP(i)
+        prop = max(A)
+        A(prop) -= P
+    
+    def ProposerSelection (vset, times):
+        // normalize the priority values
+	    if max(A)-min(A) > 2 * P:
```
milosevic: It seems that this algorithm incentives validator that is at the end of the queue to decrease it's voting power (by delegating to another validator it owns) so he can move ahead in the queue, and then to increase its voting power. The same as above, this step should be executed upon validator set changes only and not during normal proposer selection.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: fixed

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: Ok, let's do this once we wrap up the current spec.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: please check the latest to see if your comment is addressed

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: Will change in next release.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: I will add a better example (the one from #2960)

milosevic: I am wondering if the idea of delaying new validator with setting priority to large number is indeed good idea. There are several weird cases this design opens. Maybe new validator can just be blocked of becoming proposer for some number of rounds/heights maybe proportional to its voting power. During this time his priority would stay to 0, so he will not mess up with the sum of priorities and also after it is effectively added he can join with priority 0.

docs/spec/reactors/consensus/proposer-selection.md
ebuchman: Heh - https://github.com/cosmos/cosmos-sdk/tree/develop/docs/spec/staking

Basically the voting power of a validator in "Bonded Proof of Stake" systems like Cosmos is proportional to the amount of tokens "bonded" to that validator, where bonding can be done by anyone with tokens. People who bond their tokens to a validator are called "delegators"

docs/spec/reactors/consensus/proposer-selection.md
ebuchman: I was originally in favour of such an idea as well but we need to work out the details. 

What are the weird cases for the large negative initial value?

docs/spec/reactors/consensus/proposer-selection.md
Liamsi: I think the justification for the averaging is to keep the prios somewhat closer to zero. 

The justification for normalization is that without it priorities can grow (unironically due to averaging) into both directions. Validators that join early after a large power validator will accumulate lots of priority (relative to validators that join later). This penalizes freshly joined validators more than necessary. With each increment step a proposer get de-priotized in `incrementProposerPriority` by:
```go
mostest.ProposerPriority = safeSubClip(mostest.ProposerPriority, vals.TotalVotingPower())
``` 
But as the very large validator disappeared subtracting total voting power isn't much compared to the gained prio (from the existence of the early large combined with the averaging). 
Sure, this effect *eventually* wears off. But it potentially only wears off after many, many rounds where the early validator(s) were proposer (even with no power changes in the mean time). 

OK, this is not a formal requirement but I guess it can be seen as unfair if early (probably smaller) validators get to propose much more often until (eventually) this becomes fair again.

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,266 @@
...
+    ...
+    A(V)-=P  
+
+It follows that `A(V)-=sum(VP(i) for i in {v1,..,vn})`
+
+Therefore when `V` is added to the `{v1,..,vn}` set its initial priority will be:
+
+    A(V) = -sum(VP(i) for i in {v1,..,vn})
+
+Curent implementation uses a penalty factor of 0.125, so the initial priority for a newly added validator is:
```
ebuchman: Here is the justification: https://github.com/tendermint/tendermint/pull/2785#discussion_r235038971

Essentially:
- factor of 2 or 3 is probably too high
- something small like 1.1 or 1.2 is "probably enough"
- can't multiply by 1.1 or 1.2 efficiently
- can multiply by 1.125 efficiently

tada!

So maybe we can say something like, "1.125 was chosen because it provides a small punishment that is efficient to calculate. See [here](https://github.com/tendermint/tendermint/pull/2785#discussion_r235038971) for more."

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,266 @@
...
+    
+    A(V) = -1.125 * P
+
+If we consider the validator set where p3 has just been added:
+
+Validator | p1 | p2 | p3
+----------|--- |--- |---
+VP        | 1  | 3  | 8
+
+Let's assume that the last round was R and the proposer priorities were as shown in first row (R end) with their sum being 0. p3 is added with A(p3) = -4 (penalty loss due to integer division?)
```
ebuchman: Yes we should probably mention that we use integer division above so the penalty will not apply if the sum of other priorities are less than 8.

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,266 @@
...
+In the next round, p3 will still be ahead in the queue, elected as proposer and move back in the queue (same position it was just added)
```
ebuchman: Shouldn't it be moved back to -8 at this point since P=12?

docs/spec/reactors/consensus/proposer-selection.md
milosevic: You can probably remove all text between L73 and L86.

docs/spec/reactors/consensus/proposer-selection.md
Liamsi: I think @cwgoes came up with this property:

> For any sequence of proposer elections with no power changes and n blocks, the difference between n*power/totalPower and the number of proposed blocks (for each validator) is at most 1.

I think this property makes sense independent from out implementation. It basically says that a validator doesn't get to propose more than it should for any sequence of elections with no power changes (and some number of blocks n).

FWIW this property would break if we only do averaging and no normalization (the difference would be larger for some sequence of elections with no power changes as described in #2960).

Does this make any sense?

docs/spec/reactors/consensus/proposer-selection.md
milosevic: I was thinking about the case where at the end of height H we have validator with significant voting power being removed from val set and new validator with much smaller voting power being added. Then new validator starts with negative voting power of old set, that is much smaller then the current total voting power. So he might stay at the end of the queue for significant amount of time.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: removed

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: I must admit I am confused. Even after reading #3044 things are not clear. I just removed everything talking about rounds and heights and only refer to runs of the ProposerSelection(). Also not sure anymore about R1 which says: proposer_p(h,r) = proposer_q(h,r) ...After reading #3044 this is not so obvious to me. So, I left that there as is.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: Changed the example that justifies the normalization

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: I think @cwgoes' requirement is R2 exactly (for C=1). My question was how to formulate the same requirement for the proposer selections with power changes.
Just to trigger some more thoughts, I added this C=1/2 just because we double the min/max distance to 2 * voting power. Why was this 2 chosen? And why don't limit to just voting power?

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,266 @@
...
 
-p0, p0, p0, p0, p0, p0, p0, p0, p0, p0, p0, p0, p0, p1, p0, p0, p0, p0, p0, etc
+This section includes an overview of the proposer selection algorithm. 
 
-This basically means that almost all rounds have the same proposer. But in this case, the process p0 has anyway enough
-voting power to decide whatever he wants, so the fact that he coordinates almost all rounds seems correct.
+A model that gives a good intuition on how/ why the selection algorithm works and it is fair is that of a priority queue. The validators move ahead in this queue according to their voting power (the higher the voting power the faster a validator moves towards the head of the queue). At each round the following happens:
+- all validators move "ahead" according to their powers: for each validator, increase the priority by the voting power
+- first in the queue becomes the proposer: select the validator with highest priority 
+- move the proposer to the end of the queue: decrease the proposer's priority by the total voting power
```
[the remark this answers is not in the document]
ancazamfir: rephrased

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,266 @@
...
+
+If the set of the validators has not changed, at the beginning of each new round the sum of their priorities is 0.
+
+Following table shows, for the example above, how the proposer priority is calculated during each round. Only 4 rounds are shown, starting with the 5th the same values are computed.
+Each row shows the imaginary queue and the process place in it. The proposer is the closest to the head of the queue, the rightmost validator. As priorities are updated, the validators move right in the queue. At the end of the round the proposer moves left as its priority is reduced after election. At the end of each round the sum of the priorities is 0.
+
+|Round Priority  | -2| -1| 0   |  1| 2   | 3 | 4 | 5 | Alg step
+|--------------- |---|---|---- |---|---- |---|---|---|--------
+|                |   |   |p1,p2|   |     |   |   |   |Initialized to 0
+|1               |   |   |     | p1|     | p2|   |   |A(i)+=VP(i)
```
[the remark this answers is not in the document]
ancazamfir: got migrated to the end and forgot to update. Fixed now

docs/spec/reactors/consensus/proposer-selection.md
```diff
@@ -2,45 +2,266 @@
...
+
+|Round Priority |-3 | -2 | -1 | 0  | 1   | 2   | 3 | 4 | 5 | Comment
+|---------------|-- | ---|--- |--- |---  |---  |---|---|---|--------
+| H/R end       |p3 |    |    |    | p1  | p2  |   |   |   |__remove p2__
+| H+1/0         |   |    |    | p3 |     | p1  |   |   |   |A(i)+=VP(i)
+|               |   | p1 |    | p3 |     |     |   |   |   |A(p1)-= P
+| __new step__  |   |    | p1 |    | p3  |     |   |   |   |A(i) -= avg
+
+The procedure could continue without modifications. However, it is possible that after a sufficiently large number of modifications in validator set, the priority values would migrate towards maximum or minimum allowed values causing truncations due to overflow detection.
+For this reason, the selection procedure adds another step (see last row) that shifts the current priority values left or right such that the average remains close to 0. Due to integer division the priority sum is in {-1, 0, 1}.
```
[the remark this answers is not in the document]
ancazamfir: I have no idea! Who wrote this again? ;) made it (-n, n) with n number of validators.

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: The total power is of the old validator set (1+3)

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: removed

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: I recently changed this to:
f(v) = C * VP(v) / P
C=1 for stable vset,
C=1/2 for vset change

ancazamfir: ok, I changed 2 to some constant k

docs/spec/reactors/consensus/proposer-selection.md
ancazamfir: R1 is fine, Zarko confirmed and enlightened me on the round incrementPriority and pointed me to the code. I should be able to figure it out. For the spec, I removed the mentions of the rounds.

Issue #3146, opened 2019-01-18
json2wal util bug


bradyjoestar:
Currently we could use json2wal to convert the *.json to wal.File.

I found the the length of buffer of readline in json2wal is defalut

https://github.com/tendermint/tendermint/blob/aa40cfcbb97fac3832b47ccdcf99c6b7fb504cb2/scripts/json2wal/main.go#L48

And the length of a line in json will exceed the defaultSize(4096) and report the following erros　because the size of byte of BlockPart
```
panic: failed to unmarshal json: unexpected end of JSON input

goroutine 1 [running]:
main.main()
	/root/gelgo/src/github.com/tendermint/tendermint/scripts/json2wal/main.go:66 +0x73f

```

https://github.com/tendermint/tendermint/blob/aa40cfcbb97fac3832b47ccdcf99c6b7fb504cb2/types/params.go#L14

I will use `NewReaderSize(rd io.Reader, 70000)` to fix this bug.

melekes:
Merged to develop. Should land in 0.29.0. Thanks 👍

PR #3147, opened 2019-01-18
Json2wal

bradyjoestar:
#3146


scripts/json2wal/main.go
```diff
@@ -45,7 +45,7 @@ func main() {
 	}
 	defer walFile.Close()
 
-	br := bufio.NewReader(f)
+	br := bufio.NewReaderSize(f, 70000)
```
ebuchman: Can you please add a comment here explaining why this was necessary. Will help anyone who looks at it in the future :) 

Would be even better if we could make it directly a function of the relevant variable - maybe `2 * types.BlockPartSizeBytes`  for safe keeping :)

bradyjoestar: sure, I should have done it. :)

scripts/json2wal/main.go
```diff
@@ -45,7 +45,9 @@ func main() {
 	}
 	defer walFile.Close()
 
-	br := bufio.NewReader(f)
+	// the length of tendermint/wal/MsgInfo in the wal.json may exceed the defaultBufSize(4096) of bufio
+	// because of the byte array in BlockPart.
```
melekes:
```suggestion
	// because of the byte array in BlockPart
	// leading to unmarshal error: unexpected end of JSON input
```

PR #3148, opened 2019-01-18
Detecting deadlocks online

b00f:
In this PR I am using [go-deadlock](https://github.com/sasha-s/go-deadlock/) for detecting deadlock situations online. The `syns` package has no ability to detect the deadlock situations.
We were struggling a lot on an strange deadlock when updating the validator-set. Helping this package I could find the source of the issue:
```
POTENTIAL DEADLOCK: Recursive locking:
current goroutine 1947 lock 0xc0004d1910
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:196 consensus.(*ConsensusState).GetState { cs.mtx.RLock() } <<<<<
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/rpc/core/consensus.go:53 core.Validators { height := consensusState.GetState().LastBlockHeight + 1 }
~/go/src/github.com/gallactic/gallactic/core/validator/validator_set.go:25 validator._validatorListProxy.validators { result, err := tmRPC.Validators(&height) }
<autogenerated>:1 validator.(*_validatorListProxy).validators ???
~/go/src/github.com/gallactic/gallactic/core/validator/validator_set.go:108 validator.(*ValidatorSet).AdjustPower { result, err := set.proxy.validators(height) }
~/go/src/github.com/gallactic/gallactic/core/consensus/tendermint/abci/app.go:189 abci.(*App).EndBlock { set.AdjustPower(reqEndBlock.GetHeight()) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/abci/client/local_client.go:238 client.(*localClient).EndBlockSync { res := app.Application.EndBlock(req) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/proxy/app_conn.go:77 proxy.(*appConnConsensus).EndBlockSync { return app.appConn.EndBlockSync(req) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/state/execution.go:265 state.execBlockOnProxyApp { abciResponses.EndBlock, err = proxyAppConn.EndBlockSync(abci.RequestEndBlock{Height: block.Height}) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/state/execution.go:96 state.(*BlockExecutor).ApplyBlock { abciResponses, err := execBlockOnProxyApp(blockExec.logger, blockExec.proxyApp, block, state.LastValidators, blockExec.db) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:1292 consensus.(*ConsensusState).finalizeCommit { stateCopy, err = cs.blockExec.ApplyBlock(stateCopy, types.BlockID{block.Hash(), blockParts.Header()}, block) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:1223 consensus.(*ConsensusState).tryFinalizeCommit { cs.finalizeCommit(height) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:1169 consensus.(*ConsensusState).enterCommit.func1 { cs.tryFinalizeCommit(height) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:1200 consensus.(*ConsensusState).enterCommit { } }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:1625 consensus.(*ConsensusState).addVote { cs.enterCommit(height, vote.Round) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:1471 consensus.(*ConsensusState).tryAddVote { added, err := cs.addVote(vote, peerID) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:651 consensus.(*ConsensusState).handleMsg { added, err := cs.tryAddVote(msg.Vote, peerID) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:608 consensus.(*ConsensusState).receiveRoutine { cs.handleMsg(mi) }

Previous place where the lock was grabbed (same goroutine)
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:627 consensus.(*ConsensusState).handleMsg { cs.mtx.Lock() } <<<<<
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:608 consensus.(*ConsensusState).receiveRoutine { cs.handleMsg(mi) }

Other goroutines holding locks:
goroutine 1947 lock 0xc000bb9348
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/abci/client/local_client.go:235 client.(*localClient).EndBlockSync { app.mtx.Lock() } <<<<<
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/proxy/app_conn.go:77 proxy.(*appConnConsensus).EndBlockSync { return app.appConn.EndBlockSync(req) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/state/execution.go:265 state.execBlockOnProxyApp { abciResponses.EndBlock, err = proxyAppConn.EndBlockSync(abci.RequestEndBlock{Height: block.Height}) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/state/execution.go:96 state.(*BlockExecutor).ApplyBlock { abciResponses, err := execBlockOnProxyApp(blockExec.logger, blockExec.proxyApp, block, state.LastValidators, blockExec.db) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:1292 consensus.(*ConsensusState).finalizeCommit { stateCopy, err = cs.blockExec.ApplyBlock(stateCopy, types.BlockID{block.Hash(), blockParts.Header()}, block) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:1223 consensus.(*ConsensusState).tryFinalizeCommit { cs.finalizeCommit(height) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:1169 consensus.(*ConsensusState).enterCommit.func1 { cs.tryFinalizeCommit(height) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:1200 consensus.(*ConsensusState).enterCommit { } }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:1625 consensus.(*ConsensusState).addVote { cs.enterCommit(height, vote.Round) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:1471 consensus.(*ConsensusState).tryAddVote { added, err := cs.addVote(vote, peerID) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:651 consensus.(*ConsensusState).handleMsg { added, err := cs.tryAddVote(msg.Vote, peerID) }
~/go/src/github.com/gallactic/gallactic/vendor/github.com/tendermint/tendermint/consensus/state.go:608 consensus.(*ConsensusState).receiveRoutine { cs.handleMsg(mi) }


```

### DIFF
--- a/libs/sync/sync.go
+++ b/libs/sync/sync.go
@@ -1,0 +1,33 @@
+package sync
+
+/*
+// For detecting deadlock situations
+
+import deadlock "github.com/sasha-s/go-deadlock"
+import "sync"
+
+type Mutex struct {
+	deadlock.Mutex
+}
+
+type RWMutex struct {
+	deadlock.RWMutex
+}
+
+type WaitGroup struct {
+	sync.WaitGroup
+}
+*/
+
+import "sync"
+
+type Mutex struct {
+	sync.Mutex
+}
+
+type RWMutex struct {
+	sync.RWMutex
+}
+type WaitGroup struct {
+	sync.WaitGroup
+}
--- a/abci/client/local_client.go
+++ b/abci/client/local_client.go
@@ -1,10 +1,9 @@
 package abcicli
 
 import (
-	"sync"
-
 	types "github.com/tendermint/tendermint/abci/types"
 	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/libs/sync"
 )
 
 var _ Client = (*localClient)(nil)
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -3,16 +3,17 @@ package consensus
 import (
 	"fmt"
 	"reflect"
-	"sync"
+
 	"time"
 
 	"github.com/pkg/errors"
 
-	"github.com/tendermint/go-amino"
+	amino "github.com/tendermint/go-amino"
 	cstypes "github.com/tendermint/tendermint/consensus/types"
 	cmn "github.com/tendermint/tendermint/libs/common"
 	tmevents "github.com/tendermint/tendermint/libs/events"
 	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/libs/sync"
 	"github.com/tendermint/tendermint/p2p"
 	sm "github.com/tendermint/tendermint/state"
 	"github.com/tendermint/tendermint/types"
@@ -438,9 +439,9 @@ func (conR *ConsensusReactor) broadcastHasVoteMessage(vote *types.Vote) {
 
 func makeRoundStepMessage(rs *cstypes.RoundState) (nrsMsg *NewRoundStepMessage) {
 	nrsMsg = &NewRoundStepMessage{
-		Height: rs.Height,
-		Round:  rs.Round,
-		Step:   rs.Step,
+		Height:                rs.Height,
+		Round:                 rs.Round,
+		Step:                  rs.Step,
 		SecondsSinceStartTime: int(time.Since(rs.StartTime).Seconds()),
 		LastCommitRound:       rs.LastCommit.Round(),
 	}
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"reflect"
 	"runtime/debug"
-	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -13,6 +12,7 @@ import (
 	cmn "github.com/tendermint/tendermint/libs/common"
 	"github.com/tendermint/tendermint/libs/fail"
 	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/libs/sync"
 	tmtime "github.com/tendermint/tendermint/types/time"
 
 	cfg "github.com/tendermint/tendermint/config"
--- a/proxy/client.go
+++ b/proxy/client.go
@@ -1,14 +1,13 @@
 package proxy
 
 import (
-	"sync"
-
 	"github.com/pkg/errors"
 
 	abcicli "github.com/tendermint/tendermint/abci/client"
 	"github.com/tendermint/tendermint/abci/example/counter"
 	"github.com/tendermint/tendermint/abci/example/kvstore"
 	"github.com/tendermint/tendermint/abci/types"
+	"github.com/tendermint/tendermint/libs/sync"
 )
 
 // NewABCIClient returns newly connected client


melekes:
1. The idea of using go-deadlock from time to time to detect potential deadlocks is great, thanks for bringing this up 👍 
2. Does your ABCI app uses RPC to retrieve a validator list in the middle of the `EndBlock()` call? I don't think this is the right way. Your application should save an initial validator set upon receiving `InitChain()` and store it locally. cc @ebuchman

ebuchman:
I recently read about this tool - thanks for trying it out!

Indeed, calling any RPC method that grabs the consensus state from the app during block processing will cause deadlock, since the consensus lock is already held during block processing.

We have a minor comment in https://tendermint.com/docs/spec/abci/apps.html#state about how it will cause deadlock if you try to broadcast txs during abci.Commit. We should extend this to cover any other RPC methods this is true for (ie. anything that needs the ConsensusState).

We can make some changes to relieve some of this, for sure - for instance the `/validators` endpoint doesn't actually need to access the consensus state, so we can remove the particular deadlock situation you're seeing pretty easily. 

That said, making HTTP calls during block execution is in general an anti-pattern and something that should be avoided. I realize this contradicts what I said in https://github.com/tendermint/abci/issues/257, which is my bad. It also basically means you have to store this information in the app, which is redundant with what's already stored in Tendermint, so I understand the need for a solution here and using local RPC endpoints (once we free them of deadlocks) isn't terrible. Perhaps in the future, we can consider ways for apps to request data directly over ABCI, to avoid the need for them to go through the RPC.

In the short term, we should go through our RPC calls and 
- only call methods on the consensusState if we *really* need them
- explicitly mark which RPC methods are safe to call from an ABCI app

libs/sync/sync.go
```diff
@@ -0,0 +1,33 @@
+package sync
+
+/*
+// For detecting deadlock situations
```
ebuchman: Can we re-write this so it's behind a build tag? Then we can easily build binaries with deadlock detection without needing to modify any code!

b00f:
> Does your ABCI app use RPC to retrieve the validator list in a middle of the EndBlock() call? 

I believe anyone can call this method at anytime. But in our case yes, since in our architecture we have pool of validators which is running a VRF to become a validator for validating next block. Read more here: [Byzantine Agreement Proof of Stake](https://medium.com/gallactic/what-is-byzantine-agreement-https://medium.com/gallactic/what-is-byzantine-agreement-proof-of-stake-5f88253f3c78proof-of-stake-5f88253f3c78)

b00f:
> We have a minor comment in https://tendermint.com/docs/spec/abci/apps.html#state about how it will cause deadlock if you try to broadcast txs during abci.Commit.

Is there any way to solve this issue? I think this is not a good solution to tell the App to not send any tx or call any RPC method during committing bock. I try to take a look at the code, maybe I can help you on this matter. If it's possible open a channel for me in case I need to get some information. Do you have any slack group?

> While we can do the above, the officially sanctioned approach for now is to store whatever historical information you need in the application. This is certainly something we'll have to address down the road to avoid the redundant storage use.

I strongly believe that tendermint as a service should provide a **safe** and **reliable** solution to get information like transaction data or validator list. Now we are working on simple block explorer which user can search transaction by the hash. May I know what do you suggest for this request? Do you think the App should stores all transaction somewhere else. This is a big duplication.... 

Do you think gRPC has the same issue? Can we rely on gRPC rather than RPC calls?

libs/sync/sync.go
b00f: I had problem with `WaitGroup` method. Unfortunately that library doesn't support it. I need to check it more.

melekes:
> Do you have any slack group?

https://riot.im/app/#/room/#tendermint:matrix.org

> I strongly believe that tendermint as a service should provide a safe and reliable solution to get information like transaction data or validator list. Now we are working on simple block explorer which user can search transaction by the hash. May I know what do you suggest for this request?

I think we're on the same page here 👍  It's just the current implementation has an issue 😞 which can lead to a deadlock if you call specific RPC endpoints in the middle of End/BeginBlock or DeliverTx functions. I suggest you save validators list in your application. This should not take much space.

I would also like to double stress this ⚠️:
**making HTTP calls during block execution is in general an anti-pattern and something that should be avoided.**

As for the block explorer, calling `/tx` and other endpoints is completely safe.

ebuchman:
@b00f we can get rid of the deadlock for most of the rpc calls: https://github.com/tendermint/tendermint/issues/3161

libs/sync/sync.go
melekes: How about we just write a Makefile target:

```
test_with_deadlock:
    go get github.com/sasha-s/go-deadlock/...
    find . -name "*.go" | xargs -n 1 sed -i '' 's/sync.RWMutex/deadlock.RWMutex/'
    find . -name "*.go" | xargs -n 1 sed -i '' 's/sync.Mutex/deadlock.Mutex/'
    make test
```

?

b00f: I tried this solution, however I was not successful. Maybe you can do a try as well. 
Also `WaitGroup` is an issue here. Can we not use this struct inside the test files?

melekes: Can you explain why it is an issue?

b00f: `‍go-deadlock‍‍` has not such a struct but `sync` has.

melekes:
Closing in favor of https://github.com/tendermint/tendermint/pull/3218 💪

Issue #3106, opened 2019-01-11
privval: integration test for priv_val_server and for KMS


ebuchman:
This repo includes the `priv_val_server` binary under `cmd` as a simple implementation of an external validator signing process, but we don't test it anywhere. We should add tests for it.

We should also have an integration test for the external KMS process, but I'm not sure if that should live in here or in the external KMS repo - see https://github.com/tendermint/kms/issues/49

Replaces part of #2549

ebuchman:
v0.28.0 was released without this based on manual testing. bumping this to a future milestone

melekes:
Merged https://github.com/tendermint/tendermint/pull/3149 to develop. Can we close this now?

thanethomson:
> Merged #3149 to develop. Can we close this now?

I'd say so. There's always room for improvement, but we can do those in future milestones.

Issue #1921, opened 2018-07-06
tools: update documentation & cleanup the directories


zramsay:
and cleanup the Makefile / Dockerfile of these directories

zramsay:
blocking on #1930

marbar3778:
since we got rid of most of the tools that are in tendermint, this issue can be closed

PR #3149, opened 2019-01-18
Add remote signer test harness (KMS)

thanethomson:
As per #3106, this adds a new command to Tendermint to allow for us to build a standalone binary (`remote_val_harness`) to test remote signers such as [KMS](https://github.com/tendermint/kms). The aim is to eventually have this binary added into the KMS [Docker image](https://github.com/tendermint/kms/blob/master/Dockerfile) used for building KMS, and (vice versa) could be used in automated Tendermint building as well to test compatibility with a specific version of the KMS. This should allow for controlled tight coupling in integration testing from both sides, so that changes in either side won't break the other's build, and allows for testing against multiple versions of each other to be able to generate a version compatibility matrix.

Right now, all it does is test that the local public key matches the public key reported by the client, and fails at the point where it attempts to get the client to sign a proposal (still working on understanding why this is so). I'm going to try to adapt some of the relevant tests from the [`privval/client_test.go`](https://github.com/tendermint/tendermint/blob/develop/privval/client_test.go) file.

This is very much still a work-in-progress, but I'd appreciate some feedback on the approach as we go. Will document and write tests once I've iterated to something we're all happy with.


thanethomson:
To run `remote_val_harness` locally, I'm just doing the following at present in a clean environment (i.e. no Tendermint present) from your Tendermint repo:

```bash
# Configure ~/.tendermint/config/priv_validator_key.json and ~/.tendermint/data/priv_validator_state.json and ~/.tendermint/config/genesis.json
> tendermint init
> go run cmd/remote_val_harness/main.go run -a 127.0.0.1:61278
```

**KMS**:

With your `tmkms.toml` file as follows:
```toml
[[validator]]
addr = "tcp://127.0.0.1:61278" # the address for remote_val_harness
chain_id = "test-chain-0XwP5E"  # get this from the Tendermint genesis file
reconnect = true # true is the default
secret_key = "tests/support/secret_connection.key"

[[providers.softsign]]
id = "test-chain-0XwP5E"
path = "/tmp/signing2.key" # this you need to get using https://gist.github.com/Liamsi/a80993f24bff574bbfdbbfa9efa84bc7 after running "tendermint init"
```

Then build and run KMS from the other side:

```bash
> cargo build
> ./target/debug/tmkms start -c tmkms.toml
```

Liamsi:
Thanks for the explanation above 👍 I was going to ask exactly about how to run it with the kms. Will take a look now and run it against the kms, too.

thanethomson:
Awesome, thanks @Liamsi! Right now it successfully tests the public key it gets from the KMS against the one loaded from `priv_validator_key.json`, but it fails at the point where I try to submit a proposal to it for signing. I need to look into the mechanics of how the two components interact with each other in more detail to see what's going on (will be doing that today).

Liamsi:
Do you also get:
```
09:46:00 [ERROR] [test-chain-mcwStp@tcp://127.0.0.1:61278] invalid consensus message: invalid Type
```
on the kms side?

thanethomson:
Yes, I did. I'm looking into that at the moment. What do you think it could be causing that?

privval/test_harness.go
```diff
@@ -0,0 +1,263 @@
...
+	th.logger.Info("Remote", "pubKey", th.sc.GetPubKey())
+	if th.fpv.GetPubKey() != th.sc.GetPubKey() {
+		th.logger.Error("FAILED: Local and remote public keys do not match")
+		th.Shutdown(newTestHarnessError(ErrTestPublicKeyFailed, nil, ""))
+	}
+}
+
+func (th *TestHarness) TestSignProposal() {
+	th.logger.Info("TEST: Signing of proposals")
+	prop := &types.Proposal{Timestamp: time.Now()}
```
Liamsi: This won't pass validation in the kms as this does'nt represent a valid proposal:
https://github.com/tendermint/kms/blob/b4cb33d7255dfb6f4245af8ce39a1feb41299023/tendermint-rs/src/amino_types/proposal.rs#L139-L159

thanethomson: Oh I see. I haven't gotten to grips yet with the internals of the proposal mechanism. Let me take a look at that, thanks! 👍

Liamsi: Or on the tendermint side:
https://github.com/tendermint/tendermint/blob/aa40cfcbb97fac3832b47ccdcf99c6b7fb504cb2/types/proposal.go#L47-L50

Liamsi: This works for instance:

```suggestion
	prop := &types.Proposal{
		Type:     types.ProposalType, // 0x20
		Height:   12345,
		Round:    23456,
		POLRound: -1,
		BlockID: types.BlockID{
			Hash: []byte("hash"),
			PartsHeader: types.PartSetHeader{
				Hash:  []byte("parts_hash"),
				Total: 1000000,
			},
		},
		Timestamp: time.Now(),
	}
```

thanethomson: Adding a `Type` field, along with a `Height` and `Round` value, now successfully signs the proposal 👍 I now just have to validate the signature from the test harness' side, and then we'll have the first cryptographic test of the KMS by the harness 👶

privval/test_harness.go
```diff
@@ -0,0 +1,263 @@
+package privval
```
Liamsi: It's pretty awesome to see this working :-) 
I think it is a bit unfortunate that we need to add a bunch of exported types and vars to the privval package, though. IMHO, the privval package ideally would not expose anything that isn't useful solely for the test-harness. For instance an internal package for this TestHarness would be cool but then we'd need to move it to the `command` package you've created. See https://golang.org/doc/go1.4#internalpackages

Liamsi: Awesome, thanks!

Liamsi:
This will be cool! I'll take a further look later in the day. Let me know if you stumble on other errors messages like above etc.

thanethomson:
Okay, so the proposal signing validation works 🎉 But the vote signing validation isn't working just yet. I get the following error on the KMS side:

```
12:11:10 [ERROR] [test-chain-0XwP5E@tcp://127.0.0.1:61278] I/O error: failed to decode Protobuf message: TimeMsg.seconds: Vote.timestamp: SignVoteRequest.vote: invalid wire type: LengthDelimited (expected Varint)
```

My vote structure is initialised simply as follows:

```go
vote := &types.Vote{
	Type:   voteType,
	Height: 12345,
	Round:  23456,
	BlockID: types.BlockID{
		Hash: hash,
		PartsHeader: types.PartSetHeader{
			Hash:  hash,
			Total: 1000000,
		},
	},
	ValidatorIndex:   0,
	ValidatorAddress: hash[:20],
	Timestamp:        time.Now(),
}
```

It seems to be a problem with the low-level serialization of the `timestamp` field. Are there any differences between the KMS and Tendermint serialization approaches? I see KMS uses `prost-amino` for deserialization.

Liamsi:
> It seems to be a problem with the low-level serialization of the timestamp field. Are there any differences between the KMS and Tendermint serialization approaches? I see KMS uses prost-amino for deserialization.

It should be the same. `Propsoal`s also contain timestamps. So although this seem related to amino/protobuf/prost decoding it is unlikely that the timestamp is causing the problem.

Any chance you are not using the latest master of the kms? Because if you do not have the reverted commit (that reorders votes), this might cause exactly the problem you are seeing.

cmd/remote_val_harness/internal/test_harness.go
```diff
@@ -194,6 +201,56 @@ func (th *TestHarness) TestSignProposal() error {
...
+					Total: 1000000,
+				},
+			},
+			ValidatorIndex:   0,
+			ValidatorAddress: hash[:20],
+			Timestamp:        time.Now(),
+		}
+		// work out the canonicalized serialized byte form of the message
+		// without its signature
+		voteBytes, err := cdc.MarshalBinaryLengthPrefixed(vote)
```
Liamsi: Unrelated to the problem you are seeing on the kms side, this should be vote.SignBytes(th.chainID). Otherwise the verficiation below will fail.

cmd/remote_val_harness/internal/test_harness.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	return nil
+}
+
+func (th *TestHarness) TestSignVote() error {
+	th.logger.Info("TEST: Signing of votes")
+	for _, voteType := range voteTypes {
+		th.logger.Info("Testing vote type", "type", voteType)
+		// sha256 hash of "hash"
+		hash, _ := hex.DecodeString("D04B98F48E8F8BCC15C6AE5AC050801CD6DCFD428FB5F9E65C4E16E7807340FA")
```
Liamsi: This and the comment above can be replaced by:
tmhash.SumTruncated([]byte("hash"))

cmd/remote_val_harness/internal/test_harness.go
```diff
@@ -0,0 +1,363 @@
...
+			Round:  23456,
+			BlockID: types.BlockID{
+				Hash: hash,
+				PartsHeader: types.PartSetHeader{
+					Hash:  hash,
+					Total: 1000000,
+				},
+			},
+			ValidatorIndex:   0,
+			ValidatorAddress: hash[:20],
```
Liamsi:
```suggestion
			ValidatorAddress: tmhash.SumTruncated([]byte("hash")),
```

cmd/remote_val_harness/internal/test_harness.go
```diff
@@ -0,0 +1,363 @@
...
+		BlockID: types.BlockID{
+			Hash: hash,
+			PartsHeader: types.PartSetHeader{
+				Hash:  hash,
+				Total: 1000000,
+			},
+		},
+		Timestamp: time.Now(),
+	}
+	propBytes, err := cdc.MarshalBinaryLengthPrefixed(types.CanonicalizeProposal(th.chainID, prop))
```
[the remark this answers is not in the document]
thanethomson: Mostly because of the panic over [here](https://github.com/tendermint/tendermint/blob/develop/types/proposal.go#L94). I'd prefer to return any error and try to shut down gracefully, ideally.

cmd/remote_val_harness/internal/test_harness.go
thanethomson: Awesome! I'll definitely do that 👍

cmd/remote_val_harness/internal/test_harness.go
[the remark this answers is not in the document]
Liamsi: Thanks for the explanation. This should only panic if we completely messed up either amino or introduced sth obscure in `CanonicalizeProposal` that can't be encoded. IMHO, I still think this isn't sth. that this integration tests should bother about.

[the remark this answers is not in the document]
thanethomson: Okay cool, I'll use the `SignBytes` function. I suppose, like you say, if it does panic it's a sign of something pretty important that needs to be fixed ASAP 😁

thanethomson:
See https://github.com/tendermint/kms/issues/49#issuecomment-456079078 - the integrated solution is coming together now.

privval/remote_signer.go
```diff
@@ -160,6 +166,10 @@ func RegisterRemoteSignerMsg(cdc *amino.Codec) {
 	cdc.RegisterConcrete(&SignedProposalResponse{}, "tendermint/remotesigner/SignedProposalResponse", nil)
 	cdc.RegisterConcrete(&PingRequest{}, "tendermint/remotesigner/PingRequest", nil)
 	cdc.RegisterConcrete(&PingResponse{}, "tendermint/remotesigner/PingResponse", nil)
+	// TODO: Evaluate whether we need to make this an official remotesigner
+	// message. Right now it's only relevant to KMS, and is only used in
+	// testing.
```
Liamsi: The idea was that tendermint can shut down the kms via this message. I can not think of a realistic scenario where this is useful though 🤔  

CC'ing @zmanian and @tarcieri who introduced this message on the kms AFAIR. 

If we add this to tendermint, we should also document this in the spec (maybe you did that and I haven't seen this yet; reviewing each commit separately).

thanethomson: You're right - the only scenario that I'm catering for here is being able to programmatically (and gracefully) shut the KMS down once the harness is finished testing it.

An alternative approach here is to rather send a `SIGTERM` from outside the process during the testing over here: https://github.com/thanethomson/kms/blob/issues/154/tests/support/run-harness-tests.sh

It doesn't appear as though the KMS shuts down gracefully at present when I hit Ctrl+Break. Is this something related to the [Abscissa](https://github.com/iqlusioninc/abscissa/) framework KMS uses, or is this something we need to build into the KMS? It looks like it's something I'd have to build into the KMS (I'll handle that if necessary).

thanethomson: Would be great @tarcieri, if there's a relatively easy to way to handle signals generically - alternatively I could just use a library like [`signal_hook`](https://docs.rs/signal-hook/0.1.7/signal_hook/) internally in KMS for now and refactor the [start command logic](https://github.com/tendermint/kms/blob/master/src/commands/start.rs#L52).

tarcieri: Sounds good to me.

thanethomson:
See https://circleci.com/gh/tendermint/kms/449 - the second-last step. The integration testing harness (`remote_val_harness`) can now successfully run in KMS' build process on CircleCI to ensure compatibility with a specific version of Tendermint.

The only remaining thing to do here is to decide on a naming convention for KMS build images so as to make the incorporated Tendermint version explicit. For example, `tendermint/kms:build-<tendermint_version>-<image_version>`, such as `tendermint/kms:build-0.29.0-fbd1e794-v0`.

privval/remote_signer.go
```diff
@@ -147,6 +147,12 @@ func (sc *RemoteSignerClient) Ping() error {
 	return nil
 }
 
+// SendPoisonPill allows us to ask the remote signer to shut down.
+func (sc *RemoteSignerClient) SendPoisonPill() error {
+	// we need no response when sending a poison pill message
+	return writeMsg(sc.conn, &PoisonPillMsg{})
+}
```
Liamsi: I guess we can do away with all this `PoisonPill` related code now (as per https://github.com/tendermint/kms/pull/162)

thanethomson: Definitely - it's now removed 👍

thanethomson:
Oh one more important thing before I finish this: it needs some documentation 👍

cmd/remote_val_harness/internal/test_harness.go
```diff
@@ -0,0 +1,402 @@
...
+	if th.fpv.GetPubKey() != th.sc.GetPubKey() {
+		th.logger.Error("FAILED: Local and remote public keys do not match")
+		return newTestHarnessError(ErrTestPublicKeyFailed, nil, "")
+	}
+	return nil
+}
+
+// TestSignProposal makes sure the remote signer can successfully sign
+// proposals.
+func (th *TestHarness) TestSignProposal() error {
```
Liamsi: We later should add failing tests too. If the proposal is not valid for instance. Not in this PR though.

thanethomson:
> The only nit: I wish this would come with the same functionality while adding a little less code.

I think the only way we can reduce the amount of code required here is to hard-code some of the values and remove the `cobra`-based command line argument parsing? I could do something similar to what's in https://github.com/tendermint/tendermint/blob/master/cmd/priv_val_server/main.go and just use the `flag` library.

thanethomson:
`test_cover` seems to be failing because of:

```
=== RUN   TestWaitChan
--- FAIL: TestWaitChan (1.23s)
    clist_test.go:284: number of pushed items (99) not equal to number of seen items (98)
FAIL
coverage: 64.6% of statements
FAIL	github.com/tendermint/tendermint/libs/clist	1.294s
Exited with code 1
```

cmd/remote_val_harness/internal/test_harness.go
```diff
@@ -0,0 +1,391 @@
...
+		// make sure the socket doesn't exist - if so, try to delete it
+		if cmn.FileExists(addr) {
+			if err := os.Remove(addr); err != nil {
+				logger.Error("Failed to remove existing Unix domain socket", "addr", addr)
+				return nil, err
+			}
+		}
+	}
+	ln, err := net.Listen(proto, addr)
+	logger.Info("Listening at", "proto", proto, "addr", addr)
```
Liamsi: Maybe move this line underneath the error check below?

thanethomson: Yes, good catch 😊 👍

Liamsi:
> test_cover seems to be failing because of:

That is a know unrelated issue: https://github.com/tendermint/tendermint/issues/2227

cmd/remote_val_harness/internal/test_harness.go
```diff
@@ -0,0 +1,391 @@
...
+	}
+}
+
+// newTestHarnessSocketVal creates our client instance which we will use for
+// testing.
+func newTestHarnessSocketVal(logger log.Logger, cfg TestHarnessConfig) (*privval.SocketVal, error) {
+	proto, addr := cmn.ProtocolAndAddress(cfg.BindAddr)
+	if proto == "unix" {
+		// make sure the socket doesn't exist - if so, try to delete it
+		if cmn.FileExists(addr) {
```
Liamsi: Question: This is the only difference to `newSocketVal` in [client_test.go](https://github.com/tendermint/tendermint/blob/4d7b29cd8f8ef998cf8f7141f3a100b0a4fa718d/privval/client_test.go#L379-L399). 

Did you experience that the socket file was left over from previous test runs here?

thanethomson: Yes, I've seen it a few times where, even though we've closed the Unix socket, it still exists in the file system. IIRC, it happened during testing. Checking my `/tmp` folder right now, I still have 58 `.sock` files lying around from testing 😁 

Also, IIRC, I've had it happen at times that if the file already exists, the startup sequence fails because the file exists.

thanethomson:
Decided to rather change the name to `tm-signer-harness` and put it into the `tools` folder - it does seem to make sense to keep it separate from the other `cmd/*` packages. Also seems to fit better in the documentation.

Liamsi:
Sounds very good to me!

tools/tm-signer-harness/internal/test_harness.go
```diff
@@ -0,0 +1,391 @@
...
+}
+
+// Run will execute the tests associated with this test harness. The intention
+// here is to call this from one's `main` function, as the way it succeeds or
+// fails at present is to call os.Exit() with an exit code related to the error
+// that caused the tests to fail, or exit code 0 on success.
+func (th *TestHarness) Run() {
+	th.logger.Info("Starting test harness")
+	donec := make(chan struct{})
+	go func() {
```
melekes: why spawn a goroutine if you block in the end?

tools/tm-signer-harness/internal/test_harness.go
```diff
@@ -0,0 +1,391 @@
...
+			}
+		}
+	}
+	ln, err := net.Listen(proto, addr)
+	if err != nil {
+		return nil, err
+	}
+	logger.Info("Listening at", "proto", proto, "addr", addr)
+	var svln net.Listener
+	if proto == "unix" {
```
melekes:
```go
switch proto {
case "unix":
case "tcp":
default:
  // error
}
```

tools/tm-signer-harness/internal/test_harness_test.go
```diff
@@ -0,0 +1,207 @@
...
+	defaultConnDeadline   = 100
+)
+
+func TestRemoteSignerTestHarnessMaxAcceptRetriesReached(t *testing.T) {
+	cfg := makeConfig(t, 1, 2)
+	defer cleanup(cfg)
+
+	th, err := NewTestHarness(log.TestingLogger(), cfg)
+	assert.NoError(t, err)
+	donec := make(chan struct{})
```
melekes: why run in a goroutine if you block right after?

tools/tm-signer-harness/internal/test_harness_test.go
```diff
@@ -0,0 +1,207 @@
...
+	)
+}
+
+// For running relatively standard tests.
+func harnessTest(t *testing.T, rsMaker func(th *TestHarness) *privval.RemoteSigner, expectedExitCode int) {
+	cfg := makeConfig(t, 100, 3)
+	defer cleanup(cfg)
+
+	th, err := NewTestHarness(log.TestingLogger(), cfg)
+	assert.NoError(t, err)
```
melekes: better to use require here to stop the test (because next steps will probably fail to do -> make no sense to continue)

tools/tm-signer-harness/README.md
```diff
@@ -0,0 +1,137 @@
+# tm-signer-harness
```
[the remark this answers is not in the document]
zramsay: ^ ++ or another method

I also just noticed that we have duplicate files; e.g., `/tools/tm-bench/README.md` is also `/docs/tools/benchmarking.md` and same goes for `tm-monitor`.

Is there a better way we can do this? Ideally the files are in `/docs` so that they can easily be included in the build. Then the various README's within `/tools` can link to those files?

see also #1921

tools/tm-signer-harness/README.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+If the current version of Tendermint and KMS are compatible, `tm-signer-harness`
+should now exit with a 0 exit code. If they are somehow not compatible, it
+should exit with a meaningful non-zero exit code (see the exit codes below).
+
+### Step 5: Shut down KMS
+Simply hit Ctrl+Break on your KMS instance (or use the `kill` command in Linux)
+to terminate it gracefully.
+
```
zramsay: pumped to smoke test these steps!

docs/tools/remote-signer-validation.md
```diff
@@ -0,0 +1,137 @@
+# tm-signer-harness
```
zramsay: is this a duplicate file with tools/tm-signer-harness/README.md ?

docs/tools/remote-signer-validation.md
```diff
@@ -0,0 +1,137 @@
...
+```toml
+[[validator]]
+addr = "tcp://127.0.0.1:61219"         # This is where we will find tm-signer-harness.
+chain_id = "test-chain-0XwP5E"         # The Tendermint chain ID for which KMS will be signing.
+reconnect = true                       # true is the default
+secret_key = "./secret_connection.key" # Where to find our secret connection key.
+
+[[providers.softsign]]
+id = "test-chain-0XwP5E"               # The Tendermint chain ID for which KMS will be signing.
+path = "./signing.key"                 # The signing key we extracted earlier.
```
zramsay: `signer.key`, otherwise: `error: couldn't load keyring: config error: can't open ./signing.key: No such file or directory (os error 2): No such file or directory (os error 2)`

docs/tools/remote-signer-validation.md
```diff
@@ -0,0 +1,137 @@
...
+
+```toml
+[[validator]]
+addr = "tcp://127.0.0.1:61219"         # This is where we will find tm-signer-harness.
+chain_id = "test-chain-0XwP5E"         # The Tendermint chain ID for which KMS will be signing.
+reconnect = true                       # true is the default
+secret_key = "./secret_connection.key" # Where to find our secret connection key.
+
+[[providers.softsign]]
+id = "test-chain-0XwP5E"               # The Tendermint chain ID for which KMS will be signing.
```
zramsay: need to point out somewhere that these chainID's need to be changed, grabbing them from the `~/.tendermint/config/genesis.json`

docs/tools/remote-signer-validation.md
```diff
@@ -0,0 +1,137 @@
...
+
+```bash
+tm-signer-harness run \             # The "run" command executes the tests
+    -addr tcp://127.0.0.1:61219 \   # The address we promised KMS earlier
+    -tmhome /home/user/.tendermint  # Where to find our Tendermint configuration/data files.
+```
+
+If the current version of Tendermint and KMS are compatible, `tm-signer-harness`
+should now exit with a 0 exit code. If they are somehow not compatible, it
+should exit with a meaningful non-zero exit code (see the exit codes below).
```
zramsay: yup! thought i had latest `tmkms` but got exit code 5. Upgraded from 0.2.1 to 0.3.0 and it worked 💯

docs/.vuepress/config.js
```diff
@@ -81,7 +81,8 @@ module.exports = {
         children:  [
 	  "/tools/",
 	  "/tools/benchmarking",
-	  "/tools/monitoring"
+    "/tools/monitoring",
```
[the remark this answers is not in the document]
thanethomson: There's a mixture of tabs and spaces in the original 😟

tools/tm-signer-harness/internal/test_harness.go
```diff
@@ -0,0 +1,391 @@
...
+	Info string // Any additional information
+}
+
+var _ error = (*TestHarnessError)(nil)
+
+// TestHarness allows for testing of a remote signer to ensure compatibility
+// with this version of Tendermint.
+type TestHarness struct {
+	addr             string
+	sc               *privval.SocketVal
```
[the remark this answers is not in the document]
thanethomson: Changed to `spv` now for consistency 👍

tools/tm-signer-harness/internal/test_harness.go
thanethomson: Good catch! This was a remnant of an older approach that I did away with. Refactored now to remove the goroutine.

tools/tm-signer-harness/internal/test_harness.go
thanethomson: Fixed 👍

docs/tools/remote-signer-validation.md
thanethomson: Done 👍

docs/tools/remote-signer-validation.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Running against KMS
+As an example of how to use `tm-signer-harness`, the following instructions show
+you how to execute its tests against [KMS](https://github.com/tendermint/kms).
+For this example, we will make use of the **software signing module in KMS**, as
+the hardware signing module requires a physical
+[YubiHSM](https://www.yubico.com/products/yubihsm/) device.
+
+### Step 1: Install KMS on your local machine
+See the [KMS repo](https://github.com/tendermint/kms) for details on how to set
```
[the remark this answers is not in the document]
thanethomson: Sure - done 👍

docs/tools/remote-signer-validation.md
thanethomson: Good catch! I've changed it to `signing.key` on line 75 😁 (that's the convention we've been using up to now).

docs/tools/remote-signer-validation.md
thanethomson: Done 👍

tools/tm-signer-harness/internal/test_harness_test.go
thanethomson: Fixed 👍

docs/tools/remote-signer-validation.md
```diff
@@ -0,0 +1,144 @@
...
+Also, because we want KMS to connect to `tm-signer-harness`, we will need to
+provide a secret connection key from KMS' side:
+
+```bash
+tmkms keygen secret_connection.key
+```
+
+### Step 3: Configure and run KMS
+KMS needs some configuration to tell it to use the softer signing module as well
+as the `signer.key` file we just generated. Save the following to a file called
```
zramsay: signing.key

docs/tools/remote-signer-validation.md
zramsay: line 87 needs it too ;)

docs/tools/remote-signer-validation.md
thanethomson: OMG 😊 Fixed now!

melekes:
Hmm.. tests failed two times in a row with

```
--- FAIL: TestWaitingTimeoutOnNilPolka (0.04s)
panic: Timeout expired while waiting for NewTimeout event [recovered]
        panic: Timeout expired while waiting for NewTimeout event

goroutine 38971 [running]:
testing.tRunner.func1(0xc0075e0f00)
        /usr/local/go/src/testing/testing.go:792 +0x6a7
panic(0xfa79a0, 0xc0004f0520)
        /usr/local/go/src/runtime/panic.go:513 +0x1b9
github.com/tendermint/tendermint/consensus.ensureNewEvent(0xc0096af800, 0x1, 0x0, 0x1c9c380, 0x10f5946, 0x32)
        /go/src/github.com/tendermint/tendermint/consensus/common_test.go:362 +0x61f
github.com/tendermint/tendermint/consensus.ensureNewTimeout(0xc0096af800, 0x1, 0x0, 0x989680)
        /go/src/github.com/tendermint/tendermint/consensus/common_test.go:434 +0x70
github.com/tendermint/tendermint/consensus.TestWaitingTimeoutOnNilPolka(0xc0075e0f00)
        /go/src/github.com/tendermint/tendermint/consensus/state_test.go:1110 +0x362
testing.tRunner(0xc0075e0f00, 0x122c040)
        /usr/local/go/src/testing/testing.go:827 +0x163
created by testing.(*T).Run
        /usr/local/go/src/testing/testing.go:878 +0x65a
FAIL    github.com/tendermint/tendermint/consensus      45.870s
```

thanethomson:
> Hmm.. tests failed two times in a row with
> 
> ```
> --- FAIL: TestWaitingTimeoutOnNilPolka (0.04s)
> panic: Timeout expired while waiting for NewTimeout event [recovered]
>         panic: Timeout expired while waiting for NewTimeout event
> ```

Very strange. I've pulled the latest changes to this PR (including the `develop` merge) and ran `GOCACHE=off make test_cover` and all tests passed.

tools/tm-signer-harness/internal/test_harness.go
```diff
@@ -0,0 +1,392 @@
...
+
+// newTestHarnessSocketVal creates our client instance which we will use for
+// testing.
+func newTestHarnessSocketVal(logger log.Logger, cfg TestHarnessConfig) (*privval.SocketVal, error) {
+	proto, addr := cmn.ProtocolAndAddress(cfg.BindAddr)
+	if proto == "unix" {
+		// make sure the socket doesn't exist - if so, try to delete it
+		if cmn.FileExists(addr) {
+			if err := os.Remove(addr); err != nil {
+				logger.Error("Failed to remove existing Unix domain socket", "addr", addr)
```
melekes:
```suggestion
				logger.Error("Failed to remove existing Unix domain socket", "addr", addr, "err", err)
```

PR #3150, opened 2019-01-18
p2p: file descriptor leaks

### LINKED ISSUE #2967: p2p: file descriptor leaks
Running `lsof -i` on seed and public validator nodes from gaia-9002 shows lots of connections left in `CLOSE_WAIT` state, which suggests that connections aren't being closed properly in Tendermint in some cases.

I'm also seeing net_info report 0 peers but according to the logs there's at least 1 connected peer - this is a known issue (https://github.com/tendermint/tendermint/issues/2332) but perhaps these things are related.

Also previously reported memory leak in seed nodes: https://github.com/tendermint/tendermint/issues/2821

melekes:
Fixes #2967

### DIFF
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -304,8 +304,8 @@ func (sw *Switch) stopAndRemovePeer(peer Peer, reason interface{}) {
 	if sw.peers.Remove(peer) {
 		sw.metrics.Peers.Add(float64(-1))
 	}
+	sw.transport.Cleanup(peer.RemoteAddr())
 	peer.Stop()
-	sw.transport.Cleanup(peer.Addr())
 	for _, reactor := range sw.reactors {
 		reactor.RemovePeer(peer, reason)
 	}
@@ -531,14 +531,14 @@ func (sw *Switch) acceptRoutine() {
 			)
 
 			_ = p.CloseConn()
-			sw.transport.Cleanup(p.Addr())
+			sw.transport.Cleanup(p.RemoteAddr())
 
 			continue
 		}
 
 		if err := sw.addPeer(p); err != nil {
 			_ = p.CloseConn()
-			sw.transport.Cleanup(p.Addr())
+			sw.transport.Cleanup(p.RemoteAddr())
 			if p.IsRunning() {
 				_ = p.Stop()
 			}
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -210,6 +210,7 @@ func (sw *Switch) OnStart() error {
 func (sw *Switch) OnStop() {
 	// Stop peers
 	for _, p := range sw.peers.List() {
+		sw.transport.Cleanup(p)
 		p.Stop()
 		if sw.peers.Remove(p) {
 			sw.metrics.Peers.Add(float64(-1))
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -304,7 +304,7 @@ func (sw *Switch) stopAndRemovePeer(peer Peer, reason interface{}) {
 	if sw.peers.Remove(peer) {
 		sw.metrics.Peers.Add(float64(-1))
 	}
-	sw.transport.Cleanup(peer.RemoteAddr())
+	sw.transport.Cleanup(peer)
 	peer.Stop()
 	for _, reactor := range sw.reactors {
 		reactor.RemovePeer(peer, reason)
@@ -530,15 +530,13 @@ func (sw *Switch) acceptRoutine() {
 				"max", sw.config.MaxNumInboundPeers,
 			)
 
-			_ = p.CloseConn()
-			sw.transport.Cleanup(p.RemoteAddr())
+			sw.transport.Cleanup(p)
 
 			continue
 		}
 
 		if err := sw.addPeer(p); err != nil {
-			_ = p.CloseConn()
-			sw.transport.Cleanup(p.RemoteAddr())
+			sw.transport.Cleanup(p)
 			if p.IsRunning() {
 				_ = p.Stop()
 			}
@@ -599,6 +597,7 @@ func (sw *Switch) addOutboundPeerWithConfig(
 	}
 
 	if err := sw.addPeer(p); err != nil {
+		sw.transport.Cleanup(p)
 		if p.IsRunning() {
 			_ = p.Stop()
 		}
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -305,6 +305,7 @@ func (sw *Switch) stopAndRemovePeer(peer Peer, reason interface{}) {
 		sw.metrics.Peers.Add(float64(-1))
 	}
 	peer.Stop()
+	sw.transport.Cleanup(peer.Addr())
 	for _, reactor := range sw.reactors {
 		reactor.RemovePeer(peer, reason)
 	}
@@ -529,13 +530,18 @@ func (sw *Switch) acceptRoutine() {
 				"max", sw.config.MaxNumInboundPeers,
 			)
 
-			_ = p.Stop()
+			_ = p.CloseConn()
+			sw.transport.Cleanup(p.Addr())
 
 			continue
 		}
 
 		if err := sw.addPeer(p); err != nil {
-			_ = p.Stop()
+			_ = p.CloseConn()
+			sw.transport.Cleanup(p.Addr())
+			if p.IsRunning() {
+				_ = p.Stop()
+			}
 			sw.Logger.Info(
 				"Ignoring inbound connection: error while adding peer",
 				"err", err,
@@ -593,7 +599,9 @@ func (sw *Switch) addOutboundPeerWithConfig(
 	}
 
 	if err := sw.addPeer(p); err != nil {
-		_ = p.Stop()
+		if p.IsRunning() {
+			_ = p.Stop()
+		}
 		return err
 	}
 
@@ -628,7 +636,8 @@ func (sw *Switch) filterPeer(p Peer) error {
 	return nil
 }
 
-// addPeer starts up the Peer and adds it to the Switch.
+// addPeer starts up the Peer and adds it to the Switch. Error is returned if
+// the peer is filtered out or failed to start or can't be added.
 func (sw *Switch) addPeer(p Peer) error {
 	if err := sw.filterPeer(p); err != nil {
 		return err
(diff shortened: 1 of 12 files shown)


p2p/switch.go
```diff
@@ -637,10 +646,8 @@ func (sw *Switch) addPeer(p Peer) error {
 	p.SetLogger(sw.Logger.With("peer", p.NodeInfo().NetAddress()))
 
 	// All good. Start peer
-	if sw.IsRunning() {
```
melekes: did not understand this. if the switch is not running, we don't start the peer, so... it stays there presumably forever.

p2p/switch.go
```diff
@@ -529,13 +530,18 @@ func (sw *Switch) acceptRoutine() {
 				"max", sw.config.MaxNumInboundPeers,
 			)
 
-			_ = p.Stop()
+			_ = p.CloseConn()
+			sw.transport.Cleanup(p.RemoteAddr())
```
ebuchman: Why not just pass the peer to Cleanup and let it call CloseConn and do the rest?

p2p/switch.go
```diff
@@ -593,7 +599,9 @@ func (sw *Switch) addOutboundPeerWithConfig(
 	}
 
 	if err := sw.addPeer(p); err != nil {
-		_ = p.Stop()
+		if p.IsRunning() {
```
ebuchman: don't we need cleanup here too ?

p2p/switch.go
ebuchman: I think this is to handle the shut down case where the switch has stopped but we're concurrently trying to add a peer. You're right this might be a leak, but if we were shutting down anyways, it wouldn't be an issue. I'm afraid with out it this could cause a panic during the shutdown? Not sure

melekes: ok. added a comment & log statement

p2p/switch.go
melekes: yep, added

p2p/switch.go
```diff
@@ -628,19 +635,24 @@ func (sw *Switch) filterPeer(p Peer) error {
...
-	// All good. Start peer
+	// Handle the shut down case where the switch has stopped but we're
+	// concurrently trying to add a peer.
 	if sw.IsRunning() {
+		// All good. Start peer
 		if err := sw.startInitPeer(p); err != nil {
 			return err
 		}
+	} else {
+		sw.Logger.Info("Won't start a peer - switch is not running", "peer", p)
```
melekes:
```suggestion
		sw.Logger.Error("Won't start a peer - switch is not running", "peer", p)
```

melekes:
TODO:
- [ ] write a test / ensure there's no leak any more
- [ ] changelog

melekes:
tests are unstable 😞

p2p/switch_test.go
```diff
@@ -477,6 +476,67 @@ func TestSwitchFullConnectivity(t *testing.T) {
...
+	// 2. check we close new connections if we already have MaxNumInboundPeers peers
+	rp := &remotePeer{PrivKey: ed25519.GenPrivKey(), Config: cfg}
+	rp.Start()
+	conn, err := rp.Dial(sw.NodeInfo().NetAddress())
+	require.NoError(t, err)
+	// check conn is closed
+	one := make([]byte, 1)
+	conn.SetReadDeadline(time.Now().Add(10 * time.Millisecond))
+	_, err = conn.Read(one)
+	if assert.Error(t, err) {
```
ebuchman: Should this fail without the fix?

p2p/switch.go
```diff
@@ -304,6 +304,7 @@ func (sw *Switch) stopAndRemovePeer(peer Peer, reason interface{}) {
 	if sw.peers.Remove(peer) {
 		sw.metrics.Peers.Add(float64(-1))
 	}
+	sw.transport.Cleanup(peer)
```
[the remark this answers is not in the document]
melekes: I thought we could wrap `onPeerError()` function, but a) onPeerError does not necessary mean we're gonna stop the peer b) I prefer opaque things

p2p/switch.go
```diff
@@ -529,13 +530,16 @@ func (sw *Switch) acceptRoutine() {
...
-			_ = p.Stop()
+			sw.transport.Cleanup(p)
 
 			continue
 		}
 
 		if err := sw.addPeer(p); err != nil {
-			_ = p.Stop()
+			sw.transport.Cleanup(p)
+			if p.IsRunning() {
```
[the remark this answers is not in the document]
melekes: It is fine, but by adding if condition we're showing that `addPeer` can return without starting the peer.

p2p/transport.go
```diff
@@ -52,6 +52,9 @@ type Transport interface {
 
 	// Dial connects to the Peer for the address.
 	Dial(NetAddress, peerConfig) (Peer, error)
+
+	// Cleanup any resources associated with Peer.
+	Cleanup(Peer)
```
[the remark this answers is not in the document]
melekes: I think we should be extra careful when considering increasing coupling of our components. I like the fact that peer does not know anything about the transport.

Having Cleanup function do have cons:
- you can forget to call it

pros:
- very opaque
- and direct

p2p/switch_test.go
ebuchman: I tried commenting out the fix and it didn't fail. Not sure if I need to try something else, I just commented out the one Cleanup call in the max inbound peer handling.

melekes:
```
--- FAIL: TestSwitchAcceptRoutine (0.04s)
        Error Trace:    switch_test.go:511
        Error:          Not equal:
                        expected: *errors.errorString(&errors.errorString{s:"EOF"})
                        actual  : *net.OpError(&net.OpError{Op:"read", Net:"tcp", Source:(*net.TCPAddr)(0xc00022c960), Addr:(*net.TCPAddr)(0xc00022c990), Err:(*poll.TimeoutError)(0xfbdec0)})
        Test:           TestSwitchAcceptRoutine
FAIL
exit status 1
```

fails for me

p2p/transport.go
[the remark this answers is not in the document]
ebuchman: I agree with the desire for decoupling, but it seems in this case we have components that are genuinely coupled. The property that the peer doesn't know about the transport sounds nice in spirit, but in practice, as far as I understand here, the implementations of Peer and Transport are already tightly wound - a particular implementation of Transport returns a particular implementation of Peer, and a particular implementation of peer is expected to come from a particular implementation of Transport. So Peer in general doesn't need to know about Transport, but if the Transport that created it necessitates some cleanup, then it surely will.

By pushing this implicit coupling into an extra responsibility for callers, we're just asking them to forget about it. If we have transports that don't require this cleanup, we're cluttering the Transport interface unnecessarily. So I would think, in this case, `you can forget to call it` takes the cake.

We don't need to give the peer a direct reference to the transport. It could have some cleanup closure or a channel the Transport listens for.  But the current separation seems prone to misuse.

Case in point, it looks like we're forgetting to call Cleanup on peers in sw.OnStop

ebuchman:
Oh no, the tests seem to fail non-deterministically with different results:

Locally:

```
--- FAIL: TestTransportMultiplexAcceptMultiple (0.08s)
    transport_test.go:184: auth failure: secrect conn failed: EOF
panic: send on closed channel

goroutine 1623 [running]:
github.com/tendermint/tendermint/p2p.TestTransportMultiplexAcceptNonBlocking.func1(0xc0006cad80, 0xc0003e6f60, 0xc0003e7320, 0xc0003e7200)
        /Users/ethanbuchman/programming/goApps/src/github.com/tendermint/tendermint/p2p/transport_test.go:257 +0x389
created by github.com/tendermint/tendermint/p2p.TestTransportMultiplexAcceptNonBlocking
        /Users/ethanbuchman/programming/goApps/src/github.com/tendermint/tendermint/p2p/transport_test.go:232 +0x1d5
exit status 2
FAIL    github.com/tendermint/tendermint/p2p    1.073s
```

Locally again:

```
2019/01/21 17:08:22 Failed to accept conn: accept tcp 127.0.0.1:62900: use of closed network connection
--- FAIL: TestTransportMultiplexAcceptMultiple (0.06s)
    transport_test.go:184: auth failure: secrect conn failed: EOF
FAIL
exit status 1
FAIL    github.com/tendermint/tendermint/p2p    1.073s
```

And on Circle:

```
--- FAIL: TestSwitchAcceptRoutine (0.08s)
	Error Trace:	switch_test.go:501
	Error:      	Not equal: 
	            	expected: 5
	            	actual  : 0
	Test:       	TestSwitchAcceptRoutine
	Error Trace:	switch_test.go:512
	Error:      	Not equal: 
	            	expected: *errors.errorString(&errors.errorString{s:"EOF"})
	            	actual  : *net.OpError(&net.OpError{Op:"read", Net:"tcp", Source:(*net.TCPAddr)(0xc0005ee000), Addr:(*net.TCPAddr)(0xc0005ee030), Err:(*poll.TimeoutError)(0x1358340)})
	Test:       	TestSwitchAcceptRoutine
	Error Trace:	switch_test.go:513
	Error:      	Not equal: 
	            	expected: 5
	            	actual  : 1
	Test:       	TestSwitchAcceptRoutine
FAIL
coverage: 66.2% of statements
FAIL	github.com/tendermint/tendermint/p2p	2.245s
Exited with code 1
```


So we probably need to address this before merging. 

Also opened #3183 - might as well address in this PR now that we still need to fix things anyways.

melekes:
> Oh no, the tests seem to fail non-deterministically with different results:

Sorry, I thought I'd fixed them after two success builds.

melekes:
```
E[2019-01-22|07:38:39.104] Connection failed @ recvRoutine (reading byte) switch=1 peer=b378d5be5cc68eb798ade7524d1167d5adb153d7@127.0.0.1:36052 conn=MConn{127.0.0.1:34774} err=EOF
```

looks like MConnection is trying to read 1 byte? and fails because, well, `remotePeer` does not write or read anything from connection.

melekes:
Fixed the test that I wrote. Can't replicate the first two failures @ebuchman has posted. Anyway, they should be addressed separately as they are not related (I don't see how they are related)

p2p/transport.go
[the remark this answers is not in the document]
ebuchman: That's fair. But the issue about making two calls still stands. Maybe we let Cleanup call Stop ?

PR #3155, opened 2019-01-18
test proposer frequencies

ebuchman:
Add test to check that proposers are selected at the frequency we expect.

The following fixes needed to be made:
- take absolute value of the priorities diff
- normalize repeatedly until the diff is less than the threshold

### DIFF
--- a/state/state_test.go
+++ b/state/state_test.go
@@ -271,13 +271,13 @@ func TestProposerFrequency(t *testing.T) {
 	testCases := []struct {
 		powers []int64
 	}{
-		//{[]int64{1, 1}},
-		//{[]int64{1, 2}},
-		//{[]int64{1, 100}},
-		//{[]int64{5, 5}},
-		//{[]int64{5, 100}},
-		//{[]int64{50, 50}},
-		//{[]int64{50, 100}},
+		{[]int64{1, 1}},
+		{[]int64{1, 2}},
+		{[]int64{1, 100}},
+		{[]int64{5, 5}},
+		{[]int64{5, 100}},
+		{[]int64{50, 50}},
+		{[]int64{50, 100}},
 		{[]int64{1, 1000}},
 		/*{[]int64{1, 1, 1}},
 		{[]int64{1, 2, 3}},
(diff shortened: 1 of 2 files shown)


types/validator_set.go
```diff
@@ -86,11 +86,13 @@ func (vals *ValidatorSet) IncrementProposerPriority(times int) {
 	// the 2nd check (threshold > 0) is merely a sanity check which could be
 	// removed if all tests would init. voting power appropriately;
 	// i.e. threshold should always be > 0
-	if diff > threshold && threshold > 0 {
+
+	for diff > threshold && threshold > 0 {
```
ebuchman: This is guaranteed to terminate, right?

state/state_test.go
```diff
@@ -264,6 +265,118 @@ func TestOneValidatorChangesSaveLoad(t *testing.T) {
...
+	maxVals := 100
+	maxPower := 1000
+	nTestCases := 5
+	for i := 0; i < nTestCases; i++ {
+		N := cmn.RandInt() % maxVals
+		vals := make([]*types.Validator, N)
+		for j := 0; j < N; j++ {
+			votePower := int64(cmn.RandInt() % maxPower)
+			privVal := types.NewMockPV()
+			pubKey := privVal.GetPubKey()
```
Liamsi: Nit (as this is only a test): but why do we use the MockPV here to generate a pub key while below we use `ed25519.GenPrivKey().PubKey()`?

types/validator_set.go
Liamsi: I think so as the rescaling / normalisation should guarantee that after a finite (a few) number of iterations we reach diff <= thresold. I'd assume in the case of non-random prios, this should happen after one iter even almost always.

types/validator_set.go
```diff
@@ -152,7 +156,7 @@ func computeMaxMinPriorityDiff(vals *ValidatorSet) int64 {
 		}
 	}
 	diff := max - min
-	return diff
+	return int64(math.Abs(float64(diff)))
```
Liamsi: Ouch. Thanks @ebuchman!

Liamsi:
Thanks so much @ebuchman!

state/state_test.go
```diff
@@ -264,6 +265,118 @@ func TestOneValidatorChangesSaveLoad(t *testing.T) {
...
+	}{
+		// 2 vals
+		{[]int64{1, 1}},
+		{[]int64{1, 2}},
+		{[]int64{1, 100}},
+		{[]int64{5, 5}},
+		{[]int64{5, 100}},
+		{[]int64{50, 50}},
+		{[]int64{50, 100}},
+		{[]int64{1, 1000}},
```
Liamsi: It would be cool to have a test with voting power close to `MaxTotalVotingPower/2` or alike. Unfortunately, it would run forever.

Issue #2862, opened 2018-11-16
Lite client verification error in all usage: insufficient power


cwgoes:
## Summary of Bug 
`gaiacli query *` is returning error for validators who have already bonded.

## Steps to Reproduce
- Join a node to the `gaia-9000` testnet
- `gaiacli tx create-validator` to create your validator
- Run `gaiacli query signing-info $(gaiad tendermint show-validator) --chain-id gaia-9000`
- See the error:
```
ERROR: Invalid commit -- insufficient old voting power: got 200, needed 267
```

Moved from https://github.com/cosmos/cosmos-sdk/issues/2756, cc @jackzampolin.

cwgoes:
The relevant error is in `types/validator_set.go` - https://github.com/tendermint/tendermint/blob/master/types/validator_set.go#L306.

melekes:
Not immediately obvious why the commit does not have enough voting power. Also, there's no info about commit itself 😞

HaoyangLiu:
@cwgoes 
Do you have located the root of this issue?

Now in tendermint lite client, when validatorset change, the verification algorithm requires the precommits have more than 2/3 voting power in both old validator set and new validator set. However, if the validator set change is much significant, the lite client verification will be failed. 

Currently, we already have `nextValidatorHash`. If the hash of new validator set equals to `nextValidatorHash`, and the precommit has more than 2/3 voting power in new validator set, then the block commit is supposed to be valid. It is not necessary to change the voting power again old validator set.

What do you think about my opinion?

ebuchman:
The problem is this: https://github.com/tendermint/tendermint/blob/v0.28.1/lite/dynamic_verifier.go#L234

But that error never gets returned by VerifyFutureCommit, so instead of proceeding with the bisection to find a commit we can verify, it looks like we just fail.

It shouldn't be hard to fix. The more concerning thing is the lack of testing. I should have a PR up shortly

ebuchman:
I think https://github.com/tendermint/tendermint/pull/3171 should fix in. We need to add much more testing for this package.

HaoyangLiu:
@ebuchman 
I have add a [comment](https://github.com/tendermint/tendermint/pull/3171/files#r249307540) to your PR.

ebuchman:
Fixed in #3171. Will ship in v0.29

jackzampolin:
This is awesome @ebuchman !! Thank you for the fix!!!

PR #3171, opened 2019-01-20
fix DynamicVerifier for large validator set changes

ebuchman:
Fixes https://github.com/tendermint/tendermint/issues/2862

* [x] Wrote tests

### DIFF
--- a/types/validator_set.go
+++ b/types/validator_set.go
@@ -413,8 +413,7 @@ func (vals *ValidatorSet) VerifyCommit(chainID string, blockID BlockID, height i
 	if talliedVotingPower > vals.TotalVotingPower()*2/3 {
 		return nil
 	}
-	return fmt.Errorf("Invalid commit -- insufficient voting power: got %v, needed %v",
-		talliedVotingPower, vals.TotalVotingPower()*2/3+1)
+	return errTooMuchChange{talliedVotingPower, vals.TotalVotingPower()*2/3 + 1}
 }
 
 // VerifyFutureCommit will check to see if the set would be valid with a different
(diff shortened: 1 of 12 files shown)


types/validator_set.go
```diff
@@ -496,12 +495,37 @@ func (vals *ValidatorSet) VerifyFutureCommit(newSet *ValidatorSet, chainID strin
 	}
 
 	if oldVotingPower <= oldVals.TotalVotingPower()*2/3 {
-		return cmn.NewError("Invalid commit -- insufficient old voting power: got %v, needed %v",
-			oldVotingPower, oldVals.TotalVotingPower()*2/3+1)
+		return errTooMuchChange{oldVotingPower, oldVals.TotalVotingPower()*2/3 + 1}
```
HaoyangLiu: Once `errTooMuchChange` is returned, the verification algorithm would try to find out a middle point so that the validator set change is not too much for both prior one and the back one. 

But, if there is no point which can meet the above confidition, how to do with that situation?

Suppose now there are four validators(marked as A, B, C, D) and their voting power are 100. On height 1000, there is a delegator which delegate 10000 tokens to validator A. Then the voting power of validator A is 10100. In this situation, the verification algorithm will still report error.

What do you think about this situation?

ebuchman: > the verification algorithm would try to find out a middle point so that the validator set change is not too much for both prior one and the back one.

If it never finds one where the valset change is small enough, it eventually gets back to a block where the validator set actually changed, and it can verify that change directly, thanks to the NextValidatorSetHash in the header.

The test in https://github.com/tendermint/tendermint/pull/3171/files#diff-0e9c3f682613cbf3fbae4b768e9c8fdeR78 tests a case where we have 5 validators, and in one block replace all of them at once with a new 5. Because the old validator set signs a header with the NextValidatorSetHash before the new validator set takes over, we can verify it successfully. 

If it weren't for NextValidatorSetHash, we would either have to cap the validator set changes in a block, or execute blocks before proposing them.

HaoyangLiu: Great! Make sense!

PR #3182, opened 2019-01-21
[WIP]: few follow up changes to the proposer priorities

Liamsi:
- floor instead of ceil when rescaling prios
- move averaging to top
- make ValidatorSet.RescalePriorities unexported
- remove calls to RescalePriorities from bucky's tests
- re-rename diffMax to threshold
- fix some typos

WIP as `TestProposerPriorityDoesNotGetResetToZero` needs minor changes to be fixed and there is an outdated comment in the code

* [ ] Wrote tests

### DIFF
--- a/state/state_test.go
+++ b/state/state_test.go
@@ -283,6 +283,7 @@ func TestProposerFrequency(t *testing.T) {
 
 		// 3 vals
 		{[]int64{1, 1, 1}},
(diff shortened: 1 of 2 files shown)


Liamsi:
I think we can close this for now. @ancazamfir did quite some clean-up and code changes. We'll probably revisit this after we formalize more properties of the proposer selection algo.

PR #3194, opened 2019-01-23
EventBus and indexerService should be started before first block

ackratos:
* [ ] Wrote tests


consensus/replay_file.go
```diff
@@ -326,19 +326,19 @@ func newConsensusStateForReplay(config cfg.BaseConfig, csConfig *cfg.ConsensusCo
...
 
+	handshaker := NewHandshaker(stateDB, state, blockStore, eventBus, gdoc)
+	err = handshaker.Handshake(proxyApp)
+	if err != nil {
+		cmn.Exit(fmt.Sprintf("Error on handshake: %v", err))
+	}
+
 	mempool, evpool := sm.MockMempool{}, sm.MockEvidencePool{}
-	blockExec := sm.NewBlockExecutor(stateDB, log.TestingLogger(), proxyApp.Consensus(), mempool, evpool)
+	blockExec := sm.NewBlockExecutor(stateDB, log.TestingLogger(), proxyApp.Consensus(), eventBus, mempool, evpool)
```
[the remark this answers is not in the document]
melekes: Yeah, we had a discussion about this a while back. My point was a param X should not be a part of the constructor params if the object does not depend on it. E.g. most of the services we have will do just fine (perform their core logic) without logger / pubsub / ...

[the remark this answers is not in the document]
melekes: > if you think consistency is more important, I can revert related parts back

please do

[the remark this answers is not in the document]
ackratos: > please do

done dude :)

node/node.go
```diff
@@ -217,10 +217,45 @@ func NewNode(config *cfg.Config,
 		return nil, fmt.Errorf("Error starting proxy app connections: %v", err)
 	}
 
+	eventBus := types.NewEventBus()
```
melekes: we should write a comment explaining the certain order

```
EventBus and IndexerService must be started before the handshake because ...
```

consensus/replay.go
```diff
@@ -196,19 +196,21 @@ type Handshaker struct {
...
+	eventBus     types.BlockEventPublisher
 	genDoc       *types.GenesisDoc
 	logger       log.Logger
 
 	nBlocks int // number of blocks applied to the state
 }
 
 func NewHandshaker(stateDB dbm.DB, state sm.State,
-	store sm.BlockStore, genDoc *types.GenesisDoc) *Handshaker {
+	store sm.BlockStore, eventBus types.BlockEventPublisher, genDoc *types.GenesisDoc) *Handshaker {
```
melekes: This also needs to be done via `SetEventBus()` as everywhere

ackratos:
@melekes done again, could you please take another look?

node/node.go
```diff
@@ -217,11 +217,50 @@ func NewNode(config *cfg.Config,
 		return nil, fmt.Errorf("Error starting proxy app connections: %v", err)
 	}
 
+	// EventBus and IndexerService must be started before the handshake because
+	// we need index the txs of handshaker replayed block which might not done when last time node stopped
+	// i.e. endblocker panic OR stop node process after save block to db but before index tx
```
Liamsi: Suggestion (change to):
// EventBus and IndexerService must be started before the handshake because
// we might need to index the txs of the replayed block as this might not have happened 
// when the node stopped last time (i.e. the node went stopped after it saved the block 
// but before it indexed the txs, or, the endblocker panic'ed)

ackratos: done

node/node.go
```diff
@@ -217,11 +217,51 @@ func NewNode(config *cfg.Config,
...
+	if err != nil {
+		return nil, err
+	}
+
 	// Create the handshaker, which calls RequestInfo, sets the AppVersion on the state,
 	// and replays any blocks as necessary to sync tendermint with the app.
 	consensusLogger := logger.With("module", "consensus")
 	handshaker := cs.NewHandshaker(stateDB, state, blockStore, genDoc)
 	handshaker.SetLogger(consensusLogger)
+	handshaker.SetEventBus(eventBus)
```
melekes: Does this mean that all of the blocks executed during replay will be indexed? Sometimes we may index a block twice. I guess it's okay since indexing is idempotent.

Issue #3180, opened 2019-01-21
lite: cleanup the doc.go


ebuchman:
See https://godoc.org/github.com/tendermint/tendermint/lite

It's littered with markdown notation that doesn't render. Needs to be cleaned up

melekes:
Merged to develop. Thanks @infinytum 🍰

PR #3198, opened 2019-01-23
Updated Doc for issue #3180

infinytum:
I removed the Markdown parts and replaced them with godoc equivalents if possible. 
This was done for issue #3180 

* [X] Wrote tests

### DIFF
--- a/lite/doc.go
+++ b/lite/doc.go
@@ -79,6 +79,7 @@ used by the DynamicVerifier for persistence.
  }
 
 * DBProvider - persistence provider for use with any libs/DB.
+
 * MultiProvider - combine multiple providers.
 
 The suggested use for local light clients is client.NewHTTPProvider(...) for
--- a/lite/doc.go
+++ b/lite/doc.go
@@ -58,29 +58,25 @@ Provider and PersistentProvider
 
 A Provider allows us to store and retrieve the FullCommits.
 
-```go
-type Provider interface {
+ type Provider interface {
 	// LatestFullCommit returns the latest commit with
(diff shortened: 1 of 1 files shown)


lite/doc.go
```diff
@@ -56,35 +54,32 @@ greater).
...
 strategy to verify any SignedHeader of the blockchain.
 
-## Provider and PersistentProvider
+Provider and PersistentProvider
 
 A Provider allows us to store and retrieve the FullCommits.
 
-```go
-type Provider interface {
+ type Provider interface {
```
melekes: code blocks need to be indented (example https://github.com/googleapis/google-cloud-go/blob/master/datastore/doc.go#L57-L59)

lite/doc.go
```diff
@@ -15,9 +15,7 @@ for you, so you can just build nice applications.
 We design for clients who have no strong trust relationship with any Tendermint
 node, just the blockchain and validator set as a whole.
 
-# Data structures
```
melekes: why remove?

infinytum: I cant have two go-doc headlines right next to each other. I supposed it was better to highlight the title of the following, rather than the general headline. 

We could put a small text between Data structures and SignedHeader. (Unless there is a way to put two headlines together.

melekes: ok

lite/doc.go
infinytum: Its at least one extra space to the other comment content, which should be rendered as a code example:

image

I now saw  in crypto/doc.go that you have 4 spaces extra. I've applied that to the two examples  👍

PR #3207, opened 2019-01-25
WIP: Adds first remote test network (derived from localnet)

thanethomson:
This PR contains a suggestion, based on https://github.com/freshautomations/tnt, as to a remote test network configuration management structure - open for discussion.

Right now it just contains a single remote network configuration: `localnet`, which is based on the results of a `make localnet-start` command from the Tendermint base repository, with some minor modifications (documented in the `README.md` file).

This is currently up and running on @greg-szabo's servers.

* [ ] Wrote tests


test/testnet/localnet/README.md
```diff
@@ -0,0 +1,94 @@
+# Remote Version of Localnet
+
```
zramsay: can we link to this file from the docs, e.g., this file: https://github.com/tendermint/tendermint/blob/master/docs/networks/README.md

thanethomson: Will do 👍

thanethomson:
One thing that we do also need to do is include the Ansible service scripts for Tendermint. Part of the assumption of this `localnet` example is that the Tendermint `systemd` service is already available on the target machine (and, by implication, that the target machines use `systemd`).

The question is, how far do we go with these assumptions? Do we provide support for a multitude of systems, or do we only support our specific subset (i.e. CentOS, using `systemd`)?

test/testnet/localnet/Makefile
```diff
@@ -0,0 +1,47 @@
+.PHONY: deploy stop update-prereq build-tendermint status
+
+ACTIVATE_VENV := source venv/bin/activate
+TESTNET ?= tik
+
```
greg-szabo: `GOPATH ?= $(shell go env GOPATH)` is necessary for line 8 on systems that didn't set GOPATH explicitly.

test/testnet/localnet/Makefile
```diff
@@ -0,0 +1,47 @@
...
+build-tendermint:
+	@echo "Building Tendermint binary for Linux..."; \
+	cd ${GOPATH}/src/github.com/tendermint/tendermint; \
+	make build-linux; \
```
greg-szabo: `make -C ${GOPATH}/src/github.com/tendermint/tendermint build-linux` can do this in one line for you and the "done" line is unneccessary. (Makefile makes a similar report at the end if I recall.)
No need for multi-lines this way.

test/testnet/localnet/Makefile
```diff
@@ -0,0 +1,47 @@
...
+	echo "Done."
+
+# Sets up a clean instance of this test network and fires it up
+deploy: venv venv/bin/ansible
+	@$(ACTIVATE_VENV); \
```
greg-szabo: Maintaining ansible installation will become a hassle very quickly. (Especially because it's hard to do it across different operating systems.) Let's put up a big warning that ansible is a pre-requisite requirement and let the user decide how they install it. For example most of our developers already have it using `brew install ansible` which they know how to maintain, while they might never have seen Python virtualenv before. As much as I like Python, I think we should abstain from it in this Makefile.

test/testnet/localnet/README.md
```diff
@@ -0,0 +1,94 @@
...
+Remote version of the test network generated by running:
```
greg-szabo: I don't get it. `localnet` is a well established idea across all teams. Regardless if it's gaia or tendermint, `localnet` means a docker-based 4-node network brought up by docker-compose. Don't use localnet-start to start a remote network. Either call it remotenet-start or something else. (I think we did have remotenet-start a while ago.)

test/testnet/localnet/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+```bash
+> make localnet-start
+```
+
+from the Tendermint repository. There are a few additional minor tweaks, however:
+
+1. The proxy app is preconfigured to `kvstore`
+2. Logging format is JSON
+3. `persistent_peers` is preconfigured
```
greg-szabo: The point of localnet - and I would argue that the same is true for this remotenet setup - is that the developer can set configuration as they see fit. The `tendermint testnet` command was used to generate the minimum viable default configuration as a template for the developer but it is expected to be changed before a network is started. In that Thai ID issue for example, create_empty_blocks has to be set to true explicitly, if I recall correctly. Persistent peers can't really be set automatically as soon as you create a new testnet.

Instead of manually tweaking the configuration during setup (which gets cumbersome very quickly as well as not really useful for the developer), I think we should stick to the default config for now and figure out an overall method how to improve applying the config to an arbitrary testnet. This requires a bit of discussion and use-cases from the dev team.

test/testnet/localnet/README.md
```diff
@@ -0,0 +1,94 @@
...
+## Contents
+This folder contains the relevant prerequisites to deploy this test network to a
+set of predefined hosts. Here is a description of a subset of the layout of this
+folder's contents:
+
+```
+|_ Makefile           The primary entrypoint to interacting with the test network.
+|_ requirements.txt   Python requirements for running the deployment scripts.
+|_ config             Ansible configuration management.
+   |_ hosts           The inventory file for the hosts to which we will be deploying the network.
```
greg-szabo: Storing hosts is also something to discuss. TNT is a quick and dirty, short-term solution to the problem of self-managed testnets. We're definitely not going to keep updating a hosts file on a repo when we create new testnets because testnet creation should (and will) be a more flexible solution.

To be honest, I'm beginning to think, this solution is not ready to be merged into the tendermint repo and we should have more testing, use-cases and discussions around its usage.

test/testnet/localnet/README.md
```diff
@@ -0,0 +1,94 @@
...
+      node_key: "<redacted>"     # Node private key (ed25519)
+      power: "1"
+    node1:
+      address: "<redacted>"
+      public_key: "<redacted>"
+      private_key: "<redacted>"
+      node_key: "<redacted>"
+      power: "1"
+    # etc., same for node2 and node3
+```
```
greg-szabo: In AWS, this can actually be polled from an API. (At least the network parts). So, as I said, maybe it's worth rethinking this.

test/testnet/localnet/README.md
```diff
@@ -0,0 +1,94 @@
...
+**NOTE**: This will **delete** any existing Tendermint configuration/data from
+the target nodes, so make sure you first back up your Tendermint configuration
+and data prior to running this command.
+
+Once you've backed up any important data, to deploy this test net, simply run:
+
+```bash
+# This assumes that you've got Tendermint in your ${GOPATH}, and it runs the
+# `make build-linux` command in the Tendermint source folder.
+> make build-tendermint
```
greg-szabo: If I recall, dependencies are not actually depepndent on `build-linux`, so for a fresh install, you need to run `make -C $GOPATH/src/github.com/tendermint/tendermint get_vendor_deps` first. Assuming you have `dep` installed and available.

Maybe it's quicker and painless to redirect the user to the "how to build the binary" documentation instead of recreating one here. It's the same principle.

test/testnet/localnet/config/deploy.yml
```diff
@@ -0,0 +1,38 @@
+---
+- hosts: "{{ lookup('env', 'TESTNET') }}"
+  become: yes
+  vars:
+    # Where, on the remote hosts, should we put the Tendermint home folder?
+    tmhome: "/home/tendermint/.tendermint"
```
greg-szabo: Ok, this is where things got hairy the last time too. As soon as we start implementing options that noone really cares about but that define how the service runs, the setup will get complicated and users will just drop it.

The point of the TNT setup was that it is _opinionated_. You don't get to choose where the config is, because it's already set up for you. Don't focus on the infrastructure because in 99% of the developer cases, you don't care about it. You care about the application (and to be honest I don't want to submit this to the tendermint repo because we might get away from EC2 instances before this becomes really useful for the team) not about where or how it's set up. (Within reason.)

I also want to minimize ansible playbooks - or better: make them as simple as possible. Look at the `networks/remote/ansible` folder in the SDK and the tendermint repos. Huge, complex, robust ansible scripts (I wrote them) and they are effectively useless. You need to be an ansible expert to understand what they are doing and why. You need to have the inside knowledge of how ansible folder structure affects ansible running to be able to figure out how exactly it is run.

If I could, I would simply stop using ansible for config management for these testnets. This will eventually happen, we're just not there yet.

Keep It Simple. Simple?

test/testnet/localnet/config/roles/tendermint-node/templates/genesis.json.jinja2
```diff
@@ -0,0 +1,32 @@
...
+                "type": "tendermint/PubKeyEd25519",
+                "value": "{{ node.public_key }}"
+            },
+            "power": "{{ node.power }}",
+            "name": "{{ node_id }}"
+        }{% if not loop.last %},{% endif %}
+    {% endfor %}
+  ],
+  "app_hash": ""
+}
```
greg-szabo: Now this! Is the first real developers problem that we need to resolve. This is where earlier scripts eventually became unmaintainable. As soon as the genesis.json (or the config.toml for that matter) structure changes, this template is out. BUT, experience tells us that it is not going to get updated.

It was Anton's wise words that I had to understand regarding this: we manage config in the tendermint binary, in a specific package. _That's_ the authentic source for config structure, use it! That's why the `testnet` command came to be, which generates testnet configs (toml, genesis, priv_key, etc) for you so you can edit and use it as you see fit. Let's get rid of this template and see what's missing in the `testnet` command that does not allow us to use that to generate the config. - Then implement it.

test/testnet/localnet/config/roles/tendermint-node/templates/node_key.json.jinja2
```diff
@@ -0,0 +1 @@
+{"priv_key":{"type":"tendermint/PrivKeyEd25519","value":"{{ secrets.nodes[inventory_hostname].node_key }}"}}
```
greg-szabo: Same as with genesis.json. Do not use ansible templates for config.

test/testnet/localnet/Makefile
greg-szabo: Oh, sorry, one more note: you don't call `make` in a Makefile, you call `$(MAKE)`. The command looks like this:
`$(MAKE) -C ${GOPATH}/src/github.com/tendermint/tendermint build-linux`

Unless you want to include that file, in which case you can just call the `build-linux` target directly. :)

thanethomson:
Thanks @greg-szabo for the detailed feedback! 😁 I think you're right - we do need to sit down and figure out what exactly we're trying to do here and what we want to accomplish from the test network discussion.

Right now it seems as though there are many ways to deploy a Tendermint test network and, as you've said, we need to pick our battles (e.g. as I asked yesterday: Do we cater to multiple init systems/flavours of Linux? Do we also then go, once we're done with super complex Ansible setups, and try build Helm charts for Kubernetes? And then perhaps Nomad?). The permutations and complexity could skyrocket.

Or, do we just need some highly opinionated custom scripts to deal with specific issues reported by specific users?

The latter option does seem to be the better approach, which we could evolve over time, or not 😁

PR #3212, opened 2019-01-26
pubsub: fixes after Ethan's review

melekes:
in https://github.com/tendermint/tendermint/pull/3209

* [ ] Wrote tests

### DIFF
--- a/libs/pubsub/pubsub.go
+++ b/libs/pubsub/pubsub.go
@@ -211,6 +211,9 @@ func (s *Server) Subscribe(ctx context.Context, clientID string, query Query, ou
 func (s *Server) Unsubscribe(ctx context.Context, clientID string, query Query) error {
 	s.mtx.RLock()
 	clientSubscriptions, ok := s.subscriptions[clientID]
+	if ok {
+		_, ok = clientSubscriptions[query.String()]
+	}
 	s.mtx.RUnlock()
 	if !ok {
 		return ErrSubscriptionNotFound
--- a/libs/pubsub/pubsub.go
+++ b/libs/pubsub/pubsub.go
@@ -101,7 +101,7 @@ type Server struct {
 	cmdsCap int
 
 	mtx           sync.RWMutex
-	subscriptions map[string]map[string]Query // subscriber -> query (string) -> Query
+	subscriptions map[string]map[string]struct{} // subscriber -> query (string) -> empty struct
 }
 
 // Option sets a parameter for the server.
@@ -143,7 +143,7 @@ func (ts tagMap) Len() int {
 // provided, the resulting server's queue is unbuffered.
 func NewServer(options ...Option) *Server {
 	s := &Server{
-		subscriptions: make(map[string]map[string]Query),
+		subscriptions: make(map[string]map[string]struct{}),
 	}
 	s.BaseService = *cmn.NewBaseService(nil, "PubSub", s)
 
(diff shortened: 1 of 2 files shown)


libs/pubsub/pubsub.go
```diff
@@ -101,7 +101,7 @@ type Server struct {
 	cmdsCap int
 
 	mtx           sync.RWMutex
-	subscriptions map[string]map[string]Query // subscriber -> query (string) -> Query
+	subscriptions map[string]map[string]struct{} // subscriber -> query (string) -> empty struct
```
Liamsi: Do I understand this correctly: The empty struct is only necessary here because we want to be able to quickly add / delete a queries in subscribe/unsubscribe (and maps are cool for this)?

libs/pubsub/pubsub.go
```diff
@@ -332,24 +336,34 @@ loop:
...
 	// create subscription
-	state.queries[q][clientID] = ch
+	state.queryToChanMap[qStr][clientID] = ch
+
+	// initialize queries if needed
+	if _, ok := state.queries[qStr]; !ok {
+		state.queries[qStr] = &queryPlusRefCount{q: q, refCount: 0}
+	}
+	// increment reference counter
+	state.queries[qStr].refCount++
```
Liamsi: I hope everyone touching this code will be as good of a coder as you and we will never forget to increment/decrement the `refCount` (as done in this pr) 🙈

melekes: I know it looks a bit dangerous, but I believe the game is worth the candle. 

pros:
- we don't store a pointer to a query in pubsub#subscriptions ("preserve original query" hack)
- we don't store a pointer to a query if `state.queries[qStr]` already exists

cons:
- need to increment / decrement ref. counter

Note: we'll probably get rid of the pubsub in the future thanks to Zarko's suggestion. When/if this happens, I'd still like us to preserve the code, maybe even extract it to a separate repo cause I think the new design is especially good.

libs/pubsub/pubsub.go
```diff
@@ -363,43 +377,58 @@ func (state *state) remove(clientID string, q Query) {
...
-	delete(state.queries[q], clientID)
-	if len(state.queries[q]) == 0 {
-		delete(state.queries, q)
+	delete(state.queryToChanMap[qStr], clientID)
+	if len(state.queryToChanMap[qStr]) == 0 {
+		delete(state.queryToChanMap, qStr)
+	}
+
+	// decrease ref counter in queries
+	state.queries[qStr].refCount--
```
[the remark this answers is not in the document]
melekes: no

libs/pubsub/pubsub.go
```diff
@@ -211,24 +209,20 @@ func (s *Server) Subscribe(ctx context.Context, clientID string, query Query, ou
...
-	// ?
 	select {
-	case s.cmds <- cmd{op: unsub, clientID: clientID, query: origQuery}:
+	case s.cmds <- cmd{op: unsub, clientID: clientID, query: query}:
 		s.mtx.Lock()
-		// if its the only query left, should we also delete the client?
 		delete(clientSubscriptions, query.String())
+		if len(clientSubscriptions) == 0 {
+			delete(s.subscriptions, clientID)
+		}
```
Liamsi: Do you think it is worth to add a test that covers this (new) behaviour? E.g. subscribe a query for client, call unsubscribe with that query&client, expect unsubscribeAll on that client to return an `ErrSubscriptionNotFound`?

melekes: Good idea 👍

PR #3218, opened 2019-01-28
add go-deadlock tool to help detect deadlocks

melekes:
Run it with `make test_with_deadlock`. After it's done, use Git to
cleanup `git checkout .`

Link: https://github.com/sasha-s/go-deadlock/
Replaces https://github.com/tendermint/tendermint/pull/3148

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -226,11 +226,19 @@ test_race:
 	@echo "--> Running go test --race"
 	@GOCACHE=off go test -p 1 -v -race $(PACKAGES)
 
+# uses https://github.com/sasha-s/go-deadlock/ to detect potential deadlocks
 test_with_deadlock:
 	find . -name "*.go" | grep -v "vendor/" | xargs -n 1 sed -i.bak 's/sync.RWMutex/deadlock.RWMutex/'
 	find . -name "*.go" | grep -v "vendor/" | xargs -n 1 sed -i.bak 's/sync.Mutex/deadlock.Mutex/'
 	find . -name "*.go" | grep -v "vendor/" | xargs -n 1 goimports -w
 	make test
+	make cleanup_after_test_with_deadlock
+
+# cleanes up after you ran test_with_deadlock
+cleanup_after_test_with_deadlock:
+	find . -name "*.go" | grep -v "vendor/" | xargs -n 1 sed -i.bak 's/deadlock.RWMutex/sync.RWMutex/'
+	find . -name "*.go" | grep -v "vendor/" | xargs -n 1 sed -i.bak 's/deadlock.Mutex/sync.Mutex/'
+	find . -name "*.go" | grep -v "vendor/" | xargs -n 1 goimports -w
 
 ########################################
 ### Formatting, linting, and vetting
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -226,6 +226,11 @@ test_race:
 	@echo "--> Running go test --race"
 	@GOCACHE=off go test -p 1 -v -race $(PACKAGES)
 
+test_with_deadlock:
+	find . -name "*.go" | grep -v "vendor/" | xargs -n 1 sed -i.bak 's/sync.RWMutex/deadlock.RWMutex/'
+	find . -name "*.go" | grep -v "vendor/" | xargs -n 1 sed -i.bak 's/sync.Mutex/deadlock.Mutex/'
+	find . -name "*.go" | grep -v "vendor/" | xargs -n 1 goimports -w
+	make test
```
thanethomson: Just a note on this: it obviously modifies imports locally without reverting those imports back. Would it be worth mentioning this somewhere? Or perhaps reverting the imports back? This is fine for automated builds, where the codebase is ephemeral, but when running locally it becomes tedious to have to revert these changed files.

Also, I see that the `goimports` command here also modifies a whole bunch of local files to "fix" their imports. Would it be worth doing a `goimports` on the whole repo and doing a separate PR for the results?

b00f:
This will be very helpful. Thanks a lot.  :100: :1st_place_medal:

Makefile
melekes: > Would it be worth mentioning this somewhere?

I thought about running `git checkout .` in the end, but it can be dangerous since dev. might have other changes. We can also add another target like `make cleanup_deadlock`

> Would it be worth doing a goimports on the whole repo and doing a separate PR for the results?

no, goimports only needed here to remove "sync" import and add "deadlock github.com/.../go-deadlock" import for us.

most of the editors are now running `goimports` by default when you save a file (given you have a go plugin like https://github.com/fatih/vim-go)

thanethomson: > I thought about running `git checkout .` in the end, but it can be dangerous since dev. might have other changes. We can also add another target like `make cleanup_deadlock`

Having this `make cleanup_deadlock` would be best, since `make test` could exit with a non-zero exit code 👍 

> most of the editors are now running goimports by default when you save a file (given you have a go plugin like https://github.com/fatih/vim-go)

Yes, this is why I find it strange that, when I run:

```bash
find . -name '*.go' | grep -v 'vendor/' | xargs -n 1 goimports -w
```

in my local Tendermint repo it modifies almost 40 files. Most of the changes, however, look like this:

```diff
diff --git a/state/wire.go b/state/wire.go
index eeb156d6..f7a61129 100644
--- a/state/wire.go
+++ b/state/wire.go
@@ -1,7 +1,7 @@
 package state

 import (
-       "github.com/tendermint/go-amino"
+       amino "github.com/tendermint/go-amino"
        cryptoAmino "github.com/tendermint/tendermint/crypto/encoding/amino"
 )
```

melekes: > in my local Tendermint repo it modifies almost 40 files

oh. yeah, some editors (like vim) do not understand that `go-amino` and `amino` are the same thing and need a mapping! feel free to create a PR 👍

thanethomson: Will do 👍

melekes:
It could be done, I think, by wrapping standard sync package, which I'd rather not do. But feel free to open an issue.

PR #3220, opened 2019-01-28
Minor fixes to non-hyperlinked links in docs

thanethomson:
There are a few links in the docs that aren't hyperlinks, so I just fixed these quick.

Plus, the `tm-bench` utility's docs were still pointing to the deprecated tools repo.


melekes:
Cool, thanks!

Issue #3196, opened 2019-01-23
Should we implement the rewind feature in tendermint?


bradyjoestar:
I was reading the code of go-ethereum and found that it had supported the rewind feature.
```
// SetHead rewinds the local chain to a new head. In the case of headers, everything
// above the new head will be deleted and the new one set. In the case of blocks
// though, the head may be further rewound if block bodies are missing (non-archive
// nodes after a fast sync).
func (bc *BlockChain) SetHead(head uint64) error {
	log.Warn("Rewinding blockchain", "target", head)

	bc.mu.Lock()
	defer bc.mu.Unlock()

	// Rewind the header chain, deleting all block bodies until then
	delFn := func(hash common.Hash, num uint64) {
		DeleteBody(bc.db, hash, num)
	}
	bc.hc.SetHead(head, delFn)
	currentHeader := bc.hc.CurrentHeader()

	// Clear out any stale content from the caches
	bc.bodyCache.Purge()
	bc.bodyRLPCache.Purge()
	bc.blockCache.Purge()
	bc.futureBlocks.Purge()

	// Rewind the block chain, ensuring we don't end up with a stateless head block
	if currentBlock := bc.CurrentBlock(); currentBlock != nil && currentHeader.Number.Uint64() < currentBlock.NumberU64() {
		bc.currentBlock.Store(bc.GetBlock(currentHeader.Hash(), currentHeader.Number.Uint64()))
	}
	if currentBlock := bc.CurrentBlock(); currentBlock != nil {
		if _, err := state.New(currentBlock.Root(), bc.stateCache); err != nil {
			// Rewound state missing, rolled back to before pivot, reset to genesis
			bc.currentBlock.Store(bc.genesisBlock)
		}
	}
	// Rewind the fast block in a simpleton way to the target head
	if currentFastBlock := bc.CurrentFastBlock(); currentFastBlock != nil && currentHeader.Number.Uint64() < currentFastBlock.NumberU64() {
		bc.currentFastBlock.Store(bc.GetBlock(currentHeader.Hash(), currentHeader.Number.Uint64()))
	}
	// If either blocks reached nil, reset to the genesis state
	if currentBlock := bc.CurrentBlock(); currentBlock == nil {
		bc.currentBlock.Store(bc.genesisBlock)
	}
	if currentFastBlock := bc.CurrentFastBlock(); currentFastBlock == nil {
		bc.currentFastBlock.Store(bc.genesisBlock)
	}
[...]

Issue #3181, opened 2019-01-21
Treat the validator updates from the application as an unordered set


ancazamfir:
When the application updates the validator set it calls updateValidators() with an 'updates' parameter, a list of changes that is then parsed in order, one entry at a time, making updates for the corresponding validator. i.e. the list is treated as a transaction list.
For example [add v1, rem v2, add v3] is different than [rem v2, add v1, add v3] and the call with one will produce different results than with the other.
This is an issue for gaia that requires these updates to be instead interpreted as a set.

kevlubkcm:
sounds related to #3073

cwgoes:
Proposal:

- First apply all updates to the validator set, without setting priorities. Track a list of addresses of newly added validators.
- Calculate the total voting power of the new validator set.
- Set the priorities of all newly added validators to `-C * total voting power` (presently `C = 1.125`).
- Normalize priorities if necessary & subtract the average.

ancazamfir:
> Proposal for atomic application:
First apply all updates to the validator set, without setting priorities. Track a list of addresses of newly added validators.

all updates except the validator removals
> Calculate the total voting power of the new validator set.
Set the priorities of all newly added validators to -C * total voting power (presently C = 1.125).

do validator removals here
>Normalize priorities if necessary & subtract the average.
Also we should consider if we need C = 1.125 after this change.

ebuchman:
Merged! Thanks @ancazamfir !

PR #3222, opened 2019-01-28
treat validator updates as set

ancazamfir:
* [ ] Wrote tests


types/validator_set.go
```diff
@@ -94,6 +97,20 @@ func (vals *ValidatorSet) IncrementProposerPriority(times int) {
 	vals.Proposer = proposer
 }
 
+/*
+
+func (vals *ValidatorSet) RescalePriorities() {
+	tvp := vals.TotalVotingPower()
+	diff := computeMaxMinPriorityDiff(vals)
+	if diff > K * tvp {
+		ratio := float64(K * tvp) / float64(diff)
```
Liamsi: There were some concerns around on using floats (probably on slack, I'll see if I can dig those up).

state/execution.go
```diff
@@ -448,7 +383,8 @@ func updateState(
 	// Update the validator set with the latest abciResponses.
 	lastHeightValsChanged := state.LastHeightValidatorsChanged
 	if len(validatorUpdates) > 0 {
-		err := updateValidators(nValSet, validatorUpdates)
+		//err := nValSet.UpdateWithChangeSet(validatorUpdates)
+		err := nValSet.UpdateWithChangeList(validatorUpdates)
```
Liamsi: Question: Is `UpdateWithChangeSet` only commented out because it would break some of the tests? This will eventually change to `UpdateWithChangeSet`, right?

types/validator_set.go
```diff
@@ -366,6 +383,266 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
 	}
 }
 
+// If more or equal than 1/3 of total voting power changed in one block, then
+// a light client could never prove the transition externally. See
+// ./lite/doc.go for details on how a light client tracks validators.
+func (currentSet *ValidatorSet) UpdateWithChangeList(updates []*Validator) error {
```
Liamsi: Will `UpdateWithChangeList` become obsolete?

types/validator_set_test.go
```diff
@@ -599,3 +599,217 @@ func TestValidatorSetVerifyCommit(t *testing.T) {
...
+	newVals := []*Validator{v3, v4}
+	assert.Nil(t, vals.UpdateWithChangeSet(newVals))
+	assert.Nil(t, verifyValidatorSet(vals))
+
+	// Delete validators
+	v3 = newValidator([]byte("v3"), 0)
+	v4 = newValidator([]byte("v4"), 0)
+	delVals := []*Validator{v3, v4}
+	assert.Nil(t, vals.UpdateWithChangeSet(delVals))
+	assert.Nil(t, verifyValidatorSet(vals))
```
Liamsi: This looks like this could be changed to [table-driven tests](https://github.com/golang/go/wiki/TableDrivenTests).

types/validator_set_test.go
```diff
@@ -599,3 +599,217 @@ func TestValidatorSetVerifyCommit(t *testing.T) {
...
+	v3 = newValidator([]byte("v3"), 3)
+	v5 := newValidator([]byte("v4"), 55)
+
+	chVals = []*Validator{v3, v5}
+	assert.Nil(t, vals1.UpdateWithChangeSet(chVals))
+	assert.Nil(t, verifyValidatorSet(vals1))
+
+	chVals = []*Validator{v5, v3}
+	assert.Nil(t, vals2.UpdateWithChangeSet(chVals))
+	assert.Nil(t, verifyValidatorSet(vals2))
```
Liamsi: This could be changed to: `assert.NoError`. In that way the reader immediately would notice, that  `verifyValidatorSet` returns an error.

Liamsi: Alternatively (as this is a test), you could also pass around `t` and do the asserts inside of `verifyValidatorSet(t *testing.T, vals2)` etc.

types/validator_set.go
```diff
@@ -366,6 +383,266 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+func processChanges(changes []*Validator) (error, []*Validator, []*Validator) {
+
+	sort.Sort(ValidatorsByAddress(changes))
+	removals := make([]*Validator, 0)
+	updates := make([]*Validator, 0)
+	var err error
+	prevAddr := []byte(nil)
+
+	for _, valUpdate := range changes {
+		if bytes.Equal(valUpdate.Address, prevAddr) {
```
Liamsi: Are the `changes` also expected to be sorted? If not, I do not think this would catch all duplicates. If they are in arbitrary order it is not enough to check against `prevAddr`.

types/validator_set_test.go
```diff
@@ -599,3 +599,217 @@ func TestValidatorSetVerifyCommit(t *testing.T) {
...
+		}
+		vals := NewValidatorSet(vs)
+		assert.Nil(b, verifyValidatorSet(vals))
+		len := len(vals.Validators)
+		// Add m validators
+		m := 2000
+		nvs := make([]*Validator, m)
+		for j := 0; j < m; j++ {
+			nvs[j] = newValidator([]byte(fmt.Sprintf("v%d", j+len)), 1000)
+		}
```
Liamsi: If I understand this correctly, this aims to benchmark the execution of `vals.UpdateWithChangeSet(nvs)`. If so, you could add a line here: 
```suggestion
		}
		b.ResetTimer()
```
As everything until line 813 is initialization for this.

Also, you can remove the assert below (it probably uses reflection and might slow down the benchmark results).

types/validator_set_test.go
```diff
@@ -599,3 +599,217 @@ func TestValidatorSetVerifyCommit(t *testing.T) {
...
+		n := 100
+		vs := make([]*Validator, n)
+		for j := 0; j < n; j++ {
+			vs[j] = newValidator([]byte(fmt.Sprintf("v%d", j)), 100)
+		}
+		vals := NewValidatorSet(vs)
+		assert.Nil(b, verifyValidatorSet(vals))
+		len := len(vals.Validators)
+		// Add m validators
+		m := 2000
```
Liamsi: I do not think we currently care for more than a few hundred/max 1000 validators (?).

types/validator_set.go
ancazamfir: Ok, please let me know. This will most likely be considered under https://github.com/tendermint/tendermint/pull/3182/
I just did some experimentation here. The main objective was to remove the diff from the parameters and contain the priority window size  here. Not sure the caller should decide on this.

types/validator_set.go
ancazamfir: Not sure if we want to keep both APIs. Until we decide I leave it here.

types/validator_set_test.go
ancazamfir: Ok, let me try this.

types/validator_set_test.go
ancazamfir: Will try both suggestions, thanks!

types/validator_set_test.go
ancazamfir: Understood, but it's good practice IMO to try high scale to understand what could go wrong. Also I experimented a bit with the go cpu profiler here :)

types/validator_set_test.go
Liamsi: My bad. Resetting the timer inside of the loop will make the benchmarks take forever (as they do not "converge").

Liamsi: The correct way to do this is probably to move the init code out of the "b.N-loop" (and wrap it with b.StopTimer & b.StartTimer)

state/execution.go
ancazamfir: I did did the change and fixed the test. It took a long time to fix these two in state_test.go
- TestProposerPriorityDoesNotGetResetToZero
- TestProposerPriorityProposerAlternates
Because the code now changes the penalty for newly added validators to updated total power (instead of previous total power) and because the centering is done before IncrementProposerPriority() it has been a bit of a nightmare to get the numbers right.
And in the end we manually match the numbers to the actual code we are testing which is not the best thing really. Ideally these tests should move to validator_set_test.go and instead of checking exact numbers check that centering and scaling achieved the desired results, that the selection is fair etc.

types/validator_set_test.go
ancazamfir: I made the change, please take a look.

state/execution.go
Liamsi: > and instead of checking exact numbers check that centering and scaling achieved the desired results, that the selection is fair etc.

I agree. Since introduced (as requested in a review) it was a PITA to update those with every change (and there were a few since then). Maybe it is fine to have one such test as some kind of regression test which can easily updated.

types/validator_set.go
```diff
@@ -74,35 +77,57 @@ func (vals *ValidatorSet) CopyIncrementProposerPriority(times int) *ValidatorSet
...
+*/
+
 func (vals *ValidatorSet) RescalePriorities(diffMax int64) {
+	if vals.IsNilOrEmpty() {
+		return
+	}
 	// NOTE: This check is merely a sanity check which could be
 	// removed if all tests would init. voting power appropriately;
 	// i.e. diffMax should always be > 0
 	if diffMax <= 0 {
```
Liamsi: related: https://github.com/tendermint/tendermint/pull/3049#discussion_r251341684

types/validator_set.go
Liamsi: Suggestion: I would still remove it, when it is not used anywhere. If the need arises to treat updates as a list, we can readd it, or, open up a way for the application to control how updates are treated.

types/validator_set.go
ancazamfir: removed

types/validator_set.go
ancazamfir: Should we fix this in https://github.com/tendermint/tendermint/pull/3182?

Liamsi: Yes, it's OK to this in  #3182 or in another PR.

state/execution.go
```diff
@@ -448,7 +383,8 @@ func updateState(
 	// Update the validator set with the latest abciResponses.
 	lastHeightValsChanged := state.LastHeightValidatorsChanged
 	if len(validatorUpdates) > 0 {
-		err := updateValidators(nValSet, validatorUpdates)
+		//err := nValSet.UpdateWithChangeSet(validatorUpdates)
```
Liamsi: delete this line
```suggestion
	
```

types/validator_set.go
```diff
@@ -366,6 +385,202 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+		}
+	}
+
+	return nil
+}
+
+// Merges the vals' validator list with the updates list.
+// When two elements with same address are seen, the one from updates is selected.
+// Expects updates to be a list of updates sorted by address with no duplicates,
+// and to have been validated with validateUpdates().
```
Liamsi: I can't find `validateUpdates`. This probably refers to `verifyUpdatesAndComputeNewPriorities`?

types/validator_set.go
```diff
@@ -366,6 +385,202 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+		prevAddr = valUpdate.Address
+	}
+	return updates, removals, nil
+}
+
+// Verifies a list of updates against a validator set, making sure the allowed
+// total voting power would not exceed if these updates would be applied to the set.
+// Computes the proposer priority for the validators not present in the set.
+// Returns error if first step fails.
+// updates parameter must be a list of unique validators to be added or updated.
```
Liamsi: Maybe add a comment that explains that this modifies the priorities of the passed `updates` but does not change `vals`?

types/validator_set.go
```diff
@@ -173,21 +185,28 @@ func (vals *ValidatorSet) getValWithMostPriority() *Validator {
...
 	avgProposerPriority := vals.computeAvgProposerPriority()
 	for _, val := range vals.Validators {
 		val.ProposerPriority = safeSubClip(val.ProposerPriority, avgProposerPriority)
 	}
 }
 
+// Makes a copy of the validator list
+func validatorListCopy(valsList []*Validator) []*Validator {
+	valsCopy := make([]*Validator, len(valsList))
+	for i, val := range(valsList) {
```
Liamsi:
```suggestion
	for i, val := range valsList {
```

types/validator_set.go
```diff
@@ -366,6 +385,202 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+		if val == nil {
+			return fmt.Errorf("failed to find validator %X to remove", address)
+		}
+	}
+	return nil
+}
+
+// Removes the validators specified in 'deletes' from validator set 'vals'.
+// Should not fail as verification has been done before.
+func (vals *ValidatorSet)applyRemovals(deletes []*Validator) {
```
Liamsi: Formatting seems off:
```suggestion
func (vals *ValidatorSet) applyRemovals(deletes []*Validator) {
```

Liamsi: You could add goimports or gofmt to your workflow/IDE to have the code formatted in an idiomatic way.

types/validator_set.go
```diff
@@ -366,6 +385,202 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+
+	// Apply updates and rescale
+	vals.applyUpdates(updates)
+	vals.RescalePriorities(K * vals.TotalVotingPower())
+
+	// Apply removals, rescale and recenter
+	vals.applyRemovals(deletes)
+	vals.RescalePriorities(K * vals.TotalVotingPower())
+	vals.shiftByAvgProposerPriority()
+
```
Liamsi: Question: If I understand correctly, the methods `ValidatorSet.Add`, `ValidatorSet.Update` and  `ValidatorSet.Remove` should not be used anywhere, or, only in tests. We probably still want to expose them? Or should we make them unexported?

types/validator_set.go
```diff
@@ -366,6 +385,202 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+	return nil
+}
+
+// Merges the vals' validator list with the updates list.
+// When two elements with same address are seen, the one from updates is selected.
+// Expects updates to be a list of updates sorted by address with no duplicates,
+// and to have been validated with validateUpdates().
+// The validator's priorities in 'updates' should be already computed.
+func (vals *ValidatorSet) applyUpdates(updates []*Validator) {
```
Liamsi: Even if this is a private method: can we have a simple test that show-cases how a given ValidatorSet.Validators is expected to look after applyUpdates?

types/validator_set.go
ancazamfir: Ouch...I just switched to trying GoLand (from VSCode) and forgot to add gofmt to the file watchers...thanks!

types/validator_set.go
ancazamfir: I removed them and changed the tests. Also I learnt today from @melekes about reflect.DeepEqual() so I changed the tests to use that.

types/validator_set.go
```diff
@@ -366,6 +383,266 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+// separates them in updates and deletes
+// - verifies that applying the updates will not result in errors (e.g. overflows)
+// - computes the priorities of validators that are added and/or changed against the final set
+// - verifies that applying the removals will not result in errors
+//   Note: currently an error is issued if a validator to be removed is not present in the set
+// - applies the updates against the validator set and performs scaling of priority values
+// - applies the removals against the validator set and performs scaling and centering of priority values
+// If error is detected during verification steps an error is returned and the validator set
+// is not changed.
+func (vals *ValidatorSet) UpdateWithChangeSet(changes []*Validator) error {
```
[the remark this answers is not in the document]
ancazamfir: :)

types/validator_set.go
Liamsi: cool! I think, `assert.Equal` uses `reflect.DeepEqual()` internally, too.

types/validator_set_test.go
ancazamfir: changed as suggested

types/validator_set.go
ancazamfir: added tests

state/state_test.go
```diff
@@ -322,7 +322,8 @@ func TestProposerFrequency(t *testing.T) {
 		vals := make([]*types.Validator, N)
 		totalVotePower := int64(0)
 		for j := 0; j < N; j++ {
-			votePower := int64(cmn.RandInt() % maxPower)
+			// make sure votePower > 0
+			votePower := int64(cmn.RandInt()%maxPower) + 1
```
Liamsi: Can't this (rarely) exceed maxPower now?

types/validator_set_test.go
```diff
@@ -602,195 +602,316 @@ func TestValidatorSetVerifyCommit(t *testing.T) {
...
-
-	chVals = []*Validator{v5, v3}
-	assert.NoError(t, vals2.UpdateWithChangeSet(chVals))
-	verifyValidatorSet(t, vals2)
-
-	assert.True(t, reflect.DeepEqual(vals1, vals2))
-}
-
-func getTotalProposerPriority(vals *ValidatorSet) int64 {
+type TestVal struct {
```
Liamsi: Could be `testVal` (unexported)

types/validator_set.go
```diff
@@ -42,19 +45,17 @@ type ValidatorSet struct {
...
 func NewValidatorSet(valz []*Validator) *ValidatorSet {
-	validators := make([]*Validator, len(valz))
-	for i, val := range valz {
-		validators[i] = val.Copy()
-	}
-	sort.Sort(ValidatorsByAddress(validators))
-	vals := &ValidatorSet{
-		Validators: validators,
+	vals := &ValidatorSet{}
+	err := vals.UpdateWithChangeSet(valz)
```
Liamsi: It's interesting that this doesn't change the behaviour now. IIUC, calling `NewValidatorSet` would initialize priorities with one call of `incrementProposerPriority` (basically setting all prios to the voting power of that validator and decrement the mostest). Now, we detect duplicates and additionally set all initial prios to -1.125*total (then rescale, shift by average via `UpdateWithChangeSet`) and then do the `incrementProposerPriority` (which rescales / shifts again).
The test suggest this doesn't have any side effects. And as we are not constantly calling `NewValidatorSet`, I think it is negligable to do the rescaling/shifting several times here.

types/validator_set.go
```diff
@@ -366,6 +335,204 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+	return updates, removals, nil
+}
+
+// Verifies a list of updates against a validator set, making sure the allowed
+// total voting power would not be exceeded if these updates would be applied to the set.
+// Computes the proposer priority for the validators not present in the set.
+// Returns error if first step fails.
+// 'updates' parameter must be a list of unique validators to be added or updated.
+// No changes are made to the validator set 'vals'.
+func verifyUpdatesAndComputeNewPriorities(updates []*Validator, vals *ValidatorSet) error {
```
Liamsi: How do you think about splitting this into two methods? `verifyUpdates` which only does the first loop and `computePriorities` which does the 2nd?

types/validator_set.go
Liamsi: If we do not want that, and to catch the duplicates, we could call the private helper functions (processChanges, verifyUpdates) here directly (would duplicate some of the code of `UpdateWithChangeSet`)

state/execution.go
```diff
@@ -448,7 +383,7 @@ func updateState(
 	// Update the validator set with the latest abciResponses.
 	lastHeightValsChanged := state.LastHeightValidatorsChanged
 	if len(validatorUpdates) > 0 {
-		err := updateValidators(nValSet, validatorUpdates)
+		err := nValSet.UpdateWithChangeSet(validatorUpdates)
```
milosevic: Like the name.

state/state_test.go
```diff
@@ -424,49 +425,71 @@ func TestProposerPriorityDoesNotGetResetToZero(t *testing.T) {
...
+
+	// 3. Center - noop
+	// 4. IncrementProposerPriority() ->
+	// v1(10):-9+10, v2(1):9+1 -> v2 proposer so substract tvp(11)
+	// v1(10):1, v2(1):-1
+	wantVal2Prio += updatedVal2.VotingPower // 10 -> prop
+	wantVal1Prio += updatedVal1.VotingPower // 1
+	wantVal2Prio -= totalPower              // -1
+
+	assert.Equal(t, wantVal2Prio, updatedVal2.ProposerPriority)
```
milosevic: This test is very big. Can it be split into several more focused one?

types/validator_set.go
```diff
@@ -19,7 +19,10 @@ import (
 // (Proof of 1 is tricky, left to the reader).
 // It could be higher, but this is sufficiently large for our purposes,
 // and leaves room for defensive purposes.
-const MaxTotalVotingPower = int64(math.MaxInt64) / 8
+const (
+	MaxTotalVotingPower = int64(math.MaxInt64) / 8
+	K                   = 2
```
milosevic: What is K?

types/validator_set.go
```diff
@@ -42,19 +45,17 @@ type ValidatorSet struct {
 // NewValidatorSet initializes a ValidatorSet by copying over the
 // values from `valz`, a list of Validators. If valz is nil or empty,
 // the new ValidatorSet will have an empty list of Validators.
+// The addresses of vlaidators in `valz` must be unique otherwise the
```
milosevic: Typo. vlaidators -> validators.

types/validator_set.go
```diff
@@ -366,6 +335,204 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
 	}
 }
 
+// Checks changes against duplicates, splits the changes in updates and removals, sorts them by address
+func processChanges(origChanges []*Validator) (updates, removals []*Validator, err error) {
```
milosevic: Maybe changes can be split into additions, removals and updates. The potential benefit is the fact that we can easily determine size of the new valset, and we can probably also get rid of `Remove` function (like we get rid of `Add`) function.

types/validator_set.go
ancazamfir: > It's interesting that this doesn't change the behaviour now. IIUC, calling `NewValidatorSet` would initialize priorities with one call of `incrementProposerPriority` (basically setting all prios to the voting power of that validator and decrement the mostest). Now, we detect duplicates and additionally set all initial prios to -1.125*total (then rescale, shift by average via `UpdateWithChangeSet`) and then do the `incrementProposerPriority` (which rescales / shifts again).

Yes true it's an overhead. But I think we will remove the rescale from incrementProposerPriority() once we know the priority window bound. 

> The tests suggest this doesn't have any side effects. And as we are not constantly calling `NewValidatorSet`, I think it is negligable to do the rescaling/shifting several times here.

Yes, this is what creates the validator set so only called once. After this we have the updates that do pretty much the same.

>If we do not want that, and to catch the duplicates, we could call the private helper functions (processChanges, verifyUpdates) here directly (would duplicate some of the code of `UpdateWithChangeSet`)

My concern is not really the extra rescaling/ centering. That I think will get better as I mentioned above. The problem right now is the different semantics of the parameter: - --
- NewValidatorSet() : add all validators in list, vs.
- UpdateWithChangeSet(): if validator exist either update or delete, if not add

So right now, when we call NewValidatorSet() with a validator with power 0 we fail (as we should) but the error says something like "cannot delete validator with address blah.."
I will fix this soon.

types/validator_set.go
```diff
@@ -366,6 +335,204 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+// separates them in updates and deletes
+// - verifies that applying the updates will not result in errors (e.g. overflows)
+// - computes the priorities of validators that are added and/or changed against the final set
+// - verifies that applying the removals will not result in errors
+//   Note: currently an error is issued if a validator to be removed is not present in the set
+// - applies the updates against the validator set and performs scaling of priority values
+// - applies the removals against the validator set and performs scaling and centering of priority values
+// If error is detected during verification steps an error is returned and the validator set
+// is not changed.
+func (vals *ValidatorSet) UpdateWithChangeSet(changes []*Validator) error {
```
milosevic: Beautiful code refactoring. I just have question regarding contract we have with ABCI app, i.e., what happens if an application returns changes which contains duplicates for example?

types/validator_set.go
ancazamfir: I did have the same thought but didn't want to carry over the info from verify into compute.
I realize that Go programmers really like small small functions :) Sooo...I will split

state/state_test.go
ancazamfir: yes, my nightmare on this pull. In one version I had most of the last part removed.

state/state_test.go
ancazamfir: yes, and I think with +1 we can get 1000 but no more.

Liamsi: You are right! Sorry for the noise.

types/validator_set.go
ancazamfir: I added an applyRemove2() function to show how it would look without Remove().

types/validator_set.go
ancazamfir: it's 2 :)...changed to a nice long name and added comment explaining.

types/validator_set.go
ancazamfir: split and rearranged.

types/validator_set.go
ancazamfir: Made some changes so we have better errors.

types/validator_set.go
ancazamfir: Yes, the error bubbles up:
UpdateWithChangeSet() -> execution.go:updateState() -> execution.go:ApplyBlock() -> reactor.go:poolRoutine() where we panic

We could probably discard the updates, notify the application and continue without the validator set changes.
It is also possible to discard duplicates and invalid changes and apply the rest good ones...do we want to do this?

types/validator_set.go
```diff
@@ -366,6 +338,260 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+// Should not fail as verification has been done before.
+func (vals *ValidatorSet) applyRemovals2(deletes []*Validator) {
+
+	existing := make([]*Validator, len(vals.Validators))
+	copy(existing, vals.Validators)
+
+	merged := make([]*Validator, len(existing)-len(deletes))
+	i := 0
+
+	for len(existing) > 0 && len(deletes) > 0 {
```
ebuchman: What do you think of this:

    func popVal(vals []*Validator) []*Validator{
        if len(vals) > 1 { return vals[1:] }
        return vals
    }

    // ...

    for i := 0; i < len(merged); i++ {
		for existing[0] == deletes[0] {
			existing, deletes = popVal(existing), popVal(deletes)
		}
		merged[i], existing = existing[0], popVal(existing)
	}

🐙 

I think we could do something similar for applyUpdates if we separated add from update

types/validator_set.go
ebuchman: Since we don't have a good way to "notify the application" right now, and since something like duplicate validators probably indicates a serious error in application logic, we should probably treat it as is and have tendermint halt. Of course this means we need to better support rollbacks for devs with buggy apps ... #3196

types/validator_set.go
ebuchman: Is `split into additions, removals and updates` related to `also get rid of Remove function` ? 

Also you guys inspired me to write the shorted applyRemove I could. Check it out below!

types/validator_set.go
```diff
@@ -366,6 +338,260 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+		return err
+	}
+
+	// Verify that applying the 'deletes' against 'vals' will not result in error.
+	if err = verifyRemovals(deletes, vals); err != nil {
+		return err
+	}
+
+	// Verify that applying the 'updates' against 'vals' will not result in error.
+	updatedTotalVotingPower, err := verifyUpdates(updates, vals)
```
ebuchman: don't we need to include changes from deletes in the updatedTotalVotingPower ?

types/validator_set.go
ebuchman: > split into additions, removals and updates

This sounds good, but it might make things more complicated in `updateWithChangeSet` since we have to do a few things for both additions and updates. Not sure what's best.

types/validator_set.go
```diff
@@ -366,6 +338,260 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+	// Compute the priorities for updates
+	computeNewPriorities(updates, vals, updatedTotalVotingPower)
+
+	// Apply updates and removals
+	vals.applyUpdates(updates)
+	vals.applyRemovals(deletes)
+	//vals.applyRemovals2(deletes)
+
+	// Scale and center
+	vals.RescalePriorities(PriorityWindowSizeFactor * vals.TotalVotingPower())
```
ebuchman: Do we need to do this here? We call IncrementProposerPriority after UpdateWithChangeSet so this sequence would be happening twice.

ebuchman:
I merged in develop to fix a build issue. But linter failing:

```
--> Running linter
types/validator_set_test.go:899:2: Consider preallocating `valList` (prealloc)
	var valList []*Validator
	^
state/state_test.go:486:47: `substract` is a misspelling of `subtract` (misspell)
	// v1(10):-9+10, v2(1):9+1 -> v2 proposer so substract tvp(11)
	                                             ^
types/validator_set.go:437:2: redundant return statement (gosimple)
	return
	^
types/validator_set.go:510:27: func `(*ValidatorSet).applyRemovals2` is unused (unused)
func (vals *ValidatorSet) applyRemovals2(deletes []*Validator) {
                          ^
Makefile:247: recipe for target 'lint' failed
```

Also some of my review comments are responses to previous conversations, which can be confusing if you don't realize. I guess we should try to avoid putting responses inside a new review.

types/validator_set.go
```diff
@@ -366,6 +338,260 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+// must have been validated with verifyUpdates() and priorities computed with computeNewPriorities().
+func (vals *ValidatorSet) applyUpdates(updates []*Validator) {
+
+	existing := make([]*Validator, len(vals.Validators))
+	copy(existing, vals.Validators)
+
+	merged := make([]*Validator, len(existing)+len(updates))
+	i := 0
+
+	for len(existing) > 0 && len(updates) > 0 {
```
melekes: a bit hard to follow. can't we just range over `updates`?

types/validator_set.go
```diff
@@ -366,6 +338,260 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+		merged[i] = existing[j]
+		i++
+	}
+	for j := 0; j < len(updates); j++ {
+		merged[i] = updates[j]
+		i++
+	}
+
+	vals.Validators = merged[:i]
+	vals.totalVotingPower = 0
```
melekes: why reset `totalVotingPower`? Function declaration does not say anything about that

types/validator_set.go
```diff
@@ -366,6 +338,260 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+			i++
+		}
+		existing = existing[1:]
+	}
+	for j := 0; j < len(existing); j++ {
+		merged[i] = existing[j]
+		i++
+	}
+	vals.Validators = merged[:i]
+	vals.totalVotingPower = 0
```
melekes: same question

types/validator_set.go
melekes: Why do all this? For speed? If yes, I'd rather prefer safer but slower version.

types/validator_set_test.go
```diff
@@ -599,3 +599,365 @@ func TestValidatorSetVerifyCommit(t *testing.T) {
...
+	return valList
+}
+
+func createNewValidatorSet(testValList []testVal) *ValidatorSet {
+	valList := createNewValidatorList(testValList)
+	valSet := NewValidatorSet(valList)
+	return valSet
+}
+
+func getTotalProposerPriority(valSet *ValidatorSet) int64 {
```
melekes: This function should probably go into ValidatorSet itself. I.e. `ValidatorSet#TotalProposerPriority()`

types/validator_set.go
ancazamfir: Ouch!!! Totally missed this! Thanks, thanks!!!!

ancazamfir: I will write a test for this soon

types/validator_set.go
ancazamfir: Yes we could. But if we find the limit on prio distance then we can remove it from IncrementProposerPriority() and only rescale and center after changes

types/validator_set.go
```diff
@@ -42,19 +48,17 @@ type ValidatorSet struct {
...
-	for i, val := range valz {
-		validators[i] = val.Copy()
-	}
-	sort.Sort(ValidatorsByAddress(validators))
-	vals := &ValidatorSet{
-		Validators: validators,
+	vals := &ValidatorSet{}
+	err := vals.updateWithChangeSet(valz, false)
+	if err != nil {
+		panic(fmt.Sprintf("cannot create validator set: %s", err))
```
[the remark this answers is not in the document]
ancazamfir: I prefer return error but this being a bigger change I wanted to hear from you.

ancazamfir:
@ebuchman and @melekes - thanks for the review!!!
There are still a few pending issues but before I move on them, let's agree on the error handling strategy.
Now that we control all operations in validator_set, what about:
- if NewValidatorSet(validators) results in an error, we return to caller and let it deal with it. Currently somewhere up the stack we panic...or if we go with the changes in https://github.com/tendermint/tendermint/pull/3241, we exit. In the future the caller might do something else
(Note: I haven't done this yet, just added panic for now since this is a bigger change I wanted to confirm with you)
- if UpdateWithChangeSet(changes) results in error same as above. I added check that final set is not empty
- if somehow we are called with any valSet.method() with empty set, we panic since, given the above we should never run in this situation.

types/validator_set.go
ancazamfir: Also to @milosevic earlier point, should I update the ABCI documentation? Part of this PR? Or someone else needs/ wants to do it?

types/validator_set.go
```diff
@@ -366,6 +338,260 @@ func (vals *ValidatorSet) Iterate(fn func(index int, val *Validator) bool) {
...
+}
+
+// Verifies a list of updates against a validator set, making sure the allowed
+// total voting power would not be exceeded if these updates would be applied to the set.
+//
+// Returns:
+// updatedTotalVotingPower - the new total voting power if these updates would be applied
+// err - non-nil if the maximum allowed total voting power would be exceeded
+//
+// No changes are made to the validator set 'vals'.
```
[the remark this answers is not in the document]
ancazamfir: yes

types/validator_set.go
ancazamfir: Hmm, actually it was correct before. Since we start with totalVotingPower and add the deltas for changes and power for adds.
But this proves that I am definitely missing explanations...and the test.

types/validator_set.go
milosevic: This is probably leftover of old implementation. But we should probably avoid doing it and just always set it to a correct value.

types/validator_set.go
ebuchman: Can be done separately, good point though.

ebuchman:
The error handling suggestions sound good. I think they can be done in a separate PR though so we can get this merged.

types/validator_set.go
ebuchman: But we don't subtract the deletes?

types/validator_set.go
ebuchman: Oh that's great. I guess it makes sense to leave it in both for now. Calling it twice in a row has the same effect as calling it once, right?

types/validator_set.go
milosevic: It seems I also misunderstand something. Shouldn't priority of new validator be equal to -1.25*totalVotingPower of the validator set before we apply any changes?

ancazamfir: No we don't, it's on purpose. So the idea is to forget the deletes initially. And determine how the set would look like without them. This way the relative priority of the existing validators is kept and we make sure we add the new validators with smallest priority (-TVP' of this imaginary set). This keeps things fair across this set.
Then we delete the required validators, scale and center. When we do that we do not change the relative priority of remaining validators and we make sure the prio window stays centered and within the 2*TVP
We discussed this with @cwgoes and @ismail in https://github.com/tendermint/tendermint/issues/3181.

types/validator_set.go
milosevic: We probably don't need to separate updates further, but I have other proposal. What if instead of `applyUpdates` and `applyRemovals` we have single function that just applies updates and computes nextValidatorSet. This function would start from empty val set (size is equal to existing.size + additions.size - deletions.size) and then iterate over existing valset and for each validator check if it is update, addition or deletion, and then do appropriate logic. For deletions we don't do anything; for updates and additions, we add it to the new set and compute proposer priority (in place; I am not sure `computeNewPriorites` function brings a lot). To make this possible, updates and deletions should be maps, and `processChanges` would need to return size of the new val set. Verify methods would just return error and not compute modified voting power as they do in the last version, so it might be a bit cleaner. `Remove` method is not needed anymore.

types/validator_set.go
ancazamfir: You mean calling it back-to-back? Yes, second time is a noop.

ancazamfir:
Ok, I just pushed the fix for total voting power.

types/validator_set.go
ancazamfir: @milosevic - yes it used to be like that and I documented it thinking it was intentional. However, when I brought this up to @cwgoes and @Liamsi they mentioned that it is in fact a bug. So I changed it here...doc still needs to be updated.

types/validator_set.go
ancazamfir: @melekes @milosevic ...yes I will look into it
valSet.TotalVotingPower() just returns valSet.totalVotingPower if it is non-zero. It should probably be enough to call valSet.TotalVotingPower() at the end of updateWithChangeSet()

state/state_test.go
```diff
@@ -483,7 +483,7 @@ func TestProposerPriorityDoesNotGetResetToZero(t *testing.T) {
 
 	// 3. Center - noop
 	// 4. IncrementProposerPriority() ->
-	// v1(10):-9+10, v2(1):9+1 -> v2 proposer so substract tvp(11)
+	// v1(10):-9+10, v2(1):9+1 -> v2 proposer so subsract tvp(11)
```
melekes: did you mean subtract?

PR #3227, opened 2019-01-29
pubsub 2.0

melekes:
* [ ] Wrote tests


consensus/common_test.go
```diff
@@ -219,23 +219,22 @@ func validatePrevoteAndPrecommit(t *testing.T, cs *ConsensusState, thisRound, lo
...
-func subscribeToVoter(cs *ConsensusState, addr []byte) chan interface{} {
-	voteCh0 := make(chan interface{})
-	err := cs.eventBus.Subscribe(context.Background(), testSubscriber, types.EventQueryVote, voteCh0)
+func subscribeToVoter(cs *ConsensusState, addr []byte) <-chan tmpubsub.Message {
+	voteCh0Sub, err := cs.eventBus.Subscribe(context.Background(), testSubscriber, types.EventQueryVote)
 	if err != nil {
 		panic(fmt.Sprintf("failed to subscribe %s to %v", testSubscriber, types.EventQueryVote))
 	}
-	voteCh := make(chan interface{})
+	ch := make(chan tmpubsub.Message)
```
melekes: why unbuffered channel?

consensus/common_test.go
```diff
@@ -219,23 +219,22 @@ func validatePrevoteAndPrecommit(t *testing.T, cs *ConsensusState, thisRound, lo
 }
 
 // genesis
-func subscribeToVoter(cs *ConsensusState, addr []byte) chan interface{} {
-	voteCh0 := make(chan interface{})
-	err := cs.eventBus.Subscribe(context.Background(), testSubscriber, types.EventQueryVote, voteCh0)
+func subscribeToVoter(cs *ConsensusState, addr []byte) <-chan tmpubsub.Message {
+	voteCh0Sub, err := cs.eventBus.Subscribe(context.Background(), testSubscriber, types.EventQueryVote)
```
melekes: `voteCh0Sub` does not tell anything. Rename to `voteSub`?

consensus/reactor_test.go
```diff
@@ -87,11 +91,11 @@ func stopConsensusNet(logger log.Logger, reactors []*ConsensusReactor, eventBuse
 func TestReactorBasic(t *testing.T) {
 	N := 4
 	css := randConsensusNet(N, "consensus_reactor_test", newMockTickerFunc(true), newCounter)
-	reactors, eventChans, eventBuses := startConsensusNet(t, css, N)
+	reactors, eventSubs, eventBuses := startConsensusNet(t, css, N)
```
melekes: rename to `blocksSub`. `eventSub` is too abstract!

consensus/reactor_test.go
```diff
@@ -485,18 +497,21 @@ func waitForAndValidateBlockWithTx(t *testing.T, n int, activeVals map[string]st
...
 		var newBlock *types.Block
 	LOOP:
 		for {
 			css[j].Logger.Debug("waitForBlockWithUpdatedValsAndValidateIt")
-			newBlockI, ok := <-eventChans[j]
-			if !ok {
-				return
-			}
-			newBlock = newBlockI.(types.EventDataNewBlock).Block
+			msg := <-eventSubs[j].Out()
```
melekes: check if it's OK to block here (_remember that Out() channel is never closed_)

consensus/replay_file.go
```diff
@@ -94,7 +82,8 @@ func (cs *ConsensusState) ReplayFile(file string, console bool) error {
 			return err
 		}
 
-		if err := pb.cs.readReplayMessage(msg, newStepCh); err != nil {
+		// TODO: pass a subscription
```
melekes:
```suggestion
```

consensus/replay_file.go
```diff
@@ -226,27 +215,16 @@ func (pb *playback) replayConsoleLoop() int {
...
-						break LOOP
-					}
-				}
-				pb.cs.eventBus.Unsubscribe(ctx, subscriber, types.EventQueryNewRoundStep)
-			}()
+			defer pb.cs.eventBus.Unsubscribe(ctx, subscriber, types.EventQueryNewRoundStep)
 
 			if len(tokens) == 1 {
-				if err := pb.replayReset(1, newStepCh); err != nil {
+				// TODO: pass a subscription
```
melekes:
```suggestion
```

consensus/replay_file.go
```diff
@@ -256,7 +234,8 @@ func (pb *playback) replayConsoleLoop() int {
 				} else if i > pb.count {
 					fmt.Printf("argument to back must not be larger than the current count (%d)\n", pb.count)
 				} else {
-					if err := pb.replayReset(i, newStepCh); err != nil {
+					// TODO: pass a subscription
```
melekes:
```suggestion
```

consensus/state_test.go
```diff
@@ -1292,8 +1292,9 @@ func (n *fakeTxNotifier) Notify() {
 }
 
 func TestStartNextHeightCorrectly(t *testing.T) {
+	config.Consensus.SkipTimeoutCommit = false
 	cs1, vss := randConsensusState(4)
-	cs1.config.SkipTimeoutCommit = false
+	// DATA RACE
```
melekes:
```suggestion
```

libs/pubsub/pubsub.go
```diff
@@ -10,44 +10,28 @@
...
-// 		     }
-//         s.UnsubscribeAll(ctx, sub)
-//     }()
-//     for msg := range out {
-//         // handle msg
-//         if err != nil {
-//            return err
-//         }
+//     q, err := query.New("account.name='John'")
+//		 if err != nil {
```
melekes: formatting

libs/pubsub/pubsub.go
```diff
@@ -174,34 +131,61 @@ func (s *Server) BufferCapacity() int {
...
+		outCap = outCapacity[0]
+	}
+
+	return s.subscribe(ctx, clientID, query, outCap)
+}
+
+// SubscribeUnbuffered does the same as Subscribe, except it returns a
+// subscription with unbuffered channel. Use with caution as it can freeze the
+// server.
+func (s *Server) SubscribeUnbuffered(ctx context.Context, clientID string, query Query) (*Subscription, error) {
```
melekes: unused

rpc/client/httpclient.go
```diff
@@ -381,8 +418,8 @@ func (w *WSEvents) eventListener() {
 			// NOTE: writing also happens inside mutex so we can't close a channel in
 			// Unsubscribe/UnsubscribeAll.
 			w.mtx.RLock()
-			if ch, ok := w.subscriptions[result.Query]; ok {
-				ch <- result.Data
+			if sub, ok := w.subscriptions[result.Query]; ok {
+				sub.out <- tmpubsub.NewMessage(result.Data, result.Tags)
```
melekes: ugh

state/txindex/indexer_service.go
```diff
@@ -31,35 +31,44 @@ func NewIndexerService(idr TxIndexer, eventBus *types.EventBus) *IndexerService
...
-				return
-			}
-			header := e.(types.EventDataNewBlockHeader).Header
-			batch := NewBatch(header.NumTxs)
-			for i := int64(0); i < header.NumTxs; i++ {
-				e, ok := <-txsCh
-				if !ok {
-					is.Logger.Error("Failed to index all transactions due to closed transactions channel", "height", header.Height, "numTxs", header.NumTxs, "numProcessed", i)
-					return
+			select {
```
melekes: write a test for this

types/event_bus_test.go
```diff
@@ -148,18 +139,19 @@ func TestEventBusPublish(t *testing.T) {
 	require.NoError(t, err)
 	defer eventBus.Stop()
 
-	eventsCh := make(chan interface{})
-	err = eventBus.Subscribe(context.Background(), "test", tmquery.Empty{}, eventsCh)
+	// FIXME: the test fails without a buffer
```
melekes: why??

state/execution_test.go
```diff
@@ -342,13 +341,15 @@ func TestEndBlockValidatorUpdates(t *testing.T) {
...
-		require.True(t, ok, "Expected event of type EventDataValidatorSetUpdates, got %T", e)
+	case msg := <-updatesSub.Out():
+		event, ok := msg.Data().(types.EventDataValidatorSetUpdates)
+		require.True(t, ok, "Expected event of type EventDataValidatorSetUpdates, got %T", msg.Data())
 		if assert.NotEmpty(t, event.ValidatorUpdates) {
 			assert.Equal(t, pubkey, event.ValidatorUpdates[0].PubKey)
 			assert.EqualValues(t, 10, event.ValidatorUpdates[0].VotingPower)
 		}
+	case <-updatesSub.Cancelled():
+		t.Fatal(fmt.Sprintf("updatesSub was cancelled (reason: %v)", updatesSub.Err()))
```
melekes: `t.Fatalf`

melekes:
```
=== RUN   TestIndexerServiceIndexesBlocks
I[2019-02-11|12:38:01.729] Starting EventBus                            impl=EventBus
I[2019-02-11|12:38:01.730] Starting PubSub                              module=pubsub impl=PubSub
I[2019-02-11|12:38:01.730] Starting IndexerService                      impl=IndexerService
E[2019-02-11|12:38:01.730] Failed to index block. txsSub was cancelled. Did the Tendermint stop? height=1 numTxs=2 numProcessed=1 err="client is not pulling messages fast enough"
I[2019-02-11|12:38:01.832] Stopping IndexerService                      impl=IndexerService
I[2019-02-11|12:38:01.832] Stopping EventBus                            impl=EventBus
I[2019-02-11|12:38:01.832] Stopping PubSub                              module=pubsub impl=PubSub
--- FAIL: TestIndexerServiceIndexesBlocks (0.10s)
    <autogenerated>:1: ^M                          ^M   Error Trace:    indexer_service_test.go:60
        ^M      Error:          Not equal:
        ^M                      expected: &types.TxResult{Height:1, Index:0x0, Tx:types.Tx{0x66, 0x6f, 0x6f}, Result:types.ResponseDeliverTx{Code:0x0, Data:[]uint8(nil), Log:"", Info:"", GasWanted:0, GasUsed:0, Tags:[]common.KVPair(nil), Codespace:"", XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}}
        ^M                      actual  : (*types.TxResult)(nil)
        ^M
        ^M                      Diff:
        ^M                      --- Expected
        ^M                      +++ Actual
        ^M                      @@ -1,7 +1,2 @@
        ^M                      -(*types.TxResult)({
        ^M                      - Height: (int64) 1,
        ^M                      - Index: (uint32) 0,
        ^M                      - Tx: (types.Tx) (len=3) Tx{666F6F},
        ^M                      - Result: (types.ResponseDeliverTx)
        ^M                      -})
        ^M                      +(*types.TxResult)(<nil>)
        ^M
        ^M      Test:           TestIndexerServiceIndexesBlocks
    <autogenerated>:1: ^M                          ^M   Error Trace:    indexer_service_test.go:63
        ^M      Error:          Not equal:
        ^M                      expected: &types.TxResult{Height:1, Index:0x1, Tx:types.Tx{0x62, 0x61, 0x72}, Result:types.ResponseDeliverTx{Code:0x0, Data:[]uint8(nil), Log:"", Info:"", GasWanted:0, GasUsed:0, Tags:[]common.KVPair(nil), Codespace:"", XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}}
        ^M                      actual  : (*types.TxResult)(nil)
        ^M
        ^M                      Diff:
        ^M                      --- Expected
        ^M                      +++ Actual
        ^M                      @@ -1,7 +1,2 @@
        ^M                      -(*types.TxResult)({
        ^M                      - Height: (int64) 1,
        ^M                      - Index: (uint32) 1,
        ^M                      - Tx: (types.Tx) (len=3) Tx{626172},
        ^M                      - Result: (types.ResponseDeliverTx)
        ^M                      -})
        ^M                      +(*types.TxResult)(<nil>)
        ^M
        ^M      Test:           TestIndexerServiceIndexesBlocks
```

libs/pubsub/pubsub_test.go
```diff
@@ -27,16 +27,49 @@ func TestSubscribe(t *testing.T) {
...
-	assertReceive(t, "Ka-Zar", ch)
+	assertReceive(t, "Ka-Zar", subscription.Out())
 
 	err = s.Publish(ctx, "Quicksilver")
 	require.NoError(t, err)
-	assertReceive(t, "Quicksilver", ch)
+	assertReceive(t, "Quicksilver", subscription.Out())
+}
+
+func TestSubscribeWithOutCapacity(t *testing.T) {
```
thanethomson:
```suggestion
func TestSubscribeWithoutCapacity(t *testing.T) {
```

rpc/lib/types/types.go
```diff
@@ -244,19 +245,19 @@ type WSRPCConnection interface {
...
-	UnsubscribeAll(ctx context.Context, subscriber string) error
-}
-
 // websocket-only RPCFuncs take this as the first parameter.
 type WSRPCContext struct {
 	Request RPCRequest
 	WSRPCConnection
 }
 
+// EventSubscriber mirros tendermint/tendermint/types.EventBusSubscriber
```
thanethomson:
```suggestion
// EventSubscriber mirrors tendermint/tendermint/types.EventBusSubscriber
```

libs/pubsub/pubsub.go
```diff
@@ -174,34 +131,61 @@ func (s *Server) BufferCapacity() int {
...
-func (s *Server) Subscribe(ctx context.Context, clientID string, query Query, out chan<- interface{}) error {
+// Subscribe creates a subscription for the given client.
+//
+// An error will be returned to the caller if the context is canceled or if
+// subscription already exist for pair clientID and query.
+//
+// outCapacity can be used to set a capacity for Subscription#Out channel (1 by
+// default). Panics if outCapacity is less than or equal to zero. If you want
+// an unbuffered channel, use SubscribeUnbuffered.
+func (s *Server) Subscribe(ctx context.Context, clientID string, query Query, outCapacity ...int) (*Subscription, error) {
```
Liamsi: What was the reason for `outCapacity` being a variadic parameter? As far as I can see only the 1st param is used here anyways?

libs/pubsub/subscription.go
```diff
@@ -0,0 +1,74 @@
...
+func (s *Subscription) Err() error {
+	s.mtx.RLock()
+	defer s.mtx.RUnlock()
+	return s.err
+}
+
+// Message glues data and tags together.
+type Message struct {
+	data interface{}
+	tags map[string]string
```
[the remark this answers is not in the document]
melekes: That's why it's better to merge PRs quickly or reject them. I don't remember the exact reason right now why I've removed TagMap interface

a) it was an unnecessary abstraction (we may switch to lists of lists)
b) ... something with RPC

Type alias sounds nice 👍

libs/pubsub/pubsub.go
melekes: It makes the last parameter optional

Subscribe(..., "account.number=1") // 1 by default
Subscribe(..., "account.number=1", 10) // 10 - capacity

CHANGELOG_PENDING.md
```diff
@@ -21,5 +22,6 @@ Special thanks to external contributors on this release:
 ### IMPROVEMENTS:
 
 ### BUG FIXES:
+- [libs/pubsub] \#951, \#1880 use non-blocking send when dispatching messages [ADR-33](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-033-pubsub.md)
 - [consensus] \#3297 Flush WAL on stop to prevent data corruption during
-  graceful shutdown
+  graceful shutdow
```
b00f: shutdow**n**!

libs/pubsub/subscription.go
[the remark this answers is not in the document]
ebuchman: We can use commit messages to better document this stuff. The message for this commit simply says `it does not bring any additional benefits`, so unclear if there was more reasoning.

Why does switching to list of lists justify removing this? Doesn't having more explicit types usually help with refactors like that?

libs/pubsub/pubsub.go
```diff
@@ -321,75 +302,60 @@ loop:
...
 	}
 
-	ch, ok := clientToChannelMap[clientID]
+	subscription, ok := clientSubscriptions[clientID]
 	if !ok {
 		return
 	}
 
-	close(ch)
+	subscription.mtx.Lock()
```
ebuchman: Can we expose a `subscription.CancelSubscription(err)` instead of calling all the subscription innards from a `state` method

libs/pubsub/pubsub.go
```diff
@@ -400,41 +366,38 @@ func (state *state) remove(clientID string, q Query) {
...
-		ch := state.queryToChanMap[qStr][clientID]
-		close(ch)
-
-		// remove the client from query map.
-		// if query has no other clients subscribed, remove it.
-		delete(state.queryToChanMap[qStr], clientID)
-		if len(state.queryToChanMap[qStr]) == 0 {
-			delete(state.queryToChanMap, qStr)
+func (state *state) removeClient(clientID string, reason error) {
+	for qStr, clientSubscriptions := range state.subscriptions {
```
ebuchman: Why did we get rid of the clientToQueryMap? Just for simplicity? Is it worth the cost of looping over all queries to remove a client?

libs/pubsub/pubsub.go
```diff
@@ -174,34 +131,61 @@ func (s *Server) BufferCapacity() int {
...
 	if ok {
 		_, ok = clientSubscriptions[query.String()]
 	}
 	s.mtx.RUnlock()
 	if ok {
-		return ErrAlreadySubscribed
+		return nil, ErrAlreadySubscribed
 	}
 
+	subscription := &Subscription{
```
ebuchman: `NewSubscription(outCapacity)` would be nice

libs/pubsub/pubsub.go
ebuchman: Let's open a new issue describing the use case for this (eg separating subscriptions to internal consensus events from subs to user events - the former used for a local consensus explorer and synchronous testing, the later for public facing / untrusted subscriptions). I think we had some notes on it from the meeting when we discussed all this.

libs/pubsub/pubsub_test.go
```diff
@@ -27,16 +27,49 @@ func TestSubscribe(t *testing.T) {
...
-	assertReceive(t, "Ka-Zar", ch)
+	assertReceive(t, "Ka-Zar", subscription.Out())
 
 	err = s.Publish(ctx, "Quicksilver")
 	require.NoError(t, err)
-	assertReceive(t, "Quicksilver", ch)
+	assertReceive(t, "Quicksilver", subscription.Out())
+}
+
+func TestSubscribeWithoutCapacity(t *testing.T) {
```
ebuchman: Why is it called WithoutCapacity? The cap is explicitly set here as 1.

libs/pubsub/pubsub_test.go
```diff
@@ -27,16 +27,49 @@ func TestSubscribe(t *testing.T) {
...
+	s := pubsub.NewServer()
+	s.SetLogger(log.TestingLogger())
+	s.Start()
+	defer s.Stop()
+
+	ctx := context.Background()
+	assert.Panics(t, func() {
+		s.Subscribe(ctx, clientID, query.Empty{}, -1)
```
ebuchman: Is this checking two distinct error cases? Seems like the second Subscribe never runs here since the first panics? Can we separate these into their own TestXxx function to test for Subscribe errors? cases?

libs/pubsub/pubsub_test.go
```diff
@@ -27,16 +27,49 @@ func TestSubscribe(t *testing.T) {
...
+}
+
+func TestSubscribeUnbuffered(t *testing.T) {
+	s := pubsub.NewServer()
+	s.SetLogger(log.TestingLogger())
+	s.Start()
+	defer s.Stop()
+
+	ctx := context.Background()
+	subscription, err := s.SubscribeUnbuffered(ctx, clientID, query.Empty{})
```
ebuchman: Wouldn't this test also pass if we use `Subscribe` with default capacity? Should we adapt it to only pass if the channel is unbuffered? And vice versa for the previous test.

state/txindex/indexer_service.go
```diff
@@ -31,35 +31,56 @@ func NewIndexerService(idr TxIndexer, eventBus *types.EventBus) *IndexerService
...
 		return err
 	}
 
-	txsCh := make(chan interface{})
-	if err := is.eventBus.Subscribe(context.Background(), subscriber, types.EventQueryTx, txsCh); err != nil {
+	// If there're no other subscribers, goroutine below may not pull messages
+	// fast enough. To solve it, we either need to increase the capacity here or
+	// change pubsub to wait a little bit (say, 10ms) before cancelling the
+	// subscription.
+	txsSub, err := is.eventBus.Subscribe(context.Background(), subscriber, types.EventQueryTx, 100)
```
ebuchman: Or maybe we should use SubscribeUnbuffered here? This is an internal subscription and we want to provide some level of reliability to our indexing, even if we don't to the realtime web socket subscriptions.

Alternatively, as discussed, maybe we don't need the indexing to be done through the event system - we can just give it the block and results in a go routine once processing is done.

libs/pubsub/subscription.go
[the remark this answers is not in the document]
melekes: > We can use commit messages to better document this stuff. The message for this commit simply says `it does not bring any additional benefits`

good point 👍 will try to put more effort into commit messages going forward

libs/pubsub/pubsub.go
melekes: For simplicity, yes. Plus we'll most probably get rid of pubsub, so that's fine.

libs/pubsub/subscription.go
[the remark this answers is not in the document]
melekes: > Type alias sounds nice

It may be a good idea to do it in a separate PR.

state/txindex/indexer_service.go
melekes: Switched to SubscribeUnbuffered

melekes:
Thanks @ebuchman for through review! Should be good now.

ebuchman:
Please no force pushes!

image

rpc/client/httpclient.go
```diff
@@ -249,23 +249,46 @@ func (c *HTTP) Validators(height *int64) (*ctypes.ResultValidators, error) {
 
 /** websocket event stuff here... **/
 
+type subscription struct {
```
ebuchman: Why can't we use the pubsub.Subscription?

libs/pubsub/subscription.go
```diff
@@ -0,0 +1,89 @@
...
+//   and the channel returned by Out became full,
+// After Err returns a non-nil error, successive calls to Err return the same
+// error.
+func (s *Subscription) Err() error {
+	s.mtx.RLock()
+	defer s.mtx.RUnlock()
+	return s.err
+}
+
+func (s *Subscription) cancel(err error) {
```
ebuchman: By convention, since this locks and mutates state, don't we want this method to be capitalized?

CHANGELOG_PENDING.md
```diff
@@ -7,6 +7,7 @@ Special thanks to external contributors on this release:
 ### BREAKING CHANGES:
 
 * CLI/RPC/Config
+- [httpclient] Update Subscribe interface to reflect new pubsub/eventBus API [ADR-33](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-033-pubsub.md)
```
ebuchman: I think the Subscribe interface change is breaking on the Go API, but the new behaviour of non-blocking send is breaking on the RPC since now clients need to be fast or they'll get cut!

rpc/client/httpclient.go
melekes: I've removed this in https://github.com/tendermint/tendermint/pull/3269. It should've been a single PR probably

libs/pubsub/subscription.go
melekes: Making it public will allow clients to call it, which I don't want.

Issue #2695, opened 2018-10-24
KVStore in Python


saikumar-neelam:
### How can We write KVStore Application in python?

Thanks in Advance for Suggestions

PR #3229, opened 2019-01-29
add python3 kvstoreapplication

gauthamzz:
* [ ] Wrote tests

I have ported kvstore to python 3. Please check if everything is alright according to the needs. 

#2695


abci/example/kvstore python3/Readme.md
```diff
@@ -0,0 +1,28 @@
+# KVStoreApplication
+
+The KVStoreApplication is a simple merkle key-value store. Transactions of the form key=value are stored as key-value pairs in the tree. Transactions without an = sign set the value to the key. The app has no replay protection (other than what the mempool provides).
```
melekes: > Transactions without an = sign set the value to the key

what key?

gauthamzz: I just copy pasted from kvstore docs 😅

gauthamzz: Is the logic of the kv store application code right ? Is this how you write a Tendermint application 😅 I’m still very unsure.

melekes:
Could you please rename `kvstore python3` folder to `python3_kvstore`?

abci/example/kvstore python3/Readme.md
melekes: I personally think you shouldn't store db in a state, but rather have application to store a link to it.

```
self.db = NewMemDB()
self.state = State.load_state(self.db)
```

gauthamzz:
> Could you please rename `kvstore python3` folder to `python3_kvstore`?

Done :santa:

abci/example/kvstore python3/Readme.md
gauthamzz: > I personally think you shouldn't store db in a state, but rather have application to store a link to it.
That's what is happening for now.

About implementing the persistant kvstore, is it fine if i use sqlite3 instead of leveldb. https://github.com/davebryson/py-tendermint/blob/master/tendermint/db.py 

>  Using sqlite3 for key/value store...what!?  Here's why:  existing tech (leveldb, codernitydb, etc...) either doesn't  support windows, not ported to Python 3, or both.

I haven't looked into this problem yet, so i'm not sure.

gauthamzz:
I thought so too, but the project doesn’t seem to be active last commit was 3 months ago. That’s why I had send pr here instead. 

Should I send pr there ?

melekes:
> Should I send pr there ?

Yes, please! If nobody responds, we'll do something else.

Issue #3230, opened 2019-01-29
Seed nodes drop all peers with pong timeout


zmanian:
Here is what I am observing.

When I run a node with `seed_node=true` and moderate to high incoming connection load.

Every peer stops on a `pong timeout` error.

```
E[2019-01-28|23:11:41.878] Stopping peer for error                      module=p2p peer="Peer{MConn{35.178.255.1:26656} 2da253df860538fd0adcc7fd79f89f088aca7353 out}" err="pong timeout"
E[2019-01-28|23:11:43.451] Connection failed @ sendRoutine              module=p2p peer=e47b9895c092122821b0f166501b594195054f43@8.8.8.8:26656 conn=MConn{52.52.28.101:26656} err="pong timeout"
E[2019-01-28|23:11:43.451] Stopping peer for error                      module=p2p peer="Peer{MConn{52.52.28.101:26656} e47b9895c092122821b0f166501b594195054f43 out}" err="pong timeout"
E[2019-01-28|23:11:44.132] Connection failed @ sendRoutine              module=p2p peer=5b2d308c2e24dfd1d7a89c8779621839c7bd0719@13.52.136.172:26656 conn=MConn{13.52.136.172:26656} err="pong timeout"
E[2019-01-28|23:11:44.133] Stopping peer for error                      module=p2p peer="Peer{MConn{13.52.136.172:26656} 5b2d308c2e24dfd1d7a89c8779621839c7bd0719 out}" err="pong timeout"
E[2019-01-28|23:11:44.794] Connection failed @ sendRoutine              module=p2p peer=5691f066133eaa40c9fdb0f2a76106bff06469bd@0.0.0.0:26656 conn=MConn{35.234.98.115:26656} err="pong timeout"
E[2019-01-28|23:11:44.794] Stopping peer for error                      module=p2p peer="Peer{MConn{35.234.98.115:26656} 5691f066133eaa40c9fdb0f2a76106bff06469bd out}" err="pong timeout"
E[2019-01-28|23:11:45.145] Connection failed @ sendRoutine              module=p2p peer=1cf82349127cd57ae8727d9161805bf9ecf4ad5e@77.87.106.26:26656 conn=MConn{77.87.106.26:26656} err="pong timeout"
E[2019-01-28|23:11:45.145] Stopping peer for error                      module=p2p peer="Peer{MConn{77.87.106.26:26656} 1cf82349127cd57ae8727d9161805bf9ecf4ad5e out}" err="pong timeout"
[...]

Issue #3204, opened 2019-01-24
non-deterministic test failure in TestWaitingTimeoutOnNilPolka


ebuchman:
Found on circle.

```
--- FAIL: TestWaitingTimeoutOnNilPolka (0.04s)
panic: Timeout expired while waiting for NewTimeout event [recovered]
	panic: Timeout expired while waiting for NewTimeout event

goroutine 39159 [running]:
testing.tRunner.func1(0xc009c46100)
	/usr/local/go/src/testing/testing.go:792 +0x6a7
panic(0xf7f7a0, 0xc00145f7e0)
	/usr/local/go/src/runtime/panic.go:513 +0x1b9
github.com/tendermint/tendermint/consensus.ensureNewEvent(0xc006dd8a80, 0x1, 0x0, 0x1c9c380, 0x10ccf8d, 0x32)
	/go/src/github.com/tendermint/tendermint/consensus/common_test.go:362 +0x61f
github.com/tendermint/tendermint/consensus.ensureNewTimeout(0xc006dd8a80, 0x1, 0x0, 0x989680)
	/go/src/github.com/tendermint/tendermint/consensus/common_test.go:434 +0x70
github.com/tendermint/tendermint/consensus.TestWaitingTimeoutOnNilPolka(0xc009c46100)
	/go/src/github.com/tendermint/tendermint/consensus/state_test.go:1110 +0x362
testing.tRunner(0xc009c46100, 0x1203548)
	/usr/local/go/src/testing/testing.go:827 +0x163
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:878 +0x65a
FAIL	github.com/tendermint/tendermint/consensus	57.070s
Exited with code 1
```

See logs [build_42381_step_109_container_3.txt](https://github.com/tendermint/tendermint/files/2792557/build_42381_step_109_container_3.txt)

ebuchman:
Duplicate of https://github.com/tendermint/tendermint/issues/2723

PR #3232, opened 2019-01-30
p2p: fix infinite loop in addrbook

ebuchman:
Here's a failing test and a quick fix for an issue we've seen a few times which manifests as a deadlock on seed nodes, eg. #3230 and #2955.

Basically the GetSelectionWithBias method has flawed logic that led to an infinite loop because it kept trying to get more "old" addresses even though there weren't any.  The method needs a larger refactor, and the whole address book might just need to be replaced, but I believe this solves the issue for now.

* [x] Wrote tests

### DIFF
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -414,19 +414,19 @@ func (a *addrBook) GetSelectionWithBias(biasTowardsNewAddrs int) []*p2p.NetAddre
 	var newAddressesAdded int
 
 	// number of new addresses that, if possible, should be in the beginning of the selection
-	numRequiredNewAdd := int(float64(biasTowardsNewAddrs)/float64(100) * float64(numAddresses))
+	numRequiredNewAdd := int(math.Round(float64(biasTowardsNewAddrs) / float64(100) * float64(numAddresses)))
 
 	selectionIndex := 0
 ADDRS_LOOP:
 	for selectionIndex < numAddresses {
 		// biasedTowardsOldAddrs indicates if the selection can switch to old addresses
-		biasedTowardsOldAddrs := selectionIndex > numRequiredNewAdd
+		biasedTowardsOldAddrs := selectionIndex >= numRequiredNewAdd
 		// An old addresses is selected if:
 		// - the bias is for old and old addressees are still available or,
 		// - there are no new addresses or all new addresses have been selected.
 		pickFromOldBucket :=
 			(biasedTowardsOldAddrs && oldAddressesAdded < a.nOld) || //
-			a.nNew == 0 || newAddressesAdded >= a.nNew
+				a.nNew == 0 || newAddressesAdded >= a.nNew
 
 		bucket := make(map[string]*knownAddress)
 
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -414,13 +414,13 @@ func (a *addrBook) GetSelectionWithBias(biasTowardsNewAddrs int) []*p2p.NetAddre
 	var newAddressesAdded int
 
 	// number of new addresses that, if possible, should be in the beginning of the selection
-	numRequiredNewAdd := int(float64(biasTowardsNewAddrs)/float64(100)) * numAddresses
+	numRequiredNewAdd := int(float64(biasTowardsNewAddrs)/float64(100) * float64(numAddresses))
 
 	selectionIndex := 0
 ADDRS_LOOP:
 	for selectionIndex < numAddresses {
 		// biasedTowardsOldAddrs indicates if the selection can switch to old addresses
-		biasedTowardsOldAddrs := selectionIndex >= numRequiredNewAdd
+		biasedTowardsOldAddrs := selectionIndex > numRequiredNewAdd
 		// An old addresses is selected if:
 		// - the bias is for old and old addressees are still available or,
 		// - there are no new addresses or all new addresses have been selected.
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -369,6 +369,10 @@ func (a *addrBook) GetSelection() []*p2p.NetAddress {
 	return allAddr[:numAddresses]
 }
 
+func percentageOfNum(p, n int) int {
+	return int(math.Round((float64(p) / float64(100)) * float64(n)))
+}
+
 // GetSelectionWithBias implements AddrBook.
 // It randomly selects some addresses (old & new). Suitable for peer-exchange protocols.
 // Must never return a nil address.
@@ -414,7 +418,7 @@ func (a *addrBook) GetSelectionWithBias(biasTowardsNewAddrs int) []*p2p.NetAddre
 	var newAddressesAdded int
 
 	// number of new addresses that, if possible, should be in the beginning of the selection
-	numRequiredNewAdd := int(math.Round(float64(biasTowardsNewAddrs) / float64(100) * float64(numAddresses)))
+	numRequiredNewAdd := percentageOfNum(biasTowardsNewAddrs, numAddresses)
 
 	selectionIndex := 0
 ADDRS_LOOP:
@@ -424,6 +428,8 @@ ADDRS_LOOP:
 		// An old addresses is selected if:
 		// - the bias is for old and old addressees are still available or,
 		// - there are no new addresses or all new addresses have been selected.
+		// numAddresses <= a.nOld + a.nNew therefore it is guaranteed that there are enough
+		// addresses to fill the selection
 		pickFromOldBucket :=
 			(biasedTowardsOldAddrs && oldAddressesAdded < a.nOld) || //
 				a.nNew == 0 || newAddressesAdded >= a.nNew
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -411,8 +411,12 @@ func (a *addrBook) GetSelectionWithBias(biasTowardsNewAddrs int) []*p2p.NetAddre
 	selectionIndex := 0
 ADDRS_LOOP:
 	for selectionIndex < numAddresses {
+
+		// determine whether to pick from an old bucket.
+		// if there's not enough old addresses to pick from, then we cant pick from old bucket.
 		pickFromOldBucket := int((float64(selectionIndex)/float64(numAddresses))*100) >= biasTowardsNewAddrs
-		pickFromOldBucket = (pickFromOldBucket && a.nOld > 0) || a.nNew == 0
+		pickFromOldBucket = (pickFromOldBucket && a.nOld > 0 && len(oldBucketToAddrsMap) < a.nOld) || a.nNew == 0
+
 		bucket := make(map[string]*knownAddress)
 
 		// loop until we pick a random non-empty bucket
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -413,12 +413,12 @@ ADDRS_LOOP:
 	for selectionIndex < numAddresses {
 
 		// determine whether to pick from an old bucket.
-		// if there's not enough old addresses to pick from, then we cant pick from old bucket.
-		pickFromOldBucket := int((float64(selectionIndex)/float64(numAddresses))*100) >= biasTowardsNewAddrs
-		pickFromOldBucket = (pickFromOldBucket && a.nOld > 0 && len(oldBucketToAddrsMap) < a.nOld) || a.nNew == 0
+		biasedTowardsOldAddrs := int((float64(selectionIndex)/float64(numAddresses))*100) >= biasTowardsNewAddrs
+		// if there's not enough old addresses to pick from, then we can't pick from old bucket.
+		pickFromOldBucket := biasedTowardsOldAddrs && a.nOld > len(oldBucketToAddrsMap)
 
 		// if there's not enough new addrs to pick from, just return early.
-		if !pickFromOldBucket && len(newBucketToAddrsMap) >= a.nNew {
+		if !pickFromOldBucket && a.nNew <= len(newBucketToAddrsMap) {
 			return selection
 		}
 
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -431,7 +431,7 @@ ADDRS_LOOP:
 		// numAddresses <= a.nOld + a.nNew therefore it is guaranteed that there are enough
 		// addresses to fill the selection
 		pickFromOldBucket :=
-			(biasedTowardsOldAddrs && oldAddressesAdded < a.nOld) || //
+			(biasedTowardsOldAddrs && oldAddressesAdded < a.nOld) ||
 				a.nNew == 0 || newAddressesAdded >= a.nNew
 
 		bucket := make(map[string]*knownAddress)
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -413,12 +413,12 @@ ADDRS_LOOP:
 	for selectionIndex < numAddresses {
 
 		// determine whether to pick from an old bucket.
-		biasedTowardsOldAddrs := int((float64(selectionIndex)/float64(numAddresses))*100) >= biasTowardsNewAddrs
-		// if there's not enough old addresses to pick from, then we can't pick from old bucket.
-		pickFromOldBucket := biasedTowardsOldAddrs && a.nOld > len(oldBucketToAddrsMap)
+		// if there's not enough old addresses to pick from, then we cant pick from old bucket.
+		pickFromOldBucket := int((float64(selectionIndex)/float64(numAddresses))*100) >= biasTowardsNewAddrs
+		pickFromOldBucket = (pickFromOldBucket && a.nOld > 0 && len(oldBucketToAddrsMap) < a.nOld) || a.nNew == 0
 
 		// if there's not enough new addrs to pick from, just return early.
-		if !pickFromOldBucket && a.nNew <= len(newBucketToAddrsMap) {
+		if !pickFromOldBucket && len(newBucketToAddrsMap) >= a.nNew {
 			return selection
 		}
 
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -408,19 +408,25 @@ func (a *addrBook) GetSelectionWithBias(biasTowardsNewAddrs int) []*p2p.NetAddre
 	newBucketToAddrsMap := make(map[int]map[string]struct{})
 	var newIndex int
 
+	// initialize counters used to count old and new added addresses.
+	// len(oldBucketToAddrsMap) cannot be used as multiple addresses can endup in the same bucket.
+	var oldAddressesAdded int
+	var newAddressesAdded int
+
+	// number of new addresses that, if possible, should be in the beginning of the selection
+	numRequiredNewAdd := int(float64(biasTowardsNewAddrs)/float64(100)) * numAddresses
+
 	selectionIndex := 0
 ADDRS_LOOP:
 	for selectionIndex < numAddresses {
-
-		// determine whether to pick from an old bucket.
-		// if there's not enough old addresses to pick from, then we cant pick from old bucket.
-		pickFromOldBucket := int((float64(selectionIndex)/float64(numAddresses))*100) >= biasTowardsNewAddrs
-		pickFromOldBucket = (pickFromOldBucket && a.nOld > 0 && len(oldBucketToAddrsMap) < a.nOld) || a.nNew == 0
-
-		// if there's not enough new addrs to pick from, just return early.
-		if !pickFromOldBucket && len(newBucketToAddrsMap) >= a.nNew {
-			return selection
-		}
+		// biasedTowardsOldAddrs indicates if the selection can switch to old addresses
+		biasedTowardsOldAddrs := selectionIndex >= numRequiredNewAdd
+		// An old addresses is selected if:
+		// - the bias is for old and old addressees are still available or,
+		// - there are no new addresses or all new addresses have been selected.
+		pickFromOldBucket :=
+			(biasedTowardsOldAddrs && oldAddressesAdded < a.nOld) || //
+			a.nNew == 0 || newAddressesAdded >= a.nNew
 
 		bucket := make(map[string]*knownAddress)
 
@@ -459,6 +465,7 @@ ADDRS_LOOP:
 				oldBucketToAddrsMap[oldIndex] = make(map[string]struct{})
 			}
 			oldBucketToAddrsMap[oldIndex][selectedAddr.String()] = struct{}{}
+			oldAddressesAdded++
 		} else {
 			if addrsMap, ok := newBucketToAddrsMap[newIndex]; ok {
 				if _, ok = addrsMap[selectedAddr.String()]; ok {
@@ -468,6 +475,7 @@ ADDRS_LOOP:
 				newBucketToAddrsMap[newIndex] = make(map[string]struct{})
 			}
 			newBucketToAddrsMap[newIndex][selectedAddr.String()] = struct{}{}
+			newAddressesAdded++
 		}
 
 		selection[selectionIndex] = selectedAddr
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -417,6 +417,11 @@ ADDRS_LOOP:
 		pickFromOldBucket := int((float64(selectionIndex)/float64(numAddresses))*100) >= biasTowardsNewAddrs
 		pickFromOldBucket = (pickFromOldBucket && a.nOld > 0 && len(oldBucketToAddrsMap) < a.nOld) || a.nNew == 0
 
+		// if there's not enough new addrs to pick from, just return early.
+		if !pickFromOldBucket && len(newBucketToAddrsMap) >= a.nNew {
+			return selection
+		}
+
 		bucket := make(map[string]*knownAddress)
 
 		// loop until we pick a random non-empty bucket
--- a/p2p/pex/addrbook_test.go
+++ b/p2p/pex/addrbook_test.go
@@ -8,252 +8,13 @@ import (
 	"os"
 	"testing"
 
-	"github.com/stretchr/testify/require"
-
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	cmn "github.com/tendermint/tendermint/libs/common"
 	"github.com/tendermint/tendermint/libs/log"
 	"github.com/tendermint/tendermint/p2p"
 )
-
-func createTempFileName(prefix string) string {
-	f, err := ioutil.TempFile("", prefix)
-	if err != nil {
-		panic(err)
-	}
-	fname := f.Name()
-	err = f.Close()
-	if err != nil {
-		panic(err)
-	}
-	return fname
-}
-
-func deleteTempFile(fname string) {
-	err := os.Remove(fname)
-	if err != nil {
-		panic(err)
-	}
-}
-
-func TestAddrBookGetSelectionWithOneMarkedGood(t *testing.T) {
-	// create a book with 10 addresses, 1 good/ old and 9 new
-	book := createAddrBookWithMOldAndNNewAddrs(t, 1, 9)
-
-	// ensure we can get selection.
-	// this used to result in an infinite loop (!)
-	addrs := book.GetSelectionWithBias(biasToSelectNewPeers)
-	assert.NotNil(t, addrs, "expected an address")
-	nAddr := len(addrs)
-	nOld, nNew := countOldAndNewInAddrs(addrs, book)
-	assert.Equal(t, nAddr, nOld+nNew,
-		"expected only non-nil addresses, got %v new, %v old, %v nil",
-		nNew, nOld, nAddr-nOld-nNew)
-}
-
-func TestAddrBookGetSelectionWithOneNotMarkedGood(t *testing.T) {
-	// create a book with 10 addresses, 9 good/ old and 1 new
-	book := createAddrBookWithMOldAndNNewAddrs(t, 9, 1)
-
-	// ensure we can get selection.
-	// this used to result in an infinite loop (!)
-	addrs := book.GetSelectionWithBias(biasToSelectNewPeers)
-	assert.NotNil(t, addrs, "expected an address")
-	nAddr := len(addrs)
-	nOld, nNew := countOldAndNewInAddrs(addrs, book)
-	assert.Equal(t, nAddr, nOld+nNew,
-		"expected only non-nil addresses, got %v new, %v old, %v nil",
-		nNew, nOld, nAddr-nOld-nNew)
-}
-
-// Creates a Book with m old and n new addresses.
-func createAddrBookWithMOldAndNNewAddrs(t *testing.T, m, n int) *addrBook {
-	fname := createTempFileName("addrbook_test")
-	defer deleteTempFile(fname)
-
-	book := NewAddrBook(fname, true)
-	book.SetLogger(log.TestingLogger())
-	assert.Zero(t, book.Size())
-
-	// add m old addresses
-	randAddrs := randNetAddressPairs(t, m)
-	for _, addr := range randAddrs {
-		book.AddAddress(addr.addr, addr.src)
-		book.MarkGood(addr.addr)
-	}
-	// add n new addresses
-	randAddrs = randNetAddressPairs(t, n)
-	for _, addr := range randAddrs {
-		book.AddAddress(addr.addr, addr.src)
-	}
-
-	return book
-}
-
-// Counts the number of Old and New addresses
-func countOldAndNewInAddrs(addrs []*p2p.NetAddress, book *addrBook) (nOld, nNew int) {
-
-	for _, addr := range addrs {
-		if addr == nil {
-			continue
-		}
-		if book.IsGood(addr) {
-			nOld++
-		} else {
-			nNew++
-		}
-	}
-	return nOld, nNew
-}
-
-// Analyse the layout of the selection specified by 'addrs'
-// Returns:
-// - seqLens - the lengths of the sequences of addresses of same type
-// - seqTypes - the types of sequences in selection
-func analyseSelectionLayout(book *addrBook, addrs []*p2p.NetAddress) (seqLens, seqTypes []int, err error) {
-	// address types are: 0 - nil, 1 - new, 2 - old
-	prevType := 0
-
-	currentSeqLen := 0
-	for i, addr := range addrs {
-		if addr == nil {
-			err = fmt.Errorf("nil addresses in selection, index %d", i)
-			return
-		}
-		addrType := 0
-		if book.IsGood(addr) {
-			addrType = 2
-		} else {
-			addrType = 1
-		}
-
-		if addrType != prevType && prevType != 0 {
-			seqLens = append(seqLens, currentSeqLen)
-			seqTypes = append(seqTypes, prevType)
-			currentSeqLen = 0
-		}
-		currentSeqLen++
-		prevType = addrType
-	}
-	seqLens = append(seqLens, currentSeqLen)
-	seqTypes = append(seqTypes, prevType)
-	return
-}
-
-func TestEmptyAddrBookSelection(t *testing.T) {
-	book := createAddrBookWithMOldAndNNewAddrs(t, 0, 0)
-	addrs := book.GetSelectionWithBias(biasToSelectNewPeers)
-	assert.Nil(t, addrs, "expected a nil selection")
-}
-
-func testAddrBookAddressSelection(t *testing.T, booksize int) {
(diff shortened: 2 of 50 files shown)


ebuchman:
Oh, need to also fix this for the opposite case!

ebuchman:
Why is the linter broken ?

melekes:
> Why is the linter broken ?

no idea

melekes:
fixed linter issue in #3233 and rebased against develop

p2p/pex/addrbook.go
```diff
@@ -411,8 +411,17 @@ func (a *addrBook) GetSelectionWithBias(biasTowardsNewAddrs int) []*p2p.NetAddre
 	selectionIndex := 0
 ADDRS_LOOP:
 	for selectionIndex < numAddresses {
+
+		// determine whether to pick from an old bucket.
+		// if there's not enough old addresses to pick from, then we cant pick from old bucket.
 		pickFromOldBucket := int((float64(selectionIndex)/float64(numAddresses))*100) >= biasTowardsNewAddrs
-		pickFromOldBucket = (pickFromOldBucket && a.nOld > 0) || a.nNew == 0
+		pickFromOldBucket = (pickFromOldBucket && a.nOld > 0 && len(oldBucketToAddrsMap) < a.nOld) || a.nNew == 0
```
Liamsi: Maybe that is overkill but can we used named vars for the conditions here?

sth along the lines of:
```go
isBiasedTowordsOld := int((float64(selectionIndex)/float64(numAddresses))*100) >= biasTowardsNewAddrs
isNoNewAddrs := a.nNew == 0
isOldAddrs := a.nOld > 0
notEnoughOld := len(oldBucketToAddrsMap) < a.nOld

pickFromOldBucket := isBiasedTowordsOld && isOldAddrs && notEnoughOld || isNoNewAddrs
```
(with  better naming)

It would clarify the logic here, makes the code speak for itself etc.

melekes:
```suggestion
		pickFromOldBucket = (pickFromOldBucket &&  a.nOld > len(oldBucketToAddrsMap)) || a.nNew == 0
```

q: what will happen if `a.nNew == 0` -> pickFromOldBucket=true, but we don't have enough items (`a.nOld > len(oldBucketToAddrsMap)` false)?

melekes: I think it should be 

`pickFromOldBucket = (pickFromOldBucket || a.nNew == 0) && a.nOld > len(oldBucketToAddrsMap)`

melekes: Note: tests pass for the above code too!

ancazamfir: I tried to understand this code today for the first time so please bear with me a bit.

"return early" - means there will be less than numAddresses returned, might break this APIs "contract".

Then, if all good addresses (i.e. a.nNew == 0) and pickFromOldBucket is false (because of bias, e.g.first iteration), algo will return right away with empty selection (to be precise, all map entries will point to nil addresses)

I think we should also add tests for all good, all bad, and different sizes so we hit the different cases for numAddresses (bookSize, minGetSelection, bookSize*getSelectionPercent/100). And check that the selection contains in fact the promised number.

melekes: > "return early" - means there will be less than numAddresses returned, might break this APIs "contract".

GetSelectionWithBias() can return an empty collection. We don't state that it's always returns N items.

ancazamfir: The problem is we don't state either way. Before this fix, if I have a book of 10 old/ good addresses, and 0 new/ bad, I will get back the 10 old.
With the fix I get back 0. So fixing the infinite loop changed the behavior.

ancazamfir: Ok, there is this comment:
// GetSelectionWithBias implements AddrBook.
// It randomly selects some addresses (old & new). Suitable for peer-exchange protocols.
**// Must never return a nil address.**

With the example I gave above we return a list of nil addresses.

ancazamfir: GetSelectionWithBias() returns []*p2p.NetAddress, i.e. a list of pointers to addresses.
- The function determines the number of addresses to return based on the booksize. For small books it will be:
 numAddresses  = a.nNew + a.nOld. 
- Then it allocates that many entries in the list:
selection := make([]*p2p.NetAddress, numAddresses).
I believe the comment: "Must never return a nil address." means that after it runs, all these pointers must point to non-nil addresses.
- Then it starts the loop and gives preference to new addresses if there are any and randomly selects new addresses. At some point it will switch to old bucket and start picking old addresses.

The problem with the latest code in bucky/3230-fix-addrbook-loop is that now the selection only picks new addresses and it will never pick old addresses.

I believe the first commit from Ethan works fine.

ancazamfir: correction: "it will never pick old addresses." in some cases...in other cases it will do. 
It all depends on the relation between biasToSelectNewPeers and the a.nNew vs a.nOld values.
I think that if there are not enough new addresses (< ~ biasToSelectNewPeers % of total), the old addresses will not be picked.
I don't know, maybe this was the intent all along but the old code did not do that.

I think we need to specify the desired behavior clearly. And then write some code that can convince the reader of correctness via inspection also and not only exhaustive testing.

ancazamfir: Let' say we decided that we want to pick r (aka numAddresses) addresses.
I think the original intent was:
- pick i new addresses with i=min(n, k), n total number of new addresses, k number based on bias calculation (percentage of r that should be new addresses in the beginning of selection)
- pick j old addresses with j=min(m, r-i)
- i+j < r, pick r-i-j new addresses
(all picks are random)

So we end up with:
```
indices:      0...i-1   i...i+j-1    i+j...r
addresses:    N0..Ni-1  O0..Oj-1     Ni...
```

I will write some testcases for this tomorrow to check if it's true :)

melekes: > I believe the first commit from Ethan works fine.

ok, reverted my commit

ancazamfir: With the code from develop:
---------------------------
The algorithm tries to find exactly:

- k new addresses with k ~ bias/100 * r, AND
- r-i old addresses 

When it works, it can only generate:
```
indices:      0...k-1   k...r
addresses:    N0..Nk-1  O0..Or-i-1
```
This is the sequence where we have exactly bias% new addresses and the rest are old.

If the book does NOT have minimum k new addresses AND minimum r-i old addresses, the algo will enter the infite loop.

The exceptions are
- if there are 0 new addresses:
```
indices:      0...r
addresses:    O0..Or-1
```
- if there are 0 old addresses:
```
indices:      0...r
addresses:    N0..Nr-1
```

To give you an idea:
For books of size 100, the computed selection size r is 32. 
For a bias of 30%, the algo needs 10 new addresses and 22 old addresses.
So it will enter infinite loop for any case where number of old addresses m is in {1..21, 91..99}
(32 out of 100 possible combos will enter the infinte loop) 

With the last few fixes:
------------------------
The number of hogs decreased but was not completely eliminated. And now they were also not always reproducible.
The reason is the check involving len(newBucketToAddrsMap) or len(oldBucketToAddrsMap) is not correct, as many addresses are sometimes mapped to the same bucket, so in these cases len(..BucketToAddrsMap returns less than the number of addresses that were added, causing the infinite loop again.


I just committed a fix that maintains separate counters for old and new. I added a test that can take multiple booksizes and run this over all possible combinations. 
The tests need more work, some of the old tests are failing. I am looking at those now.

ancazamfir: Fixed mistake in bias to old computation ( changed to float division instead of int)

ancazamfir:
Bon, now either lint or make complain of math.Round. 
Also TestWaitingTimeoutOnNilPolka fails in the checks.
All is fine on my computer (ran all > 100 times). Any ideas?

Liamsi:
On `TestWaitingTimeoutOnNilPolka`: https://github.com/tendermint/tendermint/issues/3204
Not sure what is going on with math.Round though.

Liamsi:
As `undefined: math.Round` only seem to happen for the `localnet` job, it seems related to the environment there which uses `circleci/classic:latest`. Apparently, `circleci/classic:latest` uses golang 1.10 (different from the other jobs which use the latest golang 1.11). 1.10 is the version where math.Round was [introduced](https://golang.org/doc/go1.10#math) 🤷‍♂️

p2p/pex/addrbook.go
```diff
@@ -408,11 +408,26 @@ func (a *addrBook) GetSelectionWithBias(biasTowardsNewAddrs int) []*p2p.NetAddre
...
 	for selectionIndex < numAddresses {
-		pickFromOldBucket := int((float64(selectionIndex)/float64(numAddresses))*100) >= biasTowardsNewAddrs
-		pickFromOldBucket = (pickFromOldBucket && a.nOld > 0) || a.nNew == 0
+		// biasedTowardsOldAddrs indicates if the selection can switch to old addresses
+		biasedTowardsOldAddrs := selectionIndex >= numRequiredNewAdd
+		// An old addresses is selected if:
+		// - the bias is for old and old addressees are still available or,
+		// - there are no new addresses or all new addresses have been selected.
+		pickFromOldBucket :=
+			(biasedTowardsOldAddrs && oldAddressesAdded < a.nOld) || //
```
melekes:
```suggestion
			(biasedTowardsOldAddrs && oldAddressesAdded < a.nOld) ||
```

p2p/pex/addrbook.go
```diff
@@ -408,11 +408,26 @@ func (a *addrBook) GetSelectionWithBias(biasTowardsNewAddrs int) []*p2p.NetAddre
...
 ADDRS_LOOP:
 	for selectionIndex < numAddresses {
-		pickFromOldBucket := int((float64(selectionIndex)/float64(numAddresses))*100) >= biasTowardsNewAddrs
-		pickFromOldBucket = (pickFromOldBucket && a.nOld > 0) || a.nNew == 0
+		// biasedTowardsOldAddrs indicates if the selection can switch to old addresses
+		biasedTowardsOldAddrs := selectionIndex >= numRequiredNewAdd
+		// An old addresses is selected if:
+		// - the bias is for old and old addressees are still available or,
+		// - there are no new addresses or all new addresses have been selected.
+		pickFromOldBucket :=
```
melekes: What if there are not enough new & old addresses? In the first iteration, we'd return early thereby escaping an infinite loop. Could you explain how this works now?

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
...
+		book.AddAddress(addr.addr, addr.src)
+	}
+
+	// ensure we can get selection.
+	// this used to result in an infinite loop (!)
+	addrs := book.GetSelectionWithBias(biasToSelectNewPeers)
+	assert.NotNil(t, addrs, "expected an address")
+}
+
+func min(a, b int) int {
```
melekes: `cmn.MinInt` https://github.com/tendermint/tendermint/blob/d47094550315c094512a242445e0dde24b5a03f5/libs/common/math.go#L26

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
...
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func listsAreEqual(a, b []int) bool {
```
melekes: You can use https://golang.org/pkg/reflect/#DeepEqual in tests

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
 	}
 }
 
+func TestAddrBookGetSelectionWithOneMarkedGood(t *testing.T) {
+	fname := createTempFileName("addrbook_test")
```
melekes: can we use `createAddrBook` here?

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
...
+	}
+
+	// ensure we can get selection.
+	// this used to result in an infinite loop (!)
+	addrs := book.GetSelectionWithBias(biasToSelectNewPeers)
+	assert.NotNil(t, addrs, "expected an address")
+}
+
+func TestAddrBookGetSelectionWithOneNotMarkedGood(t *testing.T) {
+	fname := createTempFileName("addrbook_test")
```
melekes: can we use `createAddrBook` here?

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
...
+	for i := range a {
+		if a[i] != b[i] {
+			return false
+		}
+	}
+	return true
+}
+
+// Creates a Book with m old and n new addresses.
+func createAddrBook(t *testing.T, m, n int) *addrBook {
```
melekes:
```suggestion
func createAddrBookWithMOldAndNNewAddrs(t *testing.T, m, n int) *addrBook {
```

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
...
+
+	return book
+}
+
+// Computes n * bias/100
+func numExpectedAddresses(bias, n int) int {
+	if n == 0 {
+		return 0
+	}
+	return int(math.Round((float64(bias) / float64(100)) * float64(n)))
```
melekes: We should not duplicate code logic in tests.

```
numRequiredNewAdd := int(math.Round(float64(biasTowardsNewAddrs) / float64(100) * float64(numAddresses)))
```

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
...
+}
+
+// Counts the number of Old and New addresses
+func countOldAndNewInAddrs(addrs []*p2p.NetAddress, book *addrBook) (nold, nnew int) {
```
melekes:
```suggestion
func countOldAndNewInAddrs(addrs []*p2p.NetAddress, book *addrBook) (nOld, nNew int) {
```

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
...
+	nold = 0
```
melekes: no need to initialize these vars (they are 0 by default)

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
...
+	nnew = 0
+
+	for _, addr := range addrs {
+		if addr == nil {
+			continue
+		}
+		if book.IsGood(addr) {
```
melekes: good != old

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
...
+			nnew++
+		}
+	}
+	return nold, nnew
+}
+
+func analyseSelectionLayout(book *addrBook, addrs []*p2p.NetAddress) ([]int, []int, error) {
+	// address types are: 0 - nil, 1 - new, 2 - old
+	prevt := 0
+	var seqLens []int
```
melekes: what's seqLens?

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
...
+		if book.IsGood(addr) {
+			nold++
+		} else {
+			nnew++
+		}
+	}
+	return nold, nnew
+}
+
+func analyseSelectionLayout(book *addrBook, addrs []*p2p.NetAddress) ([]int, []int, error) {
```
melekes: what are the return parameters?

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
...
+	book := createAddrBook(t, 0, 0)
+	addrs := book.GetSelectionWithBias(biasToSelectNewPeers)
+	assert.Nil(t, addrs, "expected a nill selection")
+}
+
+func testAddrBookAddressSelection(t *testing.T, b int) {
+	// generate all combinations of old (m) and new (n=b-m) addresses
+	for m := 0; m < b+1; m++ {
+		n := b - m
+		dbgstr := fmt.Sprintf("book: size %d, number new % d, old %d", b, n, m)
```
melekes: what's dbgstr?

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
...
+		// create book and get selection
+		book := createAddrBook(t, m, n)
+		addrs := book.GetSelectionWithBias(biasToSelectNewPeers)
+
+		r := len(addrs)
+		assert.NotNil(t, addrs, "%s - expected a non-nill selection", dbgstr)
+		assert.NotZero(t, r, "%s - expected at least one address in selection", dbgstr)
+
+		// verify that the number of old and new addresses are the ones expected
+		numold, numnew := countOldAndNewInAddrs(addrs, book)
```
melekes: https://golang.org/doc/effective_go.html#mixed-caps

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
...
+	return seqLens, seqTypes, nil
+}
+
+func TestEmptyAddrBookSelection(t *testing.T) {
+	book := createAddrBook(t, 0, 0)
+	addrs := book.GetSelectionWithBias(biasToSelectNewPeers)
+	assert.Nil(t, addrs, "expected a nill selection")
+}
+
+func testAddrBookAddressSelection(t *testing.T, b int) {
```
melekes: this test (as of now) is a bit complicated and will be hard to modify in the future

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,264 @@ func deleteTempFile(fname string) {
...
+	for b := 1; b < N; b++ {
+		testAddrBookAddressSelection(t, b)
+	}
+	// Test for three books with sizes from following ranges
+	ranges := [...][]int{{33, 100}, {100, 175}}
+	var bsizes []int
+	for _, r := range ranges {
+		bsizes = append(bsizes, cmn.RandIntn(r[1]-r[0])+r[0])
+	}
+	fmt.Printf("Testing address selection for the following booksizes %v\n", bsizes)
```
melekes:
```suggestion
	t.Logf("Testing address selection for the following booksizes %v\n", bsizes)
```

p2p/pex/addrbook_test.go
melekes: https://talks.golang.org/2014/names.slide#1

p2p/pex/addrbook_test.go
ebuchman: How so?

```
func (a *addrBook) IsGood(addr *p2p.NetAddress) bool {
	a.mtx.Lock()
	defer a.mtx.Unlock()

	return a.addrLookup[addr.ID].isOld()
}

```

p2p/pex/addrbook.go
ebuchman: There must be enough based on the way numAddresses, was computed, ie. `numAddresses <= nOld + nNew`

p2p/pex/addrbook_test.go
melekes: What I am trying to say is it might change in the future (e.g. good addr - old addr whos been old for a month), so maybe it's better to create a `IsOld(addr *p2p.NetAddress)` function.

ancazamfir:
@melekes Thanks for the review, I am addressing the comments now.

p2p/pex/addrbook_test.go
ancazamfir: Not sure what to do here. The IsGood() function is only used by addrbook_test.go at this point and not only in the newly added tests. I think it's a good opportunity to change its name to isOld() as I found it confusing. Let me know.

p2p/pex/addrbook_test.go
ancazamfir: I looked at how to make it better but short of cutting out from the checks, couldn't find a way. I am opened to suggestions.

p2p/pex/addrbook.go
ancazamfir: added comment to clarify

p2p/pex/addrbook_test.go
ancazamfir: yup, done

p2p/pex/addrbook_test.go
ancazamfir: fixed

p2p/pex/addrbook_test.go
ancazamfir: renamed dbgStr and added comment.

p2p/pex/addrbook_test.go
ancazamfir: fixed

p2p/pex/addrbook_test.go
ancazamfir: yes, done.

p2p/pex/addrbook_test.go
ancazamfir: changed

p2p/pex/addrbook_test.go
ancazamfir: fixed

p2p/pex/addrbook_test.go
ancazamfir: removed

p2p/pex/addrbook_test.go
ancazamfir: added comment to clarify

p2p/pex/addrbook_test.go
ancazamfir: added percentageOfNum() in addrbook.go.

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +38,202 @@ func deleteTempFile(fname string) {
...
+			seqLens = append(seqLens, cnt)
+			seqTypes = append(seqTypes, prevt)
+			cnt = 0
+		}
+		cnt++
+		prevt = addrt
+	}
+	seqLens = append(seqLens, cnt)
+	seqTypes = append(seqTypes, prevt)
+	return seqLens, seqTypes, nil
```
melekes:
```suggestion
	return
```

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +38,202 @@ func deleteTempFile(fname string) {
...
+	return nOld, nNew
+}
+
+// Analyse the layout of the selection specified by 'addrs'
+// Returns:
+// - seqLens - the lengths of the sequences of addresses of same type
+// - seqTypes - the types of sequences in selection
+func analyseSelectionLayout(book *addrBook, addrs []*p2p.NetAddress) (seqLens, seqTypes []int, err error) {
+	// address types are: 0 - nil, 1 - new, 2 - old
+	prevt := 0
```
melekes: what is prevt? prevType?

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +38,202 @@ func deleteTempFile(fname string) {
...
+
+	cnt := 0
+	for i, addr := range addrs {
+		if addr == nil {
+			err = fmt.Errorf("nil addresses in selection, index %d", i)
+			return nil, nil, err
```
melekes:
```suggestion
			return
```

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +38,202 @@ func deleteTempFile(fname string) {
...
+		}
+		addrt := 0
```
melekes:
```suggestion
		addrType := 0
```

clarity/readability is more important than how short the name is

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +38,202 @@ func deleteTempFile(fname string) {
...
+
+// Analyse the layout of the selection specified by 'addrs'
+// Returns:
+// - seqLens - the lengths of the sequences of addresses of same type
+// - seqTypes - the types of sequences in selection
+func analyseSelectionLayout(book *addrBook, addrs []*p2p.NetAddress) (seqLens, seqTypes []int, err error) {
+	// address types are: 0 - nil, 1 - new, 2 - old
+	prevt := 0
+
+	cnt := 0
```
melekes: count? I see it can be reset to 0. maybe there's a better name. `seqLen`? `currentSeqLen`?

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +38,202 @@ func deleteTempFile(fname string) {
...
+		book := createAddrBookWithMOldAndNNewAddrs(t, m, n)
+		addrs := book.GetSelectionWithBias(biasToSelectNewPeers)
+
+		r := len(addrs)
+		assert.NotNil(t, addrs, "%s - expected a non-nill selection", dbgStr)
+		assert.NotZero(t, r, "%s - expected at least one address in selection", dbgStr)
+
+		// verify that the number of old and new addresses are the ones expected
+		numOld, numNew := countOldAndNewInAddrs(addrs, book)
+		assert.Equal(t, r, numOld+numNew, "%s - expected selection completely filled", dbgStr)
```
melekes: Do we actually check anything here? addr is either old or new, so sum of numOld + numNew will always be equal to len(addrs), no?

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +38,202 @@ func deleteTempFile(fname string) {
...
+		var expSeqTypes []int
+		if j == 0 {
+			// all new
+			expSeqLens = []int{r}
+			expSeqTypes = []int{1}
+		} else if i == 0 {
+			// all old
+			expSeqLens = []int{r}
+			expSeqTypes = []int{2}
+		} else if r-i-j == 0 {
```
melekes: hard to follow :( maybe you can use a switch and add a comment for each case

```go
switch {
  case j == 0: // all new addresses
  case i == 0: // all old addresses
  case r-i-j == 0: // ????
  default: // ????
}
```

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +38,202 @@ func deleteTempFile(fname string) {
...
+
+func TestMultipleAddrBookAddressSelection(t *testing.T) {
+	// test books with smaller size, < N
+	N := 32
+	for b := 1; b < N; b++ {
+		testAddrBookAddressSelection(t, b)
+	}
+	// Test for two books with sizes from following ranges
+	ranges := [...][]int{{33, 100}, {100, 175}}
+	var bsizes []int
```
melekes:
```suggestion
	var bSizes []int
```

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +38,202 @@ func deleteTempFile(fname string) {
...
+		r := len(addrs)
+		assert.NotNil(t, addrs, "%s - expected a non-nill selection", dbgStr)
+		assert.NotZero(t, r, "%s - expected at least one address in selection", dbgStr)
+
+		// verify that the number of old and new addresses are the ones expected
+		numOld, numNew := countOldAndNewInAddrs(addrs, book)
+		assert.Equal(t, r, numOld+numNew, "%s - expected selection completely filled", dbgStr)
+
+		// Given:
+		// n - num new addrs, m - num old addrs
```
melekes: why use one letter vars when you can just use numNewAddrs or numNew => developer won't be forced to seek the Given reference to understand the code

i=min(n, k) => `numNewInTheBeginning`??

p2p/pex/addrbook_test.go
melekes: see https://github.com/tendermint/tendermint/pull/3232#discussion_r253740405
if you get rid of one symbol variables, you might not need `a comment for each case` at all!

ancazamfir:
Anton, thanks again for the review comments! And your patience :) Just did a commit now, please take a look when you have time.

p2p/pex/addrbook_test.go
ancazamfir: we check that the selection is fully populated. I changed the comment, let me know how it looks.

p2p/pex/addrbook_test.go
ebuchman: I think we can leave it as IsGood because we use Good/Bad on the public interface and old/new internally. It's true that eventually they may diverge and there could be different levels of Good, but I think we can leave for now.

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,212 @@ func deleteTempFile(fname string) {
...
 }
 
+func TestAddrBookGetSelectionWithOneMarkedGood(t *testing.T) {
+	// create a book with 10 addresses, 1 good/ old and 9 new
+	book := createAddrBookWithMOldAndNNewAddrs(t, 1, 9)
+
+	// ensure we can get selection.
+	// this used to result in an infinite loop (!)
+	addrs := book.GetSelectionWithBias(biasToSelectNewPeers)
+	assert.NotNil(t, addrs, "expected an address")
```
ebuchman: We should be able to assert exactly how many we get now, and in the next test.

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,212 @@ func deleteTempFile(fname string) {
...
+	}
+
+	return book
+}
+
+// Counts the number of Old and New addresses
+func countOldAndNewInAddrs(addrs []*p2p.NetAddress, book *addrBook) (nOld, nNew int) {
+
+	for _, addr := range addrs {
+		if addr == nil {
```
ebuchman: Shouldn't this not be allowed now?

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,212 @@ func deleteTempFile(fname string) {
...
+	}
+	seqLens = append(seqLens, currentSeqLen)
+	seqTypes = append(seqTypes, prevType)
+	return
+}
+
+func TestEmptyAddrBookSelection(t *testing.T) {
+	book := createAddrBookWithMOldAndNNewAddrs(t, 0, 0)
+	addrs := book.GetSelectionWithBias(biasToSelectNewPeers)
+	assert.Nil(t, addrs, "expected a nill selection")
```
melekes:
```suggestion
	assert.Nil(t, addrs, "expected a nil selection")
```

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,212 @@ func deleteTempFile(fname string) {
...
+		n := b - m
+		// debug string to show the context for potential failures
+		dbgStr := fmt.Sprintf("book: size %d, number new % d, old %d", b, n, m)
+
+		// create book and get selection
+		book := createAddrBookWithMOldAndNNewAddrs(t, m, n)
+		addrs := book.GetSelectionWithBias(biasToSelectNewPeers)
+
+		nAddr := len(addrs)
+		assert.NotNil(t, addrs, "%s - expected a non-nill selection", dbgStr)
```
melekes:
```suggestion
		assert.NotNil(t, addrs, "%s - expected a non-nil selection", dbgStr)
```

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,212 @@ func deleteTempFile(fname string) {
...
+	return
+}
+
+func TestEmptyAddrBookSelection(t *testing.T) {
+	book := createAddrBookWithMOldAndNNewAddrs(t, 0, 0)
+	addrs := book.GetSelectionWithBias(biasToSelectNewPeers)
+	assert.Nil(t, addrs, "expected a nill selection")
+}
+
+func testAddrBookAddressSelection(t *testing.T, b int) {
```
melekes:
```suggestion
func testAddrBookAddressSelection(t *testing.T, bookSize int) {
```

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,212 @@ func deleteTempFile(fname string) {
...
+	// generate all combinations of old (m) and new (n=b-m) addresses
+	for m := 0; m < b+1; m++ {
```
melekes:
```suggestion
	for nOld := 0; nOld <= bookSize; nOld++ {
```

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,212 @@ func deleteTempFile(fname string) {
...
+		n := b - m
```
melekes:
```suggestion
		nNew := bookSize - nOld
```

p2p/pex/addrbook_test.go
```diff
@@ -36,6 +37,212 @@ func deleteTempFile(fname string) {
...
+			dbgStr, expSeqLens, seqLens)
+		assert.Equal(t, expSeqTypes, seqTypes,
+			"%s - expected sequence lengths of old/new %v, got %v",
+			dbgStr, expSeqLens, seqLens)
+	}
+}
+
+func TestMultipleAddrBookAddressSelection(t *testing.T) {
+	// test books with smaller size, < N
+	N := 32
```
melekes:
```suggestion
	const N = 32
```

p2p/pex/addrbook_test.go
ancazamfir: it's still not. This is detected in the caller, if len(addrs) > nOld+nNew there were some nil addresses and it asserts.

p2p/pex/addrbook_test.go
ancazamfir: added

melekes:
note: localnet is failing because `circleci/classic` image most likely has a Go < 1.10 (where math.Round was introduced)

melekes:
@ancazamfir can I take over this PR to fix the tests and make last stylistic changes?

ancazamfir:
@melekes yes! please go ahead, thanks

p2p/pex/addrbook_test.go
```diff
@@ -417,3 +432,198 @@ func TestPrivatePeers(t *testing.T) {
...
+			expFirstNew = cmn.MinInt(nNew, k)
+			expOld      = cmn.MinInt(nOld, nAddrs-expFirstNew)
+			expNew      = nAddrs - expOld
+			expLastNew  = expNew - expFirstNew
+		)
+
+		// Verify that the number of old and new addresses are as expected
+		// XXX: Why verify expected numbers if assertMOldAndNNewAddrsInSelection
+		// which does exact comparison passes?
+		if nNew < expNew || nNew > expNew {
```
melekes: Why verify expected numbers if assertMOldAndNNewAddrsInSelection which does exact comparison (nOld==nOld, nNew==nNew) passes?

melekes:
Fixed `localnet` circleci job. Also, brushed addrbook_test.go a bit. One question left https://github.com/tendermint/tendermint/pull/3232#discussion_r254283366

p2p/pex/addrbook_test.go
ancazamfir: I think there is a problem with the assert in assertMOldAndNNewAddrsInSelection(), it was a different assert before.
If you change TestAddrBookGetSelectionWithOneNotMarkedGood() to have 90 old and 10 new the test will fail.
It should be like before, something like:
```

func assertOnlyOldAndNewAddrsInSelection(t *testing.T, addrs []*p2p.NetAddress, book *addrBook) {
	nOld, nNew := countOldAndNewAddrsInSelection(addrs, book)
	assert.Equal(t, len(addrs), nNew+nOld, "nill addresses in selection")
}
```
or maybe assertZeroNilAddrsInSelection()

melekes: I was expecting it to fail, but it did not. Why don't we test 90 old and 10 new addresses then?

ancazamfir: There should be no failure. The selection may be smaller than the book. So we cannot assert that nOld in book is same as nOld in selection. same for new. The original intent there was just to detect that the selection does not include nil addresses.

melekes: > The original intent there was just to detect that the selection does not include nil addresses.

Oh, ok then

p2p/pex/addrbook_test.go
```diff
@@ -417,3 +432,199 @@ func TestPrivatePeers(t *testing.T) {
...
+		assert.NotZero(t, nAddrs, "%s - expected at least one address in selection", dbgStr)
+
+		// check there's no nil addresses
+		for _, addr := range addrs {
+			if addr == nil {
+				t.Fatalf("%s - got nil address in selection %v", dbgStr, addrs)
+			}
+		}
+
+		// XXX: shadowing
```
melekes: bad code smell

melekes:
> wouldn't it be more optimal to remove the random selection completely from the `GetSelectionWithBias` function and rather just randomly shuffle the address book itself at some other point in the code?

We can explore that 👍 but it can also bring new issues. e.g., what if we want to call GetSelection in the middle of the reshuffling. Should we wait? Because if don't, we may end up having duplicates or miss some addresses (depending on reshuffling implementation - copy VS cut-paste). I guess not including an address is fine. Anyway, we should really think about benefits of having our own p2p layer instead of adopting libp2p. I.e. solve global issues, then move onto local.

melekes:
https://github.com/tendermint/tendermint/issues/2067#issuecomment-429196383

PR #3239, opened 2019-02-01
[WIP] Rollback Data

bradyjoestar:
#3196 

There are still lots of work to do though it worked on the newest version.

I would continue to do it.

I try to create a new command called `rollback` but I found it need to use lots of `flag` in `config.go`.
It better to put it into the `node` Cmd

Test situation(only a node):
```
tendermint node --proxy_app=kvstore --log_level=info 

# interrupt, stop node

#this will rollback data and restart node based rollback state
tendermint node --proxy_app=kvstore --log_level=info --rollback_data=true --rollback_height=$1 --rollback_height_flag=$flag
```


bradyjoestar:
future work:
- wal cut
- potential double signing
-  recompute the correct changes to proposer priority
- more scenario testcover

Issue #3224, opened 2019-01-29
panic when running proposer election on empty validator set


ancazamfir:
func TestEmptySet(t *testing.T) {
	valList := []*Validator{}
	vals := NewValidatorSet(valList)
	vals.IncrementProposerPriority(1)
}

and run:

=== RUN   TestEmptySet
--- FAIL: TestEmptySet (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x15b2b27]

goroutine 9 [running]:
testing.tRunner.func1(0xc000167200)
	/usr/local/Cellar/go/1.11.2/libexec/src/testing/testing.go:792 +0x387
panic(0x168e260, 0x1de4340)
	/usr/local/Cellar/go/1.11.2/libexec/src/runtime/panic.go:513 +0x1b9
github.com/tendermint/tendermint/types.(*ValidatorSet).incrementProposerPriority(0xc0004c72f0, 0x0)
	/Users/ancaz/go/src/github.com/tendermint/tendermint/types/validator_set.go:146 +0xe7
github.com/tendermint/tendermint/types.(*ValidatorSet).IncrementProposerPriority(0xc0004c72f0, 0x1)
	/Users/ancaz/go/src/github.com/tendermint/tendermint/types/validator_set.go:94 +0x71
github.com/tendermint/tendermint/types.TestEmptySet(0xc000167200)
	/Users/ancaz/go/src/github.com/tendermint/tendermint/types/validator_set_test.go:606 +0x4a
testing.tRunner(0xc000167200, 0x18986d0)
	/usr/local/Cellar/go/1.11.2/libexec/src/testing/testing.go:827 +0xbf
created by testing.(*T).Run
	/usr/local/Cellar/go/1.11.2/libexec/src/testing/testing.go:878 +0x353

Process finished with exit code 1

ancazamfir:
[test and trace on 3181_treat_validator_updates_as_set off develop]
related:
https://github.com/tendermint/tendermint/issues/2951
https://github.com/tendermint/tendermint/issues/2634
I don't see any guards in execution.go:updateState() against removing all validators from the set and before running vals.IncrementProposerPriority(1)
Regardless, the code in validator_set.go must not panic.

The trace above points to:
```
func (vals *ValidatorSet) incrementProposerPriority() *Validator {
[...]

Issue #2084, opened 2018-07-27
Tendermint should fail gracefully when all validators are revoked


cwgoes:
Version: Tendermint 0.22.8
ABCI app: Cosmos SDK / Gaia 0.23.1
Network type: one-validator local testnet
Config type: default config
Expected behavior: Tendermint halts and informs user that no validators are in the validator set
Observed behavior: Tendermint crashes with an interface conversion error

https://gist.github.com/cwgoes/bedbdf51845850ea456e656a5c68c164

ebuchman:
Closing for #3224

PR #3241, opened 2019-02-01
fix panic with empty validator set

ancazamfir:
* [X] Wrote tests

Fix for https://github.com/tendermint/tendermint/issues/3224
Also address https://github.com/tendermint/tendermint/issues/2084

### DIFF
--- a/blockchain/reactor.go
+++ b/blockchain/reactor.go
@@ -8,7 +8,6 @@ import (
 
 	amino "github.com/tendermint/go-amino"
 
-	cmn "github.com/tendermint/tendermint/libs/common"
 	"github.com/tendermint/tendermint/libs/log"
 	"github.com/tendermint/tendermint/p2p"
 	sm "github.com/tendermint/tendermint/state"
@@ -338,7 +337,7 @@ FOR_LOOP:
 				state, err = bcR.blockExec.ApplyBlock(state, firstID, first)
 				if err != nil {
 					// TODO This is bad, are we zombie?
-					cmn.Exit(fmt.Sprintf("Failed to process committed block (%d:%X): %v", first.Height, first.Hash(), err))
+					panic(fmt.Sprintf("Failed to process committed block (%d:%X): %v", first.Height, first.Hash(), err))
 				}
 				blocksSynced++
 
--- a/blockchain/reactor.go
+++ b/blockchain/reactor.go
@@ -338,8 +338,7 @@ FOR_LOOP:
 				state, err = bcR.blockExec.ApplyBlock(state, firstID, first)
 				if err != nil {
 					// TODO This is bad, are we zombie?
-					cmn.PanicQ(fmt.Sprintf("Failed to process committed block (%d:%X): %v",
-						first.Height, first.Hash(), err))
+					cmn.Exit(fmt.Sprintf("Failed to process committed block (%d:%X): %v", first.Height, first.Hash(), err))
 				}
 				blocksSynced++
 
--- a/state/execution_test.go
+++ b/state/execution_test.go
@@ -354,6 +354,33 @@ func TestEndBlockValidatorUpdates(t *testing.T) {
 	}
 }
 
+// TestEndBlockValidatorUpdatesResultingInEmptySet checks that processing validator updates that
+// would result in empty set causes no panic, an error is raised and NextValidators is not updated
+func TestEndBlockValidatorUpdatesResultingInEmptySet(t *testing.T) {
+	app := &testApp{}
+	cc := proxy.NewLocalClientCreator(app)
+	proxyApp := proxy.NewAppConns(cc)
+	err := proxyApp.Start()
+	require.Nil(t, err)
+	defer proxyApp.Stop()
+
+	state, stateDB := state(1, 1)
+	blockExec := NewBlockExecutor(stateDB, log.TestingLogger(), proxyApp.Consensus(), MockMempool{}, MockEvidencePool{})
+
+	block := makeBlock(state, 1)
+	blockID := types.BlockID{block.Hash(), block.MakePartSet(testPartSize).Header()}
+
+	// Remove the only validator
+	app.ValidatorUpdates = []abci.ValidatorUpdate{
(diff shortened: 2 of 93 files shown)


types/validator_set_test.go
```diff
@@ -20,12 +20,6 @@ func TestValidatorSetBasic(t *testing.T) {
 	// empty or nil validator lists are allowed,
 	// but attempting to IncrementProposerPriority on them will panic.
 	vset := NewValidatorSet([]*Validator{})
```
Liamsi: The comment on top of this line still states that panic will still happen.

ancazamfir: fixed

ancazamfir: As discussed today I also added a check for null validator set and error after updateValidators() is called in updateState().

Liamsi: And this will panic / shut down further up in `finalizeCommit` / poolRoutine. 
Not sure but the comment in the latter also sounds a bit concerning: https://github.com/tendermint/tendermint/blob/4d7b29cd8f8ef998cf8f7141f3a100b0a4fa718d/blockchain/reactor.go#L338-L342

ancazamfir: I've seen that also. Here it looks like we decided that recovery from all errors of ApplyBlock() is not possible. In general I think it's fine to panic in places where recovery is not possible. For validator updates related failures ( updates are invalid such that they either cannot be processed or could result in a bad state) maybe we could proceed by discarding the validator updates...not sure though.
Also, in the future we could completely validate the updates in validateValidatorUpdates(). Currently this doesn't catch all problems, some come later via updateValidators().

types/validator_set.go
```diff
@@ -74,6 +74,9 @@ func (vals *ValidatorSet) CopyIncrementProposerPriority(times int) *ValidatorSet
 // proposer. Panics if validator set is empty.
 // `times` must be positive.
 func (vals *ValidatorSet) IncrementProposerPriority(times int) {
+	if vals.IsNilOrEmpty() {
+		return
```
melekes: I don't like the fact that we simply return from the function 👎 Note the function's comment says "Panics if validator set is empty.", which is not case! Moreover, if we have a contract where validator set can't be ever empty, we should panic, not silently return. Imagine someone made an error in the code and validator set is empty => we return and continue execution with WRONG DATA. It's OK to panic if the invariants are not satisfied.

If validator set can be empty sometimes, we should use null object pattern https://www.tutorialspoint.com/design_pattern/null_object_pattern.htm

types/validator_set_test.go
ebuchman: The only reason we allow the validator set to be empty is during chain initialization where there are no validators in the genesis file because the application will return the initial validator set when we call `InitChain`. 

However, any use of the ValidatorSet before calling InitChain should probably panic. So I think Anton is right that we don't want to just return from these functions if they're used on an empty validator set - though we can explicitly panic instead of waiting for it to happen from a nil dereference.

melekes:
@ancazamfir are you still working on this?

ancazamfir:
@melekes yes

blockchain/reactor.go
```diff
@@ -338,8 +338,7 @@ FOR_LOOP:
 				state, err = bcR.blockExec.ApplyBlock(state, firstID, first)
 				if err != nil {
 					// TODO This is bad, are we zombie?
-					cmn.PanicQ(fmt.Sprintf("Failed to process committed block (%d:%X): %v",
-						first.Height, first.Hash(), err))
+					cmn.Exit(fmt.Sprintf("Failed to process committed block (%d:%X): %v", first.Height, first.Hash(), err))
```
melekes:
```suggestion
					panic(fmt.Sprintf("Failed to process committed block (%d:%X): %v", first.Height, first.Hash(), err))
```

state/execution_test.go
```diff
@@ -354,6 +354,33 @@ func TestEndBlockValidatorUpdates(t *testing.T) {
...
+	blockID := types.BlockID{block.Hash(), block.MakePartSet(testPartSize).Header()}
+
+	// Remove the only validator
+	app.ValidatorUpdates = []abci.ValidatorUpdate{
+		{PubKey: types.TM2PB.PubKey(state.Validators.Validators[0].PubKey), Power: 0},
+	}
+
+	assert.NotPanics(t, func() { state, err = blockExec.ApplyBlock(state, blockID, block) })
+	assert.NotNil(t, err)
+	assert.True(t, len(state.NextValidators.Validators) > 0, "state's NextValidators is empty")
```
melekes:
```suggestion
	assert.NotEmpty(t, state.NextValidators.Validators)
```

state/execution_test.go
```diff
@@ -354,6 +354,33 @@ func TestEndBlockValidatorUpdates(t *testing.T) {
...
+	block := makeBlock(state, 1)
+	blockID := types.BlockID{block.Hash(), block.MakePartSet(testPartSize).Header()}
+
+	// Remove the only validator
+	app.ValidatorUpdates = []abci.ValidatorUpdate{
+		{PubKey: types.TM2PB.PubKey(state.Validators.Validators[0].PubKey), Power: 0},
+	}
+
+	assert.NotPanics(t, func() { state, err = blockExec.ApplyBlock(state, blockID, block) })
+	assert.NotNil(t, err)
```
melekes: not sure if this checks the above error cause it's inside a closure

maybe it'll be better to move it inside the func

```go
assert.NotPanics(t, func() { 
  state, err = blockExec.ApplyBlock(state, blockID, block) 
  assert.Error(t, err)
})
```

types/validator_set_test.go
```diff
@@ -18,6 +18,7 @@ import (
 )
 
 func TestValidatorSetBasic(t *testing.T) {
+	// empty or nil validator lists are allowed,
```
Liamsi: Was this line is duplicated by accident?

state/execution_test.go
ancazamfir: I actually tested the test before to make sure, it's all good... state and err are defined outside the anonymous function.

types/validator_set_test.go
ancazamfir: yup :)

Issue #828, opened 2017-11-09
sync: Sync current state without full replay for Applications


ebuchman:
We want to be able to sync the current state without having to replay all transactions in the blockchain. 

This can be done securely by first syncing a light client to a recent state root, and then polling peers for the pieces of the state tree.

Let's add a new `state-sync` reactor to Tendermint to handle this. It should use a new ABCI connection and message to ask the app what to ask other peers for - alternatively, there may be a case for it to just use the existing Query connection and message. The base of this could be thought of as a general purpose mechanism for letting apps control the fetching of data from Tendermint peers

cloudhead:
Specification here: 
https://docs.google.com/document/d/15MFsQtNA0MGBv7F096FFWRDzQ1vR6_dics5Y49vF8JU/edit?ts=5a0f3629

ebuchman:
@jlandrews

jlandrews:
Future reminder from https://github.com/cosmos/cosmos-sdk/issues/1516#issuecomment-402202485. When building state syncing make sure to check if node is configured for a full prune, and if so disable the sync provider reactor and don't advertise that channel to peers.

hrharder:
If @cloudhead's spec from last November is implemented (or something similar) will this support pruning the leveldb blockstore while consensus is running?

ebuchman:
That's the ultimate goal, to institute a checkpointing mechanism that will allow nodes to prune history.

ackratos:
@ebuchman is this issue & proposal still valid? I'd like to contribute (because of our organization requirement).

ValarDragon:
I'm mostly in agreement with the current proposal. As a quick summary as I understand, the proposal is essentially suggesting to do BFS between syncer and syncee. (so you go layer by layer) The data is separated into chunks, and you do merkle proofs along the way. We can't initially send anything about leaves directly, since IAVL is insertion-order dependent, and encoding the tree structure is a pain. Similarly, you can't use knowledge of prior states to help here, unlike in patricia trees, and similar. We don't skip any interior nodes, even though they are in principle functions of later data. This is because the amount of data that they occupy relative to the data at the leaves should be negligible. (This isn't currently true in the cosmos hub, but will be eventually, one example: https://github.com/cosmos/cosmos-sdk/issues/2562#issuecomment-447512158)

[...]

PR #3243, opened 2019-02-02
[WIP] issue828 state syncing

ackratos:
* [ ] Wrote tests

For #828 

This PR is still working in progress, main functionality can "work" on our app but needs testing (multiple peer edge cases and performance) and complete thread safety. 

Just raise PR for community review/comments


ackratos:
https://github.com/tendermint/tendermint/issues/828#issuecomment-459948058

zmanian:
I just wanted to give you some feedback that a number of us have looked at your implementation and hopefully we can provide in depth technical feedback soon.

At a high level, one of the issues that's been important to us is that API for state synching must provide a mechanism of verifying chunks of data incrementally against the state root Bittorrent style.

This is one of our highest priority features in Cosmos over the next few months.

ackratos:
> At a high level, one of the issues that's been important to us is that API for state synching must provide a mechanism of verifying chunks of data incrementally against the state root Bittorrent style.

Thanks for looking into this!
Yes, that's major missing part of this PR. The specification doesn't describe verify mechanism in detail, what I thought is it should be like an additional calculated check sum? i.e. concatenate chunks then check sum it?

zmanian:
The basic approach we are thinking of is requiring the underlying data store to support RangeProofs like IAVL tree does and then provide a range proof for all the leaves in the chunk.

ackratos:
> The basic approach we are thinking of is requiring the underlying data store to support RangeProofs like IAVL tree does and then provide a range proof for all the leaves in the chunk.

then we need a new abci api like `VerifyRecoveryChunk(chunk [][]byte) error`? Or write chunks as soon as we receive? (Currently this implementation write state chunks on when we received all chunks)

zmanian:
Yeah I think VerifyRecoveryChunk makes sense.

If you only write chunks when all have been received, a malicious peer can trivially send you unlimited number of invalid chunks that you can't detect.

ackratos:
Sorry for long time no update, but I was keep testing this against our application. It works (in preliminary stage, not tested in large scale) now on our environment, I have updated this PR to latest status so you can take another round of review.

And what's more, relying on iterate db on every state sync related abci calls will must be a heavy work. We want borrow the way ethereum parity warp-sync (https://wiki.parity.io/Warp-Sync) or BitTorrent organize snapshot (application db version) that load version - split it into chunk files with hash - have a  manifest know all parts' hash - peers download manifest + some files via p2p (at the same time, they can serve file parts to other nodes)

I haven't taken deep look into detail design/implementation, but want hear from your opinions on this proposal, feel free to review this PR and reply to our new idea:)

ackratos:
> Sorry for long time no update, but I was keep testing this against our application. It works (in preliminary stage, not tested in large scale) now on our environment, I have updated this PR to latest status so you can take another round of review.
> 
> And what's more, relying on iterate db on state sync related abci (ReadChunks / LatestSnapshot) will must be a heavy work. We want borrow the way ethereum parity warp-sync (https://wiki.parity.io/Warp-Sync) or BitTorrent organize snapshot (application db version) that load version - split it into chunk files with hash - have a manifest know all parts' hash - peers download manifest + some files via p2p (at the same time, they can serve file parts to other nodes)
> 
> I haven't taken deep look into detail design/implementation, but want hear from your opinions on this proposal, feel free to review this PR and reply to our new idea:)

@melekes @ebuchman any idea on this?

types/params.go
```diff
@@ -12,6 +12,12 @@ const (
 
 	// BlockPartSizeBytes is the size of one block part.
 	BlockPartSizeBytes = 65536 // 64kB
+
+	// MaxStateSizeBytes is the maximum permitted size of the blocks.
+	// TODO: reduce this by segmentation
+	MaxStateSizeBytes = 1048576000 // 1000MB
+
+	MonitorWindowInSeconds = 40
```
ebuchman: doesn't seem to below here. These are hard consensus params, but this is more like a reactor setting.

ackratos:
> * making `LatestSnapshot` and `ReadSnapshotChunk` methods just be particular uses of `Query`. Query is already supposed to be used for reading from the current and historical state, so I think we should be able to reuse it for that here too. Do you think we really need the new methods, or can we just use Query?

abci query shares rw lock with other block execution apis (begin, end, commit, deliver, check tx) and this query is very heavy according to our experience. In case of chain have 500k user, there would be 1 million keys for acc substore of app state iterate all to get num of keys would cost several seconds...

we implemented cache for state sync in app level and only update the cache on syncable waypoint. The updating and query of cache shares a lock different with abci lock.

ackratos:
> * removing `StartRecovery`. Do we really need it ?
> * change `WriteRecoveryChunk` to take a struct that also includes a `done bool` to indicate if this is the last chunk. then we don't need EndRecovery either!

Good idea, I can do this.

ackratos:
> I would also think the state sync reactor should come in its own package. We can just have `blockchain/state` for now but we probably want to rename `blockchain` to `sync` and then have `sync/blockchain` and `sync/state` and move the existing `blockchain/store*` elsewhere. Just thinking aloud here and this shouldn't all happen in the same PR.

good idea. will do this.

ackratos:
closed in favor of new implementation https://github.com/tendermint/tendermint/pull/3594 for new proposal https://docs.google.com/document/d/1npGTAa1qxe8EQZ1wG0a0Sip9t5oX2vYZNUDwr_LVRR4/edit?usp=sharing

PR #3245, opened 2019-02-03
Introduce CommitSig alias for Vote in Commit

ebuchman:
This is some precursor work towards #1648. Ultimately we want to replace Vote in Commit with a struct that has only the signature and timestamp. Here we take some steps towards that by clarifying when we really need to get a vote from the commit.

* [ ] Wrote tests

### DIFF
--- a/blockchain/reactor_test.go
+++ b/blockchain/reactor_test.go
@@ -100,8 +100,8 @@ func newBlockchainReactor(logger log.Logger, genDoc *types.GenesisDoc, privVals 
 			lastBlockMeta := blockStore.LoadBlockMeta(blockHeight - 1)
 			lastBlock := blockStore.LoadBlock(blockHeight - 1)
 
-			vote := makeVote(&lastBlock.Header, lastBlockMeta.BlockID, state.Validators, privVals[0])
-			lastCommit = &types.Commit{Precommits: []*types.Vote{vote}, BlockID: lastBlockMeta.BlockID}
+			vote := makeVote(&lastBlock.Header, lastBlockMeta.BlockID, state.Validators, privVals[0]).CommitSig()
+			lastCommit = &types.Commit{Precommits: []*types.CommitSig{vote}, BlockID: lastBlockMeta.BlockID}
 		}
 
 		thisBlock := makeBlock(blockHeight, state, lastCommit)
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -472,7 +472,7 @@ func (cs *ConsensusState) reconstructLastCommit(state sm.State) {
 		if precommit == nil {
 			continue
 		}
-		added, err := lastPrecommits.AddVote(precommit)
+		added, err := lastPrecommits.AddVote(seenCommit.ToVote(precommit))
 		if !added || err != nil {
 			cmn.PanicCrisis(fmt.Sprintf("Failed to reconstruct LastCommit: %v", err))
 		}
@@ -1330,7 +1330,7 @@ func (cs *ConsensusState) recordMetrics(height int64, block *types.Block) {
 	missingValidators := 0
 	missingValidatorsPower := int64(0)
 	for i, val := range cs.Validators.Validators {
-		var vote *types.Vote
+		var vote *types.CommitSig
 		if i < len(block.LastCommit.Precommits) {
 			vote = block.LastCommit.Precommits[i]
 		}
--- a/lite/helpers.go
+++ b/lite/helpers.go
@@ -70,7 +70,7 @@ func (pkz privKeys) ToValidators(init, inc int64) *types.ValidatorSet {
 
 // signHeader properly signs the header with all keys from first to last exclusive.
 func (pkz privKeys) signHeader(header *types.Header, first, last int) *types.Commit {
-	votes := make([]*types.CommitSig, len(pkz))
+	commitSigs := make([]*types.CommitSig, len(pkz))
 
 	// We need this list to keep the ordering.
 	vset := pkz.ToValidators(1, 0)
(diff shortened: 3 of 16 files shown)


types/vote_set.go
```diff
@@ -541,8 +541,10 @@ func (voteSet *VoteSet) MakeCommit() *Commit {
 	}
 
 	// For every validator, get the precommit
-	votesCopy := make([]*Vote, len(voteSet.votes))
-	copy(votesCopy, voteSet.votes)
+	votesCopy := make([]*CommitSig, len(voteSet.votes))
```
melekes: No longer a copy in a strict sense. Rename to `precommits`?

lite/helpers.go
```diff
@@ -70,15 +70,15 @@ func (pkz privKeys) ToValidators(init, inc int64) *types.ValidatorSet {
 
 // signHeader properly signs the header with all keys from first to last exclusive.
 func (pkz privKeys) signHeader(header *types.Header, first, last int) *types.Commit {
-	votes := make([]*types.Vote, len(pkz))
+	votes := make([]*types.CommitSig, len(pkz))
```
melekes: Rename to `precommits`?

state/execution_test.go
```diff
@@ -136,9 +136,9 @@ func TestBeginBlockByzantineValidators(t *testing.T) {
 			types.TM2PB.Evidence(ev2, valSet, now)}},
 	}
 
-	vote0 := &types.Vote{ValidatorIndex: 0, Timestamp: now, Type: types.PrecommitType}
-	vote1 := &types.Vote{ValidatorIndex: 1, Timestamp: now}
-	votes := []*types.Vote{vote0, vote1}
+	vote0 := (&types.Vote{ValidatorIndex: 0, Timestamp: now, Type: types.PrecommitType}).CommitSig()
+	vote1 := (&types.Vote{ValidatorIndex: 1, Timestamp: now}).CommitSig()
+	votes := []*types.CommitSig{vote0, vote1}
```
Liamsi: nit: could also be renamed to `commit0`, `commit1`, `commit2` (or precommit0 etc)

types/vote_set.go
ebuchman: Is this copy really any different than the original?

ebuchman: Not sure I understand what you're getting at. You're just saying we should change the name since its commitsig now instead of votes? Or are you suggesting that using the for loop instead of the `copy` is leading to some different underlying behaviour?

melekes: > You're just saying we should change the name since its commitsig now instead of votes?

yes

state/execution.go
```diff
@@ -315,7 +315,7 @@ func getBeginBlockValidatorInfo(block *types.Block, lastValSet *types.ValidatorS
 	// Collect the vote info (list of validators and whether or not they signed).
 	voteInfos := make([]abci.VoteInfo, len(lastValSet.Validators))
 	for i, val := range lastValSet.Validators {
-		var vote *types.Vote
+		var vote *types.CommitSig
```
melekes:
```suggestion
		var precommit *types.CommitSig
```

consensus/state.go
```diff
@@ -1330,7 +1330,7 @@ func (cs *ConsensusState) recordMetrics(height int64, block *types.Block) {
 	missingValidators := 0
 	missingValidatorsPower := int64(0)
 	for i, val := range cs.Validators.Validators {
-		var vote *types.Vote
+		var vote *types.CommitSig
```
```suggestion
		var precommit *types.CommitSig
```

blockchain/reactor_test.go
```diff
@@ -100,8 +100,8 @@ func newBlockchainReactor(logger log.Logger, genDoc *types.GenesisDoc, privVals
 			lastBlockMeta := blockStore.LoadBlockMeta(blockHeight - 1)
 			lastBlock := blockStore.LoadBlock(blockHeight - 1)
 
-			vote := makeVote(&lastBlock.Header, lastBlockMeta.BlockID, state.Validators, privVals[0])
-			lastCommit = &types.Commit{Precommits: []*types.Vote{vote}, BlockID: lastBlockMeta.BlockID}
+			vote := makeVote(&lastBlock.Header, lastBlockMeta.BlockID, state.Validators, privVals[0]).CommitSig()
```
melekes:
```suggestion
			commitSig := makeVote(&lastBlock.Header, lastBlockMeta.BlockID, state.Validators, privVals[0]).CommitSig()
```

types/vote_set.go
```diff
@@ -541,11 +541,13 @@ func (voteSet *VoteSet) MakeCommit() *Commit {
 	}
 
 	// For every validator, get the precommit
-	votesCopy := make([]*Vote, len(voteSet.votes))
-	copy(votesCopy, voteSet.votes)
+	commitSigs := make([]*CommitSig, len(voteSet.votes))
```
melekes: On the second thought I think we should name it `precommits` since that's what we're building and later use to construct a Commit.

state/execution_test.go
```diff
@@ -136,10 +136,10 @@ func TestBeginBlockByzantineValidators(t *testing.T) {
 			types.TM2PB.Evidence(ev2, valSet, now)}},
 	}
 
-	vote0 := &types.Vote{ValidatorIndex: 0, Timestamp: now, Type: types.PrecommitType}
-	vote1 := &types.Vote{ValidatorIndex: 1, Timestamp: now}
-	votes := []*types.Vote{vote0, vote1}
-	lastCommit := &types.Commit{BlockID: prevBlockID, Precommits: votes}
+	commitSig0 := (&types.Vote{ValidatorIndex: 0, Timestamp: now, Type: types.PrecommitType}).CommitSig()
+	commitSig1 := (&types.Vote{ValidatorIndex: 1, Timestamp: now}).CommitSig()
+	commitSigs := []*types.CommitSig{commitSig0, commitSig1}
```
melekes: same

lite/helpers.go
```diff
@@ -70,20 +70,20 @@ func (pkz privKeys) ToValidators(init, inc int64) *types.ValidatorSet {
 
 // signHeader properly signs the header with all keys from first to last exclusive.
 func (pkz privKeys) signHeader(header *types.Header, first, last int) *types.Commit {
-	votes := make([]*types.Vote, len(pkz))
+	commitSigs := make([]*types.CommitSig, len(pkz))
```
melekes: same

types/vote_set.go
ebuchman: Hmm. I don't think either name is all the great to be honest. Better might be `PrecommitWitness`. A precommit is really the whole Vote structure with the PrecommitType. I'll comment on #1648 and we can address the renaming as we go.

Issue #3089, opened 2019-01-05
Crashed validator needed to manually edit priv_validator.json to come back online


zmanian:
During Game of Stakes, we saw great 1/3rd of the voting power due to the number of transactions in mempool.

Many users crashed during round 4 of block 9755. 

A number of users did not prevote in round 5 after restarting and got the following error message.
```
Jan  4 17:52:49 Apachto gaiad[12198]: I[4016-01-04|17:52:49.082] Timed out                                    module=consensus dur=5.5s height=9755 round=5 step=RoundStepPropose
Jan  4 17:52:49 Apachto gaiad[12198]: I[4016-01-04|17:52:49.082] enterPrevote(9755/5). Current: 9755/5/RoundStepPropose module=consensus
Jan  4 17:52:49 Apachto gaiad[12198]: I[4016-01-04|17:52:49.082] enterPrevote: ProposalBlock is nil           module=consensus height=9755 round=5
Jan  4 17:52:49 Apachto gaiad[12198]: E[4016-01-04|17:52:49.082] Error signing vote                           module=consensus height=9755 round=5 vote="Vote{144:BF6CFEAB362A 9755/05/1(Prevote) 000000000000 000000000000 @ 2019-01-04T17:52:49.082158529Z}" err="Error signing vote: Conflicting data"
```

An example of the priv_validator.json state is below. Manually editing the privalidator,json state so `last_round:0` cause the validators prevote. 

![sudcubehvvinhcdanhnudebs](https://user-images.githubusercontent.com/93434/50727229-f87fab00-10cc-11e9-94fb-0abc86fb6c8f.png)
[gaiad.log](https://github.com/tendermint/tendermint/files/2729711/gaiad.log)

ebuchman:
Thanks for the report Zaki. I recall seeing some folks reporting "non-deterministic signature" as well, which could be caused by different timestamps.

Do you have a text version of that priv val so we can inspect the last_signbytes?

What's supposed to happen in the event we sign the vote and update the priv val but crash before it hits the wal or the consensus is that we re-use the last_signature with the last_signbytes, assuming its for the same vote, where the only difference might be the timestamp. It's possible there's a bug in that logic somewhere, so we should take a look and try to replicate and test this behaviour better.

It should also have been possible to recover by just reducing the last_step, without needing to reduce the round.

zmanian:
Sorry I mis wrote. they only reduced `last_step` not `last-round`

zmanian:
[...]

PR #3246, opened 2019-02-04
WAL: better errors and new fail point

ebuchman:
This PR adds a new fail point that allows us to replicate #3089 with a bit of manual work.

It seems the issue from #3089 comes from crashing during a WriteSync operation, leaving the WAL in a corrupt state.

With the new fail point, we can replicate this situation as follows:

- run using `FAIL_TEST_INDEX=0`
- delete a few bytes from the end of the WAL
- start normally

This results in logs like:

```
    I[2019-02-03|18:12:58.225] Searching for height module=consensus wal=/Users/ethanbuchman/.tendermint/data/cs.wal/wal height=1 min=0 max=0
    E[2019-02-03|18:12:58.225] Error on catchup replay. Proceeding to start ConsensusState anyway module=consensus err="failed to read data: EOF"
    I[2019-02-03|18:12:58.225] Started node module=main nodeInfo="{ProtocolVersion:{P2P:6 Block:9 App:1} ID_:35e87e93f2e31f305b65a5517fd2102331b56002 ListenAddr:tcp://0.0.0.0:26656 Network:test-chain-J8JvJH Version:0.29.1 Channels:4020212223303800 Moniker:Ethans-MacBook-Pro.local Other:{TxIndex:on RPCAddress:tcp://0.0.0.0:26657}}"
    E[2019-02-03|18:12:58.226] Couldn't connect to any seeds module=p2p
    I[2019-02-03|18:12:59.229] Timed out module=consensus dur=998.568ms height=1 round=0 step=RoundStepNewHeight
    I[2019-02-03|18:12:59.230] enterNewRound(1/0). Current: 1/0/RoundStepNewHeight module=consensus height=1 round=0
    I[2019-02-03|18:12:59.230] enterPropose(1/0). Current: 1/0/RoundStepNewRound module=consensus height=1 round=0
    I[2019-02-03|18:12:59.230] enterPropose: Our turn to propose module=consensus height=1 round=0 proposer=AD278B7767B05D7FBEB76207024C650988FA77D5 privValidator="PrivValidator{AD278B7767B05D7FBEB76207024C650988FA77D5 LH:1, LR:0, LS:2}"
    E[2019-02-03|18:12:59.230] enterPropose: Error signing proposal module=consensus height=1 round=0 err="Error signing proposal: Step regression at height 1 round 0. Got 1, last step 2"
    I[2019-02-03|18:13:02.233] Timed out module=consensus dur=3s height=1 round=0 step=RoundStepPropose
    I[2019-02-03|18:13:02.233] enterPrevote(1/0). Current: 1/0/RoundStepPropose module=consensus
    I[2019-02-03|18:13:02.233] enterPrevote: ProposalBlock is nil module=consensus height=1 round=0
    E[2019-02-03|18:13:02.234] Error signing vote module=consensus height=1 round=0 vote="Vote{0:AD278B7767B0 1/00/1(Prevote) 000000000000 000000000000 @ 2019-02-04T02:13:02.233897Z}" err="Error signing vote: Conflicting data"
```

which matches #3089.

The problem is that we fail to search for the ENDHEIGHT line because we encounter a corrupt entry. So we don't even get to replay any of the WAL. Since the WAL contains our proposal, and we've also already prevoted, when the consensus starts it tries to sign a new proposal, which of course fails with a step regression. Then since the proposal is nil, we try to sign a nil prevote, which conflicts with the last prevote we signed, thus yielding the `conflicting data` error.

* [ ] Wrote tests

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -319,8 +319,7 @@ cp $WALFILE ${WALFILE}.bak # backup the file
 go run scripts/wal2json/main.go $WALFILE > wal.json # this will panic, but can be ignored
 rm $WALFILE # remove the corrupt file
 go run scripts/json2wal/main.go wal.json $WALFILE # rebuild the file without corruption
-----
-`)
+----`)
 
 				return err
 			}


ebuchman:
If we change all the errors in `wal.Decode` to be of type DataCorruptionError, then they will be ignored during the SearchForEndHeight, and we can successfully replay the WAL messages.

This is great because it allows us to recover once:

```
I[2019-02-03|18:32:24.448] Searching for height                         module=consensus wal=/Users/ethanbuchman/.tendermint/data/cs.wal/wal height=1 min=0 max=0
E[2019-02-03|18:32:24.448] Corrupted entry. Skipping...                 module=consensus wal=/Users/ethanbuchman/.tendermint/data/cs.wal/wal err="DataCorruptionError[failed to read data: EOF]"
I[2019-02-03|18:32:24.448] Searching for height                         module=consensus wal=/Users/ethanbuchman/.tendermint/data/cs.wal/wal height=0 min=0 max=0
I[2019-02-03|18:32:24.448] Found                                        module=consensus wal=/Users/ethanbuchman/.tendermint/data/cs.wal/wal height=0 index=0
I[2019-02-03|18:32:24.448] Catchup by replaying consensus messages      module=consensus height=1
I[2019-02-03|18:32:24.448] Replay: Timeout                              module=consensus height=1 round=0 step=RoundStepNewHeight dur=997.987ms
I[2019-02-03|18:32:24.449] Need to set a buffer and log message here for sanity. module=consensus height=1 round=0 startTime=2019-02-04T02:32:25.446877Z now=2019-02-04T02:32:24.449018Z
I[2019-02-03|18:32:24.449] enterNewRound(1/0). Current: 1/0/RoundStepNewHeight module=consensus height=1 round=0
I[2019-02-03|18:32:24.449] enterPropose(1/0). Current: 1/0/RoundStepNewRound module=consensus height=1 round=0
I[2019-02-03|18:32:24.449] enterPropose: Our turn to propose            module=consensus height=1 round=0 proposer=AD278B7767B05D7FBEB76207024C650988FA77D5 privValidator="PrivValidator{AD278B7767B05D7FBEB76207024C650988FA77D5 LH:1, LR:0, LS:2}"
I[2019-02-03|18:32:24.449] Replay: New Step                             module=consensus height=1 round=0 step=RoundStepPropose
I[2019-02-03|18:32:24.449] Replay: Proposal                             module=consensus height=1 round=0 header=1:EE679B583367 pol=-1 peer=local
I[2019-02-03|18:32:24.449] Received proposal                            module=consensus proposal="Proposal{1/0 (648AE55754077FA81CA57F4450CD2D7AC3D39EBBDA17218D2F2AB4FF65C44328:1:EE679B583367, -1) 4425F05CB764 @ 2019-02-04T02:32:17.314291Z}"
I[2019-02-03|18:32:24.450] Replay: BlockPart                            module=consensus height=1 round=0 peer=local
I[2019-02-03|18:32:24.450] Received complete proposal block             module=consensus height=1 hash=648AE55754077FA81CA57F4450CD2D7AC3D39EBBDA17218D2F2AB4FF65C44328
I[2019-02-03|18:32:24.450] enterPrevote(1/0). Current: 1/0/RoundStepPropose module=consensus
I[2019-02-03|18:32:24.450] enterPrevote: ProposalBlock is valid         module=consensus height=1 round=0
I[2019-02-03|18:32:24.450] Signed and pushed vote                       module=consensus height=1 round=0 vote="Vote{0:AD278B7767B0 1/00/1(Prevote) 648AE5575407 719C7762B8D6 @ 2019-02-04T02:32:17.318336Z}" err=null
I[2019-02-03|18:32:24.450] Replay: New Step                             module=consensus height=1 round=0 step=RoundStepPrevote
E[2019-02-03|18:32:24.450] data has been corrupted in last height of consensus WAL module=consensus err="DataCorruptionError[failed to read data: EOF]" height=1
E[2019-02-03|18:32:24.450] Error on catchup replay. Proceeding to start ConsensusState anyway module=consensus err="DataCorruptionError[failed to read data: EOF]"
I[2019-02-03|18:32:24.450] Started node                                 module=main nodeInfo="{ProtocolVersion:{P2P:6 Block:9 App:1} ID_:35e87e93f2e31f305b65a5517fd2102331b56002 ListenAddr:tcp://0.0.0.0:26656 Network:test-chain-J8JvJH Version:0.29.1 Channels:4020212223303800 Moniker:Ethans-MacBook-Pro.local Other:{TxIndex:on RPCAddress:tcp://0.0.0.0:26657}}"
I[2019-02-03|18:32:24.451] Added to prevote                             module=consensus vote="Vote{0:AD278B7767B0 1/00/1(Prevote) 648AE5575407 719C7762B8D6 @ 2019-02-04T02:32:17.318336Z}" prevotes="VoteSet{H:1 R:0 T:1 +2/3:648AE55754077FA81CA57F4450CD2D7AC3D39EBBDA17218D2F2AB4FF65C44328:1:EE679B583367(1) BA{1:x} map[]}"
```

This time we skip the corrupted entry during the SearchForEndHeight, which allows us to actually replay the other messages, terminate the replay early, and then successfully start the consensus engine. In this case, the proposal msg got successfully replayed, so when we go to prevote, even though we already did prevote before crashing and the WAL entry for it was corrupted, we can just re-use the last signature from the PrivValidator. So we have no problem moving on. However, if we shutdown and restart again, now we're going to have problems:

```
I[2019-02-03|18:36:43.591] Searching for height                         module=consensus wal=/Users/ethanbuchman/.tendermint/data/cs.wal/wal height=8 min=0 max=0
E[2019-02-03|18:36:43.591] Corrupted entry. Skipping...                 module=consensus wal=/Users/ethanbuchman/.tendermint/data/cs.wal/wal err="DataCorruptionError[checksums do not match: (read: 1538959906, actual: 3713544688)]"
E[2019-02-03|18:36:43.592] Corrupted entry. Skipping...                 module=consensus wal=/Users/ethanbuchman/.tendermint/data/cs.wal/wal err="DataCorruptionError[length 201898184 exceeded maximum possible value of 1048576 bytes]"
E[2019-02-03|18:36:43.592] Corrupted entry. Skipping...                 module=consensus wal=/Users/ethanbuchman/.tendermint/data/cs.wal/wal err="DataCorruptionError[length 3633031126 exceeded maximum possible value of 1048576 bytes]"
E[2019-02-03|18:36:43.592] Corrupted entry. Skipping...                 module=consensus wal=/Users/ethanbuchman/.tendermint/data/cs.wal/wal err="DataCorruptionError[length 3011161608 exceeded maximum possible value of 1048576 bytes]"
E[2019-02-03|18:36:43.592] Corrupted entry. Skipping...                 module=consensus wal=/Users/ethanbuchman/.tendermint/data/cs.wal/wal err="DataCorruptionError[length 1869966948 exceeded maximum possible value of 1048576 bytes]"
...
```

And this continues for quite some time.

The problem is that we left the WAL with a corrupted line and then started writing to it again, so we can never actually recover since trying to read that corrupted line will also read bytes from the following non-corrupted lines. The spacing is now all off!

I think to fix this we need to remove the corrupt line from the WAL the first time we read it. Note this requires some surgery on the autofile/group.

Another possibility is to use a separator between line entries, as discussed in #785, however it's not clear how that would really help in case the separator itself gets corrupted. So we probably just need to remove the corrupted line. Thoughts @melekes ?

ebuchman:
I made the suggested change to the Decode errors in the last commit. This allows us to recover once from the kind of data corruption seen in #3089. I think we can merge this because it's strictly better than before, but to fix it fully we need to remove that corrupt line, which as mentioned will require a bit of surgery to remove the final line from the group autofile. Ie. we need to find out what line it starts at in the file and the truncate the file from there. It might turn out to be easy but on first glance it seemed non-trivial.

melekes:
> however it's not clear how that would really help in case the separator itself gets corrupted

exactly. separator won't help

> I think to fix this we need to remove the corrupt line from the WAL the first time we read it. Note this requires some surgery on the autofile/group.

Note previously we've decided to panic and let the user deal with the corrupted entry (which can be a sign of filesystem malfunctioning or disk errors) https://github.com/tendermint/tendermint/issues/573#issuecomment-353196664

What are the consequences of removing a line? If it's a `ENDHEIGHT 2` and we remove it and fail again at the next height (3), we will try to replay messages from both heights 2 and 3, right? If we're ok with removing any message, then it can be done automatically (given we can still read a msg's `length`, otherwise user intervention is inevitable).

ebuchman:
> What are the consequences of removing a line? If it's a ENDHEIGHT 2 and we remove it and fail again at the next height (3), we will try to replay messages from both heights 2 and 3, right? If we're ok with removing any message, then it can be done automatically (given we can still read a msg's length, otherwise user intervention is inevitable).

I think we should be ok to remove the last line of the WAL if its corrupted. It may be corrupted by having failed to write final bytes, so it will be less than the expected length. But so long as we can find where the previous line ended, we should be able to truncate from there.

Actually, I just realized that we can use the wal2json and json2wal tools to handle this manually as is, but then we really should exit if we detect corruption instead of trying to start anyways. Then we can do:

```
go run scripts/wal2json/main.go ~/.tendermint/data/cs.wal/wal > wal.json
rm ~/.tendermint/data/cs.wal/wal
go run scripts/json2wal/main.go wal.json ~/.tendermint/data/cs.wal/wal
```

That should give us back the same WAL file without the last (corrupted) line!

consensus/state.go
```diff
@@ -307,6 +307,26 @@ func (cs *ConsensusState) OnStart() error {
...
+
+----
+WALFILE=~/.tendermint/data/cs.wal/wal
+cp $WALFILE ${WALFILE}.bak # backup the file
+go run scripts/wal2json/main.go $WALFILE > wal.json # this will panic, but can be ignored
+rm $WALFILE # remove the corrupt file
+go run scripts/json2wal/main.go wal.json $WALFILE # rebuild the file without corruption
+----
+
+NOTE: be sure to set the WALFILE to the highest numbered file in the cs.wal directory.
```
melekes: Isn't the latest file always called `wal`?

ebuchman: Yes of course you're right!

PR #3248, opened 2019-02-04
bound mempool memory usage

### LINKED ISSUE #3079: Bound mempool memory usage
**Summary**: As a subcomponent to: https://github.com/tendermint/tendermint/issues/1696
I think it may be near-term easy and important to bound the space usage of the mempool txs. (Note I am not proposing accounting for the LRU cache overhead, or mempoolTx struct overhead, just bounding the total amount of tx data itself) 

**Problem**: At the moment, every tx in the mempool can be of size `max(maxMsgSize, BlockDataSize)`, and the former is a constant of `1MB`. In a lot configurations, block data size may be large, which would open up the mempool to taking a lot of ram. Under default mempool size, it would be `5000 * max_block_size`. With 200KB blocks, thats 1GB of RAM, with 1MB blocks, that opens yourself up to 5GB of RAM consumed. I think both of these are too much to open up sentries under default

**Proposed solution**: Add a `max_bytes` field to the mempool config options. Before adding a tx to the mempool, check if it will cause the mempool to exceed its max_bytes limit. When removing txs in Update, subtract the corresponding amount from the `max_bytes` field. `max_txs` is a useful thing to still keep, as there are many things that depend on the number of txs as well. (Recheck time, etc.)

melekes:
Closes #3079

* [x] Wrote tests


config/config.go
```diff
@@ -548,6 +550,7 @@ func DefaultMempoolConfig() *MempoolConfig {
 		// ABCI Recheck
 		Size:      5000,
 		CacheSize: 10000,
+		MaxBytes:  1000000000, // 1GB
```
milosevic: Isn't it 1073741824 for 1GB? What about writing these kind of constants as 1024 * 1024 * 1024?

config/config.go
```diff
@@ -536,6 +536,8 @@ type MempoolConfig struct {
 	WalPath   string `mapstructure:"wal_dir"`
 	Size      int    `mapstructure:"size"`
 	CacheSize int    `mapstructure:"cache_size"`
+	// Maximum size of all txs in the mempool in bytes
+	MaxBytes int `mapstructure:"max_bytes"`
```
milosevic: What is a rule of thumb regarding integer types? When we use int and when we use fixed types such as int16, int32 etc?

config/config.go
```diff
@@ -577,6 +580,9 @@ func (cfg *MempoolConfig) ValidateBasic() error {
 	if cfg.CacheSize < 0 {
 		return errors.New("cache_size can't be negative")
 	}
+	if cfg.MaxBytes <= 0 {
+		return errors.New("max_bytes must be a positive number")
```
milosevic: Shouldn't minimum mempool size be enough to store a single transaction? Maybe minimum mempool size should be `maxTxSize`?

config/config.go
```diff
@@ -536,6 +536,8 @@ type MempoolConfig struct {
 	WalPath   string `mapstructure:"wal_dir"`
 	Size      int    `mapstructure:"size"`
 	CacheSize int    `mapstructure:"cache_size"`
+	// Maximum size of all txs in the mempool in bytes
```
milosevic: It can be a bit confusing as it can be understood as max tx size. Maybe instead "Maximum size of the mempool in bytes" or "Maximum size of the mempool list in bytes" in case we want to differentiate it from the mempool cache.

config/toml.go
```diff
@@ -240,6 +240,9 @@ size = {{ .Mempool.Size }}
 # size of the cache (used to filter transactions we saw earlier)
 cache_size = {{ .Mempool.CacheSize }}
 
+# maximum size of all txs in the mempool in bytes
```
milosevic: The same as above.

docs/tendermint-core/configuration.md
```diff
@@ -189,6 +189,9 @@ size = 5000
 # size of the cache (used to filter transactions we saw earlier)
```
milosevic: Maybe we can be precise if size is in bytes or in transactions.

docs/tendermint-core/configuration.md
```diff
@@ -189,6 +189,9 @@ size = 5000
...
 cache_size = 10000
 
+# maximum size of all txs in the mempool in bytes
```
milosevic: The same as above.

docs/tendermint-core/configuration.md
```diff
@@ -189,6 +189,9 @@ size = 5000
...
+max_bytes = 1000000000
```
milosevic: The same as above.

mempool/mempool.go
```diff
@@ -147,6 +147,8 @@ type Mempool struct {
 	preCheck             PreCheckFunc
 	postCheck            PostCheckFunc
 
+	sizeBytes int64 // size of all txs in the mempool in bytes
```
milosevic: Is it on purpose here `int64` and in Mempool config `int`? The same regarding comment as above.

mempool/mempool.go
```diff
@@ -265,8 +267,13 @@ func (mem *Mempool) Size() int {
 	return mem.txs.Len()
 }
 
-// Flushes the mempool connection to ensure async resCb calls are done e.g.
-// from CheckTx.
+// SizeBytes returns the size of all txs in the mempool in bytes.
+func (mem *Mempool) SizeBytes() int64 {
+	return atomic.LoadInt64(&mem.sizeBytes)
```
milosevic: Nice.

mempool/mempool.go
```diff
@@ -280,6 +287,7 @@ func (mem *Mempool) Flush() {
 
 	for e := mem.txs.Front(); e != nil; e = e.Next() {
 		mem.txs.Remove(e)
+		_ = atomic.SwapInt64(&mem.sizeBytes, 0)
```
milosevic: Shouldn't this be outside for loop? Just a side note: in this case strictly speaking we don't need atomic access as we hold `mem.proxyMtx.Lock()` but it's would help once we start getting rid of mempool locks. Maybe it's worth leaving a comment so reader is not confused.

config/config.go
```diff
@@ -548,6 +550,7 @@ func DefaultMempoolConfig() *MempoolConfig {
 		// ABCI Recheck
 		Size:      5000,
```
milosevic: Probably we need to be clear now with units, i.e., when we talk about number of txs and when we talk about size in bytes.

mempool/mempool_test.go
```diff
@@ -450,6 +450,28 @@ func TestMempoolMaxMsgSize(t *testing.T) {
...
+	mempool.Update(1, []types.Tx{[]byte{0x01}}, nil, nil)
+	assert.EqualValues(t, 0, mempool.SizeBytes())
+
+	err = mempool.CheckTx([]byte{0x02, 0x03}, nil)
+	require.NoError(t, err)
+	assert.EqualValues(t, 2, mempool.SizeBytes())
+
+	mempool.Flush()
+	assert.EqualValues(t, 0, mempool.SizeBytes())
+}
```
milosevic: Maybe it make sense adding a test for scenario where mempool is already full so trying to add tx will fail.

config/config.go
melekes: Bytes is a common convention. Plus if we make it MB, we'll need to convert back to bytes when comparing. 1024*1024*1024 is nice.

config/config.go
melekes: Don't think we should handle such cases (clearly misconfiguration)

config/config.go
melekes: > mempool list

users don't know anything about mempool list. I will change it to "Maximum size of the mempool in bytes"

config/config.go
melekes: int32 can represent up to 2GB, so it's not enough. This can be uint, but then we'll need to type cast again in mempool

docs/tendermint-core/configuration.md
melekes: yep, added `in transactions` suffix

mempool/mempool.go
melekes: Good catch, thanks 👍

config/config.go
```diff
@@ -535,6 +535,7 @@ type MempoolConfig struct {
 	Broadcast bool   `mapstructure:"broadcast"`
 	WalPath   string `mapstructure:"wal_dir"`
 	Size      int    `mapstructure:"size"`
+	MaxBytes  int64  `mapstructure:"max_bytes"`
```
melekes: rename to `MaxTxsTotalBytes`?

melekes:
> Should we make the ErrMempoolFull more informative? Eg. it could contain the num txs and the num bytes, and the respective limits? Might also be nice to expose the byte size over the rpc endpoint, though /num_unconfirmed_txs becomes not so great a name. We can address this in new issues though.

Done. See my last commits

melekes:
Thanks @milosevic and @ebuchman for reviews 👍

melekes:
We might also want to expose txs_total_bytes via Prometheus.

config/config.go
```diff
@@ -530,12 +530,13 @@ func DefaultFuzzConnConfig() *FuzzConnConfig {
...
-	Recheck   bool   `mapstructure:"recheck"`
-	Broadcast bool   `mapstructure:"broadcast"`
-	WalPath   string `mapstructure:"wal_dir"`
-	Size      int    `mapstructure:"size"`
-	CacheSize int    `mapstructure:"cache_size"`
+	RootDir          string `mapstructure:"home"`
+	Recheck          bool   `mapstructure:"recheck"`
+	Broadcast        bool   `mapstructure:"broadcast"`
+	WalPath          string `mapstructure:"wal_dir"`
+	Size             int    `mapstructure:"size"`
```
ebuchman: Should we plan to make these names more consistent? Eg. `MaxTotalTxs` and  `MaxTotalTxBytes` or something?

mempool/mempool.go
```diff
@@ -147,6 +159,8 @@ type Mempool struct {
 	preCheck             PreCheckFunc
 	postCheck            PostCheckFunc
 
+	txsTotalBytes int64 // see TxsTotalBytes
```
ebuchman: we should note that this is an atomic integer

CHANGELOG_PENDING.md
```diff
@@ -7,6 +7,7 @@ Special thanks to external contributors on this release:
 ### BREAKING CHANGES:
 
 * CLI/RPC/Config
+  - [rpc] \#3113 rename n_txs to count in `/num_unconfirmed_txs` and `/unconfirmed_txs`
```
ebuchman: If we're planning to release v0.29.2 and we want to include this I'd say we should make this non-breaking for now and we can fix the naming in both the rpc and the config in v0.30

config/config.go
```diff
@@ -574,6 +576,9 @@ func (cfg *MempoolConfig) ValidateBasic() error {
 	if cfg.Size < 0 {
 		return errors.New("size can't be negative")
 	}
+	if cfg.MaxTxsTotalBytes < 0 {
```
ebuchman: So setting either of these to 0 is allowed and will just cause us to always reject all txs?

mempool/mempool.go
```diff
@@ -308,8 +329,11 @@ func (mem *Mempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
 	// use defer to unlock mutex because application (*local client*) might panic
 	defer mem.proxyMtx.Unlock()
 
-	if mem.Size() >= mem.config.Size {
-		return ErrMempoolIsFull
+	if mem.Size() >= mem.config.Size ||
+		int64(len(tx))+mem.TxsTotalBytes() > mem.config.MaxTxsTotalBytes {
+		return ErrMempoolIsFull{
+			mem.Size(), mem.config.Size,
+			mem.TxsTotalBytes(), mem.config.MaxTxsTotalBytes}
```
ebuchman: Isn't it possible for TxsTotalBytes() to change between when we checked and when we construct the error? 

We should probably store it in a local var and use that to avoid the potential (though relatively insignificant) race condition!

mempool/mempool_test.go
ebuchman: Would be good to add this test, and also one where the tx gets removed during recheck if possible

config/config.go
melekes: yes. I don't see why we'd want to forbid doing that

CHANGELOG_PENDING.md
melekes: Why not release v0.30?

melekes:
TODO:
- [ ] actually test the limit + value after recheck

melekes:
I think I'm done here. PTAL

milosevic:
LGTM. Great work Anton.

CHANGELOG_PENDING.md
ebuchman: To establish a better cadence of making bug fixes before making breaking changes. We should get in a habit of releasing everything on develop as non-breaking release before merging the breaking stuff

melekes: Fair enough

mempool/mempool_test.go
```diff
@@ -450,6 +452,75 @@ func TestMempoolMaxMsgSize(t *testing.T) {
...
+	require.Nil(t, err)
+	defer appConnCon.Stop()
+	res, err := appConnCon.DeliverTxSync(txBytes)
+	require.NoError(t, err)
+	require.EqualValues(t, 0, res.Code)
+	res2, err := appConnCon.CommitSync()
+	require.NoError(t, err)
+	require.NotEmpty(t, res2.Data)
+
+	// Note we say we've committed txBytes2, but actually it was txBytes.
```
ebuchman: This is a bit strange, but I guess the idea here is to get txBytes replayed.  Is txBytes2 even necessary then? Why not just have the one tx, commit it, pretend like we committed nothing so it gets rechecked and fails?

To test this without "pretending" I think we'd need a different ABCI app

rpc/core/mempool.go
```diff
@@ -255,27 +255,32 @@ func BroadcastTxCommit(tx types.Tx) (*ctypes.ResultBroadcastTxCommit, error) {
...
-//     "txs": [],
-//     "n_txs": "0"
-//   },
-//   "id": "",
-//   "jsonrpc": "2.0"
-// }
+//   "result" : {
+//       "txs" : [],
+//       "total_bytes" : "0",
+//       "count" : "0",
```
ebuchman: I feel like this would be more straight forward as:

```
txs
n_txs
total_txs
total_bytes
```

Otherwise it's not really clear what `count` and `total` are referring to. This would also make it non breaking, which is always a bonus :)

rpc/core/mempool.go
```diff
@@ -298,15 +303,19 @@ func UnconfirmedTxs(limit int) (*ctypes.ResultUnconfirmedTxs, error) {
...
-//   },
-//   "id": "",
-//   "jsonrpc": "2.0"
+//   "jsonrpc" : "2.0",
+//   "id" : "",
+//   "result" : {
+//     "count" : "0",
+//     "total_bytes" : "0",
+//     "txs" : null,
+//     "total" : "0"
```
ebuchman: Same

config/config.go
```diff
@@ -530,12 +530,13 @@ func DefaultFuzzConnConfig() *FuzzConnConfig {
...
-	WalPath   string `mapstructure:"wal_dir"`
-	Size      int    `mapstructure:"size"`
-	CacheSize int    `mapstructure:"cache_size"`
+	RootDir          string `mapstructure:"home"`
+	Recheck          bool   `mapstructure:"recheck"`
+	Broadcast        bool   `mapstructure:"broadcast"`
+	WalPath          string `mapstructure:"wal_dir"`
+	MaxTxs           int    `mapstructure:"max_txs"`
+	MaxTxsTotalBytes int64  `mapstructure:"max_txs_total_bytes"`
+	CacheSize        int    `mapstructure:"cache_size"`
```
ebuchman: CacheMaxTxs?

config/config.go
```diff
@@ -530,12 +530,13 @@ func DefaultFuzzConnConfig() *FuzzConnConfig {
...
-	Broadcast bool   `mapstructure:"broadcast"`
-	WalPath   string `mapstructure:"wal_dir"`
-	Size      int    `mapstructure:"size"`
-	CacheSize int    `mapstructure:"cache_size"`
+	RootDir          string `mapstructure:"home"`
+	Recheck          bool   `mapstructure:"recheck"`
+	Broadcast        bool   `mapstructure:"broadcast"`
+	WalPath          string `mapstructure:"wal_dir"`
+	MaxTxs           int    `mapstructure:"max_txs"`
+	MaxTxsTotalBytes int64  `mapstructure:"max_txs_total_bytes"`
```
ebuchman: Maybe `MaxTxsBytes` is sufficient, since the `Total` is kind of redundant with the `s` on `Txs`. And Tendermint has no concept of `MaxTxBytes` or `MaxTxSingleBytes`, since we just bound txs by the block size instead (full block with 1 tx).

Actually, if we're going to keep `MaxTxs` as just `Size`, then maybe this is best as `SizeBytes` for now?

config/config.go
```diff
@@ -530,12 +530,13 @@ func DefaultFuzzConnConfig() *FuzzConnConfig {
...
-	Recheck   bool   `mapstructure:"recheck"`
-	Broadcast bool   `mapstructure:"broadcast"`
-	WalPath   string `mapstructure:"wal_dir"`
-	Size      int    `mapstructure:"size"`
-	CacheSize int    `mapstructure:"cache_size"`
+	RootDir          string `mapstructure:"home"`
+	Recheck          bool   `mapstructure:"recheck"`
+	Broadcast        bool   `mapstructure:"broadcast"`
+	WalPath          string `mapstructure:"wal_dir"`
+	MaxTxs           int    `mapstructure:"max_txs"`
```
ebuchman: Sorry, I think I take back my original suggestion for now just to air on the side of breaking fewer things per release.

One day we should do a review/cleanup of the config file perhaps, and then we can fix all the name. Might be better than having one off breaking changes in naming that everyone has to fix ...

rpc/core/mempool.go
melekes: The benefit of count and total is they are the same across all the endpoints (which return an array)

config/config.go
melekes: In case of the cache, we do allocate memory, so it's completely fair to name the config parameter `CacheSize`

config/config.go
melekes: > Actually, if we're going to keep `MaxTxs` as just `Size`, then maybe this is best as `SizeBytes` for now?

SizeBytes is too ambiguous. Does it include the cache size? Internal structures? ... MaxTxsBytes is much better

melekes:
Okay, reverted some changes to make this non-breaking. PTAL

melekes:
```
--- FAIL: TestWALPeriodicSync (0.31s)
    <autogenerated>:1: ^M                          ^M   Error Trace:    wal_test.go:169
        ^M      Error:          Should not be zero, but was 0
        ^M      Test:           TestWALPeriodicSync
```
is failing. seems not related to this PR (cc @thanethomson)

rpc/core/mempool.go
ebuchman: Where else do we do that? At least for this PR since we made it non breaking we should fix the comment here

ebuchman:
```suggestion
//       "n_txs" : "0",
```

rpc/core/mempool.go
```diff
@@ -298,15 +303,19 @@ func UnconfirmedTxs(limit int) (*ctypes.ResultUnconfirmedTxs, error) {
...
-//   "result": {
-//     "txs": null,
-//     "n_txs": "0"
-//   },
-//   "id": "",
-//   "jsonrpc": "2.0"
+//   "jsonrpc" : "2.0",
+//   "id" : "",
+//   "result" : {
+//     "count" : "0",
```
ebuchman:
```suggestion
//     "n_txs" : "0",
```

PR #3252, opened 2019-02-05
[WIP] limit number of /subscribe clients and queries per client

### LINKED ISSUE #2826: rpc: too many broadcast_commit calls can lead to consensus liveness issues
### Issue 

See #2748 

pubsub currently has no buffer (capacity: 0) by default, which helps us identify issues early. One of such issues is if it (pubsub) has too many subscribers (for example, too many `/broadcast_commit` calls waiting for txs results), we can't accept a new msg until we've broadcasted a previous one (to all subscribers). 

As an example, this https://github.com/tendermint/tendermint/blob/fb10209a9630688757d84c8dd28ec32bcb4a3e7d/consensus/state.go#L708 will block until previous msg is delivered to all subscribers. 

This can potentially lead to consensus liveness issues.

### Proposed action

We should rate limit the amount of `broadcast_commit` calls and other RPC endpoints which use pubsub (or recommend a limit for our clients to set in nginx or other software) + we need to figure out correct timeout for `/broadcast_commit` and make it configurable

https://github.com/tendermint/tendermint/blob/fb10209a9630688757d84c8dd28ec32bcb4a3e7d/rpc/core/mempool.go#L188-L189

melekes:
Add the following config variables (under [rpc] section):
  * max_subscription_clients
  * max_subscriptions_per_client
  * timeout_broadcast_tx_commit

Fixes #2826

* [ ] Wrote tests

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -323,6 +323,17 @@ type RPCConfig struct {
 	// Should be < {ulimit -Sn} - {MaxNumInboundPeers} - {MaxNumOutboundPeers} - {N of wal, db and other open files}
 	// 1024 - 40 - 10 - 50 = 924 = ~900
 	MaxOpenConnections int `mapstructure:"max_open_connections"`
+
+	// Maximum number of unique clientIDs that can /subscribe
+	MaxSubscriptionClients int `mapstructure:"max_subscription_clients"`
+
+	// Maximum number of unique queries a given client can /subscribe to (note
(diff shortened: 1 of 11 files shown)


melekes:
I am going to rebase this branch against pubsub-2.0 PR and refactor HttpClient as per https://github.com/tendermint/tendermint/pull/3227#issuecomment-459707700

Issue #3217, opened 2019-01-28
Potential deadlocks as reported by go-deadlock


melekes:
```
POTENTIAL DEADLOCK: Inconsistent locking. saw this ordering in one goroutine:
happened before
bit_array.go:355 common.(*BitArray).Update { bA.mtx.Lock() } <<<<<
bit_array_test.go:198 common.TestUpdateNeverPanics { a.Update(b) }

happened after
bit_array.go:356 common.(*BitArray).Update { o.mtx.Lock() } <<<<<
bit_array_test.go:198 common.TestUpdateNeverPanics { a.Update(b) }

in another goroutine: happened before
bit_array.go:355 common.(*BitArray).Update { bA.mtx.Lock() } <<<<<
bit_array_test.go:199 common.TestUpdateNeverPanics { b.Update(a) }

happened after
bit_array.go:356 common.(*BitArray).Update { o.mtx.Lock() } <<<<<
bit_array_test.go:199 common.TestUpdateNeverPanics { b.Update(a) }

Other goroutines holding locks:
goroutine 100745 lock 0xc0022a4db0
bit_array.go:355 common.(*BitArray).Update { bA.mtx.Lock() } <<<<<
bit_array_test.go:199 common.TestUpdateNeverPanics { b.Update(a) }
```

Refs https://github.com/tendermint/tendermint/pull/3218

ebuchman:
I don't really understand this - where are the go-routines related to this test?

melekes:
It does not show go-routines (at least in this case). I guess we will need to examine TestUpdateNeverPanics.

ebuchman:
That's what I'm saying - TestUpdateNeverPanics seems to have no associated go-routines!

ancazamfir:
Yes, but I think the library assumes that if the code uses locks it is to safely access data across multiple go-routines. I think the order in bit array's Update() is wrong.

melekes:
```
panic: sync: negative WaitGroup counter

goroutine 28 [running]:
sync.(*WaitGroup).Add(0xc0001a0090, 0xffffffffffffffff)
        /usr/local/Cellar/go/1.14.4/libexec/src/sync/waitgroup.go:74 +0x2c0
sync.(*WaitGroup).Done(...)
        /usr/local/Cellar/go/1.14.4/libexec/src/sync/waitgroup.go:99
[...]

PR #3254, opened 2019-02-05
p2p/conn: don't hold stopMtx while waiting

ebuchman:
Test TestPEXReactorSeedModeFlushStop is non-deterministically failing with timeout. Note before we were holding the stopMtx while waiting for the sendRoutine to finish. Here we change it to not do that. It was also possible before to exit out of the sendRoutine due to error without ever closing the doneSendRoutine, so here we also ensure that doneSendRoutine is closed at the end of the sendRoutine!

* [ ] Wrote tests

### DIFF
--- a/p2p/conn/connection.go
+++ b/p2p/conn/connection.go
@@ -85,8 +85,8 @@ type MConnection struct {
 	errored       uint32
 	config        MConnConfig
 
-	// Closing quitSendRoutine will cause
-	// doneSendRoutine to close.
+	// Closing quitSendRoutine will cause the sendRoutine to eventually quit.
+	// doneSendRoutine is closed when the sendRoutine actually quits.
 	quitSendRoutine chan struct{}
 	doneSendRoutine chan struct{}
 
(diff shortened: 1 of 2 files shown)


melekes:
@thanethomson we have an issue about this https://github.com/tendermint/tendermint/issues/3217

p2p/conn/connection.go
```diff
@@ -200,8 +203,6 @@ func (c *MConnection) OnStart() error {
 	if err := c.BaseService.OnStart(); err != nil {
 		return err
 	}
-	c.quitSendRoutine = make(chan struct{})
```
melekes: why move? I thought initializing vars in OnStart() is safer since a service can be restarted, although we don't do it

ebuchman: Oh you're right. I wanted to avoid checking it for nil but can do something else there instead

ebuchman: Ok fixed. Those should never actually be nil by the time we're calling a stop function so we should be ok.

Issue #3256, opened 2019-02-05
TestNodeSetPrivValTCP non-deterministic test failure


ebuchman:
```
D[2019-02-05|19:58:15.769] Attempt to update stats for non-existent peer module=consensus peer=
I[2019-02-05|19:58:15.771] Received proposal                            module=consensus proposal="Proposal{73/0 (2C2FC859DACB39A7276CD5FDBFCED95B8BE085A37D760D4B45B2819ACBECB951:1:2F3FCF003935, -1) 2EA1C83119C0 @ 2019-02-05T19:58:15.766221907Z}"
I[2019-02-05|19:58:15.772] Received complete proposal block             module=consensus height=73 hash=2C2FC859DACB39A7276CD5FDBFCED95B8BE085A37D760D4B45B2819ACBECB951
I[2019-02-05|19:58:15.772] enterPrevote(73/0). Current: 73/0/RoundStepPropose module=consensus 
I[2019-02-05|19:58:15.774] enterPrevote: ProposalBlock is valid         module=consensus height=73 round=0
I[2019-02-05|19:58:15.777] Signed and pushed vote                       module=consensus height=73 round=0 vote="Vote{0:A3258DCBF45D 73/00/1(Prevote) 2C2FC859DACB E9705BAC1588 @ 2019-02-05T19:58:15.774839153Z}" err=null
D[2019-02-05|19:58:15.777] Broadcast                                    module=p2p channel=32 msgBytes=C96A6FA808491804
E[2019-02-05|19:58:15.777] OnStart                                      module=privval err="accept tcp 127.0.0.1:34236: i/o timeout"
D[2019-02-05|19:58:15.777] Attempt to update stats for non-existent peer module=consensus peer=
D[2019-02-05|19:58:15.777] addVote                                      module=consensus voteHeight=73 voteType=1 valIndex=0 csHeight=73
I[2019-02-05|19:58:15.778] Stopping RemoteSigner                        impl=RemoteSigner
--- FAIL: TestNodeSetPrivValTCP (3.01s)
    <autogenerated>:1: 
                          
	Error Trace:	node_test.go:146
        
	Error:      	Received unexpected error:
        
	            	accept tcp 127.0.0.1:34236: i/o timeout
        
	            	failed to start private validator
        
	            	github.com/tendermint/tendermint/node.createAndStartPrivValidatorSocketClient
        
	            		/go/src/github.com/tendermint/tendermint/node/node.go:915
        
[...]

PR #3257, opened 2019-02-05
node: decrease retry conn timeout in test

ebuchman:
Should fix https://github.com/tendermint/tendermint/issues/3256

Before we had the retry timeout set the same as the accept timeout, so it would fail often enough!

* [ ] Wrote tests


ebuchman:
I'm going on a crusade against our non-deterministic failures. Will kill a bunch in one PR

PR #3261, opened 2019-02-06
Revert "quick fix for CircleCI (#2279)"

melekes:
This reverts commit 1cf6712a36e8ecc843a68aa373748e89e0afecba.

Plus, switch from gometalinter to golangci-lint https://github.com/golangci/golangci-lint for speed and better support.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,65 @@
+run:
+  deadline: 1m
+
+linters:
+  enable-all: true
+  disable:
+    - gocyclo
+    - golint
+    - maligned
+    - errcheck
+    - staticcheck
+    - dupl
+    - ineffassign
+    - interfacer
+    - unconvert
+    - goconst
+    - unparam
+    - nakedret
+    - lll
+    - gochecknoglobals
+    - govet
+    - gocritic
+    - gosec
+    - gochecknoinits
+    - scopelint
+    - stylecheck
+    - deadcode
+    - prealloc
+    - unused
+    - gosimple
+
+# linters-settings:
+#   govet:
+#     check-shadowing: true
+#   golint:
+#     min-confidence: 0
+#   gocyclo:
+#     min-complexity: 10
+#   maligned:
+#     suggest-new: true
+#   dupl:
+#     threshold: 100
+#   goconst:
+#     min-len: 2
+#     min-occurrences: 2
+#   depguard:
+#     list-type: blacklist
+#     packages:
+#       # logging is allowed only by logutils.Log, logrus
+#       # is allowed to use only in logutils package
+#       - github.com/sirupsen/logrus
+#   misspell:
+#     locale: US
+#   lll:
+#     line-length: 140
+#   goimports:
+#     local-prefixes: github.com/golangci/golangci-lint
+#   gocritic:
+#     enabled-tags:
+#       - performance
+#       - style
+#       - experimental
+#     disabled-checks:
+#       - wrapperFunc
+#       - commentFormatting # https://github.com/go-critic/go-critic/issues/755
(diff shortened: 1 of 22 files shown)


Liamsi:
ref: https://github.com/tendermint/tendermint/pull/2279
background: saving cache for PR from forks was broken on circleci
Is this fixed now?

melekes:
> Is this fixed now?

looks like it

.golangci.yml
```diff
@@ -0,0 +1,65 @@
...
+    - dupl
+    - ineffassign
+    - interfacer
+    - unconvert
+    - goconst
+    - unparam
+    - nakedret
+    - lll
+    - gochecknoglobals
+    - govet
```
Liamsi: We do not pass govet? Wasn't this enabled in metalinter? 
https://github.com/tendermint/tendermint/pull/3261/commits/da33dd04cc90f60f339afa8182f59294e67d151b#diff-b67911656ef5d18c4ae36cb6741b7965L252

.golangci.yml
```diff
@@ -0,0 +1,65 @@
...
+    - gocritic
+    - gosec
```
Liamsi: same, wasn't this previously enabled?

.golangci.yml
```diff
@@ -0,0 +1,65 @@
...
+    - gochecknoinits
+    - scopelint
+    - stylecheck
+    - deadcode
```
Liamsi: same:
https://github.com/tendermint/tendermint/pull/3261/commits/da33dd04cc90f60f339afa8182f59294e67d151b#diff-b67911656ef5d18c4ae36cb6741b7965L254

.golangci.yml
melekes: probably because of the older version

Liamsi: Weird, I thought govet is just part of the golang distribution (the `go vet` command).

Liamsi:
I think, we should merge this as is and do a follow-up PR which re-enables all previously running linters (vet, gosec, ...?) and do the changes that are necessary due to the newer versions of these linters (and decide / discuss if want to enable a few more).

.golangci.yml
melekes: Or maybe gometalinter was broken

melekes:
Yep, created https://github.com/tendermint/tendermint/issues/3262

blockchain/pool.go
```diff
@@ -363,23 +363,23 @@ func (pool *BlockPool) sendError(err error, peerID p2p.ID) {
...
-			str += fmt.Sprintf("H(%v):X ", h)
-		} else {
-			str += fmt.Sprintf("H(%v):", h)
-			str += fmt.Sprintf("B?(%v) ", pool.requesters[h].block != nil)
-		}
-	}
-	return str
-}
+// for debugging purposes
+// func (pool *BlockPool) debug() string {
```
Liamsi: I wish there was a way to tell the linter: ignore this function (like for instance with clippy in rust).

Liamsi: Wait, doesn't `nolint:` work for golangci, too?

melekes: It does. Do you think we should add such labels instead of commenting?

Liamsi: Yes, (in the case of deadcode) if we consider the code to be useful, I would prefer a nolint there. If it's clear that it isn't useful and won't be used (for debugging or alike), I would prefer to delete it (even better).

melekes: Done

Liamsi: Awesome! This also makes it easier to search for these places.

PR #3269, opened 2019-02-07
limit number of /subscribe clients and queries per client

### LINKED ISSUE #2826: rpc: too many broadcast_commit calls can lead to consensus liveness issues

melekes:
Fixes #2826

Add the following config variables (under [rpc] section):
  * max_subscription_clients
  * max_subscriptions_per_client
  * timeout_broadcast_tx_commit

The PR is tagged with `breaking` label because of the changes to `rpc/client`.

```go
	// Subscribe subscribes given subscriber to query. Returns a channel onto
	// which events are published. An error is returned if it fails to subscribe.
	// outCapacity can be used optionally to set capacity for the channel.
	//
	// ctx cannot be used to unsubscribe. To unsubscribe, use either Unsubscribe
	// or UnsubscribeAll.
	Subscribe(ctx context.Context, subscriber, query string, outCapacity ...int) (out <-chan *ctypes.ResultEvent, err error)
	// Unsubscribe unsubscribes given subscriber from query.
	Unsubscribe(ctx context.Context, subscriber, query string) error
	// UnsubscribeAll unsubscribes given subscriber from all the queries.
	UnsubscribeAll(ctx context.Context, subscriber string) error

```

* [ ] ~~Wrote tests~~
✅ Tested manually using tm-bench and different modifications of https://gist.github.com/melekes/9d2c98874737fc949aa93bbca5d9db92 We still need to write unit tests for reconnection, but can be done later imho.


melekes:
I am going to refactor HttpClient as per [#3227 (comment)](https://github.com/tendermint/tendermint/pull/3227#issuecomment-459707700)

melekes:
```go
type EventsClient interface {
	Subscribe(subscriber string, query string, callback func(event *ctypes.ResultEvent, err error)) error
	Unsubscribe(subscriber string, query string) error
	UnsubscribeAll(subscriber string) error
}
```

rpc/core/pipe.go
```diff
@@ -71,6 +73,15 @@ var (
 	mempool          *mempl.Mempool
 
 	logger log.Logger
+
+	// config variables
+
+	// MaxSubscriptionClients mirrors RPCConfig.MaxSubscriptionClients
+	MaxSubscriptionClients    int
```
thanethomson: Linter is failing here:

```
+ make lint
--> Running linter
rpc/core/pipe.go:80: File is not `gofmt`-ed with `-s` (gofmt)
	MaxSubscriptionClients    int
Makefile:247: recipe for target 'lint' failed
make: *** [lint] Error 1
Exited with code 2
```

thanethomson:
```
--- FAIL: TestIndexerServiceIndexesBlocks (0.10s)
    <autogenerated>:1: 
                          
	Error Trace:	indexer_service_test.go:60
        
	Error:      	Not equal: 
        
	            	expected: &types.TxResult{Height:1, Index:0x0, Tx:types.Tx{0x66, 0x6f, 0x6f}, Result:types.ResponseDeliverTx{Code:0x0, Data:[]uint8(nil), Log:"", Info:"", GasWanted:0, GasUsed:0, Tags:[]common.KVPair(nil), Codespace:"", XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}}
        
	            	actual  : (*types.TxResult)(nil)
        
	            	
        
	            	Diff:
        
	            	--- Expected
        
	            	+++ Actual
        
	            	@@ -1,7 +1,2 @@
        
	            	-(*types.TxResult)({
        
	            	- Height: (int64) 1,
        
	            	- Index: (uint32) 0,
        
	            	- Tx: (types.Tx) (len=3) Tx{666F6F},
        
	            	- Result: (types.ResponseDeliverTx) 
        
	            	-})
        
	            	+(*types.TxResult)(<nil>)
        
	            	 
        
	Test:       	TestIndexerServiceIndexesBlocks
    <autogenerated>:1: 
                          
	Error Trace:	indexer_service_test.go:63
        
	Error:      	Not equal: 
        
	            	expected: &types.TxResult{Height:1, Index:0x1, Tx:types.Tx{0x62, 0x61, 0x72}, Result:types.ResponseDeliverTx{Code:0x0, Data:[]uint8(nil), Log:"", Info:"", GasWanted:0, GasUsed:0, Tags:[]common.KVPair(nil), Codespace:"", XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}}
        
	            	actual  : (*types.TxResult)(nil)
        
	            	
        
	            	Diff:
        
	            	--- Expected
        
	            	+++ Actual
        
	            	@@ -1,7 +1,2 @@
        
	            	-(*types.TxResult)({
        
	            	- Height: (int64) 1,
        
	            	- Index: (uint32) 1,
        
	            	- Tx: (types.Tx) (len=3) Tx{626172},
        
	            	- Result: (types.ResponseDeliverTx) 
        
	            	-})
        
	            	+(*types.TxResult)(<nil>)
        
	            	 
        
	Test:       	TestIndexerServiceIndexesBlocks
FAIL
```

rpc/core/pipe.go
melekes: Sorry, fixed it

node/node.go
```diff
@@ -658,6 +658,9 @@ func (n *Node) ConfigureRPC() {
 	rpccore.SetConsensusReactor(n.consensusReactor)
 	rpccore.SetEventBus(n.eventBus)
 	rpccore.SetLogger(n.Logger.With("module", "rpc"))
+	rpccore.MaxSubscriptionClients = n.config.RPC.MaxSubscriptionClients
```
ebuchman: Should we just set a `rpccore.Config` object with `n.config.RPC` instead?

rpc/client/httpclient.go
```diff
@@ -310,75 +289,61 @@ func (w *WSEvents) OnStart() error {
...
 
-	err := w.ws.Subscribe(ctx, q)
-	if err != nil {
+	if err := w.ws.Subscribe(ctx, query); err != nil {
 		return nil, err
 	}
 
-	outCap := 1
-	if len(outCapacity) > 0 && outCapacity[0] >= 0 {
+	outCap := 0
```
ebuchman: WSEvents is default unbuffered?

rpc/core/mempool.go
```diff
@@ -198,8 +206,6 @@ func BroadcastTxCommit(tx types.Tx) (*ctypes.ResultBroadcastTxCommit, error) {
 	}
 
 	// Wait for the tx to be included in a block or timeout.
-	// TODO: configurable?
-	var deliverTxTimeout = rpcserver.WriteTimeout / 2
```
ebuchman: We require `TimeoutBroadcastTxCommit < WriteTimeout`, so either need to add that to config too with note or make it a function of TimeoutBroadcastTxCommit. In either case it all smells a bit - maybe broadcast_tx_commit is inherently bad

rpc/core/pipe.go
```diff
@@ -71,6 +73,15 @@ var (
 	mempool          *mempl.Mempool
 
 	logger log.Logger
+
+	// config variables
```
ebuchman: RPCConfig!

rpc/lib/server/handlers.go
```diff
@@ -434,8 +433,8 @@ type wsConnection struct {
 	// Send pings to server with this period. Must be less than readWait, but greater than zero.
 	pingPeriod time.Duration
 
-	// object that is used to subscribe / unsubscribe from events
-	eventSub types.EventSubscriber
+	// callback which is called upon disconnect
+	onDisconnect func(remoteAddr string)
```
ebuchman: What prompted this?

node/node.go
melekes: Won't we have a DATA RACE then?

config/config.go
```diff
@@ -323,6 +323,17 @@ type RPCConfig struct {
...
 	// 1024 - 40 - 10 - 50 = 924 = ~900
 	MaxOpenConnections int `mapstructure:"max_open_connections"`
+
+	// Maximum number of unique clientIDs that can /subscribe
+	MaxSubscriptionClients int `mapstructure:"max_subscription_clients"`
+
+	// Maximum number of unique queries a given client can /subscribe to (note
+	// all calls to /broadcast_tx_commit uses the same client, so this would
+	// limit the number of /broadcast_tx_commit calls that can be open at once)
+	MaxSubscriptionsPerClient int `mapstructure:"max_subscriptions_per_client"`
```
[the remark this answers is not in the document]
melekes: The right thing to do, I think, is to expose IP of the client somehow and use it. Let me try to do it.

node/node.go
melekes: Not if we copy

rpc/lib/server/handlers.go
melekes: Bad design? The fact that low-level rpc library is highly-coupled with the `types` package and knows there's a pubsub (thing it can use to Subscribe/Unsubscribe) is a bad design. I realized we can Unsubscribe via a disconnect callback and Subscribe in the calling code (`lite/proxy`).

node/node.go
```diff
@@ -675,8 +676,15 @@ func (n *Node) startRPC() ([]net.Listener, error) {
 	for i, listenAddr := range listenAddrs {
 		mux := http.NewServeMux()
 		rpcLogger := n.Logger.With("module", "rpc-server")
-		wm := rpcserver.NewWebsocketManager(rpccore.Routes, coreCodec, rpcserver.EventSubscriber(n.eventBus))
-		wm.SetLogger(rpcLogger.With("protocol", "websocket"))
+		wmLogger := rpcLogger.With("protocol", "websocket")
+		wm := rpcserver.NewWebsocketManager(rpccore.Routes, coreCodec,
+			rpcserver.OnDisconnect(func(remoteAddr string) {
+				err := n.eventBus.UnsubscribeAll(context.Background(), remoteAddr)
```
melekes: same here

rpc/client/httpclient.go
```diff
@@ -310,75 +289,63 @@ func (w *WSEvents) OnStart() error {
...
-	err := w.ws.Stop()
-	if err != nil {
-		w.Logger.Error("failed to stop WSClient", "err", err)
-	}
+	_ = w.ws.Stop()
 }
 
-func (w *WSEvents) Subscribe(ctx context.Context, subscriber string, query tmpubsub.Query, outCapacity ...int) (types.Subscription, error) {
-	q := query.String()
+// Subscribe implements EventsClient by using WSClient to subscribe given
```
melekes: In the doc.go or at the top of this file write that the delivery is best effort. client will try to reconnect (with exponential timeout?)

rpc/client/httpclient.go
```diff
@@ -387,10 +354,14 @@ func (w *WSEvents) UnsubscribeAll(ctx context.Context, subscriber string) error
 // After being reconnected, it is necessary to redo subscription to server
 // otherwise no data will be automatically received.
 func (w *WSEvents) redoSubscriptions() {
+	const timeout = 5 * time.Second
```
melekes: ?

rpc/client/localclient.go
```diff
@@ -27,6 +33,7 @@ powerful control during testing, you probably want the "client/mock" package.
 */
 type Local struct {
```
melekes: don't quite understand why we need Local client

rpc/client/localclient.go
```diff
@@ -48,106 +56,177 @@ var (
...
+						if !c.IsRunning() {
+							return
+						}
+
+						ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+						defer cancel()
+						sub, err = c.EventBus.Subscribe(ctx, subscriber, q)
+						if err == nil {
+							break
+						}
```
melekes: exponential timeout?

rpc/client/httpclient.go
```diff
@@ -405,21 +376,33 @@ func (w *WSEvents) eventListener() {
 			if !ok {
 				return
 			}
+
 			if resp.Error != nil {
 				w.Logger.Error("WS error", "err", resp.Error.Error())
+				// we don't know which subscription failed, so redo all of them
+				// ErrAlreadySubscribed can be ignored
+				w.redoSubscriptions()
```
melekes: exponential timeout?

rpc/client/httpclient.go
melekes: No preference here. Changed to match pubsub's default cap = 1

lite/proxy/proxy.go
```diff
@@ -34,7 +35,12 @@ func StartProxy(c rpcclient.Client, listenAddr string, logger log.Logger, maxOpe
 	mux := http.NewServeMux()
 	rpcserver.RegisterRPCFuncs(mux, r, cdc, logger)
 
-	wm := rpcserver.NewWebsocketManager(r, cdc, rpcserver.EventSubscriber(c))
+	unsubscribeFromAllEvents := func(remoteAddr string) {
+		if err := c.UnsubscribeAll(context.Background(), remoteAddr); err != nil {
```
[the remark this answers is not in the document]
melekes: But if we blocked here forever, something is horribly wrong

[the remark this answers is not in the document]
melekes: maybe this is the sign pubsub does not need `context.Context` at all

melekes:
Every RPC function now has a *rpctypes.Context as a first parameter:

```
// Context is the first parameter for all functions. It carries a json-rpc
// request, http request and websocket connection.
//
// - JSONReq is non-nil when JSONRPC is called over websocket.
// - WSConn is non-nil when we're connected via a websocket.
// - HTTPReq is non-nil when URI or JSONRPC is called over HTTP.
type Context struct {
	// json-rpc request
	JSONReq *RPCRequest
	// websocket connection
	WSConn WSRPCConnection
	// http request
	HTTPReq *http.Request
}

// RemoteAddr returns either HTTPReq#RemoteAddr or result of the
// WSConn#GetRemoteAddr().
func (ctx *Context) RemoteAddr() string {
	if ctx.HTTPReq != nil {
		return ctx.HTTPReq.RemoteAddr
	} else if ctx.WSConn != nil {
		return ctx.WSConn.GetRemoteAddr()
	}
	return ""
}
```

config/config.go
```diff
@@ -382,6 +383,9 @@ func (cfg *RPCConfig) ValidateBasic() error {
 	if cfg.TimeoutBroadcastTxCommit < 0 {
 		return errors.New("timeout_broadcast_tx_commit can't be negative")
 	}
+	if cfg.TimeoutBroadcastTxCommit > rpcserver.WriteTimeout {
```
ebuchman: Probably need to handle this better but separate issue.

rpc/client/mock/client.go
```diff
@@ -76,61 +77,61 @@ func (c Call) GetResponse(args interface{}) (interface{}, error) {
 }
 
 func (c Client) Status() (*ctypes.ResultStatus, error) {
-	return core.Status()
+	return core.Status(&rpctypes.Context{})
```
ebuchman: Maybe it would be better if we just initialized an empty context once in `c.ctx` and pass that in everywhere? Should be easy find/replace but might be cleaner?

rpc/grpc/api.go
```diff
@@ -16,7 +17,8 @@ func (bapi *broadcastAPI) Ping(ctx context.Context, req *RequestPing) (*Response
 }
 
 func (bapi *broadcastAPI) BroadcastTx(ctx context.Context, req *RequestBroadcastTx) (*ResponseBroadcastTx, error) {
-	res, err := core.BroadcastTxCommit(req.Tx)
+	// XXX: empty context ... no remote addr!
```
ebuchman: Isn't grpc supposed to populate it with some stuff?

rpc/lib/server/handlers.go
```diff
@@ -204,13 +210,14 @@ func arrayParamsToArgs(rpcFunc *RPCFunc, cdc *amino.Codec, params []json.RawMess
...
+// raw is unparsed json (from json.RawMessage) encoding either a map or an
+// array.
 //
 // Example:
-//   rpcFunc.args = [rpctypes.WSRPCContext string]
+//   rpcFunc.args = [rpctypes.Context string]
 //   rpcFunc.argNames = ["arg"]
-func jsonParamsToArgs(rpcFunc *RPCFunc, cdc *amino.Codec, raw []byte, argsOffset int) ([]reflect.Value, error) {
+func jsonParamsToArgs(rpcFunc *RPCFunc, cdc *amino.Codec, raw []byte) ([]reflect.Value, error) {
+	const argsOffset = 1
```
ebuchman: I wonder if it was really worth adding the Context to every function or if we really only needed to add it to a select few - ie. replace it in the Subscrube/Unsubscribe and add it to the BroadcastTxCommit. Then we'd just be increasing the list of methods that have `argsOffset = 1`. 

Is there some reason we really need it on all the other methods?

rpc/lib/server/handlers.go
```diff
@@ -128,20 +128,26 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, cdc *amino.Codec, logger lo
...
 		}
+
 		rpcFunc := funcMap[request.Method]
 		if rpcFunc == nil || rpcFunc.ws {
 			WriteRPCResponseHTTP(w, types.RPCMethodNotFoundError(request.ID))
 			return
 		}
-		var args []reflect.Value
+
+		ctx := &types.Context{JSONReq: &request, HTTPReq: r}
```
ebuchman: The comments on `Context` say JSONReq is only non nil if its from a WS request.

lite/proxy/proxy.go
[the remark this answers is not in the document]
ebuchman: What does this have to do with zombie connections? What could cause this to block?

lite/proxy/proxy.go
```diff
@@ -51,13 +57,11 @@ func StartProxy(c rpcclient.Client, listenAddr string, logger log.Logger, maxOpe
...
-
 	return map[string]*rpcserver.RPCFunc{
 		// Subscribe/unsubscribe are reserved for websocket events.
-		// We can just use the core tendermint impl, which uses the
-		// EventSwitch we registered in NewWebsocketManager above
-		"subscribe":   rpcserver.NewWSRPCFunc(core.Subscribe, "query"),
-		"unsubscribe": rpcserver.NewWSRPCFunc(core.Unsubscribe, "query"),
+		"subscribe":       rpcserver.NewWSRPCFunc(c.(Wrapper).SubscribeWS, "query"),
+		"unsubscribe":     rpcserver.NewWSRPCFunc(c.(Wrapper).UnsubscribeWS, "query"),
+		"unsubscribe_all": rpcserver.NewWSRPCFunc(c.(Wrapper).UnsubscribeAllWS, ""),
```
ebuchman: Oh, this is a new route? Changelog

lite/proxy/wrapper.go
```diff
@@ -145,6 +149,53 @@ func (w Wrapper) Commit(height *int64) (*ctypes.ResultCommit, error) {
 	return res, err
 }
 
+// SubscribeWS subscribes for events using the given query and remote address as
+// a subscriber, but does not verify responses (FIXME)!
```
ebuchman: verify responses?

rpc/client/httpclient.go
ebuchman: ?

rpc/client/localclient.go
ebuchman: I would think for testing and other in-process use to avoid HTTP connections.

rpc/client/interface.go
```diff
@@ -91,7 +93,18 @@ type NetworkClient interface {
 // EventsClient is reactive, you can subscribe to any message, given the proper
 // string. see tendermint/types/events.go
 type EventsClient interface {
-	types.EventBusSubscriber
+	// Subscribe subscribes given subscriber to query. Returns an unbuffered
```
ebuchman: Doesn't it return a chan with cap 1?

rpc/client/localclient.go
```diff
@@ -48,106 +56,177 @@ var (
...
+
+	outCap := 1
+	if len(outCapacity) > 0 {
+		outCap = outCapacity[0]
+	}
 
-func (c *Local) Unsubscribe(ctx context.Context, subscriber string, query tmpubsub.Query) error {
-	return c.EventBus.Unsubscribe(ctx, subscriber, query)
+	outc := make(chan ctypes.ResultEvent, outCap)
+	go func(sub types.Subscription) {
```
ebuchman: Can this be a method instead of closure? `eventsRoutine` ?

rpc/client/httpclient.go
```diff
@@ -3,25 +3,30 @@ package client
...
+(Local), or when you want to mock out the server for test code (mock).
 
-This is the main implementation you probably want to use in
-production code.  There are other implementations when calling
-the tendermint node in-process (local), or when you want to mock
-out the server for test code (mock).
+You can subscribe for any event published by Tendermint using Subscribe method.
+Note delivery is best-effort. If you don't read events fast enough or network
+is slow, Tendermint might cancel the subscription. The client will attempt to
+resubscribe (you don't need to do anything).
```
ebuchman: Do we need to build in the resubscription right away? I guess before it wasn't supposed to be possible to loose messages, because it blocked until you read it. Now the event stream could fail if you're too slow, even if the WS is still active, so we could try to reconnect server side, but what if you meanwhile missed some events, or otherwise can't reconnect? We'll need to expose that to the client so they can figure out what they missed, or else implement much more complex logic server side to do it for them ...

rpc/client/localclient.go
```diff
@@ -48,106 +56,177 @@ var (
...
+				} else {
+					select {
+					case outc <- result:
+					default:
+						c.Logger.Error("wanted to publish ResultEvent, but out channel is full", "result", result, "query", result.Query)
+					}
+				}
+			case <-sub.Cancelled():
+				if sub.Err() != tmpubsub.ErrUnsubscribed {
+					// resubscribe
```
ebuchman: Is this tested?

rpc/client/localclient.go
ebuchman: Should we consider closing the entire WS connection if an event is missed and pushing this responsibility on the client?

rpc/core/mempool.go
```diff
@@ -165,18 +165,26 @@ func BroadcastTxSync(tx types.Tx) (*ctypes.ResultBroadcastTx, error) {
 // | Parameter | Type | Default | Required | Description     |
 // |-----------+------+---------+----------+-----------------|
 // | tx        | Tx   | nil     | true     | The transaction |
-func BroadcastTxCommit(tx types.Tx) (*ctypes.ResultBroadcastTxCommit, error) {
+func BroadcastTxCommit(ctx *rpctypes.Context, tx types.Tx) (*ctypes.ResultBroadcastTxCommit, error) {
+	subscriber := ctx.RemoteAddr()
+
+	if eventBus.NumClients() > config.MaxSubscriptionClients {
```
ebuchman: Should we deduplicate with Subscribe?

rpc/lib/server/handlers.go
ebuchman: Oh goodness, excellent point!

rpc/lib/types/types.go
```diff
@@ -1,18 +1,15 @@
...
+	"net/http"
 	"reflect"
 	"strings"
 
 	"github.com/pkg/errors"
 
 	amino "github.com/tendermint/go-amino"
-
-	tmpubsub "github.com/tendermint/tendermint/libs/pubsub"
-	tmtypes "github.com/tendermint/tendermint/types"
```
ebuchman: ++!

rpc/grpc/api.go
melekes: From https://stackoverflow.com/questions/33684570/session-and-remote-ip-address-in-grpc-go it looks like it doesn't.

rpc/lib/server/handlers.go
melekes: The code can be further simplified now that we have context everywhere.

> Is there some reason we really need it on all the other methods?

Other - no. Subscribe/Unsubscribe/UnsubscribeAll/Broadcast - yes. Note the latter group can include future methods.

But having some methods with context and others without will be super confusing for developers and inconsistent.

lite/proxy/wrapper.go
melekes: yeah, blocks/block headers, etc.

rpc/client/httpclient.go
melekes: > I guess before it wasn't supposed to be possible to loose messages, because it blocked until you read it.

If a client does not read messages, server's output queue will become full and `TryWriteRPCResponse` will return false.

> Do we need to build in the resubscription right away?

It's not idea, but a good start. We don't have to spend too much time on it, but I think having resubscription is beneficial for any client.

lite/proxy/proxy.go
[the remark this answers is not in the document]
melekes: If indexer panicked and not reading messages for example.

rpc/core/mempool.go
melekes: Probably not worth it

rpc/client/httpclient.go
melekes: Actually, when a channel is unbuffered it is harder to miss the message

melekes:
Tested manually using tm-bench and different modifications of https://gist.github.com/melekes/9d2c98874737fc949aa93bbca5d9db92

rpc/client/localclient.go
```diff
@@ -39,113 +47,189 @@ func NewLocal(node *nm.Node) *Local {
...
+		}
+	}
 }
 
-func (c *Local) Unsubscribe(ctx context.Context, subscriber string, query tmpubsub.Query) error {
-	return c.EventBus.Unsubscribe(ctx, subscriber, query)
+// try to resubscribe with exponential timeout
+func (c *Local) resubscribe(subscriber string, q tmpubsub.Query) types.Subscription {
+	attempts := 0
+	for {
```
Liamsi: Is it intentional that there is no max number of attempts?

rpc/core/abci.go
```diff
@@ -52,7 +53,7 @@ import (
 // | data      | []byte | false   | true     | Data                                           |
 // | height    | int64  | 0       | false    | Height (0 means latest)                        |
 // | prove     | bool   | false   | false    | Includes proof if true                         |
-func ABCIQuery(path string, data cmn.HexBytes, height int64, prove bool) (*ctypes.ResultABCIQuery, error) {
+func ABCIQuery(ctx *rpctypes.Context, path string, data cmn.HexBytes, height int64, prove bool) (*ctypes.ResultABCIQuery, error) {
```
Liamsi: Looks like the context is not used in rpc/core but passed around a lot?

rpc/core/abci.go
```diff
@@ -96,7 +97,7 @@ func ABCIQuery(path string, data cmn.HexBytes, height int64, prove bool) (*ctype
 // 	"jsonrpc": "2.0"
 // }
 // ```
-func ABCIInfo() (*ctypes.ResultABCIInfo, error) {
+func ABCIInfo(ctx *rpctypes.Context) (*ctypes.ResultABCIInfo, error) {
```
Liamsi: ctx not used?

rpc/core/blocks.go
```diff
@@ -68,7 +69,7 @@ import (
 // ```
 //
 // <aside class="notice">Returns at most 20 items.</aside>
-func BlockchainInfo(minHeight, maxHeight int64) (*ctypes.ResultBlockchainInfo, error) {
+func BlockchainInfo(ctx *rpctypes.Context, minHeight, maxHeight int64) (*ctypes.ResultBlockchainInfo, error) {
```
Liamsi: ctx not used?

rpc/core/blocks.go
```diff
@@ -226,7 +227,7 @@ func filterMinMax(height, min, max, limit int64) (int64, int64, error) {
 //   "jsonrpc": "2.0"
 // }
 // ```
-func Block(heightPtr *int64) (*ctypes.ResultBlock, error) {
+func Block(ctx *rpctypes.Context, heightPtr *int64) (*ctypes.ResultBlock, error) {
```
Liamsi: ctx not used?

rpc/core/events.go
```diff
@@ -111,18 +118,27 @@ func Subscribe(wsCtx rpctypes.WSRPCContext, query string) (*ctypes.ResultSubscri
...
 						resultEvent,
 					))
 			case <-sub.Cancelled():
-				wsCtx.TryWriteRPCResponse(
-					rpctypes.RPCServerError(rpctypes.JSONRPCStringID(
-						fmt.Sprintf("%v#event", wsCtx.Request.ID)),
-						fmt.Errorf("subscription was cancelled (reason: %v)", sub.Err()),
-					))
+				if sub.Err() != tmpubsub.ErrUnsubscribed {
+					// should not happen
```
Liamsi: the comment `should not happen` does not help understanding what is going on here. Looking at the lines of code below it seems like this can happen, e.g. when tendermint exited, or the subscription was cancelled?

rpc/core/health.go
```diff
@@ -31,6 +32,6 @@ import (
 // 	"jsonrpc": "2.0"
 // }
 // ```
-func Health() (*ctypes.ResultHealth, error) {
+func Health(ctx *rpctypes.Context) (*ctypes.ResultHealth, error) {
```
Liamsi: ctx n ot used. Also, the code in the comment does not use the ctx: [`result, err := client.Health()` ](https://github.com/tendermint/tendermint/pull/3269/files#diff-8f61cabd9713bfb1c674667f8dbdc189L21)

lite/proxy/wrapper.go
```diff
@@ -149,6 +153,53 @@ func (w Wrapper) RegisterOpDecoder(typ string, dec merkle.OpDecoder) {
...
+
+	go func() {
+		for {
+			select {
+			case resultEvent := <-out:
+				ctx.WSConn.TryWriteRPCResponse(
+					rpctypes.NewRPCSuccessResponse(
+						ctx.WSConn.Codec(),
+						rpctypes.JSONRPCStringID(fmt.Sprintf("%v#event", ctx.JSONReq.ID)),
+						resultEvent,
```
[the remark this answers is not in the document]
melekes: Yes. We should have a switch here that performs a validation depending on the event's type.

node/node.go
```diff
@@ -675,8 +677,15 @@ func (n *Node) startRPC() ([]net.Listener, error) {
...
+				if err != nil && err != tmpubsub.ErrSubscriptionNotFound {
```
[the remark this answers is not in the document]
melekes: > Because there are no subscriptions for remoteAddr?

yes

rpc/client/httpclient.go
```diff
@@ -310,116 +297,125 @@ func (w *WSEvents) OnStart() error {
...
 
-	err := w.ws.Subscribe(ctx, q)
-	if err != nil {
+	if err := w.ws.Subscribe(ctx, query); err != nil {
 		return nil, err
 	}
 
 	outCap := 1
-	if len(outCapacity) > 0 && outCapacity[0] >= 0 {
+	if len(outCapacity) > 0 {
```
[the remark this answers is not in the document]
melekes: I think it's OK to panic in case developer makes an error (sometimes). Like we should not cast types when `make(chan ..., int(X))` and let the code panic

rpc/core/abci.go
melekes: It's used only by a few functions.

config/config.go
```diff
@@ -337,6 +351,10 @@ func DefaultRPCConfig() *RPCConfig {
 
 		Unsafe:             false,
 		MaxOpenConnections: 900,
+
+		MaxSubscriptionClients:    100,
+		MaxSubscriptionsPerClient: 5,
+		TimeoutBroadcastTxCommit:  10 * time.Second,
```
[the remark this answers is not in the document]
melekes: No, because these were picked quasi randomly.

`TimeoutBroadcastTxCommit` should be less than 20s (`rpcserver.WriteTimeout`)
`MaxSubscriptionClients * MaxSubscriptionsPerClient` define total number of subscriptions. I don't think someone will want more than 500. Even 100 may be enough.

rpc/client/event_test.go
```diff
@@ -147,3 +147,9 @@ func TestTxEventsSentWithBroadcastTxSync(t *testing.T) {
 		})
 	}
 }
+
+// Test HTTPClient resubscribes upon disconnect && subscription error.
+// Test Local client resubscribes upon subscription error.
+func TestClientsResubscribe(t *testing.T) {
+	// TODO(melekes)
```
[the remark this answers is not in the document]
melekes: https://github.com/tendermint/tendermint/issues/3380

rpc/client/httpclient.go
```diff
@@ -310,116 +297,125 @@ func (w *WSEvents) OnStart() error {
...
 			if resp.Error != nil {
 				w.Logger.Error("WS error", "err", resp.Error.Error())
+				// Error can be ErrAlreadySubscribed or max client (subscriptions per
+				// client) reached or Tendermint exited.
+				// We can ignore ErrAlreadySubscribed, but need to retry in other
+				// cases.
+				if !isErrAlreadySubscribed(resp.Error) {
+					// Resubscribe after 1 second to give Tendermint time to restart (if
+					// crashed).
+					w.redoSubscriptionsAfter(1 * time.Second)
```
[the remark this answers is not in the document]
melekes: The logic here is a bit complex to follow. redoSubscriptionsAfter calls Subscribe. If we receive another error in eventListener, we call redoSubscriptionsAfter again.

rpc/client/localclient.go
melekes: Yes.

rpc/core/events.go
melekes: > Looking at the lines of code below it seems like this can happen, e.g. when tendermint exited, or the subscription was cancelled?

Yes, but that **should not happen** normally.

rpc/core/health.go
melekes: Comment is correct. RPC client has a different API than our core in-house RPC API.

rpc/core/pipe.go
```diff
@@ -71,6 +72,8 @@ var (
 	mempool          *mempl.Mempool
 
 	logger log.Logger
+
+	config cfg.RPCConfig
```
Liamsi: Orthogonal to this PR but why don't we wrap these properties in a struct instead of global vars?

rpc/grpc/grpc_test.go
```diff
@@ -25,9 +25,8 @@ func TestMain(m *testing.M) {
 }
 
 func TestBroadcastTx(t *testing.T) {
-	require := require.New(t)
 	res, err := rpctest.GetGRPCClient().BroadcastTx(context.Background(), &core_grpc.RequestBroadcastTx{Tx: []byte("this is a tx")})
-	require.Nil(err, "%+v", err)
-	require.EqualValues(0, res.CheckTx.Code)
-	require.EqualValues(0, res.DeliverTx.Code)
+	require.Nil(t, err, "%+v", err)
```
Liamsi:
```suggestion
	require.NoError(t, err)
```

rpc/lib/server/handlers.go
```diff
@@ -280,10 +280,13 @@ func makeHTTPHandler(rpcFunc *RPCFunc, cdc *amino.Codec, logger log.Logger) func
 // Covert an http query to a list of properly typed values.
 // To be properly decoded the arg must be a concrete type from tendermint (if its an interface).
 func httpParamsToArgs(rpcFunc *RPCFunc, cdc *amino.Codec, r *http.Request) ([]reflect.Value, error) {
-	values := make([]reflect.Value, len(rpcFunc.args))
+	// skip types.Context
```
Liamsi: Nit: In general it is cool to say _why_ we skip something, too.

rpc/core/pipe.go
melekes: I don't know, but we should definitely try to keep number of global vars as min as possible. Refs https://github.com/tendermint/tendermint/issues/3367

PR #3275, opened 2019-02-07
types.NewCommit

ebuchman:
More work towards #1648 

* [ ] Wrote tests

### DIFF
--- a/types/block.go
+++ b/types/block.go
@@ -516,6 +516,14 @@ type Commit struct {
 	bitArray *cmn.BitArray
 }
 
+// NewCommit returns a new Commit with the given blockID and precommits.
+func NewCommit(blockID BlockID, precommits []*CommitSig) *Commit {
+	return &Commit{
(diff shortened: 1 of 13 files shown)


types/block.go
```diff
@@ -543,27 +555,20 @@ func (commit *Commit) memoizeHeightRound() {
...
-// reconstruct the vote.
 func (commit *Commit) ToVote(cs *CommitSig) *Vote {
+	// TODO: use commit.validatorSet to reconstruct vote
+	// and deprecate .toVote
 	return cs.toVote()
 }
 
 // Height returns the height of the commit
 func (commit *Commit) Height() int64 {
-	if len(commit.Precommits) == 0 {
```
melekes: So now a commit always has some precommits?

ebuchman: This already should have been the case, but regardless the check exists in memoizeHeightAndRound and the default value is still 0

Issue #3273, opened 2019-02-07
Cgo build tags


ebuchman:
In `libs/db` we use `// +build gcc` for the c_level_db.

But in the recently merged `crypto/secp256k1` work we use `// +build cgo`.

Probably we want better control over these and for them to be less confusing.

Perhaps we should change them to, eg. `tm-c-leveldb` and `tm-c-secp256k1` respectively? This way folks importing tendermint can pick and choose their build flags rather than being forced to get all C libraries at once.

ebuchman:
See minor follow up review in #3234

ebuchman:
build tags need to be alphanumeric. So we can just have them be like `// +build cleveldb` and `// +build libsecp256k1`

marbar3778:
seems this was closed with https://github.com/tendermint/tendermint/pull/3628 @melekes, could you close if so.

PR #3277, opened 2019-02-08
secp256k1: change build tags

ebuchman:
Change the build tag to be specific to the lib so we don't always try to build any time CGO is enabled. 
Ref #3273

* [ ] Wrote tests


ebuchman:
> If we do not care about this being the default and explicitly want users to specify the libsecp256k1 tag for this to happen, this PR LGTM.

We can address this later. If we do this we have to update install instructions to tell them to install that C lib

Issue #3213, opened 2019-01-28
Batch requests in JSON-RPC


tomtau:
JSON-RPC 2.0 specs allow sending multiple requests at a time: https://www.jsonrpc.org/specification#batch

Currently, Tendermint RPC will return a parser error if it receives an array of requests.

melekes:
Merged https://github.com/tendermint/tendermint/pull/3534 to develop. Check out https://godoc.org/github.com/tendermint/tendermint/rpc/client once v0.31.5 is out.

PR #3280, opened 2019-02-08
Add support for batched requests/responses in JSON RPC

thanethomson:
As per #3213, this PR will add support for batched JSON RPC requests/responses.

The suggestion here is to alter the HTTP RPC client interface to have a `batching` mode, which allows for the same calls as previously, but instead of sending the requests immediately, the `JSONRPCClient` will buffer the `RPCRequest` objects until `c.SendBatch()` is called:

```go
c := client.NewHTTP(rpcAddr, "/websocket")
c.StartBatch()
_, err := c.BroadcastTxCommit(tx1)
_, err = c.BroadcastTxCommit(tx2)
// nothing sent yet
results, err := c.SendBatch()
// batch is now sent, and one can iterate through `results` here...
// ...
// this request is now sent immediately, since we're no longer in batch mode
res, err := c.BroadcastTxCommit(tx3)
```

There are a few more things I'd like to test before I'd consider this finished though - will push those tests shortly.

* [x] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -9,5 +9,6 @@ Special thanks to external contributors on this release:
 ### FEATURES:
 
 ### IMPROVEMENTS:
+- [rpc] \#3280 Add support for batched requests/responses in JSON RPC
 
 ### BUG FIXES:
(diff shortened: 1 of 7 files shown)


thanethomson:
I'm closing this here and will re-open in a separate PR from a branch within the Tendermint repo (this original PR came from my own fork of Tendermint).

PR #3284, opened 2019-02-08
types: validator set update tests

ebuchman:
Minor follow up to #3222.

It's actually already very well tested! I just separated out some of the error cases into independent tests to make it easier to add more and modify. Pretty minor. Everything is passing :)

* [ ] Wrote tests

### DIFF
--- a/types/validator_set_test.go
+++ b/types/validator_set_test.go
@@ -664,88 +664,189 @@ type testVal struct {
 	power int64
 }
 
+func testValSet(nVals int, power int64) []testVal {
+	vals := make([]testVal, nVals)
+	for i := 0; i < nVals; i++ {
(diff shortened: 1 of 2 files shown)


types/validator_set_test.go
```diff
@@ -664,88 +664,189 @@ type testVal struct {
...
 
-func TestValSetUpdatesBasicTestsExecute(t *testing.T) {
-	valSetUpdatesBasicTests := []struct {
-		startVals    []testVal
-		updateVals   []testVal
-		expectedVals []testVal
-		expError     bool
-	}{
-		// Operations that should result in error
-		0: { // updates leading to overflows
```
melekes: I'd leave the indexes (0, 1, ...). If a test case fails, it's easy to locate it by searching for the index.

Issue #1034, opened 2017-12-29
Tests should not leave any residual files


melekes:
```
drwx------  3 ubuntu ubuntu  4096 Dec 28 20:37 blockchain_reactor_test
drwx------  3 ubuntu ubuntu  4096 Dec 28 20:37 blockchain_reactor_test_bak
drwx------  3 ubuntu ubuntu  4096 Dec 28 20:37 consensus_byzantine_test
drwx------  3 ubuntu ubuntu  4096 Dec 28 20:37 consensus_byzantine_test_0
drwx------  3 ubuntu ubuntu  4096 Dec 28 19:03 consensus_byzantine_test_0_bak
drwx------  3 ubuntu ubuntu  4096 Dec 28 20:37 consensus_byzantine_test_1
drwx------  3 ubuntu ubuntu  4096 Dec 28 19:03 consensus_byzantine_test_1_bak
drwx------  3 ubuntu ubuntu  4096 Dec 28 20:37 consensus_byzantine_test_2
drwx------  3 ubuntu ubuntu  4096 Dec 28 19:03 consensus_byzantine_test_2_bak
drwx------  3 ubuntu ubuntu  4096 Dec 28 20:37 consensus_byzantine_test_3
drwx------  3 ubuntu ubuntu  4096 Dec 28 19:03 consensus_byzantine_test_3_bak
drwx------  3 ubuntu ubuntu  4096 Dec 28 19:03 consensus_byzantine_test_bak
drwx------  3 ubuntu ubuntu  4096 Dec 28 20:37 consensus_height_vote_set_test
drwx------  3 ubuntu ubuntu  4096 Dec 28 19:03 consensus_height_vote_set_test_bak
drwx------  3 ubuntu ubuntu  4096 Dec 28 20:37 consensus_mempool_test
drwx------  3 ubuntu ubuntu  4096 Dec 28 19:03 consensus_mempool_test_bak
drwx------  3 ubuntu ubuntu  4096 Dec 28 20:37 consensus_mempool_txs_available_test
drwx------  3 ubuntu ubuntu  4096 Dec 28 20:37 consensus_mempool_txs_available_test_bak
drwx------  3 ubuntu ubuntu  4096 Dec 28 20:37 consensus_reactor_test
drwx------  3 ubuntu ubuntu  4096 Dec 28 20:37 consensus_reactor_test_0
drwx------  3 ubuntu ubuntu  4096 Dec 28 20:37 consensus_reactor_test_0_bak
[...]

PR #3291, opened 2019-02-10
R4R: improve ResetTestRootWithChainID() concurrency safety

alessio:
Rely on ioutil.TempDir() to create test root directories and ensure
multiple same-chain id test cases can run in parallel.

* [ ] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -20,6 +20,8 @@ Special thanks to external contributors on this release:
 
 ### IMPROVEMENTS:
 
+* \#3291 Make config.ResetTestRootWithChainID() create concurrency-safe test directories.
+
 ### BUG FIXES:
 - [consensus] \#3297 Flush WAL on stop to prevent data corruption during
   graceful shutdown
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -20,7 +20,7 @@ Special thanks to external contributors on this release:
 
 ### IMPROVEMENTS:
 
-* \#3291 Make config.ResetTestRootWithChainID() create concurrency-safe test directories.
+- [config] \#3291 Make config.ResetTestRootWithChainID() create concurrency-safe test directories.
 
 ### BUG FIXES:
 - [consensus] \#3297 Flush WAL on stop to prevent data corruption during
--- a/blockchain/store_test.go
+++ b/blockchain/store_test.go
@@ -22,13 +22,16 @@ import (
 	tmtime "github.com/tendermint/tendermint/types/time"
 )
 
+// A cleanupFunc cleans up any config / test files created for a particular test.
+type cleanupFunc func()
+
 // make a Commit with a single vote containing just the height and a timestamp
 func makeTestCommit(height int64, timestamp time.Time) *types.Commit {
 	commitSigs := []*types.CommitSig{{Height: height, Timestamp: timestamp}}
 	return types.NewCommit(types.BlockID{}, commitSigs)
 }
 
-func makeStateAndBlockStore(logger log.Logger) (sm.State, *BlockStore, func()) {
+func makeStateAndBlockStore(logger log.Logger) (sm.State, *BlockStore, cleanupFunc) {
 	config := cfg.ResetTestRoot("blockchain_reactor_test")
 	// blockDB := dbm.NewDebugDB("blockDB", dbm.NewMemDB())
 	// stateDB := dbm.NewDebugDB("stateDB", dbm.NewMemDB())
@@ -97,7 +100,7 @@ var (
 )
 
 func TestMain(m *testing.M) {
-	var cleanup func()
+	var cleanup cleanupFunc
 	state, _, cleanup = makeStateAndBlockStore(log.NewTMLogger(new(bytes.Buffer)))
 	block = makeBlock(1, state, new(types.Commit))
 	partSet = block.MakePartSet(2)
--- a/blockchain/store_test.go
+++ b/blockchain/store_test.go
@@ -3,6 +3,7 @@ package blockchain
 import (
 	"bytes"
 	"fmt"
+	"os"
 	"runtime/debug"
 	"strings"
 	"testing"
@@ -27,7 +28,7 @@ func makeTestCommit(height int64, timestamp time.Time) *types.Commit {
 	return types.NewCommit(types.BlockID{}, commitSigs)
 }
 
-func makeStateAndBlockStore(logger log.Logger) (sm.State, *BlockStore) {
+func makeStateAndBlockStore(logger log.Logger) (sm.State, *BlockStore, func()) {
 	config := cfg.ResetTestRoot("blockchain_reactor_test")
 	// blockDB := dbm.NewDebugDB("blockDB", dbm.NewMemDB())
 	// stateDB := dbm.NewDebugDB("stateDB", dbm.NewMemDB())
@@ -37,7 +38,7 @@ func makeStateAndBlockStore(logger log.Logger) (sm.State, *BlockStore) {
 	if err != nil {
 		panic(cmn.ErrorWrap(err, "error constructing state from genesis file"))
 	}
-	return state, NewBlockStore(blockDB)
+	return state, NewBlockStore(blockDB), func() { os.RemoveAll(config.RootDir) }
 }
 
 func TestLoadBlockStoreStateJSON(t *testing.T) {
@@ -87,19 +88,32 @@ func freshBlockStore() (*BlockStore, db.DB) {
 }
 
 var (
-	state, _ = makeStateAndBlockStore(log.NewTMLogger(new(bytes.Buffer)))
-
-	block       = makeBlock(1, state, new(types.Commit))
-	partSet     = block.MakePartSet(2)
-	part1       = partSet.GetPart(0)
-	part2       = partSet.GetPart(1)
-	seenCommit1 = makeTestCommit(10, tmtime.Now())
+	state       sm.State
+	block       *types.Block
+	partSet     *types.PartSet
+	part1       *types.Part
+	part2       *types.Part
+	seenCommit1 *types.Commit
 )
+
+func TestMain(m *testing.M) {
+	var cleanup func()
+	state, _, cleanup = makeStateAndBlockStore(log.NewTMLogger(new(bytes.Buffer)))
+	block = makeBlock(1, state, new(types.Commit))
+	partSet = block.MakePartSet(2)
+	part1 = partSet.GetPart(0)
+	part2 = partSet.GetPart(1)
+	seenCommit1 = makeTestCommit(10, tmtime.Now())
+	code := m.Run()
+	cleanup()
+	os.Exit(code)
+}
 
 // TODO: This test should be simplified ...
 
 func TestBlockStoreSaveLoadBlock(t *testing.T) {
-	state, bs := makeStateAndBlockStore(log.NewTMLogger(new(bytes.Buffer)))
+	state, bs, cleanup := makeStateAndBlockStore(log.NewTMLogger(new(bytes.Buffer)))
+	defer cleanup()
 	require.Equal(t, bs.Height(), int64(0), "initially the height should be zero")
 
 	// check there are no blocks at various heights
@@ -350,7 +364,8 @@ func TestLoadBlockMeta(t *testing.T) {
 }
 
 func TestBlockFetchAtHeight(t *testing.T) {
-	state, bs := makeStateAndBlockStore(log.NewTMLogger(new(bytes.Buffer)))
+	state, bs, cleanup := makeStateAndBlockStore(log.NewTMLogger(new(bytes.Buffer)))
+	defer cleanup()
 	require.Equal(t, bs.Height(), int64(0), "initially the height should be zero")
 	block := makeBlock(bs.Height()+1, state, new(types.Commit))
 
--- a/config/toml.go
+++ b/config/toml.go
@@ -327,10 +327,10 @@ func ResetTestRootWithChainID(testName string, chainID string) *Config {
 	rootDir := mustCreateTestRootDir(testName, chainID)
 	// Create new dir
 	if err := cmn.EnsureDir(filepath.Join(rootDir, defaultConfigDir), 0700); err != nil {
-		cmn.PanicSanity(err.Error())
+		panic(err)
 	}
 	if err := cmn.EnsureDir(filepath.Join(rootDir, defaultDataDir), 0700); err != nil {
-		cmn.PanicSanity(err.Error())
+		panic(err)
 	}
 
 	baseConfig := DefaultBaseConfig()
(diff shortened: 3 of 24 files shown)


config/toml.go
```diff
@@ -372,6 +361,18 @@ func ResetTestRootWithChainID(testName string, chainID string) *Config {
 	return config
 }
 
+func mustCreateTestRootDir(testName, chainID string) string {
+	homeDir := os.ExpandEnv("$HOME/.tendermint_test")
+	if err := os.MkdirAll(homeDir, 0777); err != nil {
```
Liamsi: Shouldn't this be `0700`? At least later we check for `cmn.EnsureDir(rootDir, 0700)`. 
```suggestion
	if err := os.MkdirAll(homeDir, 0700); err != nil {
```

config/toml.go
```diff
@@ -372,6 +361,18 @@ func ResetTestRootWithChainID(testName string, chainID string) *Config {
...
+		cmn.PanicSanity(err.Error())
+	}
+	rootDir, err := ioutil.TempDir(homeDir, fmt.Sprintf("%s-%s_", chainID, testName))
```
Liamsi: Won't this clutter the home directory with test-directories (per chain-id), which are not cleaned up afterwards?Maybe add a comment which mentions that the caller is responsible of cleaning up the dirs after they are not used anymore.

config/toml.go
alessio: Depends. On UNIX this will be modifed by the user's umask anyway, it will rarely ends up as 0777 really. Furthermore, why do we negate user's group access to a test directory? Is there any particular security concern? Not a big deal though, happy to amend it!

config/toml.go
alessio: I deliberately avoided calling `defer os.RemoveAll(rootDir)` 'cause I had noticed that the latest test directory was always left on the system. On the other hand you're right, chances are that after multiple runs the user's home will be filled with test data. Happy to add a defer os.RemoveAll() :+1:

melekes: Would be great to get rid of these directories https://github.com/tendermint/tendermint/issues/1034. But this will require a significant amount of time. Not sure it's worth doing right now unless you don't have anything else to do.

config/toml.go
```diff
@@ -322,20 +323,8 @@ func ResetTestRoot(testName string) *Config {
...
-	// Move ~/.tendermint_test to ~/.tendermint_test_bak
-	if cmn.FileExists(rootDir) {
-		if err := os.Rename(rootDir, rootDir+"_bak"); err != nil {
-			cmn.PanicSanity(err.Error())
-		}
-	}
+	// create a unique, concurrency-safe test directory under $HOME/.tendermint_test/
+	rootDir := mustCreateTestRootDir(testName, chainID)
 	// Create new dir
 	if err := cmn.EnsureDir(rootDir, 0700); err != nil {
```
melekes: note you create directory twice now. 1) ioutil.Tempdir 2) cmn.EnsureDir here

config/toml.go
```diff
@@ -372,6 +361,18 @@ func ResetTestRootWithChainID(testName string, chainID string) *Config {
 	return config
 }
 
+func mustCreateTestRootDir(testName, chainID string) string {
+	homeDir := os.ExpandEnv("$HOME/.tendermint_test")
+	if err := os.MkdirAll(homeDir, 0700); err != nil {
```
melekes: can we make `0700` a const

config/toml.go
```diff
@@ -372,6 +361,18 @@ func ResetTestRootWithChainID(testName string, chainID string) *Config {
...
+		cmn.PanicSanity(err.Error())
+	}
+	rootDir, err := ioutil.TempDir(homeDir, fmt.Sprintf("%s-%s_", chainID, testName))
+	if err != nil {
+		cmn.PanicSanity(err.Error())
```
melekes: `cmn.PanicSanity` is deprecated. use `panic`

config/toml.go
```diff
@@ -372,6 +361,18 @@ func ResetTestRootWithChainID(testName string, chainID string) *Config {
 	return config
 }
 
+func mustCreateTestRootDir(testName, chainID string) string {
```
melekes: why extract this code into a function? ResetTestRootWithChainID does exactly that. see my comment above about creating a directory twice (or rather attempting to cause cmn.Ensure won't do it if dir already exist)

config/toml.go
alessio: will do, thanks

config/toml.go
alessio: I think you're right, the `EnsureDir` following the `mustCreateTestRootDir` is basically a no-op. Will remove then, thanks.

config/toml.go
alessio: done - please test

config/toml.go
alessio: reverted

config/toml.go
melekes: wow. that was fast 🐇 ⚡️

consensus/byzantine_test.go
```diff
@@ -29,7 +26,10 @@ func init() {
 func TestByzantine(t *testing.T) {
 	N := 4
 	logger := consensusLogger().With("test", "byzantine")
-	css := randConsensusNet(N, "consensus_byzantine_test", newMockTickerFunc(false), newCounter)
+	css, rootDirs := randConsensusNet(N, "consensus_byzantine_test", newMockTickerFunc(false), newCounter)
+	for _, dir := range rootDirs {
```
melekes:
```suggestion
defer func() {
  for _, dir := range configRootDirs {
    _ = os.RemoveAll()
  }
}()
```

consensus/common_test.go
```diff
@@ -574,18 +579,22 @@ func randConsensusNet(nValidators int, testName string, tickerFunc func() Timeou
...
 
 // nPeers = nValidators + nNotValidator
-func randConsensusNetWithPeers(nValidators, nPeers int, testName string, tickerFunc func() TimeoutTicker, appFunc func() abci.Application) []*ConsensusState {
+func randConsensusNetWithPeers(nValidators, nPeers int, testName string, tickerFunc func() TimeoutTicker,
+	appFunc func() abci.Application) ([]*ConsensusState, []string) {
+
 	genDoc, privVals := randGenesisDoc(nValidators, false, testMinPower)
 	css := make([]*ConsensusState, nPeers)
 	logger := consensusLogger()
+	rootDirs := []string{}
```
melekes: configRootDirs

consensus/byzantine_test.go
```diff
@@ -29,7 +26,10 @@ func init() {
 func TestByzantine(t *testing.T) {
 	N := 4
 	logger := consensusLogger().With("test", "byzantine")
-	css := randConsensusNet(N, "consensus_byzantine_test", newMockTickerFunc(false), newCounter)
+	css, rootDirs := randConsensusNet(N, "consensus_byzantine_test", newMockTickerFunc(false), newCounter)
```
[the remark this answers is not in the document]
alessio: done

consensus/byzantine_test.go
alessio: How's this better than deferring `os.RemoveAll()` kinda atomic calls individually?

melekes: smaller defer stack? faster?

alessio: done

consensus/common_test.go
alessio: done

blockchain/store_test.go
```diff
@@ -27,7 +28,7 @@ func makeTestCommit(height int64, timestamp time.Time) *types.Commit {
 	return types.NewCommit(types.BlockID{}, commitSigs)
 }
 
-func makeStateAndBlockStore(logger log.Logger) (sm.State, *BlockStore) {
+func makeStateAndBlockStore(logger log.Logger) (sm.State, *BlockStore, func()) {
```
melekes: can we create an alias type for cleanup func?

```
// A cleanupFunc cleans up any config / test files created for a particular test.
type cleanupFunc func()
```

config/toml.go
```diff
@@ -322,29 +325,17 @@ func ResetTestRoot(testName string) *Config {
...
-	if cmn.FileExists(rootDir) {
-		if err := os.Rename(rootDir, rootDir+"_bak"); err != nil {
-			cmn.PanicSanity(err.Error())
-		}
-	}
-	// Create new dir
-	if err := cmn.EnsureDir(rootDir, 0700); err != nil {
-		cmn.PanicSanity(err.Error())
+	// create a unique, concurrency-safe test directory under $HOME/.tendermint_test/
+	rootDir, err := ioutil.TempDir("", fmt.Sprintf("%s-%s_", chainID, testName))
```
melekes:
```suggestion
	rootDir, err := ioutil.TempDir(os.ExpandEnv("$HOME/.tendermint_test/"), fmt.Sprintf("%s-%s_", chainID, testName))
```

consensus/common_test.go
```diff
@@ -554,14 +555,17 @@ func consensusLogger() log.Logger {
 	}).With("module", "consensus")
 }
 
-func randConsensusNet(nValidators int, testName string, tickerFunc func() TimeoutTicker, appFunc func() abci.Application, configOpts ...func(*cfg.Config)) []*ConsensusState {
+func randConsensusNet(nValidators int, testName string, tickerFunc func() TimeoutTicker,
```
melekes: why not return a cleanup function to be consistent with the above ^^^?

consensus/replay_test.go
```diff
@@ -142,6 +155,8 @@ LOOP:
 
 		// clean up WAL file from the previous iteration
 		walFile := cs.config.WalFile()
+		fmt.Printf("%s %s\n", walFile, filepath.Dir(walFile))
```
melekes: no fmt please (or replace with `t.Logf`)

alessio: oh sure will do

alessio: debugging leftover, I'll get rid of it

config/toml.go
alessio: From https://github.com/tendermint/tendermint/pull/3291/commits/ce1db7d579c0e97ff755a91ddb259c6ee6cf0544:

```
Place test directories in OS's tempdir

In modern UNIX and UNIX-like systems /tmp is very often
mounted as tmpfs. This might speed test execution a bit.
```

Any compelling reason why you want test dirs in your $HOME?

consensus/common_test.go
alessio: done

config/toml.go
melekes: I am not 100% sure, but we had some issues exactly because it's a different partition. But let's try /tmp again. Can you please change the comment 

```
// create a unique, concurrency-safe test directory under $HOME/.tendermint_test/
```
?

blockchain/store_test.go
alessio: done

config/toml.go
alessio: Done, thanks!

melekes:
Please do not rebase the branch next time 🙇  It makes impossible for Github to track changes (i.e. there's no "Changes since you last viewed" anymore) ☝️

CHANGELOG_PENDING.md
```diff
@@ -20,6 +20,8 @@ Special thanks to external contributors on this release:
 
 ### IMPROVEMENTS:
 
+* \#3291 Make config.ResetTestRootWithChainID() create concurrency-safe test directories.
```
```suggestion
- [config] \#3291 Make config.ResetTestRootWithChainID() create concurrency-safe test directories.
```

config/toml.go
```diff
@@ -3,13 +3,16 @@ package config
...
-	"os"
+	"io/ioutil"
 	"path/filepath"
 	"text/template"
 
 	cmn "github.com/tendermint/tendermint/libs/common"
 )
 
+// DefaultDirPerm is the default permissions used when creating directories.
+const DefaultDirPerm = 0700
```
melekes: ❤️

consensus/common_test.go
```diff
@@ -554,14 +559,17 @@ func consensusLogger() log.Logger {
...
+	appFunc func() abci.Application, configOpts ...func(*cfg.Config)) ([]*ConsensusState, cleanupFunc) {
 	genDoc, privVals := randGenesisDoc(nValidators, false, 30)
 	css := make([]*ConsensusState, nValidators)
 	logger := consensusLogger()
+	configRootDirs := []string{}
```
melekes:
```suggestion
	configRootDirs := make([]string, 0, nValidators)
```

consensus/common_test.go
```diff
@@ -574,18 +582,26 @@ func randConsensusNet(nValidators int, testName string, tickerFunc func() Timeou
...
 
 // nPeers = nValidators + nNotValidator
-func randConsensusNetWithPeers(nValidators, nPeers int, testName string, tickerFunc func() TimeoutTicker, appFunc func() abci.Application) []*ConsensusState {
+func randConsensusNetWithPeers(nValidators, nPeers int, testName string, tickerFunc func() TimeoutTicker,
+	appFunc func() abci.Application) ([]*ConsensusState, cleanupFunc) {
+
 	genDoc, privVals := randGenesisDoc(nValidators, false, testMinPower)
 	css := make([]*ConsensusState, nPeers)
 	logger := consensusLogger()
+	configRootDirs := []string{}
```
melekes:
```suggestion
	configRootDirs := make([]string, 0, nPeers)
```

consensus/types/height_vote_set_test.go
```diff
@@ -11,8 +12,11 @@ import (
 
 var config *cfg.Config // NOTE: must be reset for each _test.go file
 
-func init() {
+func TestMain(m *testing.M) {
```
melekes: ❤️

mempool/mempool_test.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/tendermint/tendermint/types"
 )
 
-func newMempoolWithApp(cc proxy.ClientCreator) *Mempool {
+func newMempoolWithApp(cc proxy.ClientCreator) (*Mempool, func()) {
```
melekes: `type cleanupFunc func()`

rpc/test/helpers.go
```diff
@@ -100,8 +100,8 @@ func GetGRPCClient() core_grpc.BroadcastAPIClient {
 }
 
 // StartTendermint starts a test tendermint server in a go routine and returns when it is initialized
-func StartTendermint(app abci.Application) *nm.Node {
-	node := NewTendermint(app)
+func StartTendermint(app abci.Application) (*nm.Node, func()) {
```
melekes: Can we create `StopTendermint()` function which removes a config directory?

```go
// node/node.go
func (n *Node) Config() *cfg.Config {
	return n.config
}

//  rpc/test/helpers.go
func StopTendermint() {
	node.Stop()
	node.Wait()
	os.RemoveAll(node.Config().RootDir)
}
```

alessio: If you don't mind, I will do it in an ensuing PR

melekes: Sure 👍

consensus/replay_test.go
```diff
@@ -142,6 +153,7 @@ LOOP:
 
 		// clean up WAL file from the previous iteration
 		walFile := cs.config.WalFile()
+		ensureDir(filepath.Dir(walFile), 0700)
```
melekes: why we need this?

Issue #3262, opened 2019-02-06
Make each golangci-lint linter pass


melekes:
From `.golangci.yml`:

- [ ] gocyclo
- [ ] golint
- [x] maligned
- [ ] errcheck
- [x] staticcheck
- [x] dupl https://github.com/tendermint/tendermint/pull/3385
- [x] ineffassign https://github.com/tendermint/tendermint/pull/3386
- [x] interfacer
- [x] unconvert
- [x] goconst
- [ ] unparam
- [x] nakedret
- [x] lll
- [ ] gochecknoglobals
- [x] govet https://github.com/tendermint/tendermint/pull/3292
  - [ ] remove https://github.com/tendermint/tendermint/pull/3292#discussion_r255386460 when we bump Golang version
- [x] gocritic
- [x] gosec https://github.com/tendermint/tendermint/pull/3294
- [ ] gochecknoinits
- [x] scopelint
- [ ] stylecheck

Run `make get_tools` to install golangci-lint or follow instructions at https://github.com/golangci/golangci-lint.

Run each individual linter with: `golangci-lint run --no-config --disable-all=true --enable=XXX`. For example: `golangci-lint run --no-config --disable-all=true --enable=govet`

Run all with: `golangci-lint run --enable-all=true`

Help make Tendermint great again!

Liamsi:
I would like to go through the lints of `gosec` (if any).

Liamsi:
I would like to tackle `golint`. How would we want to treat `type name will be used as blockchain.BlockchainMessage by other packages, and that stutters; consider calling this Reactor (golint)`? I would prefer to go with the golang idiomatic style here but I can imagine some pushback. Especially on structs like `BlockchainReactor` and some other big ones...

melekes:
I'd say remove the prefixes, but let's hear from @ebuchman first.

ebuchman:
Hm. I think one concern is grepability. For instance, will the linters enforce that we call the `blockchain` pkg the same thing every time we import it? Otherwise it could be hard to search the whole codebase for use of the blockchain reactor.

Liamsi:
> For instance, will the linters enforce that we call the blockchain pkg the same thing every time we import it?

Not that I'm aware of. On the other hand there is probably no good reason to alias the `blockchain` package and we will write `blockchain.BlockchainMessage` most of the time. We could still grep for the package to find aliases. 

[...]

PR #3292, opened 2019-02-11
make govet linter pass

melekes:
Refs #3262

### DIFF
--- a/crypto/encoding/amino/encode_test.go
+++ b/crypto/encoding/amino/encode_test.go
@@ -47,6 +47,8 @@ func checkAminoJSON(t *testing.T, src interface{}, dst interface{}, isNil bool) 
 	require.Nil(t, err, "%+v", err)
 }
 
+// ExamplePrintRegisteredTypes refers to unknown identifier: PrintRegisteredTypes
+//nolint:govet
 func ExamplePrintRegisteredTypes() {
 	cdc.PrintTypes(os.Stdout)
 	// Output: | Type | Name | Prefix | Length | Notes |
--- a/crypto/secp256k1/secp256k1_nocgo.go
+++ b/crypto/secp256k1/secp256k1_nocgo.go
@@ -52,8 +52,8 @@ func (pubKey PubKeySecp256k1) VerifyBytes(msg []byte, sigStr []byte) bool {
 // that len(sigStr) == 64.
 func signatureFromBytes(sigStr []byte) *secp256k1.Signature {
 	return &secp256k1.Signature{
-		new(big.Int).SetBytes(sigStr[:32]),
-		new(big.Int).SetBytes(sigStr[32:64]),
+		R: new(big.Int).SetBytes(sigStr[:32]),
+		S: new(big.Int).SetBytes(sigStr[32:64]),
 	}
 }
 
--- a/libs/common/bit_array.go
+++ b/libs/common/bit_array.go
@@ -412,7 +412,6 @@ func (bA *BitArray) UnmarshalJSON(bz []byte) error {
 			bA2.SetIndex(i, true)
 		}
 	}
-	// assignment copies lock value to *bA: github.com/tendermint/tendermint/libs/common.BitArray contains sync.Mutex
 	*bA = *bA2 //nolint:govet
 	return nil
 }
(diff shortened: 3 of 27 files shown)


crypto/secp256k1/secp256k1_nocgo.go
```diff
@@ -52,8 +52,8 @@ func (pubKey PubKeySecp256k1) VerifyBytes(msg []byte, sigStr []byte) bool {
 // that len(sigStr) == 64.
 func signatureFromBytes(sigStr []byte) *secp256k1.Signature {
 	return &secp256k1.Signature{
-		new(big.Int).SetBytes(sigStr[:32]),
-		new(big.Int).SetBytes(sigStr[32:64]),
+		R: new(big.Int).SetBytes(sigStr[:32]),
```
melekes: @Liamsi is this correct?

p2p/conn/secret_connection_test.go
```diff
@@ -245,11 +245,11 @@ func TestSecretConnectionReadWrite(t *testing.T) {
 						}
 						*nodeReads = append(*nodeReads, string(readBuffer[:n]))
 					}
-					if err := nodeConn.PipeReader.Close(); err != nil {
-						t.Error(err)
-						return nil, err, true
-					}
-					return nil, nil, false
+					// if err := nodeConn.PipeReader.Close(); err != nil {
```
melekes: remove?

crypto/encoding/amino/encode_test.go
```diff
@@ -47,6 +47,8 @@ func checkAminoJSON(t *testing.T, src interface{}, dst interface{}, isNil bool)
 	require.Nil(t, err, "%+v", err)
 }
 
+// ExamplePrintRegisteredTypes refers to unknown identifier: PrintRegisteredTypes
+//nolint:govet
 func ExamplePrintRegisteredTypes() {
```
melekes: can't find PrintRegisteredTypes

libs/common/bit_array.go
```diff
@@ -412,6 +412,7 @@ func (bA *BitArray) UnmarshalJSON(bz []byte) error {
 			bA2.SetIndex(i, true)
 		}
 	}
-	*bA = *bA2
+	// assignment copies lock value to *bA: github.com/tendermint/tendermint/libs/common.BitArray contains sync.Mutex
+	*bA = *bA2 //nolint:govet
```
melekes: ok to copy the lock?

crypto/secp256k1/secp256k1_nocgo.go
Liamsi: Yes, that is correct!

crypto/encoding/amino/encode_test.go
Liamsi: https://github.com/golang/go/issues/29570
(fixed on tip)

p2p/conn/secret_connection_test.go
Liamsi: You can definitely remove without changing the existing semantics here. What the developer probably wanted, was that the connection gets closed on an error/when the control flow exits the loop. Sounds like this clsoing code should be moved to a defer (using the outer function though)?

Liamsi:
```go
defer func() {
				if err := nodeConn.PipeReader.Close(); err != nil {
					t.Error(err)
				}
			}()
```
on line 206?

libs/common/bit_array.go
Liamsi: I think  it is OK here: the object is just being constructed ... If you want to remove the `nolint`, you could get rid of `bA2` here instead and init the fields of `bA` directly:
```go
        bA.Bits = numBits
	bA.Elems = make([]uint64, (numBits+63)/64)
	for i := 0; i < numBits; i++ {
		if bits[i] == 'x' {
			bA.SetIndex(i, true)
		}
	}
```

melekes: no, I think it's fine to leave nolint

p2p/conn/secret_connection_test.go
melekes: https://github.com/tendermint/tendermint/pull/3292/commits/b38cbae53df73ac9b712d56431968931baf14e7c

PR #3294, opened 2019-02-11
R4R: make gosec linter pass

Liamsi:
ref #3262 

* [ ] Wrote tests


libs/db/remotedb/remotedb_test.go
```diff
@@ -14,7 +14,7 @@ import (
 func TestRemoteDB(t *testing.T) {
 	cert := "test.crt"
 	key := "test.key"
-	ln, err := net.Listen("tcp", "0.0.0.0:0")
+	ln, err := net.Listen("tcp", "localhost:0")
```
Liamsi: do not bind to all network interfaces (but only to the loopback interface) for tests

For this to work, I had to regenerate server key and cert; see commit (message): 9baec55fd9bb8f32d7d7b50de37e095b49dfb9ec

tools/tm-bench/transacter.go
```diff
@@ -1,7 +1,7 @@
 package main
 
 import (
-	"crypto/md5"
+	"crypto/sha256"
 	"encoding/binary"
 	"encoding/hex"
 	"encoding/json"
```
melekes: forgot to add

```
// it is ok here to use math/rand; we do not need cryptographically secure random
// number generator here and test run a bit faster
```

here too

Liamsi: Thanks! Added and slightly changed to 
```
// it is ok to use math/rand here: we do not need a cryptographically secure random
// number generator here and we can run the tests a bit faster
```
in all places

PR #3298, opened 2019-02-12
types: CommitVotes struct as last step towards #1648

ebuchman:
I believe this is the last commit necessary for #1648 before we make the actual breaking change.

That change will add `Height` and `Round` fields directly in the `Commit` and change the `CommitSig` to be a struct with just `Timestamp` and `Signature`.

Wahoo!

* [ ] Wrote tests


melekes:
```
types/block_test.go:211:48: invalid operation: commit (variable of type *Commit) has no field or method GetByIndex (typecheck)
	assert.Equal(t, voteSet.GetByIndex(0), commit.GetByIndex(0))
```

ebuchman:
Wow, integration test caught a bug. We should write a unit test to catch it too - basically need to be able to validate commits where some of the precommits aren't for the committed BlockID; up to 1/3 of them  could be for nil or even for a different blockID!

consensus/reactor.go
```diff
@@ -639,7 +639,15 @@ OUTER_LOOP:
 			// Load the block commit for prs.Height,
 			// which contains precommit signatures for prs.Height.
 			commit := conR.conS.blockStore.LoadBlockCommit(prs.Height)
-			if ps.PickSendVote(commit) {
+			valSet, err := conR.conS.blockExec.LoadValidators(prs.Height)
+			if err != nil {
+				// This should never happen as we should have the validator
+				// set for this height by now. In any case, don't panic.
```
melekes: what would be the consequences of that (not panicking)?

ebuchman: We might keep trying to send the peer a commit from a height where we don't have the validator set until the peer moves on (perhaps by receiving the commit votes from someone else). 

Though maybe we should sleep here before continue to avoid a hot loop.

melekes:
What's left here? Adding a sleep https://github.com/tendermint/tendermint/pull/3298#discussion_r258466966?

ebuchman:
> What's left here? 

1. Write tests for the current code (not the code in this PR) for the case where the block includes commits for nil or for the wrong block ID (this is supposed to be allowed, as long as +2/3 of the votes are for the right block ID) - #3484
2. Answer #3485 and change the block protocol accordingly. This should cause tests from (1) to fail, which should thus be updated.
3. Update this PR according to the outcome of (2)
4. Update the [ADR-25](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-025-commit.md) as necessary.

Of course updating the ADR should probably come first, and we probably actually need a new/better process altogether for making breaking changes to the block protocol (as opposed to architectural changes, which is what ADRs are for), but that's a separate issue.

If #3485 decides we do need to include nil votes (presumably we won't need votes for the explicitly wrong blockID), then we need to determine how to track whether a vote is for the blockID or for nil. I see roughly three approaches:

1. Don't track it, and when we verify the signature, if it fails for the right blockID then try verifying it against a nil vote. This requires no changes but is kind of hacky.

2.  Include a bool in the CommitSig, true if the vote is for the blockID, false if it's for nil

3. Add an array to the Commit that is a list of validator indices that signed nil. If all votes are for the blockID, this would be empty

ebuchman:
Oh whoops, we don't actually need to solve all that before merging this. We can move forward with this PR and deal with the final changes in a new one :D

ebuchman:
> What's left here? Adding a sleep #3298 (comment)?

So yes,  that's probably the only thing left here!

Thanks @melekes

consensus/reactor.go
ebuchman: So I'm not sure exactly what to do here anymore. Maybe we should panic. Maybe we can eliminate ever even having an error here. Don't think it needs to block the PR. See https://github.com/tendermint/tendermint/issues/3596

ebuchman:
Ok based on https://github.com/tendermint/tendermint/issues/3596 I think we want the CommitSig to have the ValidatorAddress, which simplifies this PR quite a bit.

types/block.go
```diff
@@ -528,11 +530,55 @@ func NewCommit(blockID BlockID, precommits []*CommitSig) *Commit {
 	}
 }
 
+// Construct a VoteSet from the Commit and validator set. Panics
+// if precommits from the commit can't be added to the voteset.
+func CommitToVoteSet(chainID string, commit *Commit, vals *ValidatorSet) *VoteSet {
```
Liamsi: These methods probably deserve tests to not decrease our test-coverage (?)

docs/architecture/adr-025-commit.md
```diff
@@ -26,50 +30,66 @@ References:
...
-## Consequences
+// indicate which BlockID the signature is for
+type BlockIDFlag int
 
-### Positive
-The size of a `Commit` transmitted over the network goes from:
+const (
+	BlockIDFlagAbsent BlockIDFlag = iota // vote is not included in the Commit.Precommits
+	BlockIDFlagCommit                    // voted for the Commit.BlockID
+	BlockIDFlagNil                       // voted for nil
```
Liamsi: 👍

docs/architecture/adr-025-commit.md
```diff
@@ -26,50 +30,66 @@ References:
...
+)
 
-|BlockID| + n * (|Address| + |ValidatorIndex| + |Height| + |Round| + |Timestamp| + |Type| + |BlockID| + |Signature|)
+```
 
-to:
+Note the need for an extra byte to indicate whether the signature is for the BlockID or for nil.
+This byte can also be used to indicate an absent vote, rather than using a nil object like we currently do,
+which has been [problematic for Amino](https://github.com/tendermint/go-amino/issues/260).
```
Liamsi: It rather is problematic for protobuf compatibility (but I think the linked issue is clear about this).

types/block.go
```diff
@@ -500,6 +500,8 @@ func (cs *CommitSig) toVote() *Vote {
 	return &v
 }
 
+//-------------------------------------
```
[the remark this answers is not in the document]
ebuchman: https://github.com/tendermint/tendermint/issues/3619

types/block.go
ebuchman: Yes you're right.

ebuchman:
Ok added a test and fixed the note about amino

Issue #3043, opened 2018-12-18
consensus: sync WAL more frequently


ebuchman:
Currently we only fsync the WAL to disk after our own votes and at the end of a block. This means if we crash or restart before we vote, or if we're not a validator, we won't persist any data during rounds.

We should:
- fsync the WAL on shutdown
- fsync the WAL periodically, eg. every 2s

ebuchman:
Done in #3300.

PR #3300, opened 2019-02-12
[R4R] consensus: sync WAL more frequently

thanethomson:
As per #3043, this adds a ticker to sync the WAL every 2s while the WAL is running.

* [x] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -21,3 +21,5 @@ Special thanks to external contributors on this release:
 ### IMPROVEMENTS:
 
 ### BUG FIXES:
+- [consensus] \#3297 Flush WAL on stop to prevent data corruption during
+  graceful shutdown
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -20,10 +20,13 @@ Special thanks to external contributors on this release:
 
 ### IMPROVEMENTS:
 
+- [config] \#3291 Make config.ResetTestRootWithChainID() create concurrency-safe test directories.
+
 ### BUG FIXES:
 
 * [consensus] \#3297 Flush WAL on stop to prevent data corruption during
   graceful shutdown
+- [consensus] \#3310 Reset TriggeredTimeoutPrecommit before starting next height
 - [rpc] \#3251 Fix /net_info#peers#remote_ip format. New format spec:
   * dotted decimal ("192.0.2.1"), if ip is an IPv4 or IP4-mapped IPv6 address
   * IPv6 ("2001:db8::1"), if ip is a valid IPv6 address
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -26,9 +26,11 @@ Special thanks to external contributors on this release:
 
 * [consensus] \#3297 Flush WAL on stop to prevent data corruption during
   graceful shutdown
-- [consensus] \#3310 Reset TriggeredTimeoutPrecommit before starting next height
+- [consensus] \#3302 Reset TriggeredTimeoutPrecommit before starting next
+  height
 - [rpc] \#3251 Fix /net_info#peers#remote_ip format. New format spec:
   * dotted decimal ("192.0.2.1"), if ip is an IPv4 or IP4-mapped IPv6 address
   * IPv6 ("2001:db8::1"), if ip is a valid IPv6 address
-* [cmd] \#3314 Return an
-  error on `show_validator` when the private validator file does not exist
+* [cmd] \#3314 Return an error on `show_validator` when the private validator
+  file does not exist
+* [p2p] \#3321 Authenticate a peer against its NetAddress.ID while dialing 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -11,6 +11,7 @@ Special thanks to external contributors on this release:
 * Apps
 
 * Go API
+- [consensus] \#3300 WAL interface now requires a `Flush` method
 
 * Blockchain Protocol
 
@@ -23,3 +24,5 @@ Special thanks to external contributors on this release:
 ### BUG FIXES:
 - [consensus] \#3297 Flush WAL on stop to prevent data corruption during
   graceful shutdown
+- [consensus] \#3300 Flush WAL periodically and prior to signing votes/proposals
+  to help prevent data corruption
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -11,7 +11,6 @@ Special thanks to external contributors on this release:
 * Apps
 
 * Go API
-- [consensus] \#3300 WAL interface now requires a `Flush` method
 
 * Blockchain Protocol
 
@@ -22,7 +21,11 @@ Special thanks to external contributors on this release:
 ### IMPROVEMENTS:
 
 ### BUG FIXES:
-- [consensus] \#3297 Flush WAL on stop to prevent data corruption during
+
+* [consensus] \#3297 Flush WAL on stop to prevent data corruption during
   graceful shutdown
-- [consensus] \#3300 Flush WAL periodically and prior to signing votes/proposals
-  to help prevent data corruption
+- [rpc] \#3251 Fix /net_info#peers#remote_ip format. New format spec:
+  * dotted decimal ("192.0.2.1"), if ip is an IPv4 or IP4-mapped IPv6 address
+  * IPv6 ("2001:db8::1"), if ip is a valid IPv6 address
+* [cmd] \#3314 Return an
+  error on `show_validator` when the private validator file does not exist
--- a/consensus/replay_test.go
+++ b/consensus/replay_test.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 	"runtime"
 	"testing"
 	"time"
@@ -30,10 +31,19 @@ import (
 	"github.com/tendermint/tendermint/types"
 )
 
-var consensusReplayConfig *cfg.Config
-
-func init() {
+func TestMain(m *testing.M) {
+	config = ResetConfig("consensus_reactor_test")
 	consensusReplayConfig = ResetConfig("consensus_replay_test")
+	configStateTest := ResetConfig("consensus_state_test")
+	configMempoolTest := ResetConfig("consensus_mempool_test")
+	configByzantineTest := ResetConfig("consensus_byzantine_test")
+	code := m.Run()
+	os.RemoveAll(config.RootDir)
+	os.RemoveAll(consensusReplayConfig.RootDir)
+	os.RemoveAll(configStateTest.RootDir)
+	os.RemoveAll(configMempoolTest.RootDir)
+	os.RemoveAll(configByzantineTest.RootDir)
+	os.Exit(code)
 }
 
 // These tests ensure we can always recover from failure at any part of the consensus process.
@@ -51,7 +61,8 @@ func init() {
 // and which ones we need the wal for - then we'd also be able to only flush the
 // wal writer when we need to, instead of with every message.
 
-func startNewConsensusStateAndWaitForBlock(t *testing.T, lastBlockHeight int64, blockDB dbm.DB, stateDB dbm.DB) {
+func startNewConsensusStateAndWaitForBlock(t *testing.T, consensusReplayConfig *cfg.Config,
+	lastBlockHeight int64, blockDB dbm.DB, stateDB dbm.DB) {
 	logger := log.TestingLogger()
 	state, _ := sm.LoadStateFromDBOrGenesisFile(stateDB, consensusReplayConfig.GenesisFile())
 	privValidator := loadPrivValidator(consensusReplayConfig)
@@ -59,7 +70,6 @@ func startNewConsensusStateAndWaitForBlock(t *testing.T, lastBlockHeight int64, 
 	cs.SetLogger(logger)
 
 	bytes, _ := ioutil.ReadFile(cs.config.WalFile())
-	// fmt.Printf("====== WAL: \n\r%s\n", bytes)
 	t.Logf("====== WAL: \n\r%X\n", bytes)
 
 	err := cs.Start()
@@ -110,21 +120,22 @@ func TestWALCrash(t *testing.T) {
 			3},
 	}
 
-	for _, tc := range testCases {
+	for i, tc := range testCases {
+		consensusReplayConfig := ResetConfig(fmt.Sprintf("%s_%d", t.Name(), i))
 		t.Run(tc.name, func(t *testing.T) {
-			crashWALandCheckLiveness(t, tc.initFn, tc.heightToStop)
+			crashWALandCheckLiveness(t, consensusReplayConfig, tc.initFn, tc.heightToStop)
 		})
 	}
 }
 
-func crashWALandCheckLiveness(t *testing.T, initFn func(dbm.DB, *ConsensusState, context.Context), heightToStop int64) {
+func crashWALandCheckLiveness(t *testing.T, consensusReplayConfig *cfg.Config,
+	initFn func(dbm.DB, *ConsensusState, context.Context), heightToStop int64) {
 	walPanicked := make(chan error)
 	crashingWal := &crashingWAL{panicCh: walPanicked, heightToStop: heightToStop}
 
 	i := 1
 LOOP:
 	for {
-		// fmt.Printf("====== LOOP %d\n", i)
 		t.Logf("====== LOOP %d\n", i)
 
 		// create consensus state from a clean slate
@@ -142,6 +153,7 @@ LOOP:
 
 		// clean up WAL file from the previous iteration
 		walFile := cs.config.WalFile()
+		ensureDir(filepath.Dir(walFile), 0700)
 		os.Remove(walFile)
 
 		// set crashing WAL
@@ -163,7 +175,7 @@ LOOP:
 			t.Logf("WAL panicked: %v", err)
 
 			// make sure we can make blocks after a crash
-			startNewConsensusStateAndWaitForBlock(t, cs.Height, blockDB, stateDB)
+			startNewConsensusStateAndWaitForBlock(t, consensusReplayConfig, cs.Height, blockDB, stateDB)
 
 			// stop consensus state and transactions sender (initFn)
 			cs.Stop()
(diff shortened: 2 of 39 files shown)


consensus/wal.go
```diff
@@ -72,6 +75,9 @@ type baseWAL struct {
 	group *auto.Group
 
 	enc *WALEncoder
+
+	syncTicker *time.Ticker
+	testChan   chan string
```
melekes: I'd prefer not to leak test concerns into implementation. If you want to test processSyncTicks function, you can create processSyncTick func and call it **OR** lower the interval in tests and assert the result after timeout.

```
func processSyncTicks() {
  for {
		select {
		case <-wal.syncTicker.C:
                  processSyncTick()
..
}
  }
}
```

```
wal := NewWAL(syncInterval: 10ms)
wal.Start()

time.Sleep(11ms)

assert.Synced(t, wal)
```

consensus/wal.go
```diff
@@ -109,13 +116,35 @@ func (wal *baseWAL) OnStart() error {
...
+// processSyncTicks allows us to periodically attempt to sync the WAL to disk.
+func (wal *baseWAL) processSyncTicks() {
+	for {
+		select {
+		case <-wal.syncTicker.C:
+			if err := wal.group.Flush(); err != nil {
+				wal.Logger.Error("Failed to flush WAL", "err", err)
+				wal.testChan <- "ERROR"
+			} else {
+				wal.Logger.Debug("Successfully sync'd WAL to disk")
```
melekes: not sure we need this

consensus/wal.go
thanethomson: This is a temporary measure until I implement that `testBus` we were talking about. I'd prefer not to add more `time.Sleep()` calls to the tests 😁

Can I just add a `TODO` note here for now rather and then remove this once I've got the `testBus` implementation started (in a separate PR)?

consensus/wal.go
thanethomson: I need some way of checking whether or not the sync happened successfully in the test, otherwise I'm just checking if the ticker actually ticked.

consensus/wal.go
melekes: Let's run the `testBus` idea against @ebuchman first before moving on with it.

thanethomson: 👍

consensus/wal.go
thanethomson: > not sure we need this

Oh, do you mean the `wal.Logger.Debug()` call?

melekes: yes

CHANGELOG_PENDING.md
```diff
@@ -11,6 +11,7 @@ Special thanks to external contributors on this release:
 * Apps
 
 * Go API
+- [consensus] \#3300 WAL interface now requires a `Flush` method
```
melekes: I think we don't consider this as a breaking change

consensus/state.go
```diff
@@ -1673,6 +1676,9 @@ func (cs *ConsensusState) addVote(vote *types.Vote, peerID p2p.ID) (added bool,
 }
 
 func (cs *ConsensusState) signVote(type_ types.SignedMsgType, hash []byte, header types.PartSetHeader) (*types.Vote, error) {
+	// Make sure we flush the WAL to disk.
```
```suggestion
	// Flush the WAL. Otherwise, we may not recompute the same vote to sign, and the privValidator will refuse to sign anything.
```

consensus/state.go
```diff
@@ -908,6 +908,9 @@ func (cs *ConsensusState) defaultDecideProposal(height int64, round int) {
 		}
 	}
 
+	// First flush the WAL to disk
```
```suggestion
	// Flush the WAL. Otherwise, we may not recompute the same proposal to sign, and the privValidator will refuse to sign anything.
```

consensus/wal.go
```diff
@@ -21,6 +21,9 @@ import (
 const (
 	// must be greater than types.BlockPartSizeBytes + a few bytes
 	maxMsgSizeBytes = 1024 * 1024 // 1MB
+
+	// how often the WAL should be sync'd during period sync'ing
+	walDefaultFlushInterval = time.Duration(2) * time.Second
```
melekes:
```suggestion
	walDefaultFlushInterval = 2 * time.Second
```

consensus/wal.go
```diff
@@ -72,6 +76,10 @@ type baseWAL struct {
 	group *auto.Group
 
 	enc *WALEncoder
+
+	flushTicker     *time.Ticker
+	flushInterval   time.Duration
+	periodicFlushFn func(*baseWAL) error // Allows for dependency injection.
```
melekes: Is there another way to check if the WAL was flushed? Maybe we can simply read from it and ensure the msg is there?

thanethomson: The ideal way I'd like to check if the WAL was flushed, if I have to avoid introducing test-related state into the data structures, would be to:

1. Create a fresh new WAL
2. Start the WAL with the periodic flush interval set to `walTestFlushInterval`
3. Generate some data in the WAL (5 blocks)
4. Ensure that nothing's been committed yet
5. Wait for `walTestFlushInterval` milliseconds to allow for the flush to happen
6. Ensure that the blocks we generated were committed

I can't find a simple way, given the existing code, to get step 4 right without introducing additional state somewhere. When I test for block height 4 immediately after the call to `consensus/wal_generator.go#WALGenerateNBlocks` (i.e. at step 4, by calling `wal.SearchForEndHeight`) it returns successfully, which is where I'd want it to fail.

Any suggestions on how I'd accomplish that with the existing code?

I've just committed the code I'd ideally like to use to test this, but it doesn't work:

```
--- FAIL: TestWALPeriodicSync (0.34s)
        Error Trace:    wal_test.go:171
        Error:      	An error is expected but got nil.
        Test:       	TestWALPeriodicSync
        Messages:   	expected to get an error for height 4
FAIL
coverage: 31.7% of statements
FAIL	github.com/tendermint/tendermint/consensus	1.189s
```

consensus/wal_test.go
```diff
@@ -141,6 +145,42 @@ func TestWALSearchForEndHeight(t *testing.T) {
...
+	require.NoError(t, err)
+
+	h := int64(4)
+	gr, found, err := wal.SearchForEndHeight(h, &WALSearchOptions{})
+	assert.Error(t, err, fmt.Sprintf("expected to get an error for height %d", h))
+
+	time.Sleep(walTestFlushInterval + (10 * time.Millisecond))
+
+	gr, found, err = wal.SearchForEndHeight(h, &WALSearchOptions{})
+	assert.NoError(t, err, fmt.Sprintf("expected not to err on height %d", h))
```
melekes:
```suggestion
	assert.NoError(t, err, "height %d should be in the WAL after it's flushed", h)
```

consensus/wal_test.go
```diff
@@ -141,6 +145,42 @@ func TestWALSearchForEndHeight(t *testing.T) {
...
+	assert.True(t, found, fmt.Sprintf("expected to find end height for %d", h))
```
melekes:
```suggestion
	assert.True(t, found)
```

consensus/wal_test.go
```diff
@@ -141,6 +145,42 @@ func TestWALSearchForEndHeight(t *testing.T) {
...
+	assert.NotNil(t, gr, "expected group not to be nil")
```
melekes:
```suggestion
	assert.NotNil(t, gr)
```

assert produces exactly the same message by default

consensus/wal_test.go
```diff
@@ -141,6 +145,42 @@ func TestWALSearchForEndHeight(t *testing.T) {
...
+		wal.Wait()
+	}()
+
+	err = WALGenerateNBlocks(wal.Group(), 5)
+	require.NoError(t, err)
+
+	h := int64(4)
+	gr, found, err := wal.SearchForEndHeight(h, &WALSearchOptions{})
+	assert.Error(t, err, fmt.Sprintf("expected to get an error for height %d", h))
+
```
melekes: can we add 

```
	if gr != nil {
		 gr.Close()
	}
```

here too

consensus/wal_test.go
```diff
@@ -141,6 +145,42 @@ func TestWALSearchForEndHeight(t *testing.T) {
...
+	time.Sleep(walTestFlushInterval + (10 * time.Millisecond))
+
+	gr, found, err = wal.SearchForEndHeight(h, &WALSearchOptions{})
+	assert.NoError(t, err, fmt.Sprintf("expected not to err on height %d", h))
+	assert.True(t, found, fmt.Sprintf("expected to find end height for %d", h))
+	assert.NotNil(t, gr, "expected group not to be nil")
+	if gr != nil {
+		defer gr.Close()
```
melekes:
```suggestion
		defer gr.Close()
```

no need for defer

consensus/wal.go
```diff
@@ -72,6 +76,9 @@ type baseWAL struct {
 	group *auto.Group
 
 	enc *WALEncoder
+
+	flushTicker   *time.Ticker
+	flushInterval time.Duration
 }
 
 func NewWAL(walFile string, groupOptions ...func(*auto.Group)) (*baseWAL, error) {
```
melekes: We should add a godoc comment saying that we sync wal every 2s. Otherwise it's not possible to know it's doing that

melekes:
```
I[2019-02-19|14:51:42.153] enterPropose: Our turn to propose            wal_generator=wal_generator height=6 round=0 proposer=A3258DCBF45DCA0DF052981870F2D1441A36D145 privValidator="PrivValidator{A3258DCBF45DCA0DF052981870F2D1441A36D145 LH:5, LR:0, LS:3}"
D[2019-02-19|14:51:42.157] WAL already stopped. Not writing message     wal_generator=wal_generator msg="{Height:6 Round:0 Step:RoundStepPropose RoundState:RoundState{\n  H:6 R:0 S:RoundStepPropose\n  StartTime:     2019-02-19 14:51:42.151546295 +0000 UTC\n  CommitTime:    2019-02-19 14:51:42.141546295 +0000 UTC\n  Validators:    ValidatorSet{\n    Proposer: Validator{A3258DCBF45DCA0DF052981870F2D1441A36D145 PubKeyEd25519{013FFE69A2F5781D38EFB32E77D24C9BC4A1F0122F39894F182F4A90852615A1} VP:10 A:0}\n    Validators:\n      Validator{A3258DCBF45DCA0DF052981870F2D1441A36D145 PubKeyEd25519{013FFE69A2F5781D38EFB32E77D24C9BC4A1F0122F39894F182F4A90852615A1} VP:10 A:0}\n  }\n  Proposal:      <nil>\n  ProposalBlock: nil-PartSet nil-Block\n  LockedRound:   -1\n  LockedBlock:   nil-PartSet nil-Block\n  ValidRound:   -1\n  ValidBlock:   nil-PartSet nil-Block\n  Votes:         HeightVoteSet{H:6 R:0~1\n    VoteSet{H:6 R:0 T:1 +2/3:<nil>(0) BA{1:_} map[]}\n    VoteSet{H:6 R:0 T:2 +2/3:<nil>(0) BA{1:_} map[]}\n    VoteSet{H:6 R:1 T:1 +2/3:<nil>(0) BA{1:_} map[]}\n    VoteSet{H:6 R:1 T:2 +2/3:<nil>(0) BA{1:_} map[]}\n  }\n  LastCommit:    VoteSet{H:5 R:0 T:2 +2/3:2FD1005A850A822E79C94915C0C0254BD1DE9EE7803F30F66C5C0E6B3AB89A55:1:2BE9973A5A31(1) BA{1:x} map[]}\n  LastValidators:ValidatorSet{\n    Proposer: Validator{A3258DCBF45DCA0DF052981870F2D1441A36D145 PubKeyEd25519{013FFE69A2F5781D38EFB32E77D24C9BC4A1F0122F39894F182F4A90852615A1} VP:10 A:0}\n    Validators:\n      Validator{A3258DCBF45DCA0DF052981870F2D1441A36D145 PubKeyEd25519{013FFE69A2F5781D38EFB32E77D24C9BC4A1F0122F39894F182F4A90852615A1} VP:10 A:0}\n  }\n}}"
E[2019-02-19|14:51:42.158] CONSENSUS FAILURE!!!                         wal_generator=wal_generator err="open /tmp/-TestWALPeriodicSync_237043158/data/write-file-atomic-03866980439807976044: no such file or directory" stack="goroutine 39226 [running]:\nruntime/debug.Stack(0x1990294, 0x1, 0xc00000000f)\n\t/usr/local/go/src/runtime/debug/stack.go:24 +0xb5\ngithub.com/tendermint/tendermint/consensus.(*ConsensusState).receiveRoutine.func2(0xc016ab6700, 0x1207a80)\n\t/go/src/github.com/tendermint/tendermint/consensus/state.go:611 +0x9a\npanic(0xfce080, 0xc009c9c4e0)\n\t/usr/local/go/src/runtime/panic.go:513 +0x1b9\ngithub.com/tendermint/tendermint/privval.(*FilePVLastSignState).Save(0xc01813a048)\n\t/go/src/github.com/tendermint/tendermint/privval/file.go:126 +0x1a5\ngithub.com/tendermint/tendermint/privval.(*FilePV).saveSigned(0xc01813a000, 0x6, 0x0, 0x1, 0xc0184e7700, 0x7d, 0xfc, 0xc018567ac0, 0x40, 0x40)\n\t/go/src/github.com/tendermint/tendermint/privval/file.go:372 +0x10f\ngithub.com/tendermint/tendermint/privval.(*FilePV).signProposal(0xc01813a000, 0xc008aad390, 0xf, 0xc0016defc0, 0xd6cded, 0xc0016df018)\n\t/go/src/github.com/tendermint/tendermint/privval/file.go:358 +0x588\ngithub.com/tendermint/tendermint/privval.(*FilePV).SignProposal(0xc01813a000, 0xc008aad390, 0xf, 0xc0016defc0, 0x20, 0x20)\n\t/go/src/github.com/tendermint/tendermint/privval/file.go:248 +0x64\ngithub.com/tendermint/tendermint/consensus.(*ConsensusState).defaultDecideProposal(0xc016ab6700, 0x6, 0x0)\n\t/go/src/github.com/tendermint/tendermint/consensus/state.go:918 +0x353\ngithub.com/tendermint/tendermint/consensus.(*ConsensusState).defaultDecideProposal-fm(0x6, 0x0)\n\t/go/src/github.com/tendermint/tendermint/consensus/state.go:168 +0x56\ngithub.com/tendermint/tendermint/consensus.(*ConsensusState).enterPropose(0xc016ab6700, 0x6, 0x0)\n\t/go/src/github.com/tendermint/tendermint/consensus/state.go:886 +0xbe3\ngithub.com/tendermint/tendermint/consensus.(*ConsensusState).enterNewRound(0xc016ab6700, 0x6, 0x0)\n\t/go/src/github.com/tendermint/tendermint/consensus/state.go:827 +0xc7a\ngithub.com/tendermint/tendermint/consensus.(*ConsensusState).addVote(0xc016ab6700, 0xc01813bc20, 0x0, 0x0, 0xc0075c8900, 0xc0099eb240, 0xc0075c8900)\n\t/go/src/github.com/tendermint/tendermint/consensus/state.go:1662 +0x1437\ngithub.com/tendermint/tendermint/consensus.(*ConsensusState).tryAddVote(0xc016ab6700, 0xc01813bc20, 0x0, 0x0, 0x5, 0xc018487420, 0x0)\n\t/go/src/github.com/tendermint/tendermint/consensus/state.go:1506 +0x90\ngithub.com/tendermint/tendermint/consensus.(*ConsensusState).handleMsg(0xc016ab6700, 0x12f48a0, 0xc00000f900, 0x0, 0x0)\n\t/go/src/github.com/tendermint/tendermint/consensus/state.go:694 +0xa1f\ngithub.com/tendermint/tendermint/consensus.(*ConsensusState).receiveRoutine(0xc016ab6700, 0x0)\n\t/go/src/github.com/tendermint/tendermint/consensus/state.go:655 +0x7de\ncreated by github.com/tendermint/tendermint/consensus.(*ConsensusState).OnStart\n\t/go/src/github.com/tendermint/tendermint/consensus/state.go:334 +0x489\n"
I[2019-02-19|14:51:42.259] Searching for height                         height=4 min=0 max=0
I[2019-02-19|14:51:42.261] Found                                        height=4 index=0
I[2019-02-19|14:51:42.261] Stopping baseWAL                             impl=baseWAL
I[2019-02-19|14:51:42.261] Stopping Group                               impl=Group
--- FAIL: TestWALPeriodicSync (0.32s)
    <autogenerated>:1: ^M                          ^M   Error Trace:    wal_test.go:169
        ^M      Error:          Should not be zero, but was 0
        ^M      Test:           TestWALPeriodicSync
```

consensus/replay_test.go
```diff
@@ -83,7 +83,7 @@ func startNewConsensusStateAndWaitForBlock(t *testing.T, consensusReplayConfig *
 	require.NoError(t, err)
 	select {
 	case <-newBlockCh:
-	case <-time.After(60 * time.Second):
+	case <-time.After(120 * time.Second):
```
melekes: why increase?

consensus/wal.go
```diff
@@ -72,8 +76,14 @@ type baseWAL struct {
 	group *auto.Group
 
 	enc *WALEncoder
+
+	flushTicker   *time.Ticker
+	flushInterval time.Duration
 }
 
+// NewWAL attempts to create a new write-ahead logger based on `baseWAL`, which
```
melekes:
```suggestion
// NewWAL returns a new write-ahead logger based on `baseWAL`, which implements WAL.
// It's flushed and synced to disk every 2s and once when being stopped.
```

consensus/wal.go
```diff
@@ -109,14 +125,37 @@ func (wal *baseWAL) OnStart() error {
 		wal.WriteSync(EndHeightMessage{0})
 	}
 	err = wal.group.Start()
+	wal.flushTicker = time.NewTicker(wal.flushInterval)
+	go wal.processFlushTicks()
 	return err
 }
 
+// processFlushTicks allows us to periodically attempt to sync the WAL to disk.
```
melekes:
```suggestion
```

comment does not bring any value

consensus/wal.go
```diff
@@ -109,14 +125,37 @@ func (wal *baseWAL) OnStart() error {
...
+func (wal *baseWAL) processFlushTicks() {
+	for {
+		select {
+		case <-wal.flushTicker.C:
+			err := wal.Flush()
```
melekes:
```
if err := wal.Flush(); err != nil {
 ...
}
```

is shorter

consensus/state.go
```diff
@@ -909,6 +909,9 @@ func (cs *ConsensusState) defaultDecideProposal(height int64, round int) {
 		}
 	}
 
+	// Flush the WAL. Otherwise, we may not recompute the same proposal to sign, and the privValidator will refuse to sign anything.
```
ebuchman: should this happen in here or right before we call `decideProposal` ? If we keep it as is, it means any other implementations of the `decideProposal` functionality besides this default need to also remember to flush the WAL first. Might not actually be an issue though.

consensus/wal_generator.go
```diff
@@ -24,13 +23,12 @@ import (
...
 )
 
-// WALGenerateNBlocks generates a consensus WAL. It does this by spining up a
+// WALGenerateNBlocks generates a consensus WAL. It does this by spinning up a
 // stripped down version of node (proxy app, event bus, consensus state) with a
 // persistent kvstore application and special consensus wal instance
 // (byteBufferWAL) and waits until numBlocks are created. If the node fails to produce given numBlocks, it returns an error.
 func WALGenerateNBlocks(t *testing.T, wr io.Writer, numBlocks int) (err error) {
 	config := getConfig(t)
-	defer os.RemoveAll(config.RootDir)
```
ebuchman: Why remove this ?

melekes: it happens later. if we remove it now, tests will fail because they are reading from the WAL.

Issue #3166, opened 2019-01-19
proposer priority fixes follow up


ebuchman:
Follow up to #3049

- Clarify whether the maximum difference should be TotalVotingPower or 2*TotalVotingPower (https://github.com/tendermint/tendermint/pull/3049#discussion_r249140957)
- Consider whether we need to avg the priorities more frequently (https://github.com/tendermint/tendermint/pull/3049#discussion_r249139319)

**Update**: TODOs we (@ancazamfir, @cwgoes, @Liamsi) collected during a call:

 - [ ] use max int32 as a bound for voting power while keeping voting power as int64; also make sure that total voting power doesn't exceed int64 (to avoid and remove clipping) (SDK-side PR: https://github.com/cosmos/cosmos-sdk/pull/3400) (@Liamsi)
 - [ ] make sure the priority computations don't exceed int64, e.g. during scaling if we do vp * tvp/.. (@Liamsi)
 - [ ] figure out bound for difference (max - min priority) in n-validator case, ensure we respect it (we should not normalize more tightly) (@cwgoes, @Liamsi )
- [ ] formalize fairness proof and priority difference bounds in LaTeX doc (@cwgoes, @Liamsi )
- [x] treat validator set changes as a set and apply them atomically #3181 #3222 (@ancazamfir )
- [x] While addressing #3181 move priorities.go:updateValidators() to validator_set.go and write unit tests in validator_set_test.go (@ancazamfir)
- [ ] update spec/documentation (@ancazamfir)

__Update__ (Feb 8):

#3181 has been fixed. But the fix included a large refactor of the validator updating code that spawned some follow up requests in the reviews. Please see the reviews in #3222. Main items for follow up:

- [ ] Better error handling instead of panicing (https://github.com/tendermint/tendermint/pull/3222#issuecomment-461757136)
- [ ] Test cleanup if possible (https://github.com/tendermint/tendermint/pull/3222#discussion_r254254894)
- [ ] Cleanup `processChanges` function (https://github.com/tendermint/tendermint/pull/3222#discussion_r255124628)
- [ ] Update ABCI documentation to specify validation rules around what is an invalid update (https://github.com/tendermint/tendermint/pull/3222#discussion_r254957923)

ancazamfir:
[...]

PR #3301, opened 2019-02-12
followup after validator set changes

ancazamfir:
New ApplyRemove() function
New test for priority order
...

See https://github.com/tendermint/tendermint/issues/3166

* [ ] Wrote tests


types/validator_set.go
```diff
@@ -510,35 +488,32 @@ func verifyRemovals(deletes []*Validator, vals *ValidatorSet) error {
...
 func (vals *ValidatorSet) applyRemovals(deletes []*Validator) {
 
-	for _, valUpdate := range deletes {
-		address := valUpdate.Address
-		_, removed := vals.Remove(address)
-		if !removed {
-			// Should never happen
-			panic(fmt.Sprintf("failed to remove validator %X", address))
+	existing := make([]*Validator, len(vals.Validators))
+	copy(existing, vals.Validators)
```
melekes: once again, what's the reason for additional structs? to make the op atomic?

types/validator_set_test.go
```diff
@@ -973,55 +1011,218 @@ func TestValSetApplyUpdatesTestsExecute(t *testing.T) {
...
+			cfg.deletedVals[i-nOld-nChanged] = testVal{fmt.Sprintf("v%d", i), 0}
+		}
+	}
+
+	for i := nBase; i < nBase+nAdd; i++ {
+		cfg.addedVals[i-nBase] = testVal{fmt.Sprintf("v%d", i), int64(cmn.RandUint()%maxPower + 1)}
+		cfg.expectedVals[i-nDel] = cfg.addedVals[i-nBase]
+	}
+
+	sort.Sort(testValsByAddress(cfg.startVals))
```
melekes: you can sort in place using the sort.Slice func

```
sort.Slice(people, func(i, j int) bool { return people[i].Name < people[j].Name })
```

types/validator_set_test.go
```diff
@@ -973,55 +1011,218 @@ func TestValSetApplyUpdatesTestsExecute(t *testing.T) {
...
+		0: {
+			startVals:    []testVal{{"v1", 1}, {"v2", 2}, {"v3", 1000}},
+			deletedVals:  []testVal{{"v3", 0}},
+			updatedVals:  []testVal{{"v2", 1}},
+			addedVals:    []testVal{{"v4", 40}, {"v5", 50}},
+			expectedVals: []testVal{{"v1", 1}, {"v2", 1}, {"v4", 40}, {"v5", 50}},
+		},
+		// generate a configuration with 100 validators,
+		// randomly select validators for updates and deletes, and
+		// generate 100 new validators to be added
```
melekes: * 10 new validators to be added

types/validator_set_test.go
```diff
@@ -658,6 +651,60 @@ type testVal struct {
...
+	permList := make([]testVal, len(valList))
+	//perm := rand.Perm(len(valList))
+	perm := cmn.RandPerm(len(valList))
+	for i, v := range perm {
+		permList[v] = valList[i]
+	}
+	return permList
+}
+
+func createNewValidatorList(testValList []testVal) []*Validator {
```
melekes: why do we use testVal? can't we use `Validator` struct?

types/validator_set_test.go
```diff
@@ -973,55 +1011,218 @@ func TestValSetApplyUpdatesTestsExecute(t *testing.T) {
...
+			startVals:    []testVal{{"v1", 1}, {"v2", 2}, {"v3", 1000}},
+			deletedVals:  []testVal{{"v3", 0}},
+			updatedVals:  []testVal{{"v2", 1}},
+			addedVals:    []testVal{{"v4", 40}, {"v5", 50}},
+			expectedVals: []testVal{{"v1", 1}, {"v2", 1}, {"v4", 40}, {"v5", 50}},
+		},
+		// generate a configuration with 100 validators,
+		// randomly select validators for updates and deletes, and
+		// generate 100 new validators to be added
+		1: randTestVSetCfg(t, 100, 10),
```
melekes: would be great to have it test more combinations

types/validator_set.go
```diff
@@ -510,35 +488,32 @@ func verifyRemovals(deletes []*Validator, vals *ValidatorSet) error {
...
+
+	merged := make([]*Validator, len(existing)-len(deletes))
+	i := 0
+
+	for len(existing) > 0 && len(deletes) > 0 {
+		if bytes.Equal(existing[0].Address, deletes[0].Address) {
+			deletes = deletes[1:]
```
melekes: shifting entire array is time consuming, no?

types/validator_set_test.go
```diff
@@ -658,6 +651,60 @@ type testVal struct {
 	power int64
 }
 
+func permutation(valList []testVal) []testVal {
+	if len(valList) == 0 {
+		return nil
+	}
+	permList := make([]testVal, len(valList))
+	//perm := rand.Perm(len(valList))
```
melekes: leftover?

types/validator_set.go
ancazamfir: no good reason currently, in an earlier version I used to do some checks and wanted the vals.Validators to stay intact. I removed the copy now.

types/validator_set_test.go
ancazamfir: I find it easier to write the test values this way and avoid the casting.

types/validator_set_test.go
ancazamfir: added a few more

types/validator_set.go
ancazamfir: I think this just changes the slice header (pointer is moved right)

types/validator_set.go
```diff
@@ -510,35 +487,32 @@ func verifyRemovals(deletes []*Validator, vals *ValidatorSet) error {
 // Should not fail as verification has been done before.
 func (vals *ValidatorSet) applyRemovals(deletes []*Validator) {
 
-	for _, valUpdate := range deletes {
-		address := valUpdate.Address
-		_, removed := vals.Remove(address)
-		if !removed {
-			// Should never happen
-			panic(fmt.Sprintf("failed to remove validator %X", address))
+	existing := vals.Validators
```
melekes: why we can't just loop over vals.Validators and apply deletes in place?

types/validator_set.go
```diff
@@ -457,8 +436,7 @@ func computeNewPriorities(updates []*Validator, vals *ValidatorSet, updatedTotal
 // must have been validated with verifyUpdates() and priorities computed with computeNewPriorities().
 func (vals *ValidatorSet) applyUpdates(updates []*Validator) {
 
-	existing := make([]*Validator, len(vals.Validators))
-	copy(existing, vals.Validators)
+	existing := vals.Validators
```
melekes: same here. why not loop over vals.Validators and apply updates? then loop over updates which add new validators.

types/validator_set.go
```diff
@@ -569,22 +548,46 @@ func (vals *ValidatorSet) updateWithChangeSet(changes []*Validator, allowDeletes
...
 	// Compute the priorities for updates
 	numNewValidators := computeNewPriorities(updates, vals, updatedTotalVotingPower)
-	if len(vals.Validators)+numNewValidators <= len(deletes) {
+	if len(vals.Validators)+numNewValidators == len(deletes) {
 		err = fmt.Errorf("applying the validator changes would result in empty set")
 		return err
 	}
 
+	if len(vals.Validators)+numNewValidators < len(deletes) {
+		panic("internal error, more deletes than validators after verification")
```
melekes:
```suggestion
		panic("more deletes than validators after verification")
```

types/validator_set.go
```diff
@@ -509,36 +486,38 @@ func verifyRemovals(deletes []*Validator, vals *ValidatorSet) error {
 // Removes the validators specified in 'deletes' from validator set 'vals'.
 // Should not fail as verification has been done before.
 func (vals *ValidatorSet) applyRemovals(deletes []*Validator) {
+	if len(deletes) > len(vals.Validators) {
+		panic("There are more deletes than validators")
```
melekes:
```suggestion
		panic("more deletes than validators when applying")
```

types/validator_set.go
```diff
@@ -569,22 +548,46 @@ func (vals *ValidatorSet) updateWithChangeSet(changes []*Validator, allowDeletes
 
 	// Compute the priorities for updates
 	numNewValidators := computeNewPriorities(updates, vals, updatedTotalVotingPower)
-	if len(vals.Validators)+numNewValidators <= len(deletes) {
+	if len(vals.Validators)+numNewValidators == len(deletes) {
 		err = fmt.Errorf("applying the validator changes would result in empty set")
```
melekes:
```suggestion
		err = errors.New("applying the validator changes would result in empty set")
```

types/validator_set_test.go
```diff
@@ -658,6 +651,62 @@ type testVal struct {
...
+	// verify that the capacity and length of validators is the same
+	assert.Equal(t, len(valSet.Validators), cap(valSet.Validators))
+
+	// verify that the vals' tvp is set to the sum of the all vals voting powers
+	tvp := valSet.totalVotingPower
+	assert.Equal(t, valSet.TotalVotingPower(), tvp,
+		"expected TVP %d. Got %d, valSet=%s", tvp, valSet.totalVotingPower, valSet)
+
+	// verify that validator priorities are centered
+	l := int64(len(valSet.Validators))
```
melekes:
```suggestion
	valsCount := int64(len(valSet.Validators))
```

types/validator_set_test.go
```diff
@@ -658,6 +651,62 @@ type testVal struct {
...
+	return valSet
+}
+
+func verifyValidatorSet(t *testing.T, valSet *ValidatorSet) {
+	// verify that the capacity and length of validators is the same
+	assert.Equal(t, len(valSet.Validators), cap(valSet.Validators))
+
+	// verify that the vals' tvp is set to the sum of the all vals voting powers
+	tvp := valSet.totalVotingPower
+	assert.Equal(t, valSet.TotalVotingPower(), tvp,
```
melekes: I wonder if we actually test anything here. In order to test `is set to the sum of the all vals voting powers` we need to sum all vals powers, not use internal variable!

types/validator_set_test.go
```diff
@@ -658,6 +651,62 @@ type testVal struct {
...
+		"expected TVP %d. Got %d, valSet=%s", tvp, valSet.totalVotingPower, valSet)
```
melekes:
```suggestion
		"expected TVP %d. Got %d, valSet=%s", tvp, valSet.TotalVotingPower(),
```

types/validator_set_test.go
```diff
@@ -658,6 +651,62 @@ type testVal struct {
...
+
+	// verify that validator priorities are centered
+	l := int64(len(valSet.Validators))
+	tpp := valSet.TotalVotingPower()
```
melekes: can we move it up and use in the second assert?

types/validator_set_test.go
```diff
@@ -973,55 +1019,243 @@ func TestValSetApplyUpdatesTestsExecute(t *testing.T) {
...
+	err := valSet.UpdateWithChangeSet(valList)
+	assert.NoError(t, err)
+}
+
+func isAddressInList(address []byte, valsList []testVal) bool {
+	for len(valsList) > 0 {
+		if bytes.Equal([]byte(valsList[0].name), address) {
+			return true
+		}
+		valsList = valsList[1:]
```
melekes: can't we just loop over valsList and return as soon as we find the address? mutating the function argument is a bad code smell

types/validator_set_test.go
ancazamfir: we have to check valSet.totalVotingPower against the sum. I added a local function to check the sum since the TotalVotingPower() only computes when totalVotingPower is 0

types/validator_set_test.go
ancazamfir: There was a mistake for tpp computation, just fixed it.

types/validator_set_test.go
ancazamfir: parameter is passed by value so it's fine. But I did change it with the last commit.

types/validator_set_test.go
```diff
@@ -677,20 +677,39 @@ func createNewValidatorSet(testValList []testVal) *ValidatorSet {
...
+		"expected TVP %d. Got %d, valSet=%s", expectedTVP, tvp, valSet)
 
 	// verify that validator priorities are centered
-	l := int64(len(valSet.Validators))
-	tpp := valSet.TotalVotingPower()
-	assert.True(t, tpp <= l || tpp >= -l,
-		"expected total priority in (-%d, %d). Got %d", l, l, tpp)
+	valsCount := int64(len(valSet.Validators))
+	tpp := valSetTotalProposerPriority(valSet)
+	assert.True(t, tpp <= valsCount || tpp >= -valsCount,
```
ancazamfir: should be &&

types/validator_set.go
```diff
@@ -569,22 +549,46 @@ func (vals *ValidatorSet) updateWithChangeSet(changes []*Validator, allowDeletes
 
 	// Compute the priorities for updates
 	numNewValidators := computeNewPriorities(updates, vals, updatedTotalVotingPower)
-	if len(vals.Validators)+numNewValidators <= len(deletes) {
-		err = fmt.Errorf("applying the validator changes would result in empty set")
+	if len(vals.Validators)+numNewValidators == len(deletes) {
+		err = errors.New("applying the validator changes would result in empty set")
```
melekes: return errors.New...

types/validator_set_test.go
```diff
@@ -658,6 +651,81 @@ type testVal struct {
...
+	return valList
+}
+
+func createNewValidatorSet(testValList []testVal) *ValidatorSet {
+	valList := createNewValidatorList(testValList)
+	valSet := NewValidatorSet(valList)
+	return valSet
+}
+
+func valSetTotalVotingPower(valSet *ValidatorSet) int64 {
```
melekes: we should not duplicate domain logic in tests

ancazamfir: In general yes. But I need the sum of voting power and cannot use the TotalVotingPower() method here.

types/validator_set.go
ancazamfir: (?) fixed with https://github.com/tendermint/tendermint/pull/3301/commits/e9a297b19af466a8e7aa1482f5169b29eae82c67

ancazamfir: ah sorry you want the return

types/validator_set_test.go
melekes: Can't you test it separately and here assume it's working properly.

```
set := construct a set with 33 tvp 
assert.Equal(t, 33, set.TotalVotingPower())
```

ancazamfir: I could but here we want to make sure that the totalVotingPower is updated as part of the updateWithChangeSet(). So even if the TotalVotingPower() works correctly it doesn't help if we don't call it.
One possibility is to make a private method in validator_set.go to compute totalVotingPower all the time and have TotalVotingPower() and UpdateWithChangeSet() call that instead. I can then use this private method in testing.

melekes: 😄 therefore we've arrived to where we started because if you "use this private method in testing." you are not testing anything.

> I could but here we want to make sure that the totalVotingPower is updated as part of the updateWithChangeSet()

Can we compare two values? old != new

ancazamfir: I am testing that vals.totalVotingPower value is correct. I am not interested to test that the private method is correct, just that it has been called.

ancazamfir: > Can we compare two values? old != new
Don't have access to old here. This function checks that after processing, the validator set has some properties:
- totalVotingPower is computed
- is scaled
- is centered
Note also how I call computeMaxMinPriorityDiff(), another private method to check the centering.

types/validator_set.go
Liamsi: Yes, I think Anton means:
```suggestion
		return errors.New("applying the validator changes would result in empty set")
```

types/validator_set_test.go
```diff
@@ -658,6 +651,81 @@ type testVal struct {
...
+func createNewValidatorList(testValList []testVal) []*Validator {
+	valList := make([]*Validator, 0, len(testValList))
+	for _, val := range testValList {
+		valList = append(valList, newValidator([]byte(val.name), val.power))
+	}
+	return valList
+}
+
+func createNewValidatorSet(testValList []testVal) *ValidatorSet {
+	valList := createNewValidatorList(testValList)
```
melekes: `return NewValidatorSet(createNewValidatorList(testValList))`

types/validator_set_test.go
```diff
@@ -837,30 +905,27 @@ func TestValSetUpdatesBasicTestsExecute(t *testing.T) {
...
 		assert.NoError(t, err, "test %d", i)
 
-		// check the parameter list has not changed
-		assert.Equal(t, valList, valListCopy, "test %v", i)
+		valListCopy := validatorListCopy(valSet.Validators)
+		// check that the voting power in the set's validators is not changing if the voting power
+		// is changed in the list of validators previously passed as parameter to UpdateWithChangeSet.
+		// this is to make sure copies of the validators are made by UpdateWithChangeSet.
+		if len(valList) > 0 {
+			valList[0].VotingPower += 1
```
melekes:
```suggestion
			valList[0].VotingPower++
```

types/validator_set.go
melekes: What @Liamsi said

types/validator_set_test.go
Liamsi: > * totalVotingPower is computed

IMHO, there could be 2 ways to approach this: either make sure `valSet.totalVotingPower != 0` (this wouldn't be the case if `TotalVotingPower()` was not called; at leaset for freshly created validator sets), or, save away `valSet.totalVotingPower` in a variable and call `TotalVotingPower()` after: both values should be the same (and the latter also computes the sum you need without replicating the logic in a test).

types/validator_set.go
```diff
@@ -569,22 +557,44 @@ func (vals *ValidatorSet) updateWithChangeSet(changes []*Validator, allowDeletes
...
 	// Compute the priorities for updates
 	numNewValidators := computeNewPriorities(updates, vals, updatedTotalVotingPower)
-	if len(vals.Validators)+numNewValidators <= len(deletes) {
-		err = fmt.Errorf("applying the validator changes would result in empty set")
-		return err
+	if len(vals.Validators)+numNewValidators == len(deletes) {
+		return errors.New("applying the validator changes would result in empty set")
+	}
+
+	if len(vals.Validators)+numNewValidators < len(deletes) {
```
Liamsi: Does it make sense to include the new validators here, too? E.g., does it make sense to add new bunch of new validators and remove all those again in the same operation? Are deletes not only for the existing vals?

ancazamfir: This basically checks for negative length after applying everything but I agree that it does look weird so I removed it. We do verify earlier that deletes are only for existing validators.
There was a panic (for len(deletes)>len(vals.Validators)) in applyRemovals that I just moved it in verifyRemovals.

types/validator_set.go
```diff
@@ -117,9 +118,9 @@ func (vals *ValidatorSet) RescalePriorities(diffMax int64) {
 	// NOTE: This may make debugging priority issues easier as well.
 	diff := computeMaxMinPriorityDiff(vals)
 	ratio := (diff + diffMax - 1) / diffMax
-	if ratio > 1 {
+	if diff > diffMax {
```
melekes: this does not change the algo, right?

types/validator_set.go
```diff
@@ -361,7 +349,11 @@ func processChanges(origChanges []*Validator) (updates, removals []*Validator, e
 			return nil, nil, err
 		}
 		if valUpdate.VotingPower < 0 {
-			err = fmt.Errorf("voting power can't be negative %v", valUpdate)
+			err = fmt.Errorf("voting power can't be negative: %v", valUpdate)
+			return nil, nil, err
+		}
+		if valUpdate.VotingPower > MaxTotalVotingPower {
+			err = fmt.Errorf("voting power can't be higher than %v: %v ", MaxTotalVotingPower, valUpdate)
```
melekes: do you think we also should explain WHY in the error msg? or it will make the msg too long?

types/validator_set.go
```diff
@@ -376,19 +368,21 @@ func processChanges(origChanges []*Validator) (updates, removals []*Validator, e
...
 // updatedTotalVotingPower - the new total voting power if these updates would be applied
+// numNew - number of new validators
 // err - non-nil if the maximum allowed total voting power would be exceeded
 //
-// 'updates' should be a list of proper validator changes, i.e. they have been scanned
+// 'updates' should be a list of proper validator changes, i.e. they have been verified
 // by processChanges for duplicates and invalid values.
 // No changes are made to the validator set 'vals'.
-func verifyUpdates(updates []*Validator, vals *ValidatorSet) (updatedTotalVotingPower int64, err error) {
+func verifyUpdates(updates []*Validator, vals *ValidatorSet) (updatedTotalVotingPower int64, numNew int, err error) {
```
melekes: if you rename numNew to numNewValidators, you won't need a comment.

types/validator_set.go
```diff
@@ -376,19 +368,21 @@ func processChanges(origChanges []*Validator) (updates, removals []*Validator, e
...
 // total voting power would not be exceeded if these updates would be applied to the set.
-// It also computes the total voting power of the set that would result after the updates but
-// before the removals.
+// It also computes:
+//  - the total voting power of the set that would result after the updates but
+// before the removals,
+//  - the number of new validators
 //
 // Returns:
 // updatedTotalVotingPower - the new total voting power if these updates would be applied
```
melekes: this is clear from the function's return params, no?

types/validator_set.go
```diff
@@ -503,88 +485,109 @@ func verifyRemovals(deletes []*Validator, vals *ValidatorSet) error {
...
+// are not allowed and will trigger an error if present in 'changes'.
+// The 'allowDeletes' flag is set to false by NewValidatorSet() and to true by UpdateWithChangeSet().
 func (vals *ValidatorSet) updateWithChangeSet(changes []*Validator, allowDeletes bool) error {
 
 	if len(changes) <= 0 {
 		return nil
 	}
 
-	// Check for duplicates within changes, split in 'updates' and 'deletes' lists (sorted)
+	// check for duplicates within changes, split in 'updates' and 'deletes' lists (sorted)
```
melekes: https://github.com/tendermint/coding/blob/master/go/coding_standard.md#comments

types/validator_set.go
ancazamfir: right

types/validator_set.go
ancazamfir: I added something, let me know.

Issue #3314, opened 2019-02-15
show_validator creates directories


folex:
**Tendermint version**: 0.25.0-0c9c3292

**Environment**:
- **OS**: Darwin mac 18.2.0 Darwin Kernel Version 18.2.0; macOS 10.14.1 (18B75)
- **Install tools**: brew

**What happened**:
I run `tendermint show_validator` without running `tendermint init` previously (i.e., no tendermint directories exist on my computer).

`tendermint show_validator` creates `~/.tendermint` directory:
```
.tendermint$ tree
.
├── config
│   ├── config.toml
│   └── priv_validator.json
└── data
```
 
**What you expected to happen**:
`tendermint show_validator` should error because there's no `priv_validator.json`. That's what `show_node_id` does.

**Have you tried the latest version**: yes

**How to reproduce it**: `brew install tendermint && tendermint show_validator`

melekes:
Merged to develop. Should land in v0.30.1

Issue #2176, opened 2018-08-07
p2p: non-deterministic TestPEXReactorRunning


ebuchman:
Triggered by https://github.com/tendermint/tendermint/pull/2174 (no code changes!)

```
ok  	github.com/tendermint/tendermint/node	1.165s	coverage: 70.8% of statements
--- FAIL: TestPEXReactorRunning (10.03s)
	pex_reactor_test.go:445: expected all switches to be connected to at least 2 peer(s) (switches: 0 => {outbound: 0, inbound: 1}, 1 => {outbound: 1, inbound: 1}, 2 => {outbound: 2, inbound: 0}, )
FAIL
coverage: 74.4% of statements
FAIL	github.com/tendermint/tendermint/p2p/pex	19.736s
```

ebuchman:
Got this again. Here's the full test output: https://gist.github.com/ebuchman/1714aab96ea2bb2b465203f471589b7d

ebuchman:
Still getting this (eg. https://circleci.com/gh/tendermint/tendermint/59449#tests/containers/3)

tessr:
Going to close this since we did a bunch of cleanup on non-deterministic tests earlier this year (+ I think we are saying goodbye to the PEX soon anyways)

PR #3316, opened 2019-02-15
Return an error on `show_validator`

alexanderbez:
closes: #3314 

----

* [ ] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -21,5 +21,9 @@ Special thanks to external contributors on this release:
 ### IMPROVEMENTS:
 
 ### BUG FIXES:
-- [consensus] \#3297 Flush WAL on stop to prevent data corruption during
+
+* [consensus] \#3297 Flush WAL on stop to prevent data corruption during
   graceful shutdown
+* [cmd] [\#3314](https://github.com/tendermint/tendermint/issues/3314) Return an
+  error on `show_validator` when the private validator file does not exist.
+  
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -24,6 +24,6 @@ Special thanks to external contributors on this release:
 
 * [consensus] \#3297 Flush WAL on stop to prevent data corruption during
   graceful shutdown
-* [cmd] [\#3314](https://github.com/tendermint/tendermint/issues/3314) Return an
+* [cmd] \#3314 Return an
   error on `show_validator` when the private validator file does not exist.
   
--- a/cmd/tendermint/commands/show_validator.go
+++ b/cmd/tendermint/commands/show_validator.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	cmn "github.com/tendermint/tendermint/libs/common"
 	"github.com/tendermint/tendermint/privval"
 )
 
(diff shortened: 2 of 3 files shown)


alexanderbez:
Looks like the `p2p` package is failing but this seems to be an existing regression on the latest `develop`?

Liamsi:
There are a bunch of known non-deterministic test-failures which are known. After a quick search through the open issues, it seems this is a new one: `TestStateLockPOLSafety1`. I'll open an issue.

CHANGELOG_PENDING.md
```diff
@@ -21,5 +21,9 @@ Special thanks to external contributors on this release:
 ### IMPROVEMENTS:
 
 ### BUG FIXES:
-- [consensus] \#3297 Flush WAL on stop to prevent data corruption during
+
+* [consensus] \#3297 Flush WAL on stop to prevent data corruption during
   graceful shutdown
+* [cmd] [\#3314](https://github.com/tendermint/tendermint/issues/3314) Return an
```
Liamsi: I think we run a script that adds the links in the end. Not sure if adding a link here directly does cause any problems with that script: https://github.com/tendermint/tendermint/blob/master/scripts/linkify_changelog.py

```suggestion
* [cmd] \#3314 Return an
```

cmd/tendermint/commands/show_validator.go
```diff
@@ -5,18 +5,29 @@ import (
...
+	cmn "github.com/tendermint/tendermint/libs/common"
 	"github.com/tendermint/tendermint/privval"
 )
 
 // ShowValidatorCmd adds capabilities for showing the validator info.
 var ShowValidatorCmd = &cobra.Command{
 	Use:   "show_validator",
 	Short: "Show this node's validator info",
-	Run:   showValidator,
+	RunE:  showValidator,
```
Liamsi: Interesting, this aren't named consistently across commands. Did never notice that.

cmd/tendermint/commands/show_validator.go
```diff
@@ -5,18 +5,29 @@ import (
...
+func showValidator(cmd *cobra.Command, args []string) error {
+	keyFilePath := config.PrivValidatorKeyFile()
+	if !cmn.FileExists(keyFilePath) {
+		return fmt.Errorf("private validator file %s does not exist", keyFilePath)
+	}
+
+	pv := privval.LoadFilePV(keyFilePath, config.PrivValidatorStateFile())
+	bz, err := cdc.MarshalJSON(pv.GetPubKey())
+	if err != nil {
+		return err
```
Liamsi: Do you think we should wrap this with something the user understands? Not sure if whatever amino would return is helpful for the caller.

Liamsi:
After restarting `TestPEXReactorRunning` fails. This one is known: https://github.com/tendermint/tendermint/issues/2176

CHANGELOG_PENDING.md
alexanderbez: Gotcha. Amended. I saw a mix case usage in the changelog.

alexanderbez:
Interesting @Liamsi -- I'd be happy to help dig into and potentially fix some of the non-deterministic tests.

melekes:
@alexanderbez here is the list https://github.com/tendermint/tendermint/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+non-deterministic

cmd/tendermint/commands/show_validator.go
melekes: Will do

melekes:

Issue #3285, opened 2019-02-08
expose proposer selection to apps


ebuchman:
We have recently done a lot of work on getting the proposer selection algorithm right in tendermint. See for instance #3049, #3140, #3222. There are still some open issues and a new `proposer-selection` label to tag them.

One thing we have discussed is exposing the proposer selection to the applications. Currently, the only control apps have is to change the validator set, but otherwise all selection is left to Tendermint. 

We should consider adding a new field to ResponseEndBlock which gives applications more control over proposer selection. This could be a list of {PubKey, Priority}, for instance, that would otherwise use the same Tendermint algorithm, but where the app specifies with every block who can be a proposer for the next block and what their respective initial priorities are.

This will need more design work and an ADR, and we need to specify when and how the default behaviour kicks in. But otherwise could be quite valuable to applications, and would for instance allow them to implement randomness in their proposer selection before we do it in Tendermint (#763), and would allow for proposers who aren't even in the validator set, which might be useful for certain configurations.

PR #3324, opened 2019-02-18
[WIP] [ADR] ADR-36: Giving the app control over proposer selection

b00f:
Adr-036 Initial commit

### DIFF
--- a/docs/architecture/adr-036-validator-set-by-app.md
+++ b/docs/architecture/adr-036-validator-set-by-app.md
@@ -1,0 +1,29 @@
+# ADR 036: Application should be in charge of validator set
+
+## Changelog
+
+
+## Context
+
+Currently Tendermint is in charge of validator set and proposer selection. Application can only update the validator set changes at EndBlock time.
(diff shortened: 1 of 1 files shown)


b00f:
In favor of #3285 and #3325

b00f:
Tendermints wants to do everything by itself and the Application just check the transaction validity. I believe Tendermint should give more responsibility to the app!

melekes:
Thank you for writing this 👍 Do you also want to expand on the interface allowing application to supply another proposer selection algorithm?

b00f:
> Do you also want to expand on the interface allowing application to supply another proposer selection algorithm?

Yes, Tendermint can give this option to the application to select the proposer for next round.

ebuchman:
This needs more work but it's fine to merge as is as a starting point. But it should be changed to ADR-041 (see https://github.com/tendermint/tendermint/issues/2313).

@b00f can you give us permission to commit to this repo?

ebuchman:
I'll just deal with it thanks.

Issue #2805, opened 2018-11-11
test: use `make localnet` instead of current p2p bash setup scripts


ebuchman:
In `/test/p2p/` we use a janky set of bash scripts to set up local docker networks to do various testing.

It would be great to eliminate much of those scripts and use the `make localnet-start` functionality as much as possible instead.

These integration tests can probably be improved in various other ways as well. 

Replaces https://github.com/tendermint/tendermint/issues/1532

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 200.0 DAI (200.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/tendermint/tendermint/2805/3142).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/tendermint/tendermint/2805/3142).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $87,038.84 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


**1) [zyfrank](https://gitcoin.co/profile/zyfrank) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/tendermint/tendermint/2805/3142?mutate_worker_action=approve&worker=zyfrank) | [reject worker](https://gitcoin.co/issue/tendermint/tendermint/2805/3142?mutate_worker_action=reject&worker=zyfrank))_.**

I have some experience on tedermint, I'd like to take this task

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/tendermint/tendermint/2805/3142).

ceresstation:
@zyfrank you're good to go, please keep us updated with your progress!

gitcoinbot:
@zyfrank Hello from Gitcoin Core - are you still working on this issue? Please submit a WIP PR or comment back within the next 3 days or you will be removed from this ticket and it will be returned to an ‘Open’ status. Please let us know if you have questions!
[...]

PR #3332, opened 2019-02-19
Rollback data by cmd

bradyjoestar:
#3196


bradyjoestar:
@ebuchman  It seems that it passed all the required thet.
Could you review this pr and tell me how to modify it? Thanks!

state/state.go
```diff
@@ -48,39 +48,39 @@ var initStateVersion = Version{
...
 // Instead, use state.Copy() or state.NextState(...).
 // NOTE: not goroutine-safe.
+
 type State struct {
-	Version Version
+	Version Version `json:"version"`
 
 	// immutable
-	ChainID string
+	ChainID string `json:"chainId"`
```
ebuchman: chain_id

state/state.go
```diff
@@ -48,39 +48,39 @@ var initStateVersion = Version{
...
 
 	// LastBlockHeight=0 at genesis (ie. block(H=0) does not exist)
-	LastBlockHeight  int64
-	LastBlockTotalTx int64
-	LastBlockID      types.BlockID
-	LastBlockTime    time.Time
+	LastBlockHeight  int64         `json:"lastBlockHeight"`
```
ebuchman: JSON tags use all lowercase and underscores please

state/store.go
```diff
@@ -23,6 +23,10 @@ func calcABCIResponsesKey(height int64) []byte {
 	return []byte(fmt.Sprintf("abciResponsesKey:%v", height))
 }
 
+func calSoftwareHeight(height int64) []byte {
```
ebuchman: calc

It seems a kind suggest changes? Did the repo this is coming from disapear?

state/store.go
```diff
@@ -94,6 +98,9 @@ func saveState(db dbm.DB, state State, key []byte) {
 		lastHeightVoteChanged := int64(1)
 		saveValidatorsInfo(db, nextHeight, lastHeightVoteChanged, state.Validators)
 	}
+	//save software info
+	saveSoftware(db, nextHeight, state.Version.Software)
```
ebuchman: Why does this need to be saved independently of the state? The point of putting the version at the beginning of the state is so we could always read it out using a simple structure even if we change the rest of the fields in the state.

state/store.go
```diff
@@ -94,6 +98,9 @@ func saveState(db dbm.DB, state State, key []byte) {
...
+	// Save next validators.
```
ebuchman: duplicate comment

state/store.go
```diff
@@ -232,12 +239,28 @@ func saveValidatorsInfo(db dbm.DB, height, lastHeightChanged int64, valSet *type
 	valInfo := &ValidatorsInfo{
 		LastHeightChanged: lastHeightChanged,
 	}
-	if lastHeightChanged == height {
-		valInfo.ValidatorSet = valSet
-	}
+	valInfo.ValidatorSet = valSet
```
ebuchman: What happened here?

test/rollback/rollbackAllPeer/rollbackAllPeer.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
ebuchman: I don't love doubling down on the bash script testing since we want to move towards using the nice docker-compose tooling we have (ie. #2805). Anyway we can use that here too?

cmd/tendermint/commands/rollback.go
```diff
@@ -0,0 +1,171 @@
+package commands
```
ebuchman: I wonder if this should be part of the `tendermint` binary or if we should consider a new `tendermint-unsafe` tool that can have the existing unsafe commands plus this command? If possible I think it would be preferable not to crowd `tendermint` with commands like this. Could you start a new `cmd/tendermint-unsafe` and put this under there. We can add the other unsafe commands there in a later PR, and eventually deprecate them from the existing binary.

Also, I don't think the rollback parameters should be part of the main config file - they should just be flags on the command line.

bradyjoestar:
Thank you for your review!
I will carefully read it and change it later on. :grin:

I'm doing it in my spare time so ... I will complete it.

cmd/tendermint/commands/rollback.go
```diff
@@ -0,0 +1,171 @@
...
+	block := blockStore.LoadBlock(rollbackHeight)
+	nextBlock := blockStore.LoadBlock(rollbackHeight + 1)
+
+	return sm.State{
+		ChainID: block.ChainID,
+		Version: sm.Version{Consensus: block.Version, Software: software},
+
+		LastBlockID:      nextBlock.LastBlockID, //? true
+		LastBlockHeight:  block.Height,
+		LastBlockTime:    block.Time,
```
guagualvcha: LoadValidators can't get correct proposer priority and proposer, releted to https://github.com/tendermint/tendermint/pull/3438.
If a validator rollback, the consensus will in chaos.

bradyjoestar: I saved it every block from 
```
	if lastHeightChanged == height {
		valInfo.ValidatorSet = valSet
	}	
```
to
```
valInfo.ValidatorSet = valSet
```
whether lastHeightChanged is true or false.

bradyjoestar: I can't say whether  it's bad or right. For it will change the storage of statedb.
And may cause some performance.

state/store.go
bradyjoestar: I saved validatorSet every block cause we don't have  proposer priority and proposer as @guagualvcha said.

bradyjoestar: I couldnt say it is a good design but I dont know whether there are some better solutions.
We should it really? What result we should bear for this change?

bradyjoestar:
@ebuchman  @guagualvcha

guagualvcha:
> @ebuchman @guagualvcha
> I found we have added a `valSetCheckpointInterval=100000` into state.go
> Could we add a `valSetRollbackInterval = 1000` to save valset every 1000 blocks?
since add`valSetCheckpointInterval=100000` into state.go, I think you can replay the valset now.

bradyjoestar:
sure, I will do it as this way.:ok_hand:

bradyjoestar:
@nolouch
I am recently focuing on libos in the intel-sgx, so sorry for delaying this pr.
I'd like to close this pr temporary.

Sorry!
webb

Issue #3337, opened 2019-02-20
rename autofile#group#Flush to FlushAndSync


melekes:
- Flush does not really imply calling `fsync`
- you have to read the function's docs in order to know that fsync is called

ebuchman:
Per the readme, autofile pkg is not covered by semver, so no need to tag `breaking` :)

melekes:
Merged to develop.

PR #3345, opened 2019-02-22
rename WAL#Flush to WAL#FlushAndSync

melekes:
- rename auto#Flush to auto#FlushAndSync
- cleanup WAL interface to not leak implementation details!
  * remove Group()
  * return io.ReadCloser instead of *auto.Group in SearchForEndHeight()
- add interface assertions

Refs #3337

* [ ] Wrote tests

### DIFF
--- a/consensus/wal.go
+++ b/consensus/wal.go
@@ -53,20 +53,13 @@ func RegisterWALMessages(cdc *amino.Codec) {
 //--------------------------------------------------------
 // Simple write-ahead logger
 
-// WALReader is an interface that embeds io.Reader and allows to close it after
-// you're done reading.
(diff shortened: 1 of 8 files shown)


consensus/wal.go
```diff
@@ -53,14 +53,22 @@ func RegisterWALMessages(cdc *amino.Codec) {
 //--------------------------------------------------------
 // Simple write-ahead logger
 
+// WALReader is an interface that embeds io.Reader and allows to close it after
+// you're done reading.
+type WALReader interface {
+	io.Reader
+	Close() error
```
ebuchman: Isn't this a `io.ReadCloser` ?

Issue #3304, opened 2019-02-12
PeerSet contains dirty peer data cause reconnection failed


guagualvcha:
**Description:** 
The topology is:
A seed node connected to serval validator nodes(Exactly two nodes). 
Two validator nodes is persistent node of seed node.
After running for serval minutes, seed nodes lose all connection to validator, and try to reconnect,
but all failed, and seed node consensus no blocks any more.
  

## Steps To Reproduce:

  1. Deploy a validator network.
  2. Deploy a seed node(seed_mode is true), connected to core network.
  3. Wait for some time or just use `iptable` to produce network joggle.

## Supporting Material/References:
We check the validator logs:
```
I[2019-02-01|13:32:37.737] Starting Peer                                module=p2p peer=77d3f2ee9b5edfd32ffc017e7e8f9e4a1dc1457e@104.193.88.123:27146 impl="Peer{MConn{10.201.41.209:2494} 77d3f2ee9b5edfd32ffc017e7e8f9e4a1dc1457e in}"
I[2019-02-01|13:32:37.737] Starting MConnection                         module=p2p peer=77d3f2ee9b5edfd32ffc017e7e8f9e4a1dc1457e@104.193.88.123:27146 impl=MConn{10.201.41.209:2494}
I[2019-02-01|13:32:37.739] Added peer                                   module=p2p peer="Peer{MConn{10.201.41.209:2494} 77d3f2ee9b5edfd32ffc017e7e8f9e4a1dc1457e in}"
E[2019-02-01|13:32:38.125] Stopping peer for error                      module=p2p peer="Peer{MConn{10.201.41.209:2494} 77d3f2ee9b5edfd32ffc017e7e8f9e4a1dc1457e in}" err="Peer (77d3f2ee9b5edfd32ffc017e7e8f9e4a1dc1457e) sent next PEX request too soon. lastReceived: 2019-02-01 13:32:38.016135243 +0000 UTC m=+41.575722596, now: 2019-02-01 13:32:38.125079205 +0000 UTC m=+41.684666557, minInterval: 10s. Disconnecting"
I[2019-02-01|13:32:38.125] Stopping Peer                                module=p2p peer=77d3f2ee9b5edfd32ffc017e7e8f9e4a1dc1457e@104.193.88.123:27146 impl="Peer{MConn{10.201.41.209:2494} 77d3f2ee9b5edfd32ffc017e7e8f9e4a1dc1457e in}"
[...]

PR #3346, opened 2019-02-22
do not panic if the peer does not have a state in Receive

melekes:
Peer does not have a state yet. We set it in AddPeer, but because the
peer is started before we add it to reactors, we can receive a message
before AddPeer is called.

Refs #3304

* [ ] Wrote tests

### DIFF
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -217,10 +217,13 @@ func (conR *ConsensusReactor) Receive(chID byte, src p2p.Peer, msgBytes []byte) 
 
 	conR.Logger.Debug("Receive", "src", src, "chId", chID, "msg", msg)
 
-	// Get peer states
+	// Get peer state
 	ps, ok := src.Get(types.PeerStateKey).(*PeerState)
 	if !ok {
-		panic(fmt.Sprintf("Peer %v has no state", src))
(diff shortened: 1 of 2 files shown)


consensus/reactor.go
```diff
@@ -217,10 +217,13 @@ func (conR *ConsensusReactor) Receive(chID byte, src p2p.Peer, msgBytes []byte)
...
 	conR.Logger.Debug("Receive", "src", src, "chId", chID, "msg", msg)
 
-	// Get peer states
+	// Get peer state
 	ps, ok := src.Get(types.PeerStateKey).(*PeerState)
 	if !ok {
-		panic(fmt.Sprintf("Peer %v has no state", src))
+		// Peer does not have a state yet. We set it in AddPeer, but because the
+		// peer is started before we add it to reactors, we can receive a message
+		// before AddPeer is called.
```
Liamsi: Is it difficult to write a test that exposes this behaviour?

melekes: Yes

guagualvcha:
seems we try to fix same issue https://github.com/tendermint/tendermint/pull/3361

ebuchman:
Ok let's close this for https://github.com/tendermint/tendermint/pull/3361#issuecomment-468957887 and figure out how to fix from there

PR #3347, opened 2019-02-22
secret connection check all zeroes

Liamsi:
Additionally, to the blacklist of lower order points, check if the shared secret is all zeros.

This mitigates #3010 until we have https://github.com/tendermint/tendermint/issues/3315 and switch to noise (which will also solve this and other issues).

ref: https://github.com/tendermint/tendermint/pull/3040#issuecomment-461978628
ref: https://github.com/tendermint/tendermint/issues/3010 in particular https://github.com/tendermint/tendermint/issues/3010#issuecomment-447123959

* [x] Wrote tests

### DIFF
--- a/p2p/conn/secret_connection.go
+++ b/p2p/conn/secret_connection.go
@@ -29,7 +29,10 @@ const aeadSizeOverhead = 16 // overhead of poly 1305 authentication tag
 const aeadKeySize = chacha20poly1305.KeySize
 const aeadNonceSize = chacha20poly1305.NonceSize
 
-var ErrSmallOrderRemotePubKey = errors.New("detected low order point from remote peer")
+var (
+	ErrSmallOrderRemotePubKey = errors.New("detected low order point from remote peer")
+	ErrSharedSecretIsZero     = errors.New("shared secret is all zeroes")
+)
 
 // SecretConnection implements net.Conn.
 // It is an implementation of the STS protocol.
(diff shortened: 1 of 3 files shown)


p2p/conn/secret_connection.go
```diff
@@ -349,9 +358,20 @@ func deriveSecretAndChallenge(dhSecret *[32]byte, locIsLeast bool) (recvSecret,
 	return
 }
 
-func computeDHSecret(remPubKey, locPrivKey *[32]byte) (shrKey *[32]byte) {
+// computeDHSecret computes a shared secret Diffie-Hellman secret
+// from a the own local private key and the others public key.
```
ebuchman:
```suggestion
// from our own local private key and the other's public key.
```

p2p/conn/secret_connection.go
```diff
@@ -230,9 +236,12 @@ func (sc *SecretConnection) SetWriteDeadline(t time.Time) error {
 
 func genEphKeys() (ephPub, ephPriv *[32]byte) {
 	var err error
+	// TODO: Probably not a problem but ask Tony: different from the rust implementation (uses x25519-dalek),
+	// we do not "clamp" the private key scalar:
+	// see: https://github.com/dalek-cryptography/x25519-dalek/blob/34676d336049df2bba763cc076a75e47ae1f170f/src/x25519.rs#L56-L74
```
Liamsi: @tarcieri: `box.GenerateKey` does not "clamp" the private key. Does this pose a problem? Or should we just ignore this as we will switch to noise soon anyways?

p2p/conn/secret_connection.go
Liamsi: Ouch 🙈 Will submit a mini-followup PR

p2p/conn/secret_connection.go
tarcieri: Curve25519 canonical scalars are always unclamped. Clamping is performed immediately prior to scalar multiplication (which avoids your exact worry about canonical scalars being "pre-clamped" or not):

https://github.com/golang/crypto/blob/master/curve25519/curve25519.go#L789

Issue #3307, opened 2019-02-13
[docs] explain different create_empty_blocks configurations better


jstuczyn:
**Tendermint version**: 0.27.4-1e1ca15b + also tried 0.30.0-28d75ec8
**ABCI app**: not really applicable, but persists on both kvstore and a self-built one
**OS**: Ubuntu 18.04.1

Hi!

I've been trying to increase the minimum time between blocks being created, however, I've repeatedly failed to achieve this. From what I've figured from the documentation (on related note: there is a very limited information regarding the matter), I've tried increasing `blocktime_iota = "10s"`, but this doesn't seem to have changed anything. Then for the test sake I've tried to increase all consensus timeouts tenfolds to see whether there would be any increase in time between blocks:
```
timeout_propose = "30s"
timeout_propose_delta = "5000ms"
timeout_prevote = "10s"
timeout_prevote_delta = "5000ms"
timeout_precommit = "10s"
timeout_precommit_delta = "5000ms"
timeout_commit = "10s"
```
But that seems to have only affected the initial blocks written upon genesis:
```
I[2019-02-13|13:02:17.416] Version info                                 module=main software=0.30.0 block=10 p2p=7
I[2019-02-13|13:02:17.437] Starting Node                                module=main impl=Node
E[2019-02-13|13:02:17.437] Couldn't connect to any seeds                module=p2p 
I[2019-02-13|13:02:17.440] Started node                                 module=main nodeInfo="{ProtocolVersion:{P2P:7 Block:10 App:1} ID_:bfd203973e22de188c505a949210d827ac9a1b87 ListenAddr:tcp://0.0.0.0:26656 Network:test-chain-Cpss48-SINGLE Version:0.30.0 Channels:4020212223303800 Moniker:testingNode Other:{TxIndex:on RPCAddress:tcp://0.0.0.0:26657}}"
I[2019-02-13|13:02:27.458] Executed block                               module=state height=1 validTxs=0 invalidTxs=0
I[2019-02-13|13:02:27.460] Committed state                              module=state height=1 txs=0 appHash=0000000000000000
I[2019-02-13|13:02:37.475] Executed block                               module=state height=2 validTxs=0 invalidTxs=0
I[2019-02-13|13:02:37.478] Committed state                              module=state height=2 txs=0 appHash=0000000000000000
```

When sending the 'regular' transactions, that delay seem to be unaffected:
```
[...]

Issue #2904, opened 2018-11-23
docs: explain how transaction ordering works


pavlovdog:
I need to specify the order of the transactions into the block. In my understanding, for now there's a `fee` field into the `ResponseCheckTx` message. So, inside the block, the transactions are ordered according to it. Am I right?

Does it mean, that e.g. I have 10 transactions: `Tx_1, Tx_2, Tx_3, ..., Tx_10`.

1. I'm broadcasting them in a random way, for example
2. Each transaction has a field with it's order in sequence (e.g. `Tx_1.order = 1`, `Tx_2.order = 2` and so on)
3. During the `CheckTx`call for each of them, I'm specifying the `fee` field of the `ResponseCheckTx` message equal to the `order` field of the transaction.

Am I right, that, if they all will be placed into the same block, their order will be `Tx_1, Tx_2, Tx_3, ..., Tx_4`? If not, how can I implement the specific order into the block?

ebuchman:
Currently, the only way to specify the order is to send them to a single node in order. Nodes will propose txs in a block in the order they received them. If the txs are split up across different nodes, there's no way to ensure they are processed in the expected order. The mempool contains no additional logic to reorder txs.

That said, if the txs contain some internal value, like an order/nonce/sequence number, the application can reject txs that are out of order. So if a node receives tx3, then tx1, it can reject tx3 and then accept tx1. The sender can then retry sending tx3, which should probably be rejected until the node has seen tx2. 

I'll leave this issue open and change the title to indicate we need better documentation around this.

RomanBrodetski:
Does it mean  that essentially it is up to the block proposer to define the order of the transactions? Other nodes don't know in which order those transactions were received by the proposer, so they don't have a reason to reject the block. So, a single byzantine validator, being a proposer, can essentially define the transaction order?

melekes:
Merged to develop.

dakom:
@ebuchman I realize this issue is old, but it seems to be the same now as it was when this issue was first opened.

I am still not clear on where this sequence number is determined.

I understand that a tx sequence number can help all nodes agree on order, but this still means that the original node which receives the tx from the client can lie about the tx order, before broadcasting, correct?

("lie" is not necessarily a fair term here, since it can also just be the actual fair order it received, and only seems out of order from a client perspective due to network latency, but my question is whether the node can also intentionally doctor the sequence order before broadcasting to its peers)

Issue #536, opened 2017-06-14
Error when doing a TX hash query over RPC


davebryson:
Tendermint 0.10.0

Calling  RPC `tx/hash=value` with curl only works if you remove the quotes around the hash value and include the `0x` prefix.  Making this work with an RPC client using JSON-RPC (say in Python) seems to be nearly impossible.  Quoting the hex string and including an 0x prefix returns a `400 bad request` error.  Any other permutation of the hash value doesn't seem to work either.  The errors are a bit misleading.

melekes:
On error codes: there is a PR https://github.com/tendermint/tendermint/pull/521 which will be a part of the next release. Our RPC will comply to http://www.jsonrpc.org/specification.

melekes:
> Quoting the hex string and including an 0x prefix returns a 400 bad request error.

@ebuchman is this supposed to work? `curl -X POST --data '{"jsonrpc":"2.0","method":"tx","params":["0x68656c6c6f20776f726c64"]}'`

If that was never implemented, I think we should do it.

davebryson:
I'm wondering if there's a better overall fix when dealing with hex strings.  It can be very confusing from the client side when pushing hex strings to the RPC server.  Obviously the challenge on the server side is to determine whether it's just a string or hex value.   If it's prefixed with '0x' you can assume it's a hex.  But then the hex value must also be of proper length:  `len(hexvalue) % 2 == 0`  otherwise the decoding fails.   Which means: `0x1` fails, but `0x01` is ok.   I guess at a minimum this should be clearly documented for clients to prevent the confusion.

melekes:
we can copy https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding section and adapt it for our use-case

davebryson:
Yes... I was actually peeking at there json-rpc approach last week when I ran into this :-)

ryansealy:
After doing some troubleshooting trying to implement a binary transaction format for my abci, and after getting an awesome, middle of the night response on the tendermint dev channel, I think this can be cleared up with some better documentation.

https://github.com/tendermint/tendermint/wiki/RPC indicates that you can just send hex for a binary transaction format, but there are a few details that would help clear this up.

Here's what I have tested.

To send a literal UTF8 transaction, you simply quote the value of the tx parameter on the GET request:
`GET http://myserver:46657/broadcast_tx_commit?tx="hello"`
[...]

Issue #2419, opened 2018-09-17
docs: switch to Swagger for RPC


zramsay:
blocking on https://github.com/cosmos/cosmos-sdk/pull/2215 
i.e., waiting to see how the SDK does it (and the Jenkins process running smoothly) then we'll port it over.

zramsay:
#1730 and #2880

zramsay:
note: no longer blocking

safareli:
Using swager would be great as it would allow others to generate client library in desired language!

Is there any updates on this, like is it gonna happen in close future or if this WIP already?

safareli:
This should be closed as of #3880

marbar3778:
just need to switch it out on the website then I believe we can close this. Aiming to do that this week.

PR #3349, opened 2019-02-25
update docs

melekes:
Fixes:
- #3307
- #536 
- #2904


docs/DOCS_README.md
```diff
@@ -66,11 +66,12 @@ To build and serve the documentation locally, run:
...
 ```
 # from this directory
-npm install
 npm install -g vuepress
 ```
 
-then change the following line in the `config.js`:
+NOTE: the command may require `sudo`.
+
+then change the following line in the `.vuepress/config.js`:
```
zramsay: lets make this change in the same file in the SDK repo

rpc/core/mempool.go
```diff
@@ -16,7 +16,8 @@ import (
 //-----------------------------------------------------------------------------
 // NOTE: tx should be signed, but this is only checked at the app level (not by Tendermint!)
 
-// Returns right away, with no response
+// Returns right away, with no response. Does not wait for CheckTx nor
```
zramsay: note to self; update slate once this is merged

docs/tendermint-core/configuration.md
```diff
@@ -268,3 +268,22 @@ max_open_connections = 3
...
 namespace = "tendermint"
 ```
+
+## Empty blocks VS no empty blocks
+
+**create_empty_blocks = true**
+
+If `create_empty_blocks` is set to `true` in your config, blocks will be
+created ~ every second (with default consensus parameters which are explained
+[here](https://forum.cosmos.network/t/consensus-timeouts-explained/1421)). You
```
ebuchman: Lol we should bring that explanation home too

docs/tendermint-core/configuration.md
```diff
@@ -268,3 +268,22 @@ max_open_connections = 3
...
+can regulate the delay between blocks by changing the `timeout_commit`. E.g.
+`timeout_commit = "10s"` should result in ~ 10 second blocks.
+
+**create_empty_blocks = false**
+
+In this setting, blocks are created when txs received. Note after the block N,
+Tendermint creates something we call a "proof block" (if the app hash changes)
+N+1 [#2487 (comment)](https://github.com/tendermint/tendermint/issues/2487#issuecomment-424899799).
```
ebuchman: Will this link render probably? Maybe we can just explain a bit better here?

docs/tendermint-core/configuration.md
```diff
@@ -268,3 +268,22 @@ max_open_connections = 3
...
+Plus, if you set `create_empty_blocks_interval` to something other than the
+default (`0`), Tendermint will be creating empty blocks even in the absence of
+transactions every `create_empty_blocks_interval`.
```
ebuchman:
```suggestion
transactions every `create_empty_blocks_interval`. For instance, with `create_empty_blocks = false` and `create_empty_blocks_interval = "30s"`, Tendermint will only create blocks if there are txs, or after waiting 30s without receiving any txs.
```

rpc/core/mempool.go
```diff
@@ -98,6 +143,50 @@ func BroadcastTxAsync(tx types.Tx) (*ctypes.ResultBroadcastTx, error) {
 // | Parameter | Type | Default | Required | Description     |
 // |-----------+------+---------+----------+-----------------|
 // | tx        | Tx   | nil     | true     | The transaction |
+//
+// To send a UTF8 string byte array, you simply quote the value of the tx parameter on the GET request:
```
ebuchman: Can we deduplicate somehow? Will slate let us link to a previous methods description or to some comment documentation?

rpc/core/mempool.go
```diff
@@ -53,6 +54,50 @@ import (
...
+//
+//     GET http://myserver:46657/broadcast_tx_async?tx="hello"
+//
+// This sends a 5 byte transaction "h e l l o" [68 65 6c 6c 6f]
+//
+//     GET http://myserver:46657/broadcast_tx_async?tx="5"
+//
+// This sends a 4 byte transaction "5" (UTF8) [e2 82 ac 35]
```
ebuchman: Huh? `"5"` is 4 bytes?!

rpc/core/mempool.go
```diff
@@ -53,6 +54,50 @@ import (
...
+//
+// ```
+// {
+//   "method": "broadcast_tx_async",
+//   "jsonrpc": "2.0",
+//   "params": {"tx": "0x01020304"},
+//   "id": ""
+// }
+// ```
+//
```
ebuchman: Lol this is really something :D

rpc/core/mempool.go
melekes: I think it's totally fine to duplicate (until we have better RPC docs / switch to swagger https://github.com/tendermint/tendermint/issues/2419) and not force a person to go somewhere else.

melekes: Actually, looks like we already had this info in "Using Tendermint". Replaced with the link

rpc/core/mempool.go
melekes: mistake

melekes:
Thanks for review guys! Fixed the comments. PTAL

Issue #2699, opened 2018-10-25
blockchain: pool.maxPeerHeight is not updated when a peer is removed


ebuchman:
In the Blockchain Reactor, peers self report their height and we track the highest one, but we can't validate it until we try to receive that block from that peer. Meanwhile, when we disconnect a peer, we don't update the max height. This means a rogue peer could tell us a ridiculously high height, then disconnect, and we will forever be stuck trying to sync to that height, and never switch to consensus. 

We need to either update the pool.maxPeerHeight to the next highest reported height when we disconnect from the peer that reported the max height, or we need a different strategy for deciding to switch to consensus.

Note we should also check what happens when a rogue peer reports a high height, and then we try to get blocks that don't exist yet from good peers. Is there an attack here that could cause us to disconnect good peers?

Thanks @milosevic for catching this in https://github.com/tendermint/tendermint/pull/2621#discussion_r224787555

james-ray:
I'm sorry but I think we already handle this?

func (pool *BlockPool) IsCaughtUp() bool {
	pool.mtx.Lock()
	defer pool.mtx.Unlock()

	// Need at least 1 peer to be considered caught up.
	if len(pool.peers) == 0 {
		pool.Logger.Debug("Blockpool has no peers")
		return false
	}

	// some conditions to determine if we're caught up
	receivedBlockOrTimedOut := (pool.height > 0 || time.Since(pool.startTime) > 5*time.Second)
	ourChainIsLongestAmongPeers := pool.maxPeerHeight == 0 || pool.height >= pool.maxPeerHeight
	isCaughtUp := receivedBlockOrTimedOut && ourChainIsLongestAmongPeers
	return isCaughtUp
}


this code time.Since(pool.startTime) > 5*time.Second  has already thought about that?
We will not 

> forever be stuck trying to sync to that height, and never switch to consensus.

?

melekes:
> We need to either update the pool.maxPeerHeight to the next highest reported height when we disconnect from the peer that reported the max height

Does this solve the problem though? Can't peer pretend to be normal? In the meantime we will be disconnecting from the good nodes because they can't send us blocks with `height = {real height} + 1`.

What if we stop syncing when `receivedBlockOrTimedOut && ourChainIsLongestAmongPeers && {none of the peers can send / sends invalid blocks (we must check it) with height = lastSuccessHeight + 1}`?

ebuchman:
> Can't peer pretend to be normal? 

[...]

PR #3350, opened 2019-02-25
blockchain: update the maxHeight when a peer is removed

melekes:
Refs #2699

* [x] Wrote tests


ancazamfir:
Changes look good to me!  I think they solve the critical issue caused by maxPeerHeight not being updated.

One minor comment. When we remove the peer with highest height, there may be requesters allocated for the higher hights, higher than the new pool.maxPeerHeight, left in pool.requesters map (up to maxTotalRequesters = 600). They will have nil block and peerID "", this is done by func (bpr *bpRequester) reset()
Also the pool.numPending includes these requesters in the count. I believe the requesters remain allocated even after blockchain reactor pool is stopped and the switch to consensus happens.
Maybe we could cleanup them in OnStop()? Also the pool.peers are left around, are they useful once the blockchain reactor is stopped?

melekes:
> One minor comment. When we remove the peer with highest height, there may be requesters allocated for the higher hights, higher than the new pool.maxPeerHeight, left in pool.requesters map (up to maxTotalRequesters = 600)

Ethan said we only request blocks from peers if {peer.height} >= {needed block height}. So there should be no requesters (e.g. `peerB.NewMaxHeight 10 ... peerA.OldMaxHeight 12` deleting peerA should result in removing all requesters with 11-12 heights) after we delete peer with max height.

> Also the pool.numPending includes these requesters in the count. I believe the requesters remain allocated even after blockchain reactor pool is stopped and the switch to consensus happens.
Maybe we could cleanup them in OnStop()?

Yeah, we normally don't clean up such things. I.e. services can't be reset. The important thing is to make sure we decrement/increment when needed, we cleanup all structs allocated for a peer when it's removed and so on.

ancazamfir:
> So there should be no requesters (e.g. `peerB.NewMaxHeight 10 ... peerA.OldMaxHeight 12` deleting peerA should result in removing all requesters with 11-12 heights) after we delete peer with max height.

Where do we remove them? I see that we "reset" them but they seem to be staying there. The only place where we remove requester is in PopRequest().

> If you want to work on `blockchain` package to simplify the code / streamline the logic, please go ahead.

I will probably try something there but after talking to Zarko today it seems that it is a bigger change that may take longer. I believe your fix should go in ASAP with or without the cleanup (if I am right which is not clear yet).

melekes:
> Where do we remove them? I see that we "reset" them but they seem to be staying there. The only place where we remove requester is in PopRequest().

hmm.. can't find it either. I don't know what's the deal with `reset` method. And you may be right that `requesters` map is never cleaned up.

ancazamfir:
> hmm.. can't find it either. I don't know what's the deal with `reset` method.

Not entirely sure either, but I think it's to retry the block request with another peer at that height or higher.

ebuchman:
Merged develop. Had to resolve conflict to support https://github.com/tendermint/tendermint/commit/976b1c2ef73e557a702ef17600e1eeaf4865f292

Issue #3107, opened 2019-01-11
privval: improve heartbeat implementation


ebuchman:
Current implementation of heartbeat spawns an independent routine to call Ping periodically. 

It would be better if the ping timer was reset every time a message was received - that way we wouldn't call Ping more frequently than we actually need to.

We might also want to serialize calls to Ping with other calls to the SignVote and SignProposal, so that they only happen from a single routine and the underlying RemoteSignerClient need not be thread safe.

That should help clarify what needs a mutex and why. See #3064 for more. 

Also, see http://250bpm.com/blog:22 for more info on the subtleties of heartbeats.

melekes:
> It would be better if the ping timer was reset every time a message was received - that way we wouldn't call Ping more frequently than we actually need to.

I think this can be applied to other connections we have as well (e.g. p2p).

Liamsi:
cc @jleni this and https://github.com/tendermint/tendermint/issues/3296 (by @amrali) are the issues I was just talking about

jleni:
I have an issue where invalid signatures leave data in the channel and affect pinging.

Very similar to what @ebuchman was saying, I would suggested to:
- split responsibilities multiplexing/demultiplexing vs command specific work
- move socket multiplexing/demultiplexing to a single go routine + channel and receives/pushes answers to other command specific channels for `ping` or `signature`, etc..
- make ping and other commands to operate on these new command specific queues/channels

We would need to have some specific header to demultiplex but talking to @Liamsi, I understand that things are already in place so the change would be small.

This will avoid thread safety issues and ordering problems in the tendermint-kms comms

jleni:
I can prepare a PR if you agree with change and you think it has enough priority at this stage.

jleni:
I now see that an almost identical discussion in taking place in https://github.com/tendermint/tendermint/issues/3296

ebuchman:
See discussion in https://github.com/tendermint/tendermint/issues/3296#issuecomment-463073468. Moving it here.

ebuchman:
Tracking https://github.com/tendermint/tendermint/issues/3486

ebuchman:
@jleni can we close this?

marbar3778:
This was solved in #3351. In 0.34 privval is migrating to gRPC and use keepalive functionality

jleni:
Excellent news about gRPC!! 
A much better and simpler design? Will tmkms catch up with this?

marbar3778:
[...]

Issue #1234, opened 2018-02-21
TestByzantine non-deterministic failure


ebuchman:
[test_integrations (35).log](https://github.com/tendermint/tendermint/files/1742885/test_integrations.35.log)

https://circleci.com/gh/tendermint/tendermint/3160#artifacts/containers/0

ebuchman:
Should be fixed by #1683

Liamsi:
bubbled up again in https://github.com/tendermint/tendermint/pull/3351#issuecomment-467357000

cmwaters:
Closing this as being stale. Feel free to open this up again if people experience this particular failure again

PR #3351, opened 2019-02-25
Improve Remote Signer implementation

jleni:
This issue is related to #3107 
This is a first renaming/refactoring step before reworking and removing heartbeats.
As discussed with @Liamsi , we preferred to go for a couple of independent and separate PRs to simplify review work.

The changes:
- Help to clarify the relation between the validator and remote signer endpoints
- Differentiate between timeouts and deadlines
- Prepare to encapsulate networking related code behind RemoteSigner in the next PR

* [x] Wrote tests


jleni:
As discussed with @Liamsi, this is a partial renaming/refactoring PR, names and final functionality will get adjusted in a second step.

jleni:
It seems that there is an sporadic issue in test_cover. I leave a reference to the logs here https://circleci.com/gh/tendermint/tendermint/47902#tests/containers/3

Liamsi:
@jleni mentioned that `TestByzantine` is failing which was deemed to be fixed in: https://github.com/tendermint/tendermint/issues/1234
I'll download the log output and restart the build in a moment.

jleni:
Here is the one where TestByzantine fails:

https://circleci.com/gh/tendermint/tendermint/47977#tests/containers/3

privval/file.go
```diff
@@ -86,25 +86,25 @@ type FilePVLastSignState struct {
 func (lss *FilePVLastSignState) CheckHRS(height int64, round int, step int8) (bool, error) {
 
 	if lss.Height > height {
-		return false, fmt.Errorf("Height regression. Got %v, last height %v", height, lss.Height)
+		return false, fmt.Errorf("height regression. Got %v, last height %v", height, lss.Height)
```
Liamsi: I'm all in favour of non-capitalized error messages (as error msg are often wrapped) but capitalized seem to be the default in this code base. I remember when I tried to change this in a PR we decided to keep them capitalized for the sake of consistency ... 
Does this still hold @melekes?

jleni: I can revert if necessary. I was just getting to many warnings and decided to reduce some :)

privval/signer_remote.go
```diff
@@ -7,20 +7,14 @@ import (
 
 	"github.com/pkg/errors"
 
-	amino "github.com/tendermint/go-amino"
+	"github.com/tendermint/go-amino"
```
Liamsi: My IDE also complans about these aliases, and goimports does not add them on my machine. But we use this alias (and similar ones) everywhere. There was even a PR that consistently added them everywhere: https://github.com/tendermint/tendermint/pull/3219
which was the result of running `find . -name '*.go' | grep -v 'vendor/' | xargs -n 1 goimports -w` 
I'm pretty sure you'll be asked to add them back.

jleni: :+1: 
maybe in the future the repos can be renamed to the aliases.
When you rename a repo in GitHub, it will automatically redirect the url is somebody tries to use the old one.

privval/file.go
melekes: > Does this still hold @melekes?

no. let's switch to govet preferred style - non-capitalized error messages

privval/doc.go
```diff
@@ -6,16 +6,16 @@ FilePV
 
 FilePV is the simplest implementation and developer default. It uses one file for the private key and another to store state.
 
-SocketVal
+SignerValidatorEndpoint
```
melekes:
```suggestion
SignerValidatorServer
```

I haven't seen Endpoint suffix before.

privval/messages.go
```diff
@@ -0,0 +1,61 @@
+package privval
+
+import (
+	amino "github.com/tendermint/go-amino"
+	"github.com/tendermint/tendermint/crypto"
+	"github.com/tendermint/tendermint/types"
+)
+
+// RemoteSignerMsg is sent between SignerServiceEndpoint and the SignerServiceEndpoint client.
```
melekes: Does it make sense to rename

SignerServiceEndpoint -> SignerRemoteServer
SignerServiceEndpoint -> SignerRemoteClient

?

privval/signer_service_endpoint.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		res, err := handleRequest(req, se.chainID, se.privVal)
+
+		if err != nil {
+			// only log the error; we'll reply with an error in res
+			se.Logger.Error("handleConnection handleRequest", "err", err)
+		}
+
+		err = writeMsg(conn, res)
```
melekes: Don't we want to reset the connection deadline again before writing?

melekes: Or maybe explicitely use conn.SetWriteDeadline and SetReadDeadline above?

privval/signer_validator_endpoint.go
```diff
@@ -0,0 +1,230 @@
...
+func (ve *SignerValidatorEndpoint) OnStart() error {
+	if closed, err := ve.reset(); err != nil {
+		ve.Logger.Error("OnStart", "err", err)
+		return err
+	} else if closed {
+		return fmt.Errorf("listener is closed")
+	}
+
+	// Start a routine to keep the connection alive
+	ve.cancelPingCh = make(chan struct{}, 1)
```
melekes: why capacity == 1?

privval/messages.go
jleni: I intentionally tried to avoid client/server here, because the concept it is actually mixed up in this implementation.

- Tendermint opens a listening socket  (server like)
- KMS dials in (client like)
- Tendermint initiates requests over the connection (client like)
- KMS answers (server like)

I tried to avoid this confusion in the names.
In the next PR, I will encapsulate this behind the Signer, which at the moment is the opposite too.

privval/signer_validator_endpoint.go
jleni: not sure, implementation was like that before.
Remember anyway that the next PR will remove the Heartbeat/Pings. so I would say this is not very important.

privval/signer_service_endpoint.go
jleni: Yes, I think that would make sense.
I tried to avoid any actual changes in the way things work in this PR to make review simpler.
Maybe I can create an issue for this and fix it in the next PR?

privval/messages.go
melekes: Makes sense. Why the endpoint suffix then? And not just `SignerService`?

Liamsi: I think both `SignerServiceEndpoint` or simply `SignerService` are better than `SocketVal`. Endpoint makes clear it is a connection endpoint. But yeah, let's get this right this time. The renaming is the main purpose of this PR.

jleni: My intention is to separate and encapsulate the "network related" code from the actual signer.

SignerRemote ---(uses/contains)--> SignerValidatorEndpoint <--(connects to)--> SignerServiceEndpoint --->  SignerService (future.. not here yet but would like to decouple too)

All reconnection/heartbeat/whatever code goes in the endpoints. Signer[Remote/Service] do not need to know about that.

I agree Endpoint may not be the perfect name. I tried to find something "Go-ish" enough. It is a common name in go-kit, kubernetes, etc. 

I am happy to change it as long as it makes clear the difference between network and non-network related.

What do you think?

Liamsi: I think your reasoning behind this makes a lot of sense. Maybe we can add this:

> SignerRemote ---(uses/contains)--> SignerValidatorEndpoint <--(connects to)--> SignerServiceEndpoint ---> SignerService (future.. not here yet but would like to decouple too)

in a doc.go package documentation.

jleni: Yes, but it is actually not working this way yet..  
that is why we agreed on rename a few things before changing/breaking things :)

SignerValidatorEndpoint: 
- handles the listener
- contains SignerRemote
- Implements the PrivValidator interface
- connects and sets a connection object in a contained SignerRemote
- delegates PrivValidator some calls to SignerRemote which in turn uses the conn object that was set externally

SignerRemote:
- Implements the PrivValidator interface
- handles heartbeats

SignerServiceEndpoint: 
- Does most things in a single place 
- delegates to a PrivValidator IIRC.

Liamsi: Alright, then let's not document this now but when the refactoring is finished :-)

melekes: Thanks for explaining this 🙇 LGTM

Issue #3353, opened 2019-02-26
 pool timer leak bug


guagualvcha:
### Description
timer leak in bpPeer which cause peer reconnection every 15 s when in fastsync mode

### Rationale
The bpPeer get timer:
```
func (peer *bpPeer) resetTimeout() {
	if peer.timeout == nil {
		peer.timeout = time.AfterFunc(peerTimeout, peer.onTimeout)
	} else {
		peer.timeout.Reset(peerTimeout)
	}
}
```
the AfterFunc will `startTimer`, even this timer is set to nil, the func will still execute:
```
func AfterFunc(d Duration, f func()) *Timer {
	t := &Timer{
		r: runtimeTimer{
			when: when(d),
			f:    goFunc,
			arg:  f,
		},
	}
	startTimer(&t.r)
	return t
}
```
When remove peer, block pool simple remove bpPeer, but do not stop timer, that cause stopError for correct peer.

Before and after fix, the reconnect won't happen again:
![image](https://user-images.githubusercontent.com/7310198/53138601-e3e66c00-35c1-11e9-9853-205d4f241013.png)

### Changes

Notable changes: 
Just stop timer when remove peer.

iphucc:
Has anyone noticed?

https://github.com/tendermint/tendermint/issues/129

https://github.com/iphucc/tendermint/commit/7443575ff9714370761ad9f30c26af8a4be1676a

PR #3358, opened 2019-02-28
fix pool timer leak bug, resolve#3353

guagualvcha:
resolve issue #3353 
When remove peer, block pool simple remove bpPeer,
but do not stop timer, that cause stopError for recorrected
peers. Stop timer when remove from pool.

* [x] Wrote tests


guagualvcha:
@melekes @ebuchman

Issue #3338, opened 2019-02-20
p2p: sent next PEX request too soon


ebuchman:
Some users are reporting seeing this error, potentially when talking to seeds. 

Properly functioning nodes should rate limit how often they send pex requests so they don't send them too frequently. It seems something about this logic is off. 

An example of this happening was reported in #3304, though it seems that issue is about a different problem.

guagualvcha:
### Description
I get some log in data-seed:

```
E[2019-02-20|02:58:48.688] Stopping peer for error                      module=p2p peer="Peer{MConn{10.203.42.19:7369} d375415f420d7a664e52ed98a30a433842be930b in}" err="Peer (d375415f420d7a664e52ed98a30a433842be930b) sent next PEX request too soon. lastReceived: 2019-02-20 02:58:48.56335545 +0000 UTC m=+134240.541996690, now: 2019-02-20 02:58:48.688950674 +0000 UTC m=+134240.667591914, minInterval: 10s. Disconnecting"
E[2019-02-20|03:01:28.898] Stopping peer for error                      module=p2p peer="Peer{MConn{10.203.42.19:57821} d375415f420d7a664e52ed98a30a433842be930b in}" err="Peer (d375415f420d7a664e52ed98a30a433842be930b) sent next PEX request too soon. lastReceived: 2019-02-20 03:01:28.876270203 +0000 UTC m=+134400.854911440, now: 2019-02-20 03:01:28.89826068 +0000 UTC m=+134400.876901920, minInterval: 10s. Disconnecting"
```
### Rationale
When get a PexMessage, it will compare the receive time and lastreceive time. But if follow the step:
1、stopForError fo some peer
2、peer is removed from peerset, PEXReactor.RemovePeer is in the way but id is till not delete from requestsSent and lastReceivedRequests.
3、the same peer is connected, and receive a PexMessage
Then we will make a wrong decision that the peer send pexMessage too fast.

```
func (r *PEXReactor) receiveRequest(src Peer) error {
	id := string(src.ID())
	v := r.lastReceivedRequests.Get(id)
	if v == nil {
		// initialize with empty time
		lastReceived := time.Time{}
		r.lastReceivedRequests.Set(id, lastReceived)
		return nil
	}

	lastReceived := v.(time.Time)
	if lastReceived.Equal(time.Time{}) {
		// first time gets a free pass. then we start tracking the time
		lastReceived = time.Now()
		r.lastReceivedRequests.Set(id, lastReceived)
[...]

PR #3361, opened 2019-02-28
p2p: peer state init too late and pex message too soon

guagualvcha:
fix two issue    https://github.com/tendermint/tendermint/pull/3346 and 
https://github.com/tendermint/tendermint/issues/3338
 1、fix reconnection pex send too fast,
    error is caused lastReceivedRequests is still
    not deleted when a peer reconnected

2、Peer does not have a state yet. We set it in AddPeer.
    We need an new interface before mconnection is started
* [x] Wrote tests

### DIFF
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -225,9 +225,6 @@ func (conR *ConsensusReactor) Receive(chID byte, src p2p.Peer, msgBytes []byte) 
 	conR.Logger.Debug("Receive", "src", src, "chId", chID, "msg", msg)
 
 	// Get peer states
-	if src.Get(types.PeerStateKey) == nil {
-		fmt.Println("============= nil result ========")
-	}
 	ps, ok := src.Get(types.PeerStateKey).(*PeerState)
 	if !ok {
 		panic(fmt.Sprintf("Peer %v has no state", src))
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -161,10 +161,10 @@ func (conR *ConsensusReactor) AddPeer(peer p2p.Peer) {
 		return
 	}
 
-	// Create peerState for peer
-	peerState := NewPeerState(peer).SetLogger(conR.Logger)
-	peer.Set(types.PeerStateKey, peerState)
-
+	peerState, ok := peer.Get(types.PeerStateKey).(*PeerState)
+	if !ok {
+		panic(fmt.Sprintf("Peer %v has no state", peer))
+	}
 	// Begin routines for this peer.
 	go conR.gossipDataRoutine(peer, peerState)
 	go conR.gossipVotesRoutine(peer, peerState)
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -163,7 +163,7 @@ func (conR *ConsensusReactor) AddPeer(peer p2p.Peer) {
 
 	peerState, ok := peer.Get(types.PeerStateKey).(*PeerState)
 	if !ok {
-		panic(fmt.Sprintf("Peer %v has no state", peer))
+		panic(fmt.Sprintf("peer %v has no state", peer))
 	}
 	// Begin routines for this peer.
 	go conR.gossipDataRoutine(peer, peerState)
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -225,6 +225,9 @@ func (conR *ConsensusReactor) Receive(chID byte, src p2p.Peer, msgBytes []byte) 
 	conR.Logger.Debug("Receive", "src", src, "chId", chID, "msg", msg)
 
 	// Get peer states
+	if src.Get(types.PeerStateKey) == nil {
+		fmt.Println("============= nil result ========")
+	}
 	ps, ok := src.Get(types.PeerStateKey).(*PeerState)
 	if !ok {
 		panic(fmt.Sprintf("Peer %v has no state", src))
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -177,7 +177,7 @@ func (conR *ConsensusReactor) AddPeer(peer p2p.Peer) {
 	}
 }
 
-func (conR *ConsensusReactor) InitAddPeer(peer p2p.Peer) p2p.Peer {
+func (conR *ConsensusReactor) InitPeer(peer p2p.Peer) p2p.Peer {
 	// Create peerState for peer
 	peerState := NewPeerState(peer).SetLogger(conR.Logger)
 	peer.Set(types.PeerStateKey, peerState)
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -177,6 +177,13 @@ func (conR *ConsensusReactor) AddPeer(peer p2p.Peer) {
 	}
 }
 
+func (conR *ConsensusReactor) InitAddPeer(peer p2p.Peer) p2p.Peer {
+	// Create peerState for peer
+	peerState := NewPeerState(peer).SetLogger(conR.Logger)
+	peer.Set(types.PeerStateKey, peerState)
+	return peer
+}
+
 // RemovePeer implements Reactor
 func (conR *ConsensusReactor) RemovePeer(peer p2p.Peer, reason interface{}) {
 	if !conR.IsRunning() {
--- a/consensus/reactor_test.go
+++ b/consensus/reactor_test.go
@@ -241,7 +241,7 @@ func TestReactorCreatesBlockWhenEmptyBlocksFalse(t *testing.T) {
 	}, css)
 }
 
-func TestReactorReconnectionPeerGetPeerState(t *testing.T) {
+func TestReactorReceiveDoesNotPanicIfAddPeerHasntBeenCalledYet(t *testing.T) {
 	N := 1
 	css, cleanup := randConsensusNet(N, "consensus_reactor_test", newMockTickerFunc(true), newCounter,
 		func(c *cfg.Config) {
--- a/consensus/reactor_test.go
+++ b/consensus/reactor_test.go
@@ -11,9 +11,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/tendermint/tendermint/crypto/ed25519"
-	"github.com/tendermint/tendermint/p2p/mock"
-
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -22,10 +19,12 @@ import (
 	abci "github.com/tendermint/tendermint/abci/types"
 	bc "github.com/tendermint/tendermint/blockchain"
 	cfg "github.com/tendermint/tendermint/config"
+	"github.com/tendermint/tendermint/crypto/ed25519"
 	dbm "github.com/tendermint/tendermint/libs/db"
 	"github.com/tendermint/tendermint/libs/log"
 	mempl "github.com/tendermint/tendermint/mempool"
 	"github.com/tendermint/tendermint/p2p"
+	"github.com/tendermint/tendermint/p2p/mock"
 	sm "github.com/tendermint/tendermint/state"
 	"github.com/tendermint/tendermint/types"
 )
--- a/consensus/reactor_test.go
+++ b/consensus/reactor_test.go
@@ -11,6 +11,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/tendermint/tendermint/crypto/ed25519"
+	"github.com/tendermint/tendermint/p2p/mock"
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -237,6 +240,36 @@ func TestReactorCreatesBlockWhenEmptyBlocksFalse(t *testing.T) {
 	timeoutWaitGroup(t, N, func(j int) {
 		<-blocksSubs[j].Out()
 	}, css)
+}
+
+func TestReactorReconnectionPeerGetPeerState(t *testing.T) {
+	N := 1
+	css, cleanup := randConsensusNet(N, "consensus_reactor_test", newMockTickerFunc(true), newCounter,
+		func(c *cfg.Config) {
+			c.Consensus.CreateEmptyBlocks = false
+		})
+	defer cleanup()
+	reactors, _, eventBuses := startConsensusNet(t, css, N)
+	defer stopConsensusNet(log.TestingLogger(), reactors, eventBuses)
+	reactor := reactors[0]
+	defer reactor.Stop()
+	sw := reactor.Switch
+	nodeKey := p2p.NodeKey{PrivKey: ed25519.GenPrivKey()}
+	nodeId := nodeKey.ID()
+
+	for i := 0; i < 100; i++ {
+		peer := mock.NewFixIdPeer(nil, nodeId)
+		err := p2p.AddPeerToSwitch(sw, peer)
+		assert.NoError(t, err)
+		assert.True(t, sw.Peers().Has(nodeId))
+		msg := cdc.MustMarshalBinaryBare(&HasVoteMessage{Height: 1, Round: 1, Index: 1, Type: types.PrevoteType})
+		p := sw.Peers().Get(nodeId)
+		assert.NotNil(t, p)
+		reactor.Receive(StateChannel, p, msg) //will not panic
+		sw.StopPeerForError(peer, "peer not available")
+		for sw.Peers().Has(peer.ID()) {
+		}
+	}
 }
 
 // Test we record stats about votes and block parts from other peers.
--- a/p2p/base_reactor.go
+++ b/p2p/base_reactor.go
@@ -29,8 +29,8 @@ type Reactor interface {
 	// CONTRACT: msgBytes are not nil.
 	Receive(chID byte, peer Peer, msgBytes []byte)
 
-	// InitAddPeer is called by switch before peer is started.
-	InitAddPeer(peer Peer) Peer
+	// InitPeer is called by switch before peer is started.
+	InitPeer(peer Peer) Peer
 }
 
 //--------------------------------------
@@ -54,4 +54,4 @@ func (*BaseReactor) GetChannels() []*conn.ChannelDescriptor        { return nil 
 func (*BaseReactor) AddPeer(peer Peer)                             {}
 func (*BaseReactor) RemovePeer(peer Peer, reason interface{})      {}
 func (*BaseReactor) Receive(chID byte, peer Peer, msgBytes []byte) {}
-func (*BaseReactor) InitAddPeer(peer Peer) Peer                    { return peer }
+func (*BaseReactor) InitPeer(peer Peer) Peer                       { return peer }
--- a/p2p/base_reactor.go
+++ b/p2p/base_reactor.go
@@ -28,6 +28,9 @@ type Reactor interface {
 	//
 	// CONTRACT: msgBytes are not nil.
 	Receive(chID byte, peer Peer, msgBytes []byte)
+
+	// InitAddPeer is called by switch before peer is started.
+	InitAddPeer(peer Peer) Peer
 }
 
 //--------------------------------------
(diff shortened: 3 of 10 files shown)


guagualvcha:
@ebuchman  @me
   https://github.com/tendermint/tendermint/issues/3338 and https://github.com/tendermint/tendermint/pull/3346
fix here

p2p/pex/pex_reactor.go
```diff
@@ -189,6 +189,13 @@ func (r *PEXReactor) logErrAddrBook(err error) {
 	}
 }
 
+func (r *PEXReactor) InitAddPeer(peer Peer) Peer {
+	id := string(peer.ID())
+	r.requestsSent.Delete(id)
```
ebuchman: Can't we just add this to AddPeer?

ebuchman:
I see. The peer is started and thus might receive a message before AddPeer is called

ebuchman:
What if we let Receive call AddPeer, and add some atomic integer so AddPeer only gets called once?

p2p/pex/pex_reactor.go
ebuchman: And why isn't it enough for this to be done in RemovePeer?

guagualvcha:
> What if we let Receive call AddPeer, and add some atomic integer so AddPeer only gets called once?

If we do like this, when the reactor want to send message to the peer first, but it have to wait until it receive some message?

p2p/pex/pex_reactor.go
guagualvcha: It may do concurrently

ebuchman:
> If we do like this, when the reactor want to send message to the peer first, but it have to wait until it receive some message?

No we'd still have AddPeer get called by the switch, but if it turns out a message is received before the AddPeer, Receive could make sure AddPeer gets called first. Then the other AddPeer called by the Switch should have no effect

guagualvcha:
> > If we do like this, when the reactor want to send message to the peer first, but it have to wait until it receive some message?
> 
> No we'd still have AddPeer get called by the switch, but if it turns out a message is received before the AddPeer, Receive could make sure AddPeer gets called first. Then the other AddPeer called by the Switch should have no effect

Maybe , I will submit another commit , and let's see if it make sense

melekes:
I can try and rewrite the code to use `sync.Once` so we can compare two versions.

guagualvcha:
@melekes  @ebuchman any news for this pr?  
You may complain I add  new interface `InitAddPeer` to `Reactor`. in my opinion, this implement is simple, easy to understand, and it works.

consensus/reactor.go
```diff
@@ -177,6 +177,13 @@ func (conR *ConsensusReactor) AddPeer(peer p2p.Peer) {
 	}
 }
 
+func (conR *ConsensusReactor) InitAddPeer(peer p2p.Peer) p2p.Peer {
```
melekes: Should we return a peer OR just state in the interface that InitAddPeer might modify the peer?

p2p/base_reactor.go
```diff
@@ -28,6 +28,9 @@ type Reactor interface {
 	//
 	// CONTRACT: msgBytes are not nil.
 	Receive(chID byte, peer Peer, msgBytes []byte)
+
+	// InitAddPeer is called by switch before peer is started.
+	InitAddPeer(peer Peer) Peer
```
melekes: I like `InitPeer` better

melekes:
I agree with @guagualvcha

consensus/reactor.go
guagualvcha: I prefer return a peer, more literally, no not need explain anything.  What do you think?

p2p/base_reactor.go
guagualvcha: updated @melekes @ebuchman @xla  would you like review again?

melekes:
Can you 

1) add a changelog entry in `CHANGELOG_PENDING.md`?

```
BUGFIXES:
- [p2p] \#3346 Init data structures for peer before starting it (previously was done in AddPeer) (fixes #3346 and #3338; @guagualvcha)
```

2) reformat code with `gofmt -s -w .`?

guagualvcha:
> Can you
> 
> 1. add a changelog entry in `CHANGELOG_PENDING.md`?
> 
> ```
> BUGFIXES:
> - [p2p] \#3346 Init data structures for peer before starting it (previously was done in AddPeer) (fixes #3346 and #3338; @guagualvcha)
> ```
> 1. reformat code with `gofmt -s -w .`?

updated

guagualvcha:
@ebuchman  @xla  could you like to review this?

consensus/reactor.go
```diff
@@ -177,6 +177,13 @@ func (conR *ConsensusReactor) AddPeer(peer p2p.Peer) {
 	}
 }
 
+func (conR *ConsensusReactor) InitPeer(peer p2p.Peer) p2p.Peer {
+	// Create peerState for peer
+	peerState := NewPeerState(peer).SetLogger(conR.Logger)
+	peer.Set(types.PeerStateKey, peerState)
```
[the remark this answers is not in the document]
guagualvcha: right. I forget delete code of line 165 and 166. Fixed

p2p/switch.go
```diff
@@ -681,6 +681,9 @@ func (sw *Switch) addPeer(p Peer) error {
 	}
 	sw.metrics.Peers.Add(float64(1))
 
+	for _, reactor := range sw.reactors {
+		p = reactor.InitPeer(p)
+	}
```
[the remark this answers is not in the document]
guagualvcha: image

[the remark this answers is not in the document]
guagualvcha: keyPoint is :  there is not handshake that tell other peer I am ready to receive message. So do some init work before the mconnection start.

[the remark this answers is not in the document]
xla: If I understand correctly is that the peer's mconnection is started too early, but the introduction of `InitPeer` won't help with that as [`p.Start()`](https://github.com/tendermint/tendermint/blob/e82f69babb977655751fbcac168cebe5daa94bc3/p2p/switch.go#L669) is called befpre that.

The only sensible way forward is to find a way to write a failing test for this scenario and then verify that any proposed solution solves.

ebuchman:
> For the change itself we introduce a new method on the already large reactor which adds more peer lifecycle management responsibiliies outside of the p2p package which only seem to affect consensus and pex. It is cocneivable to not have this a major blocker for this change, but something I find worth pointing out. Maybe there is a more isolated fix.

Mostly agree with this. Shouldn't necessarily block making this fix, but we should at least open a new issue to track this problem and see if there's a better solution that doesn't require the new method.

> The only sensible way forward is to find a way to write a failing test for this scenario and then verify that any proposed solution solves.

@guagualvcha is it conceivable to write a test for this ?

guagualvcha:
> > For the change itself we introduce a new method on the already large reactor which adds more peer lifecycle management responsibiliies outside of the p2p package which only seem to affect consensus and pex. It is cocneivable to not have this a major blocker for this change, but something I find worth pointing out. Maybe there is a more isolated fix.
> 
> Mostly agree with this. Shouldn't necessarily block making this fix, but we should at least open a new issue to track this problem and see if there's a better solution that doesn't require the new method.
> 
> > The only sensible way forward is to find a way to write a failing test for this scenario and then verify that any proposed solution solves.
> 
> @guagualvcha is it conceivable to write a test for this ?

sure, I will keep working on thins.

consensus/reactor_test.go
```diff
@@ -583,3 +615,14 @@ func capture() {
 	count := runtime.Stack(trace, true)
 	fmt.Printf("Stack of %d bytes: %s\n", count, trace)
 }
+
+func createSwitchAndAddReactors(reactors ...p2p.Reactor) *p2p.Switch {
```
golangcibot: `createSwitchAndAddReactors` is unused (from `deadcode`)

consensus/reactor_test.go
```diff
@@ -3,6 +3,8 @@ package consensus
 import (
 	"context"
 	"fmt"
+	"github.com/tendermint/tendermint/crypto/ed25519"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

p2p/pex/pex_reactor_test.go
```diff
@@ -2,6 +2,7 @@ package pex
 
 import (
 	"fmt"
+	"github.com/tendermint/tendermint/crypto/ed25519"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

guagualvcha:
@xla @ebuchman  I have finished the test case work. Please have a review then.

xla:
@guagualvcha Thanks for taking the time and following up on this. Currently the test-cases don't fail when the `InitPeer` changes are removed, which leads me to believe that they are not proving the change  is required or the tests don't fail in the right way. So we still need to find a way to show that without this change messages can be send too soon.

guagualvcha:
> @guagualvcha Thanks for taking the time and following up on this. Currently the test-cases don't fail when the `InitPeer` changes are removed, which leads me to believe that they are not proving the change is required or the tests don't fail in the right way. So we still need to find a way to show that without this change messages can be send too soon.

definitely right. I will add it then

p2p/pex/pex_reactor_test.go
```diff
@@ -35,6 +37,77 @@ func TestPEXReactorBasic(t *testing.T) {
...
+	defer teardownReactor(book)
+
+	sw := createSwitchAndAddReactors(r)
+	sw.SetAddrBook(book)
+	sw.Start()
+	defer sw.Stop()
+
+	nodeKey := p2p.NodeKey{PrivKey: ed25519.GenPrivKey()}
+	nodeId := nodeKey.ID()
+	stopForError:=0
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

xla:
@guagualvcha Please don't force push as it makes it hard to review latest changes. We have a poliyc of squash merging, so long commit histories are not a problem.

p2p/pex/pex_reactor_test.go
```diff
@@ -35,6 +37,77 @@ func TestPEXReactorBasic(t *testing.T) {
 	assert.NotEmpty(t, r.GetChannels())
 }
 
+func TestPEXReactorStopPeerWithOutInitPeer(t *testing.T) {
```
xla: This test doesn't always fail when `InitiPeer` is removed.

p2p/pex/pex_reactor_test.go
```diff
@@ -35,6 +37,77 @@ func TestPEXReactorBasic(t *testing.T) {
...
+		}
+
+		go sw.StopPeerForError(peer, "peer not available")
+		for sw.Peers().Has(peer.ID()) {
+		}
+	}
+	assert.NotEqual(t, stopForError, 0)
+}
+
+func TestPEXReactorDoNotStopReconnectionPeer(t *testing.T) {
```
xla: Same for this test, it only fails sometimes.

p2p/pex/pex_reactor_test.go
xla: It sometimes failed with `InitPeer`.

guagualvcha: The issue I fixed is also not always happen. It depends on the schedule of goroutine,  as I describe:
image 
this is happen sometimes.  That is why the case run 10000 times to make it happen.

p2p/pex/pex_reactor_test.go
guagualvcha: So do this case

p2p/pex/pex_reactor_test.go
xla: If the success of the fix is sensistive to the inner workings of the runtme we need to find a way to never fail, no matter how the routines are scheduled.

consensus/reactor.go
```diff
@@ -177,6 +177,13 @@ func (conR *ConsensusReactor) AddPeer(peer p2p.Peer) {
...
+	return peer
+}
+
```
melekes: Don't we need to remove `peer.Set(types.PeerStateKey, peerState)` from `AddPeer`? Otherwise it will override whenever was written between `InitPeer` and `AddPeer` calls.

consensus/reactor_test.go
```diff
@@ -11,6 +11,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/tendermint/tendermint/crypto/ed25519"
```
melekes: can you place these after 3rd party testify imports?

p2p/mock/peer.go
```diff
@@ -40,6 +40,24 @@ func NewPeer(ip net.IP) *Peer {
 	return mp
 }
 
+func NewFixIdPeer(ip net.IP, id p2p.ID) *Peer {
```
melekes: Can't we use `NewPeer`? Why need to provide an external ID?

p2p/pex/pex_reactor.go
```diff
@@ -216,6 +216,13 @@ func (r *PEXReactor) logErrAddrBook(err error) {
...
+func (r *PEXReactor) InitPeer(peer Peer) Peer {
+	id := string(peer.ID())
+	r.requestsSent.Delete(id)
+	r.lastReceivedRequests.Delete(id)
+	return peer
+}
+
 // RemovePeer implements Reactor.
 func (r *PEXReactor) RemovePeer(p Peer, reason interface{}) {
 	id := string(p.ID())
```
melekes: same here. why do we need to do the same thing in `RemovePeer` (I'm talking about `r.requestsSent.Delete(id)` and `r.lastReceivedRequests.Delete(id)`)

p2p/pex/pex_reactor_test.go
```diff
@@ -8,6 +8,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/tendermint/tendermint/crypto/ed25519"
```
melekes: can we place this after 3rd party testify imports, not before?

p2p/mock/peer.go
melekes: Can't we overwrite ID in any case?

```go
peer = mock.NewPeer(nil)
peer.ID = nodeID
```

p2p/pex/pex_reactor_test.go
```diff
@@ -35,6 +37,77 @@ func TestPEXReactorBasic(t *testing.T) {
...
+		// first time creates the entry
+		r.Receive(PexChannel, peer, msg)
+		assert.True(t, sw.Peers().Has(peer.ID()))
+
+		// next time sets the last time value
+		r.Receive(PexChannel, peer, msg)
+		assert.True(t, sw.Peers().Has(peer.ID()))
+
+		go sw.StopPeerForError(peer, "peer not available")
+		for sw.Peers().Has(peer.ID()) {
```
melekes: potentially infinite loop

p2p/pex/pex_reactor_test.go
```diff
@@ -35,6 +37,77 @@ func TestPEXReactorBasic(t *testing.T) {
...
+			stopForError++
+		}
+		// next time sets the last time value
+		r.Receive(PexChannel, peer, msg)
+		if !sw.Peers().Has(peer.ID()) {
+			stopForError++
+		}
+
+		go sw.StopPeerForError(peer, "peer not available")
+		for sw.Peers().Has(peer.ID()) {
```
melekes: potentially infinite loop

consensus/reactor.go
guagualvcha: right. Remove it from AddPeer

consensus/reactor_test.go
guagualvcha: yes

p2p/mock/peer.go
guagualvcha: It is an option, but we should expose ID.

guagualvcha:
@melekes @xla  Sorry for late replay, busy with other work.   I have 1. delete potentially infinite loop 2. make test case stable.  Please see my latest commit.

p2p/switch.go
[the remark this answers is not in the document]
ancazamfir: I tend to agree with @xla here, not clear the fix will work. Curious, have you tried to actually delay the peer reconnecting? (e.g. in switch.go:reconnectToPeer() bring the sw.randomSleep() just at the beginning of the first for loop.)
My understanding is that the root cause of the issue is the peer reconnecting too fast, before RemovePeer() has a chance to cleanup. I think it would be ok to delay the reconnection of a peer that was stopped for error.

consensus/reactor_test.go
```diff
@@ -239,6 +241,36 @@ func TestReactorCreatesBlockWhenEmptyBlocksFalse(t *testing.T) {
 	}, css)
 }
 
+func TestReactorReconnectionPeerGetPeerState(t *testing.T) {
```
melekes:
```suggestion
func TestReactorReceiveDoesNotPanicIfAddPeerHasntBeenCalledYet(t *testing.T) {
```

p2p/mock/peer.go
melekes: lets do that. it's a mock peer anyway

consensus/reactor_test.go
```diff
@@ -239,6 +241,36 @@ func TestReactorCreatesBlockWhenEmptyBlocksFalse(t *testing.T) {
...
+	defer stopConsensusNet(log.TestingLogger(), reactors, eventBuses)
+	reactor := reactors[0]
+	defer reactor.Stop()
+	sw := reactor.Switch
+	nodeKey := p2p.NodeKey{PrivKey: ed25519.GenPrivKey()}
+	nodeId := nodeKey.ID()
+
+	for i := 0; i < 100; i++ {
+		peer := mock.NewFixIdPeer(nil, nodeId)
+		err := p2p.AddPeerToSwitch(sw, peer)
```
melekes: we don't need to initialize switch & add peer to it in order to test that reactor does not panic anymore. we can just call reactor.InitPeer() here and write a comment (CONTRACT) saying that if InitPeer is not called, Reactor will panic!

melekes: I can do that

p2p/pex/pex_reactor.go
```diff
@@ -216,11 +216,11 @@ func (r *PEXReactor) logErrAddrBook(err error) {
...
 }
 
-// RemovePeer implements Reactor.
-func (r *PEXReactor) RemovePeer(p Peer, reason interface{}) {
-	id := string(p.ID())
+func (r *PEXReactor) InitPeer(peer Peer) Peer {
+	id := string(peer.ID())
 	r.requestsSent.Delete(id)
 	r.lastReceivedRequests.Delete(id)
+	return peer
```
melekes: do we need to remove these lines from `RemovePeer`? **if no**, we should extract them to a function call it in both places

p2p/pex/pex_reactor_test.go
```diff
@@ -35,6 +37,73 @@ func TestPEXReactorBasic(t *testing.T) {
 	assert.NotEmpty(t, r.GetChannels())
 }
 
+func TestPEXReactorStopPeerWithOutInitPeer(t *testing.T) {
```
melekes: can we rename this to something which reflects the test's purpose?

I don't understand what the test is testing right now.

p2p/pex/pex_reactor_test.go
```diff
@@ -35,6 +37,73 @@ func TestPEXReactorBasic(t *testing.T) {
...
+		r.Receive(PexChannel, peer, msg)
+		if !sw.Peers().Has(peer.ID()) {
+			stopForError++
+		}
+		p2p.RemovePeerFromSwitchPeerSet(sw, peer)
+	}
+	assert.Equal(t, stopForError, 99)
+}
+
+func TestPEXReactorDoNotStopReconnectionPeer(t *testing.T) {
```
melekes: can we rename this to something which reflects the test's purpose?

I don't understand what the test is testing right now.

melekes:
I am going to take over this.

melekes:
Closing in favor of https://github.com/tendermint/tendermint/pull/3634.

p2p/switch.go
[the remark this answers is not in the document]
melekes: > My understanding is that the root cause of the issue is the peer reconnecting too fast, before RemovePeer() has a chance to cleanup. 

**the core issue here** is that `reactor.Receive` might be called before `reactor.AddPeer`, which will lead to reactor panicking if it depends on data from`reactor.AddPeer`. **The proposed solution** is to move data initialization to a new `reactor.InitPeer` function, which is guaranteed to be executed before `reactor.Receive` is called.

Issue #3289, opened 2019-02-09
Avoid requiring entire `go-ethereum` commit in lockfile


cwgoes:
See if there's a way we can just bind to an exact version of the `secp256k1` library, which is all we need.

Perhaps copy that code out into our own repository.

Ref https://github.com/cosmos/cosmos-sdk/issues/3564

Liamsi:
> Perhaps copy that code out into our own repository.

I think that is the way to go until the related dep issue is resolved: https://github.com/golang/dep/issues/1847

Liamsi:
Addressed in https://github.com/tendermint/tendermint/pull/3371

PR #3371, opened 2019-03-01
Copy secp256k1 code from go-ethereum to avoid GPL vendoring issues in

silasdavis:
This copies the CGO secp256k1 implementation from https://github.com/ethereum/go-ethereum/tree/729bf365b5f17325be9107b63b233da54100eec6/crypto/secp256k1 added as dependency by https://github.com/tendermint/tendermint/pull/3234.

The code in question is MIT not GPL licensed but dep will still pull in the entire go-ethereum library. This causes and issue for downstream projects like http://github.com/hyperledger/burrow that vendor Tendermint and are permissively licensed. Pruning would sort-of solve this but doesn't work for CGO code, and in any case for code that is as stable over time as this is I think a little duplication is cleaner.

From my conversation with @Liamsi I believe this was plan anyway, so here we are :)

Some other justifications for this:
- go-ethereum doesn't own the core code in this package and they haven't update it for 2 years, so there is little to gain from having it as a formal dependency
- copying the bit that is needed in means that you won't unncessarily create conflicts with any downstream projects that would like to depend on a version of go-ethereum of their choosing (cough ethermint cough)
- Managing go packages doesn't get any _more_ painful for me in Burrow :)


Liamsi:
ref: https://github.com/tendermint/tendermint/issues/3289

Liamsi:
I've verified that the code under `crypto/secp256k1/internal/secp256k1` in this PR is basically the same as in https://github.com/ethereum/go-ethereum/tree/729bf365b5f17325be9107b63b233da54100eec6/crypto/secp256k1
"Basically" because there is a test-file which was deleted `secp256_test.go` and comment added to the readme:
```
+++ b/crypto/secp256k1/internal/secp256k1/README.md
@@ -0,0 +1,3 @@
+This package is copied from https://github.com/ethereum/go-ethereum/tree/729bf365b5f17325be9107b63b233da54100eec6/crypto/secp256k1
+
+Unlike the rest of go-ethereum it is MIT licensed so compatible with our Apache2.0 license. We opt to copy in here rather than depend on go-ethereum to avoid issues with vendoring of the GPL parts
 of that repository by downstream.

```
Codeowners (@melekes, @ebuchman) can verify this themselves by doing the following:

 1) checkout the PR changes locally (e.g. via `git fetch origin pull/3371/head:silasdavis/internalise-libsecp256k1 &&  git checkout silasdavis/internalise-libsecp256k1`)
 2) temporarily get go-ethereum repo: `go get -v -d github.com/ethereum/go-ethereum`
 3)  `cd $GOPATH/src/github.com/ethereum/go-ethereum/` and checkout the commit Silas' mentioned: `git checkout 729bf365b5f17325be9107b63b233da54100eec6` (will be in detached head)
 4) go back to tendermint `cd $GOPATH/src/github.com/tendermint/tendermint/`
 5)  `git diff  $GOPATH/src/github.com/ethereum/go-ethereum/crypto/secp256k1/ crypto/secp256k1/internal/`

Here is the full output of above operation (created via `git diff --no-prefix $GOPATH/src/github.com/ethereum/go-ethereum/crypto/secp256k1/ crypto/secp256k1/internal/ > diff_from_ethereum.patch`):

[diff_from_ethereum.txt](https://github.com/tendermint/tendermint/files/2930417/diff_from_ethereum.txt)

ebuchman:
Verified.

But why was the test file deleted?

PR #3374, opened 2019-03-04
Move dependency tracking into ./cmd

coriolinus:
It is a truism in go that libraries should never vendor their dependencies; dependency management is the responsibility of the binary author. Even so, tools like dep try to do the smart thing, and search the import paths of downstream projects for `Gopkg.lock`, `glide.lock`, etc. This means that if I import `github.com/tendermint/tendermint/abci/types` within my codebase, I still end up with all the dependencies that tendermint requires.

This became a problem recently: my team upgraded to go 1.12. Tendermint, however, specifies a pinned version of `golang.org/x/crypto`, which is not the version of the library which go 1.12 uses. This produces an irreconcilable dependency graph.

Moving the dep files into `./cmd` ensures that downstream projects such as mine are not affected by Tendermint's constraints, while retaining full flexibility for Tendermint to pin whatever it likes.

None of the subsequent boxes are checked because this PR does not involve any code changes. 

* [ ] Wrote tests


coriolinus:
Moving dep's files helps because it takes them out of the import path. For example:

- In my project, I wish to import `github.com/tendermint/tendermint/abci/types`.
- As things stand, dep recursively checks every component of that path, and discovers a `Gopkg.lock` within `github.com/tendermint/tendermint`; it therefore applies all those constraints to my project, including ones which break things for me.
- Moving `Gopkg.lock` to the `./cmd` folder fixes this issue because I can ensure that nothing in my project ever imports `github.com/tendermint/tendermint/cmd` or any subfolder. As by convention, the source within `cmd` just contains the code which builds executables, this isn't a problem for me.
- This is also in accordance with the principle that dependency management is the responsibility of binary packages, not libraries.

I've noted the build failures and am slowly iterating my way through them. I'd appreciate help with this, as I'm not particularly familiar with the build system used in this project. However, the sort description of how to fix this:

- ensure the CWD during `dep ensure` is `./cmd`; this produces `./cmd/vendor`
- ensure the CWD during `go build` is `./cmd`; this finds and uses `./cmd/vendor` properly to get the vendored dependencies.

coriolinus:
I've fixed what I could in the subsequent commits. Tendermint now builds properly. This doesn't fix the CI issues because Tendermint has a fairly complex test system, and the tests fail. This should at least have demonstrated the intended technique. 

I had to add `tendermint/tendermint` as a dependency and specify the current branch. That's obviously not desirable, but it is a `dep` limitation: https://github.com/golang/dep/issues/2116. 

This is the limit of what I can contribute on this issue: while it's inconvenient for my project to have Tendermint's constraints (particularly the unused one to `golang.org/x/crypto` which is incompatible with go 1.12), we already have a workaround. I hope this is sufficiently persuasive to convince the project maintainers to continue work along this path.

Liamsi:
> This is the limit of what I can contribute on this issue: while it's inconvenient for my project to have Tendermint's constraints (particularly the unused one to `golang.org/x/crypto` which is incompatible with go 1.12), we already have a workaround.

I've updated circleci to use go1.12 and did not run into any issues regarding `golang.org/x/crypto` https://github.com/tendermint/tendermint/pull/3376
Also, as far as I can see we use `golang.org/x/crypto` (for instance in secret connection). AFAIK, a few things got (rightfully) deprecated there recently. I can submit a PR that updates the dependency in tendermint, too.

coriolinus:
Thanks. This is the message I got which prompted this whole issue for me:

```
Solving failure: No versions of github.com/tendermint/tendermint met constraints:
    v0.30.1: Could not introduce github.com/tendermint/tendermint@v0.30.1, as it has a dependency on golang.org/x/crypto with constraint 505ab145d0a99da450461ae2c1a9f6cd10d1f447, which does not allow the currently selected version of release-branch.go1.12
```

Environment:

```sh
$ go version
Mon Mar  4 14:15:57 CET 2019
go version go1.12 darwin/amd64
$ dep version
Mon Mar  4 14:16:08 CET 2019
dep:
 version     : v0.5.0
 build date  : 2018-08-16
 git hash    : 224a564
 go version  : go1.10.3
 go compiler : gc
 platform    : darwin/amd64
 features    : ImportDuringSolve=false
```

Liamsi:
Looks like in your project's Gopkg.toml you use the constraint ` branch = "release-branch.go1.12"`. If there isn't any particular change you definitely need, you can simply use the same constraint as tendermint instead: https://github.com/tendermint/tendermint/blob/e0f8936455029a40287a69d5b0e7baa4d5864da1/Gopkg.toml#L98-L100
go1.12 does _not_ depend on using the release branch you are using. I'll update our dependency on `golang.org/x/crypto` shortly (to the latest commit on master and not to the `release-branch.go1.12` branch though).

coriolinus:
No, that's not the case. `x/crypto` didn't appear in our `Gopkg.toml` at all before running into this issue.

Incidentally, the fix which unbroke things for us was to add an override:

```toml
[[override]]
name = "golang.org/x/crypto"
```

This just unset the constraint you provided.

Liamsi:
Interesting, does one of your dependencies introduce `release-branch.go1.12` as a dependency?

coriolinus:
Not to the best of my knowledge, though I suppose it's possible. If there is a way to make dep report the origin of a particular constraint, I don't know it. The report from dep blamed tendermint, and the only change we've made recently in that regard is to upgrade from 1.11 to 1.12, which is why I made the PR here.

melekes:
@coriolinus Thanks for the PR 👍 A few notes here:

- We pin the deps with no releases to exact versions to prevent accidental upgrades
- I agree that importing `github.com/tendermint/tendermint/abci/types` should not bear all of Tendermint's requirements. Maybe this is the sign we should rethink our decision to bring abci repo https://github.com/tendermint/abci/ into tendermint 
- But at the same time, how a library can signal a particular dependency? For example, we (Tendermint) import `packageA` at a particular revision because other revisions break the API. If we didn't have Gopkg.toml, a user would have to try and figure what version Tendermint requires on her own, which is not great!

I think for now the short term solution is to use dep's override stanza https://golang.github.io/dep/docs/Gopkg.toml.html#override

```
[[override]]
name = "golang.org/x/crypto"
...
```

> It is a truism in go that libraries should never vendor their dependencies; dependency management is the responsibility of the binary author.

Could you please point me to the sources of that recommendation?

coriolinus:
> But at the same time, how a library can signal a particular dependency?

Agree, this is a limitation of the go language. Somehow things that in Rust, Python, JS aren't particular issues, are big deals in this language. I didn't get to pick the language, though: my boss did, based largely on the existence of tendermint. Sunk costs mean that we're never going to change the primary language at this point.

> Could you please point me to the sources of that recommendation?

There's no official directive that libraries must not vendor their dependencies, only a collection of [best practices documents](https://peter.bourgon.org/go-best-practices-2016/#dependency-management), [golang dev forum posts](https://groups.google.com/d/msg/golang-dev/4FfTBfN2YaI/M5SD8_FqCgAJ), [go-nuts](https://groups.google.com/d/msg/golang-nuts/AnMr9NL6dtc/JN_tkJFHCAAJ) [forum posts](https://groups.google.com/d/msg/golang-nuts/AnMr9NL6dtc/nqBvQTdMCAAJ), etc.

FWIW, I agree that putting all the vendoring data outside the repo root is a pain in its own way. My team ended up establishing a `commands` repository which contains nearly all our executable packages (excepting things like codegen), and the only vendoring in our codebase. All our other code lives in other repositories, which are forbidden to vendor anything. While hacking on a library, we just build from within the library's repo, meaning go uses (non-vendored) local copies of everything. It's relatively convenient for us, but it's also a much bigger change than I'd expect to be approved in a PR.

melekes:
> Somehow things that in Rust, Python, JS aren't particular issues, are big deals in this language.

😞 

We're probably going to write Tendermint version in Rust at some point. Also, we have https://github.com/tendermint/rust-abci, so people can write their applications in Rust right now.

> There's no official directive that libraries must not vendor their dependencies, only a collection of best practices documents, golang dev forum posts, go-nuts forum posts, etc.

Thanks! We should discuss this more before making a move.

melekes:
```
Importing configuration from govendor. These are only initial constraints, and are further refined during the solve process.
  Locking in v1.4.0 (66b9c49) for transitive dep github.com/gorilla/websocket
  Locking in  (505ab14) for transitive dep golang.org/x/crypto
  Locking in master (fd36f42) for transitive dep github.com/prometheus/client_model
  Locking in v1.0.1 (c12348c) for transitive dep github.com/matttproud/golang_protobuf_extensions
  Locking in  (e2704e1) for transitive dep github.com/rcrowley/go-metrics
  Locking in v1.1.0 (b4deda0) for transitive dep github.com/golang/protobuf
  Locking in v1.6.0 (9a47f48) for transitive dep github.com/rs/cors
  Locking in v1.8.0 (2fee6af) for transitive dep github.com/go-stack/stack
  Locking in v1.0.0 (853d788) for transitive dep github.com/jmhodges/levigo
  Locking in v0.14.1 (dc14acf) for transitive dep github.com/tendermint/go-amino
  Locking in v0.2.0 (cfeb6f9) for transitive dep github.com/prometheus/common
  Using ^0.30.1 as constraint for direct dep github.com/tendermint/tendermint
  Locking in v0.30.1 (e0f8936) for direct dep github.com/tendermint/tendermint
  Locking in v0.0.1 (2a8bb92) for transitive dep github.com/golang/snappy
  Locking in v1.19.0 (2fdaae2) for transitive dep google.golang.org/grpc
  Locking in  (292b43b) for transitive dep golang.org/x/net
  Locking in v0.8.1 (ba968bf) for transitive dep github.com/pkg/errors
  Locking in master (4f5b463) for transitive dep google.golang.org/genproto
  Locking in v1.1.1 (8991bc2) for transitive dep github.com/davecgh/go-spew
  Locking in v1.8.23 (c942700) for transitive dep github.com/ethereum/go-ethereum
  Locking in v0.6.0 (4dc7be5) for transitive dep github.com/go-kit/kit
  Locking in v0.4.0 (07c9b44) for transitive dep github.com/go-logfmt/logfmt
  Locking in v0.3.0 (f21a4df) for transitive dep golang.org/x/text
  Locking in v1.1.1 (636bf03) for transitive dep github.com/gogo/protobuf
  Locking in v0.9.2 (505eaef) for transitive dep github.com/prometheus/client_golang
  Locking in master (3a771d9) for transitive dep github.com/beorn7/perks
  Locking in v1.0.0 (9d007e4) for transitive dep github.com/syndtr/goleveldb
  Locking in master (30e92a1) for transitive dep golang.org/x/sys
  Locking in  (ed77733) for transitive dep github.com/btcsuite/btcd
  Locking in master (bbced96) for transitive dep github.com/prometheus/procfs
  Locking in master (b84e30a) for transitive dep github.com/kr/logfmt
```

^^ is what you get when you try to write a Go client. so many dependencies.

Liamsi:
@coriolinus any chance you could point us to your repo that imports `github.com/tendermint/tendermint/abci/types`. It would be great to fully understand what is going on and where the `release-branch.go1.12` comes from.

coriolinus:
I'm sorry, the code is all still private; having the repo URL wouldn't help
because you would get a 404 regardless.

Liamsi:
@coriolinus we completely dropped pinning to an exact version via https://github.com/tendermint/tendermint/pull/3382

Does that solve your problem?

melekes:
Given https://github.com/tendermint/tendermint/pull/3374#issuecomment-469571988 and https://github.com/tendermint/tendermint/pull/3382 (we're no longer pinning repos without releases to exact revisions), I am going to close this in favor of https://github.com/tendermint/tendermint/issues/3393. I don't think we're ready to move on with this right now.

PR #3376, opened 2019-03-04
Update to golang/1.12.0

Liamsi:
- update docker image on circleci
 - remove GOCACHE=off from Makefile (see: https://tip.golang.org/doc/go1.12#gocache)

* [ ] Wrote tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ For examples of the kinds of bugs we're looking for, see [SECURITY.md](SECURITY.
 
 Requirement|Notes
 ---|---
-Go version | Go1.12.0 or higher
+Go version | Go1.11.4 or higher
 
 ## Documentation
 
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Or [Blockchain](https://en.wikipedia.org/wiki/Blockchain_(database)), for short.
 [![API Reference](
 https://camo.githubusercontent.com/915b7be44ada53c290eb157634330494ebe3e30a/68747470733a2f2f676f646f632e6f72672f6769746875622e636f6d2f676f6c616e672f6764646f3f7374617475732e737667
 )](https://godoc.org/github.com/tendermint/tendermint)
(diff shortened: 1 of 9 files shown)


melekes:
Thanks 👍 Can you also update `https://img.shields.io/badge/go-1.10.4-blue.svg` in readme, scripts/install/*.sh files and Vagrantfile?

Liamsi:
Thanks! Will do now.

Liamsi:
There is also https://github.com/tendermint/tendermint/blob/e0f8936455029a40287a69d5b0e7baa4d5864da1/tools/build/Makefile#L7
which even contains references to deprecated (now private) repos: https://github.com/tendermint/tendermint/blob/e0f8936455029a40287a69d5b0e7baa4d5864da1/tools/build/Makefile#L79-L81

Liamsi:
And https://github.com/tendermint/tendermint/blob/e0f8936455029a40287a69d5b0e7baa4d5864da1/networks/remote/integration.sh#L13-L14

README.md
```diff
@@ -49,7 +49,7 @@ For examples of the kinds of bugs we're looking for, see [SECURITY.md](SECURITY.
 
 Requirement|Notes
 ---|---
-Go version | Go1.11.4 or higher
```
melekes: Do we require 1.12 though? I thought running with 1.11.4 is still ok

Liamsi: You are right. 1.11.4 is still OK. It's just we only test with 1.12 from this PR on.

melekes: The fact that we only test 1.12 should not force us to require 1.12.

Liamsi: OK, you are right. I can revert that change. Shouldn't we then still ensure that all tests etc. run on all versions we support (looks like currently we could run on 1.10 - 1.12 without any changes but we are only sure about the latest version we test on).

melekes: We do need to test `release-X` branches that they compile with all versions we support. But that's not a priority and can be done separately.

PR #3382, opened 2019-03-06
do not pin repos without releases to exact revisions

melekes:
We're pinning repos without releases because it's very easy to upgrade all the dependencies by executing `dep ensure --upgrade`. Instead, we should just never run this command directly, only `dep ensure --upgrade <some repo>`. And we can defend that in PRs.

Refs https://github.com/tendermint/tendermint/pull/3374

The problem with pinning to exact revisions: people who import Tendermint as a library (e.g. `abci/types`) are stuck with these revisions even though the code they import may not even use them.

* [ ] Wrote tests

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -76,34 +76,14 @@
   version = "^0.9.1"
 
 ###################################
-## Some repos dont have releases.
-## Pin to revision
+## Repos which don't have releases.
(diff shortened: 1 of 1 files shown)


melekes:
> I think github doesn't show it's contents by default.

😞

PR #3386, opened 2019-03-06
make ineffassign linter pass

melekes:
Refs #3262

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -718,7 +718,7 @@ func (cs *ConsensusState) handleMsg(mi msgInfo) {
 	}
 	if err != nil {
 		cs.Logger.Error("Error with msg", "height", cs.Height, "round", cs.Round,
-			"type", reflect.TypeOf(msg), "peer", peerID, "err", err, "msg", msg)
+			"peer", peerID, "err", err, "msg", msg)
 	}
 }
 
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -679,7 +679,7 @@ func (cs *ConsensusState) handleMsg(mi msgInfo) {
 	case *ProposalMessage:
 		// will not cause transition.
 		// once proposal is set, we can receive block parts
-		err = cs.setProposal(msg.Proposal)
+		_ = cs.setProposal(msg.Proposal)
 	case *BlockPartMessage:
 		// if the proposal is complete, we'll enterPrevote or tryFinalizeCommit
 		added, err = cs.addProposalBlockPart(msg, peerID)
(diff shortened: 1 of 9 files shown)


lite/dbprovider.go
```diff
@@ -258,14 +258,14 @@ func parseKey(key []byte) (chainID string, height int64, part string, ok bool) {
...
+	chainID, height, part, _ := parseKey(key)
 	if part != "sh" {
 		return "", 0, false
 	}
 	return chainID, height, true
 }
 
 func parseChainKeyPrefix(key []byte) (chainID string, height int64, ok bool) {
-	chainID, height, _, ok = parseKey(key)
+	chainID, height, _, _ = parseKey(key)
```
Liamsi: I'm surprised that we do not check `ok` here though. If `keyPattern.FindSubmatch(key)` doesn't match anything this will return `"", 0, "", false`.

Liamsi:
Restarted build. Previously failed: https://circleci.com/gh/tendermint/tendermint/49116#tests/containers/3

lite/dbprovider.go
melekes: Do you think we should return `ok` instead of `true`?

melekes:
wow 😨 test_cover failed 5 times in a row because of this error statement https://github.com/tendermint/tendermint/pull/3386/commits/140980047885d63efdbff3d257a8d0d46d7078fe

consensus/state.go
```diff
@@ -713,9 +716,10 @@ func (cs *ConsensusState) handleMsg(mi msgInfo) {
 	default:
 		cs.Logger.Error("Unknown msg type", reflect.TypeOf(msg))
 	}
-	if err != nil {
-		cs.Logger.Error("Error with msg", "height", cs.Height, "round", cs.Round, "type", reflect.TypeOf(msg), "peer", peerID, "err", err, "msg", msg)
-	}
+	// if err != nil {
+	// 	cs.Logger.Error("Error with msg", "height", cs.Height, "round", cs.Round,
```
melekes: uncommentting it will lead to TestReactorValidatorSetChanges failing with 5m0s timeout

lite/dbprovider.go
melekes: https://github.com/tendermint/tendermint/pull/3386/commits/77379e0cda71616da5cf32b25d44c1eceb80c869

consensus/state.go
melekes: @ebuchman any ideas why this may be the case?

Liamsi: `TestReactorValidatorSetChanges` uses a `TestingLoggerWithOutput` which inits a `NewSyncWriter`. Could it be the case that some other go-routine is writing to stdout such that this blocks and never becomes available? 

I could not reproduce this locally so not sure what is going on.

Liamsi: I removed the `-v` flag from go test, which triggerst the test to be run with a `NewNopLogger`.  Interestingly, the test passes (only tried once though): https://github.com/tendermint/tendermint/compare/develop...non-verbose/anton/ineffassign?expand=1

melekes: Maybe it has something to do with `reflect.TypeOf`

melekes: Can't replicate it anymore. Super weird

Liamsi: `reflect.TypeOf` doesn't block on anything. The logger potentially does as there is a lock involved (I have to admit I do not understand how and why but it seems more plausible than reflect).

PR #3387, opened 2019-03-07
Fix the reverse of meaning in spec doc

siburu:
PR just to fix one word in the doc.


docs/spec/blockchain/encoding.md
```diff
@@ -175,7 +175,7 @@ The differences between RFC 6962 and the simplest form a merkle tree are that:
    The leaf nodes are `SHA256(0x00 || leaf_data)`, and inner nodes are `SHA256(0x01 || left_hash || right_hash)`.
 
 2) When the number of items isn't a power of two, the left half of the tree is as big as it could be.
-   (The smallest power of two less than the number of items) This allows new leaves to be added with less
+   (The largest power of two less than the number of items) This allows new leaves to be added with less
```
Liamsi: https://tools.ietf.org/html/rfc6962#section-2.1
"The largest power of two smaller than the number of items" is actually correct!

melekes:

PR #3397, opened 2019-03-09
Close WriteBatch to prevent memory leak.

Stumble:
ClevelDB requires closing when WriteBatch is no longer needed, https://godoc.org/github.com/jmhodges/levigo#WriteBatch.Close


alexanderbez:
Looks pretty straightforward to me 🙂

PR #3401, opened 2019-03-11
cs: remove log.Error to avoid TestReactorValidatorSetChanges timing out

melekes:
See https://github.com/tendermint/tendermint/pull/3386#discussion_r263254903 for original discussion. Back then I thought #3386 has nothing to do with the TestReactorValidatorSetChanges test timeout, but now it's more or less clear. I still don't understand how/why log.Error statement is causing it. Note before #3386 we're not calling log.Error because of the shadowing.

[build_49553_step_105_container_3.txt.zip](https://github.com/tendermint/tendermint/files/2951514/build_49553_step_105_container_3.txt.zip)

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -671,8 +671,8 @@ func (cs *ConsensusState) handleMsg(mi msgInfo) {
 	defer cs.mtx.Unlock()
 
 	var (
-		err   error
 		added bool
+		err   error
 	)
 	msg, peerID := mi.Msg, mi.PeerID
 	switch msg := msg.(type) {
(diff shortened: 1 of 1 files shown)


consensus/state.go
```diff
@@ -670,31 +670,26 @@ func (cs *ConsensusState) handleMsg(mi msgInfo) {
...
-		err   error
-		added bool
-	)
 	msg, peerID := mi.Msg, mi.PeerID
 	switch msg := msg.(type) {
 	case *ProposalMessage:
 		// will not cause transition.
 		// once proposal is set, we can receive block parts
-		err = cs.setProposal(msg.Proposal)
+		cs.setProposal(msg.Proposal)
```
Liamsi: Add a comment / TODO about why we do not check the error here?

consensus/state.go
```diff
@@ -670,31 +670,26 @@ func (cs *ConsensusState) handleMsg(mi msgInfo) {
...
 
 		if err != nil && msg.Round != cs.Round {
 			cs.Logger.Debug("Received block part from wrong round", "height", cs.Height, "csRound", cs.Round, "blockRound", msg.Round)
-			err = nil
 		}
 	case *VoteMessage:
 		// attempt to add the vote and dupeout the validator if its a duplicate signature
 		// if the vote gives us a 2/3-any or 2/3-one, we transition
-		added, err = cs.tryAddVote(msg.Vote, peerID)
+		added, err := cs.tryAddVote(msg.Vote, peerID)
```
Liamsi: Add a comment that we basically ignore the error here, too?

melekes: Good point. Added in https://github.com/tendermint/tendermint/pull/3401/commits/9001464e7d624625cb77466903ea82fae1668761

PR #3403, opened 2019-03-11
remove TimeIotaMs from ABCI consensus params

melekes:


consensus/replay.go
```diff
@@ -324,7 +324,10 @@ func (h *Handshaker) ReplayBlocks(
 			}
 
 			if res.ConsensusParams != nil {
+				timeIotaMs := state.ConsensusParams.Block.TimeIotaMs
 				state.ConsensusParams = types.PB2TM.ConsensusParams(res.ConsensusParams)
+				// TimeIotaMs is not a part of ABCI consensus params
+				state.ConsensusParams.Block.TimeIotaMs = timeIotaMs
```
Liamsi: I did not verify that this is the only place where we do not want ABCI params to potentially overwrite `TimeIotaMs`.  Was this caught by a test?

consensus/replay.go
```diff
@@ -324,7 +324,10 @@ func (h *Handshaker) ReplayBlocks(
 			}
 
 			if res.ConsensusParams != nil {
+				timeIotaMs := state.ConsensusParams.Block.TimeIotaMs
 				state.ConsensusParams = types.PB2TM.ConsensusParams(res.ConsensusParams)
+				// TimeIotaMs is not a part of ABCI consensus params
```
Liamsi:
```suggestion
				// TimeIotaMs is not a part of ABCI consensus params but
				// of the actual state. Make sure it does not get overwritten.
```

consensus/replay.go
melekes: No, me executing

```
 rg "PB2TM.ConsensusParams"
consensus/replay.go
328:                            state.ConsensusParams = types.PB2TM.ConsensusParams(res.ConsensusParams)

types/protobuf_test.go
67:     cp2 := PB2TM.ConsensusParams(abciCP)
```

consensus/replay.go
```diff
@@ -324,10 +324,12 @@ func (h *Handshaker) ReplayBlocks(
 			}
 
 			if res.ConsensusParams != nil {
+				// TimeIotaMs is not exposed to the application.
+				// Make sure we preserve it.
```
Liamsi: 👍

consensus/replay.go
```diff
@@ -324,10 +324,12 @@ func (h *Handshaker) ReplayBlocks(
...
 				timeIotaMs := state.ConsensusParams.Block.TimeIotaMs
-				state.ConsensusParams = types.PB2TM.ConsensusParams(res.ConsensusParams)
-				// TimeIotaMs is not a part of ABCI consensus params but
-				// of the actual state. Make sure it does not get overwritten.
+				{
+					state.ConsensusParams = types.PB2TM.ConsensusParams(res.ConsensusParams)
+				}
```
Liamsi: Why the extra `{}`-brackets? The new scope doesn't introduce / separate anything.

melekes: Screenshot 2019-03-11 at 22 05 33

It separates reading from state.ConsensusParams and rewriting it with ABCI cs params

consensus/replay.go
```diff
@@ -324,7 +324,12 @@ func (h *Handshaker) ReplayBlocks(
 			}
 
 			if res.ConsensusParams != nil {
-				state.ConsensusParams = types.PB2TM.ConsensusParams(res.ConsensusParams)
+				// Preserve TimeIotaMs since it's not exposed to the application.
+				timeIotaMs := state.ConsensusParams.Block.TimeIotaMs
+				{
+					state.ConsensusParams = types.PB2TM.ConsensusParams(res.ConsensusParams)
```
ebuchman: I feel like a better approach would be to build this into the PB2TM.ConsensusParams method. Then we know we'll never forget to do this

ebuchman: Openned #3432

PR #3412, opened 2019-03-11
Merge master back to develop

melekes:


melekes:
> This means we currently do changelogs for minor/fix releases directly into master?

Yes. At least that is how it was done before.

melekes:
Can I squash this PR? Not sure

melekes:
https://github.com/tendermint/tendermint/pull/3339

Liamsi:
I think merging master to develop should not come with squashing:
image
The 3 commits from master won't make it to develop. Or do we want develop and master to diverge?

PR #3417, opened 2019-03-12
Release/v0.31.0

Liamsi:
- [x] do we need any upgrade instructions (prob. not)
 - [x] double check that the changelog is complete; merged without a changelog entry:
      - https://github.com/tendermint/tendermint/pull/3371 (add shoutout to @silasdavis in entry)
      - https://github.com/tendermint/tendermint/pull/3351
      - https://github.com/tendermint/tendermint/pull/3359
      - https://github.com/tendermint/tendermint/pull/3358
      - https://github.com/tendermint/tendermint/pull/3354
      - https://github.com/tendermint/tendermint/pull/3301
      - https://github.com/tendermint/tendermint/pull/3397
      - ... ?
 - [x] double check we did not forget any external contributor
 - [x] double check versions.go

https://github.com/tendermint/tendermint/blob/5483ac6b0a2a75bc140487e8405392533a3f1687/CHANGELOG.md#v0310

PR #3422, opened 2019-03-13
Minor changes / fixes to release 0.31.0

Liamsi:
See commit history

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,6 @@ Special thanks to external contributors on this release:
 - [libs/common] TrapSignal accepts logger as a first parameter and does not block anymore
   * previously it was dumping "captured ..." msg to os.Stdout
   * TrapSignal should not be responsible for blocking thread of execution
-- [privval] [\#3351](https://github.com/tendermint/tendermint/pull/3351) First part of larger refactoring that clarifies and separates concerns in the privval package.
 - [libs/db] [\#3397](https://github.com/tendermint/tendermint/pull/3397) Add possibility to `Close()` `Batch` to prevent memory leak when using ClevelDB. (@Stumble)   
 
 * Blockchain Protocol
@@ -46,6 +45,7 @@ Special thanks to external contributors on this release:
   - leveldb.openedtables
   - leveldb.alivesnaps
   - leveldb.aliveiters
+- [privval] [\#3351](https://github.com/tendermint/tendermint/pull/3351) First part of larger refactoring that clarifies and separates concerns in the privval package.
 
 ### BUG FIXES:
 - [p2p/conn] [\#3347](https://github.com/tendermint/tendermint/issues/3347) Reject all-zero shared secrets in the Diffie-Hellman step of secret-connection
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ Special thanks to external contributors on this release:
 
 * Apps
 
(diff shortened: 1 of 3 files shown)


UPGRADING.md
```diff
@@ -3,6 +3,24 @@
 This guide provides steps to be followed when you upgrade your applications to
 a newer version of Tendermint Core.
 
+## v0.31.0
+
+Since the pubsub no longer blocks on sending, some WS clients might stop working as expected. 
+If your WS client is not consuming events fast enough, Tendermint can terminate the subscription. 
+In this case, the WS client will receive an error with description:
```
Liamsi: ref: https://github.com/tendermint/tendermint/pull/3417#issuecomment-472285393

There is not much people can do about this. It is merely to inform people, right? @melekes

CHANGELOG.md
```diff
@@ -14,10 +14,14 @@ Special thanks to external contributors on this release:
 
 * Apps
 
+- [genesis] [\#2920](https://github.com/tendermint/tendermint/issues/2920) Rename `consensus_params.block_size` to `consensus_params.block` in ABCI ConsensusParams
+
 * Go API
 - [libs/common] TrapSignal accepts logger as a first parameter and does not block anymore
   * previously it was dumping "captured ..." msg to os.Stdout
   * TrapSignal should not be responsible for blocking thread of execution
+- [privval] [\#3351](https://github.com/tendermint/tendermint/pull/3351) First part of larger refactoring that clarifies and separates concerns in the privval package.
```
melekes: We can move it to improvements below. Even though it's breaking technically, we do not consider it "breaking" here

From readme
That said, breaking changes in the following packages will be documented in the
CHANGELOG even if they don't lead to MINOR version bumps:

- crypto
- types
- rpc/client
- config
- node
- libs
  - bech32
  - common
  - db
  - errors
  - log

Liamsi: Thanks! Done in https://github.com/tendermint/tendermint/pull/3422/commits/fc08a234ff78434ffca346777f04d59be14bd970

CHANGELOG.md
```diff
@@ -49,6 +52,8 @@ Special thanks to external contributors on this release:
 - [libs/pubsub] [\#951](https://github.com/tendermint/tendermint/issues/951), [\#1880](https://github.com/tendermint/tendermint/issues/1880) use non-blocking send when dispatching messages [ADR-33](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-033-pubsub.md)
 - [p2p] [\#3369](https://github.com/tendermint/tendermint/issues/3369) do not panic when filter times out
 - [cmd] [\#3408](https://github.com/tendermint/tendermint/issues/3408) Fix `testnet` command's panic when creating non-validator configs (using `--n` flag) (@srmo)
+- [blockchain] [\#3358](https://github.com/tendermint/tendermint/pull/3358) Fix timer leak in `BlockPool` (@guagualvcha)
+- [p2p] [\#3359](https://github.com/tendermint/tendermint/pull/3359) Fix order of events: start peer, then add it to `PeerSet`, and then start reactors on peer. This prevents receiving to error before actually starting or adding the peer to the set. (@guagualvcha)
```
melekes: Why not simply "Fix reconnecting report duplicate ID error due to race condition between adding peer to peerSet and starting it"?

UPGRADING.md
melekes: Right. Normally they should not see that error.

CHANGELOG.md
Liamsi: Thought it might be unclear what it means. But yeah, what I wrote is probably to verbose for a changelog entry.

PR #3424, opened 2019-03-13
grpcdb: close Iterator/ReverseIterator after use

### LINKED ISSUE #3402: Close Iterator in RemoteDB
https://github.com/tendermint/tendermint/blob/e0f8936455029a40287a69d5b0e7baa4d5864da1/libs/db/remotedb/grpcdb/server.go#L139-L161

Though this part of codes does not seem to be used for now, but in case it's used in the future, do we need to close the Iterator in `handleIterator`?

melekes:
Fixes #3402

* [ ] Wrote tests
  Maybe in the future. Code is unused right now.

### DIFF
--- a/libs/db/remotedb/grpcdb/server.go
+++ b/libs/db/remotedb/grpcdb/server.go
@@ -138,6 +138,7 @@ func (s *server) SetSync(ctx context.Context, in *protodb.Entity) (*protodb.Noth
 
 func (s *server) Iterator(query *protodb.Entity, dis protodb.DB_IteratorServer) error {
 	it := s.db.Iterator(query.Start, query.End)
+	defer it.Close()
 	return s.handleIterator(it, dis.Send)
 }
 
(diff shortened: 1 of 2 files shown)


libs/db/remotedb/grpcdb/server.go
```diff
@@ -162,6 +163,7 @@ func (s *server) handleIterator(it db.Iterator, sendFunc func(*protodb.Iterator)
 
 func (s *server) ReverseIterator(query *protodb.Entity, dis protodb.DB_ReverseIteratorServer) error {
 	it := s.db.ReverseIterator(query.Start, query.End)
+	defer it.Close()
 	return s.handleIterator(it, dis.Send)
```
Liamsi: Should we add a comment to `handleIterator` that the caller is responsible for closing the iter?

melekes: I think it should be obvious since it's passed as an argument, no?

Liamsi: Right. I’m just worried that we might forget to close it in other places that will call this (or other) functions.

melekes: Would be cool to have another program that checks this kind of stuff!

PR #3426, opened 2019-03-14
crypto: delete unused code

needkane:
* [ ] Wrote tests

### DIFF
--- a/crypto/hash.go
+++ b/crypto/hash.go
@@ -2,18 +2,10 @@ package crypto
 
 import (
 	"crypto/sha256"
-
-	"golang.org/x/crypto/ripemd160"
 )
 
 func Sha256(bytes []byte) []byte {
 	hasher := sha256.New()
-	hasher.Write(bytes)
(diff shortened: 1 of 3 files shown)


crypto/hash.go
```diff
@@ -2,18 +2,10 @@ package crypto
...
 	hasher := sha256.New()
 	hasher.Write(bytes)
 	return hasher.Sum(nil)
 }
-
-func Ripemd160(bytes []byte) []byte {
-	hasher := ripemd160.New()
-	hasher.Write(bytes)
-	return hasher.Sum(nil)
-}
```
Liamsi: This wrapper is indeed not used (besides in the example). We use ripemd160 in `PubKeySecp256k1.Address()` but  the wrapper, example code isn't useful (nor used).
https://github.com/tendermint/tendermint/blob/976819537d199aa43c6ad55ec4e8958323322bb0/crypto/secp256k1/secp256k1.go#L113-L121

PR #3430, opened 2019-03-16
rpc: client disable compression

guagualvcha:
By default, the http go client support Gzip-compressed Http response. 
The cli and light binary does not disable this support via the DisableCompression
property of the HTTPTransport object. As soon as it send a Http Request to a user-controlled
server, it become possible to cause a Denial of Service via a GZIP bomb.

Steps To Reproduce:
Code:
```
<?php
header("Content-Encoding: gzip");
echo file_get_contents("hello.gz")
?>
```

file： hello.gz
the file is 9MB in size but will expand to 9GB as soon as it decompressed.

start lightd or cli with --node to this server, the process will crash.
* [ ] Wrote tests

### DIFF
--- a/rpc/lib/client/http_client.go
+++ b/rpc/lib/client/http_client.go
@@ -74,7 +74,9 @@ func makeHTTPClient(remoteAddr string) (string, *http.Client) {
 	protocol, address, dialer := makeHTTPDialer(remoteAddr)
 	return protocol + "://" + address, &http.Client{
 		Transport: &http.Transport{
-			Dial: dialer,
+			// Set to true to prevent GZIP-bomb DoS attacks
+			DisableCompression: true,
+			Dial:               dialer,
 		},
 	}
 }


rpc/lib/client/http_client.go
```diff
@@ -74,6 +74,7 @@ func makeHTTPClient(remoteAddr string) (string, *http.Client) {
 	protocol, address, dialer := makeHTTPDialer(remoteAddr)
 	return protocol + "://" + address, &http.Client{
 		Transport: &http.Transport{
+			DisableCompression: true,
```
melekes: Could you please explain why we should not use a default here?

guagualvcha: sure. I have just finish writing the description of this pr. please have a check

melekes:
Thank you for discovering this 👍  It's a bit hard to believe Go's `http.Transport` is using an "unsafe" option by default. 

```
// Reference https://github.com/golang/go/blob/master/src/net/http/transport.go
//
// DisableCompression, if true, prevents the Transport from
// requesting compression with an "Accept-Encoding: gzip"
// request header when the Request contains no existing
// Accept-Encoding value. If the Transport requests gzip on
// its own and gets a gzipped response, it's transparently
// decoded in the Response.Body. However, if the user
// explicitly requested gzip it is not automatically
// uncompressed.
DisableCompression bool
```

If true and you add a header `Accept-Encoding: gzip` manually in the request, then you have to handle Gzip response body by yourself. 
If false and server responds with gzip file, client will attempt to decode by itself.

Note while our RPC client is vulnerable, Tendermint itself is not (because our server does not respond with gzip files / any files).

rpc/lib/client/http_client.go
melekes: We should add a comment above explaining why we set DisableCompression to true:

```suggestion
			// Set to true to prevent GZIP-bomb DoS attacks
			DisableCompression: true,
```

melekes:
Also, could you please format the code with `gofmt -s -w`?

guagualvcha:
> Also, could you please format the code with `gofmt -s -w`?

Thks for your suggestion, updated. please review again.

melekes:
Can you also add 

```
BUGFIXES:
- [rpc/lib/http_client] DisableCompression set to true to prevent DoS attacks using gzip bombs
```

to the CHANGELOG_PENDING.md?

guagualvcha:
> Can you also add
> 
> ```
> BUGFIXES:
> - [rpc/lib/http_client] DisableCompression set to true to prevent DoS attacks using gzip bombs
> ```
> to the CHANGELOG_PENDING.md?

sure

PR #3438, opened 2019-03-17
rpc: Fix response time grow over time

guagualvcha:
/validators rpc response time grow with time.

Steps To Reproduce:
1. start a new node from hight 1
2.`time curl 127.0.0.1:26657/validators`, it takes 0.005s
3. let the chain run for serveral days, and change no validators.in my scenario the height is : 1584840
4. `time curl 127.0.0.1:26857/validators`, it takes 1.25s

The api consumes cpu resouces, in the long run, a small scale of DDOS can make the node
a zombie process.

Acturally the logic to do `IncrementProposerPriority` by `height - valInfo.LastHeightChanged` times will not replay the exactly `ProposerPriority` once there are two or more round.

* [ ] Wrote tests

### DIFF
--- a/state/store.go
+++ b/state/store.go
@@ -9,6 +9,10 @@ import (
 	"github.com/tendermint/tendermint/types"
 )
 
+const (
+	ValidatorSetStoreInterval = 1 << 10
+)
+
 //------------------------------------------------------------------------
 
 func calcValidatorsKey(height int64) []byte {
@@ -182,19 +186,22 @@ func LoadValidators(db dbm.DB, height int64) (*types.ValidatorSet, error) {
 	if valInfo == nil {
 		return nil, ErrNoValSetForHeight{height}
 	}
-
 	if valInfo.ValidatorSet == nil {
-		valInfo2 := loadValidatorsInfo(db, valInfo.LastHeightChanged)
+		queryHeight := valInfo.LastHeightChanged
+		if valInfo.LastHeightChanged < (height - height%ValidatorSetStoreInterval) {
+			queryHeight = height - height%ValidatorSetStoreInterval
+		}
+		valInfo2 := loadValidatorsInfo(db, queryHeight)
 		if valInfo2 == nil {
 			panic(
 				fmt.Sprintf(
 					"Couldn't find validators at height %d as last changed from height %d",
-					valInfo.LastHeightChanged,
+					queryHeight,
 					height,
 				),
 			)
 		}
-		valInfo2.ValidatorSet.IncrementProposerPriority(int(height - valInfo.LastHeightChanged)) // mutate
+		valInfo2.ValidatorSet.IncrementProposerPriority(int(height - queryHeight)) // mutate
 		valInfo = valInfo2
 	}
 
@@ -232,7 +239,7 @@ func saveValidatorsInfo(db dbm.DB, height, lastHeightChanged int64, valSet *type
 	valInfo := &ValidatorsInfo{
 		LastHeightChanged: lastHeightChanged,
 	}
-	if lastHeightChanged == height {
+	if lastHeightChanged == height || height%ValidatorSetStoreInterval == 0 {
 		valInfo.ValidatorSet = valSet
 	}
 	db.Set(calcValidatorsKey(height), valInfo.Bytes())


state/store.go
```diff
@@ -187,15 +187,39 @@ func LoadValidators(db dbm.DB, height int64) (*types.ValidatorSet, error) {
...
-					"Couldn't find validators at height %d as last changed from height %d",
+				fmt.Errorf(
+					"couldn't find validators at height %d as last changed from height %d",
 					valInfo.LastHeightChanged,
 					height,
 				),
 			)
 		}
-		valInfo2.ValidatorSet.IncrementProposerPriority(int(height - valInfo.LastHeightChanged)) // mutate
 		valInfo = valInfo2
```
melekes: A much simpler solution here might be saving/memoizing valInfo2 `saveValidatorsInfo(db, height, valInfo2)` instead of the whole shebang with the blockstore. Thoughts?

guagualvcha: > A much simpler solution here might be saving/memoizing valInfo2 `saveValidatorsInfo(db, height, valInfo2)` instead of the whole shebang with the blockstore. Thoughts?

Ofcourse, if we can save all validatorInfo.  The previous implement only save whole latest validatorInfo .

guagualvcha: > A much simpler solution here might be saving/memoizing valInfo2 `saveValidatorsInfo(db, height, valInfo2)` instead of the whole shebang with the blockstore. Thoughts?

so should I modify code to `saveValidatorsInfo(db, height, valInfo2)` ?

state/store.go
```diff
@@ -182,22 +182,6 @@ func LoadValidators(db dbm.DB, height int64) (*types.ValidatorSet, error) {
...
-			panic(
-				fmt.Sprintf(
-					"Couldn't find validators at height %d as last changed from height %d",
-					valInfo.LastHeightChanged,
-					height,
-				),
-			)
-		}
-		valInfo2.ValidatorSet.IncrementProposerPriority(int(height - valInfo.LastHeightChanged)) // mutate
-		valInfo = valInfo2
```
melekes: I thought we'd save valInfo2 here, not save for every height! I.e. we only call save once somebody tries to access it.

guagualvcha: > I thought we'd save valInfo2 here, not save for every height! I.e. we only call save once somebody tries to access it.

But the replay result is not correct when there are two or more rounds at one height.

melekes: How so? If we're using `int(height - valInfo.LastHeightChanged)` difference

Liamsi: > But the replay result is not correct when there are two or more rounds at one height.

Is that because we call `IncrementProposerPriority` per round for the same height then? Can you give an example (ideally a test) which clearly shows what you mean here @guagualvcha? That would be awesome.

ebuchman: Note that the per-height proposer selection is *independent* of the per-round proposer selection. So the proposer at height=1 round=1 is the same as the proposer for height=2 round=0. Ie. see https://github.com/tendermint/tendermint/issues/3044

ebuchman:
Maybe we need a second implementation of the IncrementProposerPriority for when we know the validator set hasn't changed that can be computed much more quickly for this use case. cc @ancazamfir

guagualvcha:
> Maybe we need a second implementation of the IncrementProposerPriority for when we know the validator set hasn't changed that can be computed much more quickly for this use case. cc @ancazamfir

Great, that would be the final solution.

ancazamfir:
h1 - the height of previous validator set change
h - the height of the status query
h2 - the current height 

if h == h2 we could get the set from the state; else we have to do IncrementProposerPriority(h-h1) and I can't think of a much better way to compute it faster and still be accurate wrt proposer priorities.

Maybe something like this:
- call:
IncrementProposerPriority(int(height - valInfo.LastHeightChanged)%valInfo.TotalVotingPower())
- memoize the valInfo every time we call RescalePriorities(). This happens even without validator set changes although rarely.

May still be costly for high total voting power values. Alternatively we could save the valInfo every N blocks and then call Increment with max N times. 

> I thought we'd save valInfo2 here, not save for every height! I.e. we only call save once somebody tries to access it.

@melekes - could you please clarify? There may be a lot of accesses for different heights, isn't memory usage a concern?

guagualvcha:
> h1 - the height of previous validator set change
> h - the height of the status query
> h2 - the current height
> 
> if h == h2 we could get the set from the state; else we have to do IncrementProposerPriority(h-h1) and I can't think of a much better way to compute it faster and still be accurate wrt proposer priorities.
> 
> Maybe something like this:
> 
> * call:
>   IncrementProposerPriority(int(height - valInfo.LastHeightChanged)%valInfo.TotalVotingPower())
> * memoize the valInfo every time we call RescalePriorities(). This happens even without validator set changes although rarely.
> 
> May still be costly for high total voting power values. Alternatively we could save the valInfo every N blocks and then call Increment with max N times.
> 
> > I thought we'd save valInfo2 here, not save for every height! I.e. we only call save once somebody tries to access it.
> 
> @melekes - could you please clarify? There may be a lot of accesses for different heights, isn't memory usage a concern?

My idea: 1. we save  valInfo every N blocks ; 2. when this block get 2 or more round, save valInfo

guagualvcha:
> h1 - the height of previous validator set change
> h - the height of the status query
> h2 - the current height
> 
> if h == h2 we could get the set from the state; else we have to do IncrementProposerPriority(h-h1) and I can't think of a much better way to compute it faster and still be accurate wrt proposer priorities.
> 
> Maybe something like this:
> 
> * call:
>   IncrementProposerPriority(int(height - valInfo.LastHeightChanged)%valInfo.TotalVotingPower())
> * memoize the valInfo every time we call RescalePriorities(). This happens even without validator set changes although rarely.
> 
> May still be costly for high total voting power values. Alternatively we could save the valInfo every N blocks and then call Increment with max N times.
> 
> > I thought we'd save valInfo2 here, not save for every height! I.e. we only call save once somebody tries to access it.
> 
> @melekes - could you please clarify? There may be a lot of accesses for different heights, isn't memory usage a concern?

@melekes  @ancazamfir  updated as ancazamfir said. Please have a review.

melekes:
> could you please clarify? There may be a lot of accesses for different heights, isn't memory usage a concern?

yes, but not big though comparing to DoS attack vector.

melekes:
@guagualvcha please refrain from doing a rebase & force-push next time. When/if you do `git rebase` and force-push, 1) we can't see commit history 2) we can't review only new changes and forced to review everything every time. Thank you.

ancazamfir:
@guagualvcha Have you tested this? What is the response time now? What about with 1<<14 or 15?

melekes:
Closing in favor of https://github.com/tendermint/tendermint/pull/3537

sunboshan:
So with this fix, tendermint 0.31.4, how to get validators info?

```
$ curl 127.0.0.1:28221/validators
curl: (52) Empty reply from server
```

Issue #3435, opened 2019-03-17
TimeoutBroadcastTxCommit should overwrite the WriteTimeout limit


ebuchman:
Go servers only allow you to set a global WriteTimeout on the server, where the WriteTimeout starts after the request header is read, so it sets an absolute lifetime for the request.

It would be nicer if it started instead when we started writing the response, but now it can't be changed because Go v1 compatibility (see https://github.com/golang/go/issues/24461).

Worse, you can't set the WriteTimeout on a per request basis, only globally, since there's no way to access the underlying connection. There is some discussion in https://github.com/golang/go/issues/16100 but it's unclear if/when that will land.

This means the timeout for broadcast tx commit MUST be smaller than the configured WriteTimeout. This is currently hard coded at 20s. While the TimeoutBroadcastTxCommit is exposed in the config as of #3269, the WriteTimeout is not. We also throw an error if `TimeoutBroadcastTxCommit > WriteTimeout`. Instead, for now, we should just set `WriteTimeout = TimeoutBroadcastTxCommit + delta` so users can set the TimeoutBroadcastTxCommit as high as necessary, and the rpc WriteTimeout will just be equally as high, and we'll clearly document what's happening.

It's a bit silly to increase the global WriteTimeout just because you want a higher timeout for one endpoint (broadcast tx commit), but there isn't really a better alternative for now. Once https://github.com/golang/go/issues/16100 is implemented, then timeout_broadcast_tx_commit can be independent of WriteTimeout, and as large as the user wants.

melekes:
Are there any security implications? Should we inform the user about that?

Liamsi:
Just to be clear: do we want to expose `WriteTimeout` as a config (and set it depending on `TimeoutBroadcastTxCommit`), or do we just want to keep the 20 sec default and only set it to `WriteTimeout = TimeoutBroadcastTxCommit + delta` only if the `TimeoutBroadcastTxCommit` exceeds 20s?

ebuchman:
I think maybe the simplest thing for now is to just expose TimeoutBroadcastTxCommit in the config, give it a default of `10s`, and always set the `WriteTimeout = TimeoutBroadcastTxCommit + 2s`

Liamsi:
[...]

Issue #3329, opened 2019-02-19
rpcserver WriteTimeout configurable


tanhuiya:
RT

melekes:
https://github.com/tendermint/tendermint/blob/d2c7f8dbcf238ccc3a2469e350523c7ed28be154/rpc/lib/server/http_server.go#L38

melekes:
Could you please explain why you want to control these? Just curious. Do you use http_server library separately from Tendermint?

ebuchman:
As of #3443 you can increase it by setting the new `rpc.timeout_broadcast_tx_commit` value. Do you need more direct control than that?

Maybe we should just expose the read and write timeout for the global HTTP server directly through the config.

melekes:
Fixed on master https://github.com/tendermint/tendermint/blob/940e68292cd2e710525c35df08d7e007a7c04f9d/rpc/jsonrpc/server/http_server.go#L58

PR #3443, opened 2019-03-18
Ensure WriteTimeout > TimeoutBroadcastTxCommit

Liamsi:
ref #3435 

* [ ] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,52 +1,21 @@
-## v0.31.0
+## v0.32.0
 
 **
-
-Special thanks to external contributors on this release:
-@srmo
 
 ### BREAKING CHANGES:
 
 * CLI/RPC/Config
-- [rpc/client] Update Subscribe interface to reflect new pubsub/eventBus API [ADR-33](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-033-pubsub.md)
 
 * Apps
 
 * Go API
-- [libs/common] TrapSignal accepts logger as a first parameter and does not block anymore
-  * previously it was dumping "captured ..." msg to os.Stdout
-  * TrapSignal should not be responsible for blocking thread of execution
 
 * Blockchain Protocol
 
 * P2P Protocol
 
 ### FEATURES:
-- [mempool] \#3079 bound mempool memory usage (`mempool.max_txs_bytes` is set to 1GB by default; see config.toml)
-  mempool's current `txs_total_bytes` is exposed via `total_bytes` field in
-  `/num_unconfirmed_txs` and `/unconfirmed_txs` RPC endpoints.
-- [config] \#2920 Remove `consensus.blocktime_iota` parameter
-- [genesis] \#2920 Add `time_iota_ms` to block's consensus parameters (not exposed to the application)
-- [genesis] \#2920 Rename `consensus_params.block_size` to `consensus_params.block`
-- [lite] add `/unsubscribe_all` endpoint, which allows you to unsubscribe from all events
-- [rpc] `StartHTTPServer` / `StartHTTPAndTLSServer` now require a Config (use `rpcserver.DefaultConfig`)
 
 ### IMPROVEMENTS:
-- [libs/common] \#3238 exit with zero (0) code upon receiving SIGTERM/SIGINT
-- [libs/db] \#3378 CLevelDB#Stats now returns the following properties:
-  - leveldb.num-files-at-level{n}
-  - leveldb.stats
-  - leveldb.sstables
-  - leveldb.blockpool
-  - leveldb.cachedblock
-  - leveldb.openedtables
-  - leveldb.alivesnaps
-  - leveldb.aliveiters
-- [rpc] \#3435 setting `TimeoutBroadcastTxCommit` may increase the HTTP global write timeout
 
 ### BUG FIXES:
-- [p2p/conn] \#3347 Reject all-zero shared secrets in the Diffie-Hellman step of secret-connection
-- [libs/pubsub] \#951, \#1880 use non-blocking send when dispatching messages [ADR-33](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-033-pubsub.md)
-- [p2p] \#3369 do not panic when filter times out
-- [cmd] \#3408 Fix `testnet` command's panic when creating non-validator configs (using `--n` flag) (@srmo)
-- [libs/db/remotedb/grpcdb] \#3402 Close Iterator/ReverseIterator after use
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -29,7 +29,7 @@ Special thanks to external contributors on this release:
 - [genesis] \#2920 Add `time_iota_ms` to block's consensus parameters (not exposed to the application)
 - [genesis] \#2920 Rename `consensus_params.block_size` to `consensus_params.block`
 - [lite] add `/unsubscribe_all` endpoint, which allows you to unsubscribe from all events
-- [rpc] StartHTTPServer / StartHTTPAndTLSServer now require a Config (use rpcserver.DefaultConfig)
+- [rpc] `StartHTTPServer` / `StartHTTPAndTLSServer` now require a Config (use `rpcserver.DefaultConfig`)
 
 ### IMPROVEMENTS:
 - [libs/common] \#3238 exit with zero (0) code upon receiving SIGTERM/SIGINT
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -29,6 +29,7 @@ Special thanks to external contributors on this release:
 - [genesis] \#2920 Add `time_iota_ms` to block's consensus parameters (not exposed to the application)
 - [genesis] \#2920 Rename `consensus_params.block_size` to `consensus_params.block`
 - [lite] add `/unsubscribe_all` endpoint, which allows you to unsubscribe from all events
+- [rpc] StartHTTPServer / StartHTTPAndTLSServer now require a Config (use rpcserver.DefaultConfig)
 
 ### IMPROVEMENTS:
 - [libs/common] \#3238 exit with zero (0) code upon receiving SIGTERM/SIGINT
@@ -41,6 +42,7 @@ Special thanks to external contributors on this release:
   - leveldb.openedtables
   - leveldb.alivesnaps
   - leveldb.aliveiters
+- [rpc] \#3435 setting TimeoutBroadcastTxCommit may increase the HTTP global write timeout
 
 ### BUG FIXES:
 - [p2p/conn] \#3347 Reject all-zero shared secrets in the Diffie-Hellman step of secret-connection
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -42,7 +42,7 @@ Special thanks to external contributors on this release:
   - leveldb.openedtables
   - leveldb.alivesnaps
   - leveldb.aliveiters
-- [rpc] \#3435 setting TimeoutBroadcastTxCommit may increase the HTTP global write timeout
+- [rpc] \#3435 setting `TimeoutBroadcastTxCommit` may increase the HTTP global write timeout
 
 ### BUG FIXES:
 - [p2p/conn] \#3347 Reject all-zero shared secrets in the Diffie-Hellman step of secret-connection
--- a/config/config.go
+++ b/config/config.go
@@ -335,6 +335,9 @@ type RPCConfig struct {
 	MaxSubscriptionsPerClient int `mapstructure:"max_subscriptions_per_client"`
 
 	// How long to wait for a tx to be committed during /broadcast_tx_commit
+	// WARNING: Using a value larger than 10s will result in increasing the
+	// global HTTP write timeout, which applies to all connections and endpoints.
+	// See https://github.com/tendermint/tendermint/issues/3435
 	TimeoutBroadcastTxCommit time.Duration `mapstructure:"timeout_broadcast_tx_commit"`
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	rpcserver "github.com/tendermint/tendermint/rpc/lib/server"
 )
 
 const (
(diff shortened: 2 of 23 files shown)


rpc/lib/server/http_server.go
```diff
@@ -31,15 +32,41 @@ const (
...
 	// TODO: use a config instead.
-	ReadTimeout  = 3 * time.Second
-	WriteTimeout = 20 * time.Second
+	ReadTimeout = 3 * time.Second
 )
 
+var (
+	// Timeout for writing to the http connection.
+	// Use the WriteTimeout() function to read this.
+	writeTimeout = 20 * time.Second
```
Liamsi: Instead of having a global var here that gets adjusted, there could be a default (const) and we could pass in a `time.Duration` (TimeoutBroadcastTxCommit) into `StartHTTPServer`. If the passed time.Duration is bigger than the default the param passed into `http.Server` gets adjusted. That should be cleaner from what we do here. The problem with this approach is that we read the `WriteTimeout` in `rpc/core/mempool.go` and `rpc/core/events.go`. Also the comment suggests that handlers might want to read the value, too.

rpc/lib/server/http_server.go
```diff
@@ -31,15 +32,38 @@ const (
...
+)
+
+// AdjustWriteTimeout adjusts the timeout for writing to the http connection
+// if necessary.
+// That is if and only if the passed in RPCConfig contains
+// a TimeoutBroadcastTxCommit (used by /broadcast_tx_commit) which is larger
+// than the WriteTimeout.
+func AdjustWriteTimeout(cfg config.RPCConfig) (changed bool) {
```
melekes: let's move this function to node.go

rpc/lib/server/http_server.go
```diff
@@ -48,7 +72,7 @@ func StartHTTPServer(listener net.Listener, handler http.Handler, logger log.Log
 	s := &http.Server{
```
melekes: how about we refactor StartHTTPServer to accept config struct

```go
type config struct {
  WriteTimeout int
  ReadTimeout int
}
```

and return http.Server, which we'll use in RPC to get Timeout value. I don't really like global variables

rpc/core/pipe.go
```diff
@@ -137,7 +135,11 @@ func SetEventBus(b *types.EventBus) {
 }
 
 // SetConfig sets an RPCConfig.
+// If necessary it adjusts (increases) the WriteTimeout for writing to the http
+// connection to ensure rpcserver.WriteTimeout() > c.TimeoutBroadcastTxCommit.
+// For details see: https://github.com/tendermint/tendermint/issues/3435
 func SetConfig(c cfg.RPCConfig) {
+	rpcserver.AdjustWriteTimeout(c)
```
melekes: move to node.go

melekes:
Refs https://github.com/tendermint/tendermint/issues/3329

rpc/lib/server/http_server.go
Liamsi: I thought about this too. But this would come at the cost of more refactoring: 

`StartHTTPServer` in its current form blocks (does no return) as it calls `http.Server.Serve`. 
Sure, we could change that. Then, we would need to rename the method to `NewHTTPServer` as it does not start anything anymore and update all places to manually call `Serve`. Same holds for `StartHTTPAndTLSServer`. 

> which we'll use in RPC to get Timeout value

This assumes the places were we need to read the timeout value have access to the returned `http.Server` (meaning they call `NewHTTPServer`). Also, not the case currently. 

I'd vouch to keep this in mind and open an issue about a larger refactoring here. It's also likely, there are more places in the rpc package like this (there are quite a few gloabal vars).

rpc/lib/server/http_server.go
Liamsi: `AdjustWriteTimeout` needs to modify the writeTimeout. If we moved the function to node.go, we would need to make the `writeTimeout` a public global var again which isn't `const` (even worse than a global var that only the package controls). Otherwise the function can't modify the var. We shouldn't move the var over to node.go (because we read the value in the rpc package and importing the node package seems the wrong way around).

rpc/core/pipe.go
Liamsi: See https://github.com/tendermint/tendermint/pull/3443#discussion_r266789988

If instead you mean we should move the call of `rpcserver.AdjustWriteTimeout` out of here and into the node package / node.go, we might miss other places which set the config for the rpc package (including the `TimeoutBroadcastTxCommit` which this PR is all about).

rpc/lib/server/http_server.go
melekes: > StartHTTPServer in its current form blocks 

oh, right. forgot about that.

but we can still pass a config struct

Liamsi: We could indeed but what would we gain from that? Currently the values are read from a global var and directly shared with other places via that global vars (I agree, this is very bad). This PR makes one of these global vars unexported s.t. that the package is "in control". Having a config as you described makes sense but how do we then ensure that increased the writeTimeout if necessary (`WriteTimeout > TimeoutBroadcastTxCommit`)? Using the adjusted value from the global var in the passed config? It gets even more confusing IMHO.

rpc/core/events.go
```diff
@@ -106,9 +105,7 @@ func Subscribe(ctx *rpctypes.Context, query string) (*ctypes.ResultSubscribe, er
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to parse query")
 	}
-	subscribeTimeout := rpcserver.WriteTimeout() / 2
-	subCtx, cancel := context.WithTimeout(context.Background(), subscribeTimeout)
-	defer cancel()
+	subCtx := context.TODO()
```
melekes: TODO: get ctx from ctx *rpctypes.Context

rpc/core/mempool.go
```diff
@@ -198,9 +197,7 @@ func BroadcastTxCommit(ctx *rpctypes.Context, tx types.Tx) (*ctypes.ResultBroadc
 	}
 
 	// Subscribe to tx being committed in block.
-	subscribeTimeout := rpcserver.WriteTimeout() / 2
-	subCtx, cancel := context.WithTimeout(context.Background(), subscribeTimeout)
-	defer cancel()
+	subCtx := context.TODO()
```
melekes: TODO: get ctx from ctx *rpctypes.Context

rpc/lib/server/http_server.go
```diff
@@ -31,48 +44,17 @@ const (
...
-	// Use the WriteTimeout() function to read this.
-	writeTimeout = 20 * time.Second
 )
 
-// AdjustWriteTimeout adjusts the timeout for writing to the http connection
-// if necessary.
-// That is if and only if the passed in RPCConfig contains
-// a TimeoutBroadcastTxCommit (used by /broadcast_tx_commit) which is larger
-// than the WriteTimeout.
-func AdjustWriteTimeout(cfg config.RPCConfig) (changed bool) {
```
melekes: low level library should not know anything about high level clients

rpc/lib/server/http_server.go
melekes: https://github.com/tendermint/tendermint/pull/3443/commits/811d92c1bd4d32b37162ee5eccab6348f9ba5e5d#r266826680

rpc/core/events.go
Liamsi: So the subscribeTimeout isn't absolutely necessary here? That simplifies this a lot ...

rpc/core/pipe.go
melekes: I understand, but I think it's fine to expect the config to be correct until TimeoutBroadcastTxCommit is further resolved

node/node.go
```diff
@@ -689,9 +689,18 @@ func (n *Node) startRPC() ([]net.Listener, error) {
 		mux.HandleFunc("/websocket", wm.WebsocketHandler)
 		rpcserver.RegisterRPCFuncs(mux, rpccore.Routes, coreCodec, rpcLogger)
 
+		config := rpcserver.DefaultConfig()
+		config.MaxOpenConnections = n.config.RPC.MaxOpenConnections
+		// If necessary adjust global WriteTimeout to ensure it's greater than
+		// TimeoutBroadcastTxCommit.
+		// See https://github.com/tendermint/tendermint/issues/3435
+		if config.WriteTimeout <= n.config.RPC.TimeoutBroadcastTxCommit {
```
Liamsi: This currently will be true by default `WriteTimeout` is 3 seconds and `TimeoutBroadcastTxCommit` defaults to 10 seconds. I think the defaults should be adjusted to reflect the assumption that it is larger.

rpc/lib/server/http_server.go
```diff
@@ -14,14 +14,27 @@ import (
...
+	// mirrors http.Server#WriteTimeout
+	WriteTimeout time.Duration
+}
+
+// DefaultConfig returns a default configuration.
+func DefaultConfig() *Config {
+	return &Config{
+		MaxOpenConnections: 0, // unlimited
+		ReadTimeout:        3 * time.Second,
+		WriteTimeout:       3 * time.Second,
```
melekes: lowered from 20 to 3 seconds. Why it was 20 seconds in the first place?

rpc/lib/server/http_server.go
```diff
@@ -18,9 +18,23 @@ import (
...
```
Liamsi: This default should be larger than the default for `TimeoutBroadcastTxCommit` which is 10s.
See: https://github.com/tendermint/tendermint/blob/4162ebe8b586deccc0e7476d8abafb75138bfe58/config/config.go#L357-L357

rpc/lib/server/http_server.go
Liamsi: Makes sense 👍  But make sure the default of WriteTimeout > TimeoutBroadcastTxCommit (which is 10 s; see: https://github.com/tendermint/tendermint/pull/3443#discussion_r266830038).

config/config.go
```diff
@@ -335,6 +335,9 @@ type RPCConfig struct {
 	MaxSubscriptionsPerClient int `mapstructure:"max_subscriptions_per_client"`
 
 	// How long to wait for a tx to be committed during /broadcast_tx_commit
+	// WARNING: Using a value larger than 10s will result in increasing the
+	// global HTTP write timeout, which applies to all connections and endpoints.
+	// See https://github.com/tendermint/tendermint/issues/3435
```
Liamsi: 👍

Liamsi:
So the only question remains how to deal with https://github.com/tendermint/tendermint/pull/3443#discussion_r266826422
Is this sth. we want do in this PR or separately? Besides that question your changes look good to me :-)

rpc/lib/server/http_server.go
melekes: Set to slightly modified Nginx defaults.

rpc/core/events.go
melekes: No, looks like we need this timeout if we want to properly respond to the client's request.

melekes: https://github.com/tendermint/tendermint/pull/3443/commits/db0986e8ebc7fd4f1190d083227d11d9df97f011

rpc/core/events.go
```diff
@@ -105,8 +106,8 @@ func Subscribe(ctx *rpctypes.Context, query string) (*ctypes.ResultSubscribe, er
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to parse query")
 	}
-
-	subCtx, cancel := context.WithTimeout(context.Background(), subscribeTimeout)
+	// must be less than the server's write timeout (see rpcserver.DefaultConfig)
+	subCtx, cancel := context.WithTimeout(ctx.Context(), 5*time.Second)
```
Liamsi: What do you think about an exported (but documented) const `const SubscriptionTimeoutSec := 5` which you use here and in `rpc/core/mempool.go`. That way this will show up in the godoc and as a global documented constant?

melekes: I don't quite understand the benefit of doing this

Liamsi: More exposure on what callers can expect / how this behaves? Also if this every changes you only need to modify on place (the constant).

Liamsi:
Can't approve this as I've opened the PR but the latest changes made by @melekes LGTM 👍 Thanks for taking over Anton 🚀

CHANGELOG_PENDING.md
```diff
@@ -29,6 +29,7 @@ Special thanks to external contributors on this release:
 - [genesis] \#2920 Add `time_iota_ms` to block's consensus parameters (not exposed to the application)
 - [genesis] \#2920 Rename `consensus_params.block_size` to `consensus_params.block`
 - [lite] add `/unsubscribe_all` endpoint, which allows you to unsubscribe from all events
+- [rpc] StartHTTPServer / StartHTTPAndTLSServer now require a Config (use rpcserver.DefaultConfig)
```
```suggestion
- [rpc] `StartHTTPServer` / `StartHTTPAndTLSServer` now require a Config (use `rpcserver.DefaultConfig`)
```

CHANGELOG_PENDING.md
```diff
@@ -41,6 +42,7 @@ Special thanks to external contributors on this release:
   - leveldb.openedtables
   - leveldb.alivesnaps
   - leveldb.aliveiters
+- [rpc] \#3435 setting TimeoutBroadcastTxCommit may increase the HTTP global write timeout
```
```suggestion
- [rpc] \#3435 setting `TimeoutBroadcastTxCommit` may increase the HTTP global write timeout
```

Liamsi: Note: We might need to add a note to Upgrade.md? But this can happen in #3417

rpc/lib/server/http_server.go
```diff
@@ -18,9 +18,23 @@ import (
...
+	ReadTimeout time.Duration
+	// mirrors http.Server#WriteTimeout
+	WriteTimeout time.Duration
+}
+
+// DefaultConfig returns a default configuration.
+func DefaultConfig() *Config {
+	return &Config{
+		MaxOpenConnections: 0, // unlimited
+		ReadTimeout:        12 * time.Second,
```
ebuchman: Did we need to increase this from 3? 

Also we use 12 for the write timeout but refer to it in comments as 10. So I'll change it to 10 ...

ebuchman:
Also going to retarget this to the release branch ...

rpc/lib/server/http_server.go
Liamsi: No, there was no reason to increase the `ReadTimeout`. That must have slipped (only the write timeout needed to be changed back from 3 to sth. larger than TimeoutBroadcastTxCommit).

Note that while the `WriteTimeout` will be 10 with your changes as mentioned in the comments, it will effectively be 11 now via: https://github.com/tendermint/tendermint/pull/3443/files#diff-380e79adccbaf1276bf6c91cc551cd7fR698

ebuchman: Yeh that's fine, so long as the comment is correct (ie. that if broadcast tx is greater than 10 it will cause an increase)

ebuchman: The more I think about it the more I think we should probably expose the Read/WriteTimeout directly...

melekes: > The more I think about it the more I think we should probably expose the Read/WriteTimeout directly...

My refactoring should allow us to do that very easily.

PR #3446, opened 2019-03-19
consensus: call reconstructLastCommit after updateToState in SwitchToConsensus

iphucc:
Reasons:

- otherwise `cs.LastCommit` will be `nil`???
- to align it with `NewConsensusState`

* [ ] Wrote tests

### DIFF
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -100,10 +100,10 @@ func (conR *ConsensusReactor) OnStop() {
 // It resets the state, turns off fast_sync, and starts the consensus state-machine
 func (conR *ConsensusReactor) SwitchToConsensus(state sm.State, blocksSynced int) {
 	conR.Logger.Info("SwitchToConsensus")
-	conR.conS.reconstructLastCommit(state)
 	// NOTE: The line below causes broadcastNewRoundStepRoutine() to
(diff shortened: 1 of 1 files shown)


iphucc:
fix clear the cs.LastCommit by updateToState().

melekes:
> Is this OK? Otherwise, the LastCommit will be nil.

Could you elaborate on that? What were the conditions, which have lead to LastCommit being `nil`?

iphucc:
It seems that there is no need to modify this here（I am not completely sure）. Is there a problem with changing their order? Their order in NewConsensusState is different from here.

melekes:
> otherwise cs.LastCommit will be nil???

This part I don't understand. We're passing the same `state`, so the order does not seem to be important. No?

iphucc:
> > otherwise cs.LastCommit will be nil???
> 
> This part I don't understand. We're passing the same `state`, so the order does not seem to be important.I still don't fully understand the problem I encountered.

iphucc:
> > otherwise cs.LastCommit will be nil???
> 
> This part I don't understand. We're passing the same `state`, so the order does not seem to be important. No?

I still don't fully understand the problem I encountered. So, I will close this pull-request. Sorry!

PR #3448, opened 2019-03-19
replace PB2TM.ConsensusParams with a call to params#Update

### LINKED ISSUE #3444: Issue in updating ConsensusParams
The docs read:

> Each field that is not empty will be applied in full. For instance, if updating the BlockSize.MaxBytes, applications must also set the other BlockSize fields (like BlockSize.MaxGas), even if they are unchanged, as they will otherwise cause the value to be updated to 0.

But in the code, we do this: 

https://github.com/tendermint/tendermint/blob/745713330736c5c751450245d88b8037cbee3aa6/types/protobuf.go#L224-L251

Which means all the non-specified fields will be over written. 

Also, we're still referring to BlockSize, which should be Block as of v0.31.0

melekes:
Fixes #3444

### DIFF
--- a/consensus/replay.go
+++ b/consensus/replay.go
@@ -324,7 +324,7 @@ func (h *Handshaker) ReplayBlocks(
 			}
 
 			if res.ConsensusParams != nil {
-				state.ConsensusParams = types.PB2TM.ConsensusParams(res.ConsensusParams, state.ConsensusParams.Block.TimeIotaMs)
+				state.ConsensusParams = state.ConsensusParams.Update(res.ConsensusParams)
 			}
 			sm.SaveState(h.stateDB, state)
 		}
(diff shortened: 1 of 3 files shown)


consensus/replay.go
```diff
@@ -324,7 +324,7 @@ func (h *Handshaker) ReplayBlocks(
 			}
 
 			if res.ConsensusParams != nil {
-				state.ConsensusParams = types.PB2TM.ConsensusParams(res.ConsensusParams, state.ConsensusParams.Block.TimeIotaMs)
+				state.ConsensusParams = state.ConsensusParams.Update(res.ConsensusParams)
```
Liamsi: Interesting, this works the other way now: `state.ConsensusParams.Update` updates everything (or overwrites every filed if necessary)  *besides* blockTimeIotaMs. This comes with the neat side effect that we can delete `pb2tm.ConsensusParams(...)` 👍

PR #3453, opened 2019-03-20
crypto: update Sha256 defined by FIPS-202

needkane:
* [ ] Wrote tests

### DIFF
--- a/crypto/hash.go
+++ b/crypto/hash.go
@@ -1,11 +1,9 @@
 package crypto
 
-import (
-	"crypto/sha256"
-)
+import "golang.org/x/crypto/sha3"
 
 func Sha256(bytes []byte) []byte {
-	hasher := sha256.New()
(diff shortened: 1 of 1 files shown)


needkane:
golang package（crypto/sha3） defined by FIPS 180-4

crypto/hash.go
```diff
@@ -1,11 +1,9 @@
 package crypto
 
-import (
-	"crypto/sha256"
-)
+import "golang.org/x/crypto/sha3"
 
 func Sha256(bytes []byte) []byte {
-	hasher := sha256.New()
+	hasher := sha3.NewLegacyKeccak256()
```
Liamsi: Thanks @needkane. If anything we would include sha3 in a separate method / separate wrapper and not inside the sha2 wrapper. While this looks like an innocent change this needs justification in an issue where we can discuss the need and the implications of this change.

Issue #3156, opened 2019-01-18
Improve tests for CreateProposalBlock


ebuchman:
Follow up to https://github.com/tendermint/tendermint/issues/2806, see the tests in https://github.com/tendermint/tendermint/blob/07263298bd76f1ea7ad82f8e06234a450150d76c/node/node_test.go#L202

Improvements to be made include:
- use real evidence for different blockIDs instead of introducing this new mock evidence
- ensure the Commit and Header are at maximum size
- confirm this would have caught #2789 
- tests with more variability in the block structure

PR #3456, opened 2019-03-20
Refactor node.NewNode to simplify

thanethomson:
The `node.NewNode` method is pretty complex at the moment, an in order to address issues like #3156, we need to simplify the interface for partial node instantiation. In some places, we don't need to build up a full node (like in the `node.TestCreateProposalBlock` test), but the complexity of such partial instantiation needs to be reduced.

This PR aims to eventually make this easier/simpler.

See also [this gist](https://gist.github.com/thanethomson/56e1640d057a26186e38ad678a1d114c) for some background work done when starting to refactor here.

* [ ] Wrote tests


melekes:
What's left to do here?

thanethomson:
Ideally I wanted to find a way to better group the inner construction of the node to simplify it further (perhaps into higher-level constructs), but that's proven to be quite difficult. This might be the best I can do for now.

node/node.go
```diff
@@ -170,73 +172,64 @@ type Node struct {
...
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
+	return blockStore, stateDB, nil
+}
 
+func loadStateFromDBOrGenesisDoc(stateDB dbm.DB, genesisDocProvider GenesisDocProvider) (sm.State, *types.GenesisDoc, error) {
 	// Get genesis doc
 	// TODO: move to state package?
```
melekes: can we move this to `state/store.go`?

node/node.go
```diff
@@ -170,73 +172,64 @@ type Node struct {
...
-	privValidator types.PrivValidator,
-	nodeKey *p2p.NodeKey,
-	clientCreator proxy.ClientCreator,
-	genesisDocProvider GenesisDocProvider,
-	dbProvider DBProvider,
-	metricsProvider MetricsProvider,
-	logger log.Logger) (*Node, error) {
-
-	// Get BlockStore
+func getStoresForNewNode(config *cfg.Config, dbProvider DBProvider) (*bc.BlockStore, dbm.DB, error) {
```
melekes: it's unclear what the second return param is.

```suggestion
func initDBs(config *cfg.Config, dbProvider DBProvider) (blockStoreDB *bc.BlockStore, stateDB dbm.DB, err error) {
```

node/node.go
```diff
@@ -170,73 +172,64 @@ type Node struct {
...
 	}
+	return state, genDoc, nil
+}
 
-	// Create the proxyApp and establish connections to the ABCI app (consensus, mempool, query).
+func createAndStartProxyAppConns(clientCreator proxy.ClientCreator, logger log.Logger) (proxy.AppConns, error) {
 	proxyApp := proxy.NewAppConns(clientCreator)
 	proxyApp.SetLogger(logger.With("module", "proxy"))
 	if err := proxyApp.Start(); err != nil {
 		return nil, fmt.Errorf("Error starting proxy app connections: %v", err)
```
```suggestion
		return nil, fmt.Errorf("error starting proxy app connections: %v", err)
```

node/node.go
```diff
@@ -251,27 +244,23 @@ func NewNode(config *cfg.Config,
...
-	// and replays any blocks as necessary to sync tendermint with the app.
-	consensusLogger := logger.With("module", "consensus")
+func doHandshakeForNewNode(stateDB dbm.DB, state sm.State, blockStore sm.BlockStore, genDoc *types.GenesisDoc, eventBus *types.EventBus, proxyApp proxy.AppConns, logger log.Logger) error {
 	handshaker := cs.NewHandshaker(stateDB, state, blockStore, genDoc)
-	handshaker.SetLogger(consensusLogger)
+	handshaker.SetLogger(logger)
 	handshaker.SetEventBus(eventBus)
 	if err := handshaker.Handshake(proxyApp); err != nil {
-		return nil, fmt.Errorf("Error during handshake: %v", err)
+		return fmt.Errorf("Error during handshake: %v", err)
```
```suggestion
		return fmt.Errorf("error during handshake: %v", err)
```

node/node.go
```diff
@@ -307,19 +307,10 @@ func NewNode(config *cfg.Config,
...
-	if state.Validators.HasAddress(addr) {
-		consensusLogger.Info("This node is a validator", "addr", addr, "pubKey", pubKey)
-	} else {
-		consensusLogger.Info("This node is not a validator", "addr", addr, "pubKey", pubKey)
-	}
-
-	csMetrics, p2pMetrics, memplMetrics, smMetrics := metricsProvider(genDoc.ChainID)
-
-	// Make MempoolReactor
+func buildNewNodeMempoolReactor(config *cfg.Config, proxyApp proxy.AppConns, state sm.State, memplMetrics *mempl.Metrics, logger log.Logger) (*mempl.MempoolReactor, *mempl.Mempool) {
```
```suggestion
func createMempoolAndMempoolReactor(config *cfg.Config, proxyApp proxy.AppConns, state sm.State, memplMetrics *mempl.Metrics, logger log.Logger) (*mempl.Mempool, *mempl.MempoolReactor) {
```

node/node.go
```diff
@@ -339,34 +330,34 @@ func NewNode(config *cfg.Config,
 	if config.Consensus.WaitForTxs() {
 		mempool.EnableTxsAvailable()
 	}
+	return mempoolReactor, mempool
+}
 
-	// Make Evidence Reactor
+func buildNewNodeEvidenceReactor(config *cfg.Config, dbProvider DBProvider, stateDB dbm.DB, logger log.Logger) (*evidence.EvidenceReactor, *evidence.EvidencePool, error) {
```
```suggestion
func createEvidenceReactor(config *cfg.Config, dbProvider DBProvider, stateDB dbm.DB, logger log.Logger) (*evidence.EvidenceReactor, *evidence.EvidencePool, error) {
```

node/node.go
```diff
@@ -376,34 +367,19 @@ func NewNode(config *cfg.Config,
...
-			state.Version.Consensus.Block,
-			state.Version.Consensus.App,
-		),
-	)
-	if err != nil {
-		return nil, err
-	}
-
-	// Setup Transport.
+func setupNewNodeTransport(config *cfg.Config, nodeInfo p2p.NodeInfo, nodeKey *p2p.NodeKey, proxyApp proxy.AppConns) (*p2p.MultiplexTransport, []p2p.PeerFilterFunc) {
```
```suggestion
func createTransport(config *cfg.Config, nodeInfo p2p.NodeInfo, nodeKey *p2p.NodeKey, proxyApp proxy.AppConns) (*p2p.MultiplexTransport, []p2p.PeerFilterFunc) {
```

better be consistent with naming (`create`, `build`, `setup` all mean the same thing)

node/node.go
```diff
@@ -456,36 +432,40 @@ func NewNode(config *cfg.Config,
 	}
 
 	p2p.MultiplexTransportConnFilters(connFilters...)(transport)
+	return transport, peerFilters
+}
+
+func buildNewNodeSwitch(config *cfg.Config,
```
```sugggestion
func createSwitch(config *cfg.Config,
```

Why `NewNode` prefix is necessary?

node/node.go
```diff
@@ -456,36 +432,40 @@ func NewNode(config *cfg.Config,
...
-	//
-	// We need to set Seeds and PersistentPeers on the switch,
-	// since it needs to be able to use these (and their DNS names)
-	// even if the PEX is off. We can include the DNS name in the NetAddress,
-	// but it would still be nice to have a clear list of the current "PersistentPeers"
-	// somewhere that we can return with net_info.
-	//
-	// If PEX is on, it should handle dialing the seeds. Otherwise the switch does it.
-	// Note we currently use the addrBook regardless at least for AddOurAddress
+func buildNewNodeAddrBook(config *cfg.Config, nodeInfo p2p.NodeInfo, sw *p2p.Switch, logger, p2pLogger log.Logger) pex.AddrBook {
```
```suggestion
func createAddrBook(config *cfg.Config, nodeInfo p2p.NodeInfo, sw *p2p.Switch, logger, p2pLogger log.Logger) pex.AddrBook {
```

node/node.go
```diff
@@ -504,14 +484,147 @@ func NewNode(config *cfg.Config,
 	}
 
 	sw.SetAddrBook(addrBook)
+	return addrBook
+}
 
-	// run the profile server
+func runProfileServer(config *cfg.Config, logger log.Logger) {
```
```suggestion
func optionallyRunProfileServer(config *cfg.Config, logger log.Logger) {
```

node/node.go
```diff
@@ -251,27 +244,23 @@ func NewNode(config *cfg.Config,
...
+	if err := indexerService.Start(); err != nil {
+		return nil, nil, err
 	}
+	return indexerService, txIndexer, nil
+}
 
-	// Create the handshaker, which calls RequestInfo, sets the AppVersion on the state,
-	// and replays any blocks as necessary to sync tendermint with the app.
-	consensusLogger := logger.With("module", "consensus")
+func doHandshakeForNewNode(stateDB dbm.DB, state sm.State, blockStore sm.BlockStore, genDoc *types.GenesisDoc, eventBus *types.EventBus, proxyApp proxy.AppConns, logger log.Logger) error {
```
```suggestion
func doHandshake(stateDB dbm.DB, state sm.State, blockStore sm.BlockStore, genDoc *types.GenesisDoc, eventBus *types.EventBus, proxyApp proxy.AppConns, logger log.Logger) error {
```

node/node.go
```diff
@@ -504,14 +484,147 @@ func NewNode(config *cfg.Config,
...
+	}
+
+	// Reload the state. It will have the Version.Consensus.App set by the
+	// Handshake, and may have other modifications as well (ie. depending on
+	// what happened during block replay).
+	state = sm.LoadState(stateDB)
+
+	// If an address is provided, listen on the socket for a connection from an
+	// external signing process.
+	privValidator, err = optionallyCreateAndStartPrivValidator(config, privValidator, logger)
```
melekes: If we're going to use this `optionally` prefix (which I personally don't like because it brings uncertainty), we should document the expected behavior. Like `if the address is not provided, privValidator and err are both nil`. Note this won't prevent developers from making a mistake where they would assume `privValidator` can't ever be nil later in the code.

node/node.go
thanethomson: We can. It does, however, involve moving the `GenesisDocProvider`-related code from `node/node.go` to `state/store.go` to prevent circular dependencies. This changes the Go API a bit. Do you think that'd be a problem?

Technically the genesis doc represents a form of state, so it would make sense to have it in the `state` package, right?

node/node.go
xla: In the case when both are `nil` we just asking for unexpected panics. Better to encode this in an explicit error which the caller can react on.

node/node.go
thanethomson: Initially it was to make it clear where the functions were extracted from. I thought that perhaps if they were eventually used in tests that do partial node instantiations using those functions it'd also make it clearer that they were called by the `NewNode` function.

But I suppose that in itself would only be an intermediate step. The ultimate goal would be to have a simpler interface to do partial node instantiations (e.g. some kind of `NewNode` function with a whole bunch of flags to indicate which parts of the node to create, where this function would be used by tests that want partial node instantiations).

node/node.go
melekes: Hmm.. will respond tomorrow

node/node.go
```diff
@@ -286,6 +254,8 @@ func logNewNodeInfo(state sm.State, privValidator types.PrivValidator, logger, c
 	}
 }
 
+// This creates and starts the private validator socket client if, and only if,
+// `config.PrivValidatorListenAddr` is non-empty.
 func optionallyCreateAndStartPrivValidator(config *cfg.Config, privValidator types.PrivValidator, logger log.Logger) (types.PrivValidator, error) {
 	if config.PrivValidatorListenAddr != "" {
```
melekes: early return?

```go
if config.PrivValidatorListenAddr == "" {
  return nil, nil
}
```

node/node.go
```diff
@@ -298,6 +268,10 @@ func optionallyCreateAndStartPrivValidator(config *cfg.Config, privValidator typ
 	return privValidator, nil
 }
 
+// This returns whether or not fast sync is enabled based on (1) whether
+// `config.FastSync` is true, and (2) if there's only a single validator,
+// whether the given `privValidator`'s address is not the same as the sole
+// validator's address.
 func optionallyFastSync(config *cfg.Config, state sm.State, privValidator types.PrivValidator) bool {
```
```go
func onlyValidatorIsUs(state sm.State, privVal types.PrivValidator) bool {
	if state.Validators.Size() > 1 {
		return false
	}

	addr, _ := state.Validators.GetByIndex(0)
	return bytes.Equal(privVal.GetPubKey().Address(), addr)
}
```

Here is what is bad about optionallyFastSync func:

1) `optionallyFastSync` suggest that the function performs fast sync, which it does not
2) even if we rename it to `onlyValidatorIsUs(config.FastSync, ...) (fastSync bool)`, we'll be mixing responsibilities => spaghetti code. function should have only one primary responsibility

I think we should extract onlyValidatorIsUs func and use it to make a decision if we should fast sync or not inside another function or directly in NewNode func

```go
fastSync := config.FastSync && !onlyValidatorIsUs(state, privValidator)
```

node/node.go
```diff
@@ -493,7 +467,9 @@ func buildNewNodeAddrBook(config *cfg.Config, nodeInfo p2p.NodeInfo, sw *p2p.Swi
 	return addrBook
 }
 
-func runProfileServer(config *cfg.Config, logger log.Logger) {
+// This instantiates the profile server in a separate goroutine only if
+// `config.ProfileListenAddress` is non-empty.
+func optionallyRunProfileServer(config *cfg.Config, logger log.Logger) {
 	profileHost := config.ProfListenAddress
 	if profileHost != "" {
```
melekes: early return?

state/store.go
```diff
@@ -65,6 +84,52 @@ func LoadStateFromDBOrGenesisDoc(stateDB dbm.DB, genesisDoc *types.GenesisDoc) (
...
+// panics if failed to unmarshal bytes
+func loadGenesisDoc(db dbm.DB) (*types.GenesisDoc, error) {
+	bytes := db.Get(genesisDocKey)
+	if len(bytes) == 0 {
+		return nil, errors.New("Genesis doc not found")
+	}
+	var genDoc *types.GenesisDoc
+	err := cdc.UnmarshalJSON(bytes, &genDoc)
+	if err != nil {
+		cmn.PanicCrisis(fmt.Sprintf("Failed to load genesis doc due to unmarshaling error: %v (bytes: %X)", err, bytes))
```
melekes:
```suggestion
		panic(fmt.Sprintf("Failed to load genesis doc due to unmarshaling error: %v (bytes: %X)", err, bytes))
```

state/store.go
```diff
@@ -65,6 +84,52 @@ func LoadStateFromDBOrGenesisDoc(stateDB dbm.DB, genesisDoc *types.GenesisDoc) (
...
+	}
+	return genDoc, nil
+}
+
+// panics if failed to marshal the given genesis document
+func saveGenesisDoc(db dbm.DB, genDoc *types.GenesisDoc) {
+	bytes, err := cdc.MarshalJSON(genDoc)
+	if err != nil {
+		cmn.PanicCrisis(fmt.Sprintf("Failed to save genesis doc due to marshaling error: %v", err))
```
melekes:
```suggestion
		panic(fmt.Sprintf("Failed to save genesis doc due to marshaling error: %v", err))
````

File: node/node.go (hunk 14 above)
Comment (thanethomson) [reply node/node.go]: Sure, we can refine it this way if you want.

File: state/store.go (hunk 16 above)
Comment (thanethomson) [reply state/store.go]: If we change this here, shouldn't we just replace it in the whole codebase and take out the deprecated functions altogether?

File: node/node.go (hunk 13 above)
Comment (thanethomson) [reply node/node.go]: We can return early here like this (which effectively results in a pass-through of whichever `privValidator` was passed to us):

```go
if config.PrivValidatorListenAddr == "" {
    return privValidator, nil
}
```

state/store.go
melekes: I don't mind, but should probably be a separate PR

thanethomson: Done in #3595

node/node.go
```diff
@@ -316,10 +252,32 @@ func NewNode(config *cfg.Config,
...
+// `config.PrivValidatorListenAddr` is non-empty.
+func optionallyCreateAndStartPrivValidator(config *cfg.Config, privValidator types.PrivValidator, logger log.Logger) (types.PrivValidator, error) {
+	if config.PrivValidatorListenAddr == "" {
+		return privValidator, nil
+	}
+	// FIXME: we should start services inside OnStart
+	var err error
+	privValidator, err = createAndStartPrivValidatorSocketClient(config.PrivValidatorListenAddr, logger)
+	if err != nil {
+		return nil, errors.Wrap(err, "Error with private validator socket client")
```
```suggestion
		return nil, errors.Wrap(err, "error with private validator socket client")
```

PR #3463, opened 2019-03-21
p2p: remove noisy error log

guagualvcha:
When node start with persistent peer that is private ip, there will be error log like:
```
E[2019-03-21|09:02:55.560] Can't add peer's address to addrbook         module=p2p err="Cannot add non-routable address 4390afd570366e50f13e05b592a6bc06fb77d02c@172.27.41.104:27146"
```
I don't think it is an error, so use debug level log.

* [ ] Wrote tests


melekes:
> I don't think it is an error, so use debug level log.

It's an error. I've added a recommendation to the err msg here https://github.com/tendermint/tendermint/pull/3474

`If it's a private network, set addr_book_strict to false in the config`

melekes:
I am going to close this in favor of https://github.com/tendermint/tendermint/pull/3474

PR #3473, opened 2019-03-22
rpc/client: include NetworkClient interface into Client interface

### LINKED ISSUE #3458: RPC Client interface does not include NetworkClient
**Tendermint version**: Master as of commit e0f8936455029a40287a69d5b0e7baa4d5864da1 (0.31.0)

**What happened**:

As I was making my RPC interface more generic (using client.Client), I found my rpc.NetInfo calls were not compiling. I looked at the source code and found no NetworkClient in the Client interface: https://github.com/tendermint/tendermint/blob/master/rpc/client/interface.go#L72

The Client interface should include the NetworkClient, otherwise full-featured tests and any other usage that swaps out the underlying interface will fail if the user wishes to test/use NetInfo or any other NetworkClient rpc call. 

melekes:
I think it's nice when the `Client` interface has all the methods. If someone does not need a particular method/set of methods, she can use individual interfaces (e.g. NetworkClient, MempoolClient) or write her own interface.

**technically breaking**

Fixes #3458

* [x] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -5,6 +5,7 @@
 ### BREAKING CHANGES:
 
 * CLI/RPC/Config
+- [rpc/client] \#3458 Include NetworkClient interface into Client interface
 
 * Apps
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -5,12 +5,12 @@
 ### BREAKING CHANGES:
 
 * CLI/RPC/Config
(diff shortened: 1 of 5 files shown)


CHANGELOG_PENDING.md
```diff
@@ -5,6 +5,7 @@
 ### BREAKING CHANGES:
 
 * CLI/RPC/Config
+- [rpc/client] \#3458 Include NetworkClient interface into Client interface
```
ebuchman: Go API

melekes:
> Is there some reason this was left out initially?

I can't think of any

PR #3474, opened 2019-03-22
p2p: do not log err if peer is private

melekes:
Should replace https://github.com/tendermint/tendermint/pull/3463

### DIFF
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -677,21 +677,8 @@ func (a *addrBook) addAddress(addr, src *p2p.NetAddress) error {
 		return ErrAddrBookNilAddr{addr, src}
 	}
 
-	if a.routabilityStrict && !addr.Routable() {
-		return ErrAddrBookNonRoutable{addr}
-	}
-
-	if !addr.Valid() {
-		return ErrAddrBookInvalidAddr{addr}
-	}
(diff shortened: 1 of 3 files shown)


guagualvcha:
Ebarrassing scene: I get a sentry node which would be public in p2p network, it will connect to validators as its persistent peers, and validators are served as private peers.
If I set `addr_book_strict ` to be true, sentry keep report error log. If I set `addr_book_strict ` to be false, validator address will pexed by sentry node.

melekes:
Right! What if we reorder checks in `addrbook#addAddress()` to return `ErrAddrBookPrivate` if the address is private and we do not log it as error? Note you'll need to add validators ids to `private_peer_ids` in config.toml.

melekes:
@guagualvcha see https://github.com/tendermint/tendermint/pull/3474/commits/a92e4347a6f93b688389646ed05dc3e545770886

guagualvcha:
> @guagualvcha see [a92e434](https://github.com/tendermint/tendermint/commit/a92e4347a6f93b688389646ed05dc3e545770886)

thks, That will be good

melekes:
TODO:

- [ ] verify output manually

melekes:
@guagualvcha have you tried applying this patch?

PR #3475, opened 2019-03-22
refactor GetSelectionWithBias for addressbook

guagualvcha:
Why submit this pr:
1.  we have suffered from `infinite loop in addrbook` bug which takes us a long time to find out why process become a  zombie peer. It have been fixed in https://github.com/tendermint/tendermint/pull/3232. But the `ADDRS_LOOP ` is still there, risk of infinite loop is still exist.
2. The algorithm that to random pick a bucket is not stable, which means the peer may unluckily always choose the wrong bucket for a long time, the time and cpu cost is meaningless.

A simple improvement:
shuffle `bucketsNew ` and `bucketsOld`, and pick necessary number of address from them. A stable
algorithm. 

* [ ] Wrote tests

### DIFF
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -9,6 +9,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"math"
+	"math/rand"
 	"net"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


p2p/pex/addrbook.go
```diff
@@ -726,6 +648,43 @@ func (a *addrBook) addAddress(addr, src *p2p.NetAddress) error {
...
+			addresses = append(addresses, ka)
+		}
+	}
+	selection := make([]*p2p.NetAddress, 0, num)
+	chosenSet := make(map[string]bool)
+	rand.Shuffle(total, func(i, j int) {
+		addresses[i], addresses[j] = addresses[j], addresses[i]
+	})
+	for _, addr := range addresses {
+		if !chosenSet[addr.Addr.String()] {
```
melekes: I'd prefer

```go
		if chosenSet[addr.Addr.String()] {
			continue
		}
```

p2p/pex/addrbook.go
```diff
@@ -726,6 +648,43 @@ func (a *addrBook) addAddress(addr, src *p2p.NetAddress) error {
...
+		total = total + len(bucket)
+	}
+	addresses := make([]*knownAddress, 0, total)
+	for _, bucket := range buckets {
+		for _, ka := range bucket {
+			addresses = append(addresses, ka)
+		}
+	}
+	selection := make([]*p2p.NetAddress, 0, num)
+	chosenSet := make(map[string]bool)
```
melekes: We can also preallocate

```
	chosenSet := make(map[string]bool, num)
```

guagualvcha: sure, updated.

p2p/pex/addrbook.go
guagualvcha: done

p2p/pex/addrbook.go
```diff
@@ -726,6 +648,44 @@ func (a *addrBook) addAddress(addr, src *p2p.NetAddress) error {
...
+	}
+	addresses := make([]*knownAddress, 0, total)
+	for _, bucket := range buckets {
+		for _, ka := range bucket {
+			addresses = append(addresses, ka)
+		}
+	}
+	selection := make([]*p2p.NetAddress, 0, num)
+	chosenSet := make(map[string]bool, num)
+	rand.Shuffle(total, func(i, j int) {
```
Liamsi: Just noting that this is slightly more deterministic now than previously. I do not think this is a problem (doesn't look like it was) but previously the common package randomness was used which seeds the global randomness using the crypto/rand package while math/rand is seeded with `1`: https://golang.org/src/math/rand/rand.go?#L288
To get a behaviour more like the previous one could call `rand.Seed` with some int derived from crypto/rand

p2p/pex/addrbook.go
```diff
@@ -726,6 +648,44 @@ func (a *addrBook) addAddress(addr, src *p2p.NetAddress) error {
...
+		addresses[i], addresses[j] = addresses[j], addresses[i]
+	})
+	for _, addr := range addresses {
+		if chosenSet[addr.Addr.String()] {
+			continue
+		}
+		chosenSet[addr.Addr.String()] = true
+		selection = append(selection, addr.Addr)
+		if len(selection) >= num {
```
Liamsi: This would never actually exceed `num`, right? I mean the calling code would work in both cases 👍 but it was a bit confusing to me why `>=` is necessary here (instead of `==`).

guagualvcha: you are right. use "==" is enough, but ">=" is for safe.

p2p/pex/addrbook.go
```diff
@@ -405,89 +406,10 @@ func (a *addrBook) GetSelectionWithBias(biasTowardsNewAddrs int) []*p2p.NetAddre
...
-			newAddressesAdded++
-		}
-
-		selection[selectionIndex] = selectedAddr
-		selectionIndex++
-	}
-
+	numRequiredNewAdd := cmn.MaxInt(percentageOfNum(biasTowardsNewAddrs, numAddresses), numAddresses-a.nOld)
+	selection := a.randomPickAddresses(bucketTypeNew, numRequiredNewAdd)
+	selection = append(selection, a.randomPickAddresses(bucketTypeOld, numAddresses-len(selection))...)
```
ancazamfir: Before this change the selection was:
 - percentageOfNum(biasTowardsNewAddrs, numAddresses) -> new addresses
 - old addresses, and
 - if there are not enough old addresses to fill the selection, more new addresses are added at the end of the selection.
With this change the old addresses (a.k.a. good addresses) are pushed back to the end of the selection. 
This means that the pex peer that receives the selection will process the good addresses last, perhaps delaying connections to good peers. I'm not sure if this is a real issue though.
On the other hand it's a simple change to make it back the way it was.

melekes: > This means that the pex peer that receives the selection will process the good addresses last, perhaps delaying connections to good peers. I'm not sure if this is a real issue though.

Not an issue

p2p/pex/addrbook.go
melekes: I prefer >= too!

p2p/pex/addrbook.go
```diff
@@ -726,6 +648,44 @@ func (a *addrBook) addAddress(addr, src *p2p.NetAddress) error {
...
+			addresses = append(addresses, ka)
+		}
+	}
+	selection := make([]*p2p.NetAddress, 0, num)
+	chosenSet := make(map[string]bool, num)
+	rand.Shuffle(total, func(i, j int) {
+		addresses[i], addresses[j] = addresses[j], addresses[i]
+	})
+	for _, addr := range addresses {
+		if chosenSet[addr.Addr.String()] {
```
melekes: why do we have to check if we've already selected an address? Do we have duplicates in the addrbook?

p2p/pex/addrbook.go
```diff
@@ -405,89 +406,10 @@ func (a *addrBook) GetSelectionWithBias(biasTowardsNewAddrs int) []*p2p.NetAddre
...
-			}
-			newBucketToAddrsMap[newIndex][selectedAddr.String()] = struct{}{}
-			newAddressesAdded++
-		}
-
-		selection[selectionIndex] = selectedAddr
-		selectionIndex++
-	}
-
+	numRequiredNewAdd := cmn.MaxInt(percentageOfNum(biasTowardsNewAddrs, numAddresses), numAddresses-a.nOld)
```
melekes: Shouldn't this be MinInt?

E.g. bias = 30%, numAddresses = 10

Want 3 new addrs. If there are 9 new addrs, numRequiredNewAdd will be 9, BUT we only want 3!

ancazamfir: Yes but we cannot have nil entries. Even before we were returning 3 new, 1 old, 6 new...i.e. when not enough old addresses, we were padding the selection with new addresses.

p2p/pex/addrbook.go
guagualvcha: yes, we have duplicate address in newBuckets. Since the pex address may receive from different src peer, I think that is why addressbook is design to use buckets, to let address that 
from multiple peer get more chance to pex

p2p/pex/addrbook.go
melekes: Should we write a comment explaining this stuff?

melekes:
Hmm.. I've just tried to test this using 1 seed and 3 regular nodes connected to it. The network failed to start.

melekes:
> Hmm.. I've just tried to test this using 1 seed and 3 regular nodes connected to it. The network failed to start.

maybe it's a problem with my setup

p2p/pex/addrbook.go
ebuchman: Reason being because the peer who receives them doesn't know which are old and which are new, only the sending peer knows from their own perspective. So from the receiver's side, they are all just added to the addrbook as "new" until they prove themselves.

Issue #3460, opened 2019-03-21
replayBlocks checkAppHash should't wait the last block


Zerppen:
**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):
0d985ede28bd6937fa9d3613618e42cab6fc871c

**ABCI app** (name for built-in, URL for self-written if it's publicly available):

**Environment**:
- **OS** (e.g. from /etc/os-release):
- **Install tools**:
- **Others**:

**What happened**:
there are some code should be enhanced

**What you expected to happen**:
I will contribute code later
**Have you tried the latest version**: yes/no

**How to reproduce it** (as minimally and precisely as possible):

**Logs (paste a small part showing an error (< 10 lines) or link a pastebin, gist, etc. containing more of the log file)**:

**Config (you can paste only the changes you've made)**:

**node command runtime flags**:

**`/dump_consensus_state` output for consensus bugs**

**Anything else we need to know**:

Zerppen:
I will enhance it and contribute code

Liamsi:
> there are some code should be enhanced

Could you please explain a bit what you mean here?

Zerppen:
@Liamsi  current replay logic,only the last block's apphash will be checked.
However,the first block to replay may be wrong apphash.
In my opinion each block's apphash should be checked.

james-ray:
The replayed blocks are already checked either in consensus process or in blockchain reactor pulling blocks process, either process has checked the block.  
Replay a block is merely to sync the app data with Tendermint.
Why the last block need to be executed through ApplyBlock is not because it need to be checked but we need to modify the state of Tendermint for it shutdown between save block and state, the state hasn't been saved in last shutdown.

xla:
This issue is not actionable at this point, looking forward to contributions if you see fit to improve things.

xla:
@james-ray Thanks for the clarification.

melekes:
It would detect if the app changed in a way it shouldn't have.

melekes:
Merged https://github.com/tendermint/tendermint/pull/3579 to develop.

PR #3465, opened 2019-03-21
fix issue3460

Zerppen:
When replaying blocks we should check apphash,however current logic will result in replaying too many wrong blocks if current block's apphash is wrong.
To solve this problem,check each block's appash is import and necessary .
* [ ] Updated all relevant documentation in docs
* [ ] Updated all code comments where relevant
* [ ] Wrote tests
* [ ] Updated CHANGELOG_PENDING.md

PR #3477, opened 2019-03-23
consensus: Replay should check every block's apphash #3460

Zerppen:
Refs #3460

* [ ] Wrote tests

### DIFF
--- a/consensus/replay.go
+++ b/consensus/replay.go
@@ -429,6 +429,8 @@ func (h *Handshaker) replayBlocks(state sm.State, proxyApp proxy.AppConns, appBl
 				if err != nil {
 					return nil, err
 				}
+			} else {
+				panic(fmt.Errorf(" Replay block can't find stored block,height %X",  i+1).Error())
 			}
 		}
 
--- a/consensus/replay.go
+++ b/consensus/replay.go
@@ -428,7 +428,6 @@ func (h *Handshaker) replayBlocks(state sm.State, proxyApp proxy.AppConns, appBl
 			if err != nil {
 				return nil, err
 			}
-
 		}
 
 		h.nBlocks++
--- a/consensus/replay.go
+++ b/consensus/replay.go
@@ -422,11 +422,13 @@ func (h *Handshaker) replayBlocks(state sm.State, proxyApp proxy.AppConns, appBl
 			return nil, err
 		}
 		//check each block's apphash when replay
-		nextBlock := h.store.LoadBlock(i + 1)
-		if nextBlock != nil {
-			err = checkBlockAppHash(appHash, nextBlock.AppHash, i)
-			if err != nil {
-				return nil, err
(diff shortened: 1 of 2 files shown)


Liamsi:
Thanks again @Zerppen! In https://github.com/tendermint/tendermint/pull/3465#issuecomment-475271689 you asked why the tests fail:
the test `TestHandshakeReplayAll` combined with your changes:
```go
//check each block's apphash when replay
nextBlock := h.store.LoadBlock(i + 1)
```
fails. Note that the test uses the `mockBlockStore` which provides a mocked version of `LoadBlock`. Your changes call it with height 7 but there are only 6 blocks in the `chain` array (see the `mockBlockStore` struct).

Liamsi:
Also, please format your code with `gofmt -s -w`.

Zerppen:
yeah.It was my mistake

melekes:
1) Why do we need to check block's app hash?
2) I don't think we need to load n+1 block. We can check the apphash here https://github.com/tendermint/tendermint/blob/25a3c8b1724c9611d6edc175b1b0d079f5ee28c1/consensus/replay.go#L419 

```go
		block := h.store.LoadBlock(i)
		if appHash != nil {
                	checkBlockAppHash(block.AppHash, appHash)
		}
		appHash, err = sm.ExecCommitBlock(proxyApp.Consensus(), block, h.logger, state.LastValidators, h.stateDB)
		if err != nil {
			return nil, err
		}

		h.nBlocks++
	}
```

Zerppen:
1.Current block's appash is made from last block.
2.And sm.ExecCommitBlock() return next block's apphash of app
3. but load n+1 block to get storeblock's apphash
Therefore, if first block's apphash of app returned is wrong followings must be wrong.
But current logic is only check last block's apphash.
@melekes

melekes:
Okay. Sounds like checking app hash is a good idea 👍 

What about my 2nd point?

Zerppen:
I think apphash was getting angry because you don't know he much.
To cheer apphash up,I gave you 3 points.
And I think reading each line of code can make you make sure he don't be angry.I mean apphash lol.

Zerppen:
should be` h.store.LoadBlock(i + 1) `
not `h.store.LoadBlock( i )`

consensus/replay.go
```diff
@@ -421,6 +421,18 @@ func (h *Handshaker) replayBlocks(state sm.State, proxyApp proxy.AppConns, appBl
 		if err != nil {
 			return nil, err
 		}
+		//check each block's apphash when replay
+		if !(h.store.Height() < i+1) {
+			nextBlock := h.store.LoadBlock(i + 1)
```
melekes: I still think we can do this by saving prevBlock, not loading i+1 block

melekes:
Closing in favor of https://github.com/tendermint/tendermint/pull/3579

Issue #3306, opened 2019-02-13
Memory leak in p2p


hendrikhofstadt:
We have been observing memory usage increase over time (~1MB/min) on our sentries until they were OOM killed.

A heap profile shows that the `p2p` module seems to be leaking Peers with the `recvBuffers` of multiplex channels making up the largest amount of memory.

The following heap profile was taken from a normal node on the `game_of_stakes_6` network based on `v0.30.0-rc0`.

![image](https://user-images.githubusercontent.com/1405758/52703622-10293980-2f7f-11e9-9688-c3d30624dc01.png)

Rel #3183 #2967 #2821

ebuchman:
These are not seeds, correct? 

Did you see this behaviour in previous versions or it's new?

hendrikhofstadt:
@ebuchman This happens on all types of nodes. The node the heap profile was taken on was not a seed.

We've been seeing this for 2+ months.

ebuchman:
Ok thanks - any chance you have a go-routine dump for this too ?

melekes:
ping

hendrikhofstadt:
@melekes 
```
goroutine profile: total 774
172 @ 0x433bdb 0x433c83 0x44478c 0x4443c9 0x46b7f4 0xb80e9b 0x461d21
#	0x4443c8	sync.runtime_Semacquire+0x38										/usr/lib/golang/src/runtime/sema.go:56
#	0x46b7f3	sync.(*WaitGroup).Wait+0x63										/usr/lib/golang/src/sync/waitgroup.go:130
#	0xb80e9a	github.com/cosmos/cosmos-sdk/vendor/github.com/tendermint/tendermint/p2p.(*Switch).Broadcast.func2+0x2a	/go/src/github.com/cosmos/cosmos-sdk/vendor/github.com/tendermint/tendermint/p2p/switch.go:249

172 @ 0x433bdb 0x443bdd 0xb2f052 0xb2c094 0xb779ce 0xb80e18 0x461d21
#	0xb2f051	github.com/cosmos/cosmos-sdk/vendor/github.com/tendermint/tendermint/p2p/conn.(*Channel).sendBytes+0xf1	/go/src/github.com/cosmos/cosmos-sdk/vendor/github.com/tendermint/tendermint/p2p/conn/connection.go:729
[...]

PR #3478, opened 2019-03-24
R4R: Resolve broadcastTxRoutine leakage issue

HaoyangLiu:
Ref: #3306, https://github.com/irisnet/tendermint/pull/44/commits/fdbb67655db45726173895fd4ca6458180a8fc11

I ran an irishub validator. After the validator node ran several days, I dump the whole goroutine stack. I found that there were hundreds of broadcastTxRoutine. However, the connected peer quantity was less than 30. So I belive that there must be broadcastTxRoutine leakage issue.

According to my analysis, I think the root cause of this issue locate in below code:
```
		select {
		case <-next.NextWaitChan():
			// see the start of the for loop for nil check
			next = next.Next()
		case <-peer.Quit():
			return
		case <-memR.Quit():
			return
		}
```
As we know, if multiple paths are avaliable in the same time, then a random path will be selected. Suppose that `next.NextWaitChan()` and `peer.Quit()` are both avaliable, and `next.NextWaitChan()` is chosen. 
```
                // send memTx
		msg := &TxMessage{Tx: memTx.tx}
		success := peer.Send(MempoolChannel, cdc.MustMarshalBinaryBare(msg))
		if !success {
			time.Sleep(peerCatchupSleepIntervalMS * time.Millisecond)
			continue
		}
```
Then `next` will be non-empty and the peer send operation won't be success. As a result, this go routine will be track into infinite loop and won't be released. 

My proposal is to check `peer.Quit()` and `memR.Quit()` in every loop no matter whether `next` is nil.

* [ ] Wrote tests

### DIFF
--- a/mempool/reactor.go
+++ b/mempool/reactor.go
@@ -112,6 +112,10 @@ func (memR *MempoolReactor) broadcastTxRoutine(peer p2p.Peer) {
 
 	var next *clist.CElement
 	for {
+		// In case of both next.NextWaitChan() and peer.Quit() are variable at the same time
+		if !memR.IsRunning() || !peer.IsRunning() {
+			return
+		}
 		// This happens because the CElement we were looking at got garbage
 		// collected (removed). That is, .NextWait() returned nil. Go ahead and
 		// start from the beginning.


melekes:
tests failed with:

```
panic: sync: negative WaitGroup counter

goroutine 5 [running]:
sync.(*WaitGroup).Add(0xc00027a000, 0xffffffffffffffff)
	/usr/local/go/src/sync/waitgroup.go:74 +0x20c
sync.(*WaitGroup).Done(...)
	/usr/local/go/src/sync/waitgroup.go:99
github.com/tendermint/tendermint/abci/client.(*socketClient).didRecvResponse(0xc0001fc1e0, 0xc0001faa80, 0x0, 0x0)
	/go/src/github.com/tendermint/tendermint/abci/client/socket_client.go:207 +0x235
github.com/tendermint/tendermint/abci/client.(*socketClient).recvResponseRoutine(0xc0001fc1e0, 0xf52640, 0xc000010010)
	/go/src/github.com/tendermint/tendermint/abci/client/socket_client.go:176 +0x32b
created by github.com/tendermint/tendermint/abci/client.(*socketClient).OnStart
	/go/src/github.com/tendermint/tendermint/abci/client/socket_client.go:77 +0x422
FAIL	github.com/tendermint/tendermint/abci/client	0.256s
```

mempool/reactor.go
```diff
@@ -112,19 +112,21 @@ func (memR *MempoolReactor) broadcastTxRoutine(peer p2p.Peer) {
 
 	var next *clist.CElement
 	for {
+		select {
```
melekes: We should probably preserve the old select and just add

```go
if !memR.IsRunning() || !peer.IsRunning() {
	return
}
```

here. What do you think?

melekes:
@HaoyangLiu thank you investigating this 👍 See my comment https://github.com/tendermint/tendermint/pull/3478#discussion_r268421432

HaoyangLiu:
Good idea! Your suggestion has less risk.

HaoyangLiu:
```
--- FAIL: TestListenerTimeoutReadWrite (0.01s)
    socket_listeners_test.go:110: read tcp 127.0.0.1:33162->127.0.0.1:49908: i/o timeout
```
The failure case in test_cover seems have no connection with my change.

melekes:
> As a result, this go routine will be track into infinite loop and won't be released.

Check out the example I wrote: https://play.golang.org/p/tH1VofiO25L. The program does exit after 1s => no infinite loop!

Can you confirm that with this patch applied you're no longer seeing the `broadcastTxRoutine` leak?

HaoyangLiu:
[goroutine.txt](https://github.com/tendermint/tendermint/files/3002829/goroutine.txt)
[net_info.txt](https://github.com/tendermint/tendermint/files/3002837/net_info.txt)
This is a the dumped goroutine stack and net_info from my [irishub](https://github.com/irisnet/irishub/releases/tag/v0.12.3) validator node.

mempool/reactor.go
```diff
@@ -112,6 +112,10 @@ func (memR *MempoolReactor) broadcastTxRoutine(peer p2p.Peer) {
 
 	var next *clist.CElement
 	for {
+		// In case of both next.NextWaitChan() and peer.Quit() are variable at the same time
```
Liamsi: Thanks so much! Great find. Just the comment here is a bit confusing. Maybe:
```suggestion
		// Prevent infinite loop in case next != nil
```

HaoyangLiu:
This issue is found on a blockchain network which has a lot of nodes and has been run for a long time. I have tried to reproduce this issue with few nodes and want to create a bug bounty. But I failed. Currently I 'm afraid I might have no time in writing related test. Sorry for that.

PR #3487, opened 2019-03-27
mempool: add a safety check, write tests for mempoolIDs

melekes:
and document 65536 limit in the mempool reactor spec

follow-up to https://github.com/tendermint/tendermint/pull/2778

* [ ] Wrote tests


mempool/reactor.go
```diff
@@ -45,19 +48,25 @@ type mempoolIDs struct {
 	activeIDs map[uint16]struct{} // used to check if a given peerID key is used, the value doesn't matter
 }
 
-// Reserve searches for the next unused ID and assignes it to the peer.
-func (ids *mempoolIDs) ReserveForPeer(peer p2p.Peer) {
+// Reserve searches for the next unused ID and assignes it to the
+// peer. It returns the ID assigned.
+func (ids *mempoolIDs) ReserveForPeer(peer p2p.Peer) uint16 {
```
Liamsi: Looks like this is only useful for tests?

mempool/reactor.go
```diff
@@ -131,13 +144,26 @@ func (memR *MempoolReactor) GetChannels() []*p2p.ChannelDescriptor {
...
 }
 
 // RemovePeer implements Reactor.
 func (memR *MempoolReactor) RemovePeer(peer p2p.Peer, reason interface{}) {
 	memR.ids.Reclaim(peer)
+
+	// XXX (melekes): commented out for performance reasons
+	//
+	// Remove the sender if present to avoid a new node joining and not receiving
+	// a tx because we think it already has it. For that two things should happen:
```
Liamsi:
```suggestion
	// a tx because we think it already has it. This happens if the following two statements are true:
```

mempool/reactor.go
```diff
@@ -131,13 +144,26 @@ func (memR *MempoolReactor) GetChannels() []*p2p.ChannelDescriptor {
...
+	//   1) node has to connect to over 65536 peers during its lifetime
```
Liamsi:
```suggestion
	//   1) node has connected to over 65536 peers during its lifetime
```

mempool/reactor.go
```diff
@@ -131,13 +144,26 @@ func (memR *MempoolReactor) GetChannels() []*p2p.ChannelDescriptor {
...
+	//   2) tx must reside in the mempool for the above duration
```
Liamsi:
```suggestion
	//   2) tx resides in the mempool for the above duration
```

mempool/reactor.go
```diff
@@ -131,13 +144,26 @@ func (memR *MempoolReactor) GetChannels() []*p2p.ChannelDescriptor {
...
+	//
+	// for e := memR.Mempool.txs.Front(); e != nil; e = e.Next() {
+	// 	memTx := e.Value.(*mempoolTx)
+	// 	memTx.senders.Delete(id)
+	// }
```
Liamsi: So it is not only performance reasons why this is not added but this is also unlikely (that both 1) and 2) are true), right?

p2p/mock/peer.go
```diff
@@ -0,0 +1,61 @@
+package mock
```
Liamsi: Can't we re-use `p2p/dummy` for this?

mempool/reactor.go
```diff
@@ -131,13 +144,26 @@ func (memR *MempoolReactor) GetChannels() []*p2p.ChannelDescriptor {
...
-	memR.ids.ReserveForPeer(peer)
+	_ = memR.ids.ReserveForPeer(peer)
 	go memR.broadcastTxRoutine(peer)
 }
 
 // RemovePeer implements Reactor.
 func (memR *MempoolReactor) RemovePeer(peer p2p.Peer, reason interface{}) {
 	memR.ids.Reclaim(peer)
+
+	// XXX (melekes): commented out for performance reasons
```
Liamsi: It's not clear why this is added but immediately commented out. OK, it says performance reasons but why is the commented code added in the first place then?

p2p/mock/peer.go
Liamsi: Done in d2f3a306be5ca4e6890fcab18b14785cceaffb65

PR #3488, opened 2019-03-27
docs: Fix broken links (#3482)

### LINKED ISSUE #3482: docs: signature spec link is broken
On https://tendermint.com/docs/spec/blockchain/blockchain.html#signature

>>>
See the [signature spec](https://tendermint.com/docs/docs/spec/blockchain/encoding.html#key-types) for more. => 404

brapse:
A bunch of links were broken in the documentation, they included the `docs` prefix. 
Fixes #3482 

* [ ] ~~Wrote tests~~

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -26,3 +26,4 @@
 ### BUG FIXES:
 
 - [blockchain] \#2699 update the maxHeight when a peer is removed
+- [docs] \#3482 fix broken links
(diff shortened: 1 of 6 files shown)


ebuchman:
Ah, thanks for this.

The problem is that these links work on github but then fail on the website. I'm not sure if there's a way to fix them so that they work on both. cc @zramsay

melekes:
We should target website. I personally never browse github docs if there's a website.

brapse:
Ah, relative links should work on both Github and compiled website. Will fix it up.

zramsay:
https://github.com/tendermint/tendermint/blob/master/docs/DOCS_README.md#relative

PR #3498, opened 2019-03-27
Release management using CircleCI

greg-szabo:
This PR is the first step for backporting and security releases deployed using CircleCI. It moves the current Jenkins-based release management over to CircleCI and introduces the vX.X branching model. This step does not support security releases yet. The architecture of security releases is still under discussion.

* Removed golang versions from CircleCI builds. This ensures that we always use the latest Golang version provided by CircleCI. (They update it faster than we do.)
* Minor changes to the `Makefile` so the binary name of the build can be configured. This is necessary because of the parallel builds of multiple architectures.
* New CircleCI release management steps: prepare, build, release (to GitHub), release to Docker.
  * The prepare and build steps run on every branch, so when new features are added we have an indication if the release process will succeed.
  * The release steps only run on vX.X branches to add a permission model for the releases.

* [X] Ran tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 defaults: &defaults
   working_directory: /go/src/github.com/tendermint/tendermint
   docker:
-    - image: circleci/golang:1.12.0
+    - image: circleci/golang
   environment:
     GOBIN: /tmp/workspace/bin
 
@@ -13,6 +13,9 @@ docs_update_config: &docs_update_config
     - image: tendermintdev/jq_curl
   environment:
     AWS_REGION: us-east-1
+
+release_management_docker: &release_management_docker
+  machine: true
 
 jobs:
   setup_dependencies:
@@ -192,7 +195,7 @@ jobs:
             name: run localnet and exit on failure
             command: |
               set -x
-              docker run --rm -v "$PWD":/go/src/github.com/tendermint/tendermint -w /go/src/github.com/tendermint/tendermint golang:1.11.4 make build-linux
+              docker run --rm -v "$PWD":/go/src/github.com/tendermint/tendermint -w /go/src/github.com/tendermint/tendermint golang make build-linux
               make localnet-start &
               ./scripts/localnet-blocks-test.sh 40 5 10 localhost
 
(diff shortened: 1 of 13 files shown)


.circleci/config.yml
```diff
@@ -256,6 +259,105 @@ jobs:
...
+          keys:
+            - v1-release-deps-{{ .Branch }}-{{ .Revision }}
+      - attach_workspace:
+          at: /tmp/workspace
+      - run:
+          name: Build artifact
+          command: |
+            # Setting CIRCLE_TAG because we do not tag the release ourselves.
+            source /tmp/workspace/release-version.source
+            if test ${CIRCLE_NODE_INDEX:-0} == 0 ;then export GOOS=linux GOARCH=amd64   && export OUTPUT=build/tendermint_${GOOS}_${GOARCH} && make build && python -u scripts/release_management/zip-file.py ;fi
```
melekes: maybe we can use https://github.com/mitchellh/gox here?

.circleci/config.yml
```diff
@@ -256,6 +259,105 @@ jobs:
...
+      - run:
+          name: Get next release number
+          command: |
+            export LAST_TAG="`git describe --tags --abbrev=0 --match "${CIRCLE_BRANCH}.*"`"
+            echo "Last tag: ${LAST_TAG}"
+            if [ -z "${LAST_TAG}" ]; then
+              export LAST_TAG="${CIRCLE_BRANCH}"
+              echo "Last tag not found. Possibly fresh branch or feature branch. Setting ${LAST_TAG} as tag."
+            fi
+            export NEXT_TAG="`python -u scripts/release_management/bump-semver.py --version "${LAST_TAG}"`"
```
melekes: how bump-semver decides whenever it's a major or minor release?

.circleci/config.yml
```diff
@@ -256,6 +259,105 @@ jobs:
...
+            echo "---"
+            export RELEASE_ID="`python -u scripts/release_management/github-draft.py`"
+            echo "Release ID: ${RELEASE_ID}"
+            #Todo: Parallelize uploads
+            export GOOS=linux GOARCH=amd64   && python -u scripts/release_management/github-upload.py --id "${RELEASE_ID}"
+            export GOOS=darwin GOARCH=amd64  && python -u scripts/release_management/github-upload.py --id "${RELEASE_ID}"
+            export GOOS=windows GOARCH=amd64 && python -u scripts/release_management/github-upload.py --id "${RELEASE_ID}"
+            export GOOS=linux GOARCH=arm     && python -u scripts/release_management/github-upload.py --id "${RELEASE_ID}"
+            python -u scripts/release_management/github-upload.py --file "/tmp/workspace/SHA256SUMS" --id "${RELEASE_ID}"
+            python -u scripts/release_management/github-publish.py --id "${RELEASE_ID}"
```
melekes: have you considered using https://github.com/goreleaser/goreleaser ? what is the benefit of using our own Python scripts? security?

scripts/release_management/README.md
```diff
@@ -0,0 +1,65 @@
...
+The scripts in this folder are used for release management in CircleCI. Although the scripts are fully configurable using input parameters,
+the default settings were modified to accommodate CircleCI execution.
+
+# Build scripts
+These scripts help during the build process. They prepare the release files.
+
+## bump-semver.py
+Bumps the semantic version of the input `--version`. Versions are expected in vMAJOR.MINOR.PATCH format or vMAJOR.MINOR format.
+
+In vMAJOR.MINOR format, the result will be patch version 0 of that version, for example `v1.2 -> v1.2.0`.
```
melekes: when bumping major version happens?

Liamsi: We currently do not bump the major version? We do not really follow https://semver.org and misuse the minor version for API breaking changes. I assume this will change as soon as we have a v1.0.0? But if I understand correctly the scripts can create such a release. It's just that major won't be bumped.

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2
 defaults: &defaults
   working_directory: /go/src/github.com/tendermint/tendermint
   docker:
-    - image: circleci/golang:1.12.0
+    - image: circleci/golang
```
Liamsi: 👍

.circleci/config.yml
greg-szabo: Last time I did that Jae rolled back my whole PR. Let's go with this for now.

.circleci/config.yml
greg-szabo: From bump-semver.py:
# Bump the release number of a semantic version number and print it. --version is required.
# Version is
# - vA.B.C, in which case vA.B.C+1 will be returned
# - vA.B.C-devorwhatnot in which case vA.B.C will be returned
# - vA.B in which case vA.B.0 will be returned

.circleci/config.yml
greg-szabo: Multiple reasons, although it would be nice to use it:
1. This is a "grab-and-move" step from Jenkins and Jenkins used Python scripts.
2. We used Python scripts because they don't need compiling. We don't want compiling in the middle of the release process. (We can overcome that by adding the goreleaser in the docker image so when release comes around, it's already available.)
3. Goreleaser works by deploying by tags. Tenderbot currently can't push tags. (No permission.) We work around that by doing the GitHub release which creates a tag as a side-effect.

The main reason for this step is to get off of Jenkins. We can improve the build process by introducing security releases in a next step and we can look at goreleaser there too. The goal now is to shut down Jenkins.

scripts/release_management/README.md
greg-szabo: The request was to keep version management in the developer's hand. vX.Y.Z, Z is automatically calculated using `bump-semver.py`. The rest is manual, If you open a v1.0 branch, the automation will trigger a v1.0.0 release. If you push something to that branch, automation will trigger a v1.0.1 release.

PR #3491, opened 2019-03-27
Update release/v0.31.1 to include #3488

Liamsi:
**Do not squash** to include commit: https://github.com/tendermint/tendermint/pull/3491/commits/6a45379cdf7cde8654ba6524ed408b7e9d8fa7cd

* [ ] Updated all relevant documentation in docs
* [ ] Updated all code comments where relevant
* [ ] Wrote tests
* [ ] Updated CHANGELOG_PENDING.md

PR #3503, opened 2019-03-28
docs: Update contributing guideliens

ebuchman:
Minor updates to reflect squash merging and how to prepare releases

* [ ] Wrote tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -105,9 +105,13 @@ removed from the header in rpc responses as well.
 
 ## Branching Model and Release
 
-All repos should adhere to the branching model: http://nvie.com/posts/a-successful-git-branching-model/.
+We follow a variant of [git flow](http://nvie.com/posts/a-successful-git-branching-model/).
 This means that all pull-requests should be made against develop. Any merge to
 master constitutes a tagged release.
+
+Note all pull requests should be squash merged except for merging to master and
+merging master back to develop. This keeps the commit history clean and makes it
+easy to reference the pull request where a change was introduced.
 
 ### Development Procedure:
 - the latest state of development is on `develop`
@@ -120,13 +124,13 @@ master constitutes a tagged release.
 ### Pull Merge Procedure:
 - ensure pull branch is based on a recent develop
 - run `make test` to ensure that all tests pass
-- merge pull request
+- squash merge pull request
 - the `unstable` branch may be used to aggregate pull merges before fixing tests
 
 ### Release Procedure:
 - start on `develop`
 - run integration tests (see `test_integrations` in Makefile)
-- prepare changelog:
+- prepare release in a pull request against develop (to be squash merged):
     - copy `CHANGELOG_PENDING.md` to top of `CHANGELOG.md`
     - run `python ./scripts/linkify_changelog.py CHANGELOG.md` to add links for
       all issues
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -105,10 +105,14 @@ removed from the header in rpc responses as well.
...
 ## Branching Model and Release
 
-All repos should adhere to the branching model: http://nvie.com/posts/a-successful-git-branching-model/.
+We follow a variant of [git flow](http://nvie.com/posts/a-successful-git-branching-model/).
 This means that all pull-requests should be made against develop. Any merge to
 master constitutes a tagged release.
 
+Note all pull requests should be squash merged except for merging to master and
+merging master back to develop. This keeps the commit history clean and makes it
+easy to reference the pull request where a change was introduced.
```
Liamsi: What if we want to include a PR that made it into develop *after* the release branch has been created? Squashing those changes will create another commit for the same changes. For example see: https://github.com/tendermint/tendermint/pull/3491 
If we squash merge this, it will create a commit that will make it to the release branch and then to master. The same changes exist under develop with a different commit.

CONTRIBUTING.md
```diff
@@ -135,23 +139,15 @@ master constitutes a tagged release.
       the changelog. To lookup an alias from an email, try `bash
       ./scripts/authors.sh <email>`
     - reset the `CHANGELOG_PENDING.md`
-- bump versions
-- push to release/vX.X.X to run the extended integration tests on the CI
-- merge to master
-- merge master back to develop
+    - bump versions
+- push latest develop with prepared release details to release/vX.X.X to run the extended integration tests on the CI
+- if necessary, make pull requests against release/vX.X.X and squash merge them
```
Liamsi: see: https://github.com/tendermint/tendermint/pull/3503/files#r269982992
There are edge cases where we do want to merge instead of squash (we can either disallow these cases or use merge).

CONTRIBUTING.md
xla: Can we enforce this through branch protection in github settings?

ebuchman: Merging develop to release branch should be done with a normal merge, not squash. You're correct here.

Ideally we minimize this. In the ideal case, the release branch is checked out directly from develop and merged to master with no more changes. 

I suppose this means we should continue to do all review and finalization on develop before we make the release branch, so that ideally there's no PRs against only the release branch. 

But that would cause develop to freeze while we're prepping the release, which isn't ideal, so there's some tension here.

I don't think it's a big deal to have some merges from develop into release. But we can also just retarget PRs to the release branch to minimize the number of times we have new stuff on develop that we want in the release after we've already made the release branch ...

ebuchman: > Can we enforce this through branch protection in github settings?

Which part?

I don't think we can enforce the squash merge rules :(

CONTRIBUTING.md
ebuchman: this should say "new pull requests, based on the release branch, against the release branch".

The only case we don't want to squash merge (afaik) is when we need to merge develop to the release branch. But in general we should be able to avoid that by basing those fixes on the release branch in the first place, yeh ?

Liamsi: >  in general we should be able to avoid that by basing those fixes on the release branch in the first place, yeh ?

Yes, makes sense. We should base PRs directly on the release branch if we are sure, we want to include them.

ebuchman:
Need to also specify changelog entry title for changes that affect top level stuff like docker, circle, etc. https://github.com/tendermint/tendermint/pull/3512#discussion_r270661332

PR #3504, opened 2019-03-28
libs: remove useless code in group

guagualvcha:
I am reading code about wal code, auto file package contains some confusing code that is useless

* [x] Wrote tests


Liamsi:
Added the breaking label as this breaks the go-API. Deleting a bunch of code, my favourite PRs 👍

guagualvcha:
Thank you for quick review.

guagualvcha:
> @guagualvcha Great change! Deleting code like no tomorrow.
> 
> Could you update the changelog?

sure

CHANGELOG_PENDING.md
```diff
@@ -9,6 +9,7 @@
 * Apps
 
 * Go API
+- [libs/autofile] Remove useless code in autofile package
```
```suggestion
- [libs/autofile] \#3504 Remove useless code in autofile package (@guagualvcha)
```

melekes: Would also be good to enumerate the removed public functions, namely: `Group#Search, Group#FindLast, GroupReader#ReadLine, GroupReader#PushLine`

Liamsi: Very good point. I've updated the suggestion above.

PR #3506, opened 2019-03-29
[ADR] ADR-36: blockchain reactor refactor spec

milosevic:
* [ ] Wrote tests


xla:
@milosevic Corrected the markdown and fixed up some Go indentation and naming inconsistencies.

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+
+const (
+	StateUnknown State = iota
+	StateFastSyncMode
+	StateConsensusMode
+)
+
+type ControllerState struct {
+	Height             int64            // the first block that is not committed
+	State              State            // mode of operation of the state machine
```
xla: As this is part of `ControllerState` and the enum does have the suffix `Mode` already it might be beneficial streamline the naming to:

```go
type Mode int

const (
	ModeUnknown State = iota
	ModeFastSync
	ModeConsensus
)

type ControllerState struct {
	Height             int64            // the first block that is not committed
	Mode               Mode             // mode of operation of the state machine
	PeerMap            map[ID]PeerStats // map of peer IDs to peer statistics
	MaxRequestPending  int64            // maximum height of the pending requests
	FailedRequests     []int64          // list of failed block requests
	PendingRequestsNum int              // number of pending requests
	Store              []BlockInfo      // contains list of downloaded blocks
	Executor           BlockExecutor    // store, verify and executes blocks
}
```

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+	Block  Block
+	Commit Commit
+	PeerID ID // a peer from which we downloading the corresponding Block and Commit
+}
+```
+
+The `Controller` is initialized by providing initial height (`startHeight`) from which it will start downloading blocks from peers.
+
+``` go
+func ControllerInit(state ControllerState, startHeight int64) ControllerState {
```
xla: Why does init get an exisiting state? Is it not supposed to create a fresh state for us?

If so a more idiomatic name would be `NewControllerState`.

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+	state.Height = startHeight
+	state.State = FastSyncMode
+	state.MaxRequestPending = startHeight - 1
+	state.PendingRequestsNum = 0
+}
+
+func ControllerHandle(event Event, state ControllerState) (ControllerState, Message, TimeoutTrigger, Error) {
```
xla: This could be more explicit:

``` go
handleEvent(state ControllerState, event Event) (ControllerState, Message, Timeout, Error)
```

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+
+### Implementation changes
+
+We assume the following system model for "fast sync" protocol: 
+
+* a node is connected to a random subset of all nodes that represents its peer set. Some nodes are correct and some might be faulty. We don't make assumptions about ratio of  
+  faulty nodes, i.e., it is possible that all nodes in some peer set are faulty. 
+* we assume that communication between correct nodes is synchronous, i.e., if a correct node `p` sends a message `m` to a correct node `q` at time `t`, then `q` will receive
+  message the latest at time `t+Delta` where `Delta` is a system parameter that is known. `Delta` is normally order of magnitude higher than the real communication delay 
+  (maximum) between correct nodes. Therefore if a correct node `p` sends a request message to a correct node `p` at time `t` and there is no the corresponding reply at time
```
guagualvcha: `Therefore if a correct node p sends a request message to a correct node p` to `Therefore if a correct node p sends a request message to a correct node q`

xla:
```suggestion
  (maximum) between correct nodes. Therefore if a correct node `p` sends a request message to a correct node `q` at time `t` and there is no the corresponding reply at time
```

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+					peerStats.PendingRequest = -1
+					state.PendingRequestsNum--
+					if event.CurrentHeight > peerStats.Height {
+						peerStats.Height = event.CurrentHeight
+					}
+					state.Store[event.Height] = BlockInfo{event.Block, event.Commit, event.PeerID}
+					state = verifyBlocks(state) // it can lead to event.PeerID being removed from peer list
+
+					if _, ok := state.PeerMap[event.PeerID]; ok {
+						msg = createBlockRequestMessage(state, event.PeerID, peerStats.Height)
```
guagualvcha: Previous blockchain reactor will send multi block requests in the same time, but new version only send new request until last block have received, will it slow down?

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+	done = false
+	for !done {
+		block = state.Store[height]
+
+		if block != nil {
+			// TODO: verify block.Block using block.Commit
+			verified = true
+
+			if verified {
+				block.Execute()
```
guagualvcha: step of `block.Execute()` takes most of the time of state transition. Since state transition is serial. We may find a way to let `block.Execute()` tasks run async.

zmanian:
I think it should be in scope for the blockchain reactor to sync with a known root of trust at a certain height. ie. syncing with **weak subjectivity**

docs/architecture/adr-036-blockchain-reactor-refactor.md
milosevic: With the new version, there is one pending request per peer. After a response (block) arrives, we will try to send a new request to the peer. This makes protocol more reactive towards responsive peers as the peers that send responses faster will be asked more. Having multiple pending requests (especially big number of requests as in the current version) is problematic as it opens the space for DDoS attacks. It is also not clear that having many pending requests is indeed more efficient. We will need to measure this. The main goal of this ADR is correctness and improved testability, and this should hopefully make performance improvements easier after.

docs/architecture/adr-036-blockchain-reactor-refactor.md
milosevic: Agree. Verifying and executing blocks are good candidates to be isolated in a separate routine but as stated above the idea is to focus on performance improvements after (if needed).

milosevic:
@zmanian In terms of specification, does this mean that "fast-sync" will have as an input parameter hash of the block (and maybe some additional info) at some height H provided by the known root of trust, so the protocol can verify it once it reaches height H?

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+	Height int64
+	PeerID p2p.ID
+}
+
+type NewBlock struct {
+	Height        int64
+	Block         Block
+	Commit        Commit
+	PeerID        ID
+	CurrentHeight int64
```
melekes: Rename to `CurrentPeerHeight`? Otherwise not clear what `CurrentHeight` refers to.

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+
+type BlockRequest struct {
+	Height int64
+	PeerID p2p.ID
+}
+
+type NewBlock struct {
+	Height        int64
+	Block         Block
+	Commit        Commit
```
melekes: Is this last commit? Doesn't Block contain it already?

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+	MessageBlockResponse
+	MessageStatusReport
+)
+
+type BlockRequestMessage struct {
+	Height int64
+	PeerID ID
+}
+
+type BlockResponseMessage struct {
```
melekes: This seems to be the copy of `NewBlock` event? What's the reason for duplication?

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+const (
+	TimeoutUnknown TimeoutTrigger = iota
+	TimeoutResponseTrigger
+	TimeoutTerminationTrigger
+)
+```
+
+The Controller state machine can be in two modes (states): `FastSyncMode` when
+it is trying to catch up with the network by downloading committed blocks,
+and `ConsensusMode` in which it executes Tendermint consensus protocol. We
```
melekes: Do we even need consensus mode in blockchain package?

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+	StateFastSyncMode
+	StateConsensusMode
+)
+
+type ControllerState struct {
+	Height             int64            // the first block that is not committed
+	State              State            // mode of operation of the state machine
+	PeerMap            map[ID]PeerStats // map of peer IDs to peer statistics
+	MaxRequestPending  int64            // maximum height of the pending requests
+	FailedRequests     []int64          // list of failed block requests
```
melekes: why do we need this? for testing?

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+	state.PeerMap[event.PeerID] = peerStats
+	return msg
+}
+
+func createBlockRequestMessage(state ControllerState, peerID ID, peerHeight int64) BlockRequestMessage {
+	msg = nil
+	blockNumber = -1
+
+	// exist request r in state.FailedRequests such that r <= peerHeight
+	if true {
```
melekes: ???

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+		blockNumber = r
+		delete(state.FailedRequests, r)
+	} else if state.MaxRequestPending < peerHeight {
+		state.MaxRequestPending++
```
melekes: shouldn't we set it to `peerHeight`?

melekes: also, I can't find the place where we decrease it after the peer with the max height disconnects.

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+			CurrentHeight: state.Height - 1,
+		}
+	}
+	state.PeerMap[event.PeerID] = peerStats
+	return msg
+}
+
+func createBlockRequestMessage(state ControllerState, peerID ID, peerHeight int64) BlockRequestMessage {
+	msg = nil
+	blockNumber = -1
```
melekes: rename to `BlockHeight`?

docs/architecture/adr-036-blockchain-reactor-refactor.md
milosevic: To verify block at height h we need lastCommit from the block h+1.

docs/architecture/adr-036-blockchain-reactor-refactor.md
milosevic: NewBlock is event that is created by the Executor upon receiving BlockResponseMessage.

docs/architecture/adr-036-blockchain-reactor-refactor.md
milosevic: We should help other peers sync-up even during consensus mode. Idea here is to emphasise that fast-sync is a mode of operation and not one shot functionality (as it is in the current code).

docs/architecture/adr-036-blockchain-reactor-refactor.md
milosevic: If a request failed, then it is added to this list so it can be fetched from some other peer.

docs/architecture/adr-036-blockchain-reactor-refactor.md
milosevic: Fixed in the latest version.

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+
+### Implementation changes
+
+We assume the following system model for "fast sync" protocol: 
+
+* a node is connected to a random subset of all nodes that represents its peer set. Some nodes are correct and some might be faulty. We don't make assumptions about ratio of  
+  faulty nodes, i.e., it is possible that all nodes in some peer set are faulty. 
+* we assume that communication between correct nodes is synchronous, i.e., if a correct node `p` sends a message `m` to a correct node `q` at time `t`, then `q` will receive
+  message the latest at time `t+Delta` where `Delta` is a system parameter that is known. `Delta` is normally order of magnitude higher than the real communication delay 
+  (maximum) between correct nodes. Therefore if a correct node `p` sends a request message to a correct node `q` at time `t` and there is no the corresponding reply at time
```
ancazamfir: there is no the corresponding reply at time
```suggestion
  (maximum) between correct nodes. Therefore if a correct node `p` sends a request message to a correct node `q` at time `t` and there is no corresponding reply at time
```

docs/architecture/adr-036-blockchain-reactor-refactor.md
milosevic: There is no notion of max peer height in this proposal (therefore it is never reset). Termination mechanism is different (based on number of pending requests, size of the peer set and termination timeout). I have added more comments in the latest version. MaxRequestPending is the maximum height for which request is made to some peer.

docs/architecture/adr-036-blockchain-reactor-refactor.md
ancazamfir: I think we need an overview of how all these data structures are used and when. Some control flow diagrams would be useful. 
Also, a part of this ADR should be the concurrency model/ design for scale/performance and testability. I believe this would be part of what you call the Executor task. I did some experiments with a version of the reactor poolRoutine that incorporates some of the Controller/ FSM changes. Sending one request per peer is slowing things down dramatically. Having hardcoded fixed limits may not be the best approach, maybe adaptive would work better. But in order to know, we need to have a good design (e.g. collect good stats, be able to abstract the things we sync, etc.) to understand what happens. Eyeing the logs is a nightmare. Not sure what we have today to collect operational data from nodes.

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+}
+
+type StatusReport struct {
+	PeerID ID
+	Height int64
+}
+
+type RemovePeer struct {
+	PeerID ID
+	Height int64
```
ancazamfir: Do we need Height here? In what conditions is this sent? Switch removes peer on error? Heartbeat loss?
Similar questions for the messages/ types below.

docs/architecture/adr-036-blockchain-reactor-refactor.md
ancazamfir: As an alternative the FSM handle function could just return the new state and any error that occurred while processing the event from old state. Any actions (like sending a message or triggering a timeout) would be part of the handle implementation (action functions). Where needed they can make calls to interface functions where we can plug in tester or real reactors. 

I was also wondering if instead or along with the code we could have a diagram or description on what happens  on peer removal and update. There is also a case currently where we detect slow peers. We should also put in place the design to have N pending requests, I don't think 1 will be enough. This complicates things quite a bit unfortunately.

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+	error = nil
+
+	switch state.State {
+	case StateConsensusMode:
+		switch event := event.(type) {
+		case EventBlockRequest:
+			msg = createBlockResponseMessage(state, event)
+			return state, msg, timeout, error
+		default:
+			error = "Only respond to BlockRequests while in ConsensusMode!"
```
ancazamfir: Just a note that this is not the case with current implementation.

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,406 @@
...
+			return state, msg, timeout, error
+		}
+
+	case StateFastSyncMode:
+		switch event := event.(type) {
+		case EventBlockRequest:
+			msg = createBlockResponseMessage(state, event)
+			return state, msg, timeout, error
```
ancazamfir: Here we are not in consensus mode and we reply to Block response messages. Also, this (the fast sync helper functionality) could be done outside the controller.

docs/architecture/adr-036-blockchain-reactor-refactor.md
```diff
@@ -0,0 +1,438 @@
...
+	error = nil
+
+	switch state.Mode {
+	case ModeConsensus:
+		switch event := event.(type) {
+		case EventBlockRequest:
+			msg = createBlockResponseMessage(state, event)
+			return state, msg, timeout, error
+		default:
+			error = "Only respond to BlockRequests while in ModeConsensus!"
```
ancazamfir:
```suggestion
			error = "While in ModeConsensus only respond to Block Requests"
```

What about the status requests? Where do we reply to those?

milosevic: It is part of the Heartbeat mechanism that is part of the Executor logic. I assumed that Executor send periodically to each peer StatusRequest message and expects StatusReport to arrive within response timeout.

docs/architecture/adr-036-blockchain-reactor-refactor.md
milosevic: We don't need Height, good point. It could be in case switch removes peer on error or in case StatusReport is not received on time.

docs/architecture/adr-036-blockchain-reactor-refactor.md
milosevic: I will try to add control diagram similar as we have for the current version. Maybe you can share source of the current diagram? Regarding performance concerns, extending this proposal to send multiple requests to a peer should be relatively straightforward. I will try to do it. I agree that number of pending requests per peer should be adaptive in ideal case but we need to estimate the complexity of such approach. We can maybe start with adding multiple pending requests (hardcoded value) to this proposal as this will let faster peer send multiple responses, and have them faster in position to serve new requests.

docs/architecture/adr-036-blockchain-reactor-refactor.md
ancazamfir: What's in a name... "peer"?  :)
A node that had previously answered to the status request broadcast? Or the peers the switch knows about (here we don't know the peer fast sync status)? 
If the former, will Executor keep track of peers in order to know who didn't answer?

milosevic: A peer is a node with which we have open connection (at the switch level). Executor sends status requests to all peers and need to manage timeouts. Therefore, we probably need to have peer lists at the Executor level also to able to determine who didn't responded on time.

xla:
@milosevic Given that we want to stick to consistent ADR numbering captured in #2313 this ADR should get the number 40. I amended the original issue accordingly.

Issue #2317, opened 2018-09-03
Allowing multiple entries in external_address for dual-stack and/or SNA


mdyring:
I'd like to advocate for allowing multiple entries in external_address, and ideally at the same time renaming the config option to external_address**es**.

The simple use-case is dual-stack nodes, where nodes can advertise both their IPv6 and IPv4 addresses. This seems generally useful and should increase robustness of the network.

For [Sentry Node Architecture](https://forum.cosmos.network/t/sentry-node-architecture-overview/454/5), using [/status RPC based peer discovery](https://forum.cosmos.network/t/sentry-node-architecture-overview/454/44), it is also helpful to let a node advertise both public IPv4 and private IPv4 addresses.

The use case here would be for sentries to talk between themselves via private IPv4 addresses, thus avoiding the bandwidth costs of hitting the Internet (which can be substantial as number of sentries increase).

Hybrid SNA architectures using, for instance, AWS EC2 for sentries and AWS Direct Connect for dedicated bandwidth to the physical validator, will also benefit from this feature, as the validator will be able to easily discover the private IPv4 of sentries, which can be routed via Direct Connect.

I believe this is not a major undertaking and can be boiled down to:
* Allowing multiple entries in the config option
* Exposing these multiple entries via /status RPC
* Make PEX send all entries for self, instead of one today

AFAIK the peer dial logic will only accept a single connection per node ID, so whichever gets established first should "win". This seems to align well with the above.

PR #3516, opened 2019-04-01
p2p: Expose actual listener address.

silasdavis:
This useful to discover dynamically assigned ports - i.e. when passing `localhost:0`

I suppose this is technically a breaking change if anyone is relying on the exact format of `Listeners`. Let me know if you would like me to list it as such.


silasdavis:
Perhaps this ought to be exposed here: https://github.com/tendermint/tendermint/blob/master/p2p/node_info.go#L82 

We might want something like
```go
type ListenAddresses struct {
	LocalAddress p2p.NetAddress
	ExternalAddress p2p.NetAddress
}
```

silasdavis:
The second commit takes a different approach - it adds `ExternalAddress` to `DefaultNodeInfo` and it updates the node's info to include the bound `ListenAddress`, which feels cleaner. I've dropped the `Listeners` function on `Node` - it seems like the information is better presented via `NodeInfo` and the presentation handled in RPC.

node/node.go
```diff
@@ -852,7 +847,7 @@ func makeNodeInfo(
 	txIndexer txindex.TxIndexer,
 	chainID string,
 	protocolVersion p2p.ProtocolVersion,
-) (p2p.NodeInfo, error) {
+) (p2p.DefaultNodeInfo, error) {
```
silasdavis: We don't lose anything by returning concrete

rpc/core/net.go
```diff
@@ -172,9 +172,12 @@ func NetInfo(ctx *rpctypes.Context) (*ctypes.ResultNetInfo, error) {
 	// TODO: Should we include PersistentPeers and Seeds in here?
 	// PRO: useful info
 	// CON: privacy
+
+	nodeInfo := p2pTransport.NodeInfo().(p2p.DefaultNodeInfo)
+
 	return &ctypes.ResultNetInfo{
 		Listening: p2pTransport.IsListening(),
-		Listeners: p2pTransport.Listeners(),
+		Listeners: []string{fmt.Sprintf("Listener(%v@%v)", nodeInfo.ListenAddress, nodeInfo.ExternalAddress)},
```
silasdavis: I thought I'd add the `ListenAddress` here, but if you would prefer to not make a technically breaking change to the RPC we could keep exact compatibility here and just return external address or listen address when external not set.

p2p/transport.go
```diff
@@ -231,6 +231,13 @@ func (mt *MultiplexTransport) Listen(addr NetAddress) error {
 	return nil
 }
 
+// Returns the actual address to which the transport's listener is bound.
+// Useful when the port has been dynamically assigned by the OS, e.g. tcp://localhost:0
+func (mt *MultiplexTransport) ListenAddress() string {
```
silasdavis: For purposes of `NodeInfo` we don't need to add this to the `Transport` interface, but we could...

node/node.go
```diff
@@ -150,7 +150,7 @@ type Node struct {
 	transport   *p2p.MultiplexTransport
 	sw          *p2p.Switch  // p2p connections
 	addrBook    pex.AddrBook // known peers
-	nodeInfo    p2p.NodeInfo
+	nodeInfo    p2p.DefaultNodeInfo
```
silasdavis: Given this is private to node it doesn't seem like it hurts. Also the RPC depends on `NodeInfo` being `DefaultNodeInfo` so seems to make sense to enforce the type.

xla:
@silasdavis @Liamsi What's the status of this PR? Still relevant? Does it need review, or any other support?

silasdavis:
@xla yes still relevant for me, or anyone else wanting to bind `localhost:0`

xla:
@silasdavis Cool, could you merge latest develop into this branch and then we take it from there?

p2p/node_info.go
```diff
@@ -78,8 +78,9 @@ type DefaultNodeInfo struct {
 
 	// Authenticate
 	// TODO: replace with NetAddress
-	ID_        ID     `json:"id"`          // authenticated identifier
-	ListenAddr string `json:"listen_addr"` // accepting incoming
+	ID_             ID     `json:"id"`            // authenticated identifier
+	ListenAddress   string `json:"listen_addr"`   // the local listen address bound on transport
+	ExternalAddress string `json:"external_addr"` // the advertised external address
```
melekes:
```suggestion
	ExternalAddress string `json:"external_address"` // the advertised external address
```

p2p/node_info.go
```diff
@@ -78,8 +78,9 @@ type DefaultNodeInfo struct {
 
 	// Authenticate
 	// TODO: replace with NetAddress
-	ID_        ID     `json:"id"`          // authenticated identifier
-	ListenAddr string `json:"listen_addr"` // accepting incoming
+	ID_             ID     `json:"id"`            // authenticated identifier
+	ListenAddress   string `json:"listen_addr"`   // the local listen address bound on transport
```
melekes:
```suggestion
	ListenAddress   string `json:"listen_address"`   // the local listen address bound on transport
```

node/node.go
```diff
@@ -862,6 +857,8 @@ func makeNodeInfo(
 		ID_:             nodeID,
 		Network:         chainID,
 		Version:         version.TMCoreSemVer,
+		ListenAddress:   config.P2P.ListenAddress,
```
melekes: what is the point of having `config.P2P.ListenAddress` then if it will be overwritten in any case?

p2p/node_info.go
```diff
@@ -121,7 +122,7 @@ func (info DefaultNodeInfo) Validate() error {
 	// ID is already validated.
 
 	// Validate ListenAddr.
-	_, err := NewNetAddressString(IDAddressString(info.ID(), info.ListenAddr))
+	_, err := NewNetAddressString(IDAddressString(info.ID(), info.ListenAddress))
```
melekes: should we also validate external address here?

p2p/node_info.go
silasdavis: Preferable, but a breaking change, will make

silasdavis:
> Although this seems like a step in the right direction +1, I am struggling to understand the need to separate listen and external addresses. The latter does not seem to play any role anymore (except that RPC is still returning it)

Not sure I follow - why would external addresses not be interesting to caller's of node info? I would still like to know the public external route-able address for a node would I not? Perhaps I am missing the sense you mean?

silasdavis:
Rebased

p2p/node_info.go
silasdavis: Probably annoying to do so since it will fail on DNS lookup. A node's external address may not be reachable from the node itself.

node/node.go
silasdavis: I do not think it is overwritten. We set:

```go
n.nodeInfo.ListenAddress = n.transport.ListenAddress()
```

Not the p2p config version - that's the one we dial. Under most circumstances they will be the same.

melekes:
Right now we'll advertise external address "mynode.dynamic-edge-dns.com" (if present) to the other node (when connected). The other node will then save "mynode.dynamic-edge-dns.com" to its address book and maybe send it to other peers OR use it to reconnect to the original peer.

With your change, we seem to lose that part because Listen address is "[::1]:54739" which is correct from meaning point of view, but incorrect from the original design point of view (i.e. why ListenAddr exists).

There's also https://github.com/tendermint/tendermint/pull/3646, which attempts to fix "[::1]:54739" advertised listen address (not reachable) by rewriting it with the node's remote IP.

p2p/node_info.go
melekes: We can assert that error

silasdavis:
I'm missing how my change alters the original design point of view. Either it is the same as originally designed or in the case of dynamic resolution it gives the port.

Can you give a worked example?

melekes:
If I am not mistaken, then

## before

1) we create nodeInfo and set ListenAddress to ExternalAddress or ListenAddress if ExternalAddress is empty
2) we send our nodeInfo to other nodes

## after

1) we create nodeInfo and set ListenAddress to ExternalAddress or ListenAddress if ExternalAddress is empty

https://github.com/tendermint/tendermint/blob/f12f345b81d3f811c3481ec40645f0799c17752c/node/node.go#L1001-L1007

2) we **overwrite** ListenAddress with listener's address

https://github.com/tendermint/tendermint/blob/f12f345b81d3f811c3481ec40645f0799c17752c/node/node.go#L686

3) we send our nodeInfo to other nodes

silasdavis:
Sorry about delay responding I was on holiday.

I see the issue. I think your first resolution is correct - PexReactor should use the `ExternalAddress`

There is a question of where do handle the fallthrough `ExternalAddress ?? ListenAddress`.

We might want to be defensive and implement it on the setting side, like we do now, but I think I like storing both. Consumer's should generally use the `ExternalAddress` if it is set, otherwise try the `ListenAddress`.

On the other hand if we think there should only be one address - the one this node deems other nodes should use, then maybe the actual bound listen address should be exposed programmatically directly off the transport?

Simplest way to expose this would be a method on `Node` - the placement for the `Listeners` method that is always a singleton, and a string.

I think that makes more sense, if it works for you I can rebase and rework.

silasdavis:
I decided this does indeed work better. We can access `node.ListenAddress()` from Go and we expose `listen_address` over rpc/net.

Both of which replacing the singleton `Listeners() []string` which didn't make much sense.

Avoid messing with `NodeInfo` and possibly leaking private net info of pex.

marbar3778:
Could you merge master into this branch and resolve conflicts. then we will review again.

erikgrinaker:
Supporting multiple listener addresses is useful for multi-homed and/or dual-stack hosts, both in terms of which addresses to bind to and also which addresses to advertise (see e.g. #2317).

The term "listen address" is generally taken to mean which local address(es) to bind to, and should be orthogonal to the advertised addresses. I would propose always using `ExternalAddress` as the publicly advertised address, and explicitly deriving this from `ListenAddr` at startup if it is not given.

Derivation of `ExternalAddress` from `ListenAddr` would involve e.g. replacing `0.0.0.0` with actual interface addresses, and replacing `:0` or undefined port numbers with actual port numbers. There should perhaps be some filtering of advertised addresses based on the receiver as well, e.g. only advertise non-routable addresses like `127.0.0.0/8` and `192.168.0.0/24` if the peer is on the same network.

I'll write up a more comprehensive proposal for cleaning up address handling in a bit.

erikgrinaker:
I don't think this is the right approach to fixing this problem - see e.g. https://github.com/tendermint/tendermint/pull/3516#issuecomment-587555940.

We are [planning](https://github.com/tendermint/tendermint/issues/4360) an overhaul of the P2P system later this year, which might involve a [migration to libp2p](https://github.com/tendermint/tendermint/issues/3696#988) and/or a [refactor of the P2P module](https://github.com/tendermint/tendermint/issues/2067). We will make sure these issues are addressed then.

Issue #2659, opened 2018-10-18
mempool: Move interface to mempool package


ValarDragon:
Currently the mempool interface is defined in the state package. This has a couple downsides, one is that the mempool package can't compile time check that it implements the desired interface. Additionally its non-intuitive where the interface is getting defined. This originally made more sense as it was to avoid state having to import the mempool package, but now it has to anyway for the precheck filters.

I propose that we make the mempool package a similar design pattern to the crypto package. We would have the interfaces defined in `/mempool`, along with the common precheck filters. We then have a separate package for every mempool implementation. This allows us to more easily experiment with / build out the new mempool data structures iteratively, and that nodes can elect to run different versions. It also helps for nodes who want to modify the mempool, which is a fairly natural thing to expect nodes to want to do.

 `state` and `node` would only know about the top level package import, and therefore not import any of the implementation detail-specific mempool code as well.

ValarDragon:
I think we should make the `/mempool` package have the interface, filters, and reactor, but the reactor calls a method `mempool.BroadcastRoutine(peer)` for gossiping. The mempool would then decode the message type. I think this will make it straightforward to upgrade the mempool data structure, and broadcasting method, in an isolated manner.

melekes:
Q:

- should we move CListMempool to a separate package (`/clist`)?
- what should we do with reactor's mempool interface?

currently, it's defined as:

```go
type MempoolWithWait interface {
	Mempool
	TxsFront() *clist.CElement
	TxsWaitChan() <-chan struct{}
	ReapMaxTxs(n int) types.Txs
}
```

maybe we should move TxsWaitChan to Mempool interface and write a simple iterator which won't expose *clist.CElement.

ReapMaxTxs is used by rpc in UnconfirmedTxs and by reactor for testing.

Liamsi:
> should we move CListMempool to a separate package (/clist)?

I think, yes. 

> what should we do with reactor's mempool interface?
> should we do anything with rpc/core/pipe#Mempool interface?

We should re-think if we actually need those as separate interfaces? The Clist mempool-thingy implements them all (?!) And it is the only implementation... Where it makes sense we can create smaller interfaces to capture functionality to keep the mempool interface small?

melekes:
https://github.com/tendermint/tendermint/pull/3524 was merged to develop.

Issue #3433, opened 2019-03-16
rpc: pipe should use a struct to wrap all the global objects


ebuchman:
rpc/core/pipe.go allows a number of high level objects to be registered for use with the rpc. 

This creates a lot of global variables in the pkg. It might be nicer if these were contained in a single RPCEnvironment (?) struct.

Mentioned recently in https://github.com/tendermint/tendermint/pull/3269#discussion_r263348542

melekes:
Should help us fix https://github.com/tendermint/tendermint/pull/3524#discussion_r278036869

PR #3310, opened 2019-02-14
reset triggered timeout precommit

milosevic:
Fixes https://github.com/tendermint/tendermint/issues/3302.

* [x] Updated all relevant documentation in docs
* [x] Updated all code comments where relevant
* [x] Wrote tests
* [x] Updated CHANGELOG_PENDING.md

Issue #1993, opened 2018-07-17
Tendermint breaks when BeginBlock (any ABCI call?) in ABCI server is not returning


srmo:
**BUG REPORT**

*Tendermint version*
0.22.4-c64a3c74 (self compiled: make dist)

*ABCI app*
JavaCounter - part of jABCI https://github.com/jTendermint/jabci

*Environment*
- OSX 10.13.6
- go version go1.10.2 darwin/amd64

**What happened**:
I've forced the BeginBlock implementation of the ABCI app to throw an exception instead of returning a ResponseBeginBlock. This was implemented such that the exception is thrown after the third block.
After that, no more Blocks are being built, broadcastTx etc still working.
No messages in tendermint log (default loglevel) could be found, hinting at a problem with beginBlock.
Ctrl+Cing tendermint also wasn't possible, i.e. it was just hanging there after sending the signal.

*What you expected to happen*:
Tendermint should timeout on an ABCI request, maybe?
Expected this to be fixed via 
https://github.com/tendermint/tendermint/pull/1891
https://github.com/tendermint/tendermint/issues/1890

*How to reproduce it* (as minimally and precisely as possible):
Just throw an exception in your ABCI app on BeginBlock or DeliverTx

**Logs (you can paste a small part showing an error or link a pastebin, gist, etc. containing more of the log file)**:
https://gist.github.com/srmo/a647b5db2a3c1a5268edc0ec62006b08
Contains the full flow, throwing exception at height 3->4
Then doing a broadcast_tx
Then doing Ctrl+C
Then doing multiple Ctrl+C

**Config (you can paste only the changes you've made)**:
only changed loglevel to "*:debug"

srmo:
One could think about guarding the ABCI app against exceptions, but then the question is what to return on ABCI calls.
It doesn't seem to make sense to return an error ABCI response on BeginBlock, right?
Maybe the ABCI app just have to shutdown in such cases?

melekes:
> It doesn't seem to make sense to return an error ABCI response on BeginBlock, right?

Why not?

melekes:
> Maybe the ABCI app just have to shutdown in such cases?

If it's something that an app can't handle, then - yes. Process supervisor should then restart the app and Tendermint.

srmo:
@melekes Yeah, I guess I was referring to the non-deterministic part. Those probably can't be solved via ABCI.

melekes:
[...]

PR #3524, opened 2019-04-02
mempool: move interface into mempool package

melekes:
Refs #2659

Breaking changes in the mempool package:

    [mempool] #2659 Mempool now an interface
        old Mempool renamed to CListMempool
        NewMempool renamed to NewCListMempool
        Option renamed to CListOption
        MempoolReactor renamed to Reactor
        NewMempoolReactor renamed to NewReactor
        unexpose TxID method
        TxInfo.PeerID renamed to SenderID
        unexpose MempoolReactor.Mempool

Breaking changes in the state package:

    [state] #2659 Mempool interface moved to mempool package
        MockMempool moved to top-level mock package and renamed to Mempool

Non Breaking changes in the node package:

    [node] #2659 Add Mempool method, which allows you to access mempool

* [ ] ~~Wrote tests~~

====

https://en.wikipedia.org/wiki/Nighthawks
_Nighthawks by Edward Hopper_


mempool/reactor.go
```diff
@@ -31,13 +31,25 @@ const (
 	maxActiveIDs = math.MaxUint16
 )
 
+// MempoolWithWait extends the standard Mempool interface to allow reactor to
+// wait for transactions and iterate on them once there are any. Also it
+// includes ReapMaxTxs function, which is useful for testing.
+//
+// UNSTABLE
+type MempoolWithWait interface {
```
melekes: better name?

melekes:
https://github.com/tendermint/tendermint/issues/2659#issuecomment-479084707

mempool/reactor.go
melekes: https://github.com/tendermint/tendermint/issues/2659#issuecomment-479084707

blockchain/reactor_test.go
```diff
@@ -13,6 +13,7 @@ import (
 	cmn "github.com/tendermint/tendermint/libs/common"
 	dbm "github.com/tendermint/tendermint/libs/db"
 	"github.com/tendermint/tendermint/libs/log"
+	memplmock "github.com/tendermint/tendermint/mempool/mock"
```
Liamsi: I've seen this several times but never understood why we alias  mocks if there are no name clashes/conflicts. Couldn't this do without the alias. Then uses of this would read `mock.Mempool`.

rpc/core/pipe.go
```diff
@@ -49,6 +49,14 @@ type peers interface {
 	Peers() p2p.IPeerSet
 }
 
+// Mempool extends the standard Mempool interface to allow getting
+// total txs size. ReapMaxTxs is used by UnconfirmedTxs to reap N transactions.
+type Mempool interface {
```
Liamsi: Still reading through the code but are there are several implementations of this interface? 

OK, seems like there is only one implementation. Namely, `CListMempool` which sits in the `mempool` package. Shouldn't this interface be in the mempool package then?
Or shouldn't `ReapMaxTxs(n int) types.Txs` be part of the standard mempool interface? Looks like any actual (non-mock) mempool implementation has it anyways? At least it is repeated in the `MempoolWithWait` interface.

mempool/reactor.go
Liamsi: > better name?

`WithWait`. If you use it somewhere, it would be `mempool.WithWait`then.

mempool/reactor.go
```diff
@@ -31,13 +31,25 @@ const (
...
+	Mempool
+	TxsFront() *clist.CElement
+	TxsWaitChan() <-chan struct{}
+	ReapMaxTxs(n int) types.Txs
```
Liamsi: I find it a bit odd to introduce this in a public interface if this *only* for testing. This is not test-driven development IMHO (more like test-only API design).

OK, `ReapMaxTxs` is used in testing *and* in `UnconfirmedTxs` (the latter is the only actual use-case). This method should maybe be part of the standard Mempool then?

state/execution_test.go
```diff
@@ -16,10 +16,10 @@ import (
 	cmn "github.com/tendermint/tendermint/libs/common"
 	dbm "github.com/tendermint/tendermint/libs/db"
 	"github.com/tendermint/tendermint/libs/log"
-	tmtime "github.com/tendermint/tendermint/types/time"
-
+	memplmock "github.com/tendermint/tendermint/mempool/mock"
```
Liamsi: Same, why alias this? No conflict and `mock.Mempool` reads just fine.

rpc/core/pipe.go
Liamsi: It's confusing

```go
type MempoolWithWait interface {
	Mempool
	TxsFront() *clist.CElement
	TxsWaitChan() <-chan struct{}
	ReapMaxTxs(n int) types.Txs
}
```
and 
```go
type Mempool interface {
	mempl.Mempool
	ReapMaxTxs(n int) types.Txs
	TxsBytes() int64
}
```

As mentioned above `ReapMaxTxs` should live in the standard mempool. OR, if for some reason we want to keep the standard mempool interface small. Then, introduce yet another interface, e.g.

```go
type TxReaper interface {
  ReapMaxTxs(n int) types.Txs
}
```
and include this in both interfaces?

blockchain/reactor_test.go
melekes: Do you think it would make sense to move all the mocks into one top-level package?

melekes:
I think I went into the rabbit hole 🐇

mempool/mempool.go
```diff
@@ -1,26 +1,34 @@
...
-	"github.com/tendermint/tendermint/libs/log"
-	"github.com/tendermint/tendermint/proxy"
 	"github.com/tendermint/tendermint/types"
 )
 
+// Mempool defines the mempool interface.
+// Updates to the mempool need to be synchronized with committing a block so
+// apps can reset their transient state on Commit.
+type Mempool interface {
+	Lock()
```
silasdavis: ++ glad to see this is still here - we need to be able to claim this lock to avoid issues around recheck/commit.

See further: https://github.com/hyperledger/burrow/blob/develop/consensus/tendermint/abci/app.go#L321-L344

mempool/clist_mempool.go
```diff
@@ -0,0 +1,712 @@
...
+
+//--------------------------------------------------------------------------------
+
+// txKey is the fixed length array sha256 hash used as the key in maps.
+func txKey(tx types.Tx) [sha256.Size]byte {
+	return sha256.Sum256(tx)
+}
+
+// txID is the hex encoded hash of the bytes as a types.Tx.
+func txID(tx []byte) string {
```
melekes: same here. seems like an implementation detail

mempool/mempool.go
```diff
@@ -1,26 +1,86 @@
...
+
+	// TxsWaitChan returns a channel to wait on transactions. It will be closed
+	// once the mempool is not empty (ie. the internal `mem.txs` has at least one
+	// element)
+	TxsWaitChan() <-chan struct{}
+
+	// TxsFront returns the first transaction in the ordered list for peer
+	// goroutines to call .NextWait() on.
+	// FIXME: leaking implementation details!
+	TxsFront() *clist.CElement
```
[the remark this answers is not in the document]
ebuchman: can't it just return *mempoolTx?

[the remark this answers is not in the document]
ebuchman: or is that the same problem?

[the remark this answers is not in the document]
melekes: No, because later in the broadcastTxRoutine loop we're calling `next.WaitNextChan` followed by `next.Next()`. I don't think we want either of these methods in the MempoolTx interface.

What we want - **a simple interface for Reactor to subscribe/get txs from the mempool, so it can broadcast them to peers**.

There are a couple of ways to go about it:

### 1

```go
type Iterator interface {
  Next() bool
  Tx() MempoolTx
```

### 2

```go
TxsChan() <-chan MempoolTx
```
which is used to distribute txs to peers. Exposing a single channel should allow us to remove `TxsWaitChan() <-chan struct{}` method.

mempool/reactor.go
```diff
@@ -31,13 +31,13 @@ const (
...
-// MempoolReactor handles mempool tx broadcasting amongst peers.
+// Reactor handles mempool tx broadcasting amongst peers.
 // It maintains a map from peer ID to counter, to prevent gossiping txs to the
 // peers you received it from.
-type MempoolReactor struct {
+type Reactor struct {
 	p2p.BaseReactor
 	config  *cfg.MempoolConfig
-	Mempool *Mempool
+	mempool Mempool
```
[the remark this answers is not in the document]
silasdavis: Not sure the reasoning, but as mentioned above we need the Mempool locker and until you added `node.Mempool()` (in this PR)  `node.MempoolReactor().Mempool` was the only way to get it...

Come to think of it I think it was the only way to get `CheckTx` which I've always found handy :)

blockchain/reactor_test.go
Liamsi: Might be worth thinking about this. Not sure if that would be a good idea. But no, I only meant that we do not need aliases when there are no conflicts. No big thing tho.

mempool/mempool.go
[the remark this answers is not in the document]
melekes: On the second thought, maybe it's okay for MempoolReactor to require the CList implementation for now. It can be a follow-up PR.

melekes:
Tests failed with:

```
--- FAIL: TestStartNextHeightCorrectly (0.22s)
    assertions.go:247: ^M                          ^M   Error Trace:    state_test.go:1329
        ^M      Error:          Should be true
        ^M      Test:           TestStartNextHeightCorrectly
```

seems not related.

Liamsi:
This failure was seen before in: https://github.com/tendermint/tendermint/pull/3310#issuecomment-464301327

mempool/clist_mempool.go
```diff
@@ -0,0 +1,712 @@
...
+	metrics *Metrics
+}
+
+var _ Mempool = &CListMempool{}
+
+// Option sets an optional parameter on the mempool.
+type Option func(*CListMempool)
+
+// NewCListMempool returns a new mempool with the given configuration and connection to an application.
+func NewCListMempool(
```
melekes: add a changelog entry `NewMempool renamed to NewCListMempool`

mempool/clist_mempool.go
```diff
@@ -0,0 +1,712 @@
...
+
+	logger log.Logger
+
+	metrics *Metrics
+}
+
+var _ Mempool = &CListMempool{}
+
+// Option sets an optional parameter on the mempool.
+type Option func(*CListMempool)
```
melekes: rename to `CListOption`?

mempool/clist_mempool.go
```diff
@@ -0,0 +1,712 @@
...
+
+func (mem *CListMempool) TxsWaitChan() <-chan struct{} {
+	return mem.txs.WaitChan()
+}
+
+// It blocks if we're waiting on Update() or Reap().
+// cb: A callback from the CheckTx command.
+//     It gets called from another goroutine.
+// CONTRACT: Either cb will get called, or err returned.
+func (mem *CListMempool) CheckTx(tx types.Tx, cb func(*abci.Response)) (err error) {
```
melekes: remove?

mempool/mempool.go
```diff
@@ -1,26 +1,86 @@
...
+
+	// ReapMaxTxs reaps up to max transactions from the mempool.
+	// If max is negative, there is no cap on the size of all returned
+	// transactions (~ all available transactions).
+	ReapMaxTxs(max int) types.Txs
+
+	// TxsWaitChan returns a channel to wait on transactions. It will be closed
+	// once the mempool is not empty (ie. the internal `mem.txs` has at least one
+	// element)
+	TxsWaitChan() <-chan struct{}
```
melekes: remove for now.

mempool/mempool.go
melekes: remove for now.

mempool/mempool.go
```diff
@@ -1,26 +1,86 @@
...
+
+	// EnableTxsAvailable initializes the TxsAvailable channel, ensuring it will
+	// trigger once every height when transactions are available.
+	EnableTxsAvailable()
+
+	// Size returns the number of transactions in the mempool.
+	Size() int
+
+	// TxsBytes returns the total size of all txs in the mempool.
+	TxsBytes() int64
```
melekes: better abstraction for stats?

node/node.go
```diff
@@ -329,11 +330,10 @@ func NewNode(config *cfg.Config,
 		mempl.WithPostCheck(sm.TxPostCheck(state)),
 	)
 	mempoolLogger := logger.With("module", "mempool")
-	mempool.SetLogger(mempoolLogger)
 	if config.Mempool.WalEnabled() {
 		mempool.InitWAL() // no need to have the mempool wal during tests
 	}
-	mempoolReactor := mempl.NewMempoolReactor(config.Mempool, mempool)
+	mempoolReactor := mempl.NewReactor(config.Mempool, mempool)
```
melekes: add changelog entry

node/node.go
```diff
@@ -799,11 +800,16 @@ func (n *Node) ConsensusReactor() *cs.ConsensusReactor {
...
 
-// MempoolReactor returns the Node's MempoolReactor.
-func (n *Node) MempoolReactor() *mempl.MempoolReactor {
+// MempoolReactor returns the Node's mempool reactor.
+func (n *Node) MempoolReactor() *mempl.Reactor {
 	return n.mempoolReactor
 }
 
+// Mempool returns the Node's mempool.
+func (n *Node) Mempool() mempl.Mempool {
```
melekes: add changelog entry

melekes: add changelog entry

mempool/mempool.go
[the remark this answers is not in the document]
melekes: https://github.com/tendermint/tendermint/issues/3589

CHANGELOG_PENDING.md
```diff
@@ -9,12 +9,24 @@
...
+  * old Mempool renamed to CListMempool
+  * NewMempool renamed to NewCListMempool
+  * Option renamed to CListOption
+  * MempoolReactor renamed to Reactor
+  * NewMempoolReactor renamed to NewReactor
+  * unexpose TxID method
+  * TxInfo.PeerID renamed to SenderID
+  * unexpose MempoolReactor.Mempool
+- [state] \#2659 Mempool interface moved to mempool package
+  * MockMempool moved to top-level mock package and renamed to Mempool
```
Liamsi: Can you mark all symbols as code? `NewMempool` instead of NewMempool etc?

mempool/reactor_test.go
```diff
@@ -88,9 +88,9 @@ func waitForTxs(t *testing.T, txs types.Txs, reactors []*MempoolReactor) {
 }
 
 // wait for all txs on a single mempool
-func _waitForTxs(t *testing.T, wg *sync.WaitGroup, txs types.Txs, reactorIdx int, reactors []*MempoolReactor) {
+func _waitForTxs(t *testing.T, wg *sync.WaitGroup, txs types.Txs, reactorIdx int, reactors []*Reactor) {
```
Liamsi: Nit: Can we rename this to `waitForTxsFromMempool` or sth alike?

rpc/client/rpc_test.go
```diff
@@ -252,7 +252,8 @@ func TestAppCalls(t *testing.T) {
 func TestBroadcastTxSync(t *testing.T) {
 	require := require.New(t)
 
-	mempool := node.MempoolReactor().Mempool
+	// TODO (melekes): use mempool which is set on RPC rather than getting it from node
```
Liamsi: Is there an issue to track this?

node/node.go
```diff
@@ -157,9 +157,10 @@ type Node struct {
 	// services
 	eventBus         *types.EventBus // pub/sub for services
 	stateDB          dbm.DB
-	blockStore       *bc.BlockStore         // store the blockchain to disk
-	bcReactor        *bc.BlockchainReactor  // for fast-syncing
-	mempoolReactor   *mempl.MempoolReactor  // for gossipping transactions
+	blockStore       *bc.BlockStore        // store the blockchain to disk
+	bcReactor        *bc.BlockchainReactor // for fast-syncing
+	mempoolReactor   *mempl.Reactor        // for gossipping transactions
+	mempool          *mempl.CListMempool
```
Liamsi: Why is this a pointer while in `rpc/core/pipe.go` it was changed to not be a pointer?

CHANGELOG_PENDING.md
```diff
@@ -9,12 +9,24 @@
...
+  * `TxInfo.PeerID` renamed to `SenderID`
+- [state] \#2659 `Mempool` interface moved to mempool package
+  * `MockMempool` moved to top-level mock package and renamed to `Mempool`
 
 * Blockchain Protocol
 
 * P2P Protocol
 
 ### FEATURES:
+- [node] \#2659 Add Mempool method, which allows you to access mempool
```
```suggestion
- [node] \#2659 Add `node.Mempool()` method, which allows you to access mempool
```

mempool/doc.go
```diff
@@ -0,0 +1,24 @@
...
+// To manage these goroutines, there are three methods of locking.
+// 1. Mutations to the linked-list is protected by an internal mtx (CList is goroutine-safe)
+// 2. Mutations to the linked-list elements are atomic
+// 3. CheckTx() calls can be paused upon Update() and Reap(), protected by .proxyMtx
+
+// Garbage collection of old elements from mempool.txs is handlde via
+// the DetachPrev() call, which makes old elements not reachable by
+// peer broadcastTxRoutine() automatically garbage collected.
+
+// TODO: Better handle abci client errors. (make it automatically handle connection errors)
```
Liamsi: Same: is there an issue tracking this?

mempool/clist_mempool.go
Liamsi: Probably a good idea to discuss this in a separate issue.

melekes: https://github.com/tendermint/tendermint/issues/3590

rpc/client/rpc_test.go
melekes: https://github.com/tendermint/tendermint/issues/3433

mempool/doc.go
melekes: I don't think so. There's https://github.com/tendermint/tendermint/issues/1993, which is one of the bugs

node/node.go
melekes: Because node's mempool calls InitWAL/CloseWAL, but I don't want to add these methods to Mempool interface, which is already big

melekes: I created `mempoolWithWAL` interface

melekes: Made them a part of Mempool interface for now.

1. WAL is an implementation details and shouldn't really be a part of the interface.
2. But we need a way to init/close associated resources
3. And we need a way to configure them (wal path, etc.)

so we should either make node.Mempool a concrete type OR make InitWAL/CloseWAL/ConfigureWAL a part of any mempool interface

node/node.go
```diff
@@ -617,6 +614,7 @@ func NewNode(config *cfg.Config,
 		blockStore:       blockStore,
 		bcReactor:        bcReactor,
 		mempoolReactor:   mempoolReactor,
+		mempool:          mempool,
```
golangcibot: cannot use mempool (variable of type *mempool.Mempool) as mempool.Mempool value in struct literal: missing method CheckTx (from `typecheck`)

consensus/mempool_test.go
```diff
@@ -11,14 +11,15 @@ import (
 
 	"github.com/tendermint/tendermint/abci/example/code"
 	abci "github.com/tendermint/tendermint/abci/types"
+	mempl "github.com/tendermint/tendermint/mempool"
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

melekes:
> This doesn't change any functionality right, just renames and file re-org?

yes

mempool/clist_mempool.go
```diff
@@ -0,0 +1,714 @@
+package mempool
```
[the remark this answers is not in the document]
melekes: No actual changes

Issue #2840, opened 2018-11-14
consensus: proposer-based timestamps


ebuchman:
Current mechanism for BFT time includes a timestamp in every precommit, and just takes the median. 

While simple and live despite arbitrary clock drift, it has some problems:

- can't be used for [signature aggregation](https://github.com/tendermint/tendermint/issues/1319) (since each vote has a unique timestamp component)
- gives +1/3 of the validator set complete control over the timestamp (see https://github.com/tendermint/tendermint/issues/2653)

An alternative, originally proposed in https://github.com/tendermint/tendermint/pull/1615, is to let the proposer use their own local timestamp, and for validators to only vote for the block if the timestamp is within some range of their own clock. This requires all nodes to keep their clocks reasonably closely synced, and was originally rejected/postponed for this reason. But in order to solve the above two issues, we should reconsider this solution. We can also include an option to disable it, so folks that don't want to worry about synced clocks won't have to (but of course it means they might not have accurate timestamps in blocks).

We should write a new ADR for this and do some more analysis before fully moving forward with it.

Note there is also a follow up issue from the last BFT-time PR: https://github.com/tendermint/tendermint/issues/2309

mdyring:
Slack chatter:

**bucky**
One important thing to note about #2840 (part of v0.33 milestone) is that it will introduce a clock synchronization assumption into Tendermint - ie. validators will be expected to run NTP or equivalent so that their internal clocks are roughly synchronized, else there will be a liveness failure. This is quite different from Tendermint now, and is the reason we didn’t move forward with this last summer when we introduced BFT time. Currently we just take the median of the timestamps, so there is no clock sync assumption, but the new approach will make it so proposed blocks are only valid if the timestamp is within some delta around each validators local timestamp. If you have concerns with this, please comment in the issue! (edited)

**Ethan Frey**
As long as delta is reasonably large (like 2 minutes, not 2 seconds), this seems like a reasonable assumption to me.

Tendermint halts with network failures, so there are plenty of assumptions about connectivity already

**mdyring**
Tried to read up on this, perhaps below was already addressed:

What happens once validators (out of malice or clock drift), starts proposing blocks that are right at the upper boundary (future) of true time+delta? Lets call this timestamp F.

Next proposer in turn might have a more accurate view of time, and won't be able to propose a block with timestamp before F (as time must move forward), resulting in missed block.

[...]

PR #3529, opened 2019-04-03
docs: fix block.Header.Time description

### LINKED ISSUE #3514: docs: Fix incorrect note about header.Time in ABCI data types being the proposer's local time
**Tendermint version** `v0.31.1` with the patch from #3118

**ABCI app**: [Ercoin](https://gitlab.com/Ercoin/ercoin)

**What happened**:

If `create_empty_blocks` is set to `false` and a block is created, then time of that block is not the current time, but it refers to the time at which the previous block was created.

Block time should be current. Past time may make transactions invalid if their validity is time-dependent. If a transaction is created with a current timestamp, then a block in which that transaction is included must not contain a lower time value in block header.

1. Set `create_empty_blocks` to `false` and `create_empty_blocks_interval` to a reasonably high value.
2. Observe the time of the block header.

melekes:
It's not proposer local time anymore, but a weighted median

Fixes #3514

* [ ] ~~Wrote tests~~

### DIFF
--- a/docs/spec/abci/abci.md
+++ b/docs/spec/abci/abci.md
@@ -347,8 +347,10 @@ Commit are included in the header of the next block.
   - `Version (Version)`: Version of the blockchain and the application
   - `ChainID (string)`: ID of the blockchain
   - `Height (int64)`: Height of the block in the chain
-  - `Time (google.protobuf.Timestamp)`: Time of the block. It is the proposer's
-    local time when block was created.
+  - `Time (google.protobuf.Timestamp)`: Time of the previous block.
+    For heights > 1, it's the weighted median of the timestamps of the valid
+    votes in the block.LastCommit.
+    For height == 1, it's genesis time.
   - `NumTxs (int32)`: Number of transactions in the block
   - `TotalTxs (int64)`: Total number of transactions in the blockchain until
     now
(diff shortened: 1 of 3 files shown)


docs/spec/abci/abci.md
```diff
@@ -347,8 +347,10 @@ Commit are included in the header of the next block.
   - `Version (Version)`: Version of the blockchain and the application
   - `ChainID (string)`: ID of the blockchain
   - `Height (int64)`: Height of the block in the chain
-  - `Time (google.protobuf.Timestamp)`: Time of the block. It is the proposer's
-    local time when block was created.
+  - `Time (google.protobuf.Timestamp)`: Time of the previous block.
```
milosevic: Isn't this time of the block (not previous block)?

melekes: No, we're using LastCommit to calculate median time https://github.com/tendermint/tendermint/blob/f965a4db15796dfca4e504d93eaa83760ced4af2/state/state.go#L149

melekes: We should've renamed `Time` to `LastTime` or something. But now given we're switching to proposer's local time (basically reverting the changes), it does not worth it.

milosevic: We use LastCommit of block h to calculate time for block h+1. But it's still time of the current block (and shouldn't be renamed to LastTime).

b00f:
It was good if the label also changed to `PreviousBlockTime`. A little bit confusing.

melekes:
We're going to switch back to proposer-based timestamp https://github.com/tendermint/tendermint/issues/2840. Sorry for confusion

PR #3534, opened 2019-04-08
rpc: Add support for batched requests/responses

thanethomson:
Continues from #3280 in building support for batched requests/responses in the JSON RPC (as per issue #3213).

* [x] Wrote tests


rpc/client/httpclient.go
```diff
@@ -48,15 +49,53 @@ func NewHTTP(remote, wsEndpoint string) *HTTP {
...
+// CancelBatch will cancel a batch operation and return the number of pending
+// operations that were cancelled.
+func (c *HTTP) CancelBatch() int {
+	c.batching = false
+	return c.rpc.ClearEnqueued()
+}
+
+func (c *HTTP) SendBatch() ([]interface{}, error) {
+	if !c.batching {
+		return nil, errors.Errorf("Not currently in batch mode: use StartBatch() to start enqueuing requests")
```
Liamsi: Nit: 
> Error strings should not start with a capital letter because they'll often be prefixed before printing:

https://github.com/golang/go/wiki/Errors#errors
```suggestion
		return nil, errors.Errorf("currently not in batch mode: use StartBatch() to start enqueuing requests")
```

rpc/client/httpclient.go
```diff
@@ -48,15 +49,53 @@ func NewHTTP(remote, wsEndpoint string) *HTTP {
...
 }
 
 var _ Client = (*HTTP)(nil)
 
+// StartBatch will cause the HTTP client to go into batching mode, which will
+// enqueue subsequent requests (`c.Status()`, `c.ABCIQuery`,
+// `c.BroadcastTxSync`, etc.) instead of sending them immediately. These
+// requests will be enqueued until such time that they are either sent by
+// calling `c.SendBatch()`, or cleared by calling `c.CancelBatch()`.
+func (c *HTTP) StartBatch() {
```
Liamsi: Maybe add a comment that StartBatch isn't thread-safe/not meant to be used from go-rountines?

rpc/client/httpclient.go
```diff
@@ -48,15 +49,53 @@ func NewHTTP(remote, wsEndpoint string) *HTTP {
...
+
+func (c *HTTP) SendBatch() ([]interface{}, error) {
+	if !c.batching {
+		return nil, errors.Errorf("Not currently in batch mode: use StartBatch() to start enqueuing requests")
+	}
+	c.batching = false
+	return c.rpc.SendEnqueued()
+}
+
+func (c *HTTP) callRPC(method string, params map[string]interface{}, result interface{}) (interface{}, error) {
```
Liamsi: Maybe that is a slightly better name that makes clear what is going on here?
```suggestion
func (c *HTTP) callOrBatchRPC(method string, params map[string]interface{}, result interface{}) (interface{}, error) {
```

rpc/client/rpc_test.go
```diff
@@ -441,3 +443,65 @@ func TestTxSearch(t *testing.T) {
 		require.Len(t, result.Txs, 0)
 	}
 }
+
```
Liamsi: IMHO a test which tests the error'ing or rather un-expected paths would be great: 
 - calling `SendBatch` without beeing in batch mode
 - calling `CancelBatch` without having anything in the enqueued 

(OK, looks like these are the only cases not covered by the tests 👍)

rpc/lib/server/http_server.go
```diff
@@ -111,6 +111,23 @@ func WriteRPCResponseHTTP(w http.ResponseWriter, res types.RPCResponse) {
...
+// WriteRPCResponseArrayHTTP will do the same as WriteRPCResponseHTTP, except it
+// can write arrays of responses for batched request/response interactions via
+// the JSON RPC.
+func WriteRPCResponseArrayHTTP(w http.ResponseWriter, res []types.RPCResponse) {
+	if len(res) == 1 {
+		WriteRPCResponseHTTP(w, res[0])
+	} else {
+		jsonBytes, err := json.MarshalIndent(res, "", "  ")
+		if err != nil {
+			panic(err)
```
melekes: maybe return error?

rpc/lib/client/http_client.go
```diff
@@ -83,11 +83,19 @@ func makeHTTPClient(remoteAddr string) (string, *http.Client) {
...
+
 // JSONRPCClient takes params as a slice
 type JSONRPCClient struct {
-	address string
-	client  *http.Client
-	cdc     *amino.Codec
+	address  string
+	client   *http.Client
+	cdc      *amino.Codec
+	reqBatch []jsonRPCBufferedRequest
```
melekes: shouldn't we initialize it in NewJSONRPCClient?

rpc/client/httpclient.go
melekes: Agree. Was our client thread-safe before? Can't find any documentation. If yes, then we should try to preserve this property.

In any case, we should document whenever it's thread-safe to use or not (for both HTTP and Local clients).

rpc/client/httpclient.go
thanethomson: Done 👍

rpc/client/httpclient.go
thanethomson: I've refactored the batching functionality to be, IMHO, a little more user-friendly when it comes to concurrent usage. This would allow for multiple batches to be created across different goroutines, all executed by the same client without interfering with each other.

The new usage would look something like:

```go
c := client.NewHTTP("127.0.0.1:26657", "/websocket")

// Generate two random key/value pairs and their equivalent transactions
tx1 := makeTxKV()
tx2 := makeTxKV()

// Create a new batch
batch := c.NewBatch()

// Queue up the first transaction
batch.BroadcastTxCommit(tx1)
// Queue up the second transaction
batch.BroadcastTxCommit(tx2)

// Send the whole batch
batch.Send()
```

rpc/client/rpc_test.go
thanethomson: Added some tests now that should cover these paths.

rpc/lib/server/http_server.go
thanethomson: There's code in [`rpc/lib/server/http_server.go`](https://github.com/tendermint/tendermint/blob/8dd2798cfcc2fd57ed5c90132f1bfaa4c81d932a/rpc/lib/server/http_server.go#L148) (not my code) that catches panics and returns 500 Internal Server errors.

Should we refactor all of that?

rpc/lib/client/http_client.go
thanethomson: Yes, I can't understand how this wasn't caught by the tests... Very strange. Fixed now - thanks! 👍

rpc/client/httpclient.go
thanethomson: The refactoring, I think, should clarify this now?

rpc/lib/server/http_server.go
melekes: no, I think that's fine. But can you please add

```
err := w.Write(jsonBytes)
if err != nil {
  panic(err)
}
```

below and in WriteRPCResponseHTTP function and remove `// nolint: errcheck`?

rpc/client/httpclient.go
Liamsi: Yes, thank you!

rpc/lib/server/http_server.go
thanethomson: Done 👍

rpc/client/examples_test.go
```diff
@@ -0,0 +1,89 @@
...
+	results, err := batch.Send()
+	if err != nil {
+		panic(err)
+	}
+	// `results` will now contain the deserialized result objects from each of
+	// the sent BroadcastTxCommit transactions
+
+	// Do something useful with the results here
+	for _, result := range results {
+		fmt.Println(result)
```
Liamsi: if the results always come in the same order, you could use `Output` to show what the example will return:
https://blog.golang.org/examples

rpc/lib/client/http_client.go
```diff
@@ -202,6 +314,30 @@ func unmarshalResponseBytes(cdc *amino.Codec, responseBytes []byte, result inter
 	return result, nil
 }
 
+func unmarshalResponseBytesArray(cdc *amino.Codec, responseBytes []byte, results []interface{}) ([]interface{}, error) {
+	var err error
+	var responses []types.RPCResponse
+	err = json.Unmarshal(responseBytes, &responses)
+	if err != nil {
+		return nil, errors.Errorf("Error unmarshalling rpc response: %v", err)
```
Liamsi:
```suggestion
		return nil, errors.Errorf("error unmarshalling rpc response: %v", err)
```

rpc/lib/client/http_client.go
```diff
@@ -202,6 +314,30 @@ func unmarshalResponseBytes(cdc *amino.Codec, responseBytes []byte, result inter
...
+	}
+	// No response error checking here as there may be a mixture of successful
+	// and unsuccessful responses.
+
+	if len(results) != len(responses) {
+		return nil, errors.Errorf("Expected %d result objects into which to inject responses, but got %d", len(responses), len(results))
```
Liamsi:
```suggestion
		return nil, errors.Errorf("expected %d result objects into which to inject responses, but got %d", len(responses), len(results))
```

rpc/lib/client/http_client.go
```diff
@@ -202,6 +314,30 @@ func unmarshalResponseBytes(cdc *amino.Codec, responseBytes []byte, result inter
...
+	}
+
+	// Unmarshal the Result fields into the final result objects
+	for i := 0; i < len(responses); i++ {
+		err = cdc.UnmarshalJSON(responses[i].Result, results[i])
+		if err != nil {
+			return nil, errors.Errorf("Error unmarshalling rpc response result: %v", err)
```
Liamsi:
```suggestion
			return nil, errors.Errorf("error unmarshalling rpc response result: %v", err)
```

rpc/client/examples_test.go
thanethomson: I've refactored the examples (which required a bit of an extension to the `rpctest` library) to make them testable with a real Tendermint node running the `kvstore` app.

rpc/lib/client/http_client.go
thanethomson: Missed all of these somehow, sorry about that - I think I've got all of them now though!

rpc/client/examples_test.go
```diff
@@ -42,48 +48,83 @@ func ExampleHTTP_simple() {
...
+	rpctest.StopTendermint(node)
+
+	fmt.Println("Sent tx     :", string(tx))
+	fmt.Println("Queried for :", string(qres.Response.Key))
+	fmt.Println("Got value   :", string(qres.Response.Value))
+
+	// Output:
+	// Sent tx     : name=satoshi
+	// Queried for : name
+	// Got value   : satoshi
```
Liamsi: 👍 🎭

rpc/lib/client/http_client.go
Liamsi: No worries and thanks!

rpc/client/examples_test.go
Liamsi: Looks great.

rpc/client/examples_test.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	if !bytes.Equal(qres.Response.Key, k) {
+		panic("returned key does not match queried key")
+	}
+	if !bytes.Equal(qres.Response.Value, v) {
+		panic("returned value does not match sent value")
+	}
+
+	// Shut down Tendermint
+	rpctest.StopTendermint(node)
```
Liamsi: This should probably go into a defer statement after `rpctest.StartTendermint` above? There are a bunch of panics in between and (theoretically) the node would still keep running in a case of panic (same in the other test).

rpc/client/examples_test.go
```diff
@@ -0,0 +1,130 @@
...
+	for _, result := range results {
+		qr, ok := result.(*ctypes.ResultABCIQuery)
+		if !ok {
+			panic("invalid result type from ABCIQuery request")
+		}
+		fmt.Println(string(qr.Response.Key), "=", string(qr.Response.Value))
+	}
+
+	// Shut down Tendermint
+	rpctest.StopTendermint(node)
```
Liamsi: use `defer` seems safer here.

rpc/client/examples_test.go
thanethomson: Definitely, you're right - fixed this now 👍

rpc/client/examples_test.go
thanethomson: 😁

rpc/client/examples_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"bytes"
+	"fmt"
+
+	"github.com/tendermint/tendermint/abci/example/kvstore"
+	"github.com/tendermint/tendermint/rpc/client"
+	ctypes "github.com/tendermint/tendermint/rpc/core/types"
+	rpctest "github.com/tendermint/tendermint/rpc/test"
+)
+
+func ExampleHTTP_simple() {
```
melekes: This is not according to Go naming. Why not `ExampleHTTPSimple`?

rpc/client/examples_test.go
```diff
@@ -0,0 +1,126 @@
...
+	fmt.Println("Queried for :", string(qres.Response.Key))
+	fmt.Println("Got value   :", string(qres.Response.Value))
+
+	// Output:
+	// Sent tx     : name=satoshi
+	// Queried for : name
+	// Got value   : satoshi
+}
+
+func ExampleHTTP_batching() {
```
melekes: same

rpc/client/httpclient.go
```diff
@@ -18,67 +18,170 @@ import (
...
+// an example.
+type BatchHTTP struct {
+	rpcBatch *rpcclient.JSONRPCRequestBatch
+	*baseRPCClient
+}
+
+// rpcClient is an internal interface to which our RPC clients (batch and
+// non-batch) must conform. Acts as an additional code-level sanity check to
+// make sure the implementations stay coherent.
+type rpcClient interface {
```
melekes: Why not use already available public interface?

rpc/client/rpc_test.go
```diff
@@ -441,3 +444,100 @@ func TestTxSearch(t *testing.T) {
...
+func TestClearingEmptyJSONRPCRequestBatch(t *testing.T) {
+	c := getHTTPClient()
+	batch := c.NewBatch()
+	require.Zero(t, batch.Clear(), "clearing an empty batch of JSON RPC requests should result in a 0 result")
+}
+
+func TestConcurrentJSONRPCBatching(t *testing.T) {
+	c := getHTTPClient()
+	doneg := &sync.WaitGroup{}
```
melekes:
```suggestion
	var wg sync.WaitGroup
```

rpc/lib/server/handlers.go
```diff
@@ -113,49 +113,60 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, cdc *amino.Codec, logger lo
...
-		ctx := &types.Context{JSONReq: &request, HTTPReq: r}
-		args := []reflect.Value{reflect.ValueOf(ctx)}
-		if len(request.Params) > 0 {
-			fnArgs, err := jsonParamsToArgs(rpcFunc, cdc, request.Params)
+			// next, try to unmarshal as a single request
+			var request types.RPCRequest
+			err = json.Unmarshal(b, &request)
 			if err != nil {
-				WriteRPCResponseHTTP(w, types.RPCInvalidParamsError(request.ID, errors.Wrap(err, "Error converting json params to arguments")))
+				WriteRPCResponseHTTP(w, types.RPCParseError(types.JSONRPCStringID(""), errors.Wrap(err, "Error unmarshalling request")))
```
melekes: should we return request.ID instead of `types.JSONRPCStringID("")` here?

rpc/lib/server/handlers_test.go
```diff
@@ -154,6 +154,80 @@ func TestRPCNotification(t *testing.T) {
...
 
+func TestRPCNotificationInBatch(t *testing.T) {
+	mux := testMux()
+	tests := []struct {
+		payload     string
+		expectCount int
+	}{
+		{
+			`[
+				{"jsonrpc": "2.0","id": ""},
```
melekes: do we even need to test notification the second time? Shouldn't it be enough to test that we don't respond to it?

rpc/lib/server/handlers.go
```diff
@@ -113,49 +113,60 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, cdc *amino.Codec, logger lo
...
+			if !ok || rpcFunc.ws {
+				responses = append(responses, types.RPCMethodNotFoundError(request.ID))
+				continue
+			}
+			ctx := &types.Context{JSONReq: &request, HTTPReq: r}
+			args := []reflect.Value{reflect.ValueOf(ctx)}
+			if len(request.Params) > 0 {
+				fnArgs, err := jsonParamsToArgs(rpcFunc, cdc, request.Params)
+				if err != nil {
+					responses = append(responses, types.RPCInvalidParamsError(request.ID, errors.Wrap(err, "Error converting json params to arguments")))
```
melekes: lowercase err

rpc/lib/client/http_client.go
```diff
@@ -189,19 +301,43 @@ func unmarshalResponseBytes(cdc *amino.Codec, responseBytes []byte, result inter
...
 	err = cdc.UnmarshalJSON(response.Result, result)
 	if err != nil {
-		return nil, errors.Errorf("Error unmarshalling rpc response result: %v", err)
+		return nil, errors.Errorf("error unmarshalling rpc response result: %v", err)
 	}
 	return result, nil
 }
 
+func unmarshalResponseBytesArray(cdc *amino.Codec, responseBytes []byte, results []interface{}) ([]interface{}, error) {
+	var err error
```
```go
var (
...
)
```

rpc/lib/client/http_client.go
```diff
@@ -136,6 +197,57 @@ func (c *JSONRPCClient) SetCodec(cdc *amino.Codec) {
...
+		b.clear()
+		b.mtx.Unlock()
+	}()
+	return b.client.sendBatch(b.requests)
+}
+
+// Call enqueues a request to call the given RPC method with the specified
+// parameters, in the same way that the `JSONRPCClient.Call` function would.
+func (b *JSONRPCRequestBatch) Call(method string, params map[string]interface{}, result interface{}) (interface{}, error) {
+	request, err := types.MapToRequest(b.client.cdc, types.JSONRPCStringID("jsonrpc-client"), method, params)
```
melekes: Should we make "jsonrpc-client" a field in RPCClient? and append some nonce for every method call?

rpc/lib/client/http_client.go
```diff
@@ -122,10 +149,44 @@ func (c *JSONRPCClient) Call(method string, params map[string]interface{}, resul
...
-	// 	log.Info(Fmt("RPC response: %v", string(responseBytes)))
 	return unmarshalResponseBytes(c.cdc, responseBytes, result)
 }
 
+// NewRequestBatch starts a batch of requests for this client.
+func (c *JSONRPCClient) NewRequestBatch() *JSONRPCRequestBatch {
+	return &JSONRPCRequestBatch{
+		requests: make([]*JSONRPCBufferedRequest, 0),
+		client:   c,
+		mtx:      &sync.Mutex{},
```
melekes: no need to initialize mutex

rpc/lib/client/http_client.go
```diff
@@ -83,13 +84,39 @@ func makeHTTPClient(remoteAddr string) (string, *http.Client) {
 
 //------------------------------------------------------------------------------------
 
+// JSONRPCBufferedRequest encapsulates a single buffered request, as well as its
+// anticipated response structure.
+type JSONRPCBufferedRequest struct {
```
melekes: why expose it IF there's no access to it NOR it's an argument to some public function?

rpc/client/rpc_test.go
```diff
@@ -441,3 +444,100 @@ func TestTxSearch(t *testing.T) {
...
+	for i := 0; i < 50; i++ {
+		doneg.Add(1)
+		go func(_i int) {
```
melekes: why pass `i`?

rpc/client/rpc_test.go
```diff
@@ -441,3 +444,100 @@ func TestTxSearch(t *testing.T) {
...
+			testBatchedJSONRPCCalls(t, c)
+			doneg.Done()
```
melekes: `defer wg.Done()` in the beginning of the function please

rpc/client/examples_test.go
Liamsi: See under "Example function names":
https://blog.golang.org/examples

melekes: ooo, okay 👍 I'll read

thanethomson: See [here](https://golang.org/pkg/testing/#hdr-Examples) - since I'm not documenting a specific function, using a `_suffix` makes more sense, since these are package-level examples that are conceptually differentiated, instead of being specific to a particular function/type/method.

rpc/client/httpclient.go
Liamsi: Yeah, that is a good point 👍  I think it is OK to have this private `rpcClient` interface but it should make use of the existing interfaces by embedding them here.

thanethomson: I can make `rpcClient` like this:

```go
type rpcClient interface {
	ABCIClient
	HistoryClient
	NetworkClient
	SignClient
	StatusClient
}
```

But I'm trying to avoid including the `EventsClient` interface in the `BatchHTTP` struct.

rpc/lib/server/handlers.go
thanethomson: Will the ID be in there if parsing of `request` fails? I suppose if it isn't an empty string will be returned anyways, so I'll put `request.ID` in there 👍

rpc/lib/server/handlers_test.go
thanethomson: I was just testing for different permutations of notifications in the messages - I could remove the second test case, but I wanted to ensure that single and multiple notifications are also ignored.

rpc/lib/server/handlers.go
thanethomson: Done 👍

rpc/lib/server/handlers.go
melekes: > I suppose if it isn't an empty string will be returned anyways, so I'll put request.ID in there

that's why I'm asking and not just telling you to do so) wanted you to have a look at the code. thank you

rpc/client/rpc_test.go
thanethomson: Ah damn, I used `i` previously and somehow forgot to remove it when I refactored this code.

rpc/lib/server/handlers.go
thanethomson: This test fails now:

```
--- FAIL: TestRPCParams (0.00s)
        Error Trace:    handlers_test.go:84
        Error:      	Not equal:
                    	expected: rpctypes.JSONRPCStringID("")
                    	actual  : <nil>(<nil>)
        Test:       	TestRPCParams
        Messages:   	#2: expected ID not matched in RPCResponse
```

Seems like it results in a `nil` value instead of an empty string, so it's necessary 😁

rpc/lib/client/http_client.go
thanethomson: Not sure I understand - is this to make each batched request's response uniquely identifiable and able to be matched to a corresponding request?

rpc/lib/client/http_client.go
thanethomson: Sure - I generally tend to err on the side of making more things publicly accessible than private, just so other devs using the API have options, but in the case of `JSONRPCBufferedRequest` specifically I suppose we could consider it a purely internal concern for now. I'll make it private 👍

rpc/lib/server/handlers_test.go
Liamsi: I would have kept the test case. It doesn't harm and what you say makes sense to me. It might be worth to add a comment for the reasoning behind it. Le's not revert (again) it though :-)

rpc/lib/client/http_client.go
melekes: Better identify it on the server? I don't remember what ID should be according to JSONRPC spec, but I think it should be unique per client, so we should probably use cmn.RandStr? to generate suffix or something

CHANGELOG_PENDING.md
```diff
@@ -17,7 +17,8 @@
 ### FEATURES:
 
 ### IMPROVEMENTS:
-- [p2p] [\#3463](https://github.com/tendermint/tendermint/pull/3463) Do not log "Can't add peer's address to addrbook" error for a private peer
+- [rpc] \#3534 Add support for batched requests/responses in JSON RPC
+- [p2p] \#3463 Do not log "Can't add peer's address to addrbook" error for a private peer
```
melekes: please return the URL. I doubt that our script can expand links to PRs (note the link refers to a PR, not an issue)

rpc/test/helpers.go
```diff
@@ -23,7 +23,18 @@ import (
 	rpcclient "github.com/tendermint/tendermint/rpc/lib/client"
 )
 
+// Options helps with specifying some parameters for our RPC testing for greater
+// control.
+type Options struct {
```
melekes: this is a little bit different from what is common in the rest of the codebase 

see https://github.com/tendermint/tendermint/blob/a453628c4e43c67c747d5078704479aa73716182/libs/log/filter.go for example

better to be consistent unless there are strong points for Options struct

xla: A good resource to get familiar with options: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

And usage in our codebase for inspiration: https://github.com/tendermint/tendermint/blob/master/p2p/transport.go#L94-L117

CHANGELOG_PENDING.md
Liamsi: It can because github redirects to prs; e.g., try this `https://github.com/tendermint/tendermint/issue/3534`

rpc/lib/client/http_client.go
thanethomson: Oh so you mean that the Tendermint server can identify different clients? So yes, from the JSON RPC spec:

> The Server MUST reply with the same value in the Response object if included. This member is used to correlate the context between the two objects.

So I don't see any context being tracked on the Tendermint side at the moment for RPC clients (apart from logging with the ID), but should we need that in future it'd be critical to identify clients from each other.

Busy adding this functionality now 👍

CHANGELOG_PENDING.md
thanethomson: Ah, fixing it now 😁

rpc/test/helpers.go
thanethomson: I originally started with the standard approach of using options to influence the primary object in question (in this case, `*nm.Node`), but the trouble is that that approach assumes all you're only influencing state within that one struct.

For instance, if instead of making options of type `func(*Options)`, I make them of type `func(*nm.Node)` (the way @xla is suggesting), how do I influence state outside of `nm.Node` in `rpctest.GetConfig()` or `rpctest.StartTendermint()` (without introducing global variables and potentially leaking this configuration into other tests)?

There's a bunch of stuff going on inside the `rpc/test/helpers.go` package that isn't encapsulated in a higher-level structure that'd be more amenable to the traditional options approach, I think. Unless you want to refactor to create a higher-level object, something like `RPCTestNode` like so:

```go
type RPCTestNode struct {
    *nm.Node
    suppressStdout bool // This makes the examples work
    recreateConfig bool // This also makes the examples work
}

type Option func(*RPCTestNode)
```

Same goes for `log.filter`: the option function would make sense to be of type `func(*filter)` because all the necessary state to be influenced is contained within `filter`, but that won't allow you to influence any state outside of `filter`.

I thought this alternative approach would work better because it doesn't change the return types: it just adds variadic options to the functions. It also requires less refactoring.

But I can refactor this if consistency is more important?

CHANGELOG_PENDING.md
thanethomson: Thought I'd just fix it anyways - as long as it doesn't cause the scripts to try to double-link it somehow (i.e. insert a link to a non-existent issue within the PR link). Haven't taken a look at how those scripts work yet.

thanethomson: Weren't we supposed to only include links here to PRs though? I thought that's what we decided from one of the discussions earlier in the year? Unless I got them confused.

Liamsi: > Haven't taken a look at how those scripts work yet.

https://github.com/tendermint/tendermint/blob/75ffa2bf1c7d5805460d941a75112e6a0a38c039/scripts/linkify_changelog.py#L4-L13

You can link bot, PRs, and issues (due to github redirecting).

> Weren't we supposed to only include links here to PRs though?

@ebuchman often links PRs. Everyone else links only issues. We haven't really decided a strict policy here. I think whatever makes more sense in the particular context is fine 👍

rpc/test/helpers.go
xla: Thanks for all the context, I understand your reasoning. The refactoring you propose would be desirable, but would defo. conflate the scope of this PR. We should capture it in a follow-up issue and focus on the objective of the PR itself for now. Would that make sense?

Liamsi: IMHO sounds like this is worth consider doing in a separate PR (?) @melekes @xla

xla: @Liamsi see my comment, prior to yours.

Liamsi: heh, your comment didn't show while I posted mine. Agree as you can see 👍

rpc/lib/client/http_client.go
```diff
@@ -181,27 +295,84 @@ func (c *URIClient) SetCodec(cdc *amino.Codec) {
...
+
+	if len(results) != len(responses) {
+		return nil, errors.Errorf("expected %d result objects into which to inject responses, but got %d", len(responses), len(results))
+	}
+
+	// From the JSON-RPC 2.0 spec:
+	//  id: It MUST be the same as the value of the id member in the Request Object.
+	if len(expectedID) > 0 {
+		for i, response := range responses {
+			if response.ID == nil {
```
melekes: can we extract this code into a function and use it in both unmarshalResponseBytes and unmarshalResponseBytesArray?

thanethomson: Done 👍

rpc/test/helpers.go
thanethomson: Added #3565 to address this 👍

rpc/lib/client/http_client.go
thanethomson: Done adding this now 👍

rpc/client/httpclient.go
```diff
@@ -18,67 +18,154 @@ import (
...
+	StatusClient
+}
+
+// baseRPCClient implements the basic RPC method logic without the actual
+// underlying RPC call functionality, which is provided by `caller`.
+type baseRPCClient struct {
+	caller rpcclient.JSONRPCCaller
+}
+
+// HTTP implements rpcClient
```
melekes:
```suggestion
```

comment does not add any value

rpc/client/httpclient.go
```diff
@@ -18,67 +18,154 @@ import (
...
+var _ rpcClient = (*HTTP)(nil)
+
+// BatchHTTP implements rpcClient
```
melekes:
```suggestion
```

rpc/client/httpclient.go
```diff
@@ -18,67 +18,154 @@ import (
...
+var _ rpcClient = (*BatchHTTP)(nil)
+
+// baseRPCClient implements rpcClient
```
melekes:
```suggestion
```

rpc/client/httpclient.go
```diff
@@ -18,67 +18,154 @@ import (
...
+var _ rpcClient = (*baseRPCClient)(nil)
+
+//-----------------------------------------------------------------------------
+// HTTP
+
+// NewHTTP takes a remote endpoint in the form tcp://<host>:<port> and the
```
melekes: protocol can also be http, https, ws, wss

	protoHTTP  = "http"
	protoHTTPS = "https"
	protoWSS   = "wss"
	protoWS    = "ws"
	protoTCP   = "tcp"

rpc/lib/client/http_client.go
```diff
@@ -83,35 +85,63 @@ func makeHTTPClient(remoteAddr string) (string, *http.Client) {
...
+type jsonRPCBufferedRequest struct {
+	request types.RPCRequest
+	result  interface{} // The result will be deserialized into this object.
+}
+
+// JSONRPCRequestBatch allows us to buffer multiple request/response structures
+// into a single batch request. Note that this batch acts like a FIFO queue, and
+// is thread-safe.
+type JSONRPCRequestBatch struct {
+	requests []*jsonRPCBufferedRequest
```
melekes: the convention we're using in Tendermint - fields protected by the mutex should be declared after it. E.g.

```go
mtx      sync.Mutex
A typeA
B typeB

V typeC // fields, which are not protected by mutex, get separated by a new line
```

rpc/lib/client/http_client.go
```diff
@@ -122,8 +152,41 @@ func (c *JSONRPCClient) Call(method string, params map[string]interface{}, resul
...
+		requests: make([]*jsonRPCBufferedRequest, 0),
+		client:   c,
+	}
+}
+
+func (c *JSONRPCClient) sendBatch(requests []*jsonRPCBufferedRequest) ([]interface{}, error) {
+	reqCount := len(requests)
+	reqs := make([]types.RPCRequest, reqCount)
+	results := make([]interface{}, reqCount)
+	for i := 0; i < reqCount; i++ {
```
melekes: why not use `for range` which is generally safer?

rpc/lib/client/http_client.go
```diff
@@ -181,27 +295,83 @@ func (c *URIClient) SetCodec(cdc *amino.Codec) {
...
+	// From the JSON-RPC 2.0 spec:
+	//  id: It MUST be the same as the value of the id member in the Request Object.
+	for i, response := range responses {
+		if err := validateResponseID(&response, expectedID); err != nil {
+			return nil, errors.Errorf("failed to validate response ID in response %d: %v", i, err)
+		}
+	}
+
+	// Unmarshal the Result fields into the final result objects
+	for i := 0; i < len(responses); i++ {
```
melekes: same question. why not `range`?

also can't we merge this and the above loop?

rpc/lib/client/http_client.go
```diff
@@ -181,27 +295,83 @@ func (c *URIClient) SetCodec(cdc *amino.Codec) {
...
+		if err != nil {
+			return nil, errors.Errorf("error unmarshalling rpc response result: %v", err)
+		}
+	}
+	return results, nil
+}
+
+func validateResponseID(res *types.RPCResponse, expectedID types.JSONRPCStringID) error {
+	// we only validate a response ID if the expected ID is non-empty
+	if len(expectedID) > 0 {
```
```suggestion
if len(expectedID) == 0 {
	return nil
}
```

rpc/lib/server/handlers.go
```diff
@@ -113,49 +113,60 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, cdc *amino.Codec, logger lo
 			return
 		}
 
-		var request types.RPCRequest
-		err = json.Unmarshal(b, &request)
+		// first try to unmarshal the incoming request as an array of RPC requests
+		var requests []types.RPCRequest
```
melekes:
```go
var (
)
```

rpc/lib/server/handlers.go
```diff
@@ -113,49 +113,60 @@ func makeJSONRPCHandler(funcMap map[string]*RPCFunc, cdc *amino.Codec, logger lo
...
 		}
 
-		returns := rpcFunc.f.Call(args)
-
-		logger.Info("HTTPJSONRPC", "method", request.Method, "args", args, "returns", returns)
-		result, err := unreflectResult(returns)
-		if err != nil {
-			WriteRPCResponseHTTP(w, types.RPCInternalError(request.ID, err))
-			return
+		for i := 0; i < len(requests); i++ {
```
melekes: why not `range`?

rpc/client/httpclient.go
thanethomson: Should I just add a note about this in the comment?

melekes: no, just change tcp to <protocol>

rpc/lib/client/http_client.go
thanethomson: For me it's just been a convention: by default I'd use the C-style loop, and when I want to take advantage of the more short-hand 2-tuple variables unpacked by `for range` then I go with that.

Also, over here specifically it's just as safe as the `for range` option, unless `requests` is changing in size, or `len(requests)` comes back with an incorrect value, or `reqs` and/or `results` have been incorrectly allocated by the `make` function - all of which seem impossible to me here (and if those things do occur, then we have much bigger problems 😁).

I can stick to `for range` if that's the style we'd generally prefer.

rpc/lib/client/http_client.go
thanethomson: As above.

I could've merged the loop with the one above it, but I was doing all the error checking first for ID validation, and if all responses pass that check then go through the effort of unmarshalling.

rpc/lib/client/http_client.go
melekes: Yeah, I think we want to use `for range` where we can because it provides additional guarantees like the code won't panic with `out of bound` exception

rpc/lib/client/http_client.go
```diff
@@ -164,12 +165,11 @@ func (c *JSONRPCClient) NewRequestBatch() *JSONRPCRequestBatch {
 }
 
 func (c *JSONRPCClient) sendBatch(requests []*jsonRPCBufferedRequest) ([]interface{}, error) {
-	reqCount := len(requests)
-	reqs := make([]types.RPCRequest, reqCount)
-	results := make([]interface{}, reqCount)
-	for i := 0; i < reqCount; i++ {
-		reqs[i] = requests[i].request
-		results[i] = requests[i].result
+	reqs := make([]types.RPCRequest, 0)
```
```suggestion
	reqs := make([]types.RPCRequest, 0, len(requests))
```

rpc/lib/client/http_client.go
```diff
@@ -164,12 +165,11 @@ func (c *JSONRPCClient) NewRequestBatch() *JSONRPCRequestBatch {
...
+	results := make([]interface{}, 0)
```
```suggestion
	results := make([]interface{}, 0, len(requests))
```

Issue #3400, opened 2019-03-10
blockchain: non-deterministic test fail


ebuchman:
```
=== RUN   TestBadBlockStopsPeer
...
I[2019-03-10|09:19:32.335] Executed block                               height=146 validTxs=10 invalidTxs=0
I[2019-03-10|09:19:32.335] Committed state                              height=146 txs=10 appHash=
I[2019-03-10|09:19:32.336] Executed block                               height=147 validTxs=10 invalidTxs=0
I[2019-03-10|09:19:32.336] Committed state                              height=147 txs=10 appHash=
I[2019-03-10|09:19:32.339] Starting P2P Switch                          switch=0 impl="P2P Switch"
panic: Panicked on a Sanity Check: BlockStore can only save contiguous blocks. Wanted 149, got 147

goroutine 1251 [running]:
github.com/tendermint/tendermint/libs/common.PanicSanity(0xecf420, 0xc00130ebf0)
	/go/src/github.com/tendermint/tendermint/libs/common/errors.go:224 +0xc6
github.com/tendermint/tendermint/blockchain.(*BlockStore).SaveBlock(0xc000524840, 0xc0012cd680, 0xc000bacd20, 0xc00017c680)
	/go/src/github.com/tendermint/tendermint/blockchain/store.go:151 +0x807
github.com/tendermint/tendermint/blockchain.(*BlockchainReactor).poolRoutine(0xc0013abba0)
	/go/src/github.com/tendermint/tendermint/blockchain/reactor.go:332 +0xfb0
created by github.com/tendermint/tendermint/blockchain.(*BlockchainReactor).OnStart
	/go/src/github.com/tendermint/tendermint/blockchain/reactor.go:117 +0x129
FAIL	github.com/tendermint/tendermint/blockchain	7.102s
Exited with code 1

```

Seen in https://circleci.com/gh/tendermint/tendermint/49647#tests/containers/2

[build_49647_step_105_container_2.txt.zip](https://github.com/tendermint/tendermint/files/2949953/build_49647_step_105_container_2.txt.zip)

melekes:
Just got this locally:

```
panic: Panicked on a Sanity Check: BlockStore can only save contiguous blocks. Wanted 149, got 1

goroutine 1395 [running]:
github.com/tendermint/tendermint/libs/common.PanicSanity(...)
        /home/vagrant/go/src/github.com/tendermint/tendermint/libs/common/errors.go:224
[...]

PR #3537, opened 2019-04-09
rpc: Fix response time grow over time

melekes:
/validators rpc response time grow with time.

Steps To Reproduce:

    start a new node from hight 1
    2.time curl 127.0.0.1:26657/validators, it takes 0.005s
    let the chain run for serveral days, and change no validators.in my scenario the height is : 1584840
    time curl 127.0.0.1:26857/validators, it takes 1.25s

The api consumes cpu resouces, in the long run, a small scale of DDOS can make the node
a zombie process.

Acturally the logic to do IncrementProposerPriority by height - valInfo.LastHeightChanged times will not replay the exactly ProposerPriority once there are two or more round.

Replaces https://github.com/tendermint/tendermint/pull/3438

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -20,7 +20,8 @@
 
 ### BUG FIXES:
 
-- [state] Persist validators every 100000 blocks even if no changes to the set
+- [state] [\#3438](https://github.com/tendermint/tendermint/pull/3438) 
+  Persist validators every 100000 blocks even if no changes to the set
   occurred (@guagualvcha). This
   1) Prevents possible DoS attack using `/validators` or `/status` RPC
   endpoints. Before response time was growing linearly with height if no
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -21,8 +21,11 @@
 ### BUG FIXES:
 
 - [state] Persist validators every 100000 blocks even if no changes to the set
-  occurred. This prevents possible DoS attack using /validators RPC endpoint.
-  Before /validators response time was growing linearly if no changes were made
-  to validator set. (@guagualvcha)
+  occurred (@guagualvcha). This
+  1) Prevents possible DoS attack using `/validators` or `/status` RPC
+  endpoints. Before response time was growing linearly if no changes were made
+  to validator set.
+  2) Fixes performance degradation in `ExecCommitBlock` where we call
+  `LoadValidators`.
 - [p2p] \#2716 Check if we're already connected to peer right before dialing it (@melekes)
 - [docs] \#3514 Fix block.Header.Time description (@melekes)
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -27,3 +27,7 @@ mempool load for external ABCI apps.
 ### BUG FIXES:
 
 - [mempool] \#3512 Fix panic from concurrent access to txsMap, a regression for external ABCI apps introduced in v0.31.1
+- [state] Persist validators every 100000 blocks even if no changes to the set
+  occurred. This prevents possible DoS attack using /validators RPC endpoint.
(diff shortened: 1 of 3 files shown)


state/store.go
```diff
@@ -182,25 +190,40 @@ func LoadValidators(db dbm.DB, height int64) (*types.ValidatorSet, error) {
...
-		valInfo2.ValidatorSet.IncrementProposerPriority(int(height - valInfo.LastHeightChanged)) // mutate
+		valInfo2.ValidatorSet.IncrementProposerPriority(int(height - lastStoreHeight)) // mutate
 		valInfo = valInfo2
 	}
 
 	return valInfo.ValidatorSet, nil
 }
 
+func lastStoreHeightFor(height, lastHeightChanged int64) int64 {
+	lastStoreHeight := height - height%valSetStoreInterval
```
melekes: rename to `intermediateHeight` (or `checkpointHeight`)?

state/store.go
```diff
@@ -9,6 +9,14 @@ import (
 	"github.com/tendermint/tendermint/types"
 )
 
+const (
+	// persist validators every valSetStoreInterval blocks to avoid
+	// LoadValidators taking too much time.
+	// https://github.com/tendermint/tendermint/pull/3438
+	// 100000 results in ~ 100ms to get 100 validators (see BenchmarkLoadValidators)
+	valSetStoreInterval = 100000
```
Liamsi: Maybe,`valSetStoreBlockInterval` or `storeValSetHeightInterval` (it makes clear that every soandso blocks/height we store)

Liamsi:
ref: https://github.com/tendermint/tendermint/issues/3400 (restarting the build)

melekes:
> So should we see better performance also during execution? I see the LoadValidators() is being also called in ApplyBlock() and ExecCommitBlock() via execBlockOnProxyApp(). Trying to figure out where these changes will have an impact (other than the RPC queries).

omg, good point 👍 we're calling LoadValidators in /status and /validators in RPC and ExecCommitBlock and VerifyEvidencein state => 1s possible delay per block?

CHANGELOG_PENDING.md
```diff
@@ -19,5 +19,14 @@
 ### IMPROVEMENTS:
 
 ### BUG FIXES:
+
+- [state] [\#3438](https://github.com/tendermint/tendermint/pull/3438) 
+  Persist validators every 100000 blocks even if no changes to the set
+  occurred (@guagualvcha). This
+  1) Prevents possible DoS attack using `/validators` or `/status` RPC
+  endpoints. Before response time was growing linearly with height if no
+  changes were made to validator set.
```
```suggestion
  changes were made to the validator set.
```

state/store.go
```diff
@@ -232,7 +256,7 @@ func saveValidatorsInfo(db dbm.DB, height, lastHeightChanged int64, valSet *type
 	valInfo := &ValidatorsInfo{
 		LastHeightChanged: lastHeightChanged,
 	}
-	if lastHeightChanged == height {
+	if height == lastHeightChanged || height%valSetCheckpointInterval == 0 {
```
melekes:
```go
func valSetChangedOrIsCheckpointHeight(height) bool {
	return height == lastHeightChanged || height%valSetCheckpointInterval == 0
}
```

Liamsi: Yeah, I also thought about introducing vars to make this more readable. BUt thought it is OK as it is. 

Either the function you proposed or sth along the lines of:
```go
 hasValSetChanged := height == lastHeightChanged 
 isCheckPointHeight := height%valSetCheckpointInterval == 0
 if hasValSetChanged || isCheckPointHeight {
    // ...
 }
```
I prefer the vars instead of the function but both are fine.

sunboshan:
So with this fix, tendermint 0.31.4, how to get validators info?

```
$ curl 127.0.0.1:28221/validators
curl: (52) Empty reply from server
```

The current test node is on height about 270k and the validators rpc returns empty reply from server.

PR #3539, opened 2019-04-09
[ADR] ADR-037: Peer Behaviour

brapse:
* [ ] Wrote tests

### DIFF
--- a/docs/architecture/adr-037-peer-behaviour.md
+++ b/docs/architecture/adr-037-peer-behaviour.md
@@ -6,8 +6,7 @@
 ## Context
 
 The responsibility for signaling and acting upon peer behaviour lacks a single 
-owning component and is heavily coupled with the network stack, (cf.
-[#2067](https://github.com/tendermint/tendermint/issues/2067)). Reactors
+owning component and is heavily coupled with the network stack[^1]. Reactors
 maintain a reference to the `p2p.Switch` which they use to call 
 `switch.StopPeerForError(...)` when a peer misbehaves and 
 `switch.MarkAsGood(...)` when a peer contributes in some meaningful way. 
@@ -76,7 +75,7 @@ func NewSwitchedPeerBehaviour(sw *Switch) *SwitchedPeerBehaviour {
 }
 ```
 
-Reactors, which are often difficult to unit test (Cf. [#3506](https://github.com/tendermint/tendermint/pull/3506)). could use an implementation which exposes the signals produced by the reactor in
+Reactors, which are often difficult to unit test[^2]. could use an implementation which exposes the signals produced by the reactor in
 manufactured scenarios:
 
 ```go
@@ -142,5 +141,5 @@ Proposed
 
 ## References
 
-* [#2067](https://github.com/tendermint/tendermint/issues/2067): P2P Refactor
-* [#3506](https://github.com/tendermint/tendermint/pull/3506): ADR 036: Blockchain Reactor Refactor
+* [^1] Issue [#2067](https://github.com/tendermint/tendermint/issues/2067): P2P Refactor
+* [^2] PR: [#3506](https://github.com/tendermint/tendermint/pull/3506): ADR 036: Blockchain Reactor Refactor
--- a/docs/architecture/adr-037-peer-behaviour.md
+++ b/docs/architecture/adr-037-peer-behaviour.md
@@ -6,7 +6,7 @@
 ## Context
 
 The responsibility for signaling and acting upon peer behaviour lacks a single 
-owning component and is heavily coupled with the network stack[^1]. Reactors
+owning component and is heavily coupled with the network stack[<sup>1</sup>](#references). Reactors
 maintain a reference to the `p2p.Switch` which they use to call 
 `switch.StopPeerForError(...)` when a peer misbehaves and 
 `switch.MarkAsGood(...)` when a peer contributes in some meaningful way. 
@@ -75,7 +75,7 @@ func NewSwitchedPeerBehaviour(sw *Switch) *SwitchedPeerBehaviour {
 }
 ```
 
-Reactors, which are often difficult to unit test[^2]. could use an implementation which exposes the signals produced by the reactor in
+Reactors, which are often difficult to unit test[<sup>2</sup>](#references). could use an implementation which exposes the signals produced by the reactor in
 manufactured scenarios:
 
 ```go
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-037-peer-behaviour.md
```diff
@@ -0,0 +1,146 @@
...
+    * Reduce the coupling of reactors and the Switch and the network
+      stack
+    * The responsibility of managing peer behaviour can be migrated to
+      a single component instead of split between the switch and the
+      address book.
+
+### Negative
+
+    * The first iteration will simply wrap the Switch and introduce a
+      level of indirection.
```
Liamsi: You could could also argue that this is not negative as you propose to improve the status quo incrementally (which is a good thing IMHO).

docs/architecture/adr-037-peer-behaviour.md
```diff
@@ -0,0 +1,146 @@
...
+`switch.StopPeerForError(...)` when a peer misbehaves and 
+`switch.MarkAsGood(...)` when a peer contributes in some meaningful way. 
+While the switch handles `StopPeerForError` internally, the `MarkAsGood` 
+method delegates to another component, `p2p.AddrBook`. This scheme of delegation 
+across Switch obscures the responsibility for handling peer behaviour
+and ties up the reactors in a larger dependency graph when testing.
+
+## Decision
+
+Introduce a PeerBehaviour interface and concrete implementations which
```
Liamsi: nit:
```suggestion
Introduce a `PeerBehaviour` interface and concrete implementations which
```

docs/architecture/adr-037-peer-behaviour.md
```diff
@@ -0,0 +1,146 @@
...
+provide methods for reactors to signal peer behaviour without direct
+coupling `p2p.Switch`.  Introduce a ErrPeer to provide 
+concrete reasons for stopping peers.
+
+### Implementation Changes
+
+PeerBehaviour then becomes an interface for signaling peer errors as well
+as marking peers for `good`.
```
Liamsi:
```suggestion
as for marking peers as `good`.
```

docs/architecture/adr-037-peer-behaviour.md
```diff
@@ -0,0 +1,146 @@
...
+### Negative
+
+    * The first iteration will simply wrap the Switch and introduce a
+      level of indirection.
+
+### Neutral
+
+## References
+
+* [#2067](https://github.com/tendermint/tendermint/issues/2067): P2P Refactor
```
Liamsi: Nit: you can use the footnote style for linking (prob. there is an actual name for this) these to increase readability in the text above and to re-use the link. 
E.g.
```
Mention [#2067] in the text like this and use it again somewhere else, 
e.g. here: [#2067] and link to it once at the end (or somewhere):

[#2067]: https://github.com/tendermint/tendermint/issues/2067
```

docs/architecture/adr-037-peer-behaviour.md
xla: :+1:

docs/architecture/adr-037-peer-behaviour.md
```diff
@@ -0,0 +1,146 @@
...
+
+func (spb StorePeerBehaviour) Errored(peer Peer, reason ErrPeer) {
+    if _, ok := spb.pe[peer]; !ok {
+        spb.pe[peer] = []ErrPeer{reason}
+    } else {
+        spb.pe[peer] = append(spb.pe[peer], reason)
+    }
+}
+
+func (mpb *StorePeerBehaviour) GetPeerErrors() PeerErrors {
```
xla: Are the `Get` methods used to inspect the peer topology from test code?

docs/architecture/adr-037-peer-behaviour.md
brapse: Do you mean like this?
```suggestion
* [#2067]: https://github.com/tendermint/tendermint/issues/2067
```
or possibly use  [the markdown footnote](https://www.markdownguide.org/extended-syntax/#footnotes) to link the text to the footnote to the actual link?

docs/architecture/adr-037-peer-behaviour.md
brapse: Yes that's correct.

docs/architecture/adr-037-peer-behaviour.md
Liamsi: > or possibly use the markdown footnote to link the text to the footnote to the actual link? 

This one. I think it makes the document slightly easier to read and you can re-use the reference in the text.

brapse: Maybe I'm missing something, but Github markdown doesn't appear to support that particular `[^#]` features but we can accomplish something similar by linking to the Reference section with sub text. 

4d7

Liamsi: Haha, where did that awesome meme go! I really just meant to replace the 2 (or more) occurrences of links to issue with what is described in https://github.com/tendermint/tendermint/pull/3539#discussion_r273869696
(and for above reasons)
github supports this.
But NVM. The PR is really good as it is and it was just a tiny nit.

docs/architecture/adr-037-peer-behaviour.md
```diff
@@ -0,0 +1,145 @@
...
+as for marking peers as `good`.
+
+XXX: It might be better to pass p2p.ID instead of the whole peer but as
+a first draft maintain the underlying implementation as much as
+possible.
+
+```go
+type PeerBehaviour interface {
+    Errored(peer Peer, reason ErrPeer)
+    MarkPeerAsGood(peer Peer)
```
melekes: The first func now tells what happened (errored), the second one - what to do (mark). Maybe we should rename `MarkPeerAsGood` to `BecomeGood`?

Liamsi: Hmm, consistent would be sth like `BehavedWell`, `WasUseful`, `Contributed`? I think the actual name stems from the fact that the switch carries around an addrBook and the method there is `MarkGood`. Sound like we could improve the naming here @melekes 👍 Let's keep this in mind and wait for the implementation though.

xla: Good point! Let's ammend the ADR to not have it fall through the cracks during implementation. What about:

``` go
type PeerBehaviour interface {
    Behaved(id ID, reason GoodBehaviourPeer)
    Errored(id ID, reason ErrorBehaviourPeer)
}
```

@brapse & @milosevic Thoughts on this?

brapse: I really like the `Behaved`, `Errored` dichotomy!

The change from `peer p2p.Peer` -> `id p2p.ID` also further decouples the reactor from the network concern heavy peer interface. 👍 

The introduction of `GoodBehaviourPeer` also seems to complete the symmetry 👍 

Happy to amend:
image

PR #3544, opened 2019-04-11
add missing changelog entry

melekes:
https://github.com/tendermint/tendermint/pull/3474

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -17,6 +17,7 @@
 ### FEATURES:
 
 ### IMPROVEMENTS:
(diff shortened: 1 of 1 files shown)


CHANGELOG_PENDING.md
```diff
@@ -17,6 +17,7 @@
 ### FEATURES:
 
 ### IMPROVEMENTS:
+- [p2p] Do not log "Can't add peer's address to addrbook" error for a private peer
```
Liamsi: Shouldn't this link to the PR/issue?

melekes: There's no issue :)

Liamsi: But sure there is a PR? #3474

Issue #3532, opened 2019-04-08
p2p: Too many unsuccessful attempts


danil-lashin:
**Tendermint version**: 0.31.3 

**What happened**: p2p module of Tendermint trying to connect to peers from addrbook too many times.

For example, I found this record in addrbook.json in one of my nodes: 
```
{
	"addr": {
		"id": "7f880203aacfc405d926ea9a76163a9f284e8a5e",
		"ip": "37.153.1.214",
		"port": 26656
	},
	"src": {
		"id": "8c866c4cd346f55b38cafd007862b6279d2e8000",
		"ip": "136.243.33.21",
		"port": 48236
	},
	"attempts": 497,
	"last_attempt": "2019-04-08T14:51:24.699155733+03:00",
	"last_success": "0001-01-01T00:00:00Z",
	"bucket_type": 1,
	"buckets": [
		210,
		166,
		238,
		51
	]
}
```

**497 attempts** and no success one (`"last_success": "0001-01-01T00:00:00Z"`). It happens alot especially on start of new test network. Some nodes not wipe addrbook and some wipe node_key file.

**What you expected to happen**: Tendermint should remove peer from addrbook when max dial attempts ([16](https://github.com/tendermint/tendermint/blob/75ffa2bf1c7d5805460d941a75112e6a0a38c039/p2p/pex/pex_reactor.go#L49)) reached.

**Have you tried the latest version**: yes

danil-lashin:
Additional note: this peer is not persistent and not a seed.

xla:
Ref. #2067

melekes:
> Additional note: this peer is not persistent and not a seed.

you mean 7f880203aacfc405d926ea9a76163a9f284e8a5e or the node with this record in addrbook.json?

danil-lashin:
> you mean 7f880203aacfc405d926ea9a76163a9f284e8a5e or the node with this record in addrbook.json?

7f880203aacfc405d926ea9a76163a9f284e8a5e is not a persistent node in config of node with given addrbook.json

melekes:
What about the node with the given addrbook.json? Is it a seed?

danil-lashin:
Yes, it is.

melekes:
That's what I thought. Thanks for a quick response 🙇

melekes:
Merged https://github.com/tendermint/tendermint/pull/3573 to develop.

PR #3545, opened 2019-04-11
p2p: Bring back NodeInfo NetAddress form the dead

xla:
A prior change to address accidental DNS lookups introduced the SocketAddr on peer, which was then used to add it to the addressbook. Which in turn swallowed the self reported port of the peer, which is important on a reconnect. This change revives the NetAddress on NodeInfo which the Peer carries, but now returns an error to avoid nil dereferencing another issue observed in the past. Additionally we could potentially address #3532, yet the original problem statemenf of that issue stands.

As a drive-by optimisation `MarkAsGood` now takes only a `p2p.ID` which makes it interface a bit stricter and leaner.

### DIFF
--- a/p2p/node_info.go
+++ b/p2p/node_info.go
@@ -24,7 +24,12 @@ func MaxNodeInfoSize() int {
 // and determines if we're compatible.
 type NodeInfo interface {
 	ID() ID
+	nodeInfoAddress
 	nodeInfoTransport
+}
+
+type nodeInfoAddress interface {
+	NetAddress() (*NetAddress, error)
 }
 
 // nodeInfoTransport validates a nodeInfo and checks
@@ -209,20 +214,9 @@ OUTER_LOOP:
 // it includes the authenticated peer ID and the self-reported
 // ListenAddr. Note that the ListenAddr is not authenticated and
 // may not match that address actually dialed if its an outbound peer.
-func (info DefaultNodeInfo) NetAddress() *NetAddress {
+func (info DefaultNodeInfo) NetAddress() (*NetAddress, error) {
 	idAddr := IDAddressString(info.ID(), info.ListenAddr)
-	netAddr, err := NewNetAddressString(idAddr)
-	if err != nil {
-		switch err.(type) {
-		case ErrNetAddressLookup:
-			// XXX If the peer provided a host name and the lookup fails here
-			// we're out of luck.
-			// TODO: use a NetAddress in DefaultNodeInfo
-		default:
-			panic(err) // everything should be well formed by now
-		}
-	}
-	return netAddr
+	return NewNetAddressString(idAddr)
 }
 
 //-----------------------------------------------------------
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -55,7 +55,7 @@ type AddrBook interface {
 	PickAddress(biasTowardsNewAddrs int) *p2p.NetAddress
 
 	// Mark address
-	MarkGood(*p2p.NetAddress)
+	MarkGood(p2p.ID)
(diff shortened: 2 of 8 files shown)


xla:
Ref. #2067

p2p/pex/pex_reactor.go
```diff
@@ -170,12 +170,18 @@ func (r *PEXReactor) AddPeer(p Peer) {
 		}
 	} else {
 		// inbound peer is its own source
-		addr := p.SocketAddr()
+		addr, err := p.NodeInfo().NetAddress()
```
Liamsi: That's the line fixing the problem, right?

xla: Exactly. Funny enough the `connection refused` errors popped up in the pex tests, unfortunately in a context where we already expected the connection to fail.

p2p/node_info.go
```diff
@@ -209,20 +214,9 @@ OUTER_LOOP:
...
 	idAddr := IDAddressString(info.ID(), info.ListenAddr)
-	netAddr, err := NewNetAddressString(idAddr)
-	if err != nil {
-		switch err.(type) {
-		case ErrNetAddressLookup:
-			// XXX If the peer provided a host name and the lookup fails here
-			// we're out of luck.
-			// TODO: use a NetAddress in DefaultNodeInfo
-		default:
-			panic(err) // everything should be well formed by now
```
melekes: Do you think this panic was not necessary?

xla: By the looks of it, this panic was never hit and @ebuchmann hinted at that as well.

p2p/pex/addrbook.go
```diff
@@ -55,7 +55,7 @@ type AddrBook interface {
 	PickAddress(biasTowardsNewAddrs int) *p2p.NetAddress
 
 	// Mark address
-	MarkGood(*p2p.NetAddress)
+	MarkGood(p2p.ID)
 	MarkAttempt(*p2p.NetAddress)
```
melekes: Should we refactor MarkAttempt and MarkBad to accept p2p.ID instead of NetAddress? Otherwise interface is inconsistent.

melekes: Looks like we can't do that because `MarkBad` calls `RemoveAddress`, which requires `NetAddress`.

p2p/pex/pex_reactor.go
```diff
@@ -312,7 +318,10 @@ func (r *PEXReactor) ReceiveAddrs(addrs []*p2p.NetAddress, src Peer) error {
 	}
 	r.requestsSent.Delete(id)
 
-	srcAddr := src.SocketAddr()
+	srcAddr, err := src.NodeInfo().NetAddress()
```
ebuchman: Actually I'm not sure we need or want to do this here. We only care about the src IP, not port, so this fix isn't as relevant. Would be nice to formalize that in the type system somehow so it's clear.

This is fine as is, but it means we will do the DNS lookup for the source (if they have a name instead of just an IP) when we receive addresses from them, and if it fails, we'll ignore the addresses they sent us and we will disconnect that peer.

xla: Sounds like two distinct problems:

* only store src IP when receiving addresses
* use src.IP to reconnect and not DNS

Or do I miss-understand?

ebuchman: > use src.IP to reconnect and not DNS

Where did you get this from ? Aren't we already doing this ?

xla: From your comment:

> This is fine as is, but it means we will do the DNS lookup for the source (if they have a name instead of just an IP)

PR #3547, opened 2019-04-12
Fix a couple of typos

mdyring:
The secREKT is the most annoying one, as it turns up in log output a lot. ;-)

* [ ] Wrote tests

### DIFF
--- a/p2p/transport.go
+++ b/p2p/transport.go
@@ -364,7 +364,7 @@ func (mt *MultiplexTransport) upgrade(
 	if err != nil {
 		return nil, nil, ErrRejected{
 			conn:          c,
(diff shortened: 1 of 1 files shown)


p2p/transport.go
```diff
@@ -364,7 +364,7 @@ func (mt *MultiplexTransport) upgrade(
 	if err != nil {
 		return nil, nil, ErrRejected{
 			conn:          c,
-			err:           fmt.Errorf("secrect conn failed: %v", err),
+			err:           fmt.Errorf("secret conn failed: %v", err),
```
Liamsi: 😆

PR #3552, opened 2019-04-12
p2p: PeerBehaviour implementation (#3539) 

brapse:
Implementation of ADR-039 #3539 

* [x] Concurrency test for `StoredPeerBehaviour`
* [x] Wrote tests


p2p/peer_behaviour.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (spb *SwitchedPeerBehaviour) Errored(peer Peer, reason ErrorBehaviourPeer) {
+	spb.sw.StopPeerForError(peer, reason)
+}
+
+func (spb *SwitchedPeerBehaviour) Behaved(peer Peer, reason GoodBehaviourPeer) {
+	spb.sw.MarkPeerAsGood(peer)
+}
+
+func NewSwitchedPeerBehaviour(sw *Switch) *SwitchedPeerBehaviour {
```
melekes: should we return an interface here? then we'll be able to make the struct SwitchedPeerBehaviour private

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,83 @@
...
+
+// StorePeerBehaviour serves a mock concrete implementation of the
+// PeerBehaviour interface used in reactor tests to ensure reactors
+// produce the correct signals in manufactured scenarios.
+type StorePeerBehaviour struct {
+	eb ErrorBehaviours
+	gb GoodBehaviours
+}
+
+func NewStorePeerBehaviour() *StorePeerBehaviour {
```
melekes: same question

p2p/peer_behaviour.go
xla: You can make it private and still return the concrete type. There is merrit to avoid return of interfaces.

Liamsi: What are the benefits of avoiding to returning interfaces?

xla: If a caller needs the concrete type they can still acquire it without being forced to constrcut it manually in a potential inconsistent way. Additionally the implementation should not be couple to the interface which again might just be of interest to the caller.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,89 @@
...
+type GoodBehaviourPeer int
+
+const (
+	GoodBehaviourVote = iota
+	GoodBehaviourBlockPart
+)
+
+// PeerBehaviour provides an interface for reactors to signal the behaviour
+// of peers synchronously to other components.
+type PeerBehaviour interface {
```
Liamsi: Does it make sense to move the interface in a separate package?

p2p/peer_behaviour.go
Liamsi: Thanks, that makes sense! If the type is private though, the caller can't do anything with the concrete type. Unless the caller is in the same package (using internals of the implementation in tests etc).

p2p/peer_behaviour.go
xla: Only if we bundle all things peer together I'd say. But as long as peer is part of p2p it's not really helping much.

p2p/peer_behaviour.go
xla: Exactly right and inside the package boundary that might be okay.

melekes: > What are the benefits of not returning interfaces (but private types instead)?

Being able to call methods not specified in the interface (private - within the package, public - everywhere). If this is not desired, which I think is the case here, returning an interface is preferable.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,89 @@
...
+func NewswitchedPeerBehaviour(sw *Switch) PeerBehaviour {
+	return &switchedPeerBehaviour{
+		sw: sw,
+	}
+}
+
+type ErrorBehaviours map[Peer][]ErrorBehaviourPeer
+type GoodBehaviours map[Peer][]GoodBehaviourPeer
+
+type IStorePeerBehaviour interface {
```
xla: I assume that you want to expose this for consistent testing tooling in packages that depend on `PeerBehaviour`. Is that correct? If so I welcome this effort as we've seen with `Peer` we otherwise might end up with heaps of local `mock` implementations, which leads to unnecessary duplicated effort esp. with more complext types and harder to navigate test code for the uninitiated (often even the initiated).

Why is the name prefixed with an `I`?

Also embedding the `PeerBehaviour` interface in here makes it less flexible. I'd omit it and either have the caller construct a combined interface a la:

``` go
type myLocalTestBehaviour interface {
    p2p.PeerBehaviour
    p2p.StorePeerBehaviour
}
```
or expose a combined one from here (not as elegant).

At first I was on the fence about the name and was tempted to say we should signal more strongly that it is intended to be used in testing, but that might not be true.  :>

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,89 @@
...
+	PeerBehaviour
+	GetErrored() ErrorBehaviours
+	GetBehaved() GoodBehaviours
+}
+
+// StorePeerBehaviour serves a mock concrete implementation of the
+// PeerBehaviour interface used in reactor tests to ensure reactors
+// produce the correct signals in manufactured scenarios.
+type storePeerBehaviour struct {
```
xla: As with our design method on this will be called concurrently and we therefore need to evaluate if in it's current is thread-safe and consistent.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func NewswitchedPeerBehaviour(sw *Switch) PeerBehaviour {
+	return &switchedPeerBehaviour{
+		sw: sw,
+	}
+}
+
+type ErrorBehaviours map[Peer][]ErrorBehaviourPeer
+type GoodBehaviours map[Peer][]GoodBehaviourPeer
```
xla: As those are public they need proper doc strings.

I'd also advice to store `ID` instead of `Peer` to continue with our ambitions to not depend on an entire `Peer` for the behavioural part.

p2p/peer_behaviour_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"net"
+	"testing"
+)
+
+func TestStorePeerBehaviour(t *testing.T) {
+	peer := newMockPeer(net.IP{127, 0, 0, 1})
+	pb := NewStorePeerBehaviour()
+	pb.Errored(peer, ErrorBehaviourUnknown)
+
+	peerErrors := pb.GetErrored()
```
xla: For some extra safety we could check if the returned list has at least one element. Same for the call below.

p2p/peer_behaviour_test.go
```diff
@@ -0,0 +1,23 @@
+package p2p
+
+import (
+	"net"
+	"testing"
+)
+
+func TestStorePeerBehaviour(t *testing.T) {
```
xla: Would be good to test that there could be more than one error per peer and that the lists are empty with a fresh instance.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,89 @@
...
+func (spb *storePeerBehaviour) Errored(peer Peer, reason ErrorBehaviourPeer) {
+	if _, ok := spb.eb[peer]; !ok {
+		spb.eb[peer] = []ErrorBehaviourPeer{reason}
+	} else {
+		spb.eb[peer] = append(spb.eb[peer], reason)
+	}
+}
+
+func (mpb *storePeerBehaviour) GetErrored() ErrorBehaviours {
+	return mpb.eb
```
xla: We should think about what it means that we return the real slice here and what can happen in case the caller manipulates it.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,89 @@
...
+// produce the correct signals in manufactured scenarios.
+type storePeerBehaviour struct {
+	eb ErrorBehaviours
+	gb GoodBehaviours
+}
+
+func NewStorePeerBehaviour() IStorePeerBehaviour {
+	return &storePeerBehaviour{
+		eb: make(ErrorBehaviours),
+		gb: make(GoodBehaviours),
```
xla: The literal form is preferred: `ErrorBehaviours{}`.

p2p/peer_behaviour.go
xla: For your consideration: http://idiomaticgo.com/post/best-practice/accept-interfaces-return-structs/

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (spb *switchedPeerBehaviour) Errored(peer Peer, reason ErrorBehaviourPeer) {
+	spb.sw.StopPeerForError(peer, reason)
+}
+
+func (spb *switchedPeerBehaviour) Behaved(peer Peer, reason GoodBehaviourPeer) {
+	spb.sw.MarkPeerAsGood(peer)
+}
+
+func NewswitchedPeerBehaviour(sw *Switch) PeerBehaviour {
```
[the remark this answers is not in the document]
xla: Naming nit: `NewSwitchPeerBehaviour`.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// PeerBehaviour provides an interface for reactors to signal the behaviour
+// of peers synchronously to other components.
+type PeerBehaviour interface {
+	Behaved(peer Peer, reason GoodBehaviourPeer)
+	Errored(peer Peer, reason ErrorBehaviourPeer)
+}
+
+type switchedPeerBehaviour struct {
```
xla: Naming nit: `switchPeerBehaviour`.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,88 @@
...
+// storedPeerBehaviour serves a mock concrete implementation of the
+// PeerBehaviour interface used in reactor tests to ensure reactors
+// produce the correct signals in manufactured scenarios.
+type storedPeerBehaviour struct {
+	eb ErrorBehaviours
+	gb GoodBehaviours
+}
+
+type StoredPeerBehaviour interface {
+    GetErrored() ErrorBehaviours
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

p2p/peer_behaviour.go
brapse: What if we return a copy here?

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,95 @@
...
+	spb.mtx.Lock()
+	defer spb.mtx.Unlock()
+	if _, ok := spb.eb[peer]; !ok {
+		spb.eb[peer] = []ErrorPeerBehaviour{reason}
+	} else {
+		spb.eb[peer] = append(spb.eb[peer], reason)
+	}
+}
+
+func (spb storedPeerBehaviour) GetErrored() ErrorBehaviours {
```
golangcibot: copylocks: GetErrored passes lock by value: github.com/tendermint/tendermint/p2p.storedPeerBehaviour contains sync.Mutex (from `govet`)

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,95 @@
...
+	spb.mtx.Lock()
+	defer spb.mtx.Unlock()
+	if _, ok := spb.gb[peer]; !ok {
+		spb.gb[peer] = []GoodPeerBehaviour{reason}
+	} else {
+		spb.gb[peer] = append(spb.gb[peer], reason)
+	}
+}
+
+func (spb storedPeerBehaviour) GetBehaved() GoodBehaviours {
```
golangcibot: copylocks: GetBehaved passes lock by value: github.com/tendermint/tendermint/p2p.storedPeerBehaviour contains sync.Mutex (from `govet`)

p2p/peer_behaviour.go
brapse: I discussed this offline with @xla and pulled in @peterbourgon for another perspective to augment the article cited above. The most convincing argument I've heard is that interfaces are best used as consumer contracts, allowing the receiver to specify expectations, while return  types are producer specific concerns. Returning an interface couples the consumer expectation with the producing implementation. With this in mind I agree with @xla that we should return a private struct here.

p2p/peer_behaviour.go
brapse: Indeed. The target benefactor of this interface is the `reactor.Receive` method which is called within multiple go routines. I've placed a coarse lock around both write methods which should be sufficient for correctness of the `Behaved` and `Errored` methods. However, the `GetBehaved` and `GetErrored` methods return the underlying map, which exposes the struct to a separate write path in which the receive takes the map and mutates it willy-nilly. Given that this is specifically for testing, is it sufficient to specify these limitations in a doc string or should we deep copy the structure before returning it?

p2p/peer_behaviour.go
xla: Or as @peterbourgon pointed out a public struct with private fields.

p2p/peer_behaviour.go
xla: As discussed offline, changing the API to only return errors for a specific peer id will be cleaner and less error prone to implement.

p2p/peer_behaviour.go
Liamsi: Let's go with a public struct (with private fields). Returning a private struct doesn't seem uncommon but it still feels a bit off. For instance the documentation won't show in godoc. Sorry for the back and forth @brapse.

p2p/peer_behaviour_test.go
```diff
@@ -0,0 +1,41 @@
...
+		t.Errorf("Expected error to be ErrorPeerBehaviourUnknown")
+	}
+
+	peerGoods := pb.GetGoodBehaviours(peer.ID())
+	if len(peerGoods) != 0 {
+		t.Errorf("Expected the peer have zero error behaviour")
+	}
+
+	pb.Behaved(peer.ID(), GoodPeerBehaviourVote)
+	peerGoods = pb.GetGoodBehaviours(peer.ID())
```
golangcibot: ineffectual assignment to `peerGoods` (from `ineffassign`)

p2p/peer_behaviour.go
peterbourgon: @brapse and @xla asked why returning a private type was annoying. The primary reason is that it's difficult for callers to use, because the type can't escape the function boundary where it was received. That is, a function can't return a private type from another package, nor can it receive a private type from another package as a parameter.

melekes: What a great discussion 👍 Thanks for your thoughts guys

> Returning an interface couples the consumer expectation with the producing implementation.

that's not always a bad thing (e.g. https://github.com/go-kit/kit/blob/22a2d43a4ffeb2460fdf0d332b4d4f0638253fb2/log/logfmt_logger.go#L37)

@peterbourgon nice to see you here 🧙‍♂️

p2p/peer_behaviour_test.go
```diff
@@ -0,0 +1,183 @@
...
+			if aBehaviour == bBehaviour {
+				same[i] = aBehaviour
+			}
+		}
+	}
+
+	return len(same) == len(a)
+}
+
+// TestStoredPeerBehaviourConcurrency constrcuts a scenario in which
```
xla:
```suggestion
// TestStoredPeerBehaviourConcurrency constructs a scenario in which
```

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,137 @@
+package p2p
+
+import (
+	"errors"
+	"sync"
+)
+
+// Type of error behaviour a peer can perform.
```
xla:
```suggestion
// ErrorPeerBehaviour are types of observable erroneous peer behaviours.
```

This adheres to common go doc format.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,137 @@
...
+type ErrorPeerBehaviour int
+
+const (
+	ErrorPeerBehaviourUnknown = iota
+	ErrorPeerBehaviourBadMessage
+	ErrorPeerBehaviourMessageOutofOrder
+)
+
+// Type of good behaviour a peer can perform.
```
xla:
```suggestion
// GoodPeerBehaviour are types of observable positive peer behaviours.
```

Adhere to golang doc comment format.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,137 @@
...
+import (
+	"errors"
+	"sync"
+)
+
+// Type of error behaviour a peer can perform.
+type ErrorPeerBehaviour int
+
+const (
+	ErrorPeerBehaviourUnknown = iota
```
xla: Will we ever want to construct the unknown version of this behaviour? Can we make it impossible somehow to construct it?

One way to somewhat encode this is to do `iota + 1` for the first known one, so that the zero value for `int` is accidentally interpreted as a supported value.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,137 @@
...
+type PeerBehaviour interface {
+	Behaved(peerID ID, reason GoodPeerBehaviour) error
+	Errored(peerID ID, reason ErrorPeerBehaviour) error
+}
+
+type SwitchPeerBehaviour struct {
+	sw *Switch
+}
+
+// Reports the ErrorPeerBehaviour of peer identified by peerID to the Switch.
```
xla:
```suggestion
// Errored reports the ErrorPeerBehaviour of peer identified by ID to the Switch.
```

Adhere to golang doc comment format.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,137 @@
...
+	peer := spb.sw.Peers().Get(peerID)
+	if peer == nil {
+		return errors.New("Peer not found")
+	}
+
+	spb.sw.StopPeerForError(peer, reason)
+	return nil
+}
+
+// Reports the GoodPeerBehaviour of peer identified by peerID to the Switch.
```
xla:
```suggestion
// Behaved reports the GoodPeerBehaviour of peer identified by ID to the Switch.
```

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,137 @@
...
+	GoodPeerBehaviourBlockPart
+)
+
+// PeerBehaviour provides an interface for reactors to signal the behaviour
+// of peers synchronously to other components.
+type PeerBehaviour interface {
+	Behaved(peerID ID, reason GoodPeerBehaviour) error
+	Errored(peerID ID, reason ErrorPeerBehaviour) error
+}
+
```
xla:
```suggestion
// SwitchPeerBehaviour relays peer behaviour signals to a Switch, which is
// needed until Switch satisfies the PeerBehaviour interface.
```

All public types need a proper doc comment.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,137 @@
...
+	peer := spb.sw.Peers().Get(peerID)
+	if peer == nil {
+		return errors.New("Peer not found")
+	}
+
+	spb.sw.MarkPeerAsGood(peer)
+	return nil
+}
+
+// Return a new switchPeerBehaviour instance which wraps the Switch.
```
xla:
```suggestion
// NewswitchPeerBehaviour wraps a Switch and returns a PeerBehaviour.
```

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,137 @@
...
+func NewSwitchPeerBehaviour(sw *Switch) *SwitchPeerBehaviour {
```
xla: This constructor should be moved above right after the type def of `SwitchPeerBehaviour`.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,137 @@
...
+// produce the correct signals in manufactured scenarios.
+type StoredPeerBehaviour struct {
+	eb  map[ID][]ErrorPeerBehaviour
+	gb  map[ID][]GoodPeerBehaviour
+	mtx sync.RWMutex
+}
+
+// GettablePeerBehaviour provides an interface for accessing ErrorPeerBehaviours
+// and GoodPeerBehaviour recorded by an implementation of PeerBehaviour
+type GettablePeerBehaviour interface {
```
xla: By now it seems like this interface is superfluous. The comment already reveals a tight coupling to PeerBehaviour, but really the only place we going to need this is for `StoredPeerBehaviour`.  Why don't we dissolve this interface entirely and leave those two methods just as sturct methods on `StoredPeerBehaviour`, if the caller wants to call them they need to be aware they are handling that struct, esp. now that's public this will work fine. Should we end up with more concrete implementations exposing the same methods we can reassess the need for such an abstraction.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,137 @@
...
+	GetErrorBehaviours(peerID ID) []ErrorPeerBehaviour
+	GetGoodBehaviours(peerID ID) []GoodPeerBehaviour
+}
+
+// Creates a new storedPeerBehaviour instance.
```
xla:
```suggestion
// NewStoredPeerBehaviour returns a PeerBehaviour which records all observed
// behaviour in memory.
```

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,137 @@
...
+func NewStoredPeerBehaviour() *StoredPeerBehaviour {
+	return &StoredPeerBehaviour{
+		eb: map[ID][]ErrorPeerBehaviour{},
+		gb: map[ID][]GoodPeerBehaviour{},
+	}
+}
+
+// Stores the ErrorPeerBehaviour produced by the peer identified by peerID.
```
xla:
```suggestion
// Errored stores the ErrorPeerBehaviour produced by the peer identified by ID.
```

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,137 @@
...
+	spb.mtx.Lock()
+	defer spb.mtx.Unlock()
+	if _, ok := spb.eb[peerID]; !ok {
+		spb.eb[peerID] = []ErrorPeerBehaviour{reason}
+	} else {
+		spb.eb[peerID] = append(spb.eb[peerID], reason)
+	}
+}
+
+// Return all the ErrorBehaviours produced by peer identified by peerID.
```
xla:
```suggestion
// ErrorBehaviours returns all erorrs produced by peer identified by ID.
```

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,137 @@
...
+		result := make([]ErrorPeerBehaviour, len(items))
+		copy(result, items)
+
+		return result
+	} else {
+		return []ErrorPeerBehaviour{}
+	}
+}
+
+// Stores the GoodPeerBehaviour of the peer identified by peerID.
```
xla:
```suggestion
// Behaved stores the GoodPeerBehaviour of peer identified by ID.
```

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,137 @@
...
+	spb.mtx.Lock()
+	defer spb.mtx.Unlock()
+	if _, ok := spb.gb[peerID]; !ok {
+		spb.gb[peerID] = []GoodPeerBehaviour{reason}
+	} else {
+		spb.gb[peerID] = append(spb.gb[peerID], reason)
+	}
+}
+
+// Returns all the GoodPeerBehaviours produced by the peer identified by peerID.
```
xla:
```suggestion
// GetGoodPeerBehaviours returns all positive behaviours produced by the peer
// identified by peerID.
```

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,140 @@
...
+	GoodPeerBehaviourVote = iota + 100
+	GoodPeerBehaviourBlockPart
+)
+
+// PeerBehaviour provides an interface for reactors to signal the behaviour
+// of peers synchronously to other components.
+type PeerBehaviour interface {
+	Behaved(peerID ID, reason GoodPeerBehaviour) error
+	Errored(peerID ID, reason ErrorPeerBehaviour) error
+}
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

xla: oh shit

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,140 @@
...
+	"sync"
+)
+
+// ErrorPeerBehaviour are types of observable erroneous peer behaviours.
+type ErrorPeerBehaviour int
+
+const (
+	ErrorPeerBehaviourUnknown = iota
+	ErrorPeerBehaviourBadMessage
+	ErrorPeerBehaviourMessageOutofOrder
```
ebuchman: How do reactors specify reactor specific values for this, and should they need to? We can try to capture the common things, like timeouts and invalid messages, but presumably we may want to distinguish between those happening in the SecretConnection/MConn from those happening in specific reactors.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+// GoodPeerBehaviour are types of observable positive peer behaviours.
+type GoodPeerBehaviour int
+
+const (
+	GoodPeerBehaviourVote = iota + 100
+	GoodPeerBehaviourBlockPart
```
ebuchman: This kind of knowledge shouldn't be in the peer package - ie. the p2p package shouldn't depend on (or have knowledge of) consensus reactor behaviour.

Do we need per-channel peer behaviour? Then we can let these behaviours be defined where the channels are set up (eg. in the specific reactors)

(Recall each reactor can utilize multiple channels, where each channel is a stream multiplexed on the underlying MConn)

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+// PeerBehaviour provides an interface for reactors to signal the behaviour
+// of peers synchronously to other components.
+type PeerBehaviour interface {
+	Behaved(peerID ID, reason GoodPeerBehaviour) error
```
ebuchman: I'd go for something stronger than behaved. What we currently have is more like `BehavedForALongTime`, but maybe there's a better word for that.

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,140 @@
...
+func NewSwitchPeerBehaviour(sw *Switch) *SwitchPeerBehaviour {
+	return &SwitchPeerBehaviour{
+		sw: sw,
+	}
+}
+
+// storedPeerBehaviour serves a mock concrete implementation of the
+// PeerBehaviour interface used in reactor tests to ensure reactors
+// produce the correct signals in manufactured scenarios.
+type StoredPeerBehaviour struct {
```
ebuchman: awesome

p2p/peer_behaviour.go
```diff
@@ -0,0 +1,140 @@
...
+	eb  map[ID][]ErrorPeerBehaviour
+	gb  map[ID][]GoodPeerBehaviour
+	mtx sync.RWMutex
```
ebuchman: Our convention is to put mtxs immediately above the block of fields they protect

p2p/peer_behaviour.go
xla: Is it important to encode the temporal part? Sounds very narrow to me.

p2p/peer_behaviour.go
xla: Initially the reason for the introduction of the behaviour is to avoid disparate places in the codebase stopping a peer for a non-formalised error. We need to enumerate all possible errors in one place, ideally the place which decides if the error is severe enough to punish a peer. If we strive for a reactor centric definition of those errors we end up with the same kid of dynamic glue we currently try to reduce.

Do you have ideas how it could be expressed?

p2p/peer_behaviour.go
xla: We can't really express this with the go type system and it might likely lead to hard to debug code paths again, we wouldn't gain much, compared to the current implementation.

One approach is to extrat all possible behvaviours into it's own package.

p2p/peer_behaviour.go
ebuchman: `Behaved` is the normal behaviour. Many peers are behaving even if we're not calling this method. We only call this method once a peer has behaved for a sufficiently long period of time. It's actually stronger than behave ... currently it's a peer we've been receiving valuable unique information from. I'm not sure what a better name would be, `behave` is probably sufficient for now, but we can almost certainly do better :)

ebuchman: Also the interface is called `PeerBehaviour` so `Behaved` doesn't feel informative. How did it behave? Maybe this would be better with just `Good` and `Bad` methods?

brapse: I agree with of issues raised here around naming.  go
[guidelines](https://golang.org/doc/effective_go.html#interface-names)
suggest that interfaces should have `-er` post fix indicating what they
do. If we are reporting peer behaviour the interface should probably be
`PeerReporter`. In which case the only thing the PeerReporter needs
to do is ...Report.  This frees the caller from distinguishing
between good and bad behaviour. Here is one path we could take:

```go
type PeerBehaviour int
const (
	BadMessage = iota
	MessageOutOfOrder
	Vote
	BlockPart
)

type PeerReporter interface {
	Report(peerID ID, reason PeerBehaviour) error
}

type SwitchPeerReporter struct {
	sw *Switch
}

func (spb *SwitchPeerReporter) Report(peerID ID, behaviour PeerBehaviour) error {
	peer := spb.sw.Peers().Get(peerID)
	if peer == nil {
		return errors.New("Peer not found")
	}
    switch behaviour {
    case Vote:
		spb.sw.MarkPeerAsGood(peer, behaviour)))
    case BlockPart:
		spb.sw.MarkPeerAsGood(peer, behaviour))
    case BadMessage:
		spb.sw.StopPeerForError(peer, BlockPart)
    case MessageOutOfOrder:
		spb.sw.StopPeerForError(peer, BlockPart)
    default:
		return errors.New("Unknown Behaviour")
    }

	return nil
}
```

This approach still produces a central set of peer behaviours. I tend to agree with @xla that there is value in having central authority of stopping peers.

Liamsi: Maybe `PeerBehaviourReporter`? (it's a bit long but tells what it reports about too). And I wanted to throw in yet another verb/iface name: track / Tracker. As the thing keeps track of peer behaviours 🤔 

I like what @brapse and @xla proposed above. You should probably collapse the cases in the actual code.

brapse: From the Reactors perspective, I can imagine them taking on responsibility for reporting peers when they witness some behaviour. `Tracking` does have a nice ring to it but to me  it implies some process is being triggered which might not be the case. 
`PeerBehaviourReporter` also sounds nice, and would help distinguish between other reporters. (Metrics perhaps?).

p2p/peer_behaviour.go
ebuchman: I don't think we can avoid having some reactor-centric feature here without losing significant functionality or over-coupling our domains. 

At the least, we could standardize where in the respective reactor packages such errors are defined. For instance it could be part of channel registration: ie https://github.com/tendermint/tendermint/blob/develop/consensus/reactor.go#L125

p2p/peer_behaviour.go
ebuchman: > One approach is to extrat all possible behvaviours into it's own package.

This actually might not be a bad idea. A higher level package that deals with all these things. My main concern is just to avoid having reactor-specific ideas trickle into the `p2p` package.

p2p/peer_behaviour.go
ebuchman: Other possibilites: PeerMarker, PeerEvaluator, PeerRater, PeerAppraiser, etc.

p2p/peer_behaviour_test.go
```diff
@@ -0,0 +1,115 @@
+package p2p
```
Liamsi: As far as I can see the tests below only use the public API 👍 If you want you can rename this test package here to `p2p_test`.

p2p/peer_behaviour_test.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	pr.Report(peer.ID(), PeerBehaviourBadMessage)
+	behaviours = pr.GetBehaviours(peer.ID())
+	if len(behaviours) != 1 {
+		t.Errorf("Expected the peer have one reported behaviour")
+	}
+
+	if behaviours[0] != PeerBehaviourBadMessage {
+		t.Errorf("Expected PeerBehaviourBadMessage to have been reported")
```
Liamsi: Nit: not necessary to use `Errorf` (no formatting directive in the string).

p2p/peer_behaviour_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+	for i := 0; i < numConsumers; i++ {
+		done <- 1
+	}
+
+	receiveWg.Wait()
+
+	for _, items := range behaviourScript {
+		if !equalBehaviours(pr.GetBehaviours(items.PeerID), items.Behaviours) {
+			t.Errorf("Expected peer %s to have behaved \n", items.PeerID)
```
Liamsi: Nit: In case this fails: instead of only shoding the pperID here, wouldn't it be nicer if the test showed you what "script" was expected/wanted and what the actual/got behaviour observed was?

p2p/peer_behaviour.go
brapse: I think this makes sense. This could be moved to a `behaviour` package which effectively maps the reactor specific events to peer management.

I wonder if we couldn't implement a similar pattern as consensus message handling in [adr-30]("https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-030-consensus-refactor.md") 🤔 

Implementation behaviour/peer_behaviour.go:
```go
struct PeerBehaviour interface {}
...

type PeerBehaviourReporter interface {
	Report(peerID p2p.ID, behaviour PeerBehaviour) error
}
...
func (spbr *SwitchPeerBehaviourReporter) Report(peerID ID, behaviour PeerBehaviour) error {
	...
	switch behaviour.(type) {
	case PeerBehaviourVote
		spbr.sw.MarkPeerAsGood(peer)
		break
	...
	}
}
```


Example consensus/reactor.go:
```go
type PeerBehaviourVote struct {...}
...
peerReporter.Report(peerID, PeerBehaviourVote{...})
```

This would also leave room for reactors to provide richer information to the reporter in reactor specific ways. Thoughts?

brapse: I think this makes sense. This could be moved to a `behaviour` package which effectively maps the reactor specific events to peer management.

I wonder if we couldn't implement a similar pattern as the consensus message handling in [adr-30](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-030-consensus-refactor.md) 🤔 

Implementation behaviour/peer_behaviour.go:
```go
struct PeerBehaviour interface {}
...

type PeerBehaviourReporter interface {
	Report(peerID p2p.ID, behaviour PeerBehaviour) error
}
...
func (spbr *SwitchPeerBehaviourReporter) Report(peerID ID, behaviour PeerBehaviour) error {
	...
	switch behaviour.(type) {
	case PeerBehaviourVote
		spbr.sw.MarkPeerAsGood(peer)
		break
	...
	}
}
```


Example consensus/reactor.go:
```go
type PeerBehaviourVote struct {...}
...
peerReporter.Report(peerID, PeerBehaviourVote{...})
```
This might also resolve the [ #3552 (comment)](https://github.com/tendermint/tendermint/pull/3552#discussion_r280940857) and allow  reactor specific behaviour. Thoughts?

p2p/peer_behaviour.go
brapse: What if we move this to a higher level package as mentioned in
[#3552 comment](https://github.com/tendermint/tendermint/pull/3552#discussion_r280941065)? We could then follow a similar pattern as specified in [adr-30](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-030-consensus-refactor.md) to provide reactor-centric features:

Implementation: `behaviour/peer_behaviour.go:`
```go
struct PeerBehaviour interface {}

type PeerBehaviourVote struct {...}
...

type PeerBehaviourReporter interface {
	Report(peerID p2p.ID, behaviour PeerBehaviour) error
}
...
func (spbr *PeerBehaviourReporterXXX)Report(peerID ID, behaviour PeerBehaviour) error {
	...
	switch behaviour.(type) {
	case PeerBehaviourVote
        ..
		break
	...
	}
}
```

Example `consensus/reactor.go`:
```go
...
peerReporter.Report(peerID, PeerBehaviourVote{...})
```

Thoughts?

PR #3554, opened 2019-04-13
state: Rollback valset height

bradyjoestar:
#3555 

* [Yes] Updated all code comments where relevant
* [Yes] Wrote tests


melekes:
Can't we just set `valSetCheckpointInterval` to 1000 for this purpose?

Issue #2723, opened 2018-10-29
consensus: non-deterministic test failure


ebuchman:
Found in https://circleci.com/gh/tendermint/tendermint/31318#tests/containers/3

```
--- FAIL: TestWaitingTimeoutOnNilPolka (0.05s)
panic: Timeout expired while waiting for NewTimeout event [recovered]
	panic: Timeout expired while waiting for NewTimeout event

goroutine 38820 [running]:
testing.tRunner.func1(0xc4278aa3c0)
	/usr/local/go/src/testing/testing.go:742 +0x567
panic(0xefb1a0, 0xc4225429d0)
	/usr/local/go/src/runtime/panic.go:502 +0x24a
github.com/tendermint/tendermint/consensus.ensureNewEvent(0xc4227b3440, 0x1, 0x0, 0x1c9c380, 0x104fcd1, 0x32)
	/go/src/github.com/tendermint/tendermint/consensus/common_test.go:358 +0x69f
github.com/tendermint/tendermint/consensus.ensureNewTimeout(0xc4227b3440, 0x1, 0x0, 0x989680)
	/go/src/github.com/tendermint/tendermint/consensus/common_test.go:414 +0x70
github.com/tendermint/tendermint/consensus.TestWaitingTimeoutOnNilPolka(0xc4278aa3c0)
	/go/src/github.com/tendermint/tendermint/consensus/state_test.go:986 +0x370
testing.tRunner(0xc4278aa3c0, 0x1185580)
	/usr/local/go/src/testing/testing.go:777 +0x16e
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:824 +0x565
FAIL	github.com/tendermint/tendermint/consensus	36.265s
```

ebuchman:
This thing happens so often now

Liamsi:
ref: https://github.com/tendermint/tendermint/pull/3560

melekes:
Should be fixed in https://github.com/tendermint/tendermint/pull/3582.

PR #3560, opened 2019-04-14
state: Use last height changed if validator set is empty

melekes:
What happened:

New code v0.31.4 was supposed to fall back to last height changed when/if it
fails to find validators at checkpoint height (to make release
non-breaking).

But because we did not check if validator set is empty (only checking if we have
 an info is not enough, we save info for every block), the fall back
logic was never executed => resulting in LoadValidators returning an
empty validator set for cases where `lastStoredHeight` is checkpoint
height (i.e. almost all heights if the application does not change
validator set often).

How it was found:

one of our users - @sunboshan reported a bug here
https://github.com/tendermint/tendermint/pull/3537#issuecomment-482711833

* [x] Wrote tests

### DIFF
--- a/state/store_test.go
+++ b/state/store_test.go
@@ -6,34 +6,50 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	cfg "github.com/tendermint/tendermint/config"
 	dbm "github.com/tendermint/tendermint/libs/db"
 	"github.com/tendermint/tendermint/types"
 )
 
-func TestSaveValidatorsInfo(t *testing.T) {
-	// test we persist validators every valSetCheckpointInterval blocks
+func TestStoreLoadValidators(t *testing.T) {
 	stateDB := dbm.NewMemDB()
 	val, _ := types.RandValidator(true, 10)
 	vals := types.NewValidatorSet([]*types.Validator{val})
 
-	// TODO(melekes): remove in 0.33 release
-	// https://github.com/tendermint/tendermint/issues/3543
+	// 1) LoadValidators loads validators using a height where they were last changed
 	saveValidatorsInfo(stateDB, 1, 1, vals)
 	saveValidatorsInfo(stateDB, 2, 1, vals)
+	loadedVals, err := LoadValidators(stateDB, 2)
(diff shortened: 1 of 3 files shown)


Liamsi:
we are seeing https://github.com/tendermint/tendermint/issues/2723 again; restarting the build to be certain all other tests pass

state/store_test.go
```diff
@@ -6,34 +6,50 @@ import (
...
 	assert.NotPanics(t, func() {
-		_, err := LoadValidators(stateDB, 2)
+		saveValidatorsInfo(stateDB, valSetCheckpointInterval+1, 1, vals)
+		loadedVals, err := LoadValidators(stateDB, valSetCheckpointInterval+1)
 		if err != nil {
-			panic(err)
+			t.Fatal(err)
+		}
+		if loadedVals.Size() == 0 {
+			t.Fatal("Expected validators to be non-empty")
```
Liamsi: This code would never be reached, right? Because `panic("empty validator set")` would happen before? I like that these two lines make this more explicit! (keep them, I'm just asking).

melekes: If you remove ` || valInfo2.ValidatorSet == nil` check, it will reach this code and fail.

With ` || valInfo2.ValidatorSet == nil`, the code does not panic and returns non-empty validator set.

Liamsi: OK, just tried it locally: I did not add the `|| valInfo2.ValidatorSet == nil` check but added the changes in the test. The function will panic before the `loadedVals.Size() == 0` is reached because `LoadValidators` calls `IncrementProposerPriority` anyways. Which will panic with `empty validator set`:
https://github.com/tendermint/tendermint/blob/a453628c4e43c67c747d5078704479aa73716182/types/validator_set.go#L82-L85

melekes: You're right!

melekes: I think it's still fine to leave the check just in case.

Liamsi:
Now failing: 
```
=== RUN   TestSIGHUP
MustReadFile failed: open /tmp/sighup_test259331718: no such file or directory
FAIL	github.com/tendermint/tendermint/libs/autofile	0.052s
Exited with code 1
```

PR #3561, opened 2019-04-15
blockchain: Reorg reactor

ancazamfir:
This is the draft PR for a possible new implementation of the blockchain reactor. It is still work in progress and parts of the design might change. 
One piece missing are tests fo the reactor.go:poolRoutine(), currently still covered mainly in reactor_test.go
I also need to write design notes and add more comments.


blockchain_old/reactor_test.go
```diff
@@ -0,0 +1,416 @@
...
+			lastBlock := blockStore.LoadBlock(blockHeight - 1)
+
+			vote := makeVote(&lastBlock.Header, lastBlockMeta.BlockID, state.Validators, privVals[0]).CommitSig()
+			lastCommit = types.NewCommit(lastBlockMeta.BlockID, []*types.CommitSig{vote})
+		}
+
+		thisBlock := makeBlock(blockHeight, state, lastCommit)
+
+		thisParts := thisBlock.MakePartSet(types.BlockPartSizeBytes)
+		blockID := types.BlockID{thisBlock.Hash(), thisParts.Header()}
```
golangcibot: composites: `github.com/tendermint/tendermint/types.BlockID` composite literal uses unkeyed fields (from `govet`)

blockchain/reactor_fsm_test.go
```diff
@@ -0,0 +1,720 @@
...
+			peerId: peerID,
+			height: height,
+			length: len(msgBytes),
+		},
+	}
+
+	_ = sendMessageToFSMSync(fsm, msgData)
+}
+
+func sendStatusResponse2(fsm *bReactorFSM, peerID p2p.ID, height int64) {
```
golangcibot: `sendStatusResponse2` is unused (from `deadcode`)

blockchain/reactor_fsm_test.go
```diff
@@ -0,0 +1,720 @@
...
+		data: bReactorEventData{
+			peerId: peerID,
+			height: height,
+			length: len(msgBytes),
+		},
+	}
+	_ = fsm.handle(msgData)
+}
+
+func sendStateTimeout(fsm *bReactorFSM, name string) {
```
golangcibot: `sendStateTimeout` is unused (from `deadcode`)

blockchain/reactor_fsm_test.go
```diff
@@ -0,0 +1,720 @@
...
+
+	failStatusReq       bool
+	shouldSendStatusReq bool
+
+	failBlockReq      bool
+	blockReqIncreased bool
+	blocksAdded       []int64
+	peersNotInPool    []p2p.ID
+
+	expectedLastBlockReq *lastBlockRequestT
```
golangcibot: `expectedLastBlockReq` is unused (from `structcheck`)

blockchain/reactor_fsm_test.go
```diff
@@ -0,0 +1,720 @@
...
+	numStatusRequests = 0
+	numBlockRequests = 0
+	lastBlockRequest.peerID = ""
+	lastBlockRequest.height = 0
+	lastPeerError.peerID = ""
+	lastPeerError.err = nil
+}
+
+type fsmStepTestValues struct {
+	startingHeight int64
```
golangcibot: `startingHeight` is unused (from `structcheck`)

blockchain/reactor_fsm_test.go
```diff
@@ -0,0 +1,720 @@
...
+	maxTotalPendingRequestsTest = 600
+	maxNumPeersTest             = 1000
+	maxNumBlocksInChainTest     = 50000
+)
+
+type testFields struct {
+	name               string
+	startingHeight     int64
+	maxRequestsPerPeer int32
+	maxPendingRequests int32
```
golangcibot: `maxPendingRequests` is unused (from `structcheck`)

blockchain/pool_test.go
```diff
@@ -1,224 +1,572 @@
...
+	height int64
 }
 
-type testPeer struct {
-	id        p2p.ID
-	height    int64
-	inputChan chan inputData //make sure each peer's data is sequential
+type testPeerResult struct {
+	id         p2p.ID
+	height     int64
```
golangcibot: `height` is unused (from `structcheck`)

blockchain/pool_test.go
```diff
@@ -1,224 +1,572 @@
...
+	numPending int32
+	blocks     map[int64]*types.Block
```
golangcibot: `blocks` is unused (from `structcheck`)

blockchain_old/store.go
```diff
@@ -0,0 +1,247 @@
+package blockchain_old
```
golangcibot: 1-247 lines are duplicate of `blockchain/store.go:1-247` (from `dupl`)

blockchain/pool_test.go
```diff
@@ -1,224 +1,572 @@
...
 	}
-	assert.EqualValues(t, 10, pool.MaxPeerHeight())
+}
 
-	// remove not-existing peer
-	assert.NotPanics(t, func() { pool.RemovePeer(p2p.ID("Superman")) })
+func TestBlockPoolGetNextTwoBlocks(t *testing.T) {
+	testBcR := newTestBcR()
+
+	tests := []struct {
```
golangcibot: 427-470 lines are duplicate of `blockchain/pool_test.go:500-543` (from `dupl`)

blockchain/pool_test.go
```diff
@@ -1,224 +1,572 @@
...
+	}
+}
 
-	// remove all peers
-	for peerID := range peers {
-		pool.RemovePeer(peerID)
+func TestBlockPoolInvalidateFirstTwoBlocks(t *testing.T) {
+	testBcR := newTestBcR()
+
+	tests := []struct {
```
golangcibot: 500-543 lines are duplicate of `blockchain/pool_test.go:427-470` (from `dupl`)

blockchain_old/store_test.go
```diff
@@ -0,0 +1,421 @@
+package blockchain_old
```
golangcibot: 1-421 lines are duplicate of `blockchain/store_test.go:1-421` (from `dupl`)

blockchain/reactor_fsm_test.go
```diff
@@ -0,0 +1,704 @@
...
+type fsmStepTestValues struct {
+	startingHeight int64
+	currentState   string
+	event          bReactorEvent
+	data           bReactorEventData
+	errWanted      error
+	expectedState  string
+
+	shouldSendStatusReq bool
+	failBlockReq        bool
```
golangcibot: `failBlockReq` is unused (from `structcheck`)

blockchain/reactor_fsm_test.go
```diff
@@ -0,0 +1,1185 @@
...
+				// statusResponseEv from P2
+				makeStepStatusEv("waitForBlock", "waitForBlock", "P2", 3, nil),
+				// statusResponseEv from P3
+				makeStepStatusEv("waitForBlock", "waitForBlock", "P3", 3, nil),
+				makeStepPeerRemoveEv("waitForBlock", "waitForBlock", "P2", errSwitchRemovesPeer, []p2p.ID{"P2"}),
+			},
+		},
+	}
+
+	for _, tt := range tests {
```
golangcibot: 602-633 lines are duplicate of `blockchain/reactor_fsm_test.go:679-710` (from `dupl`)

blockchain/reactor_fsm_test.go
```diff
@@ -0,0 +1,1185 @@
...
+				makeStepStartFSMEv(),
+				// statusResponseEv from P1
+				makeStepStatusEv("waitForPeer", "waitForBlock", "P1", 200, nil),
+				// statusResponseEv from P1
+				makeStepStatusEv("waitForBlock", "waitForPeer", "P1", 3, errPeerTooShort),
+			},
+		},
+	}
+
+	for _, tt := range tests {
```
golangcibot: 679-710 lines are duplicate of `blockchain/reactor_fsm_test.go:735-766` (from `dupl`)

blockchain/reactor_fsm_test.go
```diff
@@ -0,0 +1,1185 @@
...
+				makeStepStartFSMEv(),
+				// statusResponseEv from P1
+				makeStepStatusEv("waitForPeer", "waitForBlock", "P1", 20000000, nil),
+				makeStepMakeRequestsEvErrored("waitForBlock", "waitForBlock",
+					maxNumPendingRequests, nil, []p2p.ID{"P1"}),
+			},
+		},
+	}
+
+	for _, tt := range tests {
```
golangcibot: 735-766 lines are duplicate of `blockchain/reactor_fsm_test.go:602-633` (from `dupl`)

blockchain/reactor_fsm_test.go
```diff
@@ -0,0 +1,1174 @@
...
+				// add another peer
+				makeStepStatusEv("waitForBlock", "waitForBlock", "P2", 3, nil),
+
+				makeStepBlockRespEvErrored("waitForBlock", "waitForBlock",
+					"P2", 1, []int64{}, errBadDataFromPeer, []p2p.ID{"P2"}),
+			},
+		},
+	}
+
+	for _, tt := range tests {
```
golangcibot: 529-567 lines are duplicate of `blockchain/reactor_fsm_test.go:612-650` (from `dupl`)

blockchain/reactor_fsm_test.go
```diff
@@ -0,0 +1,1174 @@
...
+	msgData := bReactorMessageData{
+		event: statusResponseEv,
+		data: bReactorEventData{
+			peerId: peerID,
+			height: height,
+			length: len(msgBytes),
+		},
+	}
+
+	_ = sendMessageToFSMSync(fsm, msgData)
```
golangcibot: undeclared name: `sendMessageToFSMSync` (from `typecheck`)

blockchain/peer.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+type bpPeer struct {
+	logger log.Logger
+	id     p2p.ID
+
+	height      int64                  // the peer reported height
+	numPending  int32                  // number of requests still waiting for block responses
+	blocks      map[int64]*types.Block // blocks received or expected to be received from this peer
+	timeout     *time.Timer
```
milosevic: Can we have less generic name for this?

blockchain/peer.go
```diff
@@ -0,0 +1,134 @@
...
+	peerWindowSize = 40 * peerSampleRate
+)
+
+type bpPeer struct {
+	logger log.Logger
+	id     p2p.ID
+
+	height      int64                  // the peer reported height
+	numPending  int32                  // number of requests still waiting for block responses
+	blocks      map[int64]*types.Block // blocks received or expected to be received from this peer
```
milosevic: The fact that we keep in this data structure downloaded blocks and pending requests is a bit confusing. Maybe it should be split into two data structures.

blockchain/peer.go
```diff
@@ -0,0 +1,134 @@
...
+	if peer.timeout == nil {
+		peer.timeout = time.AfterFunc(peerTimeout, peer.onTimeout)
+	} else {
+		peer.timeout.Reset(peerTimeout)
+	}
+}
+
+func (peer *bpPeer) incrPending() {
+	if peer.numPending == 0 {
+		peer.resetMonitor()
```
milosevic: Why we reset monitor in this case?

blockchain/peer.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (peer *bpPeer) decrPending(recvSize int) {
+	if peer.numPending == 0 {
+		panic("cannot decrement, peer does not have pending requests")
+	}
+
+	peer.numPending--
+	if peer.numPending == 0 {
+		peer.timeout.Stop()
```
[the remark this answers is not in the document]
milosevic: Probably not as we call peer.resetTimeout (which creates timer) when we increment `numPending`.

blockchain/peer.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+}
+
+func (peer *bpPeer) onTimeout() {
+	peer.errFunc(errNoPeerResponse, peer.id)
+}
+
+func (peer *bpPeer) isGood() error {
+	if peer.didTimeout {
+		return errNoPeerResponse
```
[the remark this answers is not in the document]
milosevic: The first time we detect an error we remove the peer (as this function is called only from `removeBadPeers`) so there will be no next time.

blockchain/peer.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (peer *bpPeer) resetTimeout() {
+	if peer.timeout == nil {
+		peer.timeout = time.AfterFunc(peerTimeout, peer.onTimeout)
+	} else {
+		peer.timeout.Reset(peerTimeout)
+	}
+}
+
+func (peer *bpPeer) incrPending() {
```
milosevic: As a more general comment, I don't like the fact that we are mixing-in timeout logic here. Timeout management is important protocol aspect and ideally should be part of Controller logic. But this is outside the scope of this PR.

blockchain/peer_test.go
```diff
@@ -0,0 +1,241 @@
...
+	assert.NotNil(t, peer.timeout)
+	// make sure timer is running and stop it
+	checkByStoppingPeerTimer(t, peer, true)
+
+	// reset with non nil expired timer
+	peer.resetTimeout()
+	assert.NotNil(t, peer.timeout)
+	// make sure timer is running and stop it
+	checkByStoppingPeerTimer(t, peer, true)
+	resetErrors()
```
milosevic: Is this needed? What error can be created at this point?

blockchain/peer_test.go
```diff
@@ -0,0 +1,241 @@
...
+	checkByStoppingPeerTimer(t, peer, false)
+
+	peerTestMtx.Lock()
+	// ... check an error has been sent, error is peerNonResponsive
+	assert.Equal(t, 1, numErrFuncCalls)
+	assert.Equal(t, lastErr, errNoPeerResponse)
+	peerTestMtx.Unlock()
+
+	// Restore the peerTimeout to its original value
+	peerTimeout = 15 * time.Second
```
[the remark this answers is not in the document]
milosevic: Do we really need to restore this value? Every test should set it initially.

blockchain/peer.go
```diff
@@ -0,0 +1,134 @@
...
+	if peer.numPending == 0 {
+		peer.resetMonitor()
+		peer.resetTimeout()
+	}
+	peer.numPending++
+}
+
+func (peer *bpPeer) decrPending(recvSize int) {
```
milosevic: We are mixing concerns here. It's confusing receiving int as a parameter as it does not correspond to decrement amount. `recvSize` is monitor related concern so we should probably have it somewhere else.

blockchain/peer.go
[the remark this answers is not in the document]
milosevic: Do we need to stop monitor here?

blockchain/peer_test.go
```diff
@@ -0,0 +1,241 @@
...
+	peer := &bpPeer{
+		id:      p2p.ID(cmn.RandStr(12)),
+		height:  10,
+		errFunc: errFunc,
+		logger:  log.TestingLogger(),
+	}
+
+	peerTimeout = time.Millisecond
+	peer.incrPending()
+	time.Sleep(2 * time.Millisecond)
```
milosevic: Could we have non deterministic test failure in this case?

blockchain/peer.go
```diff
@@ -0,0 +1,134 @@
...
+			// consider the peer timedout
+			peer.didTimeout = true
+			return errSlowPeer
+		}
+	}
+
+	return nil
+}
+
+func (peer *bpPeer) cleanup() {
```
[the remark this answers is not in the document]
milosevic: Should probably be removed.

blockchain/peer_test.go
```diff
@@ -0,0 +1,241 @@
...
+	for i := 0; i < 10; i++ {
+		peer.decrPending(9)
+		time.Sleep(100 * time.Millisecond)
+	}
+	// check peer is considered slow
+	assert.Equal(t, errSlowPeer, peer.isGood())
+
+}
+
+func TestPeerCleanup(t *testing.T) {
```
milosevic: Should probably be removed.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
-				peer.didTimeout = true
-			}
-		}
-		if peer.didTimeout {
-			pool.removePeer(peer.id)
-		}
+type blockPool struct {
+	logger log.Logger
+	peers  map[p2p.ID]*bpPeer
+	blocks map[int64]p2p.ID
```
milosevic: Can we have a comment explaining what this variable represent?

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
+
+	requests          map[int64]bool // list of blocks to be assigned peers for blockRequest
```
milosevic: Maybe it's better to call it failedRequests?

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
-		}
-		*/
-		r.Stop()
-		delete(pool.requesters, pool.height)
-		pool.height++
-	} else {
-		panic(fmt.Sprintf("Expected requester to pop, got nothing at height %v", pool.height))
-	}
+func (pool *blockPool) reachedMaxHeight() bool {
+	return pool.maxPeerHeight == 0 || pool.height >= pool.maxPeerHeight
```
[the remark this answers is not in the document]
milosevic: As we need blocks h and h+1 to be able to verify and execute block h, shouldn't the second condition be `pool.height >= pool.maxPeerHeight - 1`? I think that we were having condition like this in the previous version of the code.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
-	if requester == nil {
-		pool.Logger.Info("peer sent us a block we didn't expect", "peer", peerID, "curHeight", pool.height, "blockHeight", block.Height)
-		diff := pool.height - block.Height
-		if diff < 0 {
-			diff *= -1
-		}
-		if diff > maxDiffBetweenCurrentAndReceivedBlockHeight {
-			pool.sendError(errors.New("peer sent us a block we didn't expect with a height too far ahead/behind"), peerID)
+// Updates the pool's max height. If no peers are left maxPeerHeight is set to 0.
+func (pool *blockPool) updateMaxPeerHeight() {
```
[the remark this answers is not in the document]
milosevic: Making this logic peer based (upon peer update or remove) will make also logic of failed requests probably simpler as we only care about requests sent to this peer. We probably don't need to care about requests sent to other peers as we don't send a request to a peer if he hasn't reported bigger height.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
 
-	peer := pool.peers[peerID]
-	if peer != nil {
-		peer.height = height
-	} else {
-		peer = newBPPeer(pool, peerID, height)
-		peer.setLogger(pool.Logger.With("peer", peerID))
-		pool.peers[peerID] = peer
+		// Don't add or update a peer that is not useful.
+		if peer != nil {
```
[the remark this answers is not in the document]
milosevic: That's in case we call updatePeer to add a peer.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
-		if peer.didTimeout {
-			pool.removePeer(peer.id)
-			continue
-		}
-		if peer.numPending >= maxPendingRequestsPerPeer {
-			continue
-		}
-		if peer.height < minHeight {
-			continue
+		if err := peer.isGood(); err != nil {
```
milosevic: It seems that result of `peer.isGood` depends on the instant at which it is called. Maybe it would be better isolating failure detection in a separate module that would inform other part of the code in case it detects the failure.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
 
-func (pool *BlockPool) makeNextRequester() {
-	pool.mtx.Lock()
-	defer pool.mtx.Unlock()
-
-	nextHeight := pool.height + pool.requestersLen()
-	if nextHeight > pool.maxPeerHeight {
-		return
+func (pool *blockPool) makeRequestBatch(maxNumPendingRequests int32) []int {
+	pool.removeBadPeers()
```
milosevic: It might be easier (to understand and test) if these kind of actions are event driven.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
+	// If running low on planned requests, make more.
+	numNeeded := int32(cmn.MinInt(int(maxNumPendingRequests), len(pool.peers)*int(maxRequestsPerPeer))) - pool.numPending
+	for int32(len(pool.requests)) < numNeeded {
+		if pool.nextRequestHeight > pool.maxPeerHeight {
```
milosevic: Do you handle a case where a request failed and you can find a new peer that can serve that request at that point of time, but some time after there is a peer that can provide it? Imagine for example that there is a peer p that can serve failed request but we can't send to him as number of pending peer requests for p is reached. The problematic case is the one where there is a hole in a sequence. In that case you want to send request for such a block as soon as possible, and this can be lower than `pool.nextRequestHeight`.

blockchain/peer.go
```diff
@@ -0,0 +1,134 @@
...
+	if peer.numPending == 0 {
+		peer.timeout.Stop()
+	} else {
+		peer.recvMonitor.Update(recvSize)
+		peer.resetTimeout()
+	}
+}
+
+func (peer *bpPeer) onTimeout() {
+	peer.errFunc(errNoPeerResponse, peer.id)
```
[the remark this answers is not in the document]
ancazamfir: removed didTimeout

blockchain/peer.go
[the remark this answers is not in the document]
ancazamfir: removed didTimeout

blockchain/peer.go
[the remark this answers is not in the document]
ancazamfir: called from pool.cleanup() with next commit.

blockchain/peer_test.go
```diff
@@ -0,0 +1,241 @@
...
+		if running {
+			assert.True(t, stopped)
+		} else {
+			assert.False(t, stopped)
+		}
+	})
+}
+
+func TestPeerResetMonitor(t *testing.T) {
+	peer := &bpPeer{
```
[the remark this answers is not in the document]
ancazamfir: fixed

blockchain/peer_test.go
```diff
@@ -0,0 +1,241 @@
...
+		errFunc: errFunc,
+	}
+	peer.resetMonitor()
+	assert.NotNil(t, peer.recvMonitor)
+}
+
+func TestPeerTimer(t *testing.T) {
+	peerTimeout = 2 * time.Millisecond
+
+	peer := &bpPeer{
```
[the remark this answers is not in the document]
ancazamfir: fixed

blockchain/pool.go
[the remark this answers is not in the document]
ancazamfir: Not sure, this is called when we just removed the tallest peer (there is a check in the caller). We need to go through the remaining peer set to figure out if there is a different max and set newMaxPeerHeight

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
+	if peer == nil {
+		// Add new peer.
+		peer = newBPPeer(peerID, height, pool.toBcR.sendPeerError)
+		peer.setLogger(pool.logger.With("peer", peerID))
+		pool.peers[peerID] = peer
+	} else {
+		// Update existing peer.
+		// Remove any requests made for heights in range (height, peer.height].
+		for h, block := range pool.peers[peerID].blocks {
+			if h <= height {
```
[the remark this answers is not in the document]
ancazamfir: changed to not allow peers to lower their heights

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
 
-func (pool *BlockPool) removePeer(peerID p2p.ID) {
-	for _, requester := range pool.requesters {
-		if requester.getPeerID() == peerID {
-			requester.redo(peerID)
+// Stops the peer timer and deletes the peer. Recomputes the max peer height.
+func (pool *blockPool) deletePeer(peerID p2p.ID) {
+	if p, ok := pool.peers[peerID]; ok {
+		if p.timeout != nil {
+			p.timeout.Stop()
```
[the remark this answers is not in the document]
ancazamfir: yup, changed

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
+		if p.height == pool.maxPeerHeight {
+			pool.updateMaxPeerHeight()
 		}
+	}
+}
 
-		delete(pool.peers, peerID)
+// Removes any blocks and requests associated with the peer and deletes the peer.
+// Also triggers new requests if blocks have been removed.
+func (pool *blockPool) removePeer(peerID p2p.ID, err error) {
```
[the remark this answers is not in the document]
ancazamfir: added in the log

blockchain/pool.go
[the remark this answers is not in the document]
ancazamfir: IIRC, the old code was starting from store.Height() while here we start with store.Height()+1.
So the meaning of pool.height changes, it is the next block the pool would need to execute and not the latest executed.

blockchain/peer_test.go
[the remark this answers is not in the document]
ancazamfir: Ok I added the const for now. This was a pain for me as I would forget this set at msec and other tests (reactor_test.go) would fail when running on the full package. I think it should be included as a field in reactor or pool and in general I would love to get rid of all global vars.

blockchain/peer_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/p2p"
+)
+
+var (
+	numErrFuncCalls int
+	lastErr         error
+	peerTestMtx     sync.Mutex
```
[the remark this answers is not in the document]
ancazamfir: added

blockchain/pool.go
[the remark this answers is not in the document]
ancazamfir: > Is the first part of this condition even necessary?
No, removed.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
-			// sleep for a bit.
-			time.Sleep(requestIntervalMS * time.Millisecond)
-			// check for timed out peers
-			pool.removeTimedoutPeers()
-		} else {
-			// request for more blocks.
-			pool.makeNextRequester()
+	if bd.block == nil {
+		if bd.peer == nil {
+			return fmt.Sprintf("block: nil peer: nil")
```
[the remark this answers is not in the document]
ancazamfir: I will look at this again, it had commas before but with many peers and block it was harder to read. Will get back on this.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
-	if peerID != p2p.ID("") {
-		// RemovePeer will redo all requesters associated with this peer.
-		pool.removePeer(peerID)
-	}
-	return peerID
+func (pool *blockPool) rescheduleRequest(peerID p2p.ID, height int64) {
+	pool.logger.Debug("reschedule requests made to peer for height ", "peerID", peerID, "height", height)
+	pool.requests[height] = true
+	delete(pool.blocks, height)
+	delete(pool.peers[peerID].blocks, height)
```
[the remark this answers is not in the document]
ancazamfir: Not sure what you mean by "stopped". The blockRequest message may have been sent out. A block may have been received or not.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
 	}
 
-	if requester.setBlock(block, peerID) {
-		atomic.AddInt32(&pool.numPending, -1)
-		peer := pool.peers[peerID]
-		if peer != nil {
-			peer.decrPending(blockSize)
+	if newMax < pool.maxPeerHeight {
+		// Remove any planned requests for heights over the new maxPeerHeight.
+		// This may happen if a peer has updated with lower height.
```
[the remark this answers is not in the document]
ancazamfir: I updated to not allow peers to lower their height. But there is still the case where the tallest peer is removed.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
+	return nil
 }
 
-func (pool *BlockPool) removePeer(peerID p2p.ID) {
-	for _, requester := range pool.requesters {
-		if requester.getPeerID() == peerID {
-			requester.redo(peerID)
+// Stops the peer timer and deletes the peer. Recomputes the max peer height.
+func (pool *blockPool) deletePeer(peerID p2p.ID) {
+	if p, ok := pool.peers[peerID]; ok {
```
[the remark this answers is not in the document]
ancazamfir: We cannot since the caller currently makes sure the peer exists. I changed to pass *bpPeer...but still check for nil.

blockchain/peer.go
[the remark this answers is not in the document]
ancazamfir: We could but I don't think it is needed. We reset the monitor when numPending goes from 0 to 1. We check the peer for slowness (by checking the monitor) only when numPending > 0.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
+// Also triggers new requests if blocks have been removed.
+func (pool *blockPool) removePeer(peerID p2p.ID, err error) {
+	pool.logger.Debug("removing peer", "peerID", peerID)
 
-		// Find a new peer with the biggest height and update maxPeerHeight if the
-		// peer's height was the biggest.
-		if peer.height == pool.maxPeerHeight {
-			pool.updateMaxPeerHeight()
+	peer := pool.peers[peerID]
+	if peer == nil {
```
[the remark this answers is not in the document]
ancazamfir: If switch removes a peer we already removed. Or if we receive a block from a peer we don't have, or was removed.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
+		return
+	}
+	// Reschedule the requests for all blocks waiting for the peer, or received and not processed yet.
+	for h, block := range pool.peers[peerID].blocks {
+		if block == nil {
+			pool.numPending--
 		}
+		pool.rescheduleRequest(peerID, h)
 	}
+	pool.deletePeer(peerID)
```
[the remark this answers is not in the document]
ancazamfir: changed

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
-			continue
-		}
-		if peer.numPending >= maxPendingRequestsPerPeer {
-			continue
-		}
-		if peer.height < minHeight {
-			continue
+		if err := peer.isGood(); err != nil {
+			pool.removePeer(peer.id, err)
+			if err == errSlowPeer {
```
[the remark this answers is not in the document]
ancazamfir: Right now the only err is errSlowPeer but you are right it should be on any err. Removed the check.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
-}
-
-func (peer *bpPeer) resetTimeout() {
-	if peer.timeout == nil {
-		peer.timeout = time.AfterFunc(peerTimeout, peer.onTimeout)
-	} else {
-		peer.timeout.Reset(peerTimeout)
+// Validates that the block comes from the peer it was expected from and stores it in the 'blocks' map.
+func (pool *blockPool) addBlock(peerID p2p.ID, block *types.Block, blockSize int) error {
+	if _, ok := pool.peers[peerID]; !ok {
```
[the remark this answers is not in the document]
ancazamfir: changed.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
-func (peer *bpPeer) onTimeout() {
-	peer.pool.mtx.Lock()
-	defer peer.pool.mtx.Unlock()
 
-	err := errors.New("peer did not send us anything")
-	peer.pool.sendError(err, peer.id)
-	peer.logger.Error("SendTimeout", "reason", err, "timeout", peerTimeout)
-	peer.didTimeout = true
+	pool.peers[peerID].blocks[block.Height] = block
+	pool.blocks[block.Height] = peerID
```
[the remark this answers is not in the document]
ancazamfir: yup, removed.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
+	first, err1 := pool.getBlockAndPeerAtHeight(pool.height)
+	second, err2 := pool.getBlockAndPeerAtHeight(pool.height + 1)
 
-// Returns true if the peer matches and block doesn't already exist.
-func (bpr *bpRequester) setBlock(block *types.Block, peerID p2p.ID) bool {
-	bpr.mtx.Lock()
-	if bpr.block != nil || bpr.peerID != peerID {
-		bpr.mtx.Unlock()
-		return false
+	if err1 == nil {
```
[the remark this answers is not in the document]
ancazamfir: They shouldn't be non-nil but the logic is in the caller and a bit far away, in the reactor processBlock(). Indeed it will never call this if either blocks are missing. 
At this point invalidateFirstTwoBlocks() is more generic than the caller needs it to be and invalidates the blocks if they exists, does nothing if they don't. I even wrote tests for these cases :)
I could remove the checks and tests if that's what you have in mind (??)

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
-					continue WAIT_LOOP
-				}
-			case <-bpr.gotBlockCh:
-				// We got a block!
-				// Continue the for-loop and wait til Quit.
-				continue WAIT_LOOP
-			}
+func (pool *blockPool) cleanup() {
+	for _, peer := range pool.peers {
+		if peer.timeout != nil {
```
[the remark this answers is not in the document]
ancazamfir: changed

blockchain/pool_test.go
```diff
@@ -1,224 +1,669 @@
...
+			poolWanted: makeBlockPool(testBcR, 100, []bpPeer{{id: "P1", height: 101}}, map[int64]tPBlocks{}),
+		},
+		{
+			name:       "increase the height of P1 from 120 to 123",
+			pool:       makeBlockPool(testBcR, 100, []bpPeer{{id: "P1", height: 120}}, map[int64]tPBlocks{}),
+			args:       testPeer{"P1", 123},
+			poolWanted: makeBlockPool(testBcR, 100, []bpPeer{{id: "P1", height: 123}}, map[int64]tPBlocks{}),
+		},
+		{
+			name:       "decrease the height of P1 from 120 to 110 (still over current processing height)",
```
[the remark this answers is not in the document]
ancazamfir: fixed

blockchain/reactor.go
```diff
@@ -37,35 +32,73 @@ const (
...
-	return fmt.Sprintf("error with peer %v: %s", e.peerID, e.err.Error())
-}
-
 // BlockchainReactor handles long-term catchup syncing.
 type BlockchainReactor struct {
 	p2p.BaseReactor
 
 	// immutable
 	initialState sm.State
+	state        sm.State
```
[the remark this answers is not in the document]
ancazamfir: yes, fixed, it was for the initialState.

blockchain/reactor.go
```diff
@@ -140,23 +166,17 @@ func (bcR *BlockchainReactor) GetChannels() []*p2p.ChannelDescriptor {
 // AddPeer implements Reactor by sending our state to peer.
 func (bcR *BlockchainReactor) AddPeer(peer p2p.Peer) {
 	msgBytes := cdc.MustMarshalBinaryBare(&bcStatusResponseMessage{bcR.store.Height()})
-	if !peer.Send(BlockchainChannel, msgBytes) {
+	if !peer.TrySend(BlockchainChannel, msgBytes) {
```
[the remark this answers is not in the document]
ancazamfir: ah, a leftover from an earlier wild goose chase. changed back.

blockchain/peer.go
ancazamfir: peerTimer?

blockchain/peer.go
ancazamfir: but it's just one field: numPending and it reflects the number of entries in blocks for which we don't have blocks yet. If we have two datastructs we still need to bring them together under the peer. Or maybe I don't understand?

blockchain/peer.go
ancazamfir: We monitor when we have pending requests (numPending > 0), so when we go from 0 to 1 we (re)start the monitor.

blockchain/peer.go
ancazamfir: Yes, we discussed this and I kept delaying making the change, my bad. I will look into it soon!

blockchain/peer.go
ancazamfir: What if I change the name to receivedBlock(recvSize int)...?

blockchain/peer_test.go
ancazamfir: I was thinking to keep it around and do a commit dedicated to cleanup :)
The question I have is what do we do once we are in finished state. We leave quite a few stale things around right now and I plan to clean. We do stop timers I think, but peers are left around, etc. IMO we should scan the pool when finished and verify it is in the correct state and remove peers.

blockchain/pool.go
ancazamfir: done

blockchain/pool.go
ancazamfir: If a request cannot be sent it is left in the requests[] list and retried later. The list is sorted by height so the lowest block in the list will be the one requested first.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
+		for h := range pool.requests {
+			if h > newMax {
+				delete(pool.requests, h)
+			}
 		}
-	} else {
-		pool.Logger.Info("invalid peer", "peer", peerID, "blockHeight", block.Height)
-		pool.sendError(errors.New("invalid peer"), peerID)
 	}
+	if pool.nextRequestHeight > newMax {
```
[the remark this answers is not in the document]
ancazamfir: Here are the steps:
1. All the block requests made to the 150-tall peer or blocks received from the 150-tall peer that are still around (not executed) are removed. Let's say blocks 45..100 (can be a mixture of received and requested)
2. All the heights of these removed blocks are added to the requests slice, i.e. it will be [45..100]
3. The peer is deleted and the new max is set to 50
4. Any of the heights in the requests slice that are higher than the new max are removed from, ending up in this example with [45..50].
5. nextRequestHeight is set to 51.

I have reorganized the code here, hoping it's easier to read..it was all over the place before.
And will add some tests for these next.

blockchain/pool.go
ancazamfir: changed to plannedRequests...since not all are really failed, some we have left over from previous sending of requests processing if we go over the maxNumPending or per peer limit. So these are planned for the next time we are asked for new request by the reactor.

blockchain/pool.go
```diff
@@ -1,630 +1,383 @@
...
+	oldMaxPeerHeight := pool.maxPeerHeight
+	// Delete the peer. This operation may result in the pool's maxPeerHeight being lowered.
+	pool.deletePeer(peer)
+
+	// Check if the pool's maxPeerHeight has been lowered.
+	// This may happen if the tallest peer has been removed.
+	if oldMaxPeerHeight > pool.maxPeerHeight {
+		// Remove any planned requests for heights over the new maxPeerHeight.
+		for h := range pool.plannedRequests {
+			if h > pool.maxPeerHeight {
```
milosevic: There is a lot of logic that is `maxPeerHeight` related. I am still not convinced that it's a good idea managing this variable.

blockchain/pool.go
```diff
@@ -1,630 +1,383 @@
...
+// Make a batch of requests sorted by height. The parameter 'maxNumRequests' includes the number of block requests
+// already made.
+func (pool *blockPool) makeRequestBatch(maxNumRequests int32) []int {
+	pool.removeBadPeers()
+	// At this point pool.requests may include heights for requests to be redone due to removal of peers:
+	// - peers timed out or were removed by switch
+	// - FSM timed out on waiting to advance the block execution due to missing blocks at h or h+1
+	// Check if more requests should be tried by subtracting the number of requests already made from the maximum allowed
+	numNeeded := int(maxNumRequests) - len(pool.blocks)
+	for len(pool.plannedRequests) < numNeeded {
```
milosevic: I find hard understanding this logic. I guess part of the problem might be the fact that we use `len(pool.blocks)` and `len(pool.plannedRequests)` as something semantic and it's not obvious what it corresponds to.

blockchain/reactor_fsm.go
```diff
@@ -0,0 +1,430 @@
...
+}
+
+// Blockchain Reactor State Machine
+type bReactorFSM struct {
+	logger log.Logger
+	mtx    sync.Mutex
+
+	startTime time.Time
+
+	state      *bReactorFSMState
```
milosevic: It is surprising that state of FSM is bReactorFSMState. I would expect state to hold protocol specific variables.

blockchain/reactor_fsm.go
```diff
@@ -0,0 +1,430 @@
...
+	"github.com/tendermint/tendermint/p2p"
+	"github.com/tendermint/tendermint/types"
+)
+
+// Blockchain Reactor State
+type bReactorFSMState struct {
+	name string
+
+	// called when transitioning out of current state
+	handle func(*bReactorFSM, bReactorEvent, bReactorEventData) (next *bReactorFSMState, err error)
```
milosevic: What is a reason for not using a single parameter of type`bcReactorMessage` instead of two parameters?

blockchain/reactor_fsm.go
```diff
@@ -0,0 +1,430 @@
...
+	case stateTimeoutEv:
+		return "stateTimeoutEv"
+	default:
+		return "event unknown"
+	}
+
+}
+
+// states
+var (
```
milosevic: Isn't it simpler to have a field in FSM that capture state (of Enum type for example)?

blockchain/reactor_fsm.go
```diff
@@ -0,0 +1,430 @@
...
+			}
+		},
+	}
+
+	waitForPeer = &bReactorFSMState{
+		name:    "waitForPeer",
+		timeout: waitForPeerTimeout,
+		enter: func(fsm *bReactorFSM) {
+			// Stop when leaving the state.
+			fsm.resetStateTimer()
```
milosevic: How we can test that timer is in expected state?

blockchain/reactor_fsm.go
```diff
@@ -0,0 +1,430 @@
...
+				if data.stateName != "waitForPeer" {
+					fsm.logger.Error("received a state timeout event for different state", "state", data.stateName)
+					return waitForPeer, errTimeoutEventWrongState
+				}
+				// There was no statusResponse received from any peer.
+				// Should we send status request again?
+				return finished, errNoTallerPeer
+
+			case statusResponseEv:
+				if err := fsm.pool.updatePeer(data.peerId, data.height); err != nil {
```
milosevic: Aren't state update supposed to be reflected in output (return) state?

blockchain/reactor_fsm.go
```diff
@@ -0,0 +1,430 @@
...
+					// Ignore block, remove peer and send error to switch.
+					fsm.pool.removePeer(data.peerId, err)
+					fsm.toBcR.sendPeerError(err, data.peerId)
+				}
+				return waitForBlock, err
+
+			case processedBlockEv:
+				fsm.logger.Debug("processedBlockEv", "err", data.err)
+				first, second, _ := fsm.pool.getNextTwoBlocks()
+				if data.err != nil {
```
milosevic: It's not clear where we do block validation.

blockchain/reactor_fsm.go
```diff
@@ -0,0 +1,430 @@
...
+}
+
+// Starts the FSM.
+func (fsm *bReactorFSM) start() {
+	_ = fsm.handle(&bcReactorMessage{event: startFSMEv})
+}
+
+// handle processes messages and events sent to the FSM.
+func (fsm *bReactorFSM) handle(msg *bcReactorMessage) error {
+	fsm.mtx.Lock()
```
milosevic: Why do you need lock on fsm?

melekes:
Looks like tests are failing

blockchain/peer.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/p2p"
+	"github.com/tendermint/tendermint/types"
+)
+
+const defaultPeerTimeout = 15 * time.Second
+
+//--------
+// Peer
+var (
```
melekes: we should try to avoid mutable global state. `peerTimeout` can be a config variable passed during initialization.

blockchain/peer.go
```diff
@@ -0,0 +1,130 @@
...
+	peerSampleRate = time.Second
+	peerWindowSize = 40 * peerSampleRate
+)
+
+type bpPeer struct {
+	logger log.Logger
+	id     p2p.ID
+
+	height      int64                  // the peer reported height
+	numPending  int32                  // number of requests still waiting for block responses
```
melekes: `numPendingBlockRequests`? comments are good, but we should try to capture the meaning in the name as much as possible

blockchain/peer.go
```diff
@@ -0,0 +1,130 @@
...
+	blocks      map[int64]*types.Block // blocks received or expected to be received from this peer
+	timeout     *time.Timer
+	recvMonitor *flow.Monitor
+
+	errFunc func(err error, peerID p2p.ID) // function to call on error
```
melekes: `onErr`?

blockchain/peer.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func newBPPeer(
+	peerID p2p.ID, height int64, errFunc func(err error, peerID p2p.ID)) *bpPeer {
+	peer := &bpPeer{
+		id:         peerID,
+		height:     height,
+		blocks:     make(map[int64]*types.Block, maxRequestsPerPeer),
+		numPending: 0,
```
melekes: can be removed

blockchain/peer.go
```diff
@@ -0,0 +1,130 @@
...
+	blocks      map[int64]*types.Block // blocks received or expected to be received from this peer
+	timeout     *time.Timer
+	recvMonitor *flow.Monitor
+
+	errFunc func(err error, peerID p2p.ID) // function to call on error
+}
+
+func newBPPeer(
+	peerID p2p.ID, height int64, errFunc func(err error, peerID p2p.ID)) *bpPeer {
+	peer := &bpPeer{
```
melekes: `return &bpPeer{`

blockchain/peer.go
```diff
@@ -0,0 +1,130 @@
...
+		height:     height,
+		blocks:     make(map[int64]*types.Block, maxRequestsPerPeer),
+		numPending: 0,
+		logger:     log.NewNopLogger(),
+		errFunc:    errFunc,
+	}
+	return peer
+}
+
+func (peer *bpPeer) setLogger(l log.Logger) {
```
melekes: Why it's not public? Same questions for other bpPeer methods. We can keep bpPeer private, but I see no reasons to not make its methods public - at least those what form public API for modifying its state. like incrPending, decrPending, ...

blockchain/peer.go
```diff
@@ -0,0 +1,130 @@
...
+		peer.resetTimeout()
+	}
+}
+
+func (peer *bpPeer) onTimeout() {
+	peer.errFunc(errNoPeerResponse, peer.id)
+}
+
+func (peer *bpPeer) isGood() error {
+	if peer.numPending > 0 {
```
melekes:
```go
if peer.numPending == 0 {
  return nil
}
```

blockchain/peer.go
```diff
@@ -0,0 +1,130 @@
...
+		curRate := peer.recvMonitor.Status().CurRate
+		// curRate can be 0 on start
+		if curRate != 0 && curRate < minRecvRate {
+			err := errSlowPeer
+			peer.logger.Error("SendTimeout", "peer", peer.id,
+				"reason", err,
+				"curRate", fmt.Sprintf("%d KB/s", curRate/1024),
+				"minRate", fmt.Sprintf("%d KB/s", minRecvRate/1024))
+			// consider the peer timedout
```
melekes: Is this a TODO or what? not sure I understand

blockchain/peer.go
melekes: The struct already has Peer in its name. No need to duplicate. `blockResponseTimer`?

blockchain/peer_test.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/stretchr/testify/require"
+
+	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/p2p"
+)
+
+// These vars are used to record and verify different events generated during tests
+var (
+	numErrFuncCalls int        // number of calls to the errFunc
```
melekes: avoid mutable global state. This should probably be a part of `bpTestPeer` or something

blockchain/peer_test.go
```diff
@@ -0,0 +1,216 @@
...
+	_ = peerID
+	lastErr = err
+	numErrFuncCalls++
+}
+
+// check if peer timer is running or not (a running timer can be successfully stopped)
+// Note: it does stop the timer!
+func checkByStoppingPeerTimer(t *testing.T, peer *bpPeer, running bool) {
+	assert.NotPanics(t, func() {
+		stopped := peer.timeout.Stop()
```
melekes: please avoid accessing internal state in tests as much as possible. otherwise it will be almost impossible to change the struct without breaking tests.

blockchain/peer.go
ancazamfir: I changed them but not convinced as they are internal to the package.

blockchain/peer.go
ancazamfir: changed to blockResponseTimer

blockchain/peer.go
ancazamfir: Forgot to remove this comment, code used to set a didTimeout field to true...removed now.

blockchain/peer.go
```diff
@@ -0,0 +1,185 @@
...
+	flow "github.com/tendermint/tendermint/libs/flowrate"
+	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/p2p"
+	"github.com/tendermint/tendermint/types"
+)
+
+//--------
+// Peer
+type bpPeerParams struct {
+	peerTimeout    time.Duration
```
melekes: peer prefix is not necessary imho

blockchain/peer.go
```diff
@@ -0,0 +1,185 @@
...
+	minRecvRate    int64
+	peerSampleRate time.Duration
```
melekes: same

blockchain/peer.go
```diff
@@ -0,0 +1,185 @@
...
+	peerWindowSize time.Duration
```
melekes: same

blockchain/peer.go
```diff
@@ -0,0 +1,185 @@
...
+	id     p2p.ID
+
+	height                  int64                  // the peer reported height
+	numPendingBlockRequests int32                  // number of requests still waiting for block responses
+	blocks                  map[int64]*types.Block // blocks received or expected to be received from this peer
+	blockResponseTimer      *time.Timer
+	recvMonitor             *flow.Monitor
+	parameters              *bpPeerParams // parameters for timer and monitor
+
+	errFunc func(onErr error, peerID p2p.ID) // function to call on error
```
melekes:
```suggestion
	onErr func(err error, peerID p2p.ID)
```

blockchain/peer.go
```diff
@@ -0,0 +1,185 @@
...
+type bpPeer struct {
+	logger log.Logger
+	id     p2p.ID
+
+	height                  int64                  // the peer reported height
+	numPendingBlockRequests int32                  // number of requests still waiting for block responses
+	blocks                  map[int64]*types.Block // blocks received or expected to be received from this peer
+	blockResponseTimer      *time.Timer
+	recvMonitor             *flow.Monitor
+	parameters              *bpPeerParams // parameters for timer and monitor
```
melekes:
```suggestion
	params              *bpPeerParams // parameters for timer and monitor
```

blockchain/peer.go
```diff
@@ -0,0 +1,185 @@
...
+
+	errFunc func(onErr error, peerID p2p.ID) // function to call on error
+
+}
+
+func NewBPPeer(
```
melekes: godoc for exposed functions

blockchain/peer.go
```diff
@@ -0,0 +1,185 @@
...
+
+func (peer *bpPeer) ID() p2p.ID {
+	return peer.id
+}
+
+func (peer *bpPeer) SetLogger(l log.Logger) {
+	peer.logger = l
+}
+
+func (peer *bpPeer) GetHeight() int64 {
```
melekes: https://golang.org/doc/effective_go.html#Getters

blockchain/peer.go
```diff
@@ -0,0 +1,185 @@
...
+	return peer.height
+}
+
+func (peer *bpPeer) SetHeight(height int64) {
+	peer.height = height
+}
+
+func (peer *bpPeer) GetNumPendingBlockRequests() int32 {
```
melekes: https://golang.org/doc/effective_go.html#Getters

blockchain/peer.go
```diff
@@ -0,0 +1,185 @@
...
+	return peer.numPendingBlockRequests
+}
+
+func (peer *bpPeer) GetBlockAtHeight(height int64) (*types.Block, error) {
```
melekes: https://golang.org/doc/effective_go.html#Getters

blockchain/peer.go
melekes: Alternatively, you can just expose `Height` field. Then you won't need getters and setters

blockchain/peer.go
```diff
@@ -0,0 +1,185 @@
...
+		peer.recvMonitor.Update(recvSize)
+		peer.resetTimeout()
+	}
+}
+
+func (peer *bpPeer) onTimeout() {
+	peer.errFunc(errNoPeerResponse, peer.id)
+}
+
+func (peer *bpPeer) IsGood() error {
```
melekes: weird to see a method with `Is` prefix return an error (when you expect it to return boolean). Maybe add a comment?

blockchain/peer_test.go
```diff
@@ -0,0 +1,219 @@
...
+	})
+}
+
+func TestPeerResetMonitor(t *testing.T) {
+	peer := NewBPPeer(
+		p2p.ID(cmn.RandStr(12)), 10,
+		func(err error, _ p2p.ID) {},
+		nil)
+	peer.SetLogger(log.TestingLogger())
+	peer.resetMonitor()
```
melekes: do we have to test this method separately from desc/incrPending functions? As far as I understand, we only call it there.

blockchain/peer.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	return peer.blocks[height], nil
+}
+
+func (peer *bpPeer) SetBlockAtHeight(height int64, block *types.Block) {
+	peer.blocks[height] = block
+}
+
+func (peer *bpPeer) RemoveBlockAtHeight(height int64) {
+	if _, ok := peer.blocks[height]; !ok {
```
melekes: no need for a check. from delete doc: `If m is nil or there is no such element, delete is a no-op.`

blockchain/peer.go
```diff
@@ -0,0 +1,185 @@
...
+
+func (peer *bpPeer) StopBlockResponseTimer() bool {
+	if peer.blockResponseTimer == nil {
+		return false
+	}
+	return peer.blockResponseTimer.Stop()
+}
+
+func (peer *bpPeer) DecrPending(recvSize int) {
+	if peer.GetNumPendingBlockRequests() == 0 {
```
melekes:
```suggestion
	if peer.numPendingBlockRequests == 0 {
```

blockchain/peer.go
```diff
@@ -0,0 +1,185 @@
...
+		panic("cannot decrement, peer does not have pending requests")
+	}
+
+	peer.numPendingBlockRequests--
+	if peer.GetNumPendingBlockRequests() == 0 {
```
melekes:
```suggestion
	if peer.numPendingBlockRequests == 0 {
```

blockchain/peer.go
melekes: Confusion partially comes from the name: `resetMonitor`
when in fact it's more like: `initOrResetMonitor`

blockchain/peer_test.go
```diff
@@ -0,0 +1,219 @@
...
+	"github.com/stretchr/testify/require"
+
+	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/p2p"
+)
+
+// check if peer timer is running or not (a running timer can be successfully stopped)
+// Note: it does stop the timer!
+func checkByStoppingPeerTimer(t *testing.T, peer *bpPeer, running bool) {
```
melekes: private funcs usually are located at the end of the file, not at the top

brapse:
This is a large undertaking that has already delivers big improvement. It looks like we've already split up the blockchain reactor into an reactor and a fsm, the later of which exposes a `handle` method
which performs state transitions. In the current iteration it appears that we are mutating the message with this handle function and ignoring the return value. This appears to diverge from
[adr-40](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-040-blockchain-reactor-refactor.md). Is it safe to assume that this is just a temporary thing still trending towards ADR-40/36?

I ask because the test coverage has already improved dramatically and seems to suggest that the API is stable enough to to start codifying contract in test assertions. However, The current tests rely on a coupling between the `peer`, `reactor` and `reactor_fsm` which I didn't notice in adr-36/40. One example is the reactor_fsm which mutates `pool` internal state of assert correctness in it's tests.I use this example as it was explicitly brought up by the author ;)

https://github.com/tendermint/tendermint/blob/81d121b7d25cddaea6f35cacd5864583f1dada11/blockchain/reactor_fsm_test.go#L191-L202

With tests:
https://github.com/tendermint/tendermint/blob/81d121b7d25cddaea6f35cacd5864583f1dada11/blockchain/reactor_fsm_test.go#L231-L235

I will echo the concerns of @melekes [here](https://github.com/tendermint/tendermint/pull/3561#discussion_r289592041) that These tests might become liabilities as we evolve the dependency graph to ideally provide encapsulation between the `peer`, `reactor` and `reactor_fsm` components of the blockchain package.

So how what's the best way to review the current iteration, as an intermediary implementation of ADR-36/40 or as alternative direction?

blockchain/peer_test.go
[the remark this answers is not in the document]
ancazamfir: removed.

blockchain/peer.go
ancazamfir: changed to startMonitor(), added stopMonitor() for symmetry.

blockchain/pool.go
```diff
@@ -1,630 +1,372 @@
...
-			// check for timed out peers
-			pool.removeTimedoutPeers()
-		} else {
-			// request for more blocks.
-			pool.makeNextRequester()
+	if bd.block == nil {
+		if bd.peer == nil {
+			return fmt.Sprintf("block: nil peer: nil")
 		}
+		return fmt.Sprintf("block: nil peer: %v", bd.peer.id)
```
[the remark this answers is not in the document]
ancazamfir: removed as it was not used anymore.

blockchain/peer_test.go
ancazamfir: Shouldn't IMO

blockchain/reactor_fsm.go
ancazamfir: Because processing blocks (CPU intensive) is in a different goroutine.

blockchain/peer.go
ancazamfir: changed to CheckRate()

ancazamfir:
>  In the current iteration it appears that we are mutating the message with this handle function 

Could you point to the code please?
> and ignoring the return value. 

you mean the error returned from Handle()? At this point it seems that FSM can handle the errors and the return value is informative. May change as we work more on the reactor.go (the neglected part so far). FSM tests do some checks on the return value.

> This appears to diverge from
> [adr-40](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-040-blockchain-reactor-refactor.md). Is it safe to assume that this is just a temporary thing still trending towards ADR-40/36?

Not sure I understand the exact piece of the ADR you refer to. In general the goal is to have this reorg and the ADR be the same. While I am evolving the code from the old one towards the ADR, I also expect some changes and adjustments to the ADR along the way.

> 
> I ask because the test coverage has already improved dramatically and seems to suggest that the API is stable enough to to start codifying contract in test assertions. However, The current tests rely on a coupling between the `peer`, `reactor` and `reactor_fsm` which I didn't notice in adr-36/40. One example is the reactor_fsm which mutates `pool` internal state of assert correctness in it's tests.I use this example as it was explicitly brought up by the author ;)

Oh, you picked on the rotten apple there :) That is a patch to the test as the code that selects the peer for a request is not smart or deterministic at this point and basically depends on golang range mechanics (same as before). So the tests make some requests and then they want to simulated blocks arriving from the correct peer (or not). So yeah, currently it's not nice...but should be fixed.

Can you find other examples so I understand the concern with the "coupling". 
> 
...

> I will echo the concerns of @melekes [here](https://github.com/tendermint/tendermint/pull/3561#discussion_r289592041) that These tests might become liabilities as we evolve the dependency graph to ideally provide encapsulation between the `peer`, `reactor` and `reactor_fsm` components of the blockchain package.
> 
Yes, let's fix as much as we can. 
> So how what is the best way to review the current iteration; as an intermediary implementation of ADR-36/40 or as alternative direction?

Intermediary. Note the PR is in draft for a reason, it is work in progress. All the commits are transformations of old code towards our ideal. Maybe this is not the approach that some would have taken but there was effort and knowledge put in the old code that I didn't want to lose.
It's also always nice to be able to "ship" at any point a version that is better, even if not perfect.

Please feel free to jump in, make comments, but also make changes, etc. This is valid for all ... convinced some Go experts could do wonders here :)

brapse:
Sorry I should have been more specific 🙈 

> Could you point to the code please?

https://github.com/tendermint/tendermint/blob/81d121b7d25cddaea6f35cacd5864583f1dada11/blockchain/reactor.go#L314

By "appears" I mean we are passing in a pointer and ignoring the return value. Looking a bit deeper down the call chain from the handle function, we are then mutating state on the fsm and delegating to an internal `fsm.state.handle`.  This chain is deep. 

https://github.com/tendermint/tendermint/blob/81d121b7d25cddaea6f35cacd5864583f1dada11/blockchain/reactor_fsm.go#L364-L377

This diverges from ADR-40 in so far as the handle function takes a state and returns a state, this chain is shallow.

```go
func handleEvent(state ControllerState, event Event) (ControllerState, Message, TimeoutTrigger, Error) {
   msg = nil
   timeout = nil
   error = nil
   ...
   return state, msg, timeout, error
}
```

The ADR-40 version, where we want to get to, will definitely be easier to test as we won't have any internal structures to mock out or worry about, just list of transformations that can be stated as assertions ☺️.

> Can you find other examples so I understand the concern with the "coupling".

https://github.com/tendermint/tendermint/blob/81d121b7d25cddaea6f35cacd5864583f1dada11/blockchain/reactor_fsm.go#L243-L249
Here the fsm is checking how many peers in a pool.

https://github.com/tendermint/tendermint/blob/81d121b7d25cddaea6f35cacd5864583f1dada11/blockchain/reactor_fsm.go#L286-L294

Here the fsm is changing the internal structure of the pool.

https://github.com/tendermint/tendermint/blob/81d121b7d25cddaea6f35cacd5864583f1dada11/blockchain/reactor.go#L461-L468

Here the blockchain reactor looks like it's telling the consensus reactor what to do. 

In these cases we are connecting state via method calls.  These connections form a dependency graph which we end up constructing and encoding in tests. 

Maybe this is crazy but perhaps the probelm is that we don't have one but several distinct state machines that are tangled up in the blockchain package 🤔?

What if, Instead of having a single state machine which manipulate state in nested objects (fsm, pool, peer), we have a state machine per state, ie: `fsm.handle`,` pool.handle`, `peer.handle`. and model dependencies exclusively with events. Peer communication could be triggered by events but handled exclusively by the reactor. This schema would flatten the dependency graph and allow each state machine to be tested independently.  The reactor could stitch them together and the reactor test would constitute an integration test between state machines.

If this is totally crazy then please ignore, I just wanted to provide some concrete suggestion on de-coupling the state above. There are probably many paths to ADR-40 and this might only be one. Would be interested in your thoughts here and contributing code to fleshing it out if that might be helpful.

blockchain/peer.go
```diff
@@ -0,0 +1,212 @@
...
+	sampleRate  time.Duration
+	windowSize  time.Duration
+}
+
+type bpPeer struct {
+	logger log.Logger
+	ID     p2p.ID
+
+	Height                  int64                  // the peer reported height
+	NumPendingBlockRequests int32                  // number of requests still waiting for block responses
```
melekes: why `int32` specifically? not just `int`?

blockchain/peer.go
```diff
@@ -0,0 +1,212 @@
...
+	peer.NumPendingBlockRequests++
+}
+
+// BlockReceived handles a block arrival at peer level. Block must be non-nil and recvSize a positive integer
+func (peer *bpPeer) BlockReceived(block *types.Block, recvSize int) error {
+	if block == nil || recvSize < 0 {
+		panic("bad parameters")
+	}
+	if peer.NumPendingBlockRequests == 0 {
+		panic("peer does not have pending requests")
```
melekes: shouldn't we log and return errMissingRequest here too? otherwise any malicious peer will be able to crash our node by sending us a block

blockchain/peer.go
```diff
@@ -0,0 +1,212 @@
...
+	return fmt.Sprintf("peer: %v height: %v pending: %v", peer.ID, peer.Height, peer.NumPendingBlockRequests)
+}
+
+// SetLogger sets the logger of the peer.
+func (peer *bpPeer) SetLogger(l log.Logger) {
+	peer.logger = l
+}
+
+// SetHeight sets the height of the peer.
+func (peer *bpPeer) SetHeight(height int64) {
```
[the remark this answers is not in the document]
ancazamfir: yes, removed

blockchain/peer.go
ancazamfir: Not sure, it was like this in old code. Changed to int.

blockchain/peer.go
ancazamfir: Ouch, broken from last commit. The checks under should be before this.

blockchain/reactor_fsm.go
```diff
@@ -0,0 +1,443 @@
...
+func (msg *bcReactorMessage) String() string {
+	var dataStr string
+
+	switch msg.event {
+	case startFSMEv:
+		dataStr = ""
+	case statusResponseEv:
+		dataStr = fmt.Sprintf("peer=%v height=%v", msg.data.peerId, msg.data.height)
+	case blockResponseEv:
+		dataStr = fmt.Sprintf("peer=%v block.height=%v lenght=%v",
```
golangcibot: `lenght` is a misspelling of `length` (from `misspell`)

blockchain/reactor_fsm.go
```diff
@@ -0,0 +1,443 @@
...
+		dataStr = ""
+	case stopFSMEv:
+		dataStr = ""
+	case peerRemoveEv:
+		dataStr = fmt.Sprintf("peer: %v is being removed by the switch", msg.data.peerId)
+	case stateTimeoutEv:
+		dataStr = fmt.Sprintf("state=%v", msg.data.stateName)
+
+	default:
+		dataStr = fmt.Sprintf("cannot interpret message data")
```
golangcibot: ineffectual assignment to `dataStr` (from `ineffassign`)

blockchainexp/reactor_fsm.go
```diff
@@ -0,0 +1,444 @@
...
+func (msg *bcReactorMessage) String() string {
+	var dataStr string
+
+	switch msg.event {
+	case startFSMEv:
+		dataStr = ""
+	case statusResponseEv:
+		dataStr = fmt.Sprintf("peer=%v height=%v", msg.data.peerID, msg.data.height)
+	case blockResponseEv:
+		dataStr = fmt.Sprintf("peer=%v block.height=%v lenght=%v",
```
golangcibot: `lenght` is a misspelling of `length` (from `misspell`)

blockchainexp/reactor_fsm.go
```diff
@@ -0,0 +1,444 @@
...
+		dataStr = ""
+	case stopFSMEv:
+		dataStr = ""
+	case peerRemoveEv:
+		dataStr = fmt.Sprintf("peer: %v is being removed by the switch", msg.data.peerID)
+	case stateTimeoutEv:
+		dataStr = fmt.Sprintf("state=%v", msg.data.stateName)
+
+	default:
+		dataStr = fmt.Sprintf("cannot interpret message data")
```
golangcibot: ineffectual assignment to `dataStr` (from `ineffassign`)

marbar3778:
Can we merge this? If not, let's make a list of things that need to be completed prior to merging.

ancazamfir:
> Can we merge this? If not, let's make a list of things that need to be completed prior to merging.

- review for the blockchain version configuration flag changes and integration with `behavior` from @brapse 
- I plan to do more testing on gaia once SDK release for tendermint v0.32.0 is out

blockchain/v1/reactor.go
```diff
@@ -137,6 +141,7 @@ func (bcR *BlockchainReactor) SetLogger(l log.Logger) {
 
 // OnStart implements cmn.Service.
 func (bcR *BlockchainReactor) OnStart() error {
+	bcR.swReporter = behaviour.NewSwitcReporter(bcR.BaseReactor.Switch)
```
brapse: How do we know the switch has been setup by the time `OnStart` is called?

blockchain/v1/peer.go
```diff
@@ -13,14 +13,17 @@ import (
...
+type BpPeerParams struct {
 	timeout     time.Duration
 	minRecvRate int64
 	sampleRate  time.Duration
 	windowSize  time.Duration
 }
 
-type bpPeer struct {
+// BpPeer is the datastructure associated with a fast sync peer.
+type BpPeer struct {
```
brapse: Would it be possible to make BpPeer private?

brapse:
@ancazamfir Besides the questions above the behaviour and config flags LGTM 👍

blockchain/v1/reactor.go
ancazamfir: Through code inspection :) I could add a check, but probably we don't reach here with nil Switch and crash much earlier.

The switch is created in `NewNode()` by createSwitch() that adds the reactors and calls `SetSwitch()`  for each reactor  :
```
main() -> 
     NewRunNodeCmd(nodeProvider nm.NodeProvider) ->
        n, err := nodeProvider(config, logger) ->
           node.go:NewNode() -> 
              createSwitch(.., bcReactor, ...) -> 
                 sw.AddReactor("BLOCKCHAIN", bcReactor) ->
                     sw.reactors[name] = reactor
                     reactor.SetSwitch(sw)
```

Then the node is started and that causes the reactor to start:
```
main() -> 
     NewRunNodeCmd(nodeProvider nm.NodeProvider) ->
          n, err := nodeProvider(config, logger)
          err := n.Start() ->
              (n *Node) OnStart() -> 
                    n.sw.Start()
                          (sw *Switch) OnStart() -> 
                                 reactor.Start() -> 
                                     (bs *BaseService) Start() -> 
                                           bs.impl.OnStart()
```

blockchain/v1/reactor_test.go
```diff
@@ -0,0 +1,337 @@
...
+	"sort"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	abci "github.com/tendermint/tendermint/abci/types"
+	cfg "github.com/tendermint/tendermint/config"
+	cmn "github.com/tendermint/tendermint/libs/common"
+	dbm "github.com/tendermint/tm-cmn/db"
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

CHANGELOG_PENDING.md
```diff
@@ -18,6 +18,8 @@ program](https://hackerone.com/tendermint).
 
 ### FEATURES:
 
+- [blockchain] \#3561 Blockchain Reorg Refactor, the new reactor currently sits under a feature flag, see [here](https://github.com/tendermint/tendermint/blob/master/config/toml.go#L297) on how to use it, for further information about the versions please see: [ADR-40](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-040-blockchain-reactor-refactor.md) & [ADR-43](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-043-blockchain-riri-org.md)
```
melekes: I'd write smth like

```
- [blockchain] \#3561 Add early version of the new blockchain reactor, which is supposed to be more modular and testable compared to the old version. To try it, you'll have to turn on the `??` feature flag in the config file. NOTE: It's not ready for a production yet. For further information, see [ADR-40](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-040-blockchain-reactor-refactor.md) & [ADR-43](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-043-blockchain-riri-org.md).
```

PR #3575, opened 2019-04-17
bandaid for non-deterministic clist test

kevlubkcm:
Very simple bandaid for #2227
I think the errors are caused from a race condition in the test, not the actual `clist` data structure

* [ ] Wrote tests

### DIFF
--- a/libs/clist/clist_test.go
+++ b/libs/clist/clist_test.go
@@ -262,7 +262,7 @@ func TestWaitChan(t *testing.T) {
 			time.Sleep(time.Duration(cmn.RandIntn(25)) * time.Millisecond)
 		}
 		// apply a deterministic pause so the counter has time to catch up
-		time.Sleep(time.Duration(25) * time.Millisecond)
+		time.Sleep(25 * time.Millisecond)
 		close(done)
 	}()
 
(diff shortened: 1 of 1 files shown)


libs/clist/clist_test.go
```diff
@@ -261,6 +261,8 @@ func TestWaitChan(t *testing.T) {
 			pushed++
 			time.Sleep(time.Duration(cmn.RandIntn(25)) * time.Millisecond)
 		}
+		// apply a deterministic pause so the counter has time to catch up
+		time.Sleep(time.Duration(25) * time.Millisecond)
```
melekes:
```suggestion
		time.Sleep(25 * time.Millisecond)
```

Liamsi: Ideally, we wouldn't use a timer but some channel to signal that the counter actually caught up. But given that this test fails quite often (and we already use sleep with a random timeout 🙈), it's probably worth trying this out 👍 
Thanks @kevlubkcm!

Liamsi:
circle failed with:
```
=== RUN   TestListenerTimeoutReadWrite
--- FAIL: TestListenerTimeoutReadWrite (0.01s)
    socket_listeners_test.go:110: read tcp 127.0.0.1:38914->127.0.0.1:60772: i/o timeout
```
This looks unrelated but is a new non-deterministic test-failure.

libs/clist/clist_test.go
kevlubkcm: yup. was thinking that too but simple change is simple :D

xla:
> This looks unrelated but is a new non-deterministic test-failure.

Non-deterministic test failures be like

hydra

PR #3578, opened 2019-04-17
ci: Pin golang version to 1.12 line

### LINKED ISSUE #3564: ci: Pin the golang version
Currently the CI just uses `circleci/golang` for the docker container.

This means we just pull the latest version of this image, so our golang version will upgrade whenever circle updates this image.

There was some concern over this as it means there are variables that can change in our build process with out our awareness, which may cause extra debugging headaches. So we probably want to pin the go version used.

xla:
We want to avoid accidental upgrades of the runtime beyond patch versions and rather delibrately upgrade once we have confidence the new minor version is stable enough to be considered for our test and build automation.

Fixes #3564 

* [ ] Wrote tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 defaults: &defaults
   working_directory: /go/src/github.com/tendermint/tendermint
   docker:
-    - image: circleci/golang
+    - image: circleci/golang:1.12
   environment:
     GOBIN: /tmp/workspace/bin
 
@@ -195,7 +195,7 @@ jobs:
             name: run localnet and exit on failure
             command: |
               set -x
-              docker run --rm -v "$PWD":/go/src/github.com/tendermint/tendermint -w /go/src/github.com/tendermint/tendermint golang make build-linux
+              docker run --rm -v "$PWD":/go/src/github.com/tendermint/tendermint -w /go/src/github.com/tendermint/tendermint golang:1.12 make build-linux
               make localnet-start &
               ./scripts/localnet-blocks-test.sh 40 5 10 localhost
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -18,6 +18,8 @@
 
 ### IMPROVEMENTS:
 - [rpc] [\#3534](https://github.com/tendermint/tendermint/pull/3534) Add support for batched requests/responses in JSON RPC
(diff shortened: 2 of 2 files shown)


greg-szabo:
Please do not pin the go version. There was an explicit change to remoev that.

xla:
@greg-szabo It's important to understand the decision, could you share the motivation behind it and where it was decided to stick to this strategy?

greg-szabo:
This is the outcome of the release management project we did a few weeks ago. The project (as most projects currently) didn't record a decision graph. Obviously, this can be improved, feel free to implement it for any upcoming projects. Also, you're welcome to join any of the projects we do around infrastructure if you want to add value.

Without trying to recreate the full decision matrix for this, here are a few key points that had an impact on the decision:

Fact: the team forgets to update the version number for CircleCI. We have missed 1.11.5, 1.12.1, 1.12.2, and 1.12.3 at least. This means that some of our releases were built with sub-par Golang compilers with possible security bugs in them at the time.

Fact: we forget to update our build documentation regarding the required Golang version.

Fact: the team prepares for new major releases in advance. We have discussions on new upcoming features (see the discussions about 1.12 before, from Zaki, Ismail, and others) and even discovery projects to see how it impacts our code (see Ismail's 1.12 discovery regarding go modules).

So patch releases are necessary to be updated because we forget them and minor releases can be updated because we already prepared for them.
Our documentation can state "use the latest version of Golang to build Tendermint" without any need for regular documentation updates.

Fact: there is no proof in our issues that updating the version numbers ever impacted our testing code.

Hence the idea that automatic updates ever cause non-deterministic bugs has no basis.

Fact: when a release fails, the first reaction from the team is to ping me.

Hence I consider myself the responsible person for releases. As the responsible person, my decision is to always build releases with the latest Go compiler provided by CircleCI. It is a compromise between running the latest provided by golang.org and fixing it to a version that we're not going to update. CircleCI was quicker in updating the compiler version than we were but they can be somewhat slower than golang.org.
Feel free to take over that responsibility and make different decisions.

If this means that releases can break with the newest version of the compiler, it's all the better. It means we have a _direct feedback_ from our own systems that we need to update the code before we do a release.
If a discovery before a new minor version finds that we _cannot_ support a new version, we can always lock down the Go version for that release. But that's a workaround since we will have to support newer releases, later if not sooner.

melekes:
Looks like this deserves more discussion before we change anything. Let's move to the issue. I'll close the PR in the meantime.

PR #3579, opened 2019-04-18
cs/replay: check appHash for each block

### LINKED ISSUE #3460: replayBlocks checkAppHash should't wait the last block

melekes:
Fixes #3460

Not really needed, but it would detect if the app changed in a way it
shouldn't have.

* [ ] Wrote tests

### DIFF
--- a/consensus/replay.go
+++ b/consensus/replay.go
@@ -332,7 +332,7 @@ func (h *Handshaker) ReplayBlocks(
 
 	// First handle edge cases and constraints on the storeBlockHeight.
 	if storeBlockHeight == 0 {
-		assertAppHashEqualsStateHash(appHash, state)
+		assertAppHashEqualsOneFromState(appHash, state)
 		return appHash, nil
 
 	} else if storeBlockHeight < appBlockHeight {
@@ -360,7 +360,7 @@ func (h *Handshaker) ReplayBlocks(
 
 		} else if appBlockHeight == storeBlockHeight {
 			// We're good!
-			assertAppHashEqualsStateHash(appHash, state)
+			assertAppHashEqualsOneFromState(appHash, state)
 			return appHash, nil
 		}
 
@@ -422,7 +422,7 @@ func (h *Handshaker) replayBlocks(state sm.State, proxyApp proxy.AppConns, appBl
 
 		// Extra check to ensure the app was not changed in a way it shouldn't have.
 		if len(appHash) > 0 {
-			assertAppHashEqualsBlockHash(appHash, block)
+			assertAppHashEqualsOneFromBlock(appHash, block)
 		}
 
 		appHash, err = sm.ExecCommitBlock(proxyApp.Consensus(), block, h.logger, state.LastValidators, h.stateDB)
@@ -442,7 +442,7 @@ func (h *Handshaker) replayBlocks(state sm.State, proxyApp proxy.AppConns, appBl
 		appHash = state.AppHash
 	}
 
-	assertAppHashEqualsStateHash(appHash, state)
+	assertAppHashEqualsOneFromState(appHash, state)
 	return appHash, nil
 }
 
@@ -465,7 +465,7 @@ func (h *Handshaker) replayBlock(state sm.State, height int64, proxyApp proxy.Ap
 	return state, nil
 }
 
-func assertAppHashEqualsBlockHash(appHash []byte, block *types.Block) {
+func assertAppHashEqualsOneFromBlock(appHash []byte, block *types.Block) {
 	if !bytes.Equal(appHash, block.AppHash) {
 		panic(fmt.Sprintf(`block.AppHash does not match AppHash after replay. Got %X, expected %X.
 
@@ -475,7 +475,7 @@ Block: %v
 	}
 }
 
-func assertAppHashEqualsStateHash(appHash []byte, state sm.State) {
+func assertAppHashEqualsOneFromState(appHash []byte, state sm.State) {
 	if !bytes.Equal(appHash, state.AppHash) {
 		panic(fmt.Sprintf(`state.AppHash does not match AppHash after replay. Got
 %X, expected %X.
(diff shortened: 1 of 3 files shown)


consensus/replay.go
```diff
@@ -456,9 +462,26 @@ func (h *Handshaker) replayBlock(state sm.State, height int64, proxyApp proxy.Ap
 	return state, nil
 }
 
-func checkAppHash(state sm.State, appHash []byte) error {
-	if !bytes.Equal(state.AppHash, appHash) {
-		panic(fmt.Errorf("Tendermint state.AppHash does not match AppHash after replay. Got %X, expected %X", appHash, state.AppHash).Error())
```
Liamsi: `panic(fmt.Errorf().Error())`, that was a weird way to panic 🙈

consensus/replay.go
```diff
@@ -456,9 +462,26 @@ func (h *Handshaker) replayBlock(state sm.State, height int64, proxyApp proxy.Ap
...
+func checkAppHashAgainstBlock(appHash []byte, block *types.Block) error {
```
Liamsi: Why does the methiod return an error? The only thing it does, is panic in an error case and return nil if everything seems right.

consensus/replay.go
```diff
@@ -417,6 +417,12 @@ func (h *Handshaker) replayBlocks(state sm.State, proxyApp proxy.AppConns, appBl
 	for i := appBlockHeight + 1; i <= finalBlock; i++ {
 		h.logger.Info("Applying block", "height", i)
 		block := h.store.LoadBlock(i)
+
+		// Extra check to ensure the app was not changed in a way it shouldn't have.
+		if len(appHash) > 0 {
+			checkAppHashAgainstBlock(appHash, block)
```
Liamsi: Why is the error never checked here? And why does the method never return an error (only panic).

consensus/replay.go
```diff
@@ -456,9 +462,26 @@ func (h *Handshaker) replayBlock(state sm.State, height int64, proxyApp proxy.Ap
...
+		panic(fmt.Sprintf(`block.AppHash does not match AppHash after replay. Got %X, expected %X.
+
+Block: %v
+`,
+			appHash, block.AppHash, block))
+	}
+	return nil
+}
+
+func checkAppHashAgainstState(appHash []byte, state sm.State) error {
```
Liamsi: Same question: This will never return an error.

consensus/replay.go
melekes: Good catch. No need to return an error

consensus/replay.go
```diff
@@ -456,11 +465,26 @@ func (h *Handshaker) replayBlock(state sm.State, height int64, proxyApp proxy.Ap
 	return state, nil
 }
 
-func checkAppHash(state sm.State, appHash []byte) error {
-	if !bytes.Equal(state.AppHash, appHash) {
-		panic(fmt.Errorf("Tendermint state.AppHash does not match AppHash after replay. Got %X, expected %X", appHash, state.AppHash).Error())
+func checkAppHashAgainstBlock(appHash []byte, block *types.Block) {
```
Liamsi: The name is slightly misleading. This isn't a simple check anymore as we only panic. Don't know what is the best name here to signal this but I think `assertXXX`works.  E.g.
`assertAppMatchesBlockHash`
`assertAppMatchesStateHash`

You could even combine this in a single method:
`assertAppHashEq(appHash []byte, otherHash []byte, panicMsg string)`

melekes: Updated. Thanks!

melekes:
I will probably take a lot of time, but I'll try. You must understand that requiring to write a test for each PR will slow down dev. speed to 0 until codebase is easily testable (then it will be fast again)

melekes:
@Liamsi added a test case. PTAL

consensus/replay_test.go
```diff
@@ -499,7 +637,7 @@ func makeBlockchainFromWAL(wal WAL) ([]*types.Block, []*types.Commit, error) {
 		return nil, nil, err
 	}
 	if !found {
-		return nil, nil, fmt.Errorf("WAL does not contain height %d.", 1)
+		return nil, nil, fmt.Errorf("WAL does not contain height %d", 1)
```
melekes:
```suggestion
		return nil, nil, fmt.Errorf("WAL does not contain height %d", 0)
```

consensus/replay_test.go
```diff
@@ -489,6 +495,138 @@ func buildTMStateFromChain(config *cfg.Config, stateDB dbm.DB, state sm.State, c
...
+		defer proxyApp.Stop()
+
+		assert.Panics(t, func() {
+			h := NewHandshaker(stateDB, state, store, genDoc)
+			h.Handshake(proxyApp)
+		})
+	}
+}
+
+func makeBlocks(n int, state *sm.State, privVal types.PrivValidator) []*types.Block {
```
melekes: https://github.com/tendermint/tendermint/issues/3584

consensus/replay.go
```diff
@@ -393,8 +395,7 @@ func (h *Handshaker) ReplayBlocks(
 
 	}
 
-	cmn.PanicSanity("Should never happen")
-	return nil, nil
+	panic("should never happen")
```
melekes: explain why it should never happen?

consensus/replay_test.go
```diff
@@ -489,6 +495,138 @@ func buildTMStateFromChain(config *cfg.Config, stateDB dbm.DB, state sm.State, c
...
+	//		- 0x03
+	{
+		app := &badApp{numBlocks: 3, allHashesAreWrong: true}
+		clientCreator := proxy.NewLocalClientCreator(app)
+		proxyApp := proxy.NewAppConns(clientCreator)
+		err := proxyApp.Start()
+		require.NoError(t, err)
+		defer proxyApp.Stop()
+
+		assert.Panics(t, func() {
```
melekes: how to check panic msg using regexp?

Issue #3576, opened 2019-04-17
test: TestListenerTimeoutReadWrite non-deterministic test failure


Liamsi:
https://github.com/tendermint/tendermint/pull/3575#issuecomment-484149537
```
=== RUN   TestListenerTimeoutReadWrite
--- FAIL: TestListenerTimeoutReadWrite (0.01s)
    socket_listeners_test.go:110: read tcp 127.0.0.1:38914->127.0.0.1:60772: i/o timeout
```

Liamsi:
The cause probably is that the timeout isn't sufficient in rare cases:
https://github.com/tendermint/tendermint/blob/f2aa1bf50e3eb61f9cf27ee4e867fbd1daf9069f/privval/socket_listeners_test.go#L99-L113

Liamsi:
The reason here is, that in rare cases, the `Accept` times out faster than the dial happens. This is weird as the accept timeout is set to one full second (and the test only runs for 0.01s ...). 
https://github.com/tendermint/tendermint/blob/f2aa1bf50e3eb61f9cf27ee4e867fbd1daf9069f/privval/socket_listeners_test.go#L99-L110
Looking at the internals of the `Accept` it seems that during the accept we run into the timeoutReadWrite (which is only 1 ms):
https://github.com/tendermint/tendermint/blob/f2aa1bf50e3eb61f9cf27ee4e867fbd1daf9069f/privval/socket_listeners.go#L67-L80

melekes:
Hopefully fixed by https://github.com/tendermint/tendermint/pull/3580.

PR #3580, opened 2019-04-18
test: Increase timeout to mitigate non-deterministic test failure

Liamsi:
This should fix https://github.com/tendermint/tendermint/issues/3576 (ran it many times locally but only time will tell). The test actually only checked for the opcode of the error. From the name of the test we actually want to test if we see a timeout after a pre-defined time. 

* [ ] Wrote tests

### DIFF
--- a/privval/socket_listeners_test.go
+++ b/privval/socket_listeners_test.go
@@ -97,7 +97,15 @@ func TestListenerTimeoutAccept(t *testing.T) {
 }
 
 func TestListenerTimeoutReadWrite(t *testing.T) {
-	for _, tc := range listenerTestCases(t, time.Second, time.Millisecond) {
+	var (
+		// This needs to be long enough s.t. the Accept will definitely succeed:
+		timeoutAccept = time.Second
+		// This can be really short but in the TCP case, the accept can
+		// also trigger a timeoutReadWrite. Hence, we need to give it some time.
+		// Note: this controls how long this test actually runs.
+		timeoutReadWrite = 10 * time.Millisecond
+	)
+	for _, tc := range listenerTestCases(t, timeoutAccept, timeoutReadWrite) {
 		go func(dialer SocketDialer) {
 			_, err := dialer()
 			if err != nil {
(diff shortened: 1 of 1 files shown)


privval/socket_listeners_test.go
```diff
@@ -97,7 +97,15 @@ func TestListenerTimeoutAccept(t *testing.T) {
 }
 
 func TestListenerTimeoutReadWrite(t *testing.T) {
-	for _, tc := range listenerTestCases(t, time.Second, time.Millisecond) {
+	var (
```
melekes:
```suggestion
	const (
```

privval/socket_listeners_test.go
```diff
@@ -122,5 +129,9 @@ func TestListenerTimeoutReadWrite(t *testing.T) {
 		if have, want := opErr.Op, "read"; have != want {
 			t.Errorf("for %s listener, have %v, want %v", tc.description, have, want)
 		}
+
+		if have, want := opErr.Timeout(), true; have != want {
+			t.Errorf("for %s listener, got unexpected error: have %v, want %v", tc.description, have, want)
```
melekes:
```suggestion
			t.Errorf("for %s listener, got unexpected error: have %v, want timeout error", tc.description, opErr)
```

privval/socket_listeners_test.go
```diff
@@ -122,5 +129,9 @@ func TestListenerTimeoutReadWrite(t *testing.T) {
 		if have, want := opErr.Op, "read"; have != want {
 			t.Errorf("for %s listener, have %v, want %v", tc.description, have, want)
 		}
+
+		if have, want := opErr.Timeout(), true; have != want {
```
melekes:
```suggestion
		if !opErr.Timeout() {
```

Liamsi: I like to have this consistent with the cases above (would you also change to `if opErr.Op != "read"`)

melekes: Yes, I would

xla: The reason for this consistent form is that when you reformulate the expectations, you can't forget to change the error reprted. Less error prone and less editing.

melekes: `for %s listener, got unexpected error: have false want true`
in it's current form does not make sense for person running tests

melekes: @xla agree if you're talking about above case where we compare Op == read, but not here

xla: Agreed, it doesn't always produces the most readable output. A balance must be struck.

PR #3593, opened 2019-04-25
make persistent prop independent of conn direction

### LINKED ISSUE #3362: p2p: Seed node will disconnect from persistent peer if inbound
refer to #3304

For a seed node that want catch up with chain, it will only connected to a persistent peer,
but even a peer is a persistent peer of seed according to it's config file, in the runtime,
the peer may be inbounded peer(which is not persistent) or address pick from addressbook(also not persistent), witch cause the seed node disconnected to its persistent peer.

melekes:
Previously only outbound peers can be persistent. 

Now, even if the peer is inbound, if it's marked as persistent, when/if conn is lost,
Tendermint will try to reconnect. **This part is actually optional and can be reverted.**

Plus, seed won't disconnect from inbound peer if it's marked as
persistent. Fixes #3362

* [x] Wrote tests

### DIFF
--- a/docs/spec/p2p/config.md
+++ b/docs/spec/p2p/config.md
@@ -12,14 +12,14 @@ and upon incoming connection shares some peers and disconnects.
 
 ## Seeds
 
-`--p2p.seeds “1.2.3.4:26656,2.3.4.5:4444”`
+`--p2p.seeds “id100000000000000000000000000000000@1.2.3.4:26656,id200000000000000000000000000000000@2.3.4.5:4444”`
 
 Dials these seeds when we need more peers. They should return a list of peers and then disconnect.
 If we already have enough peers in the address book, we may never need to dial them.
 
 ## Persistent Peers
 
-`--p2p.persistent_peers “1.2.3.4:26656,2.3.4.5:26656”`
+`--p2p.persistent_peers “id100000000000000000000000000000000@1.2.3.4:26656,id200000000000000000000000000000000@2.3.4.5:26656”`
 
 Dial these peers and auto-redial them if the connection fails.
 These are intended to be trusted persistent peers that can help
@@ -30,9 +30,9 @@ backoff and will give up after a day of trying to connect.
 the user will be warned that seeds may auto-close connections
 and that the node may not be able to keep the connection persistent.
 
-## Private Persistent Peers
+## Private Peers
 
-`--p2p.private_persistent_peers “1.2.3.4:26656,2.3.4.5:26656”`
+`--p2p.private_peer_ids “id100000000000000000000000000000000,id200000000000000000000000000000000”`
 
-These are persistent peers that we do not add to the address book or
-gossip to other peers. They stay private to us.
+These are IDs of the peers that we do not add to the address book or gossip to
+other peers. They stay private to us.
--- a/node/node.go
+++ b/node/node.go
@@ -471,6 +471,11 @@ func NewNode(config *cfg.Config,
 	sw.AddReactor("EVIDENCE", evidenceReactor)
 	sw.SetNodeInfo(nodeInfo)
 	sw.SetNodeKey(nodeKey)
+	errs := sw.AddPersistentPeers(splitAndTrimEmpty(config.P2P.PersistentPeers, ",", " "))
+	if len(errs) > 0 {
+		// do not bother with returning all errors at once
+		return nil, errors.Wrap(errs[0], "error in persistent_peers")
+	}
 
 	p2pLogger.Info("P2P Node ID", "ID", nodeKey.ID(), "file", config.NodeKeyFile())
 
@@ -591,12 +596,8 @@ func (n *Node) OnStart() error {
 	}
 
 	// Always connect to persistent peers
-	if n.config.P2P.PersistentPeers != "" {
-		err = n.sw.DialPeersAsync(n.addrBook, splitAndTrimEmpty(n.config.P2P.PersistentPeers, ",", " "), true)
-		if err != nil {
-			return err
-		}
-	}
+	// parsing errors are handled above by AddPersistentPeers
+	_ = n.sw.DialPeersAsync(splitAndTrimEmpty(n.config.P2P.PersistentPeers, ",", " "))
 
 	return nil
 }
--- a/node/node.go
+++ b/node/node.go
@@ -473,7 +473,7 @@ func NewNode(config *cfg.Config,
 	sw.SetNodeKey(nodeKey)
 	err = sw.AddPersistentPeers(splitAndTrimEmpty(config.P2P.PersistentPeers, ",", " "))
 	if err != nil {
-		return nil, errors.Wrap(err, "error in persistent_peers")
+		return nil, errors.Wrap(err, "could not add peers from persistent_peers config")
 	}
 
 	p2pLogger.Info("P2P Node ID", "ID", nodeKey.ID(), "file", config.NodeKeyFile())
--- a/node/node.go
+++ b/node/node.go
@@ -425,10 +425,6 @@ func createSwitch(config *cfg.Config,
 	sw.AddReactor("EVIDENCE", evidenceReactor)
 	sw.SetNodeInfo(nodeInfo)
 	sw.SetNodeKey(nodeKey)
-	err = sw.AddPersistentPeers(splitAndTrimEmpty(config.P2P.PersistentPeers, ",", " "))
-	if err != nil {
-		return nil, errors.Wrap(err, "could not add peers from persistent_peers config")
-	}
 
 	p2pLogger.Info("P2P Node ID", "ID", nodeKey.ID(), "file", config.NodeKeyFile())
 	return sw
@@ -582,6 +578,11 @@ func NewNode(config *cfg.Config,
 		config, transport, p2pMetrics, peerFilters, mempoolReactor, bcReactor,
 		consensusReactor, evidenceReactor, nodeInfo, nodeKey, p2pLogger,
 	)
+
+	err = sw.AddPersistentPeers(splitAndTrimEmpty(config.P2P.PersistentPeers, ",", " "))
+	if err != nil {
+		return nil, errors.Wrap(err, "could not add peers from persistent_peers field")
+	}
 
 	addrBook := createAddrBookAndSetOnSwitch(config, sw, p2pLogger)
 
--- a/node/node.go
+++ b/node/node.go
@@ -471,10 +471,9 @@ func NewNode(config *cfg.Config,
 	sw.AddReactor("EVIDENCE", evidenceReactor)
 	sw.SetNodeInfo(nodeInfo)
 	sw.SetNodeKey(nodeKey)
-	errs := sw.AddPersistentPeers(splitAndTrimEmpty(config.P2P.PersistentPeers, ",", " "))
-	if len(errs) > 0 {
-		// do not bother with returning all errors at once
-		return nil, errors.Wrap(errs[0], "error in persistent_peers")
+	err = sw.AddPersistentPeers(splitAndTrimEmpty(config.P2P.PersistentPeers, ",", " "))
+	if err != nil {
+		return nil, errors.Wrap(err, "error in persistent_peers")
 	}
 
 	p2pLogger.Info("P2P Node ID", "ID", nodeKey.ID(), "file", config.NodeKeyFile())
--- a/p2p/pex/pex_reactor_test.go
+++ b/p2p/pex/pex_reactor_test.go
@@ -398,7 +398,7 @@ func TestPEXReactorSeedModeFlushStop(t *testing.T) {
 	reactor := switches[0].Reactors()["pex"].(*PEXReactor)
 	peerID := switches[1].NodeInfo().ID()
 
-	err = switches[1].DialPeerWithAddress(switches[0].NetAddress(), false)
+	err = switches[1].DialPeerWithAddress(switches[0].NetAddress())
 	assert.NoError(t, err)
 
 	// sleep up to a second while waiting for the peer to send us a message.
--- a/p2p/pex/pex_reactor_test.go
+++ b/p2p/pex/pex_reactor_test.go
@@ -322,6 +322,42 @@ func TestPEXReactorSeedMode(t *testing.T) {
 	assert.Equal(t, 0, sw.Peers().Size())
 }
 
+func TestPEXReactorDoesNotDisconnectFromPersistentPeerInSeedMode(t *testing.T) {
+	// directory to store address books
+	dir, err := ioutil.TempDir("", "pex_reactor")
+	require.Nil(t, err)
+	defer os.RemoveAll(dir) // nolint: errcheck
+
+	pexR, book := createReactor(&PEXReactorConfig{SeedMode: true, SeedDisconnectWaitPeriod: 1 * time.Millisecond})
+	defer teardownReactor(book)
+
+	sw := createSwitchAndAddReactors(pexR)
+	sw.SetAddrBook(book)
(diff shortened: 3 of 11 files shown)


p2p/transport.go
```diff
@@ -37,11 +37,15 @@ type accept struct {
 // events.
 // TODO(xla): Refactor out with more static Reactor setup and PeerBehaviour.
 type peerConfig struct {
-	chDescs              []*conn.ChannelDescriptor
-	onPeerError          func(Peer, interface{})
-	outbound, persistent bool
-	reactorsByCh         map[byte]Reactor
-	metrics              *Metrics
+	chDescs      []*conn.ChannelDescriptor
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

rpc/core/pipe.go
```diff
@@ -44,7 +44,8 @@ type transport interface {
 }
 
 type peers interface {
-	DialPeersAsync(p2p.AddrBook, []string, bool) error
+	AddPersistentPeers([]string) []error
```
Liamsi: Everywhere this is used, only the first error `errs[0]` is considered. Why return a list of errors then?

rpc/core/pipe.go
```diff
@@ -44,7 +44,8 @@ type transport interface {
...
+	DialPeersAsync([]string) []error
```
Liamsi: Similarly, `DialPeersAsync`: if the errors are not ignored (which is well explained tho), only the first error is considered.

p2p/switch.go
```diff
@@ -401,28 +416,37 @@ func isPrivateAddr(err error) bool {
 	return ok && te.PrivateAddr()
 }
 
-// DialPeersAsync dials a list of peers asynchronously in random order (optionally, making them persistent).
+// DialPeersAsync dials a list of peers asynchronously in random order.
 // Used to dial peers from config on startup or from unsafe-RPC (trusted sources).
-// TODO: remove addrBook arg since it's now set on the switch
-func (sw *Switch) DialPeersAsync(addrBook AddrBook, peers []string, persistent bool) error {
+// ErrNetAddressLookup errors are ignored. However, if there are other errors,
+// all of them are returned.
```
Liamsi: What for?

rpc/core/pipe.go
melekes: Because someone (external users of our p2p lib) could need all the errors?

p2p/switch.go
melekes: To handle them?

rpc/core/pipe.go
Liamsi: Not sure if there are users of the p2p library. But it is likely that they will use it as it we use it internally too: probably, they will only check "was there an error" (and use the first).

If it is important that the interface method takes several addrs, then only one `error`(the first occurring?) should be retuned. This reflects the current implementation. If you think people must have the possibility to treat the errors separately, why not have a method that takes one addr string and returns one error instead?

melekes: > probably, they will only check "was there an error"

I wound not make such statements.

> why not have a method that takes one addr string and returns one error instead?

being able to add multiple addresses is more convenient, I guess.

Liamsi: > I wound not make such statements.

Sorry, you are right. It's far too vague and sounds like wild guessing. What I'm trying to say, is that in golang it is very idiomatic that a method returns one error if sth. went wrong. Ideally, in some cases it returns a set of known errors users can switch over and handle in a particular way. Returning a list of errors is less idiomatic. Hence, it isn't unlikely that a user will treat the list of errors like "did an error occur" (in fact your code even does that).

> being able to add multiple addresses is more convenient, I guess.

Absolutely! But it obfuscates the connection between addr <-> potentially occurred error. Letting the caller use a for-loop and range over the addresses (and deal with those errors), leaves no doubt about this connection.

melekes: > What I'm trying to say, is that in golang it is very idiomatic that a method returns one error

good argument

p2p/switch.go
```diff
@@ -583,11 +641,10 @@ func (sw *Switch) acceptRoutine() {
 // add the peer.
 // if dialing fails, start the reconnect loop. If handhsake fails, its over.
 // If peer is started succesffuly, reconnectLoop will start when
```
Liamsi:
```suggestion
// If peer is started successfully, reconnectLoop will start when
```

p2p/switch.go
```diff
@@ -583,11 +641,10 @@ func (sw *Switch) acceptRoutine() {
 // add the peer.
 // if dialing fails, start the reconnect loop. If handhsake fails, its over.
```
Liamsi: While at it:
```suggestion
// if dialing fails, start the reconnect loop. If handshake fails, it's over.
```
or: it stops instead of the dramatic "it's over"

p2p/pex/pex_reactor_test.go
```diff
@@ -322,6 +322,42 @@ func TestPEXReactorSeedMode(t *testing.T) {
...
+
+	err = sw.AddPersistentPeers([]string{peerSwitch.NetAddress().String()})
+	require.NoError(t, err)
+
+	// 1. Test crawlPeers dials the peer
+	pexR.crawlPeers([]*p2p.NetAddress{peerSwitch.NetAddress()})
+	assert.Equal(t, 1, sw.Peers().Size())
+	assert.True(t, sw.Peers().Has(peerSwitch.NodeInfo().ID()))
+
+	time.Sleep(10 * time.Millisecond)
```
Liamsi: Can you add a comment why we need to sleep 10 ms. I'm worried that we might have introduced a new non-deterministic test-failure 🙈

node/node.go
```diff
@@ -471,6 +471,10 @@ func NewNode(config *cfg.Config,
 	sw.AddReactor("EVIDENCE", evidenceReactor)
 	sw.SetNodeInfo(nodeInfo)
 	sw.SetNodeKey(nodeKey)
+	err = sw.AddPersistentPeers(splitAndTrimEmpty(config.P2P.PersistentPeers, ",", " "))
+	if err != nil {
+		return nil, errors.Wrap(err, "error in persistent_peers")
```
Liamsi:
```suggestion
		return nil, errors.Wrap(err, "could not add peers from persistent_peers config")
```
`err` should contain what was wrong (e.g. seperator not found etc)

docs/spec/p2p/config.md
```diff
@@ -30,9 +30,9 @@ backoff and will give up after a day of trying to connect.
 the user will be warned that seeds may auto-close connections
 and that the node may not be able to keep the connection persistent.
 
-## Private Persistent Peers
+## Private Peers
```
Liamsi: Aren't private peers still persistent?

melekes: Any peer can be private, no?)

melekes:
> Any reasons why this behaviour could be undesired?

No.

malicious node could report different address (say address our node has in `persistent_peers`), but this won't give it any advantages.

node/node.go
```diff
@@ -425,6 +425,10 @@ func createSwitch(config *cfg.Config,
 	sw.AddReactor("EVIDENCE", evidenceReactor)
 	sw.SetNodeInfo(nodeInfo)
 	sw.SetNodeKey(nodeKey)
+	err = sw.AddPersistentPeers(splitAndTrimEmpty(config.P2P.PersistentPeers, ",", " "))
```
golangcibot: undeclared name: `err` (from `typecheck`)

node/node.go
```diff
@@ -425,6 +425,10 @@ func createSwitch(config *cfg.Config,
...
+	if err != nil {
+		return nil, errors.Wrap(err, "could not add peers from persistent_peers config")
```
golangcibot: wrong number of return values (want 1, got 2) (from `typecheck`)

p2p/pex/pex_reactor_test.go
```diff
@@ -315,13 +316,48 @@ func TestPEXReactorSeedMode(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// 1. Test crawlPeers dials the peer
+	pexR.crawlPeers([]*p2p.NetAddress{peerSwitch.NetAddress()})
+	assert.Equal(t, 1, sw.Peers().Size())
+	assert.True(t, sw.Peers().Has(peerSwitch.NodeInfo().ID()))
+
+	// 2. attemptDisconnects should not disconnect because the peer is persistent
+	pexR.attemptDisconnects()
+	assert.Equal(t, 1, sw.Peers().Size())
```
ebuchman: This passes even if we comment out the AddPersistenPeers line. I think the `attemptDisconnects` isn't causing a disconnect because of the timing constraints. If we sleep before it, then it fails without the AddPersistentPeers, but maybe there's a better way ?

p2p/switch.go
```diff
@@ -495,13 +520,46 @@ func (sw *Switch) IsDialingOrExistingAddress(addr *NetAddress) bool {
...
+	sw.Logger.Info("Adding persistent peers", "addrs", addrs)
+	netAddrs, errs := NewNetAddressStrings(addrs)
+	// report all the errors
+	for _, err := range errs {
+		sw.Logger.Error("Error in peer's address", "err", err)
+	}
+	// return first non-ErrNetAddressLookup error
+	for _, err := range errs {
+		if _, ok := err.(ErrNetAddressLookup); ok {
+			continue
```
ebuchman: Doesn't this mean we might include `nil` in the list? What is the value of ignoring this error here if we don't even get a netAddr out of it?

p2p/switch.go
```diff
@@ -495,13 +520,46 @@ func (sw *Switch) IsDialingOrExistingAddress(addr *NetAddress) bool {
...
+		}
+		return err
+	}
+	sw.persistentPeersAddrs = netAddrs
+	return nil
+}
+
+func (sw *Switch) isPeerPersistentFn() func(*NetAddress) bool {
```
ebuchman: I wonder if this should just check the ID instead. Since for instance if there's DNS involved the IP address could change between when it got added to persistent peers and when we connected to it and then this would fail.

ebuchman: Also why is this a closure?

p2p/switch_test.go
```diff
@@ -390,49 +394,33 @@ func TestSwitchStopPeerForError(t *testing.T) {
...
-	require.Nil(err)
-
-	require.Nil(sw.addPeer(p))
+	err = sw.AddPersistentPeers([]string{rp.Addr().String()})
+	require.NoError(t, err)
 
-	require.NotNil(sw.Peers().Get(rp.ID()))
+	err = sw.DialPeerWithAddress(rp.Addr())
+	require.Nil(t, err)
+	time.Sleep(50 * time.Millisecond)
```
ebuchman: Can we do better than a sleep here?

p2p/switch_test.go
```diff
@@ -443,23 +431,68 @@ func TestSwitchReconnectsToPersistentPeer(t *testing.T) {
 	rp.Start()
 	defer rp.Stop()
 
-	// simulate first time dial failure
 	conf := config.DefaultP2PConfig()
 	conf.TestDialFail = true
-	err = sw.addOutboundPeerWithConfig(rp.Addr(), conf, true)
-	require.NotNil(err)
-
+	err = sw.addOutboundPeerWithConfig(rp.Addr(), conf)
```
ebuchman: don't we need to make this peer persistent to preserve the test case?

ebuchman:
> malicious node could report different address (say address our node has in persistent_peers), but this won't give it any advantages.

I think we can address this by just using IDs for persistent peers.

node/node.go
```diff
@@ -675,12 +680,8 @@ func (n *Node) OnStart() error {
...
 
 	// Always connect to persistent peers
-	if n.config.P2P.PersistentPeers != "" {
-		err = n.sw.DialPeersAsync(n.addrBook, splitAndTrimEmpty(n.config.P2P.PersistentPeers, ",", " "), true)
-		if err != nil {
-			return err
-		}
-	}
+	// parsing errors are handled above by AddPersistentPeers
+	_ = n.sw.DialPeersAsync(splitAndTrimEmpty(n.config.P2P.PersistentPeers, ",", " "))
```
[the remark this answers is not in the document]
melekes: > We should still handle this anyways for completeness, no?

ok

> Also is there a way we can avoid having to use the splitAndTrimEmpty again?

A bit annoying, but I think we can live with it for now.

p2p/switch.go
```diff
@@ -401,28 +416,41 @@ func isPrivateAddr(err error) bool {
 	return ok && te.PrivateAddr()
 }
 
-// DialPeersAsync dials a list of peers asynchronously in random order (optionally, making them persistent).
+// DialPeersAsync dials a list of peers asynchronously in random order.
 // Used to dial peers from config on startup or from unsafe-RPC (trusted sources).
-// TODO: remove addrBook arg since it's now set on the switch
-func (sw *Switch) DialPeersAsync(addrBook AddrBook, peers []string, persistent bool) error {
+// It ignores ErrNetAddressLookup. However, if there are other errors, first
```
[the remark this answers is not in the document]
melekes: Imagine you're starting a k8s cluster and DNS does not work yet => lookup fails => Tendermint fails to start.

If we ignore ErrNetAddressLookup, we won't be able to dial peers immediately, but reconnect will connect us to them eventually (only to persistent peers though).

Maybe we should add a flag `ignoreNetAddressLookupErrors`

p2p/switch.go
melekes: Good catch 👍 I'll need to think about it. See my other comment https://github.com/tendermint/tendermint/pull/3593#discussion_r280971995

p2p/switch.go
melekes: > I wonder if this should just check the ID instead. Since for instance if there's DNS involved the IP address could change between when it got added to persistent peers and when we connected to it and then this would fail.

This deserves a separate issue. One argument is favor of IP addresses is that it's more static. Maybe there are some DNS attacks we're not considering.

p2p/switch.go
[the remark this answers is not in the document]
melekes: > but reconnect will connect us to them eventually (only to persistent peers though).

this won't work in the current implementation because addr is `nil`. hmmm

p2p/switch.go
melekes: > Also why is this a closure?

I thought it would be weird to pass persistent_peers list to `peerConfig`.

p2p/switch_test.go
melekes: No. Because we'll reconnect anyway if `TestDialFail` is set
https://github.com/tendermint/tendermint/blob/5051a1f7bcda164835592baccee3caf0d02984ec/p2p/switch.go#L653

p2p/switch_test.go
melekes: removed it

p2p/switch.go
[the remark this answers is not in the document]
melekes: https://github.com/tendermint/tendermint/issues/880

p2p/switch.go
melekes: > Doesn't this mean we might include nil in the list?

no https://github.com/tendermint/tendermint/blob/develop/p2p/netaddress.go#L134-L138

p2p/switch.go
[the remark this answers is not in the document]
melekes: I think NetAddress should resolve IP right before dialing + resolve it again when IP is no longer reachable.

p2p/switch.go
ebuchman: https://github.com/tendermint/tendermint/issues/3692

PR #3603, opened 2019-04-29
pex: dial seeds when address book needs more addresses

defunctzombie:
If we are low on addresses for peering, we need to discover more peers. The
previous behavior would query existing peers; however, if an existing peer
does not participate in peer exchange, then our node will not discover more peers.

This change consults both existing peers as well as seeds when there is a deficit
in address book addresses. This allows for discovering peers though existing channels
as well as via seeds if existing peers do not share addresses.

---

* [ ] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -19,3 +19,5 @@
 ### IMPROVEMENTS:
 
 ### BUG FIXES:
+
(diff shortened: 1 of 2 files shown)


defunctzombie:
I'd like to settle on direction for the code changes before tackling the remaining TODO items from the PR checklist.

jackzampolin:
This looks like a great change!

melekes:

PR #3604, opened 2019-04-30
add boltdb for storage

CrocdileChan:
add  boltdb  implement for db interface.

The boltdb is safe and high performence embedded KV database.  It is based on B+tree and Mmap.
Now it is maintained by etcd-io org. https://github.com/etcd-io/bbolt
It is much better than LSM tree (levelDB)  when RANDOM and SCAN read.

### DIFF
--- a/libs/db/boltdb.go
+++ b/libs/db/boltdb.go
@@ -2,31 +2,32 @@ package db
 
 import (
 	"bytes"
+	"os"
 	"path/filepath"
 	"sync"
 
 	"github.com/etcd-io/bbolt"
 )
 
-var bucket = []byte("boltdb_bucket")
+var bucket = []byte("tm")
 
 func init() {
 	registerDBCreator(BoltDBBackend, func(name string, dir string) (DB, error) {
-		return NewBoltdb(name, dir)
+		return NewBoltDB(name, dir)
 	}, false)
 }
 
-type Boltdb struct {
+type BoltDB struct {
 	db *bbolt.DB
 }
 
-func NewBoltdb(name, dir string) (DB, error) {
-	return NewBoltdbWithOpts(name, dir, bbolt.DefaultOptions)
+func NewBoltDB(name, dir string) (DB, error) {
+	return NewBoltDBWithOpts(name, dir, bbolt.DefaultOptions)
 }
 
-func NewBoltdbWithOpts(name, dir string, opts *bbolt.Options) (DB, error) {
+func NewBoltDBWithOpts(name string, dir string, opts *bbolt.Options) (DB, error) {
 	dbPath := filepath.Join(dir, name+".db")
-	db, err := bbolt.Open(dbPath, 0600, opts)
+	db, err := bbolt.Open(dbPath, os.ModePerm, opts)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


libs/db/boltdb_test.go
```diff
@@ -0,0 +1,95 @@
...
+	// Open the db twice in a row to test read-only locks
+	ro1, err := NewBoltdbWithOpts(name, "", &bbolt.Options{ReadOnly: true})
+	defer ro1.Close()
+	require.Nil(t, err)
+	ro2, err := NewBoltdbWithOpts(name, "", &bbolt.Options{ReadOnly: true})
+	defer ro2.Close()
+	require.Nil(t, err)
+}
+
+func BenchmarkBoltdbRandomReadsWrites(b *testing.B) {
```
golangcibot: 34-95 lines are duplicate of `libs/db/go_level_db_test.go:35-96` (from `dupl`)

libs/db/boltdb.go
```diff
@@ -0,0 +1,245 @@
+package db
+
+import (
+	"bytes"
+	"github.com/etcd-io/bbolt"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

libs/db/boltdb.go
```diff
@@ -0,0 +1,253 @@
+package db
```
Liamsi: Not blocking this PR but shouldn't we create a package per implementation (bolt, golevel, mem, etc)?

libs/db/boltdb.go
```diff
@@ -0,0 +1,253 @@
...
+func (bdb *Boltdb) DeleteSync(key []byte) {
+	bdb.Delete(key)
+}
+
+func (bdb *Boltdb) Close() {
+	bdb.db.Close()
+}
+
+func (bdb *Boltdb) Print() {
+	panic("boltdb.print not yet implemented")
```
Liamsi: If we merge this, we should open follow-up issues for these.

libs/db/boltdb_test.go
Liamsi: We can probably refactor the tests to benchmark them via t.Run to remove code duplication. Can happen in a follow-up PR too

libs/db/boltdb.go
melekes: Yes to ^. We should also use conditional compilation to only include deps & code when X db is requested (`// +build boltdb`).

CrocdileChan:
I've requested a PR yesterday `490981f`.
Please review it :-p    
Can it be merged?

Thanks  a lot :-p

melekes:
> Can it be merged?

Of course! But before we do that could you please address the above comments.

melekes:
> If you want me to modify the benchmark case, please teach me how to do it.

No problem. I can work on this.

melekes:
Closing in favor of https://github.com/tendermint/tendermint/pull/3610

libs/db/boltdb.go
Liamsi: I meant that if these were not implemented here, we should track these in an issue and do implement these functions in a follow-up PR.

libs/db/boltdb.go
melekes: https://github.com/tendermint/tendermint/issues/3611

PR #3606, opened 2019-04-30
cli: Add option to not clear address book with unsafe reset

climber73:
Adds `--keep-addr-book` flag to `unsafe_reset_all` command allowing not to clear address book


climber73:
Not sure, maybe it's better to split the definitions of ResetAllCmd and ResetPrivValidatorCmd

melekes:

Liamsi:
Awesome! Thanks ❤️ Even if the code is self-explanatory: Could you add a short description on what was changed and why in the PR description?

Issue #3531, opened 2019-04-04
p2p: Session should terminate on nonce wrapping


zebambam:
Hi!

The nonce that peers use to communicate in secret sessions is 64 bit (rather than 96 bit) and looks like it might wrap?

``` go
func incrNonce(nonce *[aeadNonceSize]byte) {
	counter := binary.LittleEndian.Uint64(nonce[4:])
	counter++
	binary.LittleEndian.PutUint64(nonce[4:], counter)
}
```

Seems like the session should terminate before re-using a nonce.

Doesn't look like there's a terribly practical attack available but may as well belt-and-braces since it's crypto that's important.

Thanks

Liamsi:
Thanks for reporting this. Yes, you are right it will wrap (after the 4 bytes are full). We are moving away from our implementation of the station to station protocol (secret connection). Hence, I'm not sure it is worth the effort. 
FWIW, the easiest would be to panic before we overflow (that is what we did in the Rust implementation: https://github.com/tendermint/kms/blob/752184b5bfa139cde4c9013d836e0f3cda897110/tendermint-rs/src/secret_connection/nonce.rs#L68). This would end the session and makes sure the nonce would not be re-used.

melekes:
Merged https://github.com/tendermint/tendermint/pull/3609 to develop.

zebambam:
Good stuff, thanks!

PR #3609, opened 2019-05-01
p2p: Session should terminate on nonce wrapping (#3531)

climber73:
Adds checking if nonce would overflow by next increasing. Panics if true

Refs #3531

### DIFF
--- a/p2p/conn/secret_connection.go
+++ b/p2p/conn/secret_connection.go
@@ -8,6 +8,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"io"
+	"math"
 	"net"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


p2p/conn/secret_connection.go
```diff
@@ -439,6 +440,10 @@ func shareAuthSignature(sc *SecretConnection, pubKey crypto.PubKey, signature []
 // (little-endian in nonce[4:]).
 func incrNonce(nonce *[aeadNonceSize]byte) {
 	counter := binary.LittleEndian.Uint64(nonce[4:])
+	if counter == math.MaxUint64 {
+		// to end the session and makes sure the nonce would not be re-used:
```
```suggestion
		// Terminates the session and makes sure the nonce would not re-used.
		// See https://github.com/tendermint/tendermint/issues/3531
```

climber73: 👍

Liamsi:
@climber73 do you mind adding a changelog_pending entry?

Issue #1835, opened 2018-06-29
Tendermint shows time dependent performance degredation


z-bowen:
**BUG REPORT**

**Tendermint version**: 0.20.0-27bd1dea

**Environment**:
- **OS**: Ubuntu 18.04 LTS (Bionic Beaver)
- **Processor**: Intel(R) Xeon(R) CPU E5-2650 v3 @ 2.30GHz (10 cores)
- **Memory**: 15GB

**What happened**:
The tm-bench tool shows a performance degradation over time. Initially, I can get performance on the order of 2000tx/s, but this drops off rapidly as the test duration increases. After 1000s, I see transaction rates of only ~200tx/s, and this continues to fall as the test duration increases.

**What you expected to happen**:
Tendermint transaction rates should be consistent, regardless of test duration.

**How to reproduce it**:
A detailed report of the study can be found [here](https://github.com/bigchaindb/BEPs/tree/ecdf09868be7b8c922f893aa09e1acdd1e0d5912/19)

This report contains scripts to reproduce the problem, as well as the config.toml from the test runs.

**Anything else do we need to know**:
I'll be out of the office and unable to check emails for the next week, but I'll be happy to communicate with your team about this issue when I'm back on 9.7.18

ValarDragon:
We've recently created patches to tm-bench to fix this. There was an issue with the tm-bench code. (Basically stats were only being collected from the last 20 blocks, not over all blocks in that duration) Thanks for making the report, its awesome!

The graphs you made look roughly like a `1/x` curve which is what we'd expect since tm-bench was only reading the last 20 blocks

melekes:
> Any pauses needed to maintain throughput on the Tendermint end should be invisible to the user, as the memory buffer stays available at all times.

given you are limiting the number of requests per IP, right? Otherwise, any DDOS attack will lead to OOM (out of memory) errors.

melekes:
I run the scripts on my dev. machine and can confirm there is a slight decline in performance.

Tendermint: https://github.com/tendermint/tendermint/tree/release/v0.22.0
tm-bench: https://github.com/tendermint/tools/commit/db419a308e6618eb52d0df6193fbbce482417311

![screen shot 2018-07-02 at 14 52 41](https://user-images.githubusercontent.com/1282182/42160320-d41a6a96-7e07-11e8-87a1-3a62e45bb428.png)

red - std.dev
yellow - max

melekes:
Not immediately apparent of why this is happening.

[...]

PR #3610, opened 2019-05-01
bbolt (etcd's fork of bolt)

melekes:
Original description:

> add boltdb implement for db interface.
> 
> The boltdb is safe and high performence embedded KV database. It is based on B+tree and Mmap.
> Now it is maintained by etcd-io org. https://github.com/etcd-io/bbolt
> It is much better than LSM tree (levelDB) when RANDOM and SCAN read.

Replaced https://github.com/tendermint/tendermint/pull/3604
Refs https://github.com/tendermint/tendermint/issues/1835

* [ ] Wrote tests

### DIFF
--- a/libs/db/boltdb.go
+++ b/libs/db/boltdb.go
@@ -180,7 +180,8 @@ func (bdb *boltDBBatch) WriteSync() {
 
 func (bdb *boltDBBatch) Close() {}
 
-// Any concurrent writes (Set, SetSync) will block until the Iterator is closed.
+// WARNING: Any concurrent writes (Set, SetSync) will block until the Iterator
+// is closed.
 func (bdb *BoltDB) Iterator(start, end []byte) Iterator {
 	tx, err := bdb.db.Begin(false)
 	if err != nil {
@@ -190,7 +191,8 @@ func (bdb *BoltDB) Iterator(start, end []byte) Iterator {
 	return newBoltDBIterator(c, start, end, false)
 }
 
-// Any concurrent writes (Set, SetSync) will block until the Iterator is closed.
+// WARNING: Any concurrent writes (Set, SetSync) will block until the Iterator
+// is closed.
 func (bdb *BoltDB) ReverseIterator(start, end []byte) Iterator {
 	tx, err := bdb.db.Begin(false)
 	if err != nil {
--- a/libs/db/boltdb.go
+++ b/libs/db/boltdb.go
@@ -2,31 +2,32 @@ package db
 
 import (
 	"bytes"
+	"os"
 	"path/filepath"
 	"sync"
 
 	"github.com/etcd-io/bbolt"
 )
 
-var bucket = []byte("boltdb_bucket")
+var bucket = []byte("tm")
 
 func init() {
 	registerDBCreator(BoltDBBackend, func(name string, dir string) (DB, error) {
-		return NewBoltdb(name, dir)
+		return NewBoltDB(name, dir)
 	}, false)
 }
 
-type Boltdb struct {
+type BoltDB struct {
 	db *bbolt.DB
 }
 
-func NewBoltdb(name, dir string) (DB, error) {
-	return NewBoltdbWithOpts(name, dir, bbolt.DefaultOptions)
+func NewBoltDB(name, dir string) (DB, error) {
+	return NewBoltDBWithOpts(name, dir, bbolt.DefaultOptions)
 }
 
-func NewBoltdbWithOpts(name, dir string, opts *bbolt.Options) (DB, error) {
+func NewBoltDBWithOpts(name string, dir string, opts *bbolt.Options) (DB, error) {
 	dbPath := filepath.Join(dir, name+".db")
-	db, err := bbolt.Open(dbPath, 0600, opts)
+	db, err := bbolt.Open(dbPath, os.ModePerm, opts)
 	if err != nil {
 		return nil, err
 	}
@@ -37,10 +38,10 @@ func NewBoltdbWithOpts(name, dir string, opts *bbolt.Options) (DB, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &Boltdb{db: db}, nil
+	return &BoltDB{db: db}, nil
 }
 
-func (bdb *Boltdb) Get(key []byte) (value []byte) {
+func (bdb *BoltDB) Get(key []byte) (value []byte) {
 	key = nonNilBytes(key)
 	err := bdb.db.View(func(tx *bbolt.Tx) error {
 		b := tx.Bucket(bucket)
@@ -53,11 +54,11 @@ func (bdb *Boltdb) Get(key []byte) (value []byte) {
 	return
 }
 
-func (bdb *Boltdb) Has(key []byte) bool {
+func (bdb *BoltDB) Has(key []byte) bool {
 	return bdb.Get(key) != nil
 }
 
-func (bdb *Boltdb) Set(key, value []byte) {
+func (bdb *BoltDB) Set(key, value []byte) {
 	key = nonNilBytes(key)
 	value = nonNilBytes(value)
 	err := bdb.db.Update(func(tx *bbolt.Tx) error {
@@ -69,11 +70,11 @@ func (bdb *Boltdb) Set(key, value []byte) {
 	}
 }
 
-func (bdb *Boltdb) SetSync(key, value []byte) {
+func (bdb *BoltDB) SetSync(key, value []byte) {
 	bdb.Set(key, value)
 }
 
-func (bdb *Boltdb) Delete(key []byte) {
+func (bdb *BoltDB) Delete(key []byte) {
 	key = nonNilBytes(key)
 	err := bdb.db.Update(func(tx *bbolt.Tx) error {
 		return tx.Bucket(bucket).Delete(key)
@@ -83,28 +84,28 @@ func (bdb *Boltdb) Delete(key []byte) {
 	}
 }
 
-func (bdb *Boltdb) DeleteSync(key []byte) {
+func (bdb *BoltDB) DeleteSync(key []byte) {
 	bdb.Delete(key)
 }
 
-func (bdb *Boltdb) Close() {
+func (bdb *BoltDB) Close() {
 	bdb.db.Close()
 }
 
-func (bdb *Boltdb) Print() {
+func (bdb *BoltDB) Print() {
 	panic("boltdb.print not yet implemented")
 }
 
-func (bdb *Boltdb) Stats() map[string]string {
+func (bdb *BoltDB) Stats() map[string]string {
 	panic("boltdb.stats not yet implemented")
 }
 
 type BoltdbBatch struct {
 	buffer *sync.Map
-	db     *Boltdb
+	db     *BoltDB
 }
 
-func (bdb *Boltdb) NewBatch() Batch {
+func (bdb *BoltDB) NewBatch() Batch {
 	return &BoltdbBatch{
 		buffer: &sync.Map{},
 		db:     bdb,
@@ -139,7 +140,7 @@ func (bdb *BoltdbBatch) WriteSync() {
 
 func (bdb *BoltdbBatch) Close() {}
 
-func (bdb *Boltdb) Iterator(start, end []byte) Iterator {
+func (bdb *BoltDB) Iterator(start, end []byte) Iterator {
 	tx, err := bdb.db.Begin(false)
 	if err != nil {
 		panic(err)
@@ -148,7 +149,7 @@ func (bdb *Boltdb) Iterator(start, end []byte) Iterator {
 	return newBoltdbIterator(c, start, end, false)
 }
 
-func (bdb *Boltdb) ReverseIterator(start, end []byte) Iterator {
+func (bdb *BoltDB) ReverseIterator(start, end []byte) Iterator {
 	tx, err := bdb.db.Begin(false)
 	if err != nil {
 		panic(err)
--- a/libs/db/boltdb.go
+++ b/libs/db/boltdb.go
@@ -1,0 +1,245 @@
+package db
+
+import (
+	"bytes"
+	"github.com/etcd-io/bbolt"
+	"path/filepath"
+	"sync"
+)
+
+var bucket = []byte("boltdb_bucket")
+
+func init() {
+	registerDBCreator(BoltDBBackend, func(name string, dir string) (DB, error) {
+		return NewBoltdb(name, dir)
+	}, false)
+}
+
+type Boltdb struct {
+	db *bbolt.DB
+}
+
+func NewBoltdb(name, dir string) (DB, error) {
+	return NewBoltdbWithOpts(name, dir, bbolt.DefaultOptions)
+}
+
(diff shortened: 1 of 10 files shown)


libs/db/boltdb.go
```diff
@@ -0,0 +1,305 @@
...
+func init() {
+	registerDBCreator(BoltDBBackend, func(name, dir string) (DB, error) {
+		return NewBoltDB(name, dir)
+	}, false)
+}
+
+// BoltDB is a wrapper around etcd's fork of bolt
+// (https://github.com/etcd-io/bbolt).
+//
+// NOTE: All operations (including Set, Delete) are syncronous by default. One
```
golangcibot: `syncronous` is a misspelling of `synchronous` (from `misspell`)

libs/db/boltdb.go
```diff
@@ -0,0 +1,305 @@
...
+	m["FreelistInuse"] = fmt.Sprintf("%v", stats.FreelistInuse)
+
+	// Transaction stats
+	m["TxN"] = fmt.Sprintf("%v", stats.TxN)
+	m["OpenTxN"] = fmt.Sprintf("%v", stats.OpenTxN)
+
+	return m
+}
+
+// boltDBBatch stores key values in syncronous map and dumps them to the
```
golangcibot: `syncronous` is a misspelling of `synchronous` (from `misspell`)

libs/db/boltdb.go
```diff
@@ -0,0 +1,305 @@
...
+
+func (bdb *boltDBBatch) Set(key, value []byte) {
+	bdb.buffer.Store(key, value)
+}
+
+func (bdb *boltDBBatch) Delete(key []byte) {
+	bdb.buffer.Delete(key)
+}
+
+// NOTE: the operation is syncronous (see BoltDB for reasons)
```
golangcibot: `syncronous` is a misspelling of `synchronous` (from `misspell`)

libs/db/boltdb_test.go
```diff
@@ -0,0 +1,95 @@
...
+	ro2, err := NewBoltDBWithOpts(name, "", &bbolt.Options{ReadOnly: true})
+	defer ro2.Close()
+	require.Nil(t, err)
+}
+
+func TestBoltDBGetSet(t *testing.T) {
+	// TODO
+}
+
+func BenchmarkBoltDBRandomReadsWrites(b *testing.B) {
```
golangcibot: 38-95 lines are duplicate of `libs/db/go_level_db_test.go:35-92` (from `dupl`)

libs/db/common_test.go
```diff
@@ -188,3 +188,66 @@ func (mockIterator) Value() []byte {
...
+}
+
+func int642Bytes(i int64) []byte {
+	buf := make([]byte, 8)
+	binary.BigEndian.PutUint64(buf, uint64(i))
+	return buf
+}
+
+func bytes2Int64(buf []byte) int64 {
+	return int64(binary.BigEndian.Uint64(buf))
```
golangcibot: undeclared name: `binary` (from `typecheck`)

libs/db/common_test.go
```diff
@@ -188,3 +188,66 @@ func (mockIterator) Value() []byte {
...
+		internal[int64(i)] = int64(0)
+	}
+
+	// fmt.Println("ok, starting")
+	b.StartTimer()
+
+	for i := 0; i < b.N; i++ {
+		// Write something
+		{
+			idx := int64(cmn.RandInt()) % numItems
```
golangcibot: undeclared name: `cmn` (from `typecheck`)

libs/db/common_test.go
```diff
@@ -188,3 +188,66 @@ func (mockIterator) Value() []byte {
...
+
+		// Read something
+		{
+			idx := int64(cmn.RandInt()) % numItems
+			valExp := internal[idx]
+			idxBytes := int642Bytes(int64(idx))
+			valBytes := db.Get(idxBytes)
+			//fmt.Printf("Get %X -> %X\n", idxBytes, valBytes)
+			if valExp == 0 {
+				if !bytes.Equal(valBytes, nil) {
```
golangcibot: undeclared name: `bytes` (from `typecheck`)

melekes:
@CrocdileChan TestDBIterator test is failing with:

```
=== RUN   TestDBIterator/boltdb
panic: runtime error: index out of range [recovered]
	panic: runtime error: index out of range

goroutine 32 [running]:
testing.tRunner.func1(0xc00019c200)
	/usr/local/go/src/testing/testing.go:830 +0x69d
panic(0xb86820, 0x1134a10)
	/usr/local/go/src/runtime/panic.go:522 +0x1b5
encoding/binary.bigEndian.Uint64(...)
	/usr/local/go/src/encoding/binary/binary.go:124
github.com/tendermint/tendermint/libs/db.bytes2Int64(...)
	/go/src/github.com/tendermint/tendermint/libs/db/common_test.go:255
github.com/tendermint/tendermint/libs/db.verifyIterator(0xc00019c200, 0xce4060, 0xc0001669a0, 0xc0000262d0, 0x9, 0x9, 0xc11619, 0x10)
	/go/src/github.com/tendermint/tendermint/libs/db/backend_test.go:218 +0x425
github.com/tendermint/tendermint/libs/db.testDBIterator(0xc00019c200, 0xc0d9c4, 0x6)
	/go/src/github.com/tendermint/tendermint/libs/db/backend_test.go:179 +0x34c
github.com/tendermint/tendermint/libs/db.TestDBIterator.func1(0xc00019c200)
	/go/src/github.com/tendermint/tendermint/libs/db/backend_test.go:162 +0x6b
testing.tRunner(0xc00019c200, 0xc000ef0590)
	/usr/local/go/src/testing/testing.go:865 +0x164
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:916 +0x65b
FAIL	github.com/tendermint/tendermint/libs/db	0.061s
```

CrocdileChan:
It seems that go_leveldb is wrong when you  run  the  TestIterator()  :-p

BY THE WAY.
Am I going to be a contributor of yours when this PR merged ?   😂

libs/db/boltdb.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+}
+
+func (bdb *boltDBBatch) WriteSync() {
+	bdb.Write()
+}
+
+func (bdb *boltDBBatch) Close() {}
+
+// WARNING: Any concurrent writes (Set, SetSync) will block until the Iterator
```
melekes: this is concerning. should we rewrite Iterator to execute transaction for each Next() call rather then holding a lock?

libs/db/boltdb.go
```diff
@@ -0,0 +1,313 @@
...
+func (itr *boltDBIterator) Close() {}
+
+func (itr *boltDBIterator) assertIsValid() {
+	if !itr.Valid() {
+		panic("Boltdb-iterator is invalid")
+	}
+}
+
+// nonEmptyKey returns a []byte("nil") if key is empty.
+// WARNING: this may collude with "nil" user key!
```
melekes: bolt does not support empty keys!

melekes:
> It seems that go_leveldb is wrong when you run the TestIterator() :-p

what do you mean?

> Am I going to be a contributor of yours when this PR merged ?

yes

libs/db/boltdb.go
CrocdileChan: I think that we should judge if the key is empty.  
If the key is empty,  we should return an error (key must not be empty)
In fact, many embedded KVDBs do not allow key to be empty.

We should modify the nonEmptyKey() that return an error when key is empty.

Liamsi: Are we using non-empty keys anywhere (the sdk)? If not, returning an error makes sense. Otherwise, we would need to support existing use-cases nonetheless.

libs/db/boltdb.go
CrocdileChan: Do you mean that `*boltDBBatch.Iterator(start,end)`  will block  the concurrent writes？
But why ？  `*boltDBBatch.Iterator(start,end)`  only creates an read-only transaction.   Why do you think that concurrent writes will block?

melekes: > Do you mean that *boltDBBatch.Iterator(start,end) will block the concurrent writes？

yes. to verify comment out `if backend = BoltDBBackend` condition in `TestPrefixIteratorNoMatch1`

libs/db/boltdb.go
```diff
@@ -0,0 +1,320 @@
...
+	}, false)
+}
+
+// BoltDB is a wrapper around etcd's fork of bolt
+// (https://github.com/etcd-io/bbolt).
+//
+// NOTE: All operations (including Set, Delete) are synchronous by default. One
+// can globally turn it off by using NoSync config option (not recommended).
+//
+// A sigle bucket is used per a database instance. This could lead to
```
Liamsi:
```suggestion
// A single bucket is used per a database instance. This could lead to
```

libs/db/boltdb.go
melekes: > Are we using non-empty keys anywhere (the sdk)?

This logic was explicitly added by Jae. Not sure what for. Maybe for SDK?

Liamsi: OK, then let's keep it like you've implemented it (for now).

libs/db/boltdb.go
```diff
@@ -0,0 +1,320 @@
...
+// performance issues when/if there will be lots of keys.
```
Liamsi: Could it be important to document that the single bucket is `tm`?

libs/db/boltdb.go
CrocdileChan: I want to know where to use empty-key in cosmos-sdk.  
`[]byte("nil")` is not a good idea because you probably judge the key in other places.   
And if you let the empty-key keep it ,  it will be incompatible with other better KVDBs.

libs/db/boltdb.go
```diff
@@ -0,0 +1,320 @@
...
+
+func (bdb *BoltDB) Close() {
+	bdb.db.Close()
+}
+
+func (bdb *BoltDB) Print() {
+	stats := bdb.db.Stats()
+	fmt.Printf("%v\n", stats)
+
+	bdb.db.View(func(tx *bbolt.Tx) error {
```
Liamsi: Doesn't View return an error?

libs/db/boltdb_test.go
```diff
@@ -0,0 +1,35 @@
...
+	cmn "github.com/tendermint/tendermint/libs/common"
+)
+
+func TestBoltDBNewBoltDB(t *testing.T) {
+	name := fmt.Sprintf("test_%x", cmn.RandStr(12))
+	dir := os.TempDir()
+	defer cleanupDBDir(dir, name)
+
+	db, err := NewBoltDB(name, dir)
+	require.Nil(t, err)
```
Liamsi:
```suggestion
	require.NoError(t, err)
```
(slightly more expressive)

libs/db/boltdb_test.go
```diff
@@ -0,0 +1,35 @@
...
+	db.Close()
+}
+
+func BenchmarkBoltDBRandomReadsWrites(b *testing.B) {
```
Liamsi: BTW, my mbp:
```
BenchmarkBoltDBRandomReadsWrites-12    	     100	  13983447 ns/op
BenchmarkGoLevelDBRandomReadsWrites-12    	  100000	     13278 ns/op
```
Isn't this supposed to be faster?

melekes: LOL

melekes: 1) BenchmarkBoltDBRandomReadsWrites does not reflect our load
2) I'll double check. If bolt is slower for our load, I'll be 😢 because I spent a good portion of the day working on top of https://github.com/tendermint/tendermint/pull/3604

melekes: 3. It's not fair to compare goleveldb and bolt in this case (BenchmarkBoltDBRandomReadsWrites) because Set is synchronous in bolt! We can turn sync in this benchmark by using NoSync option

CrocdileChan: Sounds great.   By the way, the bolt is better than leveldb in random read because it's based on B+TREE. 
  I will find some authority benchmark reports for you tomorrow. Now it's too late in China I will go for sleep.
HAPPY  MAY DAY  😁😁

libs/db/boltdb.go
melekes: cc @alexanderbez @jaekwon

melekes: > Ethan: also how do we handle the empty key (non nil) on the fsdb?
> Jae: mmm lets panic for fsdb.

melekes: 4ce8448d7fcf92b040046f894474ce2f7e779b67

melekes:
> I would wait for the benchmarks (as per https://github.com/tendermint/tendermint/pull/3610/files#r280132399) before we merge this

benchmarks should not hold us from merging this.

CrocdileChan:
> > I would wait for the benchmarks (as per https://github.com/tendermint/tendermint/pull/3610/files#r280132399) before we merge this
> 
> benchmarks should not hold us from merging this.

So now what's blocking merging this ?  The NIL key?

libs/db/boltdb.go
melekes: https://github.com/tendermint/tendermint/issues/3614

CrocdileChan:
I have a question 😂   why  am not  the contributor of yours now?   The number of contributors is still only 133.    You promise that I will be the contributor of you when this PR is merged.

melekes:
> why am not the contributor of yours now?

please wait till this gets merged to master (0.31.6 release)

CrocdileChan:
ok!  Thank you very much.

libs/db/boltdb.go
leoluk: Yes, BoltDB does not support concurrent transactions. This code does not work, as the transactions created are never closed, and the next write will deadlock. The tx needs to be included in the iterator struct and closed.

Liamsi:
Thanks a lot for the extensive testing and feedback @leoluk! 
Would you argue that there is absolutely to reason to add boltdb support? Remember that tendermint is not only used by the cosmos-hub.

CrocdileChan:
@leoluk   thanks for your benchmark and test.  Is it possible that tendermint will be used in read-more-an-write-less ?   b+tree will performs better than lsm-tree in read-heavy.  I think that we should show more storage implementions in tendermint.

libs/db/boltdb.go
melekes: Wait. Didn't I do that?

melekes: Oh no. I will fix. Thank you so much for pointing this out 💚

leoluk:
Demonstrating a specific use case where boltdb actually outperforms goleveldb would be good to justify the expense of maintaining it. The most common use-case - syncing a chain - is write-heavy due to how the IAVL tree works, which is not specific to cosmos-sdk.

melekes:
> The most common use-case - syncing a chain - is write-heavy due to how the IAVL tree works, which is not specific to cosmos-sdk.

most of the users will use state-sync once it's implemented.

PR #3615, opened 2019-05-03
rpc: /block_results fix docs + write test + restructure response

melekes:
BREAKING

Example response:

```json
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    "height": "2109",
    "txs_results": null,
    "begin_block_events": null,
    "end_block_events": null,
    "validator_updates": null,
    "consensus_param_updates": null
  }
}
```

Old result consisted of ABCIResponses struct and height. Exposing
internal ABCI structures (which we store in state package) in RPC seems
bad to me for the following reasons:

1) high risk of breaking the API when somebody changes internal structs
(This has already happened before! We changed the ABCIResponses struct at some point in the past => /block_results response format got updated. There's a high chance we didn't even notice it (~ no changelog entry))
2) RPC is aware of ABCI, which I'm not sure we want

Refs https://github.com/tendermint/tendermint/issues/3616

* [x] Wrote tests

### DIFF
--- a/rpc/core/blocks.go
+++ b/rpc/core/blocks.go
@@ -124,11 +124,14 @@ func BlockResults(ctx *rpctypes.Context, heightPtr *int64) (*ctypes.ResultBlockR
 		return nil, err
 	}
 
-	res := &ctypes.ResultBlockResults{
-		Height:  height,
-		Results: results,
-	}
-	return res, nil
+	return &ctypes.ResultBlockResults{
+		Height:                height,
+		TxsResults:            results.DeliverTxs,
+		BeginBlockEvents:      results.BeginBlock.Events,
+		EndBlockEvents:        results.EndBlock.Events,
+		ValidatorUpdates:      results.EndBlock.ValidatorUpdates,
+		ConsensusParamUpdates: results.EndBlock.ConsensusParamUpdates,
+	}, nil
 }
 
 func getHeight(currentHeight int64, heightPtr *int64) (int64, error) {
--- a/rpc/core/blocks_test.go
+++ b/rpc/core/blocks_test.go
@@ -4,11 +4,18 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	abci "github.com/tendermint/tendermint/abci/types"
+	ctypes "github.com/tendermint/tendermint/rpc/core/types"
+	rpctypes "github.com/tendermint/tendermint/rpc/lib/types"
+	sm "github.com/tendermint/tendermint/state"
+	"github.com/tendermint/tendermint/types"
+	dbm "github.com/tendermint/tm-db"
 )
 
 func TestBlockchainInfo(t *testing.T) {
-
 	cases := []struct {
 		min, max     int64
 		height       int64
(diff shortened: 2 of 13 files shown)


rpc/core/blocks_test.go
```diff
@@ -54,5 +62,59 @@ func TestBlockchainInfo(t *testing.T) {
...
+	testCases := []struct {
+		height  int64
+		wantErr bool
+		wantRes *ctypes.ResultBlockResults
+	}{
+		{-1, true, nil},
+		{0, true, nil},
+		{101, true, nil},
+		{100, false, &ctypes.ResultBlockResults{
+			Height:     100,
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

rpc/core/types/responses.go
```diff
@@ -38,8 +37,10 @@ type ResultCommit struct {
 
 // ABCI results from a block
 type ResultBlockResults struct {
-	Height  int64                `json:"height"`
-	Results *state.ABCIResponses `json:"results"`
+	Height                int64                     `json:"height"`
+	TxsResults            []*abci.ResponseDeliverTx `json:"txs_results"`
+	ValidatorUpdates      []abci.ValidatorUpdate    `json:"validator_updates"`
+	ConsensusParamUpdates *abci.ConsensusParams     `json:"consensus_param_updates"`
```
Liamsi: 👍

melekes:
Note: response no longer contains EndBlock.Tags and BeginBlock.Tags. a) not sure people will need it in RPC b) it looks like another implementation detail

ebuchman:
I agree that in general we don't want to just expose internal structs over external interfaces. But was there something particular that precipitated this change, and if so can we get an issue for it? We shouldn't have breaking changes like this without clear rationale for them. 

Also what are you referring to with `(HAPPENED HERE!)` ?

> Note: response no longer contains EndBlock.Tags and BeginBlock.Tags. a) not sure people will need it in RPC b) it looks like another implementation detail

Not sure I understand this - how is it an implementation detail? It's fundamental data about block execution. We probably shouldn't remove it.

melekes:
> Also what are you referring to with (HAPPENED HERE!) ?

we've changed the ABCIResponses struct => /block_results response format got changed. I am not sure, but there's a high chance we haven't even noticed it (~ no changelog entry)

> Not sure I understand this - how is it an implementation detail? It's fundamental data about block execution. We probably shouldn't remove it.

implementation in the sense that we process tags using the indexer

I'll add them back. But I still think we shouldn't expose Begin/EndBlock in RPC

ebuchman:
Yeh good points. Could you write up an issue about why you don't think the Begin/EndBlock should be in RPC? How else would users get that data?

marbar3778:
What is the status here? I see label:`breaking`,  when is the plan to include this

melekes:
I will work on this tomorrow.

rpc/core/blocks.go
```diff
@@ -364,25 +364,38 @@ func Commit(ctx *rpctypes.Context, heightPtr *int64) (*ctypes.ResultCommit, erro
...
+//         "gas_used": 2,
+//         "tags": [
+//           {
+//             "key": "YXBwLmNyZWF0b3I=",
+//             "value": "Q29zbW9zaGkgTmV0b3dva28="
+//           },
+//         ]
+//       },
+//    ],
+//    "begin_block_events": null,
```
melekes: Should we omit empty fields?

state/store.go
```diff
@@ -165,8 +165,9 @@ func LoadABCIResponses(db dbm.DB, height int64) (*ABCIResponses, error) {
 
 // SaveABCIResponses persists the ABCIResponses to the database.
 // This is useful in case we crash after app.Commit and before s.Save().
-// Responses are indexed by height so they can also be loaded later to produce Merkle proofs.
-func saveABCIResponses(db dbm.DB, height int64, abciResponses *ABCIResponses) {
+// Responses are indexed by height so they can also be loaded later to produce
+// Merkle proofs.
+func SaveABCIResponses(db dbm.DB, height int64, abciResponses *ABCIResponses) {
```
melekes: Had to export this method in order to test /block_results.

rpc/core/blocks.go
marbar3778: Don't think it is needed here

rpc/core/blocks.go
```diff
@@ -364,25 +364,38 @@ func Commit(ctx *rpctypes.Context, heightPtr *int64) (*ctypes.ResultCommit, erro
...
-//         }
-//       ],
-//       "end_block": {
-//         "validator_updates": null
+//     "height": "437",
+//     "txs_results": [
+//       {
+//         "gas_wanted": 1,
+//         "gas_used": 1,
+//         "tags": [
```
AdityaSripal: shouldn't this be replaced by "events"

rpc/core/blocks.go
```diff
@@ -396,11 +409,14 @@ func BlockResults(ctx *rpctypes.Context, heightPtr *int64) (*ctypes.ResultBlockR
...
 	}
 
-	res := &ctypes.ResultBlockResults{
-		Height:  height,
-		Results: results,
-	}
-	return res, nil
+	return &ctypes.ResultBlockResults{
+		Height:                height,
+		TxsResults:            results.DeliverTx,
```
AdityaSripal: I think `ABCIResponses` in `tendermint/state/store.go` can be refactored to:
```
type ABCIResponses struct {
	TxsResults  []*abci.ResponseDeliverTx `json:"deliver_tx"`
	EndBlock    *abci.ResponseEndBlock    `json:"end_block"`
	BeginBlock *abci.ResponseBeginBlock  `json:"begin_block"`
}
```

marbar3778:
**note:** update swagger

PR #3613, opened 2019-05-02
Switch to go mod

jackzampolin:
This PR switches dependency management to using `go mod`.

PR #3623, opened 2019-05-05
Fix #3613

Liamsi:
This PR builds on top of #3613. It:

 - reverts 45117bd which deactivated all caches
 - brings back the dependency cache (but make it use modules instead of the `vendor/` directory)
 - bumps versions of caches related to modules as recommended by https://circleci.com/docs/2.0/caching/#clearing-cache (to clear all caches)


alessio:
Seems fine to me. ACK

Issue #2185, opened 2018-08-08
mempool: punish peers sending txs that return certain ABCI codes in mempool


ebuchman:
Currently we just ignore invalid txs sent in the mempool since an honest peer could send txs that are invalid according to our local mempool state. 

However some txs may be invalid no matter what the state, like invalid encoding, and perhaps invalid signature (though careful with how you define signature).

Perhaps we could consider punishing peers for sending us txs that are invalid in this way by making better use of standardized ABCI codes.

This was brought up in https://github.com/tendermint/tendermint/issues/2175

1kevinxu1:
Is it possible to establish a "threshold" of invalid/valid messages that resets/decays per "threshold period", in such a way that if a validator ever crosses the "threshold" set by the network, they are slashed?

ValarDragon:
The easiest way to logically separate these may be declaring a set of ABCI error codes for "This tx was invalid, and never could have been valid". (e.g. bad signature, bad amino encoding)

Then if we get that error code, it stays in the mempool cache, and your peer score is reduced. Otherwise its removed from the mempool. It means that spamming attacks have to have a valid signature at minimum, and therefore can be related to an on-chain identity.

> Is it possible to establish a "threshold" of invalid/valid messages that resets/decays per "threshold period", in such a way that if a validator ever crosses the "threshold" set by the network, they are slashed?

That sounds more related to

marbar3778:
Open Question:

- When and under what circumstances should we disconnect from a peer or mark as a bad peer?

PR #3625, opened 2019-05-06
mempool: remove only valid (Code==0) txs on Update

### LINKED ISSUE #3322: Evil proposers can drop valid txs in Commit stage
In the end of commit stage, we will update mempool to remove all the txs in current block.

```go
// Update mempool.
err = blockExec.mempool.Update(
	block.Height,
	block.Txs,
	TxPreCheck(state),
	TxPostCheck(state),
)
```

Assum an account has 3 transactions in the mempool, the sequences are `100`, `101` and `102` separately, 
So an evil proposal can only package the 101 and 102 transactions into its proposal block, and leave 100 still in mempool, then the two txs will be removed from all validators' mempool when commit.  
So the account lost the two valid txs.

melekes:
so evil proposers can't drop valid txs in Commit stage.

Also remove invalid (Code!=0) txs from the cache so they can be
resubmitted.

Fixes #3322

@rickyyangz:

In the end of commit stage, we will update mempool to remove all the txs
in current block.

```go
// Update mempool.
err = blockExec.mempool.Update(
	block.Height,
	block.Txs,
	TxPreCheck(state),
	TxPostCheck(state),
)
```

Assum an account has 3 transactions in the mempool, the sequences are
100, 101 and 102 separately, So an evil proposal can only package the
101 and 102 transactions into its proposal block, and leave 100 still in
mempool, then the two txs will be removed from all validators' mempool
when commit. So the account lost the two valid txs.

@ebuchman:

In the longer term we may want to do something like #2639 so we can
validate txs before we commit the block. But even in this case we'd only
want to run the equivalent of CheckTx, which means the DeliverTx could
still fail even if the CheckTx passes depending on how the app handles
the ABCI Code semantics. So more work will be required around the ABCI
code. See also #2185

* [x] Wrote tests

### DIFF
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -519,6 +519,7 @@ func (mem *CListMempool) ReapMaxTxs(max int) types.Txs {
 func (mem *CListMempool) Update(
 	height int64,
 	txs types.Txs,
+	deliverTxResponses []*abci.ResponseDeliverTx,
 	preCheck PreCheckFunc,
 	postCheck PostCheckFunc,
 ) error {
@@ -533,20 +534,37 @@ func (mem *CListMempool) Update(
 		mem.postCheck = postCheck
 	}
 
-	// Add committed transactions to cache (if missing).
-	for _, tx := range txs {
-		_ = mem.cache.Push(tx)
-	}
+	for i, tx := range txs {
+		if deliverTxResponses[i].Code == abci.CodeTypeOK {
+			// Add valid committed tx to the cache (if missing).
+			_ = mem.cache.Push(tx)
 
-	// Remove committed transactions.
-	txsLeft := mem.removeTxs(txs)
+			// Remove valid committed tx from the mempool.
+			if e, ok := mem.txsMap.Load(txKey(tx)); ok {
+				mem.removeTx(tx, e.(*clist.CElement), false)
+			}
+		} else {
+			// Allow invalid transactions to be resubmitted.
+			mem.cache.Remove(tx)
+
+			// Don't remove invalid tx from the mempool.
+			// Otherwise evil proposer can drop valid txs.
+			// Example:
+			//   100 -> 101 -> 102
+			// Block, proposed by evil proposer:
+			//   101 -> 102
+			// Mempool (if you remove txs):
+			//   100
+			// https://github.com/tendermint/tendermint/issues/3322.
+		}
+	}
 
 	// Either recheck non-committed txs to see if they became invalid
 	// or just notify there're some txs left.
-	if len(txsLeft) > 0 {
+	if mem.Size() > 0 {
 		if mem.config.Recheck {
-			mem.logger.Info("Recheck txs", "numtxs", len(txsLeft), "height", height)
-			mem.recheckTxs(txsLeft)
+			mem.logger.Info("Recheck txs", "numtxs", mem.Size(), "height", height)
+			mem.recheckTxs()
 			// At this point, mem.txs are being rechecked.
 			// mem.recheckCursor re-scans mem.txs and possibly removes some txs.
 			// Before mem.Reap(), we should wait for mem.recheckCursor to be nil.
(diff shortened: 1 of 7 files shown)


mempool/clist_mempool_test.go
```diff
@@ -170,23 +170,37 @@ func TestMempoolFilters(t *testing.T) {
...
+	// 2. Removes valid txs from the mempool
+	err = mempool.CheckTx([]byte{0x02}, nil)
+	require.NoError(t, err)
+	mempool.Update(1, []types.Tx{[]byte{0x02}}, abciResponses(1, abci.CodeTypeOK), nil, nil)
+	assert.Zero(t, mempool.Size())
+
+	// 3. Removes invalid transactions from the cache, but leaves them in the mempool (if present)
+	err = mempool.CheckTx([]byte{0x03}, nil)
+	require.NoError(t, err)
+	mempool.Update(1, []types.Tx{[]byte{0x03}}, abciResponses(1, 1), nil, nil)
```
Liamsi: Wouldn't this be more consistent: 
```suggestion
	mempool.Update(1, []types.Tx{[]byte{0x03}}, abciResponses(1, abci.CodeTypeEncodingError), nil, nil)
```

mempool/clist_mempool_test.go
```diff
@@ -170,23 +170,37 @@ func TestMempoolFilters(t *testing.T) {
...
+	mempool.Update(1, []types.Tx{[]byte{0x01}}, abciResponses(1, abci.CodeTypeOK), nil, nil)
 	err := mempool.CheckTx([]byte{0x01}, nil)
 	if assert.Error(t, err) {
 		assert.Equal(t, ErrTxInCache, err)
 	}
+
+	// 2. Removes valid txs from the mempool
+	err = mempool.CheckTx([]byte{0x02}, nil)
+	require.NoError(t, err)
+	mempool.Update(1, []types.Tx{[]byte{0x02}}, abciResponses(1, abci.CodeTypeOK), nil, nil)
```
Liamsi: The Update is what is removing the valid tx. The comment (`// 2. Removes valid txs from the mempool`) should probably go there, or not?

mempool/clist_mempool_test.go
```diff
@@ -170,23 +170,37 @@ func TestMempoolFilters(t *testing.T) {
...
+func TestMempoolUpdate(t *testing.T) {
 	app := kvstore.NewKVStoreApplication()
 	cc := proxy.NewLocalClientCreator(app)
 	mempool, cleanup := newMempoolWithApp(cc)
 	defer cleanup()
-	mempool.Update(1, []types.Tx{[]byte{0x01}}, nil, nil)
+
+	// 1. Adds valid txs to the cache
+	mempool.Update(1, []types.Tx{[]byte{0x01}}, abciResponses(1, abci.CodeTypeOK), nil, nil)
 	err := mempool.CheckTx([]byte{0x01}, nil)
```
Liamsi: Maybe stupid question: why is the order here: 1) Update 2) CheckTx while it is the other way around in the remaining parts of the test?

mempool/clist_mempool_test.go
Liamsi: Oh, I see that is a different package. NVM then.

mempool/clist_mempool.go
```diff
@@ -533,20 +534,37 @@ func (mem *CListMempool) Update(
...
-	// Remove committed transactions.
-	txsLeft := mem.removeTxs(txs)
+			// Don't remove invalid tx from the mempool.
+			// Otherwise evil proposer can drop valid txs.
+			// Example:
+			//   100 -> 101 -> 102
+			// Block, proposed by evil proposer:
+			//   101 -> 102
+			// Mempool (if you remove txs):
+			//   100
```
Liamsi: Can we write this down as a test case?

melekes: TestMempoolUpdate will fail without the changes

mempool/clist_mempool_test.go
melekes: if we execute mempool.CheckTx([]byte{0x01}, nil) before Update, it will be added to the cache and mempool
 - Update function will remove it from the mempool (2. test case), but leave it in the cache (noop)

now because we want to actually test that Update adds tx to cache, we can't add tx (call CheckTx) before

melekes: https://github.com/tendermint/tendermint/pull/3625/commits/87db96b96d9f838e16d6843a4eb32910fcee0d7a

mempool/clist_mempool_test.go
melekes: https://github.com/tendermint/tendermint/pull/3625/commits/87db96b96d9f838e16d6843a4eb32910fcee0d7a

melekes: Is https://github.com/tendermint/tendermint/commit/87db96b96d9f838e16d6843a4eb32910fcee0d7a better?

mempool/clist_mempool_test.go
Liamsi: Thanks for the explanation 💚

mempool/clist_mempool.go
Liamsi: I understand but I wondered if we should have the above example in a test instead of what is there in `TestMempoolUpdate`. But NVM. TestMempoolUpdate is actually better as it is more minimalistic!

Issue #3626, opened 2019-05-06
Follow-up on boltdb integration


Liamsi:
Particularly:
 - https://github.com/tendermint/tendermint/pull/3610#pullrequestreview-233751678
 - https://github.com/tendermint/tendermint/pull/3610#discussion_r280995538
 - https://github.com/tendermint/tendermint/pull/3610#discussion_r281201274

Maybe we should mark boltdb support as highly experimental and not throughly testes in the changelog, too.
ref: #3610, #3611

leoluk:
Another follow-up is adding missing tests: https://codecov.io/gh/tendermint/tendermint/pull/3610/diff?src=pr&el=tree#diff-bGlicy9kYi9ib2x0ZGIuZ28=

PR #3627, opened 2019-05-06
close boltDBIterator

melekes:
Refs https://github.com/tendermint/tendermint/pull/3610#discussion_r281201274

If we do not close, other txs will be stuck forever.

* [ ] Wrote tests

### DIFF
--- a/libs/db/boltdb.go
+++ b/libs/db/boltdb.go
@@ -190,8 +190,8 @@ func (bdb *boltDBBatch) WriteSync() {
 
 func (bdb *boltDBBatch) Close() {}
 
-// WARNING: Any concurrent writes (Set, SetSync) or reads will block until the
-// Iterator is closed.
+// WARNING: Any concurrent writes or reads will block until the iterator is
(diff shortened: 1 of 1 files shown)


melekes:
We should probably write more tests, but can be done later at this particular case.

Liamsi:
ref: https://github.com/tendermint/tendermint/issues/3626

libs/db/boltdb.go
```diff
@@ -190,31 +190,31 @@ func (bdb *boltDBBatch) WriteSync() {
 
 func (bdb *boltDBBatch) Close() {}
 
-// WARNING: Any concurrent writes (Set, SetSync) will block until the Iterator
-// is closed.
+// WARNING: Any concurrent writes (Set, SetSync) or reads will block until the
```
Liamsi: Is a `Delete` not also a write op? I think, I'd remove `(Set, SetSync)` from the comment, or, add all write ops and read ops, too.

Issue #3611, opened 2019-05-01
libs/db: conditional compilation


melekes:
We should also use conditional compilation to only include deps & code when X db is requested (// +build boltdb).
(https://github.com/tendermint/tendermint/pull/3604#discussion_r279671629)

Refs https://github.com/tendermint/tendermint/issues/3273

Also we should move implementations into separate packages (bolt, goleveldb, memdb, etc).

melekes:
Adding this to v0.31.6 milestone mostly because I don't want to include any more DBs dependencies into the binary. E.g. https://github.com/tendermint/tendermint/pull/3610 boltdb.

melekes:
Merged https://github.com/tendermint/tendermint/pull/3628 to develop.

PR #3628, opened 2019-05-06
libs/db: conditional compilation

melekes:
Fixes https://github.com/tendermint/tendermint/issues/3611

* [ ] ~~Wrote tests~~

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -41,7 +41,9 @@
 - [p2p] \#3531 Terminate session on nonce wrapping (@climber73)
 - [libs/db] \#3611 Conditional compilation
   * Use `cleveldb` tag instead of `gcc` to compile Tendermint with CLevelDB or
-    use `make build_c` / `make install_c`
+    use `make build_c` / `make install_c` (full instructions can be found at
+    https://tendermint.com/docs/introduction/install.html#compile-with-cleveldb-support)
+  * Use `boltdb` tag to compile Tendermint with bolt db
 
 ### BUG FIXES:
 - [p2p] \#3532 limit the number of attempts to connect to a peer in seed mode
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -22,7 +22,6 @@
   * `MockMempool` moved to top-level mock package and renamed to `Mempool`
 - [libs/common] Removed `PanicSanity`, `PanicCrisis`, `PanicConsensus` and `PanicQ`
 - [node] Moved `GenesisDocProvider` and `DefaultGenesisDocProviderFunc` to state package
-- [libs/db] \#3611 Removed LevelDBBackend
 
 * Blockchain Protocol
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -22,6 +22,7 @@
   * `MockMempool` moved to top-level mock package and renamed to `Mempool`
 - [libs/common] Removed `PanicSanity`, `PanicCrisis`, `PanicConsensus` and `PanicQ`
 - [node] Moved `GenesisDocProvider` and `DefaultGenesisDocProviderFunc` to state package
+- [libs/db] \#3611 Removed LevelDBBackend
 
 * Blockchain Protocol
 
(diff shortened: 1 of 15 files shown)


config/config.go
```diff
@@ -153,7 +153,18 @@ type BaseConfig struct {
...
+	//   - pure go
+	//   - stable
+	// * cleveldb (uses levigo wrapper)
+	//   - fast
+	//   - requires gcc
+	//   - use cleveldb build tag (go build -tags cleveldb)
+	// * boltdb (uses etcd's fork of bolt - github.com/etcd-io/bbolt)
+	//   - EXPERIMENTAL
+	//   - may be faster is some use-cases (random reads - indexer)
+	//   - use boltdb build tag (go build -tags boltdb)
```
Liamsi: 👍

docs/tendermint-core/configuration.md
```diff
@@ -30,7 +30,18 @@ moniker = "anonymous"
...
+#   - pure go
+#   - stable
+# * cleveldb (uses levigo wrapper)
+#   - fast
+#   - requires gcc
+#   - use cleveldb build tag (go build -tags cleveldb)
+# * boltdb (uses etcd's fork of bolt - github.com/etcd-io/bbolt)
+#   - EXPERIMENTAL
+#   - may be faster is some use-cases (random reads - indexer)
+#   - use boltdb build tag (go build -tags boltdb)
```
Liamsi: Maybe link to this in the changelog file?

CHANGELOG_PENDING.md
```diff
@@ -38,6 +39,9 @@
 - [rpc] \#3362 `/dial_seeds` & `/dial_peers` return errors if addresses are incorrect (except when IP lookup fails)
 - [node] \#3362 returns an error if `persistent_peers` list is invalid (except when IP lookup fails)
 - [p2p] \#3531 Terminate session on nonce wrapping (@climber73)
+- [libs/db] \#3611 Conditional compilation
+  * Use `cleveldb` tag instead of `gcc` to compile Tendermint with CLevelDB or
```
Liamsi: We have to be careful that by moving from `gcc` to some custom build flag we did not change the default behaviour. When `CGO_ENABLED=1` the `cgo` (and the `gcc`?) flag would have been implicitly be set. Not true for `cleveldb`. 
Should the changelog state which backend will be used by default?

libs/db/db.go
```diff
@@ -5,18 +5,33 @@ import (
...
+	// FSDBBackend represents filesystem database
+	//	 - EXPERIMENTAL
+	//   - slow
+	FSDBBackend DBBackendType = "fsdb"
+	// BoltDBBackend represents bolt (uses etcd's fork of bolt -
+	// github.com/etcd-io/bbolt)
+	//   - EXPERIMENTAL
+	//   - may be faster is some use-cases (random reads - indexer)
+	//   - use boltdb build tag (go build -tags boltdb)
+	BoltDBBackend DBBackendType = "boltdb"
```
Liamsi: Isn't removing a backend (`LevelDBBackend`) breaking?

melekes: It is

CHANGELOG_PENDING.md
melekes: > Should the changelog state which backend will be used by default?

mm, I thought everybody knows it is goleveldb and we did not change it here, so why state it?

melekes: > When CGO_ENABLED=1 the cgo (and the gcc?) flag would have been implicitly be set. Not true for cleveldb.

good point 👍 but I think we still want to make it implicit - i.e. if you want to compile with cleveldb, please provide cleveldb build constraint, etc.

libs/db/db.go
Liamsi: Should add an entry in under the `BREAKING CHANGES` section then?

melekes: I did `[libs/db] \#3611 Removed LevelDBBackend`

melekes:
Made this non-breaking.

CHANGELOG_PENDING.md
ebuchman: *explicit

Makefile
```diff
@@ -132,7 +132,7 @@ clean_certs:
 	rm -f db/remotedb/::.crt db/remotedb/::.key
 
 test_libs: gen_certs
-	go test -tags gcc $(PACKAGES)
+	go test -tags clevedb boltdb $(PACKAGES)
```
ebuchman: Shouldn't these tags be in quotes? Otherwise it looks like we're trying to run the `boltdb` package ... shuold be `-tags "cleveldb boltdb"`

PR #3629, opened 2019-05-06
libs/db: fix boltdb batching

leoluk:
`[]byte` is unhashable and needs to be casted to a string.

See https://github.com/tendermint/tendermint/pull/3610#discussion_r280995538

Ref https://github.com/tendermint/tendermint/issues/3626

CC @melekes @Liamsi

* [ ] Wrote tests

### DIFF
--- a/libs/db/boltdb.go
+++ b/libs/db/boltdb.go
@@ -161,11 +161,11 @@ func (bdb *BoltDB) NewBatch() Batch {
 }
 
 func (bdb *boltDBBatch) Set(key, value []byte) {
-	bdb.buffer.Store(key, value)
+	bdb.buffer.Store(string(key), value)
 }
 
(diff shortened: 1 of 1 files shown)


melekes:

Note we should probably use sync.Mutex and array of KV pairs instead of casting.

Liamsi:
> []byte is unhashable and needs to be casted to a string.

As far as I understand, the reason for this change is rather that `[]byte` is not comparable via `==`. This would be necessary as `bdb.buffer.Store(...)` operates on a (sync) map.

Liamsi:
> Note we should probably use sync.Mutex and array of KV pairs instead of casting.

Agree with @melekes here. The map only forces us to cast top strings only to cast back to []byte later.

PR #3633, opened 2019-05-07
libs/db: boltdb: use slice instead of sync.Map

### LINKED ISSUE #3631: libs/db: get rid of superfluous key casts
Follow-up on https://github.com/tendermint/tendermint/pull/3629#issuecomment-489902374
[]byte keys are casted to strings to be batched and later casted back to bytes to be stored.

melekes:
for storing batch keys and values in boltDBBatch.

NOTE: batch does not have to be safe for concurrent access. Delete may
be slow, but given it should not be used often, it's ok.

Fixes #3631

* [ ] Wrote tests

### DIFF
--- a/libs/db/boltdb.go
+++ b/libs/db/boltdb.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"sync"
 
 	"github.com/etcd-io/bbolt"
 )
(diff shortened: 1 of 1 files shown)


libs/db/boltdb.go
```diff
@@ -150,36 +149,58 @@ func (bdb *BoltDB) Stats() map[string]string {
...
 	}
 }
 
+// It is safe to modify the contents of the argument after Set returns but not
+// before.
 func (bdb *boltDBBatch) Set(key, value []byte) {
-	bdb.buffer.Store(string(key), value)
+	bdb.buffer = append(bdb.buffer, struct {
+		k []byte
+		v []byte
```
Liamsi: I guess you didn't want to use `KVPair` because it is protobuf and you do not need any encoding here etc?

melekes: yes

PR #3634, opened 2019-05-07
p2p: peer state init too late and pex message too soon

melekes:
Original issue:

fix two issue #3346 and
#3338
1、fix reconnection pex send too fast,
error is caused lastReceivedRequests is still
not deleted when a peer reconnected

2、Peer does not have a state yet. We set it in AddPeer.
We need an new interface before mconnection is started

Replaced https://github.com/tendermint/tendermint/pull/3361

* [ ] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -22,6 +22,7 @@
   * `MockMempool` moved to top-level mock package and renamed to `Mempool`
 - [libs/common] Removed `PanicSanity`, `PanicCrisis`, `PanicConsensus` and `PanicQ`
 - [node] Moved `GenesisDocProvider` and `DefaultGenesisDocProviderFunc` to state package
+- [p2p] \#3346 Reactor#InitPeer method is added to Reactor interface
 
 * Blockchain Protocol
 
@@ -48,9 +49,8 @@
 - [p2p] \#3532 limit the number of attempts to connect to a peer in seed mode
   to 16 (as a result, the node will stop retrying after a 35 hours time window)
 - [consensus] \#2723, \#3451 and \#3317 Fix non-deterministic tests
-<<<<<<< HEAD
-- [p2p] \#3346 Init data structures for peer before starting it (previously was done in AddPeer) (fixes #3346 and #3338; @guagualvcha)
-=======
+- [p2p] \#3346 Init data structures for peer before starting it (previously was
+  done in AddPeer) (fixes #3346 and #3338; @guagualvcha)
 - [p2p] \#3362 make persistent prop independent of conn direction
   * `Switch#DialPeersAsync` now only takes a list of peers
   * `Switch#DialPeerWithAddress` now only takes an address
@@ -60,4 +60,3 @@
   * `Mempool#Update` and `BlockExecutor#Commit` now accept
     `[]*abci.ResponseDeliverTx` - list of `DeliverTx` responses, which should
     match `block.Txs`
->>>>>>> develop
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -50,7 +50,8 @@
   to 16 (as a result, the node will stop retrying after a 35 hours time window)
 - [consensus] \#2723, \#3451 and \#3317 Fix non-deterministic tests
 - [p2p] \#3346 Init data structures for peer before starting it (previously was
-  done in AddPeer) (fixes #3346 and #3338; @guagualvcha)
+  done in AddPeer) (@guagualvcha)
+- [p2p] \#3338 Ensure RemovePeer is always called before InitPeer
 - [p2p] \#3362 make persistent prop independent of conn direction
   * `Switch#DialPeersAsync` now only takes a list of peers
   * `Switch#DialPeerWithAddress` now only takes an address
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -27,3 +27,4 @@
 - [p2p] \#3532 limit the number of attempts to connect to a peer in seed mode
   to 16 (as a result, the node will stop retrying after a 35 hours time window)
 - [consensus] \#2723, \#3451 and \#3317 Fix non-deterministic tests
+- [p2p] \#3346 Init data structures for peer before starting it (previously was done in AddPeer) (fixes #3346 and #3338; @guagualvcha)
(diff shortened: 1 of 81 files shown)


p2p/base_reactor.go
```diff
@@ -5,23 +5,36 @@ import (
...
+	// GetChannels returns the list of MConnection.ChannelDescriptor. Make sure
+	// that each ID is unique across all the reactors added to the switch.
 	GetChannels() []*conn.ChannelDescriptor
 
-	// AddPeer is called by the switch when a new peer is added.
+	// InitPeer is called by the switch before the peer is started. Use it to
+	// initialize data for the peer (e.g. peer state).
+	InitPeer(peer Peer) Peer
+
+	// AddPeer is called by the switch after the peer is added and successfuly
```
golangcibot: `successfuly` is a misspelling of `successfully` (from `misspell`)

p2p/pex/pex_reactor.go
```diff
@@ -204,6 +204,13 @@ func (r *PEXReactor) AddPeer(p Peer) {
 	}
 }
 
+// RemovePeer implements Reactor by resetting peer's requests info.
+func (r *PEXReactor) RemovePeer(p Peer) {
+	id := string(peer.ID())
```
golangcibot: undeclared name: `peer` (from `typecheck`)

consensus/reactor.go
```diff
@@ -438,9 +446,9 @@ func (conR *ConsensusReactor) broadcastHasVoteMessage(vote *types.Vote) {
 
 func makeRoundStepMessage(rs *cstypes.RoundState) (nrsMsg *NewRoundStepMessage) {
 	nrsMsg = &NewRoundStepMessage{
-		Height:                rs.Height,
-		Round:                 rs.Round,
-		Step:                  rs.Step,
+		Height: rs.Height,
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

melekes:
Ok, I think the correct fix for #3338 is to remove the peer from the Switch last (after we've stopped it and removed from all reactors)

p2p/base_reactor.go
golangcibot: `successfuly` is a misspelling of `successfully` (from `misspell`)

p2p/switch.go
```diff
@@ -324,14 +324,20 @@ func (sw *Switch) StopPeerGracefully(peer Peer) {
...
+
 	for _, reactor := range sw.reactors {
 		reactor.RemovePeer(peer, reason)
 	}
+
+	// Removing a peer should go last to avoid a situation where a peer
+	// reconnect to our node and the switch calls InitPeer before
+	// RemovePeer is finished.
+	// https://github.com/tendermint/tendermint/issues/3338
+	if sw.peers.Remove(peer) {
```
melekes: TODO: write a test for it

p2p/base_reactor.go
```diff
@@ -5,23 +5,36 @@ import (
...
 
-	// GetChannels returns the list of channel descriptors.
+	// GetChannels returns the list of MConnection.ChannelDescriptor. Make sure
+	// that each ID is unique across all the reactors added to the switch.
 	GetChannels() []*conn.ChannelDescriptor
 
-	// AddPeer is called by the switch when a new peer is added.
+	// InitPeer is called by the switch before the peer is started. Use it to
+	// initialize data for the peer (e.g. peer state).
+	InitPeer(peer Peer) Peer
```
melekes: I think we should only accept a small part of the peer interface here like

```
	Set(string, interface{})
	Get(string) interface{}
```

this should prevent us from misusing InitPeer function and making an error

CHANGELOG_PENDING.md
```diff
@@ -48,6 +49,9 @@
 - [p2p] \#3532 limit the number of attempts to connect to a peer in seed mode
   to 16 (as a result, the node will stop retrying after a 35 hours time window)
 - [consensus] \#2723, \#3451 and \#3317 Fix non-deterministic tests
+- [p2p] \#3346 Init data structures for peer before starting it (previously was
+  done in AddPeer) (@guagualvcha)
+- [p2p] \#3338 Ensure RemovePeer is always called before InitPeer
```
melekes: ... upon a peer reconnecting to our node

CHANGELOG_PENDING.md
```diff
@@ -48,6 +49,9 @@
 - [p2p] \#3532 limit the number of attempts to connect to a peer in seed mode
   to 16 (as a result, the node will stop retrying after a 35 hours time window)
 - [consensus] \#2723, \#3451 and \#3317 Fix non-deterministic tests
+- [p2p] \#3346 Init data structures for peer before starting it (previously was
```
melekes: - [consensus] Create a peer state in consensus reactor before the peer is started

Liamsi:
I see the test `TestSwitchInitPeerIsNotCalledBeforeRemovePeer ` passes because it only ensures that `reactor.InitPeer` is always only called *after* `reactor.RemovePeer` by checking if it was called at all. This passes on develop because there is no InitPeer method to call.

melekes:
I brought back timeouts, sorry. Maybe there's a way to get rid of them, need to think about it.

p2p/switch_test.go
```diff
@@ -603,6 +604,69 @@ func TestSwitchAcceptRoutineErrorCases(t *testing.T) {
...
+
+	return peer
+}
+
+func (r *mockReactor) InitCalledBeforeRemoveFinished() bool {
+	return atomic.LoadUint32(&r.initCalledBeforeRemoveFinished) == 1
+}
+
+// see stopAndRemovePeer
+func TestSwitchInitPeerIsNotCalledBeforeRemovePeer(t *testing.T) {
```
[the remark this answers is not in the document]
melekes: test passes on develop because there's no InitPeer method
if you replace InitPeer with AddPeer
it will fail on develop

melekes:
Brought back timeouts, sorry. Maybe there's a way to get rid of them, need to think.

Liamsi:
Thanks for removing the timeouts 👍 The test in question paniced now though.

brapse:
Looking back over the previous iteration of this PR it appears that I'm a bit of a broken record. It looks like a better solution might be outside to scope of this particular issue and could be better serviced in a follow up PR. I've tried to capture my reservations in a separate issue #3679 such that at  we can get the current bug fixed. 👍

PR #3638, opened 2019-05-07
libs: Sort Genesis on file SaveAs

jackzampolin:

### DIFF
--- a/libs/common/json_sort.go
+++ b/libs/common/json_sort.go
@@ -1,4 +1,4 @@
-package types
+package common
 
 import "encoding/json"
 
--- a/types/genesis.go
+++ b/types/genesis.go
@@ -47,7 +47,7 @@ func (genDoc *GenesisDoc) SaveAs(file string) error {
 	if err != nil {
 		return err
 	}
-	return cmn.WriteFile(file, genDocBytes, 0644)
+	return cmn.WriteFile(file, MustSortJSON(genDocBytes), 0644)
 }
 
 // ValidatorHash returns the hash of the validator set contained in the GenesisDoc
--- a/types/genesis.go
+++ b/types/genesis.go
@@ -47,7 +47,7 @@ func (genDoc *GenesisDoc) SaveAs(file string) error {
 	if err != nil {
 		return err
 	}
-	return cmn.WriteFile(file, MustSortJSON(genDocBytes), 0644)
+	return cmn.WriteFile(file, cmn.MustSortJSON(genDocBytes), 0644)
 }
 
 // ValidatorHash returns the hash of the validator set contained in the GenesisDoc
--- a/types/utils.go
+++ b/types/utils.go
@@ -1,0 +1,31 @@
+package types
+
+import "encoding/json"
+
+// SortedJSON takes any JSON and returns it sorted by keys. Also, all white-spaces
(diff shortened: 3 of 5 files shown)


jackzampolin:
Related to https://github.com/cosmos/cosmos-sdk/pull/4294

types/utils.go
```diff
@@ -0,0 +1,31 @@
...
+
+// MustSortJSON is like SortJSON but panic if an error occurs, e.g., if
+// the passed JSON isn't valid.
+func MustSortJSON(toSortJSON []byte) []byte {
+	js, err := SortJSON(toSortJSON)
+	if err != nil {
+		panic(err)
+	}
+	return js
+}
```
Liamsi: This is basically copy & pasted from the SDK (added here: https://github.com/cosmos/cosmos-sdk/commit/b7f902fba5fcb367d82a2dd9e59600e91d2dfc72#diff-da7a1d1eef56dc7672476c992ce1e56c) 

I think we should also copy over the tests then. Probably these methods (and tests) do not need to live in both repositories (but that is up to the SDK team).

melekes: Should we move them to `libs/common` package?

Liamsi: Yes, that is probably a good place for this to live in.

jackzampolin: I'm super OK with not duplicating code. I didn't want to import the SDK into tendermint tho... How should I move forward here?

Liamsi: I think we should move those methods into libs/common together with their tests. Whenever the sdk updates to a Tendermint Version including this PR, they could delete the code from the SDK.

Liamsi: https://github.com/tendermint/tendermint/pull/3650

melekes:
What's the rationale behind this change?

jackzampolin:
@melekes During the last network hardfork we ran into an issue where when exporting, and then generating the genesis file we were getting different hashes. This would sort the JSON keys by default and should help avoid a similar issue in the future.

libs/common/json_sort.go
```diff
@@ -0,0 +1,31 @@
+package common
```
melekes: I'd just call the file `json.go`. If you look at other files, you'll see they are named similarly - `int.go`, ...

types/genesis.go
```diff
@@ -47,7 +47,7 @@ func (genDoc *GenesisDoc) SaveAs(file string) error {
 	if err != nil {
 		return err
 	}
-	return cmn.WriteFile(file, genDocBytes, 0644)
+	return cmn.WriteFile(file, cmn.MustSortJSON(genDocBytes), 0644)
```
melekes: I think we should make this explicit by adding `sort bool` parameter.

melekes: Also, "SaveAs is a utility method for saving GenensisDoc as a JSON file." comment needs to be updated

melekes:
> During the last network hardfork we ran into an issue where when exporting, and then generating the genesis file we were getting different hashes.

Because you did not sort `app_state`? I don't understand. The order should always be the same for `GenesisDoc` struct keys when encoded as json:

```
genesis_time
chain_id
consensus_params
validators
app_hash
app_state
```

PR #3641, opened 2019-05-08
abci/example: use base64 for update validator set

needkane:
reference : https://github.com/tendermint/tendermint/blob/master/abci/example/kvstore/README.md
"val:pubkey1/power1,addr2/power2,addr3/power3"
There is a problem, pubkey contains forward slash("/").

* [x] Wrote tests
curl -s 'localhost:26657validators'
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    "block_height": "91",
    "validators": [
      {
        "address": "75F6B901A666ABDB0AB12BDB5D4B1CF8F512997B",
        "pub_key": {
          "type": "tendermint/PubKeyEd25519",
          "value": "Vy3mgS0jodu/5zbs0dz99DYTgq/8vOHpFMsjxZSJtqw="
        },
        "voting_power": **"10",**
        "proposer_priority": "0"
      }
    ]
  }
}

if I want update validator voting power,but pubkey contains forward slash so that split  result is wrong.
curl -s 'localhost:26657/broadcast_tx_commit?tx="val:Vy3mgS0jodu/5zbs0dz99DYTgq/8vOHpFMsjxZSJtqw=/6"'

**My modify:**
curl -s 'localhost:26657/broadcast_tx_commit?tx="val:Vy3mgS0jodu/5zbs0dz99DYTgq/8vOHpFMsjxZSJtqw=!6"'
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    "check_tx": {
      "gasWanted": "1"
    },
    "deliver_tx": {},
    "hash": "3463B10BC220888CEA883761B33DE22F3C186C21DC5BBDB1E401F6CC5ACBD6A3",
    "height": "11"
  }
}

 Pubkey default display use base64 encode by rpc(localhost:26657/validators) or read configure file(~/.tendermint/config/priv_validator_key.json)， but update validator set function use hex.

https://github.com/tendermint/tendermint/issues/3659


ebuchman:
Would you mind opening a short issue to describe the need for this change? Also need to update the kvstore README where this tx type is documented (is it documented anywhere else that would need changing? )

needkane:
> 
> 
> Would you mind opening a short issue to describe the need for this change? Also need to update the kvstore README where this tx type is documented (is it documented anywhere else that would need changing? )

https://github.com/tendermint/tendermint/issues/3659

melekes:
> if I want update validator voting power,but pubkey contains forward slash so that split result is wrong. curl -s 'localhost:26657/broadcast_tx_commit?tx="val:Vy3mgS0jodu/5zbs0dz99DYTgq/8vOHpFMsjxZSJtqw=/6"'

Aren't we expecting hex-encoded pubkey? i.e.

`Vy3mgS0jodu/5zbs0dz99DYTgq/8vOHpFMsjxZSJtqw=` will result in `5679336D6753306A6F64752F357A627330647A393944595467712F38764F4870464D736A785A534A7471773D`

It's not documented, but from the looks of it we do.

ebuchman:
> Aren't we expecting hex-encoded pubkey? i.e.

We are, but the problem is that we don't expose the pubkey as hex anywhere, so users would have to manually convert it, which is kind of a pain. I think we do want the input format here to match what we use in RPC. Since we probably don't want to change the RPC (large breaking change, inconsistent with previous decisions to use base64 there), we should probably change this app to take base64 input.

But since this is technically breaking, we'll save it for v0.32. Thanks!

needkane:
> 
> 
> > if I want update validator voting power,but pubkey contains forward slash so that split result is wrong. curl -s 'localhost:26657/broadcast_tx_commit?tx="val:Vy3mgS0jodu/5zbs0dz99DYTgq/8vOHpFMsjxZSJtqw=/6"'
> 
> Aren't we expecting hex-encoded pubkey? i.e.
> 
> `Vy3mgS0jodu/5zbs0dz99DYTgq/8vOHpFMsjxZSJtqw=` will result in `5679336D6753306A6F64752F357A627330647A393944595467712F38764F4870464D736A785A534A7471773D`
> 
> It's not documented, but from the looks of it we do.

Problem is default use base64 （rpc result and priv）

needkane:
> 
> 
> > Aren't we expecting hex-encoded pubkey? i.e.
> 
> We are, but the problem is that we don't expose the pubkey as hex anywhere, so users would have to manually convert it, which is kind of a pain. I think we do want the input format here to match what we use in RPC. Since we probably don't want to change the RPC (large breaking change, inconsistent with previous decisions to use base64 there), we should probably change this app to take base64 input.
> 
> But since this is technically breaking, we'll save it for v0.32. Thanks!

Do you have any plan, maybe I can coding it. And this PR has solved it, use another format (pubkey1!power1,pubkey2!power2)

melekes:
@needkane do you mind updating https://github.com/tendermint/tendermint/blob/develop/abci/example/kvstore/README.md ?

needkane:
> 
> 
> @needkane do you mind updating https://github.com/tendermint/tendermint/blob/develop/abci/example/kvstore/README.md ?

done

abci/example/kvstore/persistent_kvstore.go
```diff
@@ -129,33 +129,34 @@ func (app *PersistentKVStoreApplication) Validators() (validators []types.Valida
...
+	return []byte(fmt.Sprintf("val:%s!%d", pubStr, power))
 }
 
 func isValidatorTx(tx []byte) bool {
 	return strings.HasPrefix(string(tx), ValidatorSetChangePrefix)
 }
 
-// format is "val:pubkey/power"
+// format is "val:pubkey!power"
 // pubkey is raw 32-byte ed25519 key
```
melekes: this is not true, right? it should be base64 encoded

abci/example/kvstore/README.md
```diff
@@ -22,7 +22,7 @@ and the Handshake allows any necessary blocks to be replayed.
 Validator set changes are effected using the following transaction format:
 
 ```
-val:pubkey1/power1,addr2/power2,addr3/power3"
+"val:pubkey1!power1,pubkey2!power2,pubkey3!power3"
 ```
 
 where `power1` is the new voting power for the validator with `pubkey1` (possibly a new one).
```
melekes:
```suggestion
where `pubkeyN` is a base64 encoded raw pubkey and `powerN` is a new voting power for the validator with `pubkeyN` (possibly a new one). To remove a validator from the validator set, set power to `0`.
```

abci/example/kvstore/persistent_kvstore.go
melekes:
```suggestion
// pubkey is a base64 encoded 32-byte ed25519 key
```

melekes:
Can you also add 

```
- [abci/examples] \#3659 Change validator update tx format (incl. expected pubkey format, which is base64 now) (@needkane)
```

to `CHANGELOG_PENDING.md` under IMPROVEMENTS section?

abci/example/kvstore/README.md
needkane: "Validators can be removed by setting their power to `0`"  already exist

Issue #2905, opened 2018-11-24
Missing documentation: non-unique tag keys


feldgendler:
**Tendermint version**: 0.22.8

Sorry for not following the bug template. This is a documentation bug, so most of that is irrelevant here.

`DeliverTx` can return, for a single transaction, several tags with the same key, since `Tags` is a list, not a map. The documentation is silent on whether it's allowed, and what will happen.

My current understanding from reading the code is that it works, and causes the transaction to be searchable using `TxSearch("key='value'", ...)` with either value. However, I'm not sure about ranges for numeric tags, and about whether subscriptions are going to work correctly.

I would like to use this if it's possible, but only if it's a documented feature rather than an unindended side effect.

On the other hand, if it _is_ an unintended side effect, then duplicate tags should better be prohibited. The best way to do that is to define the tag set as a map rather than a slice.

melekes:
We do not support duplicated keys right now. And you're right. We could've done a better job documenting this.

Refs #1859
Possible duplicate of #1385

feldgendler:
In this case, is there a reason why `Tags` in `ResponseDeliverTx` should be a list rather than a map?

ebuchman:
Primarily because maps are not ordered, and we need to use deterministic structures that can be hashed consistently on all nodes.

We have been considering changing the list to a list of lists. We can also consider a map, and use a deterministic ordering, eg. alphabetical sorting by key. 

This is still an open design question. Obviously things can be improved here, so we'd love more input. See https://github.com/tendermint/tendermint/issues/1859 and [ADR-021](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-021-abci-events.md) for an initial write-up of the list-of-list idea.

ValarDragon:
> Primarily because maps are not ordered, and we need to use deterministic structures that can be hashed consistently on all nodes.

hashmaps aren't ordered, tree maps are. We could use a tree map.

feldgendler:
Unless a single transaction has thousands of tags, the differences in efficiency between different kinds of map-like structures shouldn't matter.

melekes:
Done in https://github.com/tendermint/tendermint/pull/3643

We now support non-unique key tags through something called `Events`. Please refer to 
https://github.com/alexanderbez/tendermint/blob/91aab82e1c1b2d3cb185a30127277606ed6a2d66/docs/spec/abci/abci.md#events for details (later will be available at https://tendermint.com/docs/spec/abci/abci.html#events)

Issue #1385, opened 2018-03-28
Subscribing to tx event works for multiple kvpairs using same key


ethanfrey:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):

Feature Request

**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):

Last tested on rpc bugfix branch
`0.17.0-34c800c1`
(but also with 0.16.1 and 0.17.1)

**ABCI app** (name for built-in, URL for self-written if it's publicly available):

https://github.com/confio/weave - mycoind

**Use Case**

I have a wallet and want to be notified of changes, so I can subscribe to all tx that affect the balance.

Ideally, I would do something like `tm.event = 'Tx' AND wallet='<my hex address>'`. However, this would require the transaction to create two events `wallet=<sender hex address>` and `wallet=<recipient hex address>`. This is currently not supported.

A previous release of cosmos-sdk:basecoind resolved this by using one tag for sender and one for recipient. This is a temporary work-around, but brings up two issues. First, because queries do not support OR, we need to make two separate subscriptions (`sender=<my hex address>` and `recipient=<my hex address>`). These both will make separate callbacks, and when I send some cash to myself it will show up twice. (This is a bit annoying, ask @faboweb and his desire for the txhash on subscriptions to detect duplicates, now fixed #1367).

A bigger issue is multi-account sends. `basecoind` is set up to allow many input accounts and many output accounts, similar to the bitcoin model, which allows for secure swaps in one tx with multiple signatures. However, if there ever was a transaction with two senders, the above work-around wouldn't work and it is impossible to subscribe to all multi-output transactions that affect your account.

Also, please note that the `/search_tx` functionality currently supports the same key showing up on a given tx multiple times with different values, and searching for any of them will match the tx. Making this change will maintain consistency between the search and subscribe interfaces.

**Proposed Changes**

Look at https://github.com/tendermint/tendermint/blob/develop/types/event_bus.go#L88-L120

`func (b *EventBus) PublishEventTx(event EventDataTx) error` builds up a `map[string]interface{}` of all the tags. The contents are currently either `string` or `int64` (only for height). (Actually also an unmatchable `[]byte` but that should be fixed in another issue).

I would propose another case `[]string`, which is used on all `event.Result.Tags`. When matching, the subscription will match if the query matches any of the strings in the slice. We still require multiple kvpairs and do not complicate that interface, we only aggragate them when building the map in `PublishEventTx`.

[...]

Issue #1040, opened 2017-12-30
consensus: TestWALCrash non-deterministic failure


ebuchman:
possibly due to `go sendTxs` routines and the indeterminate amount of txs sent. 

let's try to take advantage of the fact that mempool is an interface to preload the right deterministic number of transactions to satisfy the test cases.

melekes:
Can't replicate anymore. 

> preload the right deterministic number of transactions

**we do that now**, but I wonder if instead we should use a non-deterministic number. It can help uncover bugs (if any?) connected to number of transactions

Liamsi:
Looks like this is back? 
https://circleci.com/gh/tendermint/tendermint/52571#tests/containers/3

Liamsi:
[build_52571_step_105_container_3.txt.zip](https://github.com/tendermint/tendermint/files/3022551/build_52571_step_105_container_3.txt.zip)

jackzampolin:
Looks like this is still an issue?

tessr:
Gonna close this for now, but will reopen if we see it again!

PR #3643, opened 2019-05-08
abci: Refactor tagigng events using list of lists

alexanderbez:
This PR introduces a fundamental breaking change to the structure of ABCI response and tx tags and the way they're processed. Namely, the SDK can support more complex and aggregated events for distribution and slashing. In addition, block responses can include duplicate keys in events.

* Implement new `Event` type. An event has a `type` and a list of KV pairs (ie. list-of-lists). Typical events may look like:

```json
"rewards": [{"amount": "5000uatom", "validator": "...", "recipient": "..."}]
"sender": [{"address": "...", "balance": "100uatom"}]
```

The events are indexed by `{even.type}.{even.attribute[i].key}/...`. In this case a client would subscribe or query for `rewards.recipient='...'`

* ABCI response types and related types now include `Events []Event` instead of `Tags []cmn.KVPair`.
* PubSub logic now publishes/matches against `map[string][]string` instead of `map[string]string` to support duplicate keys in response events (from #1385). A match is successful if the value is found in the slice of strings.

closes: #1859 
closes: #2905

----

* [ ] Wrote tests


libs/pubsub/pubsub.go
```diff
@@ -392,18 +392,18 @@ func (state *state) removeAll(reason error) {
 	}
 }
 
-func (state *state) send(msg interface{}, tags map[string]string) {
+func (state *state) send(msg interface{}, events map[string][]string) {
 	for qStr, clientSubscriptions := range state.subscriptions {
 		q := state.queries[qStr].q
-		if q.Matches(tags) {
+		if q.Matches(events) {
```
golangcibot: cannot use events (variable of type map[string][]string) as map[string]string value in argument to q.Matches (from `typecheck`)

types/event_bus.go
```diff
@@ -90,51 +91,58 @@ func (b *EventBus) UnsubscribeAll(ctx context.Context, subscriber string) error
 func (b *EventBus) Publish(eventType string, eventData TMEventData) error {
 	// no explicit deadline for publishing events
 	ctx := context.Background()
-	b.pubsub.PublishWithTags(ctx, eventData, map[string]string{EventTypeKey: eventType})
+	_ = b.pubsub.PublishWithEvents(ctx, eventData, map[string][]string{EventTypeKey: []string{eventType}})
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

types/event_bus.go
```diff
@@ -90,51 +91,58 @@ func (b *EventBus) UnsubscribeAll(ctx context.Context, subscriber string) error
...
-		if len(tag.Key) == 0 {
-			logger.Debug("Got tag with an empty key (skipping)", "tag", tag)
+		if len(event.Type) == 0 {
+			logger.Debug("Got tag with an empty type (skipping)", "event", event)
 			continue
 		}
-		result[string(tag.Key)] = string(tag.Value)
+
+		for _, attr := range event.Attributes {
+			compositeKey := fmt.Sprintf("%s.%s", event.Type, string(attr.Key))
```
alexanderbez: We should probably skip here too if the key is empty?

types/event_bus.go
```diff
@@ -90,51 +91,58 @@ func (b *EventBus) UnsubscribeAll(ctx context.Context, subscriber string) error
...
+// map of stringified events where each key is composed of the event
+// type and each of the event's attributes keys in the form of
+// "{event.Type}.{attribute.Key}" and the value is each attribute's value.
+func (b *EventBus) validateAndStringifyEvents(events []types.Event, logger log.Logger) map[string][]string {
+	result := make(map[string][]string)
+	for _, event := range events {
 		// basic validation
-		if len(tag.Key) == 0 {
-			logger.Debug("Got tag with an empty key (skipping)", "tag", tag)
+		if len(event.Type) == 0 {
```
melekes: don't we need to validate attribute.key as well?

state/txindex/kv/kv.go
```diff
@@ -107,17 +116,26 @@ func (txi *TxIndex) AddBatch(b *txindex.Batch) error {
...
 	hash := result.Tx.Hash()
 
 	// index tx by tags
-	for _, tag := range result.Result.Tags {
-		if txi.indexAllTags || cmn.StringInSlice(string(tag.Key), txi.tagsToIndex) {
-			b.Set(keyForTag(tag, result), hash)
+	for _, event := range result.Result.Events {
+		// only index events with a non-empty type
+		for _, attr := range event.Attributes {
+			if len(event.Type) > 0 {
```
melekes: same here

types/event_bus.go
alexanderbez: Yes, I'll amend.

state/txindex/kv/kv.go
alexanderbez: Updated.

state/txindex/kv/kv_test.go
```diff
@@ -21,7 +21,7 @@ func TestTxIndex(t *testing.T) {
 	indexer := NewTxIndex(db.NewMemDB())
 
 	tx := types.Tx("HELLO WORLD")
-	txResult := &types.TxResult{1, 0, tx, abci.ResponseDeliverTx{Data: []byte{0}, Code: abci.CodeTypeOK, Log: "", Tags: nil}}
+	txResult := &types.TxResult{1, 0, tx, abci.ResponseDeliverTx{Data: []byte{0}, Code: abci.CodeTypeOK, Log: "", Events: nil}}
```
golangcibot: composites: `github.com/tendermint/tendermint/types.TxResult` composite literal uses unkeyed fields (from `govet`)

state/txindex/kv/kv_test.go
```diff
@@ -36,7 +36,7 @@ func TestTxIndex(t *testing.T) {
 	assert.Equal(t, txResult, loadedTxResult)
 
 	tx2 := types.Tx("BYE BYE WORLD")
-	txResult2 := &types.TxResult{1, 0, tx2, abci.ResponseDeliverTx{Data: []byte{0}, Code: abci.CodeTypeOK, Log: "", Tags: nil}}
+	txResult2 := &types.TxResult{1, 0, tx2, abci.ResponseDeliverTx{Data: []byte{0}, Code: abci.CodeTypeOK, Log: "", Events: nil}}
```
golangcibot: composites: `github.com/tendermint/tendermint/types.TxResult` composite literal uses unkeyed fields (from `govet`)

docs/spec/abci/abci.md
```diff
@@ -38,20 +38,30 @@ Finally, `Query`, `CheckTx`, and `DeliverTx` include a `Codespace string`, whose
...
+Events can be used to index transactions and blocks according to what happened
+during their execution. Note that the set of events returned for a block from
 `BeginBlock` and `EndBlock` are merged. In case both methods return the same
 tag, only the value defined in `EndBlock` is used.
 
-Keys and values in tags must be UTF-8 encoded strings (e.g.
-"account.owner": "Bob", "balance": "100.0",
-"time": "2018-01-02T12:30:00Z")
+Each event has a `type` which should be unique and is meant to categorize the
+series of events for a particular `Response*` or tx. Keys and values in tags
```
fedekunze:
```suggestion
series of events for a particular `Response*` or tx. Keys and values in events
```

alexanderbez:
Looks like CI needs to be kicked again. Last commit was only a doc change. Not sure but there is probably some unrelated non-determinism. @Liamsi

Liamsi:
> Not sure but there is probably some unrelated non-determinism. @Liamsi

Yes, it was this issue `TestWALCrash`: https://github.com/tendermint/tendermint/issues/1040

restarting the build.

abci/types/types.proto
```diff
@@ -212,7 +212,7 @@ message ConsensusParams {
   ValidatorParams validator = 3;
 }
 
-// BlockParams contains limits on the block size.
+// BlockParams contains limits on the block size and timestamp.
```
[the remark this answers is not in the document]
alexanderbez: Weird. I don't remember making this change?

docs/spec/abci/abci.md
```diff
@@ -38,20 +38,30 @@ Finally, `Query`, `CheckTx`, and `DeliverTx` include a `Codespace string`, whose
...
-during their execution. Note that the set of tags returned for a block from
+Events can be used to index transactions and blocks according to what happened
+during their execution. Note that the set of events returned for a block from
 `BeginBlock` and `EndBlock` are merged. In case both methods return the same
 tag, only the value defined in `EndBlock` is used.
 
-Keys and values in tags must be UTF-8 encoded strings (e.g.
-"account.owner": "Bob", "balance": "100.0",
-"time": "2018-01-02T12:30:00Z")
+Each event has a `type` which should be unique and is meant to categorize the
```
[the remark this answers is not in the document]
alexanderbez: Ok, so technically you can have multiple events with the same `type`. I guess what I was trying to state and motivate here is that it is generally unadvisable to do so. ie. it wouldn't make sense to have two or more events with type `rewards` when you can just include all the attributes under a single `rewards` event.

state/txindex/kv/kv.go
```diff
@@ -75,18 +75,30 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
 	defer storeBatch.Close()
 
 	for _, result := range b.Ops {
 		hash := result.Tx.Hash()
 
-		// index tx by tags
-		for _, tag := range result.Result.Tags {
-			if txi.indexAllTags || cmn.StringInSlice(string(tag.Key), txi.tagsToIndex) {
-				storeBatch.Set(keyForTag(tag, result), hash)
+		// index tx by events
```
ethanfrey: Looks good, and seems that AND will work over different types as well, correct?

I would love to see an high-level integration test of this (or the pubsub system in general), adding a batch of transactions and subscribing to events (parsed from query strings) to see if they match. This will be the interface uses receive, good to make sure it is useable (especially the mapping from go types to query strings)

docs/spec/abci/abci.md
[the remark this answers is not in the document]
ebuchman: > you can just include all the attributes under a single rewards event.

Not sure I follow. Isn't the whole point of this to have eg. discrete withdraw events where each has a different set of attributes, eg. one is for DelegatorA (where attributes relate to DelegatorA and the amount they withdrew) and a different event, also of the same type, but for DelegatorB, with attributes for DelegatorB ?

libs/pubsub/pubsub.go
```diff
@@ -63,7 +63,7 @@ var (
 
 // Query defines an interface for a query to be used for subscribing.
 type Query interface {
-	Matches(tags map[string]string) bool
+	Matches(events map[string][]string) bool
```
ebuchman: Would benefit from documentation as to why this `map[string][]string` differs from what's in the ABCI format, which is essentially `[]map[string]string`. 

The latter case allows event types to repeat themselves, with the same set of keys and different values.

The former case consolidates all values for  a given `type+key` into a single list.

libs/pubsub/query/query.go
```diff
@@ -148,12 +148,14 @@ func (q *Query) Conditions() []Condition {
 	return conditions
 }
 
-// Matches returns true if the query matches the given set of tags, false otherwise.
+// Matches returns true if the query matches against any event in the given set
+// of events, false otherwise. For each event, a match exists if the query is
+// matches against *any* value in a slice of values.
```
ebuchman:
```suggestion
// matched against *any* value in a slice of values.
```

libs/pubsub/query/query.go
```diff
@@ -232,105 +235,119 @@ func (q *Query) Matches(tags map[string]string) bool {
...
-// match returns true if the given triplet (tag, operator, operand) matches any tag.
+// match returns true if the given triplet (tag, operator, operand) matches any
+// value in an event for that key.
 //
-// First, it looks up the tag in tags and if it finds one, tries to compare the
-// value from it to the operand using the operator.
+// First, it looks up the key in the events and if it finds one, tries to compare
+// all the values from it to the operand using the operator.
 //
 // "tx.gas", "=", "7", { "tx.gas": 7, "tx.ID": "4AE393495334" }
```
ebuchman: The final argument on this line should be updated for the new format

rpc/core/events.go
```diff
@@ -22,19 +22,20 @@ import (
...
 // Tendermint provides a few predefined keys: tm.event, tx.hash and tx.height.
-// Note for transactions, you can define additional keys by providing tags with
+// Note for transactions, you can define additional keys by providing events with
 // DeliverTx response.
 //
 //		DeliverTx{
-//			Tags: []*KVPair{
-//				"agent.name": "K",
+//			Events: []Event{
+//				{"agent": {{"name": "K"}},
```
ebuchman: Don't we have a `type` field in the event now? Not clear how this maps to the ABCI struct.

state/txindex/kv/kv.go
```diff
@@ -75,18 +75,30 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+		for _, event := range result.Result.Events {
+			// only index events with a non-empty type
+			if len(event.Type) > 0 {
```
ebuchman: We could do:

```
if len(event.Type) == 0{
 continue
}
```

And then avoid everything else needing the extra indent

state/txindex/kv/kv.go
```diff
@@ -75,18 +75,30 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+				for _, attr := range event.Attributes {
+					if len(attr.Key) > 0 {
```
ebuchman: Same, just need to ensure we continue the inner loop!

state/txindex/kv/kv.go
```diff
@@ -75,18 +75,30 @@ func (txi *TxIndex) Get(hash []byte) (*types.TxResult, error) {
...
+						compositeTag := fmt.Sprintf("%s.%s", event.Type, string(attr.Key))
+						if txi.indexAllTags || cmn.StringInSlice(compositeTag, txi.tagsToIndex) {
+							storeBatch.Set(keyForEvent(compositeTag, attr.Value, result), hash)
```
ebuchman: Ah. This means the change is breaking for the indexer. Not sure if there's a way to avoid that (prob not), which means we may need to re-index the tx db on startup or else somehow prevent queries against the older data ... (probably want the former)

state/txindex/kv/kv.go
```diff
@@ -107,17 +119,28 @@ func (txi *TxIndex) AddBatch(b *txindex.Batch) error {
...
 
 	hash := result.Tx.Hash()
 
 	// index tx by tags
-	for _, tag := range result.Result.Tags {
-		if txi.indexAllTags || cmn.StringInSlice(string(tag.Key), txi.tagsToIndex) {
-			b.Set(keyForTag(tag, result), hash)
+	for _, event := range result.Result.Events {
+		// only index events with a non-empty type
+		if len(event.Type) > 0 {
```
ebuchman: same re checking for 0 and using continue instead of big indents

ebuchman: also it seems this logic is exactly repeated. can we dedup?

state/txindex/kv/kv_test.go
```diff
@@ -51,10 +67,10 @@ func TestTxSearch(t *testing.T) {
 	allowedTags := []string{"account.number", "account.owner", "account.date"}
 	indexer := NewTxIndex(db.NewMemDB(), IndexTags(allowedTags))
 
-	txResult := txResultWithTags([]cmn.KVPair{
-		{Key: []byte("account.number"), Value: []byte("1")},
-		{Key: []byte("account.owner"), Value: []byte("Ivan")},
-		{Key: []byte("not_allowed"), Value: []byte("Vlad")},
+	txResult := txResultWithEvents([]abci.Event{
```
ebuchman: I feel like maybe we should preserve the original behaviour here of having multiple attributes for a single event (eg. owner, number), and then having multiple events with distinct attributes (eg. (Ivan, 1) and (Gru, 2).

types/event_bus.go
```diff
@@ -90,51 +91,62 @@ func (b *EventBus) UnsubscribeAll(ctx context.Context, subscriber string) error
 func (b *EventBus) Publish(eventType string, eventData TMEventData) error {
 	// no explicit deadline for publishing events
 	ctx := context.Background()
-	b.pubsub.PublishWithTags(ctx, eventData, map[string]string{EventTypeKey: eventType})
+	_ = b.pubsub.PublishWithEvents(ctx, eventData, map[string][]string{EventTypeKey: {eventType}})
```
ebuchman: Huh should we really be ignoring the err here ?

types/event_bus.go
```diff
@@ -90,51 +91,62 @@ func (b *EventBus) UnsubscribeAll(ctx context.Context, subscriber string) error
...
+	events := b.validateAndStringifyEvents(resultEvents, b.Logger.With("block", data.Block.StringShort()))
 
-	// add predefined tags
-	logIfTagExists(EventTypeKey, tags, b.Logger)
-	tags[EventTypeKey] = EventNewBlock
+	// add predefined new block event
+	events[EventTypeKey] = append(events[EventTypeKey], EventNewBlock)
 
-	b.pubsub.PublishWithTags(ctx, data, tags)
+	_ = b.pubsub.PublishWithEvents(ctx, data, events)
```
ebuchman: Not sure if there's a reason we weren't checking errors before but probably should be ?

docs/spec/abci/abci.md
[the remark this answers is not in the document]
alexanderbez: @ebuchman yes -- I believe we're talking about the same thing here.

What I'm saying is the following:

It's best to have events such as this because they're all aggregated under the same `type`:

```json
"events": [
  {
    "type": "rewards",
    "attributes": [ 
        {"delegator": "delegatorA", "amount": "...", "another_attribute": "..."},
        {"delegator": "delegatorB", "amount": "...", "another_attribute": "..."},
    ]
  }
]
```


Opposed to this:


```json
"events": [
  {
    "type": "rewards",
    "attributes": [ 
        {"delegator": "delegatorA", "amount": "...", "another_attribute": "..."}
    ]
  },
  {
    "type": "rewards",
    "attributes": [ 
        {"delegator": "delegatorB"},
        {"amount": "..."},
        {"another_attribute": "..."},
    ]
  }
]
```

Does that make sense? Or am I thinking about the UX in the wrong way?

libs/pubsub/query/query_test.go
```diff
@@ -19,30 +19,30 @@ func TestMatches(t *testing.T) {
...
+		{"peaches.kg < 4", map[string][]string{"peaches.kg": {"5"}}, false, false},
+
+		{"tx.date > DATE 2017-01-01", map[string][]string{"tx.date": {time.Now().Format(query.DateLayout)}}, false, true},
+		{"tx.date = DATE 2017-01-01", map[string][]string{"tx.date": {txDate}}, false, true},
+		{"tx.date = DATE 2018-01-01", map[string][]string{"tx.date": {txDate}}, false, false},
+
+		{"tx.time >= TIME 2013-05-03T14:45:00Z", map[string][]string{"tx.time": {time.Now().Format(query.TimeLayout)}}, false, true},
+		{"tx.time = TIME 2013-05-03T14:45:00Z", map[string][]string{"tx.time": {txTime}}, false, false},
+
+		{"abci.owner.name CONTAINS 'Igor'", map[string][]string{"abci.owner.name": {"Igor,Ivan"}}, false, true},
```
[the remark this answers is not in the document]
alexanderbez: Updated 👍

libs/pubsub/pubsub_test.go
```diff
@@ -312,7 +312,7 @@ func benchmarkNClients(n int, b *testing.B) {
 	b.ReportAllocs()
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		s.PublishWithTags(ctx, "Gamora", map[string]string{"abci.Account.Owner": "Ivan", "abci.Invoices.Number": string(i)})
+		s.PublishWithEvents(ctx, "Gamora", map[string][]string{"abci.Account.Owner": {"Ivan"}, "abci.Invoices.Number": {string(i)}})
```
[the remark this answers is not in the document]
alexanderbez: The suggestion you've tagged this to is in regards to a benchmark, so I just implemented a new unit test `TestSubscribeDuplicateKeys`.

libs/pubsub/query/query_test.go
```diff
@@ -19,30 +19,30 @@ func TestMatches(t *testing.T) {
...
-		{"tx.date > DATE 2017-01-01", map[string]string{"tx.date": time.Now().Format(query.DateLayout)}, false, true},
-		{"tx.date = DATE 2017-01-01", map[string]string{"tx.date": txDate}, false, true},
-		{"tx.date = DATE 2018-01-01", map[string]string{"tx.date": txDate}, false, false},
-
-		{"tx.time >= TIME 2013-05-03T14:45:00Z", map[string]string{"tx.time": time.Now().Format(query.TimeLayout)}, false, true},
-		{"tx.time = TIME 2013-05-03T14:45:00Z", map[string]string{"tx.time": txTime}, false, false},
-
-		{"abci.owner.name CONTAINS 'Igor'", map[string]string{"abci.owner.name": "Igor,Ivan"}, false, true},
-		{"abci.owner.name CONTAINS 'Igor'", map[string]string{"abci.owner.name": "Pavel,Ivan"}, false, false},
+		{"tm.events.type='NewBlock'", map[string][]string{"tm.events.type": {"NewBlock"}}, false, true},
```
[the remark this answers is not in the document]
alexanderbez: Agreed. Updated.

rpc/core/events.go
```diff
@@ -22,19 +22,20 @@ import (
...
 //
 // Tendermint provides a few predefined keys: tm.event, tx.hash and tx.height.
-// Note for transactions, you can define additional keys by providing tags with
+// Note for transactions, you can define additional keys by providing events with
 // DeliverTx response.
 //
 //		DeliverTx{
-//			Tags: []*KVPair{
-//				"agent.name": "K",
+//			Events: []Event{
```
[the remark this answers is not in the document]
alexanderbez: I've amended this slightly. Can you lmk your thoughts? Any examples or queries you think are missing or should be adjusted?

state/txindex/kv/kv.go
alexanderbez: > Looks good, and seems that AND will work over different types as well, correct?

Yes.

As far as integration tests, is there anything in `types/event_bus_test.go` or `libs/pubsub/pubsub_test.go` that you believe doesn't cover what you expect it to? Both of those closely resemble an integration test(s) in that they create queries (from strings) and subscribe or query for events.

libs/pubsub/pubsub.go
alexanderbez: Hmmm, how does it differ from the ABCI format? The ABCI format is a list of KV pairs, which is essentially a list of values for each given key in `attributes`.

In either case, I've updated the godoc to reflect this change 👍

libs/pubsub/query/query.go
alexanderbez: Updated.

state/txindex/kv/kv.go
alexanderbez: Sure!

types/event_bus.go
alexanderbez: We were before! I simply added a blank identifier for linting reasons. But since this function signature returns an error already, I'll handle it and return it to the caller.

state/txindex/kv/kv.go
alexanderbez: I'm not necessarily sure that old queries will break or no longer work -- I think they should, but I have yet to test this. The key format is still `key/value/height/index`.

state/txindex/kv/kv_test.go
alexanderbez: Sure, you believe that will provide a better UX?

docs/spec/abci/abci.md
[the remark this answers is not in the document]
alexanderbez: I believe I understand what you're referring to now. I'll update the doc(s).

rpc/core/events.go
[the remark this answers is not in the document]
alexanderbez: Updated

rpc/core/events.go
alexanderbez: Updated.

alexanderbez:
@ebuchman @ethanfrey I've updated various tests, implemented new ones, and updated docs. Mind giving this another look?

libs/pubsub/query/query_test.go
```diff
@@ -51,10 +61,12 @@ func TestMatches(t *testing.T) {
 			require.Nil(t, err)
 		}
 
+		require.NotNil(t, q, "Query '%s' should not be nil", tc.s)
```
melekes: do we have to check if q is nil? if later it will fail anyway when trying to access q.Matches method.

types/event_bus.go
```diff
@@ -90,52 +91,61 @@ func (b *EventBus) UnsubscribeAll(ctx context.Context, subscriber string) error
...
 	ctx := context.Background()
 
-	resultTags := append(data.ResultBeginBlock.Tags, data.ResultEndBlock.Tags...)
+	resultTags := append(data.ResultBeginBlock.Events, data.ResultEndBlock.Events...)
 	// TODO: Create StringShort method for Header and use it in logger.
-	tags := b.validateAndStringifyTags(resultTags, b.Logger.With("header", data.Header))
+	events := b.validateAndStringifyEvents(resultTags, b.Logger.With("header", data.Header))
 
-	// add predefined tags
-	logIfTagExists(EventTypeKey, tags, b.Logger)
```
melekes: why remove?

types/event_bus.go
```diff
@@ -153,19 +163,14 @@ func (b *EventBus) PublishEventTx(data EventDataTx) error {
 	// no explicit deadline for publishing events
 	ctx := context.Background()
 
-	tags := b.validateAndStringifyTags(data.Result.Tags, b.Logger.With("tx", data.Tx))
+	events := b.validateAndStringifyEvents(data.Result.Events, b.Logger.With("tx", data.Tx))
 
 	// add predefined tags
-	logIfTagExists(EventTypeKey, tags, b.Logger)
```
melekes: same here. this method allows user to see if some tag will be overwritten by Tendermint

libs/pubsub/query/query_test.go
alexanderbez: We don't technically have to, `q` may be nil, so I see it as no harm in checking since we can avoid a panic.

types/event_bus.go
alexanderbez: It was removed because the semantics changed.

Prior, we have had a `map[string]string` to represent tags/events. With this change set, we now have `map[string][]string`, so the notion of overwriting a key doesn't apply anymore as the value will simply just be appended.

types/event_bus.go
alexanderbez: See response [here](https://github.com/tendermint/tendermint/pull/3643#discussion_r291264222).

types/event_bus.go
melekes: Missed this somehow, sorry.

> Maybe we can still warn that a Tendermint reserved duplicate will be added?

don't think there's a need.

state/txindex/kv/kv.go
ebuchman: But if txs have already been indexed under the old format, then they won't be accessible under the new format. Ie if an existing node upgrades, won't queries for older heights fail since the key is now a composite?

alexanderbez: Hmmm, so just spitballing here, but I don't immediately see why queries wont work for older heights (pre-upgrade) as the format _structure_ remains unchanged -- on the key is changed.

Say for example in the context of the SDK/Gaia, we emit a tag `"sender": "cosmos0xxx"` and now is updated  to emit an equivalent event `"transfer": [{"sender": "cosmos0xxx"}]`, the old tag would be indexed under `sender/cosmos0xxx/H/I` and the new event would be indexed under `transfer.sender/cosmos0xxx/H/I`.

Now when we search via `*TxIndex#Search`, both queries `sender = "cosmos0xxx"` and `transfer.sender = "cosmos0xxx"` should work as we just call `startKeyForCondition` for the given `Condition` which in turn does a prefix lookup with the give bytes from `startKeyForCondition`.

Maybe I'm completely wrong here and nodes will have to resync...

melekes: > Hmmm, so just spitballing here, but I don't immediately see why queries wont work

Maybe we should add a small test to ensure?

alexanderbez: I've implemented a test `TestTxSearchDeprecatedIndexing` that indexes in the deprecated fashion which shows querying still works on old txs/keys.

crossle:
if use the github.com/tendermint/tendermint v0.32.3, parse the transaction will error `unknown field "tags" in types.ResponseDeliverTx` how to compatible？

melekes:
You will need to switch to a new format (i.e. rename `tags` to `events` and change the structure a bit)

```go
events := []types.Event{
		{
			Type: "app",
			Attributes: []cmn.KVPair{
				{Key: []byte("creator"), Value: []byte("Cosmoshi Netowoko")},
				{Key: []byte("key"), Value: key},
			},
		},
	}
```

PR #3644, opened 2019-05-08
p2p: Fix error logging for connection stop

defunctzombie:
This changeset fixes two types of false-positive errors occurring during
connection shutdown.

The first occurs when the process invokes FlushStop() or Stop() on a
connection. While the previous behavior did properly wait for the sendRoutine
to finish, it did not notify the recvRoutine that the connection was shutting
down. This would cause the recvRouting to receive and error when reading and
log this error. The changeset fixes this by notifying the recvRoutine that
the connection is shutting down.

The second occurs when the connection is terminated (gracefully) by the other side.
The recvRoutine would get an EOF error during the read, log it, and stop the connection
with an error. The changeset detects EOF and gracefully shuts down the connection.

---
I am not intimately familiar with idiomatic go practices so if there is a better approach for performing the signaling I did please let me know and I will update the PR. I used a channel buffer size of 1 to ensure the signal is received and non-blocking messaging to avoid breaking existing logic.

* [x] Updated all relevant documentation in docs (There is no documentation for this)
* [ ] Wrote tests (The tests would need to check log output - I could not find a good example of how to do this).


p2p/conn/connection.go
```diff
@@ -550,7 +559,24 @@ FOR_LOOP:
...
+			// connection is closed (likely by the other side)
+			// shut down our side
+			if err == io.EOF {
+				c.stopServices()
+				c.Stop()
+				break FOR_LOOP
+			}
+
+			// stopServices was invoked and we are shutting down
+			// receving is excpected to fail since we will close the connection
```
golangcibot: `receving` is a misspelling of `receiving` (from `misspell`)

defunctzombie: fixed

p2p/conn/connection.go
```diff
@@ -249,9 +261,6 @@ func (c *MConnection) FlushStop() {
 		// so we dont race on calling sendSomePacketMsgs
 		<-c.doneSendRoutine
 
-		// Send and flush all pending msgs.
-		// By now, IsRunning == false,
```
defunctzombie: I deleted this comment because it is false. In my testing `IsRunning` was still true - that is why I needed to add the additional signaling. Maybe a better fix is to identify why `IsRunning` is still true.

defunctzombie: I looked closer and the reason is that FlushStop is outside of the `Stop` base service routine - this the base does not yet know that it is being stopped.

defunctzombie:
Thoughts on this change? While cosmetic - it does help cleanup log output which is useful for node and validator operators that scrape logs for errors and anomalies.

p2p/conn/connection.go
```diff
@@ -220,16 +225,23 @@ func (c *MConnection) stopServices() (alreadyStopped bool) {
...
 	select {
 	case <-c.quitSendRoutine:
-		// already quit via FlushStop or OnStop
+		// already quit
 		return true
 	default:
 	}
 
+	// inform the recvRouting that we are shutting down
+	select {
```
ebuchman: does this need to be sent on here or can we just close it ?

p2p/conn/connection.go
```diff
@@ -550,7 +559,24 @@ FOR_LOOP:
...
+			// connection is closed (likely by the other side)
+			// shut down our side
+			if err == io.EOF {
+				c.stopServices()
+				c.Stop()
+				break FOR_LOOP
+			}
+
+			// stopServices was invoked and we are shutting down
+			// receiving is excpected to fail since we will close the connection
```
ebuchman: So this happens when err != io.EOF ? What is the error here?

p2p/conn/connection.go
defunctzombie: This needs to be sent here to inform the receive routine that it needs to stop. The is no mechanism in go (that I have found) to receive socket events therefore you need to roll your own notifications to the go routine for receive that was setup previously.

p2p/conn/connection.go
defunctzombie: This is to handle graceful shutdown via `stopServices()`. Go does not have a nice way to see that we have closed the socket - only when we go to read the socket and it fails do we need to see if we expected this failure or not. This logic queries the channel for the quit signal to notify us that we are expecting to exit and do not care about the error.

p2p/conn/connection.go
ebuchman: Oh, you can use `close(ch)` as a way to send on a channel one time, and to have many receivers be able to receive the signal - https://golang.org/ref/spec#Close

ebuchman: But does the signal need to be sent here, or can it be sent after the block of `.Stop()` calls? Calling `close(ch)` will send a signal on the channel.

p2p/conn/connection.go
```diff
@@ -550,7 +559,24 @@ FOR_LOOP:
 		var err error
 		_n, err = cdc.UnmarshalBinaryLengthPrefixedReader(c.bufConnReader, &packet, int64(c._maxPacketMsgSize))
 		c.recvMonitor.Update(int(_n))
+
 		if err != nil {
+			// connection is closed (likely by the other side)
+			// shut down our side
+			if err == io.EOF {
+				c.stopServices()
```
ebuchman: You shouldn't have to call this separately. `c.Stop()` calls `c.OnStop()` which calls `stopServices` for us.

defunctzombie: Removed.

p2p/conn/connection.go
defunctzombie: Ah - I did not know that about channels. I have removed this select that manually sent a bool in favor of closing the channel (already done lower in the same method). I have changed the channel to a `struct{}` type since nothing is sent on it anymore.

p2p/conn/connection.go
```diff
@@ -90,6 +90,9 @@ type MConnection struct {
 	quitSendRoutine chan struct{}
 	doneSendRoutine chan struct{}
 
+	// Closing quitRecvRouting will cause the recvRouting to evnetually quit.
```
melekes: evnetually -> eventually

p2p/conn/connection.go
```diff
@@ -249,9 +255,6 @@ func (c *MConnection) FlushStop() {
 		// so we dont race on calling sendSomePacketMsgs
 		<-c.doneSendRoutine
 
-		// Send and flush all pending msgs.
```
melekes: why remove `// Send and flush all pending msgs.`? looks like we're still flushing the messages

defunctzombie: Yea - I will put this back. The `IsRunning` comment is the part that was wrong. I accidentally removed this when I should not have.

melekes: Ok, thanks

melekes:
@defunctzombie will you have time to fix https://github.com/tendermint/tendermint/pull/3644#discussion_r291102287 and https://github.com/tendermint/tendermint/pull/3644#discussion_r291100481?

@ebuchman ok to merge?

p2p/conn/connection.go
melekes:
```suggestion
	// Closing quitRecvRouting will cause the recvRouting to eventually quit.
```

melekes:
Replaced with https://github.com/tendermint/tendermint/pull/3824

PR #3646, opened 2019-05-08
p2p/pex: use RemoteIP for peers who report unroutable IPs as their IP

defunctzombie:
A peer which is has a default configuration and is unable to detect its own ip will
    report an IP of 0.0.0.0. This IP address is unroutable and not helpful in establishing
    a stronger connectivity graph. Instead, we record the peer's RemoteIP which is routable
    and we hope accepting inbound connections for peering.

This change uses the RemoteIP for any unroutable address.

* [ ] Wrote tests


hendrikhofstadt:
You could try to check whether the given address is routable and use the remoteIP if it is not.

defunctzombie:
> You could try to check whether the given address is routable and use the remoteIP if it is not.

Good suggestion - I have updated the PR accordingly.

defunctzombie:
Any feedback on this from maintainers?

p2p/pex/pex_reactor.go
```diff
@@ -194,6 +194,12 @@ func (r *PEXReactor) AddPeer(p Peer) {
 			return
 		}
 
+		// If the node is reporting a non-routable address then we use the remote IP used in establishing the connection
+		// While this doesn't prevent adding bogus IPs, it does help allow us to grow the address book with routable peers
+		if !addr.Routable() {
```
melekes: I'm becoming increasingly worried about adding such `if` constructions. In two years, nobody will be able to understand all the intricacies of the code.

Can't we reject the peer if it reports non-routable IP? https://github.com/tendermint/tendermint/blob/5997e75c84f643a65cc411331e1f47201ebb5772/p2p/node_info.go#L122

ebuchman: Agree with the sentiment re such conditions but I don't think we can just reject peers for not knowing their own IP (which is quite common on p2p networks, especially for peers behind NATs).

Perhaps what we need to do is fix the address earlier on, like when we first unmarshal the NodeInfo for the peer ? Would it make sense to check if its routable as soon as we receive it, and if not update it to the remote addr? I'm not sure on first take if this will have unintended side effects.

defunctzombie: > Can't we reject the peer if it reports non-routable IP?

That approach would be undesirable for a few of the reasons @ebuchman has mentioned. Sockets already provide a nice mechanism for the receiver to identify the source without requiring the source to properly self-report its IP address.

I agree that this _fix_ is more of a band-aid and I too would prefer to set this properly earlier when first receiving the NodeInfo. If that is the desired direction I can make the change for review.

Orthogonal to this fix - I would possibly suggest ignoring or removing a node's self-reporting IP address feature. I see this feature as one that is more ripe for miss-use than benefit. But don't want the discussion around that clouding the fix that I believe should happen for nodes reporting non-routable addresses.

melekes: > Perhaps what we need to do is fix the address earlier on, like when we first unmarshal the NodeInfo for the peer

👍 

> I would possibly suggest ignoring or removing a node's self-reporting IP address feature. 

I'm in favor of removing, but I think we're using it in several places.

defunctzombie: https://github.com/tendermint/tendermint/blob/master/p2p/transport.go#L469-L506 Would this be the location where the peer node info is first unmarshalled or is there another or multiple locations?

melekes: This should be the only location. The correct place to add the check is probably https://github.com/tendermint/tendermint/blob/master/p2p/transport.go#L438

defunctzombie: @melekes I tried to port the change to the referenced area in transport.go; however NodeInfo is not mutable in the required way at that point in the code so I opted for a change in the handshake portion where the correct fields are mutable.

defunctzombie:
I've updated the changeset to perform the routability check inside of the handshake where NodeInfo is first received from the peer.

A note has been added to indicate that this does not strictly prevent unroutable addresses, but it does prevent "invalid" ones (like 0.0.0.0) from being set as the ListenAddr for NodeInfo.

I would probably advocate for removing the `addr.Routable()` condition and unconditionally setting the ListenAddr to the remote address - but that needs stronger consideration of the tradeoffs and side-effects given existing behavior.

p2p/pex/pex_reactor.go
melekes: > however NodeInfo is not mutable in the required way at that point in the code

it's not mutable, but still can be overwritten

`nodeInfo.ListenAddress =` should work

melekes: I'd prefer to keep handshake function clean and do not add any non-related functionality.

defunctzombie: `it's not mutable, but still can be overwritten` I believe I tried that and it doesn't work because the type within the upgrade is `NodeInfo` which does not know about a `ListenAddress` field. Would it be appropriate to cast it?

melekes:
Wait, in config.toml if you set `addr_book_strict` to `true` then you should not see non-routable addresses in your addrbook.

p2p/pex/pex_reactor.go
melekes: Oh, I see what you mean now.

melekes: > Would it be appropriate to cast it?

no

melekes:
Would https://github.com/tendermint/tendermint/pull/3516 be the better fix for the issue described here?

defunctzombie:
> Would #3516 be a better fix for the issue described here? @defunctzombie what do you think?

Based on a quick reading - I would say "not quite". Still seems like it relies on the peer to provide the address via node info rather than the receiver looking it up on the socket.

melekes:
@defunctzombie what are you actually trying to solve here? How exactly overriding ListenAddr with remote IP helps with "stronger connectivity graph"?

melekes:
@defunctzombie ping (https://github.com/tendermint/tendermint/pull/3646#issuecomment-508400879)

defunctzombie:
@melekes the idea is that instead of the default behavior of having a node self-report the IP address - the receiver will use the source IP of the node (found from their connection info) as the peer IP. This avoids spoofing IPs by nodes and rely on the actual on-wire connection info.

p2p/transport.go
```diff
@@ -537,6 +537,18 @@ func handshake(
...
+	addr, err := peerNodeInfo.NetAddress()
+	if err != nil {
+		return nil, err
+	}
+
+	// If the node is reporting a non-routable address then we use the remote IP from the connection as the ListenAddr.
+	// While this doesn't prevent adding bogus IPs, it does help ensure we use routable IPs for inbound node info.
+	// Note: If the remote address is not routable (i.e. private network) then ListenAddr will still be set to this private address.
+	if !addr.Routable() {
+		peerNodeInfo.ListenAddr = c.RemoteAddr().String()
```
erikgrinaker: Unfortunately, `RemoteAddr()` will include the arbitrarily chosen outbound port number of the remote IP, not the port number which the peer is (possibly) listening on. I believe this will wind up in the address book, and may be gossiped to other peers who will not be able to connect.

As a best effort we may discard the port number and fall back to the default Tendermint port number, but even this is somewhat brittle.

erikgrinaker:
I don't think this is the right approach to fixing this problem - see e.g. https://github.com/tendermint/tendermint/pull/3646#discussion_r380772361. A more suitable approach might involve improved auto-detection of advertised addresses, and filtering of non-routable addresses like `192.168.0.0/16` advertised to peers on other networks.

We are [planning](https://github.com/tendermint/tendermint/issues/4360) an overhaul of the P2P system later this year, which might involve a [migration to libp2p](https://github.com/tendermint/tendermint/issues/3696#988) and/or a [refactor of the P2P module](https://github.com/tendermint/tendermint/issues/2067). We will make sure these issues are addressed then.

PR #3647, opened 2019-05-08
p2p/pex: consult seeds in crawlPeersRoutine

defunctzombie:
This changeset alters the startup behavior for crawlPeersRoutine. Previously
the routine would crawl a random selection of peers on startup. For a
new seed node, there are no peers. As a result, new seed nodes are unable
to bootstrap themselves with a list of peers until another node with a list
of peers connects to the seed. If this node relies on the seed node for peers,
then the two will not discover more peers.

This changeset makes the startup behavior for crawlPeersRoutine connect to
any seed nodes. Upon connecting, a request for peers will be sent to the seed node
thus helping bootstrap our seed node.

* [x] Updated all relevant documentation in docs (no documentation that I have found)
* [x] Updated all code comments where relevant (might make sense to add some of the above description directly into the source)
* [ ] Wrote tests

### DIFF
--- a/p2p/pex/pex_reactor.go
+++ b/p2p/pex/pex_reactor.go
@@ -145,7 +145,7 @@ func (r *PEXReactor) OnStart() error {
 	if err != nil {
 		return err
 	} else if numOnline == 0 && r.book.Empty() {
-		return errors.New("Address book is empty and no seed nodes")
+		return errors.New("Address book is empty and couldn't resolve any seed nodes")
 	}
 
 	r.seedAddrs = seedAddrs
--- a/p2p/pex/pex_reactor.go
+++ b/p2p/pex/pex_reactor.go
@@ -145,7 +145,7 @@ func (r *PEXReactor) OnStart() error {
 	if err != nil {
 		return err
 	} else if numOnline == 0 && r.book.Empty() {
-		return errors.New("Address book is empty, and could not connect to any seed nodes")
+		return errors.New("Address book is empty and no seed nodes")
 	}
 
 	r.seedAddrs = seedAddrs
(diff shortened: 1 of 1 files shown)


p2p/pex/pex_reactor.go
```diff
@@ -145,7 +145,7 @@ func (r *PEXReactor) OnStart() error {
 	if err != nil {
 		return err
 	} else if numOnline == 0 && r.book.Empty() {
-		return errors.New("Address book is empty, and could not connect to any seed nodes")
+		return errors.New("Address book is empty and no seed nodes")
```
ebuchman: This error message could be more clear and better synchronized with the result of checkSeeds. 

For instance, checkSeeds could return -1 and a nil error to indicate there are no seeds, which isn't captured here. 

checkSeeds should probably do a better job of returning errors, and for instance return an error indicating there are no seeds rather than using the special case of -1.

defunctzombie: What would you prefer the error message to say?

Do you want me to make changes to return values for checkSeeds - largely I tried to minimize the diff but if you have a strategy/direction you want this to go I am happy to make the changes. I am not familiar with the proper way to return and handle different "types" of errors in Go or what you prefer for tendermint so just let me know.

ebuchman:
```suggestion
		return errors.New("Address book is empty and couldn't resolve any seed nodes")
```

ebuchman: No worries I realized we actually don't want to error on empty seeds list since that would cause the node not to start, which isn't what we want. So everything is fine. If you could just accept the one suggestion we can merge. Thanks for the fix!

defunctzombie:
@ebuchman I have accepted the suggestion.

PR #3657, opened 2019-05-11
Added votes to header + added secp256k1 eth sigs

vaibhavchellani:


vaibhavchellani:
As a layer 2, we are validating votes on Ethereum chain. And for that we want tendermint to use RLP in votes and allow ETH addresses. Currently we are maintaining a fork, but happy to send PR if you guys are okay.

Issue #3660, opened 2019-05-14
Allow for greater hostname control in "tendermint testnet" command


thanethomson:
At present, the `tendermint testnet` command allows one to change the hostname prefix of the peers in the `persistent_peers` list. This assumes that there is no suffix after the prefix+digit hostname, and that all of the hostnames are homogeneous, which is not necessarily true as one starts generating more realistic and diverse test networks.

It would be useful to have the ability to:
1. Specify a hostname suffix
2. Completely override the hostnames

PR #3661, opened 2019-05-14
tendermint testnet: Allow for better hostname control

thanethomson:
This attempts to address issue #3660. It adds two additional flags to the `tendermint testnet` command:

1. `--hostname-suffix` - to automatically append a suffix to each and every generated (non-IP) hostname
2. `--hostname` - a [StringArray](https://godoc.org/github.com/spf13/pflag#StringArrayVar) parameter to manually override all peer hostnames. Using this flag causes `tendermint testnet` to completely ignore the `--hostname-prefix`, `--hostname-suffix` and `--starting-ip-address` fields.

**Example 1 with `--hostname-suffix`:**

```bash
> tendermint testnet --v 4 --n 0 --o /tmp/testnet \
    --hostname-prefix mynode \
    --hostname-suffix .xyz.com

> cat /tmp/testnet/node0/config/config.toml | grep persistent_peers
persistent_peers = "42311234da9693ff1affa073f7ef9984a0674582@mynode0.xyz.com:26656,470c1c72deacad25c61673c0b452d5ef99b6053a@mynode1.xyz.com:26656,4a60e1b25e8af75d8688f8c81e93b6c45a8e2991@mynode2.xyz.com:26656,a8d586a3ed4b47b0b04b049445d27113ac031f03@mynode3.xyz.com:26656"
```

**Example 2 with `--hostname`:**

```bash
> tendermint testnet --v 4 --n 0 --o /tmp/testnet \
    --hostname alpha.xyz.com \
    --hostname beta.xyz.com \
    --hostname alpha.abcd.com \
    --hostname beta.abcd.com

> cat /tmp/testnet/node0/config/config.toml | grep persistent_peers
persistent_peers = "42311234da9693ff1affa073f7ef9984a0674582@alpha.xyz.com:26656,470c1c72deacad25c61673c0b452d5ef99b6053a@beta.xyz.com:26656,4a60e1b25e8af75d8688f8c81e93b6c45a8e2991@alpha.abcd.com:26656,a8d586a3ed4b47b0b04b049445d27113ac031f03@beta.abcd.com:26656"
```

* [ ] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -35,7 +35,7 @@
 - [p2p] [\#3552](https://github.com/tendermint/tendermint/pull/3552) Add PeerBehaviour Interface (@brapse)
 - [rpc] [\#3534](https://github.com/tendermint/tendermint/pull/3534) Add support for batched requests/responses in JSON RPC
 - [cli] \#3585 Add option to not clear address book with unsafe reset (@climber73)
-- [cli] [\#3160](https://github.com/tendermint/tendermint/issues/3160) Add `-config=<path-to-config>` option to `testnet` cmd (@gregdhill)
+- [cli] [\#3160](https://github.com/tendermint/tendermint/issues/3160) Add `--config=<path-to-config>` option to `testnet` cmd (@gregdhill)
 - [cs/replay] \#3460 check appHash for each block
 - [rpc] \#3362 `/dial_seeds` & `/dial_peers` return errors if addresses are incorrect (except when IP lookup fails)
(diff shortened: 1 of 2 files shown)


thanethomson:
The latest commit now adds support for the generation of unique monikers for each node configuration generated using the same scheme as for generating peer hostnames (except if `--starting-ip-address` is used, it will create random hexadecimal strings for monikers). It also adds another command line flag (`--random-monikers`) which overrides all other configuration to force the tool to generate random hexadecimal strings for monikers.

Previously the command would just use your local hostname, resulting in the same moniker for all nodes. This is not particularly useful when generating configurations for remote nodes, and results in needing to use tools like `sed` to overwrite the monikers.

Issue #3653, opened 2019-05-10
PeerBehaviour: follow up and integration


ebuchman:
ADR-039 outlined a new interface to better capture peer behaviour. Initial implementation is in #3552. The implementation is just the basic data types and does not hook up the interface to the rest of the code.

There is some follow up necessary re the extensive discussion in #3552, particularly around:
- naming
- where the errors should live (ie. consensus-domain-specific errors shouldn't be in P2P package)

Finally, we need to integrate the new interface with the rest of the code.

tessr:
Is this really in the p2p layer, or is it part of the blockchain v2 work?

marbar3778:
With the new p2p stack peer stopping is unified to a single location. We have opted to not use the behavior pkg for new reactors and instead rely on the p2p v2.

PR #3663, opened 2019-05-14
p2p: Peerbehaviour follow up (#3653)

brapse:
This PR implements some of the changes previously mentioned in (#3653) and in comment of (#3552) 
to address a few key questions:

### Where do peer behaviours live?
The current implementation places reactor specific terms/concepts in the p2p package. In the ideal, the `p2p` package would be as domain agnostic as possible.

This PR creates a new package `behaviour` to  intermediate the relationship between the p2p package and reactors. PeerBehaviours are decidedly simple, containing only an explanation string. The handling of PeerBehaviour is centralized within the `behaviour` package providing clear separation between the reactors and the responsibility of managing peers based on domain knowledge.

### How are PeerBehaviours mapped to actions?
The original PR #3552 had each PeerBehaviour as an integer. This meant that passing `3` to `Report` could potentially stop a peer. This PR maps each response private struct ensuring that only explicit constructions of PeerBehaviour can trigger an action.

* [x] Wrote tests

### DIFF
--- a/behaviour/peer_behaviour.go
+++ b/behaviour/peer_behaviour.go
@@ -1,12 +1,10 @@
 package behaviour
 
 import (
-	"errors"
-	"sync"
-
 	"github.com/tendermint/tendermint/p2p"
 )
 
+// PeerBehaviour is a struct describing the behaviour of a peer.
 type PeerBehaviour struct {
 	peerID p2p.ID
 	reason interface{}
@@ -16,6 +14,7 @@ type badMessage struct {
 	explanation string
 }
 
+// BadMessage returns a badMessage PeerBehaviour.
 func BadMessage(peerID p2p.ID, explanation string) PeerBehaviour {
 	return PeerBehaviour{peerID: peerID, reason: badMessage{explanation}}
 }
@@ -24,6 +23,7 @@ type messageOutOfOrder struct {
 	explanation string
 }
 
+// MessageOutOfOrder returns a messagOutOfOrder PeerBehaviour.
 func MessageOutOfOrder(peerID p2p.ID, explanation string) PeerBehaviour {
 	return PeerBehaviour{peerID: peerID, reason: badMessage{explanation}}
 }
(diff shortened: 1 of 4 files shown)


behaviour/peer_behaviour.go
```diff
@@ -0,0 +1,123 @@
...
+type badMessage struct {
+	explanation string
+}
+
+func BadMessage(peerID p2p.ID, explanation string) PeerBehaviour {
+	return PeerBehaviour{peerID: peerID, reason: badMessage{explanation}}
+}
+
+type messageOutOfOrder struct {
+	explanation string
```
golangcibot: `explanation` is unused (from `structcheck`)

behaviour/peer_behaviour_test.go
```diff
@@ -0,0 +1,191 @@
...
+		t.Error("Expected the peer have one reported behaviour")
+	}
+
+	if behaviours[0] != badMessage {
+		t.Error("Expected Bad Message to have been reported")
+	}
+}
+
+type scriptedBehaviours struct {
+	peerID     p2p.ID
```
golangcibot: `peerID` is unused (from `structcheck`)

behaviour/peer_behaviour_test.go
```diff
@@ -0,0 +1,191 @@
...
+	behaviours []bh.PeerBehaviour
```
golangcibot: `behaviours` is unused (from `structcheck`)

behaviour/peer_behaviour_test.go
```diff
@@ -0,0 +1,191 @@
...
+	if len(behaviours) != 1 {
+		t.Error("Expected the peer have one reported behaviour")
+	}
+
+	if behaviours[0] != badMessage {
+		t.Error("Expected Bad Message to have been reported")
+	}
+}
+
+type scriptedBehaviours struct {
```
golangcibot: `scriptedBehaviours` is unused (from `deadcode`)

behaviour/reporter.go
```diff
@@ -0,0 +1,84 @@
...
+	return &SwitchReporter{
+		sw: sw,
+	}
+}
+
+// Report reports the behaviour of a peer to the Switch.
+func (spbr *SwitchReporter) Report(behaviour PeerBehaviour) error {
+	peer := spbr.sw.Peers().Get(behaviour.peerID)
+	if peer == nil {
+		return errors.New("Peer not found")
```
melekes:
```suggestion
		return errors.New("peer not found")
```

behaviour/reporter.go
```diff
@@ -0,0 +1,84 @@
...
+
+	switch reason := behaviour.reason.(type) {
+	case consensusVote, blockPart:
+		spbr.sw.MarkPeerAsGood(peer)
+	case badMessage:
+		spbr.sw.StopPeerForError(peer, reason.explanation)
+	case messageOutOfOrder:
+		spbr.sw.StopPeerForError(peer, reason.explanation)
+	default:
+		return errors.New("Unknown reason reported")
```
melekes:
```suggestion
		return errors.New("unknown reason reported")
```

brapse:
Adding some notes from a discussion with @xla held offline:
An implication of moving PeerBehaviour outside of the `p2p` package is that the response to reporting to behaviour must be limited to the public interface exposed by the `p2p` package.

This means that the the `p2p` package will continue to expose Internal peer state and that developers can accidentally use API thats are exclusively intended to be used by PeerBehaviour.

While this consequence was never discussed in the original ADR or PRs, it's clear that the intention of this work is to minimize  the surface area of peer modifying methods. The current implementation just creates "Yet Another Way" of messing with Peers.

We'll have to zoom out a bit and look at the ideal surface area we intend the `p2p` package to expose in the mid and long term.

brapse:
After follow up feedback with @xla, we decided that the problem outlined above can be handled by a follow up issue: #3719. We can merge this in, start improving the reactors and then update make peer lifecycle APIs private through evolution of the `peerReporter`. With the remaining conflicts resolved this should be good to go 👍

PR #3664, opened 2019-05-17
p2p/conn: Add Bufferpool

guagualvcha:
According to benchmark, the read/write performance improved:

**Read** : 
Before:
BenchmarkReadSecretConnection-4   	  300000	      4747 ns/op	    4672 B/op	       6 allocs/op
After
BenchmarkReadSecretConnection-4   	  300000	      3896 ns/op	       0 B/op	       0 allocs/op

**Write**:
Before:
BenchmarkWriteSecretConnection-4   	  200000	      8477 ns/op	    8006 B/op	      10 allocs/op
After:
BenchmarkWriteSecretConnection-4   	  200000	      6765 ns/op	       0 B/op	       0 allocs/op

We test this with 11 validators:

we do not see obvious improvement in TPS,  but the go gc decrease from 30-25 to 25-10 when we run 250 tps.
 
* [x] Wrote tests


p2p/conn/secret_connection.go
```diff
@@ -143,35 +154,37 @@ func (sc *SecretConnection) Write(data []byte) (n int, err error) {
...
-		aead.Seal(sealedFrame[:0], sc.sendNonce[:], frame, nil)
-		incrNonce(sc.sendNonce)
-		// end encryption
-
-		_, err = sc.conn.Write(sealedFrame)
-		if err != nil {
+		if err := func() error {
+			var sealedFrame = pool.Get(aeadSizeOverhead + totalFrameSize)
+			var frame = pool.Get(totalFrameSize)
+			defer pool.Put(sealedFrame)
```
melekes: can we unify these

```go
defer func() {
}
```

melekes:
from https://github.com/libp2p/go-buffer-pool#disadvantages-over-gc
> Can leak memory contents. Unlike the go GC, we don't zero memory.

This is not a concern here, right?

guagualvcha:
> from https://github.com/libp2p/go-buffer-pool#disadvantages-over-gc
> 
> > Can leak memory contents. Unlike the go GC, we don't zero memory.
> 
> This is not a concern here, right?

I think so.  The buffer length we use is not dynamic, but two fixed length.

p2p/conn/secret_connection.go
guagualvcha: changed.

marbar3778:
@guagualvcha could you resolve the merge conflict here?

guagualvcha:
> @guagualvcha could you resolve the merge conflict here?

resolved

marbar3778:
Thank you! Could the gopkg.toml and .lock be removed? These are not being used anymore in the repo.

guagualvcha:
> Thank you! Could the gopkg.toml and .lock be removed? These are not being used anymore in the repo.

done

melekes:
Would be great if we could add this with a feature flag. I am a little bit concerned about adding yet another dependency. But gains in GC could be worth it.

marbar3778:
I don't think we should be concerned with the added dependency here as we are starting to explore libp2p as an alternative to our current p2p stack and will have to add many of the libraries from libp2p as dependencies. We also don't want to start throwing many things under feature flags, IMO. Only when needed

guagualvcha:
hi @marbar3778 @melekes  is this pr ready for merge?

marbar3778:
sorry for the delay, could you add an entry into the changelog under improvements then I will merge

guagualvcha:
sure, will do

PR #3668, opened 2019-05-21
p2p/conn: Make non-malleable

zmanian:
This is an approach to fixing secret connection that is more noise-ish than actually noise.

but it essentially fixes the problem that #3315 is trying to solve by making the secret connection handshake non-malleable. It's easy to understand and I think will be acceptable to @jaekwon 

Lots to bikeshed here so starting this as a draft.


Liamsi:
Cool! Thanks Zaki. Could you add a brief explanation how this solves the malleability issue? I'll add a commit to make this test code compile.

p2p/conn/secret_connection.go
```diff
@@ -98,9 +111,14 @@ func MakeSecretConnection(conn io.ReadWriteCloser, locPrivKey crypto.PrivKey) (*
 	if err != nil {
 		return nil, err
 	}
+	io.ReadFull(hkdfHiEphPub, hdkf_state[:])
+
+	hkdfSecret := hkdf.New(hash, dhSecret[:], hdkf_state[:], []byte("TENDERMINT_SECRET_CONNECTION_TRANSCRIPT_HASH"))
+
+	io.ReadFull(hkdfSecret, hdkf_state[:])
 
 	// generate the secret used for receiving, sending, challenge via hkdf-sha2 on dhSecret
```
Liamsi: Comment needs updating too. Something along the lines of?
```suggestion
	// generate the secret used for receiving, sending, challenge via hkdf-sha2 on the transcript state
```

p2p/conn/secret_connection.go
```diff
@@ -98,9 +111,14 @@ func MakeSecretConnection(conn io.ReadWriteCloser, locPrivKey crypto.PrivKey) (*
...
-	recvSecret, sendSecret, challenge := deriveSecretAndChallenge(dhSecret, locIsLeast)
+	recvSecret, sendSecret, challenge := deriveSecretAndChallenge(hdkf_state, locIsLeast)
```
Liamsi: As far as I understand, the changes make some kind of transcript hashing (similar to what noise does). As far as I can tell this is an improvement to the protocol. 

It would be cool, if there was a test that shows how this previously failed and is now fixed. Ideally, this would come with some formal reasoning, too. This is even more so important as we do not have any test-vectors to compare against (?).

p2p/conn/secret_connection.go
```diff
@@ -87,7 +92,15 @@ func MakeSecretConnection(conn io.ReadWriteCloser, locPrivKey crypto.PrivKey) (*
 	}
 
 	// Sort by lexical order.
-	loEphPub, _ := sort32(locEphPub, remEphPub)
+	loEphPub, hiEphPub := sort32(locEphPub, remEphPub)
+
+	io.ReadFull(hkdfInit, hdkf_state[:])
```
Liamsi: `io.ReadFull` can error. Hence, this should be
```suggestion
	if err := io.ReadFull(hkdfInit, hdkf_state[:]); err != nil {
	  return nil, err
	}
```

p2p/conn/secret_connection.go
```diff
@@ -87,7 +92,15 @@ func MakeSecretConnection(conn io.ReadWriteCloser, locPrivKey crypto.PrivKey) (*
...
+
+	hkdfLoEphPub := hkdf.New(hash, loEphPub[:], hdkf_state[:], []byte("TENDERMINT_SECRET_CONNECTION_TRANSCRIPT_HASH"))
+
+	io.ReadFull(hkdfLoEphPub, hdkf_state[:])
```
Liamsi:
```suggestion
	if err := io.ReadFull(hkdfInit, hdkf_state[:]); err != nil {
		return nil, err
	}
```

p2p/conn/secret_connection.go
```diff
@@ -98,9 +111,14 @@ func MakeSecretConnection(conn io.ReadWriteCloser, locPrivKey crypto.PrivKey) (*
 	if err != nil {
 		return nil, err
 	}
+	io.ReadFull(hkdfHiEphPub, hdkf_state[:])
```
Liamsi:
```suggestion
	if err := io.ReadFull(hkdfHiEphPub, hdkf_state[:]); err != nil {
		return nil, err
	}
```

p2p/conn/secret_connection.go
```diff
@@ -98,9 +111,14 @@ func MakeSecretConnection(conn io.ReadWriteCloser, locPrivKey crypto.PrivKey) (*
...
+
+	hkdfSecret := hkdf.New(hash, dhSecret[:], hdkf_state[:], []byte("TENDERMINT_SECRET_CONNECTION_TRANSCRIPT_HASH"))
+
+	io.ReadFull(hkdfSecret, hdkf_state[:])
```
Liamsi:
```suggestion
	if err := io.ReadFull(hkdfSecret, hdkf_state[:]); err != nil {
		return nil, err
	}
```

zmanian:
I can commit the reasoning.

But the formal reasoning is basically, if the "view" of the transcript between diverges between the sender and the receiver at any point in the protocol, the handshake would terminate.

The base protocol of Station to Station mistakenly assumes that if the sender and receiver arrive at shared secret they have the same view. This is only true for a DH on prime order groups.

This robustly solves the problem by having each cryptographic operation commit to operators view of the protocol.

p2p/conn/secret_connection.go
```diff
@@ -87,7 +92,19 @@ func MakeSecretConnection(conn io.ReadWriteCloser, locPrivKey crypto.PrivKey) (*
 	}
 
 	// Sort by lexical order.
-	loEphPub, _ := sort32(locEphPub, remEphPub)
+	loEphPub, hiEphPub := sort32(locEphPub, remEphPub)
+
+	if err := io.ReadFull(hkdfInit, hdkf_state[:]); err != nil {
```
golangcibot: cannot initialize 1 variables with 2 values (from `typecheck`)

Liamsi: My bad. My suggestion didn't consider the 1st return value (which can be ignored as the documentation says: "// On return, n == len(buf) if and only if err == nil."

Follow up commit in a few seconds.

tarcieri:
Another nice thing about a transcript is it provides the basis for "secure" (barring cryptographic breakages, horrible design flaws, or implementation bugs) downgrades, where a backwards compatible handshake can be used to offer newer protocol features/extensions, peers agree to the common subset of what they support, and both sides have to agree on what the other offered for the transcript MAC to verify.

With something like Protos/Amino you already get "extensions" for free (TLS uses a [simple TLV format](https://tools.ietf.org/html/rfc8446#section-4.2) for extensions not too far off from Protos/Amino), so as long as you cryptographically commit to what they contain in the transcript, it should be possible to extend the protocol in a backwards-compatible manner.

marbar3778:
What is the status with this PR?

p2p/conn/secret_connection.go
```diff
@@ -98,9 +115,19 @@ func MakeSecretConnection(conn io.ReadWriteCloser, locPrivKey crypto.PrivKey) (*
...
 
-	// generate the secret used for receiving, sending, challenge via hkdf-sha2 on dhSecret
-	recvSecret, sendSecret, challenge := deriveSecretAndChallenge(dhSecret, locIsLeast)
+	hkdfSecret := hkdf.New(hash, dhSecret[:], hdkf_state[:], []byte("TENDERMINT_SECRET_CONNECTION_TRANSCRIPT_HASH"))
+
+	if _, err := io.ReadFull(hkdfSecret, hdkf_state[:]); err != nil {
+		return nil, err
+	}
+
+	// generate the secret used for receiving, sending, challenge via HKDF-SHA2
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	// generate the secret used for receiving, sending, challenge via HKDF-SHA2
```

p2p/conn/secret_connection.go
marbar3778: `bump` could we add a test, it can also live with the p2p tests `make test_p2p`

tarcieri:
One note on this design: it appears to mix the transcript hash into the key derivation itself. There are a few reasons it might be good to avoid that, namely it gives the attacker wiggle room they wouldn't otherwise have.

The alternative is to use the transcript solely for authentication (i.e. MAC), rather than mixing it into the key material.

zmanian:
I just noticed this got merged.

I could make this code considerably simpler using https://github.com/gtank/merlin at the cost of adding sha3 as a cryptographic dependency.

marbar3778:
I would be for this change, but if it could come with more tests

zmanian:
It's somewhat unclear to me how to test this?

it succeeds or fails in pretty obvious ways.

Issue #3669, opened 2019-05-21
Handlers in proxy routes miss context parameter


yutianwu:
Routes of proxy in lite pkg:

https://github.com/tendermint/tendermint/blob/86cf8ee3f96fbff39ebf79d58e186201a2bd2222/lite/proxy/proxy.go#L61-L86

actually the handlers need an extra context parameter:

https://github.com/tendermint/tendermint/blob/86cf8ee3f96fbff39ebf79d58e186201a2bd2222/rpc/lib/server/handlers.go#L269-L279

so, if we call method in proxy, it will panic and it will complain that we call the func with too many parameters.

PR #3670, opened 2019-05-21
[R4R] Fix missing context parameter in proxy

yutianwu:
Ref: https://github.com/tendermint/tendermint/issues/3669

* [ ] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -59,3 +59,4 @@
   * `Mempool#Update` and `BlockExecutor#Commit` now accept
     `[]*abci.ResponseDeliverTx` - list of `DeliverTx` responses, which should
     match `block.Txs`
+- [rpc] \#3669 Handlers in proxy routes miss context parameter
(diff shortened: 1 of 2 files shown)


melekes:

yutianwu:
> Thanks for contributing to Tendermint 👍

never mind :)

CHANGELOG_PENDING.md
```diff
@@ -59,3 +59,4 @@
   * `Mempool#Update` and `BlockExecutor#Commit` now accept
     `[]*abci.ResponseDeliverTx` - list of `DeliverTx` responses, which should
     match `block.Txs`
+- [rpc] \#3669 Handlers in proxy routes miss context parameter
```
```suggestion
- [rpc] \#3669 Add context parameter to Handlers in proxy routes (@yutianwu)
```

Liamsi:
Thanks! We should add integration tests that would have caught this.

CHANGELOG_PENDING.md
yutianwu: thx, updated

Issue #3666, opened 2019-05-17
p2p: Improve observalibility of channel sending/receiving


brapse:
Each Peer has an MConnection which multiplexes reactor channel communication. For debugging the progress of the various subsystems it would be helpful to monitor when each channel last sent/received messages. 

The peer currently exposes sending/receiving telemetry but it doesn't include a Channel specific label.
Receiving:
https://github.com/tendermint/tendermint/blob/4253e67c07c69be6d7f7263ab03944ce30a9fc90/p2p/peer.go#L249-L253

Sending:
https://github.com/tendermint/tendermint/blob/4253e67c07c69be6d7f7263ab03944ce30a9fc90/p2p/peer.go#L365-L373

The goal then is to add a label to these metrics to allow channel specific observations.

brapse:
This was resolved by #3677

PR #3677, opened 2019-05-24
p2p: Per channel metrics (#3666)

brapse:
Reactors  run in parallel exchanging messages with peers across channels. To understand the behaviour of the system as a whole, it would be helpful that have visibility into the per channel (and thus per reactor) message exchange. With per channel metrics, we can observe when a given reactor has stopped sending/receiving and correlate cpu usage with potential thread saturation.

Cf. #3666

* [ ] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -45,6 +45,8 @@
     use `make build_c` / `make install_c` (full instructions can be found at
     https://tendermint.com/docs/introduction/install.html#compile-with-cleveldb-support)
   * Use `boltdb` tag to compile Tendermint with bolt db
+- [p2p] [\#3666](https://github.com/tendermint/tendermint/issues/3666)
+  Add per channel telemtry to improve reactor observability
 
 ### BUG FIXES:
 - [p2p] \#3532 limit the number of attempts to connect to a peer in seed mode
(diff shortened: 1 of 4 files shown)


CHANGELOG_PENDING.md
```diff
@@ -45,6 +45,8 @@
     use `make build_c` / `make install_c` (full instructions can be found at
     https://tendermint.com/docs/introduction/install.html#compile-with-cleveldb-support)
   * Use `boltdb` tag to compile Tendermint with bolt db
+- [p2p] [\#3666](https://github.com/tendermint/tendermint/issues/3666)
```
Liamsi: No need to change anything but just FYI: you do not need to add the link yourself here. There is a little python script that generates the link to the issue from `[\#3666]`.

brapse:
> Awesome thanks - should this be considered a breaking change? Ie. will anyones current setup break with this change, or does it just add new/more info ?

This should be a non breaking changing. The only concern which might be worth noting is this will create additional metrics and increased data for prometheus server pulling the data in. So long as the peer set is bounded and has low churn, This shouldn't be too big a deal.

melekes:
> So long as the peer set is bounded and has low churn, This shouldn't be too big a deal.

not the case for seed nodes

brapse:
Would it be something in the region of:
`1000s of peers, 10s of channels`?
Which I think is ok to pull and could be queried efficiently with a [rule](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/)

PR #3705, opened 2019-06-03
docs: update RPC docs for /subscribe & /unsubscribe

### LINKED ISSUE #3678: Example in docs on Websocket RPC Subscribe
**Tendermint version**: 0.30.2

**Environment**:
- **OS**: macOS 10.14.4
- **Install tools**: docker 

**Have you tried the latest version**: no

I'm trying to run example from https://tendermint.com/rpc/#subscribe , but I'm not sure if I'm using correct imports. In doc, there is a `client`, but no import, and I use `rpc/client`:

```go
rpcclient "github.com/tendermint/tendermint/rpc/client"
client := rpcclient.NewHTTP(...)
ctx, _ := context.WithTimeout(context.Background(), 10 * time.Second)
query := "tm.event = 'Tx' AND tx.height = 3"
resultCh, err := client.Subscribe(ctx, "test-client", query, 1)
```

I had to tweak the code a bit (query is string, not Query, txs => resultCh), and now I'm stuck with SIGSEGV.

Minified stack trace:

```
panic SIGSEGV
rpc/lib/client.(*WSClient).Call(0x0, ...) ws_client.go:217
rpc/lib/client.(*WSClient).Subscribe(0x0, ...) ws_client.go:485
rpc/client.(*WSEvents).Subscribe(...) httpclient.go:309
test
```

P.S. Also, there's shell example missing. I could contribute it, but it will require usage of [websocat](https://github.com/vi/websocat). Would you accept such an example? It could look like this:
```bash
# installation info: https://github.com/vi/websocat
websocat -t --jsonrpc ws://localhost:26657/websocket
subscribe "tm.event = 'Tx'"
```

melekes:
Fixes #3678

* [ ] Wrote tests

### DIFF
--- a/rpc/core/events.go
+++ b/rpc/core/events.go
@@ -61,7 +61,7 @@ import (
 // ctx, cancel := context.WithTimeout(context.Background(), 1 * time.Second)
 // defer cancel()
 // query := "tm.event = 'Tx' AND tx.height = 3"
-// txs, err = client.Subscribe(ctx, "test-client", query, txs)
+// txs, err := client.Subscribe(ctx, "test-client", query)
 // if err != nil {
 //   // handle error
 // }
--- a/rpc/core/events.go
+++ b/rpc/core/events.go
@@ -204,6 +204,9 @@ func Unsubscribe(ctx *rpctypes.Context, query string) (*ctypes.ResultUnsubscribe
 // }
 // defer client.Stop()
 // err = client.UnsubscribeAll(context.Background(), "test-client")
+// if err != nil {
+//   // handle error
+// }
 // ```
 //
 // > The above command returns JSON structured like this:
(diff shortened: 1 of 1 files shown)


rpc/core/events.go
```diff
@@ -59,15 +58,17 @@ import (
...
 // }
 // defer client.Stop()
-// ctx, cancel := context.WithTimeout(context.Background(), timeout)
+// ctx, cancel := context.WithTimeout(context.Background(), 1 * time.Second)
 // defer cancel()
-// query := query.MustParse("tm.event = 'Tx' AND tx.height = 3")
-// txs := make(chan interface{})
-// err = client.Subscribe(ctx, "test-client", query, txs)
+// query := "tm.event = 'Tx' AND tx.height = 3"
+// txs, err = client.Subscribe(ctx, "test-client", query, txs)
```
ebuchman:
```suggestion
// txs, err := client.Subscribe(ctx, "test-client", query)
```

Issue #3656, opened 2019-05-10
Exit if SwitchToConsensus fails


ebuchman:
It's possible the `ConsensusState.Start()` returns an error. However, when the BlockchainReactor calls SwitchToConsensus, this error is simply logged and ignored. Since this leaves the node in an unintended state (ie. not participating in consensus), we should probably exit instead, perhaps via panicing.

This is partial follow up to #3295

ebuchman:
Should still try to add an integration test: https://github.com/tendermint/tendermint/pull/3706/#issuecomment-498831577

PR #3706, opened 2019-06-03
cs: exit if SwitchToConsensus fails

melekes:
Refs #3656

* [ ] Wrote tests


melekes:
Not sure if we need to write a test here. I mean we do want to test all failure scenarios in ConsensusState#OnStart, but these should probably be located in consensus_state_test.go.

thanethomson:
> Not sure if we need to write a test here. I mean we do want to test all failure scenarios in ConsensusState#OnStart, but these should probably be located in consensus_state_test.go.

It'd be ideal if there was an easy test to write here (e.g. causing the WAL to fail on start), but that may be a bit too contrived? A great integration test would be one that would start a node, kick off a fast sync, and then during the switch-over to consensus cause the ConsensusState WAL to fail.

Where would such an integration test be best placed?

PR #3710, opened 2019-06-03
docs: Add check validators spec

josef-widder:
* [ ] Wrote tests


docs/spec/consensus/checkvalidators.md
```diff
@@ -0,0 +1,224 @@
...
+    return false // old header was once trusted but it is expired
+  }
+
+  PAs := compute_all_PAs(oh)  // Definition 1
+  PAs := reduce (PAs)         // remove every PA that is a subset of another PA
+  UV := compute_UV(oh,nh)     // Definition 2
+
+  vpUV := votingpower(UV,nh)  // second sum in Theorem 1
+  vpNH := votingpower(nh.V,nh) // right hand side of premise of Theorem 1
+  vpMaxPA := maximumvotingpower(PAs,nh) // voting powers of all PA and big max
```
ancazamfir: Not sure if you looked into the complexity here yet. Maybe we can use the 0/1 knapsack algorithm (pseudo-polynomial), i.e. determine vpMaxPA, the maximum power that can fill a knapsack no bigger than 1/3 total voting power.
It would be O(n*vpNH), better than the O(2**n) for compute_all_PAs(oh) but still not great. 
n - number validators (~100)
vpNH - 1/3 total voting power, would have to look into real numbers but I think it is high...maybe can be optimized using normalization of powers (weights in knapsack problem)?.

docs/spec/consensus/checkvalidators.md
```diff
@@ -0,0 +1,224 @@
...
+
+  return vpMaxPA + vpUV < 1/3 * vpNH  // Theorem 1. It must be smaller for all
+                                        // so it must be smaller for the max
```
ancazamfir: Maybe we can memoize (vpMaxPA + vpUV) as the vpMaxPA of next iteration's old header so we would only need to do this once for initial header.

docs/spec/consensus/checkvalidators.md
```diff
@@ -0,0 +1,224 @@
...
+\]
+
+
+
+
+**Theorem 1.** If for all potential adversaries PA, in $nh$ the combined voting
+powers of PA and UV is less than a third of the total voting power, then in
+$nh$, more than 2/3 of the voting power is held by correct processes. Formally,
+if for all PA
+
```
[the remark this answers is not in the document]
milosevic: Note that in the expression below, we fix PA from old header, but we take into account only its voting power in the new header.

docs/spec/consensus/checkvalidators.md
milosevic: Haven't been thinking about efficient way of implementing it. We would need your help there :)

docs/spec/consensus/checkvalidators.md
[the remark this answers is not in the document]
ancazamfir: But the calculation on the old header just gives us the maximum possible bad voting power. It doesn't tell us that in fact the validators in the PA that gives that maximum are faulty. It can happen that actual bad validators outside PA have increased in power and good validators have decreased in power. I will try to find an example for this.

docs/spec/consensus/checkvalidators.md
```diff
@@ -0,0 +1,224 @@
...
+  return vpMaxPA + vpUV < 1/3 * vpNH  // Theorem 1. It must be smaller for all
+                                        // so it must be smaller for the max
+}
+```  
+
+
+## Remarks
+
+**Remark.** Computing all PAs might be too expensive (all subsets of $oh.V$ that have a certain combined voting power in oh). Similarly, we then have to compute all voting powers of PAs in nh to get the maximum. This is disturbing, as right now, based on the examples, I expect that CheckVS will mostly return false, assuming that there are frequent changes in the validator sets. However, $oh.V=nh.V$ might be the common case.
+
```
ancazamfir: Should this spec also cover what the lite client should do if CheckVS returns false? There might be just doing bisection or maybe a smarter  strategy

docs/spec/consensus/checkvalidators.md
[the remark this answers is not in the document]
milosevic: We don't do calculation on the old header. Note that the sum is only on p which is voting power of validators from PA in the new header. We only intersect with the old header.

docs/spec/consensus/checkvalidators.md
milosevic: Yes, it should be complete algorithm that includes bisection most probably. We need to integrate this with the old spec so it complete. There are also some parts missing regarding counter factual slashing.

docs/spec/consensus/checkvalidators.md
```diff
@@ -0,0 +1,224 @@
...
+## Definitions
+
+  * header fields
+    - $height$
+    - $bfttime$: the chain time when the header (block) was generated
+    - $V$: validator set containing validators.
+    - $nextV$: next validators
+  * $tp$: trusting period   
+  * for a time $t$, the predicate $correct(v,t)$ is true if the validator $v$
+    follows the protocol until time $t$ (we will see about recovery later).
```
cwgoes: What is `t` in reference to? A local clock or a global clock?

docs/spec/consensus/checkvalidators.md
```diff
@@ -0,0 +1,224 @@
...
+    Similarly we define $faulty(v,t)$.
+  * For each header $h$ it has locally stored, the LightClient stores whether
+    it trusts $h$. We write $trust(h) = true$, if this is the case.
+  * Validator fields. We will write a validator as a tuple $(v,p)$ such that
+    + $v$ is the identifier (we assume identifiers are unique in each validator set)
+    + $p$ is its voting power
+
+## LightClient Trusting Spec
+
+### LightClient Invariant
```
cwgoes: I am not sure this is the best place to start the light client definition. Is this how we define full nodes? I think we just want the light client header acceptance algorithm to reflect what a full node would have done given the same set of messages (condensed into headers in the case of a light client).

In particular, this does not even hold for subsequent blocks (block 11 following block 10) if the validator set changes by more than 1/3, which is possible. It also doesn't capture fault-accountability (which we should be able to capture precisely, such that a light client could only diverge from a full node if a slashable signature event by at least a well-defined fraction of stake had occurred).

docs/spec/consensus/checkvalidators.md
```diff
@@ -0,0 +1,224 @@
...
+
+  Formally,
+  \[\sum_{(v,p) \in h.V \wedge \\correct(v,h.bfttime + tp)} p > 2/3 \sum_{(v,p) \in h.V} p\]
+
+  Equivalently,
+  \[\sum_{(v,p) \in h.V \wedge\\ faulty(v,h.bfttime + tp)} p < 1/3 \sum_{(v,p) \in h.V} p\]
+
+**Question:** What should be the precise assumption here. Is the invariant on $h.V$ or on $h.NextV$ or both?
+
+*Assumption:* If a header is properly generated, then the above equations hold.
```
cwgoes: What does this mean? No, I don't think they do, if the validator set changes by enough even in a single block.

docs/spec/consensus/checkvalidators.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Liveness
+
+*Draft:* If a header $h$ as been properly generated by the blockchain (and its age is less than the trusting period), then a correct LightClient will eventually set $trust(h) = true$.
```
cwgoes: We need to assume synchronous data availability of headers, which is fine. Maybe we should emphasize that headers can be received out of order and the light client will determine which it needs to query for at what times?

docs/spec/consensus/checkvalidators.md
```diff
@@ -0,0 +1,224 @@
...
+social consensus. It is assumed that *inithead* satisfies the LightClient
+Invariant.
+
+When a LightClients sees a new header it has to decide whether to trust the new
+header. Trust can be obtained by (possibly) the combination of two methods.
+
+1. **an uninterrupted sequence of proof.** If a block is appended to the chain, where the last block
+is trusted (and properly committed by the old validator set in the next block),
+and the new block contains a new validator set, the new block is trusted if the LightClient knows all headers in the prefix.
+Intuitively, a trusted validator set is assumed to not chose a new validator set
```
cwgoes: OK, but it's also the case, critically, that headers contain the hash of the next validator set, so the light client can verify that the validator set of block `n` (which they trust) signed off on the validator set of block `n + 1`, which provides some level of assurance (in practice) on what has happened in the state machine, e.g. that the new validators have put up stake which could be slashed.

docs/spec/consensus/checkvalidators.md
```diff
@@ -0,0 +1,224 @@
...
+that violates the fault assumption.
+
+2. **trusting period.** Based on a trusted block *h*, and the LightClient
+Invariant, which ensures the fault assumption during the trusting period, we can
+try to infer wether it is guaranteed that the validator set of the new block
```
cwgoes:
```suggestion
try to infer whether it is guaranteed that the validator set of the new block
```

docs/spec/consensus/checkvalidators.md
```diff
@@ -0,0 +1,224 @@
...
+
+####Example: Changed Voting powers
+
+\[
+oh.V = \{(1,1), ... (4,1)\}\\
+nh.V = \{(1,1), ... (4,2)\}
+\]
+
+
+Validator 4 has more than a third voting power in nh.V. As trusting oh does not
```
cwgoes: This could occur on subsequent blocks.

docs/spec/consensus/checkvalidators.md
```diff
@@ -0,0 +1,224 @@
...
+
+  return vpMaxPA + vpUV < 1/3 * vpNH  // Theorem 1. It must be smaller for all
+                                        // so it must be smaller for the max
+}
+```  
+
+
+## Remarks
+
+**Remark.** Computing all PAs might be too expensive (all subsets of $oh.V$ that have a certain combined voting power in oh). Similarly, we then have to compute all voting powers of PAs in nh to get the maximum. This is disturbing, as right now, based on the examples, I expect that CheckVS will mostly return false, assuming that there are frequent changes in the validator sets. However, $oh.V=nh.V$ might be the common case.
```
cwgoes: I *think* we can just calculate the intersection between `oh.V` and `nh.V` (a set of validator keys), sum the total stake, subtract largest-first until we reach `1/3` (maximally conservative, assume the largest-stake faulty nodes intersected), and check if the resulting total is `>= 2/3` of stake at `nh.V`. Should be double-checked, but that algorithm should be linear time.

docs/spec/consensus/checkvalidators.md
cwgoes: See below comment.

docs/spec/consensus/checkvalidators.md
ancazamfir: Not sure I understand well your proposal, but I see two issues:
1) I believe the algorithm does not guarantee that the maximal subset with voting power < 1/3 is computed (as at some point those removed may be actually part of the maximal subset)
2) I ran a similar more complex but exact algorithm on the cosmos hub validators and since block 1 it can find a subset with total voting power = 1/3 - epsilon, with typical epsilon = 1. This means that the smallest change in voting power will cause (in theory) the "bad" voting power to be >= 1/3rd of nh.V. 
In hindsight this is expected with enough number of validators with wealth distribution for voting powers.
This brings us back to computing all subsets which is impractical (O(2**n)).
So in the end some form of periodic checkpoints + linear validation is more practical...unless I am completely wrong.

cwgoes: Ah yes, I see, thanks.

marbar3778:
Closing this PR and https://github.com/tendermint/tendermint/pull/3796, as @milosevic & @josef-widder will open a new pr with these two documents combined.

PR #3728, opened 2019-06-16
Missing 'b' in python 3 command

andynog:
In Python 3 the command outlined in this doc `import codecs; codecs.decode("YWJjZA==", 'base64').decode('ascii')` throws an error:
TypeError: decoding with 'base64' codec failed (TypeError: expected bytes-like object, not str), needs to add 'b' before the encoded string
`import codecs; codecs.decode(b"YWJjZA==", 'base64').decode('ascii')` to make it work

* [ ] Wrote tests


melekes:
Thanks for contributing to Tendermint 🚀 🌔 🍸

PR #3729, opened 2019-06-17
p2p: Stop peers gracefully when stopping switch

SebastianElvis:
This PR fixed `TODO: gracefully disconnect from peers.` in `p2p/switch.go`.

* [ ] Wrote tests

### DIFF
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -221,8 +221,7 @@ func (sw *Switch) OnStart() error {
 func (sw *Switch) OnStop() {
 	// Stop peers
 	for _, p := range sw.peers.List() {
-		sw.StopPeerGracefully(p)
-		sw.metrics.Peers.Add(float64(-1))
+		sw.stopAndRemovePeer(p, nil)
 	}
 
 	// Stop reactors
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -221,11 +221,8 @@ func (sw *Switch) OnStart() error {
 func (sw *Switch) OnStop() {
 	// Stop peers
 	for _, p := range sw.peers.List() {
-		sw.transport.Cleanup(p)
-		p.Stop()
-		if sw.peers.Remove(p) {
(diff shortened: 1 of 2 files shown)


p2p/switch.go
```diff
@@ -221,11 +221,8 @@ func (sw *Switch) OnStart() error {
 func (sw *Switch) OnStop() {
 	// Stop peers
 	for _, p := range sw.peers.List() {
-		sw.transport.Cleanup(p)
-		p.Stop()
-		if sw.peers.Remove(p) {
-			sw.metrics.Peers.Add(float64(-1))
-		}
+		sw.StopPeerGracefully(p)
+		sw.metrics.Peers.Add(float64(-1))
```
melekes: `StopPeerGracefully` already decrements this counter

p2p/switch.go
```diff
@@ -221,11 +221,8 @@ func (sw *Switch) OnStart() error {
 func (sw *Switch) OnStop() {
 	// Stop peers
 	for _, p := range sw.peers.List() {
-		sw.transport.Cleanup(p)
-		p.Stop()
-		if sw.peers.Remove(p) {
-			sw.metrics.Peers.Add(float64(-1))
-		}
+		sw.StopPeerGracefully(p)
```
melekes: I think it's better to call `sw.stopAndRemovePeer(peer, nil)` to avoid excessive logging

p2p/switch.go
SebastianElvis: resolved

p2p/switch.go
SebastianElvis: resolved

SebastianElvis:
> 👍
> 
> Although I am not sure this PR actually changes anything except removing duplicated code. I think we already do "gracefully disconnect from peers", so the TODO can be removed.

Yeah, it seems that the original code works.
Anyway, this PR can remove some duplicated code.
I have addressed your change requests. How about now?

PR #3731, opened 2019-06-19
Changes to files that had linting issue

marbar3778:
- Govet issues fixed
- 1 gosec issue solved using nolint

Refs #3262

* [ ] Wrote tests

### DIFF
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -351,10 +351,6 @@ func (conR *ConsensusReactor) Receive(chID byte, src p2p.Peer, msgBytes []byte) 
 	default:
 		conR.Logger.Error(fmt.Sprintf("Unknown chId %X", chID))
 	}
-
-	if err != nil {
-		conR.Logger.Error("Error in Receive()", "err", err)
-	}
 }
 
 // SetEventBus sets event bus.
--- a/p2p/upnp/upnp.go
+++ b/p2p/upnp/upnp.go
@@ -197,7 +197,7 @@ func localIPv4() (net.IP, error) {
 }
 
 func getServiceURL(rootURL string) (url, urnDomain string, err error) {
-	r, err := http.Get(rootURL)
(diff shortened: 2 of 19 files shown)


consensus/reactor.go
```diff
@@ -351,10 +351,6 @@ func (conR *ConsensusReactor) Receive(chID byte, src p2p.Peer, msgBytes []byte)
 	default:
 		conR.Logger.Error(fmt.Sprintf("Unknown chId %X", chID))
 	}
-
-	if err != nil {
```
melekes: why remove?

p2p/upnp/upnp.go
```diff
@@ -197,7 +197,7 @@ func localIPv4() (net.IP, error) {
 }
 
 func getServiceURL(rootURL string) (url, urnDomain string, err error) {
-	r, err := http.Get(rootURL)
+	r, err := http.Get(rootURL) // nolint: gosec
```
melekes: is this something we should worry about?

consensus/reactor.go
marbar3778: `nilness: impossible condition: nil != nil (govet)
        if err != nil {`

this is the linting error I was receiving, when looking through the function everywhere err is being set it is being checked as well

p2p/upnp/upnp.go
marbar3778: the lint error arises because rootURL is a variable, but this is the assumption being made with the parameter of getServiceURL being rootURL

melekes: Can this be exploited? Probably yes, but I guess validators who's using upnp know about this.

marbar3778: I don't believe it can be exploited, it is only used in one place in the code space and that is to get the serviceURL and urnDomain for NAT. I'm not in tune with possible exploits of this, but I may be wrong.

Issue #3618, opened 2019-05-03
rpc: landing page incorrectly reports all endpoints as needing args


ebuchman:
The landing page for the RPC used to differentiate between endpoints that need args and those that don't. But now it lists them all as needing args. Unclear when this changed.

The idea was really to indicate the distinction on the main page (ie. <url>:26657/). It used to look like what's in the documentation you linked to, but in the current build the section Available endpoints: is empty and all endpoints are listed under Endpoints that require arguments:. This is what we're looking to fix! https://github.com/tendermint/tendermint/pull/3737#issuecomment-524081361

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 100.0 DAI (100.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/tendermint/tendermint/3618/3138).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/tendermint/tendermint/3618/3138).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $86,178.27 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


**1) [nanspro](https://gitcoin.co/profile/nanspro) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/tendermint/tendermint/3618/3138?mutate_worker_action=approve&worker=nanspro) | [reject worker](https://gitcoin.co/issue/tendermint/tendermint/3618/3138?mutate_worker_action=reject&worker=nanspro))_.**

I am familiar with go(not very advanced level). I have worked upon fabric and go-spacemesh projects in the past and that's when i got familiar with it. I have used rpc calls before many times to interact with blockchains running on different consensus like quorum, sawtooth, fabric so i'd like to take a look and identify the problem.

I want to work on tendermint as i have heard about it from a lot of people and this issue is tagged as good first issue so i think it's the perfect opportunity to learn more about tendermint.

[...]

PR #3737, opened 2019-06-20
[WIP] Specifying endpoints query parameters everywhere

nanspro:

Relates to #3618


nanspro:
@ebuchman i didn't get any response to this https://github.com/tendermint/tendermint/issues/3618#issuecomment-503451865 and i was not sure what the scope of this issue was, differentiating between endpoints having no parameters and endpoints having some can be done in many ways so i choose to specify query parameters for all the api calls.
One more differentiation can be seen [here](https://tendermint.com/rpc/?shell#get-the-list), in the shell you can see which calls require parameters or not.

marbar3778:
`bump` any update with this PR?

ebuchman:
Thanks for this. The extra documentation is certainly super helpful.

The idea was really to indicate the distinction on the main page (ie. `<url>:26657/`). It used to look like what's in the documentation you linked to, but in the current build the section `Available endpoints:` is empty and all endpoints are listed under `Endpoints that require arguments:`.  This is what we're looking to fix!

marbar3778:
`Ping` is there an update on this PR?

melekes:
Closing this since no updates have been made.

Issue #1182, opened 2018-02-03
clean deps to minimize client binary size


ebuchman:
We want to minimize how much code a client binary has to compile so we can run on small devices and so the code remains easily auditable.

ValarDragon:
I don't think the amount of deps we have is an issue for small devices which it makes sense to be compiling from. Golang will only compile the modules we are actually using.

Dep will download the entire repo, but the space taken by `tendermint/tendermint` with all of its vendors (plus overhead due to me having git history + stashes) is `61.5 MB`. That seems fine for expecting someone who wants to compile from source to download.

marbar3778:
No longer using dep, @melekes can you close this issue

melekes:
> No longer using dep

I think the issue is we're importing a lot of stuff, not that there's something wrong with dep/go modules.

marbar3778:
Seems like everything is being used within the go.mod file. Is the intention here to bring some of these packages into tendermint so they aren't being imported? there a few indirect imports.

marbar3778:
The current size of the binary is 28mb when I add `-s -w` i get 23, is this small enough for a binary, or is our aim smaller?

marbar3778:
Closing this issue as the most recent release of tendermint (v0.32.0) binary size is 23 mb, please reopen if you believe we should work on reducing it more.

PR #3742, opened 2019-06-21
Added flags '-s -w' to buildflags

marbar3778:
ref #1182 
- `-s -w` reduces the binary size by removing symbols that are only used by debuggers. All panics, stacktraces, etc. stay the same.  
       - -s : Omit the symbol table and debug information.
       - -w : Omit the DWARF symbol table.
The reduction in the size of the binary is 18%, this primarily comes from removing the DWARF symbol table.

- Ran go mod tidy on the repo to help clean up the go.mod file

* [ ] Wrote tests


melekes:
People who want to use debuggers (developers) should be able to build binaries without `-s -w` themselves (this will require modifying Makefile). I think it's fine.

alessio:
I'd rather do something conditional like the following:

```
ifeq ($(NOSTRIP),)
  BUILD_FLAGS += -w -s
endif
```

So that such flags can be added back in by just

```
make NOSTRIP=y
```

(Sent from mobile - excuse the brevity)

Makefile
```diff
@@ -11,7 +11,12 @@ export GO111MODULE = on
 
 INCLUDE = -I=. -I=${GOPATH}/src -I=${GOPATH}/src/github.com/gogo/protobuf/protobuf
 BUILD_TAGS?='tendermint'
-BUILD_FLAGS = -mod=readonly -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short=8 HEAD`"
+LD_FLAGS = -X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short=8 HEAD`
+BUILD_FLAGS = -mod=readonly -ldflags "$(LD_FLAGS)"
+
+ifeq ($(NOSTRIP),)
+	LD_FLAGS += -s -w
+endif
```
marbar3778: @alessio can you verify this

melekes: I think ifeq should go above the `BUILD_FLAGS =`

Makefile
```diff
@@ -11,7 +11,12 @@ export GO111MODULE = on
 
 INCLUDE = -I=. -I=${GOPATH}/src -I=${GOPATH}/src/github.com/gogo/protobuf/protobuf
 BUILD_TAGS?='tendermint'
-BUILD_FLAGS = -mod=readonly -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short=8 HEAD`"
+LD_FLAGS = -X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short=8 HEAD`
+BUILD_FLAGS = -mod=readonly -ldflags "$(LD_FLAGS)"
+
+ifeq ($(NOSTRIP),)
+	LD_FLAGS += -s -w
```
alessio:
```suggestion
  BUILD_FLAGS += -s -w
```

@melekes I reckon it had better initialise BUILD_FLAGS first, then add all the conditional stuff at later stage

marbar3778: @alessio if I make it add to build flags then it adds outside the quotation marks so that's why I went with this. 

Unless you mean put "-mod=readonly -ldflags" on the build commands and have the build flags be only `-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short=8 HEAD`
`

Makefile
```diff
@@ -11,7 +11,11 @@ export GO111MODULE = on
 
 INCLUDE = -I=. -I=${GOPATH}/src -I=${GOPATH}/src/github.com/gogo/protobuf/protobuf
 BUILD_TAGS?='tendermint'
-BUILD_FLAGS = -mod=readonly -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short=8 HEAD`"
+BUILD_FLAGS = -X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short=8 HEAD`
```
melekes: LDFLAGS?

Makefile
```diff
@@ -21,19 +25,19 @@ check: check_tools
 ### Build Tendermint
 
 build:
-	CGO_ENABLED=0 go build $(BUILD_FLAGS) -tags $(BUILD_TAGS) -o $(OUTPUT) ./cmd/tendermint/
+	CGO_ENABLED=0 go build -mod=readonly -ldflags "$(BUILD_FLAGS)" -tags $(BUILD_TAGS) -o $(OUTPUT) ./cmd/tendermint/
```
melekes: I'd prefer to avoid duplication

Makefile
```diff
@@ -11,7 +11,11 @@ export GO111MODULE = on
 
 INCLUDE = -I=. -I=${GOPATH}/src -I=${GOPATH}/src/github.com/gogo/protobuf/protobuf
 BUILD_TAGS?='tendermint'
-BUILD_FLAGS = -mod=readonly -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short=8 HEAD`"
+BUILD_FLAGS = -X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short=8 HEAD`
+
+ifeq ($(NOSTRIP),)
+	BUILD_FLAGS += -s -w
```
ebuchman: Need to document this in the build instructions

Makefile
```diff
@@ -21,19 +21,19 @@ check: check_tools
 ### Build Tendermint
 
 build:
-	CGO_ENABLED=0 go build $(BUILD_FLAGS) -tags $(BUILD_TAGS) -o $(OUTPUT) ./cmd/tendermint/
+	CGO_ENABLED=0 go build -mod=readonly -ldflags "$(LD_FLAGS)" -tags $(BUILD_TAGS) -o $(OUTPUT) ./cmd/tendermint/
```
melekes: BUILD_FLAGS = -mod=readonly -ldflags "$(LD_FLAGS)"

PR #3744, opened 2019-06-21
abci: Refactor CheckTx to notify of recheck

thanethomson:
As per #2127, this refactors the `RequestCheckTx` ProtoBuf struct to allow for a flag indicating whether a query is a recheck or not (and allows for possible future, more nuanced states).

In order to pass this extended information through to the ABCI app, the `proxy.AppConnMempool` (and, for consistency, the `proxy.AppConnConsensus`) interface seems to need to be refactored along with `abcicli.Client`.

And, as per [this comment](https://github.com/tendermint/tendermint/issues/2127#issuecomment-456661420), I've made the following modification to the protobuf definition for the `RequestCheckTx` structure:

```protobuf
enum CheckTxType {
  New = 0;
  Recheck = 1;
}

message RequestCheckTx {
  bytes tx = 1;
  CheckTxType type = 2;
}
```

* [ ] Wrote tests


mempool/clist_mempool.go
```diff
@@ -279,7 +279,10 @@ func (mem *CListMempool) CheckTxWithInfo(tx types.Tx, cb func(*abci.Response), t
 		return err
 	}
 
-	reqRes := mem.proxyAppConn.CheckTxAsync(tx)
+	reqRes := mem.proxyAppConn.CheckTxAsync(abci.RequestCheckTx{
+		Tx: tx,
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

thanethomson:
In what ways could `RequestCheckTx.additional_data` be used to compromise security? Should we perhaps be considering a maximum length constraint there?

abci/types/types.proto
```diff
@@ -81,8 +81,15 @@ message RequestBeginBlock {
...
 
+enum CheckTxType {
+  Unchecked = 0;
+  Checked = 1;
+}
+
 message RequestCheckTx {
   bytes tx = 1;
+  CheckTxType type = 2;
+  bytes additional_data = 3;
```
thanethomson: Is it possible for `additional_data` to introduce new security threats? If so, how would we go about mitigating them?

mempool/clist_mempool.go
```diff
@@ -591,7 +591,10 @@ func (mem *CListMempool) recheckTxs() {
 	// NOTE: globalCb may be called concurrently.
 	for e := mem.txs.Front(); e != nil; e = e.Next() {
 		memTx := e.Value.(*mempoolTx)
-		mem.proxyAppConn.CheckTxAsync(memTx.tx)
+		mem.proxyAppConn.CheckTxAsync(abci.RequestCheckTx{
+			Tx:   memTx.tx,
+			Type: abci.CheckTxType_Checked,
```
thanethomson: Is this the only place where a recheck happens? Just checking in case I've missed something.

abci/types/types.proto
melekes: Why add this field now if there are no use-cases currently? we can always add it later (this won't be a breaking change)

abci/types/types.proto
```diff
@@ -81,8 +81,15 @@ message RequestBeginBlock {
   repeated Evidence byzantine_validators = 4 [(gogoproto.nullable)=false];
 }
 
+enum CheckTxType {
+  Unchecked = 0;
```
melekes: I'd call these `New` and `Recheck`. Having `CheckTxRequest` with type `Checked` is confusing to me.

mempool/clist_mempool.go
melekes: This is the only place.

abci/types/types.proto
thanethomson: Removed it now 👍

abci/types/types.proto
thanethomson: Renamed it as such 👍

docs/spec/abci/abci.md
```diff
@@ -296,6 +296,12 @@ Commit are included in the header of the next block.
 
 - **Request**:
   - `Tx ([]byte)`: The request transaction bytes
+  - `Type (CheckTxType)`: What type of `CheckTx` request is this? At present,
+    there are two possible values: `CheckTx_Unchecked` (the default, which says
```
ebuchman: Oh, seems this is not aligned with the changes actually made to the proto file, ie. the types are `New` and `Recheck`

docs/spec/abci/abci.md
```diff
@@ -296,6 +296,12 @@ Commit are included in the header of the next block.
...
+    that a full check is required), and `CheckTx_Checked` (when the mempool is
+    initiating a normal recheck of a transaction).
+  - `AdditionalData ([]byte)`: Reserved for future use. See
```
ebuchman: This wasn't implemented yet

docs/spec/abci/abci.md
ebuchman: We should actually also add an entry to `docs/spec/abci/apps.md` explaining the impact of this field on application developers (ie. they may be able to skip signature verification)

Issue #3724, opened 2019-06-13
Make the RPC host localhost:26657 to prevent DDOS attacks


jackzampolin:

melekes:
`0.0.0.0` -> `localhost`?

thanethomson:
Please elaborate on this issue. This is [configurable](https://github.com/tendermint/tendermint/blob/master/config/config.go#L290).

thanethomson:
So this refers to the **default** value for the configuration.

marbar3778:
PR #3746 closes this issue.

PR #3746, opened 2019-06-23
Make RPC bind to localhost by default

thanethomson:
Addresses #3724 

* [ ] Wrote tests


thanethomson:
Changing the default RPC binding address to `tcp://127.0.0.1:26657` breaks the localnet and p2p tests, because both rely on the `tendermint testnet` command and implicitly assume that the RPC is going to be available externally.

I've now added a flag to the `testnet` subcommand (`--rpc-laddr`) to allow one to manually override this, but by default it still binds to localhost to keep with the security-oriented theme of making localhost the default.

The relevant CircleCI tests have been modified now to cater for this change.

cmd/tendermint/commands/testnet.go
```diff
@@ -64,6 +65,8 @@ func init() {
 		"P2P Port")
 	TestnetFilesCmd.Flags().BoolVar(&randomMonikers, "random-monikers", false,
 		"Randomize the moniker for each generated node")
+	TestnetFilesCmd.Flags().StringVar(&rpcListenAddr, "rpc-laddr", "tcp://127.0.0.1:26657",
```
melekes: I'm against of adding any more options to testnet cmd. We should just pass the whole config to it in tests instead of adding config variables one by one.

thanethomson: Sure, I'll take it out.

melekes:
We can also move https://github.com/tendermint/tendermint/blob/3e7752c29d33c1c11bb2ba0ba9ac6985d47d83e1/cmd/tendermint/commands/testnet.go#L195-L196 into config template.

networks/local/localnode/Dockerfile
```diff
@@ -13,4 +13,4 @@ CMD ["node", "--proxy_app", "kvstore"]
 STOPSIGNAL SIGTERM
 
 COPY wrapper.sh /usr/bin/wrapper.sh
-
+COPY config-template.toml /etc/tendermint/config-template.toml
```
thanethomson: I added this here but to no effect (see the [CircleCI build logs](https://circleci.com/gh/tendermint/tendermint/61887?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)).

Upon investigation, it appears as though the `localnet` test pulls the `localnode` Docker image, which was [last updated 6 months ago](https://cloud.docker.com/u/tendermint/repository/docker/tendermint/localnode).

@greg-szabo: Is there somewhere where `localnode` gets built/pushed automatically? Or is it a manual process?

I'm going to make it a prerequisite for `localnet-start` to build `localnode` locally for now and see if that helps the build process.

thanethomson: Seems like it did help the build process 😁

melekes: why /etc?

PR #3753, opened 2019-06-25
[ADR - 43] blockchain riri-org

brapse:
The blockchain reactor has multiple high level responsibilities that would be easier to understand/test if seperated. This ADR contains a proposal splitting up the block processing and scheduling requests for fast-sync in order to achieve the proposed API in [ADR-40](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-040-blockchain-reactor-refactor.md).

giphy (2)

* [ ] Wrote tests


docs/architecture/adr-042-blockchain-riri-org.md
```diff
@@ -0,0 +1,290 @@
+# ADR 042: Blockhchain Reactor Riri-Org
+
+## Changelog
+* 18-06-2019: Initial draft
+
+## Context
+
+The blockchain reactor is responsible for two high level processes:sending/receiving blocks from peers and FastSync-ing blocks to catch upnode who is far behind.  The goal of[ADR-40](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-040-blockchain-reactor-refactor.md) was to refactor these twoprocesses by seperating business logic currently wrapped up ingo-channels into pure `handle*` functions.  While the ADR specified what the final formof the reactor might look like it lacked guidance on intermediary stepsto get there.
```
marbar3778:
```suggestion
The blockchain reactor is responsible for two high level processes:sending/receiving blocks from peers and FastSync-ing blocks to catch upnode who is far behind.  The goal of[ADR-40](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-040-blockchain-reactor-refactor.md) was to refactor these two processes by separating business logic currently wrapped up ingo-channels into pure `handle*` functions.  While the ADR specified what the final form of the reactor might look like it lacked guidance on intermediary steps to get there. 
```

docs/architecture/adr-042-blockchain-riri-org.md
```diff
@@ -0,0 +1,290 @@
...
+
+The blockchain reactor is responsible for two high level processes:sending/receiving blocks from peers and FastSync-ing blocks to catch upnode who is far behind.  The goal of[ADR-40](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-040-blockchain-reactor-refactor.md) was to refactor these twoprocesses by seperating business logic currently wrapped up ingo-channels into pure `handle*` functions.  While the ADR specified what the final formof the reactor might look like it lacked guidance on intermediary stepsto get there. 
+The following diagram illustrates the state of the
+
+[blockchain-reorg](https://github.com/tendermint/tendermint/pull/35610),which will be refered to as `v1`
+
+![v1 Blockchain Reactor Architecture
+Diagram](https://github.com/tendermint/tendermint/blob/f9e556481654a24aeb689bdadaf5eab3ccd66829/docs/architecture/img/blockchain-reactor-v1.png)
+
+While `v1` of the blockchan reactor has shown significant improvementsin erms of simplifying the concurrency model, the current PR has runinto few roadblocks.
```
marbar3778:
```suggestion
While `v1` of the blockchain reactor has shown significant improvements in terms of simplifying the concurrency model, the current PR has run into few roadblocks.
```

docs/architecture/adr-042-blockchain-riri-org.md
```diff
@@ -0,0 +1,290 @@
...
+
+* The current PR large and difficult to review.
+* Block gossiping and fast sync processes are highly coupled to the shared `Pool` data structure.
+* Peer communication is spread over multiple components creating complex dependency graph which must be mocked out during testing.
+* Timeouts modeled as stateful tickers introduce non-determinism in tests
+
+## Decision
+
+Partition the responsibilities of the blockchain reactor into a set ofcomponents which communicate exclusively with events. Events will containtimestamps allowing each component to track time as internal state.Internal state will be mutated by a set of `handle*` which will produce event(s).The integration between components will happen in the reactor andreactor tests will then become integration tests between components. This design will be known as `v2`
```
marbar3778:
```suggestion
Partition the responsibilities of the blockchain reactor into a set of components which communicate exclusively with events. Events will contain timestamps allowing each component to track time as internal state. The internal state will be mutated by a set of `handle*` which will produce event(s). The integration between components will happen in the reactor and reactor tests will then become integration tests between components. This design will be known as `v2`
```

docs/architecture/adr-042-blockchain-riri-org.md
```diff
@@ -0,0 +1,292 @@
...
+The processor will be responsible for validating and processing blocks.
+The Processor will maintain an internal cursor `height` of the last
+processed block. As a set of unordered blocks arrive, the the Processor
+will check if it has `height+1` nessary to process the next block.
+
+```go
+type Proccesor struct {
+    height ...
+    state ...
+    blocks [height]*Block
```
melekes: how many blocks will be stored in the worst case?

docs/architecture/adr-042-blockchain-riri-org.md
```diff
@@ -0,0 +1,292 @@
...
+An io handling routine within the reactor will isolate peer communication.
+
+```go
+func (r *BlockchainReacor) ioRoutine(chan ioMsgs, ...) {
+    ...
+    for {
+        msg := <-ioMsgs
+        switch msg := msg.(type) {
+            case scBlockRequestMessage:
+                r.sendBlockToPeer(...)
```
melekes: what if IO is slow? won't this slow down `demuxRoutine` in effect?

docs/architecture/adr-042-blockchain-riri-org.md
```diff
@@ -0,0 +1,292 @@
+# ADR 042: Blockhchain Reactor Riri-Org
```
ebuchman: ADR numbers should be registered in #2313 before being used. Just edit the comment there to add a new one.

docs/architecture/adr-042-blockchain-riri-org.md
brapse: Good point. The worst case would be unbounded. We can look to provide back pressure by counting unprocessed messages and sending `stopScheduling` messages to the `scheduler` when we exceed some threshold. Was this your concern?

docs/architecture/adr-042-blockchain-riri-org.md
brapse: IIRC this goes down results in a `peer.TrySend` which are queued and sent later until the queue fills up. Similar to the comment above, we can use the return value from trySend to send `stopScheduling` messages to the `scheduler` until the queue flushes.

docs/architecture/adr-042-blockchain-riri-org.md
```diff
@@ -0,0 +1,292 @@
...
+    // Delete all unprocessed blocks from peerID
+    for i = height; i < len(blocks); i++ {
+        if peers[i] == peerID {
+            events = append(events, pcBlockReschedule{height})
+            delete block[height]
+        }
+    }
+    return events
+}
+
```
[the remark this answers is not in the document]
brapse: From my understanding, The processor is responsible for removing/rescheduling all unprocessed block received from an errored peer. Is that true?

docs/architecture/adr-042-blockchain-riri-org.md
```diff
@@ -0,0 +1,292 @@
...
+	schedule.touchPeer(peerID, time)
+    // reschedule that block, punish peer...
+    ...
+}
+
+func handleTimeCheckEv(time) {
+	// clean peer list
+
+    events = []
+	for peerID := range schedule.peersTouchedSince(time) {
```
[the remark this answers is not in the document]
brapse: This loop is intended to iterate over peers that have timed out, change their state to timed out and reschedule blocks pending from those timed out peer to other peers.

docs/architecture/adr-042-blockchain-riri-org.md
```diff
@@ -0,0 +1,292 @@
...
+	pending map[height]struct{}
+	height height // max height for the peer
+	state {
+		pending,   // we know the peer but not the height
+		active,    // we know the height
+		timeout    // the peer has timed out
+	}
+}
+```
+
```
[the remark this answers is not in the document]
brapse: Yes I think we the `schedule.peersNotTouchedSince` loop could be done on `schedule.peersSlowerThan(minSpeed)` to prune peers in a more sophisticated way. It could be as simple as that  or if we may eventually need something more sophisticated. We could have a sort of `PeerPruning` strategy which could be applied to the schedule. Not sure if that's needed now or if we need to emit events indicating why the peer was pruned 🤔.

docs/architecture/adr-042-blockchain-riri-org.md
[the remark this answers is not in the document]
ancazamfir: IMHO we need to clarify this. Looks like both the processor and the scheduler maintain some information about the fast sync peers and blocks. They may both detect errors and trigger peer removal and eventually the rescheduling of requests. Crossing messages may cause their view of peers and blocks to get out of sync. At this point the scheduler struct is not filled in, also how the actual scheduling works is missing. I think we need more details on the scheduler and then make the analysis.

docs/architecture/adr-042-blockchain-riri-org.md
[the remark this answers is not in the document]
ancazamfir: Both v0 and v1 have a monitor and remove slow peers. Events are sent to the switch to remove the slow peer. It is probably ok if we want to leave this aside initially for this refactor.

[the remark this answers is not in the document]
brapse: 👍 Yeah I think we can do this without much trouble.

docs/architecture/adr-042-blockchain-riri-org.md
melekes: Yeah. I was thinking if there are any ways for malicious peer (group of peers) to exploit this.

docs/architecture/adr-043-blockchain-riri-org.md
```diff
@@ -0,0 +1,372 @@
+# ADR 043: Blockhchain Reactor Riri-Org
+
+## Changelog
+* 18-06-2019: Initial draft
+
+## Context
+
+The blockchain reactor is responsible for two high level processes:sending/receiving blocks from peers and FastSync-ing blocks to catch upnode who is far behind.  The goal of [ADR-40](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-040-blockchain-reactor-refactor.md) was to refactor these two processes by separating business logic currently wrapped up in go-channels into pure `handle*` functions.  While the ADR specified what the final form of the reactor might look like it lacked guidance on intermediary steps to get there.
```
milosevic: Link to ADR-40 is broken as develop branch is deleted. It should be https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-040-blockchain-reactor-refactor.md.

docs/architecture/adr-043-blockchain-riri-org.md
```diff
@@ -0,0 +1,372 @@
...
+Diagram](https://github.com/tendermint/tendermint/blob/f9e556481654a24aeb689bdadaf5eab3ccd66829/docs/architecture/img/blockchain-reactor-v1.png)
+
+While `v1` of the blockchain reactor has shown significant improvements in terms of simplifying the concurrency model, the current PR has run into few roadblocks.
+
+* The current PR large and difficult to review.
+* Block gossiping and fast sync processes are highly coupled to the shared `Pool` data structure.
+* Peer communication is spread over multiple components creating complex dependency graph which must be mocked out during testing.
+* Timeouts modeled as stateful tickers introduce non-determinism in tests
+
+This ADR is meant to specify the missing components and control nessesary to acheive [ADR-40](https://github.com/tendermint/tendermint/blob/develop/docs/architecture/adr-040-blockchain-reactor-refactor.md).
```
milosevic: typo "achieve". Link to ADR-40 broken as above.

docs/architecture/adr-043-blockchain-riri-org.md
```diff
@@ -0,0 +1,372 @@
...
+	}
+}
+
+func processRoutine(input chan Message, output chan Message) {
+	processor := NewProcessor(..)
+	for {
+		msg := <- input
+		switch msg := msg.(type) {
+			case bcBlockRequestMessage:
+				output <- processor.handleBlockRequest(msg, time.Now())
```
milosevic: Why current time is not appended to a message by demuxRoutine? This might be relevant as we had a scenario where msg was received by a reactor but delayed to be processed for quite some time. Also for debugging and testing purposes might be interesting.

docs/architecture/adr-043-blockchain-riri-org.md
```diff
@@ -0,0 +1,372 @@
...
+}
+```
+
+## Lifecycle management
+
+A set of routines for individual processes allow processes to run in parallel with clear lifecycle management. `Start`, `Stop`, and `AddPeer` hooks currently present in the reactor will delegate to the sub-routines allowing them to manage internal state independant without further coupling to the reactor.
+
+```go
+func (r *BlockChainReactor) Start() {
+	r.msgs := make(chan Message, maxInFlight)
```
milosevic: What this message buffer corresponds to?

docs/architecture/adr-043-blockchain-riri-org.md
```diff
@@ -0,0 +1,372 @@
...
+func (r *BlockchainReactor) Stop() {
+	...
+	r.msgs <- stop
+	...
+}
+...
+
+func (r *BlockchainReactor) AddPeer(peer p2p.Peer) {
+	...
+	r.msgs <- bcAddPeerEv{peer.ID}
```
milosevic: Why we write here to a buffer which is used as an outgoing buffer for routines?

docs/architecture/adr-043-blockchain-riri-org.md
```diff
@@ -0,0 +1,372 @@
...
+	...
+}
+
+```
+
+## IO handling
+An io handling routine within the reactor will isolate peer communication.
+
+```go
+func (r *BlockchainReacor) ioRoutine(chan ioMsgs, ...) {
```
milosevic: ioRoutine has different signature above, i.e., it assumes outgoing channel. We directly call sendX on reactor here.

docs/architecture/adr-043-blockchain-riri-org.md
```diff
@@ -0,0 +1,372 @@
...
+	...
+	r.msgs <- bcAddPeerEv{peer.ID}
+	...
+}
+
+```
+
+## IO handling
+An io handling routine within the reactor will isolate peer communication.
+
```
milosevic: Can we add a bit more info about responsibilities of this routine. It feels like it send messages (created by other routines?) and inform switch to stop peer upon error (detected by other routines?).

docs/architecture/adr-043-blockchain-riri-org.md
```diff
@@ -0,0 +1,372 @@
...
+			case bcFinished
+				break;
+		}
+	}
+}
+
+```
+### Processor internals 
+
+The processor will be responsible for validating and processing blocks. The Processor will maintain an internal cursor `height` of the last processed block. As a set of blocks arrive unordered, the the Processor will check if it has `height+1` nessary to process the next block. The processor also maintains the map `blockPeers` of peers to height, to keep track of which peer provided the block at `height`. `blockPeers` can be used in`handleRemovePeer(...)` to reschedule all unprocessed blocks provided by a peer who has errored.
```
milosevic: `height` seems to be the next block that should be processed and `height-1` the last processed block. Typo two "the" in the second sentence.

docs/architecture/adr-043-blockchain-riri-org.md
```diff
@@ -0,0 +1,372 @@
...
+	if blocks[height] && blocks[height+1] {
+		... = state.Validators.VerifyCommit(...)
+		... = store.SaveBlock(...)
+		state, err = blockExec.ApplyBlock(...)
+		...
+		if err == nil {
+			delete blocks[height]
+			height++
+			lastTouch = time
+			return pcBlockProcessed{height}
```
milosevic: What if there are several blocks that can be processed? Should we have a loop here or have external mechanism to trigger this check?

docs/architecture/adr-042-blockchain-riri-org.md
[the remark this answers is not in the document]
milosevic: > The processor is responsible for removing/rescheduling all unprocessed block received from an errored peer. Is that true?
One way to think about Processor is that it should order, verify and execute blocks. If it fails by doing this for some block it will notify scheduler (and io) about error so they can do corresponding actions.

milosevic:
In general looks good to me. I think it gives sufficient details on how to proceed. Then some details or concerns can be addressed on the way when we focus on concrete routine.

docs/architecture/adr-043-blockchain-riri-org.md
brapse: This is a very good point! It seems like the demuxRoutine should be the sole component responsible for time ensuring a consistent view across *routines. One consequence of this change is that we need to ensure that the we channels are never blocked. If the demuxRoutine is responsible for dispatching correct timed messages to other channels, a block channel would block and starve the other channels for messages. We can do this by having a `soft limit` on processes that institute back pressure before their `hard limit` channel size is reached. I've added some inline comments to ensure we don't miss this during implementation.

docs/architecture/adr-043-blockchain-riri-org.md
brapse: This will be the channel which funnels messages from the `Receive` method to the `demuxRoutine`.

docs/architecture/adr-042-blockchain-riri-org.md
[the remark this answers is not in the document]
ancazamfir: Could you add a comment on how this is handled in the IoRoutine?

docs/architecture/adr-043-blockchain-riri-org.md
brapse: For some concrete hints about how we can handle this, one technique we can use (which we also used in `v1`) is [overflowing channel](https://stackoverflow.com/questions/25657207/golang-how-to-know-a-buffered-channel-is-full)

docs/architecture/adr-043-blockchain-riri-org.md
brapse: Sorry if that is not clear.This "buffer" would be used as the input for the routines. Out of curiosity,  what would be the motivation for calling this a buffer here instead of channel?

brapse:
Documented the pending concern about processing invalidated blocks in #3777. Looks like this is ready to merge 👍

docs/architecture/adr-042-blockchain-riri-org.md
jaekwon: We shouldn't be using len() on channels to "provide backpressure".  It should only be used in certain cases... (1) when you're the only reader routine, and you want to ensure that there exists something in the channel before trying something, but otherwise not pull from it.  Even still, a value of len()=0 does not mean that a pull will fail.  (2) for gathering statistics that does not affect the logic.  Even if there was a save way to ensure that there is space in a channel before pushing to it, such a method would not be safe in the face of concurrent pushes to that channel.  In other words, even if it works today, it is brittle and can easily break for a variety of reasons in the future.

https://groups.google.com/forum/#!topic/golang-nuts/yQw1Wx6BoUU

The only good general way to deal with fullness of channels is to use select{default:}.

brapse: This design has evolved with the implementation to resolve a bunch of these open questions. In particular we explicitly use [priority queue](https://github.com/tendermint/tendermint/blob/master/blockchain/v2/routine.go#L22] within the each which is now bounded.

jaekwon: I'm not seeing anything that is bounding the size of the queue.  bufferSize is used as a hint in constructing a priority queue.  The `// XXX backpressure` comment and issue still exists in reactor.go.

Also, https://github.com/tendermint/tendermint/issues/4142

PR #3512, opened 2019-03-30
Bucky/mempool txsmap

ebuchman:
Addressing #3509

* [ ] Updated all relevant documentation in docs
* [x] Updated all code comments where relevant
* [x] Wrote tests
* [x] Updated CHANGELOG_PENDING.md

PR #3758, opened 2019-06-27
abci: minor cleanups in the socket client

### LINKED ISSUE #3513: abci: minor cleanups in the socket client
Follow up from https://github.com/tendermint/tendermint/pull/3512

Specifically:
- `cli.conn.Close()` need not be under the mutex (https://github.com/tendermint/tendermint/pull/3512#discussion_r270639920)
- call the reqRes callback after the resCb so they always happen in the same order (https://github.com/tendermint/tendermint/pull/3512/#discussion_r270640328)

melekes:
Follow up from #3512

Specifically:

    cli.conn.Close() need not be under the mutex (#3512 (comment))
    call the reqRes callback after the resCb so they always happen in the same order (#3512)

Fixes #3513

* [ ] ~~Wrote tests~~


abci/client/socket_client.go
```diff
@@ -82,15 +78,12 @@ RETRY_LOOP:
 }
 
 func (cli *socketClient) OnStop() {
-	cli.BaseService.OnStop()
```
marbar3778: should we also remove this function and `BaseService.OnStart()` as they don't do anything, then they can be removed elsewhere?

melekes: These functions act as defaults and sometimes overwritten (e.g. in BaseReactor). But we can safely remove BaseService.OnStart and OnStop calls (what I did here).

marbar3778: 👍 ah okay

melekes:
> Also, does this warrant a (pending) change log entry?

I'd say no

ebuchman:
> > Also, does this warrant a (pending) change log entry?
> 
> I'd say no

The re-ordering change here might be significant enough to warrant a changelog entry

PR #3769, opened 2019-07-02
[ADR - 42] State-sync

marbar3778:
Initial State-sync discussions to be held in this PR. This is not the final stage of the PR. 

* [ ] Wrote tests


docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,119 @@
...
+  the state or block header.
+
+### Next Steps
+
+To move forward with a light-client secure state sync based on key-ranges, here
+are some next steps:
+
+- The state sync reactor should follow the same architecture as the
+  new blockchain reactor ala [#3753](https://github.com/tendermint/tendermint/pull/3753).
+- The proposal in issue #3639 should be fleshed out and written up more formally as an ADR (ie.
```
AdityaSripal: Make issue link here

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,119 @@
...
+  in this one)
+- Need to address the initial light client sync (this will depend on a working
+  light client!)
+- Need to address assembling the Tendermint `State` object
```
AdityaSripal: >  - Need to address assembling the Tendermint `State` object

From my understanding this means trying to figure out how to take TM state chunks and  resurrect the TM `State` object?

If so can we change to:
 ```suggestion
- Need to address assembling the Tendermint `State` object from `StateChunk` objects
```

File: docs/architecture/adr-042-state-sync.md [hunk 3]
```diff
@@ -0,0 +1,119 @@
...
+range are ommitted, proofs for the range will fail to verify.
+This way a node can start by requesting the entire tree from one peer,
+and that peer can respond with say the first few keys, and the ranges to request
+from other peers.
+
+Finally, there has been significant work over the last few months rearchitecting
+the blockchain reactor to make it more easily testable and verifiable. This has
+culminated in a most recent design,
+[#3575](https://github.com/tendermint/tendermint/pull/3753),
+where the reactor is split into three indpendent routines: one for IO with
```
AdityaSripal:
```suggestion
where the reactor is split into three independent routines: one for IO with
```

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,119 @@
...
+just download a recent application state. Nodes would first perform a
+light-client sync to find a recent height for which they could sync the state.
+Then they would request verifiable pieces of that state concurrently from their peers.
+Once they have a complete state, they can use "Block Sync" to download
+and execute any more recent blocks.
+
+Note that Tendermint does not actually know anything about the application
+state; it is only available over ABCI. State Sync will thus require new ABCI
+methods and/or connections, and will need to support arbitrary Merkle trees.
+
```
AdityaSripal: Suggested addition:

Tendermint is responsible for downloading and verifying chunks of `AppState` from peers. ABCI Application is responsible for taking `AppStateChunk` objects from TM and constructing a valid state tree whose root corresponds with the `AppHash` of syncing block.

Aside:

This is only true of `AppStateChunks` correct? Any `BlockChunk` or `StateChunk` that gets sent by peers can be handled entirely in TM

AdityaSripal:
Would be useful to have a `Security Models` section that compares the Warp-Sync security model with a standard light-client security model. Discussion on how choosing one security model or the other also has trade-offs in performance/simplicity/etc. would be helpful

docs/architecture/adr-042-state-sync.md
ebuchman: Not necessarily. Note this is a much smaller object than what we have to sync from the application. Also it's possible that the light client sync will give us everything we need here (ie. the validator set and consensus params). So we might not actually have to sync much else ...

docs/architecture/adr-042-state-sync.md
ebuchman: Right

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,119 @@
...
+[some concerns](https://docs.google.com/document/d/1npGTAa1qxe8EQZ1wG0a0Sip9t5oX2vYZNUDwr_LVRR4/edit)
+about this design, somewhat specific to the IAVL tree, and mainly concerning
+performance of random reads and of iterating through the tree to determine element numbers
+(ie. elements aren't indexed by the element number).
+
+This design could be generalized to have numbered chunks instead of elements,
+where chunks are indexed by their chunk number and determined by some algorithm
+for grouping the elements in the tree. This would allow pre-computation and
+persistence of the chunks that could alleviate the concerns about real-time iteration and random reads.
+However, there seems to be a larger issue with the "numbered chunks" approach:
```
ebuchman: After further thought it seems definitely possible to do this, but it might require that the first few chunks come in order and provide information about how the rest of the chunks are indexed. For instance the first chunk might provide the first 5 layers of the tree, and then there might be 32 more chunks, one for each subtree. Those would then be verifiable against their index without any more commitments.

ebuchman: cc https://github.com/tendermint/iavl/issues/135

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,209 @@
...
+* What are the performance characteristics
+    * Random vs sequential reads
+    * How parallelizeable is the scheduling algorithm
+
+### Proposals
+Broadly speaking there are two approaches to this problem which have had
+varying degrees of discussion and progress. These approach can be
+summarized as:
+
+* Lazy: Where snapshots are produced dynamically at request time
```
ebuchman: Might be helpful to clarify that `Lazy` leans towards using the native interface of the tree, ie. key-value pairs, where as `Eager` leans towards constructing something more specifically for state syncing purposes, ie. chunks.

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,209 @@
...
+based on their initial implementation in
+[PR #3243](https://github.com/tendermint/tendermint/pull/3243)
+and [some learnings](https://docs.google.com/document/d/1npGTAa1qxe8EQZ1wG0a0Sip9t5oX2vYZNUDwr_LVRR4/edit).
+Note this still requires the honest majority peer assumption.
+One major advantage of the warp-sync approach is that chunks can be compressed before sending.
+
+Ideally State Sync in Tendermint would not require an honest majority of peers, and instead
+would rely on the standard light client security, meaning state chunks would be
+verifiable against the merkle root of the state. Such a design for tendermint
+was originally tracked in [#828](https://github.com/tendermint/tendermint/issues/828).
```
ebuchman: More like, `Note this property tends to come more naturally to the Lazy approach since it tends to use the existing structure of the tree (ie. keys or nodes) rather than state-sync specific chunks.`

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,209 @@
...
+### Analysis of Lazy vs Eager
+
+Lazy vs Eager have more in common than they differ. They all require
+reactors on the tendermint side, a set of ABCI messages and a method for
+serializing/deserializing snapshots facilitated by a SnapshotFormat.
+
+The biggest difference between Lazy and Eager proposals is in the
+read/write patterns necessitated by when serving a snapshot chunk.
+Specifically, Lazy State Sync random reads to the underlying data
+structure while Eager can optimize for sequential reads.
```
ebuchman: Heh, I think this might only be true for poorly implemented Merkle trees (ie. most of them :P). I believe this might be what Matt was alluding to here: https://github.com/tendermint/tendermint/issues/3639#issuecomment-491531005

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,209 @@
...
+follow up implementation of [Warp Sync](http://github.com/tendermint/tendermint/pull/3594).
+
+One proposed difference between Lazy and Eager state sync is the safety
+model. It had been suggested that Lazy State Sync could use light client
+validation while Warp Sync was unsuitable for public networks as it
+relies on the majority of peers to produce a manifest. This manifest
+would require complete syncing before validation of the data against the
+merkle root were possible. This could create DDOS attack vector in which
+a peer publishes an invalid manifest and making peers download invalid
+data. For this reason, WarpSync is faster but less safety-efficient; It
```
ebuchman:
```suggestion
data. For this reason, WarpSync is faster but less safety-efficient;
```

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,209 @@
...
+as it takes longer to realize that data is invalid.
```
ebuchman:
```suggestion
it takes longer to realize that data is invalid.
```

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,209 @@
...
+
+However, upon analysis of the approaches with the orthogonal axis of
```
ebuchman: Might help if this axes was put higher up to paint the landscape. And I think its "eager/lazy" by "light-client/honest-peers".

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,209 @@
...
+eager/lazy and chunk/snapshot validation suggest that a version of
+WarpSync with per chunk light client validation might get the optimal
```
ebuchman: Maybe something like:

```
While warp sync in general does not permit continuous light-client validation of chunks
(since chunks are typically agnostic of underlying tree structure), 
it may be possible to provide this by taking care to chunk the state in a way that preserves 
enough structure of the tree to allow chunks to be fully validated on receipt. 
Thus an Eager approach that can still harness light-client validation may get the optimal performance and safety.
```

ebuchman: Are we really sure that other implementations of warp sync don't let you validate chunks until the very end?

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,209 @@
...
+### Negative
+* Static Scheduling lacks opportunity for real time chunk availability optimizations
+
+## References
+[sync: Sync current state without full replay for Applications](https://github.com/tendermint/tendermint/issues/828) - original issue
+[tendermint state sync proposal](https://docs.google.com/document/d/15MFsQtNA0MGBv7F096FFWRDzQ1vR6_dics5Y49vF8JU/edit?ts=5a0f3629) - Cloudhead proposal
+[tendermint state sync proposal 2](https://docs.google.com/document/d/1npGTAa1qxe8EQZ1wG0a0Sip9t5oX2vYZNUDwr_LVRR4/edit) - ackratos proposal
+[proposal 2 implementation](https://github.com/tendermint/tendermint/pull/3243)  - ackratos implementation
+[WIP General/Lazy State-Sync pseudo-spec](https://github.com/tendermint/tendermint/issues/3639) - Jae Proposal
+[Warp Sync Implementation](https://github.com/tendermint/tendermint/pull/3594) - ackratos
```
ackratos: we have some more stuff:
1. document for this enhancement: https://github.com/binance-chain/BEPs/pull/18, included more chunking strategy we adopted
2. to try out state sync in binance chain prod and testnet: https://github.com/binance-chain/docs-site/blob/master/docs/fullnode.md#state-sync

docs/architecture/adr-042-state-sync.md
milosevic: Agree with Bucky here. It's probably not lazy vs eager, but generating snapshot (potentially different data structure) vs serving requests from original data store. And then in the former case we can differentiate between lazy and eager, where lazy would generate snapshot only when asked.

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,209 @@
...
+about this design, somewhat specific to the IAVL tree, and mainly concerning
+performance of random reads and of iterating through the tree to determine element numbers
+(ie. elements aren't indexed by the element number).
+
+This design could be generalized to have numbered chunks instead of
+elements, where chunks are indexed by their chunk number and determined
+by some algorithm for grouping the elements in the tree. This would
+allow pre-computation and persistence of the chunks that could alleviate
+the concerns about real-time iteration and random reads.  However, there
+seems to be a larger issue with the "numbered chunks" approach: it does
```
milosevic: Maybe this sentence should be removed as in the meantime it turns out that this is possible.

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,209 @@
...
+Hence, to go with this chunking approach to state sync, information
+about how state is chunked would need to be hashed into the state
+itself, or into the block header, so that chunk numbers could be
+directly verified.
+
+An alternative design was suggested by Jae Kwon in
+[#3639](https://github.com/tendermint/tendermint/issues/3639) where chunking
+happens lazily and in a dynamic way: nodes request key ranges from their peers,
+and peers respond with some subset of the
+requested range and with notes on how to request the rest in parallel from other
```
milosevic: This part of this design is super tricky to implement correctly in my view as we ask set of peers and each of them come back with notes how to download rest.

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,209 @@
...
+out the state is incorrect until you download the whole thing and compare it
+with a verified block header.
+
+A similar solution was implemented by Binance in
+[#3594](https://github.com/tendermint/tendermint/pull/3594)
+based on their initial implementation in
+[PR #3243](https://github.com/tendermint/tendermint/pull/3243)
+and [some learnings](https://docs.google.com/document/d/1npGTAa1qxe8EQZ1wG0a0Sip9t5oX2vYZNUDwr_LVRR4/edit).
+Note this still requires the honest majority peer assumption.
+One major advantage of the warp-sync approach is that chunks can be compressed before sending.
```
milosevic: Is this really the feature of warp-sync (only)? What prevents you from compressing chunks that corresponds to key ranges for example?Unless I am missing something, compression is property of communication protocol.

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,209 @@
...
+
+Ideally State Sync in Tendermint would not require an honest majority of peers, and instead
```
milosevic: Maybe it's worth being more precise here and say that in public systems we normally don't make assumptions about ratio of faulty processes among node peers. This might be reasonable assumption in enterprise blockchain systems, although even in that case I would challenge that assumption. For example, what if system tolerates 10 faulty processes (f=10, n=31), and you have most of those processes in node peer set (for example peer set size is also 10).

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,209 @@
...
+read/write patterns necessitated by when serving a snapshot chunk.
+Specifically, Lazy State Sync random reads to the underlying data
+structure while Eager can optimize for sequential reads.
+
+This distinctin between approaches was demonstrated by Binance's [ackratos](https://github.com/ackratos) in their
+implementation of [Lazy State sync](https://github.com/tendermint/tendermint/pull/3243), The [analysis](https://docs.google.com/document/d/1npGTAa1qxe8EQZ1wG0a0Sip9t5oX2vYZNUDwr_LVRR4/) of the performance, and
+follow up implementation of [Warp Sync](http://github.com/tendermint/tendermint/pull/3594).
+
+One proposed difference between Lazy and Eager state sync is the safety
+model. It had been suggested that Lazy State Sync could use light client
```
milosevic: I am not sure to agree with the first sentence. You could use light client validation in both approaches.

docs/architecture/adr-042-state-sync.md
milosevic: WarpSync is more efficient compared to what?

docs/architecture/adr-042-state-sync.md
milosevic: I don't understand what is meant with WarpSync approach here. Maybe it's better explaining the properties of what we see as an optimal design without referring to existing approaches with different designs as this just creates confusion. For example, optimal design would rely on light client verification for verifying root hash of the snapshot state and provide possibility for chunk based validation with respect to the root hash.

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,209 @@
...
+However, upon analysis of the approaches with the orthogonal axis of
+eager/lazy and chunk/snapshot validation suggest that a version of
+WarpSync with per chunk light client validation might get the optimal
+performance and safety.
+
+## Decision: WarpSync With Per Chunk Light Client Validation
+
+The conclusion after thorough concideration of the
+advantages/disadvances of eager/lazy is to produce a state sync which
+replicates the performance IO characteristic of WarpSync and the security of
```
milosevic: I would also avoid referring to WarpSync here.

docs/architecture/adr-042-state-sync.md
brapse: Yeah in the case of [alternative trees](https://github.com/nomic-io/merk) the implementation might provide good range scan characteristics which would alleviate any need to materialize a snapshot on disk. Since we've taken the approach to define a set of expectations by way of interface specification and push as many material concerns down to the application, we need not specify where/how the snapshots are stored. I think/hope that our current design supports the ability for an ABCI application to serve chunk requests directly from the tree if they so chose, we only insist that the the indexing of chunks be consistent across peers.

brapse:
> Would be useful to have a `Security Models` section that compares the Warp-Sync security model with a standard light-client security model. Discussion on how choosing one security model or the other also has trade-offs in performance/simplicity/etc. would be helpful

https://github.com/tendermint/tendermint/blob/834ecd53da78433dd4e0a6a30944d7b964ae0444/docs/architecture/adr-042-state-sync.md#compairing-security-models

Is this what you had in mind @AdityaSripal ?

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,225 @@
...
+### Implementation Question
+
+* What is the format of a snapshot
+    * Complete snapshot 
+    * Ordered IAVL key ranges
+    * Compressed individually chunks which can be validated
+* How is data validated
+    * Trust a peer with it's data blindly
+    * Trust a majority of peers
+    * Use light client validation to validate any key based on consensus
```
AdityaSripal:
```suggestion
    * Use light client validation to validate each chunk against consensus
```

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,225 @@
...
+peers download invalid data. For this reason, WarpSync is faster but
+less safety-efficient; It as it takes longer to realize that data is
+invalid.
+
+However, upon analysis of the approaches with the orthogonal axis of
+eager/lazy and chunk/snapshot validation suggest that a version of
+WarpSync with per chunk light client validation might get the optimal
+performance and safety.
+
+## Decision: Eager StateSync With Per Chunk Light Client Validation
```
AdityaSripal: Would this require bringing manifest hash into consensus?

AdityaSripal: Will we only require validator nodes to compute and validate this hash?

docs/architecture/adr-042-state-sync.md
ackratos: this could be rephrased to `One major advantage of the warp-sync approach is that chunks can be prepared before sending.` The point is this approach is eager computation of what's going to serve compared with lazily computed for each single request

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,225 @@
...
+
+This distinctin between approaches was demonstrated by Binance's [ackratos](https://github.com/ackratos) in their
+implementation of [Lazy State sync](https://github.com/tendermint/tendermint/pull/3243), The [analysis](https://docs.google.com/document/d/1npGTAa1qxe8EQZ1wG0a0Sip9t5oX2vYZNUDwr_LVRR4/) of the performance, and
+follow up implementation of [Warp Sync](http://github.com/tendermint/tendermint/pull/3594).
+
+#### Compairing Security Models
+One proposed difference between Lazy and Eager state sync is the
+security model. It had been suggested that Lazy State Sync could use
+light client validation while Warp Sync was unsuitable for public
+networks as it relies on the majority of peers to produce a manifest.
```
melekes: not necessary. putting manifest hash into block's header solves this problem. A separate manifest commitment server, where validators commit (sign manifest hash) to particular manifest hash + height (kinda like https://sum.golang.org/) also does the job I believe.

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,225 @@
...
+  and the ABCI application
+* A set of ABCI Messages
+* Design SnapshotFormat as an interface which can:
+    * validate chunks
+    * read/write chunks from file
+    * read/write chunks to/from application state store
+    * convert manifests into chunkRequest ABCI messages
+* Implement SnapshotFormat for cosmos-hub with concrete implementation for:
+    * read/write chunks in a way which can be:
+        * parallelized accros peers
```
melekes:
```suggestion
        * parallelized accross peers
```

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,225 @@
...
+* A Storage format with validation and performance considerations - ABCI application
+
+### Implementation Properties
+Beyond the approach, any implementation of StateSync can be evaluated
+across different criteria:
+
+* Speed: Expected throughput of producing and consuming snapshots
+* Safety: Cost of pushing invalid snapshots to a node
+* Liveness: Cost of preventing a node from receiving/constructing a snapshot
+* Effort: How much effort does an implementation require
```
cwgoes: Also versatility - whether any additional requirements are imposed upon the ABCI application about e.g. storage layout.

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Implementation Question
+
+* What is the format of a snapshot
+    * Complete snapshot 
+    * Ordered IAVL key ranges
+    * Compressed individually chunks which can be validated
+* How is data validated
```
cwgoes: Are these intended to be mutually exclusive? There's a continuum w.r.t. validation of subcomponents of state, e.g. after receiving a few chunks, validate one at random & ban the peer if invalid, etc.

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,225 @@
...
+
+This design could be generalized to have numbered chunks instead of
+elements, where chunks are indexed by their chunk number and determined
+by some algorithm for grouping the elements in the tree. This would
+allow pre-computation and persistence of the chunks that could alleviate
+the concerns about real-time iteration and random reads.  However, there
+seems to be a larger issue with the "numbered chunks" approach: it does
+not seem possible, in general, to verify that the chunk corresponds to
+the chunk number, even if it can be verified that the chunk is a valid
+set of nodes in the tree. A malicious peer could provide a chunk that
```
cwgoes: What if there is a deterministic function from subsets of key spaces to the natural numbers (chunk numbers)?

Then the downloading node can check the keys they've received in the chunk match the number.

It requires a bit of prior agreement on the function but I don't see why that wouldn't work?

It could even vary; e.g. the client could pick a modulus `n` and the chunk could be the hash of the key mod `n`.

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,225 @@
...
+itself, or into the block header, so that chunk numbers could be
+directly verified.
+
+An alternative design was suggested by Jae Kwon in
+[#3639](https://github.com/tendermint/tendermint/issues/3639) where chunking
+happens lazily and in a dynamic way: nodes request key ranges from their peers,
+and peers respond with some subset of the
+requested range and with notes on how to request the rest in parallel from other
+peers. Unlike chunk numbers, keys can be verified directly. And if some keys in the
+range are ommitted, proofs for the range will fail to verify.
```
cwgoes:
```suggestion
range are omitted, proofs for the range will fail to verify.
```

docs/architecture/adr-042-state-sync.md
```diff
@@ -0,0 +1,225 @@
...
+This way a node can start by requesting the entire tree from one peer,
+and that peer can respond with say the first few keys, and the ranges to request
+from other peers.
+
+#### Eager StateSync
+Warp Sync as implemented in Parity
+["Warp Sync"](https://wiki.parity.io/Warp-Sync-Snapshot-Format.html) to rapidly
+download both blocks and state snapshots from peers. Data is carved into ~4MB
+chunks and snappy compressed. Hashes of snappy compressed chunks are stored in a
+manifest file which co-ordinates the state-sync. Obtaining a correct manifest
```
cwgoes: That's lame. We should be able to verify random subsets of the state (random keys) pretty cheaply by keeping light client headers and requesting Merkle proofs occasionally, then ban any peer which sends invalid state, and recheck the whole state at the end.

cwgoes: Does Parity use this manifest file instead of a SPV Ethereum client?

docs/architecture/adr-042-state-sync.md
cwgoes: Chunks will contain some parts of underlying tree structure, right (keys & values)? As long as the client & server agree on the chunking algorithm I don't see why random parts of each chunk (key-value pairs) can't be validated occasionally, but maybe I'm missing something.

docs/architecture/adr-042-state-sync.md
marbar3778: across**

docs/architecture/adr-042-state-sync.md
brapse: No it would not, the application state would be validated against the merkle root produced from light client sync.

docs/architecture/adr-042-state-sync.md
brapse: Would you say that could somehow be captured by engineering effort or should it be a distinct dimension?

docs/architecture/adr-042-state-sync.md
brapse: I suspect this conversation continued in #3799

docs/architecture/adr-042-state-sync.md
brapse: Not at all. exactly as you state, the chunking scheme will have some impact on how soon data can be validated and peers banned.

docs/architecture/adr-042-state-sync.md
brapse: Parity sync module takes a hybrid approach which will block header verification at the end but will sync intermediate state from snapshots first if WarpSync is enabled. 

The general strategy is outlined here:
https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/chain/mod.rs#L21-L34

Snapshots are vetted by rules and then taken from majority of peers here:
https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/chain/mod.rs#L800-L828

PR #3770, opened 2019-07-02
Infrastructure for reproducible builds

alessio:
This change set introduces support for building tendermint with gitian
on the following GOOS/GOARCH pairs:

- darwin/386
- darwin/amd64
- linux/386
- linux/amd64
- linux/arm
- linux/arm64
- windows/386
- windows/amd64

scripts/gitian-descriptors/ contains gitian descriptor files.

scripts/gitian-keys/ contains:

* a keys.txt file that is meant to list core developers and gitian builders PGP keys.
* README.me to provide instructions on how to import the keys into one's
personal GPG keyring.

--

* [ ] Wrote tests


alessio:
To test this just run:

```
./scripts/gitian-build.sh all
```

.circleci/config.yml
```diff
@@ -364,6 +389,12 @@ workflows:
       - upload_coverage:
           requires:
             - test_cover
+      - reproducible_builds:
+          filters:
+            branches:
+              only:
+                - master
+                - develop
```
melekes: we don't have develop anymore

.circleci/config.yml
```diff
@@ -331,6 +331,31 @@ jobs:
...
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run:
+          name: Build tendermint
+          no_output_timeout: 20m
```
melekes: Does this mean circleci could hang for 20 min?

.circleci/config.yml
```diff
@@ -331,6 +331,31 @@ jobs:
...
             docker logout
 
+  reproducible_builds:
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
```
melekes: "We now offer Docker Layer Caching on CircleCI 2.0 for an additional fee. "

Are we paying CircleCI for this?

scripts/gitian-build.sh
```diff
@@ -0,0 +1,201 @@
...
+g_cached_go_tarball=''
+g_sign_identity=''
+g_sigs_dir=''
+g_flag_commit=''
+
+
+f_help() {
+  cat >&2 <<EOF
+Usage: $(basename $0) [-h] PLATFORM
+Launch a gitian build from the current source directory for the given PLATFORM.
```
melekes: Isn't there a separate tool for this? I'd imagine many people interested in a tool like this. Maybe rewritten in Golang instead of bash.

.circleci/config.yml
alessio: It would not hang, it just does not produce output while building. The build may last at most 20m, circleci default value is 10m. I could do some work upstream for this, though it will take some time.

.circleci/config.yml
alessio: No idea. CC'ing @greg-szabo

scripts/gitian-build.sh
alessio: I made alternative versions in both Golang and Python. I opted for a shell script because it ended up being the most coincise among all the alternatives. In the end what we needed was scripting away the execution of a bunch of instructions, and a scripting language appeared to be the right tool for the task at hand.

.circleci/config.yml
alessio: removed thanks

There is another occurrence here:

```
363       - deploy_docs:
364           filters:
365             branches:
366               only:
367                 - master
368                 - develop
```

Should I remove that too?

scripts/gitian-build.sh
alessio: Note that bitcoin made up a python script for this. The thing is: Bash is available anywhere, in many platforms Python needs to be installed.

.circleci/config.yml
melekes: Yes please

.circleci/config.yml
melekes: No need

scripts/gitian-build.sh
melekes: Ok, thanks!

melekes:
@greg-szabo do you have time to look at this?

.circleci/config.yml
marbar3778: @ebuchman can you check? I believe Greg is on vacation

jackzampolin:
@greg-szabo is on vacation for 2 weeks. Lets not block this PR on him.

.circleci/config.yml
```diff
@@ -331,6 +331,31 @@ jobs:
...
+      - run:
+          name: Build tendermint
+          no_output_timeout: 20m
+          command: |
+            sudo apt-get install -y ruby
```
melekes: can't find where ruby is being used

.circleci/config.yml
melekes: > No idea

Why did you add it then?

.circleci/config.yml
alessio: `gitian` is written in ruby:
https://github.com/devrandom/gitian-builder

.circleci/config.yml
alessio: If the feature is enabled, it would be used. Else it's a no-op.

.circleci/config.yml
alessio: done

alessio:
@mircea-c any thoughts on this?

.circleci/config.yml
```diff
@@ -331,6 +331,31 @@ jobs:
             docker push "tendermint/tendermint"
             docker logout
 
+  reproducible_builds:
+    <<: *defaults
```
mircea-c: This image is set to `circleci/golang`. Since this updates automatically, you might have issues in the future. The image should be locked down to a version.

.circleci/config.yml
mircea-c: ```You must open a support ticket to have a CircleCI Sales representative enable Docker Layer Caching on your circleci.com account for an additional fee.```

I doubt anyone did this as we never looked at using the feature.

.circleci/config.yml
```diff
@@ -331,6 +331,31 @@ jobs:
...
+            for os in darwin linux windows; do
+              cp gitian-build-${os}/result/tendermint-${os}-res.yml .
+              rm -rf gitian-build-${os}/
+            done
+      - store_artifacts:
+          path: /go/src/github.com/tendermint/tendermint/tendermint-darwin-res.yml
+      - store_artifacts:
+          path: /go/src/github.com/tendermint/tendermint/tendermint-linux-res.yml
+      - store_artifacts:
+          path: /go/src/github.com/tendermint/tendermint/tendermint-windows-res.yml
```
mircea-c: Why not store the binaries here also? They can be manually downloaded and pushed to github afterwards.

.circleci/config.yml
```diff
@@ -364,6 +388,11 @@ workflows:
       - upload_coverage:
           requires:
             - test_cover
+      - reproducible_builds:
+          filters:
+            branches:
+              only:
+                - master
```
mircea-c: Are we not doing builds for release branches?

mircea-c:
Left some comments on the circle config. The rest LGTM.

.circleci/config.yml
melekes: We should. Master is being used for development now. `vX` are stable branches

.circleci/config.yml
alessio: It does not really affect this build step. No Go compilation takes place here, everything happens in a docker container that is created and run by `./scripts/gitian-build.sh`

.circleci/config.yml
alessio: Brilliant, will certainly do

alessio: Done, can you double check please?

.circleci/config.yml
alessio: Will amend accordingly, thanks

alessio: Done

PR #3730, opened 2019-06-18
[node] allow registration of custom reactors while creating node

ParthDesai:
* [ ] Updated all relevant documentation in docs
* [x] Updated all code comments where relevant
* [ ] Wrote tests
* [x] Updated CHANGELOG_PENDING.md

PR #3771, opened 2019-07-03
node: allow registration of custom reactors while creating node

melekes:
See https://github.com/tendermint/tendermint/pull/3730#issuecomment-503073088

Replaces https://github.com/tendermint/tendermint/pull/3730

* [x] Wrote tests


melekes:
\cc @ParthDesai

ebuchman:
> It's great how this solution doesn't break the Golang API 

Truly fantastic.

> Should this be added into the docs somewhere?

Good question. I don't think we have explicit documentation for the Go APIs other than comments in the code. We should probably write a doc.go for the `node` package to explain more about how everything works and how to make customize more.

PR #3779, opened 2019-07-09
allow using a custom http client in rpc client

### LINKED ISSUE #2010: rpc: allow setting a Timeout for client requests
Refs https://github.com/tendermint/tendermint/issues/1772#issuecomment-405572697

```
        // Timeout specifies a time limit for requests made by this
        // Client. The timeout includes connection time, any
        // redirects, and reading the response body. The timer remains
        // running after Get, Head, Post, or Do return and will
        // interrupt reading of the Response.Body.
        //
        // A Timeout of zero means no timeout.
        //
        // The Client cancels requests to the underlying Transport
        // using the Request.Cancel mechanism. Requests passed
        // to Client.Do may still set Request.Cancel; both will
        // cancel the request.
        //
        // For compatibility, the Client will also use the deprecated
        // CancelRequest method on Transport if found. New
        // RoundTripper implementations should use Request.Cancel
        // instead of implementing CancelRequest.
        Timeout time.Duration
```

gracenoah:
fixes #2010

This also fixes another issue I had: We would like to be able to close all http connections between tests by calling `client.Transport.(*http.Transport).CloseIdleConnections()` but the http client is created inside the various constructors and not exposed anywhere. It seemed much better to allow users to set an entirely custom http client than to try to predict all possible options that might need to be set on it.

I tried to preserve backwards compatibility, which is why I used an options pattern in NewJSONRPCClient and NewHTTP.

If you don't like the use of the options pattern on NewHTTP, consider the PR with just the first commit.

If you prefer different naming, let me know. I know the names got pretty ugly.


rpc/client/rpc_test.go
```diff
@@ -24,6 +25,11 @@ func getHTTPClient() *client.HTTP {
 	return client.NewHTTP(rpcAddr, "/websocket")
 }
 
+func getHTTPClientCustom() *client.HTTP {
```
golangcibot: `getHTTPClientCustom` is unused (from `deadcode`)

gracenoah:
I don't see how my change could have caused the failure. Any ideas? It's coming from
```
func getLocalClient() *client.Local {
	return client.NewLocal(node)
}
```
and seems to be an issue because `node` is global and being accessed by concurrent tests.

melekes:
I view functional options as a way to modify some minor properties of the object (like changing car's color). If we're going to replace the car's engine (the most important detail; maybe a bad analogy), then I would rather use a constructor:

```go
NewHTTPWithClient(c http.Client)
NewJSONRPCClientWithHTTPClient(c http.Client)
```

Agree/disagree?

gracenoah:
I see functional options as a way of not breaking backwards compatibility while still allowing for constructors to be customized. If you don't expect to ever need any other options and you prefer alternative constructors, I'm happy to change the PR. Just let me know.

Any tips for fixing the race?

melekes:
> If you don't expect to ever need any other options and you prefer alternative constructors, I'm happy to change the PR. Just let me know.

Please do

melekes:
> Any tips for fixing the race?

no idea. But I can take a look once you're done with the code.

marbar3778:
is there an update to this PR?

gracenoah:
Sorry for the delay. Updated with separate constructors.
The tests failed with a race:
```
==================
WARNING: DATA RACE
Read at 0x00c001005b70 by goroutine 471:
  github.com/tendermint/tendermint/blockchain/v0.(*BlockchainReactor).poolRoutine()
      /go/src/github.com/tendermint/tendermint/blockchain/v0/reactor.go:336 +0xeca

Previous write at 0x00c001005b70 by goroutine 463:
  github.com/tendermint/tendermint/blockchain/v0.TestBadBlockStopsPeer()
      /go/src/github.com/tendermint/tendermint/blockchain/v0/reactor_test.go:223 +0x978
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:865 +0x163

Goroutine 471 (running) created at:
  github.com/tendermint/tendermint/blockchain/v0.(*BlockchainReactor).OnStart()
      /go/src/github.com/tendermint/tendermint/blockchain/v0/reactor.go:118 +0x127
  github.com/tendermint/tendermint/libs/common.(*BaseService).Start()
      /go/src/github.com/tendermint/tendermint/libs/common/service.go:139 +0x504
  github.com/tendermint/tendermint/blockchain/v0.(*BlockchainReactor).Start()
      <autogenerated>:1 +0x43
  github.com/tendermint/tendermint/p2p.(*Switch).OnStart()
      /go/src/github.com/tendermint/tendermint/p2p/switch.go:223 +0x120
  github.com/tendermint/tendermint/libs/common.(*BaseService).Start()
      /go/src/github.com/tendermint/tendermint/libs/common/service.go:139 +0x504
  github.com/tendermint/tendermint/p2p.StartSwitches()
      /go/src/github.com/tendermint/tendermint/p2p/test_util.go:160 +0x75
  github.com/tendermint/tendermint/p2p.MakeConnectedSwitches()
      /go/src/github.com/tendermint/tendermint/p2p/test_util.go:81 +0x18e
  github.com/tendermint/tendermint/blockchain/v0.TestBadBlockStopsPeer()
      /go/src/github.com/tendermint/tendermint/blockchain/v0/reactor_test.go:198 +0x72b
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:865 +0x163

Goroutine 463 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:916 +0x65a
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:1157 +0xa8
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:865 +0x163
  testing.runTests()
      /usr/local/go/src/testing/testing.go:1155 +0x523
  testing.(*M).Run()
      /usr/local/go/src/testing/testing.go:1072 +0x2eb
  main.main()
      _testmain.go:102 +0x334
==================
I[2019-08-11|18:26:51.152] Committed state                              height=130 txs=10 appHash=
panic: BlockStore can only save contiguous blocks. Wanted 149, got 147

goroutine 1546 [running]:
github.com/tendermint/tendermint/store.(*BlockStore).SaveBlock(0xc000e4c8a0, 0xc00217ab40, 0xc000101220, 0xc002182700)
	/go/src/github.com/tendermint/tendermint/store/store.go:152 +0x98b
github.com/tendermint/tendermint/blockchain/v0.(*BlockchainReactor).poolRoutine(0xc001005a00)
	/go/src/github.com/tendermint/tendermint/blockchain/v0/reactor.go:336 +0xf32
created by github.com/tendermint/tendermint/blockchain/v0.(*BlockchainReactor).OnStart
	/go/src/github.com/tendermint/tendermint/blockchain/v0/reactor.go:118 +0x128
FAIL	github.com/tendermint/tendermint/blockchain/v0	18.812s
```
The race doesn't seem related to my change. Please advise on how to proceed.

gracenoah:
I think I fixed the test. There was some weird global state being modified. I don't understand what it was, but at least the test passes now.

rpc/client/httpclient.go
```diff
@@ -85,7 +86,12 @@ var _ rpcClient = (*baseRPCClient)(nil)
 // NewHTTP takes a remote endpoint in the form <protocol>://<host>:<port> and
 // the websocket path (which always seems to be "/websocket")
 func NewHTTP(remote, wsEndpoint string) *HTTP {
-	rc := rpcclient.NewJSONRPCClient(remote)
+	return NewHTTPWithHTTPClient(remote, wsEndpoint, nil)
+}
+
+// NewHTTPWithHTTPClient allows for setting a custom http client. See NewHTTP
```
melekes:
```suggestion
// NewHTTPWithClient allows for setting a custom http client. See NewHTTP
```

rpc/client/httpclient.go
```diff
@@ -85,7 +86,12 @@ var _ rpcClient = (*baseRPCClient)(nil)
...
+func NewHTTPWithHTTPClient(remote, wsEndpoint string, client *http.Client) *HTTP {
```
melekes:
```suggestion
func NewHTTPWithClient(remote, wsEndpoint string, client *http.Client) *HTTP {
```

rpc/client/rpc_test.go
```diff
@@ -37,6 +42,7 @@ func getLocalClient() *client.Local {
 func GetClients() []client.Client {
 	return []client.Client{
 		getHTTPClient(),
+		getHTTPClientCustom(),
```
melekes: we don't have to test every single RPC method just to ensure providing custom client is possible. A single test where we use NewHTTPWithClient and call some method will do!

rpc/lib/client/http_client.go
```diff
@@ -124,7 +124,21 @@ var _ JSONRPCCaller = (*JSONRPCRequestBatch)(nil)
 
 // NewJSONRPCClient returns a JSONRPCClient pointed at the given address.
 func NewJSONRPCClient(remote string) *JSONRPCClient {
-	address, client := makeHTTPClient(remote)
+	return NewJSONRPCClientWithHTTPClient(remote, nil)
+}
+
+// NewJSONRPCClient returns a JSONRPCClient pointed at the given address using a custom http client
+func NewJSONRPCClientWithHTTPClient(remote string, clientOverride *http.Client) *JSONRPCClient {
```
```suggestion
func NewJSONRPCClientWithHTTPClient(remote string, httpClient *http.Client) *JSONRPCClient {
```

rpc/client/httpclient.go
```diff
@@ -85,7 +86,12 @@ var _ rpcClient = (*baseRPCClient)(nil)
 // NewHTTP takes a remote endpoint in the form <protocol>://<host>:<port> and
 // the websocket path (which always seems to be "/websocket")
 func NewHTTP(remote, wsEndpoint string) *HTTP {
-	rc := rpcclient.NewJSONRPCClient(remote)
+	return NewHTTPWithClient(remote, wsEndpoint, nil)
+}
+
+// NewHTTPWithClient allows for setting a custom http client. See NewHTTP
+func NewHTTPWithClient(remote, wsEndpoint string, client *http.Client) *HTTP {
+	rc := rpcclient.NewJSONRPCClientWithClient(remote, client)
```
melekes:
```suggestion
	rc := rpcclient.NewJSONRPCClientWithHTTPClient(remote, client)
```

rpc/lib/client/http_client.go
```diff
@@ -124,7 +124,21 @@ var _ JSONRPCCaller = (*JSONRPCRequestBatch)(nil)
 
 // NewJSONRPCClient returns a JSONRPCClient pointed at the given address.
 func NewJSONRPCClient(remote string) *JSONRPCClient {
-	address, client := makeHTTPClient(remote)
+	return NewJSONRPCClientWithClient(remote, nil)
+}
+
+// NewJSONRPCClient returns a JSONRPCClient pointed at the given address using a custom http client
+func NewJSONRPCClientWithClient(remote string, httpClient *http.Client) *JSONRPCClient {
```
```suggestion
func NewJSONRPCClientWithHTTPClient(remote string, httpClient *http.Client) *JSONRPCClient {
```

rpc/lib/client/http_client.go
```diff
@@ -124,7 +124,21 @@ var _ JSONRPCCaller = (*JSONRPCRequestBatch)(nil)
 
 // NewJSONRPCClient returns a JSONRPCClient pointed at the given address.
 func NewJSONRPCClient(remote string) *JSONRPCClient {
-	address, client := makeHTTPClient(remote)
+	return NewJSONRPCClientWithClient(remote, nil)
+}
+
+// NewJSONRPCClient returns a JSONRPCClient pointed at the given address using a custom http client
```
```suggestion
// NewJSONRPCClientWithHTTPClient returns a JSONRPCClient pointed at the given address using a custom http client.
```

gracenoah:
🤦‍♂ Sorry I guess I did the renaming a bit too mechanically

rpc/client/rpc_test.go
```diff
@@ -25,8 +25,8 @@ import (
 )
 
 func getHTTPClient() *client.HTTP {
-	rpcAddr := rpctest.GetConfig().RPC.ListenAddress
-	return client.NewHTTP(rpcAddr, "/websocket")
+	rpcAddr := strings.Replace(rpctest.GetConfig().RPC.ListenAddress, "tcp", "http", -1)
```
melekes: we should use the default constructor here

gracenoah: Ah, sorry, I misunderstood.

melekes:
I think we should 

1. rename `makeHTTPClient` to `DefaultHTTPClient` and expose it
2. forbid passing `nil` http.Client. I see it as an antipattern when function called `XWithY` allows passing `Y=nil` (https://en.wikipedia.org/wiki/Null_Object_pattern is somewhat related)

I can do it myself or @gracenoah if he/she wants to.

Plus we still need a test here. Actually two tests

gracenoah:
@melekes What do you recommend doing in case of a nil http client? I don't want to complicate the signature of the function by adding an error return and panicing wouldn't be significantly different from doing nothing and letting it panic somewhere else.

For tests, are unit tests that check the private member variables sufficient? Do you want me to write some kind of integration test? What did you have in mind?

melekes:
> What do you recommend doing in case of a nil http client?

panic

> For tests, are unit tests that check the private member variables sufficient?

```go
c := NewHTTPWithClient(remote, http.DefaultClient())
status, err := c.Status()
require.NoError(t, err)
```

gracenoah:
Oops. Added tests.

gracenoah:
That ended up messy. I think the next step here is to pull out the protocol selection from makeHTTPDialer so that the caller of NewHTTPWithClient doesn't have to do that conversion on their own.

rpc/lib/client/http_client.go
```diff
@@ -36,10 +36,37 @@ type HTTPClient interface {
...
+		return "", "", err
+	}
+
 	// protocol to use for http operations, to support both http and https
 	clientProtocol := protoHTTP
+	// accept http as an alias for tcp and set the client protocol
+	switch protocol {
+	case protoHTTP, protoHTTPS:
+		clientProtocol = protocol
+		protocol = protoTCP
```
golangcibot: ineffectual assignment to `protocol` (from `ineffassign`)

gracenoah:
I had to do a bunch of refactoring to untangle the parsing of addresses from the creation of the dialer and preserve the use of the "address" vs "remote" throughout. I had to add a bunch of panics because these functions used to not error. They used to fail later at dial time if given invalid parameters. Obviously, I'm not happy with the result, but I'm not sure what to do next. The only alternative I see it to require the caller to know both the "remote" and the "address" when using a custom http client. Maybe we can expose the clientAddress function and require the caller of NewJSONRPCClientWithHTTPClient to use it instead of passing the remote in?

rpc/lib/client/http_client.go
```diff
@@ -36,10 +36,36 @@ type HTTPClient interface {
 }
 
 // TODO: Deprecate support for IP:PORT or /path/to/socket
-func makeHTTPDialer(remoteAddr string) (string, string, func(string, string) (net.Conn, error)) {
+func clientAddressParts(remoteAddr string) (string, string, error) {
```
```suggestion
// protocol - client's protocol (for example, "http", "https", "wss", "ws", "tcp")
// trimmedS - rest of the address (for example, "192.0.2.1:25", "[2001:db8::1]:80") with "/" replaced with "."
func toClientAddrAndParse(remoteAddr string) (network string, trimmedS string, err error) {
```

rpc/lib/client/http_client.go
```diff
@@ -49,38 +75,44 @@ func makeHTTPDialer(remoteAddr string) (string, string, func(string, string) (ne
 	case len(parts) == 2:
 		protocol, address = parts[0], parts[1]
 	default:
-		// return a invalid message
-		msg := fmt.Sprintf("Invalid addr: %s", remoteAddr)
-		return clientProtocol, msg, func(_ string, _ string) (net.Conn, error) {
-			return nil, errors.New(msg)
-		}
+		return "", "", fmt.Errorf("Invalid addr: %s", remoteAddr)
```
```suggestion
		return "", "", fmt.Errorf("invalid addr: %s", remoteAddr)
```

rpc/lib/client/http_client.go
```diff
@@ -36,10 +36,36 @@ type HTTPClient interface {
...
+	case protoWS, protoWSS:
+		clientProtocol = protocol
+	}
+
+	// replace / with . for http requests (kvstore domain)
+	trimmedAddress := strings.Replace(address, "/", ".", -1)
+	return clientProtocol, trimmedAddress, nil
+}
 
+func clientAddress(remoteAddr string) (string, error) {
```
```suggestion
func toClientAddr(remoteAddr string) (string, error) {
```

rpc/lib/client/http_client.go
```diff
@@ -49,38 +75,44 @@ func makeHTTPDialer(remoteAddr string) (string, string, func(string, string) (ne
...
 	default:
-		// return a invalid message
-		msg := fmt.Sprintf("Invalid addr: %s", remoteAddr)
-		return clientProtocol, msg, func(_ string, _ string) (net.Conn, error) {
-			return nil, errors.New(msg)
-		}
+		return "", "", fmt.Errorf("Invalid addr: %s", remoteAddr)
 	}
 
 	// accept http as an alias for tcp and set the client protocol
```
```suggestion
	// accept http/https as an alias for tcp
```

rpc/lib/client/http_client.go
```diff
@@ -36,10 +36,36 @@ type HTTPClient interface {
...
 
+func clientAddress(remoteAddr string) (string, error) {
+	clientProtocol, trimmedAddress, err := clientAddressParts(remoteAddr)
+	if err != nil {
+		return "", err
+	}
+	return clientProtocol + "://" + trimmedAddress, nil
+}
+
+func remoteParts(remoteAddr string) (string, string, error) {
```
```suggestion
// network - name of the network (for example, "tcp", "unix")
// s - rest of the address (for example, "192.0.2.1:25", "[2001:db8::1]:80")
func parseRemoteAddr(addr string) (network string, s string, err error) {
```

rpc/lib/client/http_client.go
```diff
@@ -36,10 +36,36 @@ type HTTPClient interface {
...
+func clientAddressParts(remoteAddr string) (string, string, error) {
+	protocol, address, err := remoteParts(remoteAddr)
+	if err != nil {
+		return "", "", err
+	}
+
 	// protocol to use for http operations, to support both http and https
 	clientProtocol := protoHTTP
+	// accept http as an alias for tcp and set the client protocol
+	switch protocol {
```
melekes: don't think we need this switch now

rpc/client/httpclient.go
```diff
@@ -85,7 +86,16 @@ var _ rpcClient = (*baseRPCClient)(nil)
 // NewHTTP takes a remote endpoint in the form <protocol>://<host>:<port> and
 // the websocket path (which always seems to be "/websocket")
 func NewHTTP(remote, wsEndpoint string) *HTTP {
-	rc := rpcclient.NewJSONRPCClient(remote)
+	httpClient := rpcclient.DefaultHTTPClient(remote)
+	return NewHTTPWithClient(remote, wsEndpoint, httpClient)
+}
+
+// NewHTTPWithClient allows for setting a custom http client. See NewHTTP
+func NewHTTPWithClient(remote, wsEndpoint string, client *http.Client) *HTTP {
```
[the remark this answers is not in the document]
gracenoah: http.Client is meant to be passed around as a pointer and all methods are on the pointer version of the struct: https://golang.org/pkg/net/http/#Client
It would be weird to require it to be passed as a copy and might have unintended consequences.

gracenoah:
I can make those fixes, but what do you think about the new potential error case that I had to add a panic for? Should I:
1. leave it as-is to preserve API compatibility
2. change the constructor signature to return an error
3. change the constructor to not call a function that can fail (take in the address and remote as separate parameters)

rpc/lib/client/http_client.go
gracenoah: Oops, yes, it can be simplified now.

rpc/client/httpclient.go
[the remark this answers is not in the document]
melekes: > It would be weird to require it to be passed as a copy and might have unintended consequences.

what consequences are u talking about?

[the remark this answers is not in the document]
viktorwork: That would make copies of all properties in the struct. The struct is at: https://golang.org/src/net/http/client.go
It turns out that they all seem safe to copy because they are all pointers (interfaces). It's pretty unusual for someone to be modifying the http client after it's passed in.

Structs passed around as pointer versions of themselves is a pretty standard pattern, especially with stateful structs like long lived clients. That's why it's non-idomatic to dereference them, copy them and then pass them as arguments to another function. Every function that uses an http.Client would normally require the pointer version because all methods are on the pointer version.

For the caller to pass in the non-pointer, the caller would pretty much always have to just check for nil, dereference it and pass it in, which is not that different from checking for nil after it's passed in, but requires way more duplication of the check in all the callers.

If you still want to do this, I'll do it, but it feels wrong.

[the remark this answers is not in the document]
gracenoah: That would make copies of all properties in the struct. The struct is at: https://golang.org/src/net/http/client.go
It turns out that they all seem safe to copy because they are all pointers (interfaces). It's pretty unusual for someone to be modifying the http client after it's passed in.

Structs passed around as pointer versions of themselves is a pretty standard pattern, especially with stateful structs like long lived clients. That's why it's non-idomatic to dereference them, copy them and then pass them as arguments to another function. Every function that uses an http.Client would normally require the pointer version because all methods are on the pointer version.

For the caller to pass in the non-pointer, the caller would pretty much always have to just check for nil, dereference it and pass it in, which is not that different from checking for nil after it's passed in, but requires way more duplication of the check in all the callers.

If you still want to do this, I'll do it, but it feels wrong.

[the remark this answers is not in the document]
melekes: Right, DefaultClient returns a pointer. Okay, let's not do that. Thanks!

rpc/lib/client/http_client.go
```diff
@@ -260,10 +309,13 @@ type URIClient struct {
 }
 
 func NewURIClient(remote string) *URIClient {
```
```suggestion
// The function panics if the provided remote is invalid.
func NewURIClient(remote string) *URIClient {
```

rpc/lib/client/http_client.go
```diff
@@ -124,9 +160,22 @@ var _ JSONRPCCaller = (*JSONRPCRequestBatch)(nil)
 
 // NewJSONRPCClient returns a JSONRPCClient pointed at the given address.
 func NewJSONRPCClient(remote string) *JSONRPCClient {
-	address, client := makeHTTPClient(remote)
+	return NewJSONRPCClientWithHTTPClient(remote, DefaultHTTPClient(remote))
+}
+
+// NewJSONRPCClientWithHTTPClient returns a JSONRPCClient pointed at the given address using a custom http client
```
```suggestion
// NewJSONRPCClientWithHTTPClient returns a JSONRPCClient pointed at the given address using a custom http client.
// The function panics if the provided client is nil or remote is invalid.
```

rpc/client/httpclient.go
```diff
@@ -85,7 +86,16 @@ var _ rpcClient = (*baseRPCClient)(nil)
 // NewHTTP takes a remote endpoint in the form <protocol>://<host>:<port> and
 // the websocket path (which always seems to be "/websocket")
 func NewHTTP(remote, wsEndpoint string) *HTTP {
-	rc := rpcclient.NewJSONRPCClient(remote)
+	httpClient := rpcclient.DefaultHTTPClient(remote)
+	return NewHTTPWithClient(remote, wsEndpoint, httpClient)
+}
+
+// NewHTTPWithClient allows for setting a custom http client. See NewHTTP
```
melekes:
```suggestion
// NewHTTPWithClient allows for setting a custom http client. See NewHTTP
// The function panics if the provided client is nil or remote is invalid.
```

rpc/client/httpclient.go
```diff
@@ -85,7 +86,16 @@ var _ rpcClient = (*baseRPCClient)(nil)
 // NewHTTP takes a remote endpoint in the form <protocol>://<host>:<port> and
 // the websocket path (which always seems to be "/websocket")
```
melekes:
```suggestion
// the websocket path (which always seems to be "/websocket").
// The function panics if the provided remote is invalid.
```

melekes:
In the next major version we should probably modify functions to return errors if remote is invalid.

melekes:
Can you also add an entry to `CHANGELOG_PENDING.md`?

CHANGELOG_PENDING.md
```diff
@@ -19,4 +19,6 @@ program](https://hackerone.com/tendermint).
 
 ### IMPROVEMENTS:
 
+- [rpc] [\#3779](https://github.com/tendermint/tendermint/pull/3779) Allow using a custom http client in rpc client
```
```suggestion
- [rpc] \#2010 Add NewHTTPWithClient and NewJSONRPCClientWithHTTPClient (note these and NewHTTP, NewJSONRPCClient functions panic if remote is invalid) (@gracenoah)
```

PR #3780, opened 2019-07-09
Remove heap.go

### LINKED ISSUE #2432: Use stdlibs heap instead of the heap in common
Currently the only place we use our own heap implementation is for the validator set. We could easily refactor to utilize to use the stdlibs heap implementation, which would likely end up being more efficient. (As that does the O(N) heapify unlike our code which does an `N log(N)` process. If the stdlib's heap interface is deemed as a pain to implement, then we should adapt cmn.Heap to utilize the stdlib's heap implementation imo.

marbar3778:
-  cmn.Heap is not being used in cosmos-sdk, iavl nor tendermint repo.

closes #2432 

* [ ] Wrote tests


CHANGELOG_PENDING.md
```diff
@@ -9,33 +9,37 @@ program](https://hackerone.com/tendermint).
...
+- Go API
 
-* Go API
   - [abci] \#2127 ABCI / mempool: Add a "Recheck Tx" indicator. Breaks the ABCI
     client interface (`abcicli.Client`) to allow for supplying the ABCI
     `types.RequestCheckTx` and `types.RequestDeliverTx` structs, and lets the
     mempool indicate to the ABCI app whether a CheckTx request is a recheck or
     not.
   - [libs] Remove unused `db/debugDB` and `common/colors.go` & `errors/errors.go` files (@marbar3778)
+  - [libs] Remove unused `Heap` file (@marbar3778)
```
```suggestion
  - [libs] \#2432 Remove unused `common/heap.go` file (@marbar3778)
```

PR #3781, opened 2019-07-09
Add consensus_params to testnet config generation and document time_iota_ms (fixes #3723)

### LINKED ISSUE #3723: ABCI: RequestBeginBlock.Header.Time is sometimes far in the future
* Tendermint version: 0.31.5 (d2eab536)
* ABCI app: self-written, not currently publicly available
* OS: Ubuntu 16.04.5 LTS x86_64

**What happened**: The block header time starts drifting into the future. The below graph shows `now - BlockHeaderTime`, so if the value is negative, `Block.Header.Time` is ahead of real time.

timediff

**What was expected to happen**: The block header time should match real time, to within a couple of seconds.

**Tried the latest version**: no

**How to reproduce it**: I have yet to create a minimal ABCI app that reproduces this problem.

**Logs**: At the default log levels (`main:info,state:info,*:error`) there are no useful errors. I have changed the config to `*:debug` and am waiting for the next occurrence of time drift.

**Config**: Default config, as generated by `tendermint testnet`

**node command runtime flags**: Just `tendermint node`

Our tendermint cluster consists of 3 nodes, in GB, SG and US. Transactions are submitted at a rate of under 10 per second. If under heavy load, it would be reasonable to expect the block header time to lag real time (ie drift into the past), but this issue is drift in the opposite direction.

ashleyvega:
* [x] Referenced an issue explaining the need for the change - #3723
* [x] Updated all relevant documentation in docs - **Yes**
* [x] Updated all code comments where relevant - **n/a**
* [x] Wrote tests - **n/a**
* [x] Updated CHANGELOG_PENDING.md - **Yes**


CHANGELOG_PENDING.md
```diff
@@ -43,3 +43,4 @@ program](https://hackerone.com/tendermint).
 - [p2p] \#3338 Prevent "sent next PEX request too soon" errors by not calling
   ensurePeers outside of ensurePeersRoutine
 - [behaviour] Return correct reason in MessageOutOfOrder (@jim380)
+- [config] \#3723 Add consensus_params to testnet config generation; document time_iota_ms (@ashleyvega)
```
```suggestion
- [testnet] \#3723 Add consensus_params to testnet config generation; document time_iota_ms (@ashleyvega)
```

PR #3783, opened 2019-07-10
Removal total tx num tx

### LINKED ISSUE #2521: abci: Remove TotalTxs from BlockHeader
The TotalTxs is a nice stat, but doesn't seem like it needs to be sent in every block. This isn't needed for lite client verification, so I think we should remove it from the header.

I think we should also remove `NumTxs`, again a nice stat but isn't needed for lite clients. All full nodes can derive it from the block data.

marbar3778:
- Removal totalTx and numTx

closes #2521

* [ ] Wrote tests


consensus/state.go
```diff
@@ -1280,8 +1280,7 @@ func (cs *ConsensusState) finalizeCommit(height int64) {
 		panic(fmt.Sprintf("+2/3 committed an invalid block: %v", err))
 	}
 
-	cs.Logger.Info(fmt.Sprintf("Finalizing commit of block with %d txs", block.NumTxs),
-		"height", block.Height, "hash", block.Hash(), "root", block.AppHash)
+	cs.Logger.Info("height", block.Height, "hash", block.Hash(), "root", block.AppHash)
```
melekes: the format of Info is `message, "arg1", arg1, ...`

```suggestion
	cs.Logger.Info("Finalizing commit of block", "height", block.Height, "hash", block.Hash(), "root", block.AppHash)

File: consensus/state.go [hunk 2]
```diff
@@ -1386,9 +1385,7 @@ func (cs *ConsensusState) recordMetrics(height int64, block *types.Block) {
 		)
 	}
 
-	cs.metrics.NumTxs.Set(float64(block.NumTxs))
```
melekes: Why remove the metric?

consensus/state.go
```diff
@@ -1386,9 +1385,7 @@ func (cs *ConsensusState) recordMetrics(height int64, block *types.Block) {
...
 	cs.metrics.BlockSizeBytes.Set(float64(block.Size()))
-	cs.metrics.TotalTxs.Set(float64(block.TotalTxs))
```
melekes: Why remove the metric?

docs/spec/abci/abci.md
```diff
@@ -396,8 +396,6 @@ Commit are included in the header of the next block.
     votes in the block.LastCommit.
     For height == 1, it's genesis time.
   - `NumTxs (int32)`: Number of transactions in the block
```
melekes: remove

docs/spec/abci/abci.md
```diff
@@ -500,17 +498,15 @@ Commit are included in the header of the next block.
...
 
 ### EvidenceParams
 
 - **Fields**:
   - `MaxAge (int64)`: Max age of evidence, in blocks. Evidence older than this
-    is considered stale and ignored.
-        - This should correspond with an app's "unbonding period" or other
-          similar mechanism for handling Nothing-At-Stake attacks.
-        - NOTE: this should change to time (instead of blocks)!
+    is considered stale and ignored. - This should correspond with an app's "unbonding period" or other
```
melekes: remove `-`

state/txindex/indexer_service.go
```diff
@@ -48,22 +48,22 @@ func (is *IndexerService) OnStart() error {
 	go func() {
 		for {
 			msg := <-blockHeadersSub.Out()
```
melekes: this needs to be changed

state/txindex/indexer_service_test.go
```diff
@@ -34,8 +34,10 @@ func TestIndexerServiceIndexesBlocks(t *testing.T) {
 	defer service.Stop()
 
 	// publish block with txs
-	eventBus.PublishEventNewBlockHeader(types.EventDataNewBlockHeader{
-		Header: types.Header{Height: 1, NumTxs: 2},
+	eventBus.PublishEventNewBlock(types.EventDataNewBlock{
+		Block: &types.Block{
+			Header: types.Header{Height: 1},
```
melekes: where are two txs?

tools/tm-bench/statistics.go
```diff
@@ -66,9 +66,10 @@ func calculateStatistics(
 		// increase number of blocks for that second
 		numBlocksPerSec[sec]++
 
+		// TODO: add tracking for tx/sec
 		// increase number of txs for that second
-		numTxsPerSec[sec] += blockMeta.Header.NumTxs
-		logger.Debug(fmt.Sprintf("%d txs at block height %d", blockMeta.Header.NumTxs, blockMeta.Header.Height))
+		numTxsPerSec[sec] += 1
```
melekes: ???

marbar3778: below on the next line theres a TODO: so still need to get to this part

consensus/state.go
marbar3778: `cs.metrics.TotalTxs.Set(float64(len(block.Data.Txs)))` added this section in, NumTxs and totalTxs aren't present in the block anymore

consensus/state.go
marbar3778: same as https://github.com/tendermint/tendermint/pull/3783/files#r302011315

state/txindex/indexer_service_test.go
marbar3778: added two txs into data:txs array

types/block.go
```diff
@@ -18,7 +18,7 @@ import (
 
 const (
 	// MaxHeaderBytes is a maximum header size (including amino overhead).
-	MaxHeaderBytes int64 = 653
+	MaxHeaderBytes int64 = 632
```
ebuchman: How do we verify this again?

state/txindex/indexer_service.go
```diff
@@ -35,7 +35,7 @@ func (is *IndexerService) OnStart() error {
 	// cancelled due to not pulling messages fast enough. Cause this might
 	// sometimes happen when there are no other subscribers.
 
-	blockHeadersSub, err := is.eventBus.SubscribeUnbuffered(context.Background(), subscriber, types.EventQueryNewBlockHeader)
+	blocksSub, err := is.eventBus.SubscribeUnbuffered(context.Background(), subscriber, types.EventQueryNewBlock)
```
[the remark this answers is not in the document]
melekes: good idea!

types/block.go
melekes: I think we have a test which calculates it and fails if the const is wrong

marbar3778: https://github.com/tendermint/tendermint/blob/master/types/block_test.go#L239

state/tx_filter_test.go
```diff
@@ -24,10 +24,10 @@ func TestTxFilter(t *testing.T) {
 		isErr bool
 	}{
 		{types.Tx(cmn.RandBytes(250)), false},
-		{types.Tx(cmn.RandBytes(1809)), false},
-		{types.Tx(cmn.RandBytes(1810)), false},
-		{types.Tx(cmn.RandBytes(1811)), true},
-		{types.Tx(cmn.RandBytes(1812)), true},
+		{types.Tx(cmn.RandBytes(1811)), false},
```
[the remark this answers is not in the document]
marbar3778: What is the question here?

state/txindex/indexer_service.go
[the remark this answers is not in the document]
marbar3778: would it be of interest to also add numTxs as a field in the `BlockMeta struct`?

abci/types/types.proto
```diff
@@ -256,26 +256,24 @@ message Header {
   string chain_id = 2 [(gogoproto.customname)="ChainID"];
   int64 height = 3;
   google.protobuf.Timestamp time = 4 [(gogoproto.nullable)=false, (gogoproto.stdtime)=true];
-  int64 num_txs = 5;
-  int64 total_txs = 6;
 
   // prev block info
-  BlockID last_block_id = 7 [(gogoproto.nullable)=false];
+  BlockID last_block_id = 5 [(gogoproto.nullable)=false];
```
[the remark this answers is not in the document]
marbar3778: Will bring them back for compatibility purposes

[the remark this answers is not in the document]
melekes: This will be breaking anyway, so why not do this?!

tools/tm-monitor/monitor/network_test.go
```diff
@@ -13,7 +13,7 @@ import (
 func TestNetworkNewBlock(t *testing.T) {
 	n := monitor.NewNetwork()
 
-	n.NewBlock(tmtypes.Header{Height: 5, NumTxs: 100})
+	n.NewBlock(tmtypes.Header{Height: 5})
```
golangcibot: cannot use (tmtypes.Header literal) (value of type types.Header) as *types.Block value in argument to n.NewBlock (from `typecheck`)

tools/tm-monitor/monitor/network_test.go
```diff
@@ -13,7 +13,9 @@ import (
 func TestNetworkNewBlock(t *testing.T) {
 	n := monitor.NewNetwork()
 
-	n.NewBlock(tmtypes.Header{Height: 5, NumTxs: 100})
+	n.NewBlock(tmtypes.Block{
```
golangcibot: cannot use (tmtypes.Block literal) (value of type types.Block) as *types.Block value in argument to n.NewBlock (from `typecheck`)

state/tx_filter_test.go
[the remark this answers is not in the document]
ValarDragon: Its unclear to me why these test lines were changed

state/execution.go
```diff
@@ -215,7 +215,6 @@ func (blockExec *BlockExecutor) Commit(
 	blockExec.logger.Info(
 		"Committed state",
 		"height", block.Height,
-		"txs", block.NumTxs,
```
melekes:
```
"txs", len(block.Txs),
```

state/txindex/indexer_service.go
```diff
@@ -35,7 +35,7 @@ func (is *IndexerService) OnStart() error {
 	// cancelled due to not pulling messages fast enough. Cause this might
 	// sometimes happen when there are no other subscribers.
 
-	blockHeadersSub, err := is.eventBus.SubscribeUnbuffered(context.Background(), subscriber, types.EventQueryNewBlockHeader)
+	blocksHeaderSub, err := is.eventBus.SubscribeUnbuffered(context.Background(), subscriber, types.EventQueryNewBlockHeader)
```
melekes: why change the name?

state/txindex/indexer_service.go
```diff
@@ -47,23 +47,23 @@ func (is *IndexerService) OnStart() error {
 
 	go func() {
 		for {
-			msg := <-blockHeadersSub.Out()
-			header := msg.Data().(types.EventDataNewBlockHeader).Header
-			batch := NewBatch(header.NumTxs)
-			for i := int64(0); i < header.NumTxs; i++ {
+			msg := <-blocksHeaderSub.Out()
+			eventDataHeader := msg.Data().(types.EventDataNewBlockHeader)
```
melekes:
```
height := eventDataHeader.Header.Height
```

types/block_meta.go
```diff
@@ -2,15 +2,17 @@ package types
 
 // BlockMeta contains meta information about a block - namely, it's ID and Header.
 type BlockMeta struct {
-	BlockID BlockID `json:"block_id"` // the block hash and partsethash
-	Header  Header  `json:"header"`   // The block's Header
+	BlockID BlockID `json:"block_id"`   // the block hash and partsethash
+	Header  Header  `json:"header"`     // The block's Header
+	NumTxs  int64   `json:"number_txs"` //
```
melekes: missing comment

melekes:
also, please remove `consensus_total_txs` from the metrics list https://github.com/tendermint/tendermint/blob/master/docs/tendermint-core/metrics.md#L32

abci/types/types.proto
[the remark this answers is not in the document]
ebuchman: Because we're talking about breaking different things. The original change is breaking on the blockchain, but not necessarily on any APIs. So all you have to do is start a new blockchain to take advantage of the changes. But with a change like this, it's breaking on the ABCI, which means everyone has to update their ABCI stuff.

I don't think we should take a "breaking is breaking" attitude because we have like 6 different interfaces that can break and breaking one does not mean we should break another because each one has different implications for users. So the analysis should be based on the implications of a particular break and always be case specific, not just "breaking is breaking".

This is especially true here since we plan on breaking the header again to remove more stuff. I'd rather this change be non-breaking on the ABCI if possible, and a later change when we do more cleanup be breaking. At least then it might be worth it (a significantly simpler Header structure) vs just removing two fields now ...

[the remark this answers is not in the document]
ebuchman: Our philosophy on breaking needs to be documented and committed to the repo somewhere ...

state/tx_filter_test.go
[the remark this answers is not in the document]
ebuchman: Yes, why did these change?

types/block_meta.go
```diff
@@ -1,16 +1,18 @@
 package types
 
-// BlockMeta contains meta information about a block - namely, it's ID and Header.
+// BlockMeta contains meta information.
 type BlockMeta struct {
-	BlockID BlockID `json:"block_id"` // the block hash and partsethash
-	Header  Header  `json:"header"`   // The block's Header
+	BlockID BlockID `json:"block_id"`
+	Header  Header  `json:"header"`
+	NumTxs  int64   `json:"number_txs"`
```
ebuchman: For consistency, this should probably be `num_txs`. Eg. grep for `num_` vs. `number_` ...

types/events.go
```diff
@@ -73,6 +73,7 @@ type EventDataNewBlock struct {
 type EventDataNewBlockHeader struct {
 	Header Header `json:"header"`
 
+	NumTxs           int64                   `json:"number_txs"` // Number of txs in a block
```
ebuchman: Same

abci/types/types.proto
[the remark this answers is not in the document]
ebuchman: Ok I'm more sympathetic to this because otherwise we'd have to annotate the `types#Header` to match the non-breaking removal of the fields and that would be painful and I'm not even sure amino supports it. In any case, the thoughts about "breaking is breaking" still stand :)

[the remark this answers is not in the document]
melekes: @tessr Do you want to try to write such a document? It may help you get deeper understanding of various things that are breaking ("6 different interfaces")

PR #3785, opened 2019-07-10
Clean up `fail.go`

### LINKED ISSUE #2729: Cleanup `libs/fail`
We brought `ebuchman/fail-test` into the repo to have one less dependency, but it could probably use some cleanup. See https://github.com/tendermint/tendermint/pull/2728

marbar3778:
- Clean up fail.go
- Remove unneeded Func `FailRand()`

closes #2729

* [ ] Wrote tests

### DIFF
--- a/libs/fail/fail.go
+++ b/libs/fail/fail.go
@@ -2,36 +2,30 @@ package fail
 
 import (
 	"fmt"
-	"math/rand"
 	"os"
 	"strconv"
 )
 
-var callIndexToFail int
-
-func init() {
+func envSet() int {
 	callIndexToFailS := os.Getenv("FAIL_TEST_INDEX")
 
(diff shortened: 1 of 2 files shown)


libs/fail/fail.go
```diff
@@ -43,37 +37,7 @@ func Fail() {
...
-
-	if callRandIndex == n {
-		callIndex += 1
-	}
-}
-
 func Exit() {
 	fmt.Printf("*** fail-test %d ***\n", callIndex)
 	os.Exit(1)
-	//	proc, _ := os.FindProcess(os.Getpid())
```
melekes: not sure but this could be useful

marbar3778: leave it commented in the file?

melekes: yes please

PR #3786, opened 2019-07-11
Updating the default private validator key JSON filename

andynog:
The filename has changed in v0.28 when priva_validator.json was split into two files and I believe the documentation has not been updated to reflect that. This change only renames the file priv_validator.json to priv_validator_key.json, but in the future I believe it might be important to also add information about priv_validator_state.json file in the documentation.

* [ ] Wrote tests


melekes:
Note: we'll most probably revert the change because of https://github.com/tendermint/tendermint/issues/3418.

PR #3787, opened 2019-07-11
Release branch for v0.32.1

marbar3778:

* [ ] Wrote tests


version/version.go
```diff
@@ -20,7 +20,7 @@ const (
 	// Must be a string because scripts like dist.sh read this file.
 	// XXX: Don't change the name of this variable or you will break
 	// automation :)
-	TMCoreSemVer = "0.32.0"
+	TMCoreSemVer = "0.32.1"
 
 	// ABCISemVer is the semantic version of the ABCI library
 	ABCISemVer  = "0.16.0"
```
marbar3778: @thanethomson @ebuchman should we bump the version on the ABCI

CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
...
+## v0.32.1
+
+*July 12, 2019*
+
+Special thanks to external contributors on this release: 
+@ParthDesai, @climber73, @jim380, @ashleyvega
+
+This release contains breaking changes to our libs folder and ABCI namely:
+- Add a "Recheck Tx" indicator 
+- Removal of `db/debugDB`, `common/colors.go`, `errors/errors.go`, `common/heap.go`, `date.go`, `io.go` and
```
melekes: I would remove this and expand on RecheckTx indicator instead (how applications can use it with examples)

CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
...
+  `GoPath()`, `Prompt()` and `IsDirEmpty()` functions from `os.go`
+
+Friendly reminder, we have a [bug bounty
+program](https://hackerone.com/tendermint).
+
+### BREAKING CHANGES:
+
+- CLI/RPC/Config
```
melekes: no need for empty sections here

CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
...
 
+## v0.32.1
+
+*July 12, 2019*
+
+Special thanks to external contributors on this release: 
+@ParthDesai, @climber73, @jim380, @ashleyvega
+
+This release contains breaking changes to our libs folder and ABCI namely:
+- Add a "Recheck Tx" indicator
```
ebuchman:
```suggestion
- CheckTx requests include a `CheckTxType` enum that can be set to `Recheck` to indicate to the application that this transaction was already checked/validated and certain expensive operations (like checking signatures) can be skipped
```

CHANGELOG.md
ebuchman: `Removed various 
```suggestion
- Removed various functions from `libs` pkgs
```

CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
...
+
+### BREAKING CHANGES:
+
+- CLI/RPC/Config
+
+- Apps
+
+- Go API
+
+  - [abci] [\#2127](https://github.com/tendermint/tendermint/issues/2127) ABCI / mempool: Add a "Recheck Tx" indicator. Breaks the ABCI
```
ebuchman: The breaking change here should be described more clearly as introducing the RequestCheckTx and RequestDeliverTx types as the args for those methods in the `abci.Application` interface. Ie. this is specifically about the breaking change to the Go API.

Adding the CheckTxType field is something that should be logged as an `[abci]` feature. It also corresponds to a bump in the ABCI version (see comment below).

CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
...
+  - [libs] [\#2432](https://github.com/tendermint/tendermint/issues/2432) Remove unused `common/heap.go` file (@marbar3778)
+  - [libs] Remove unused `date.go`, `io.go`. Remove `GoPath()`, `Prompt()` and `IsDirEmpty()` functions from `os.go` (@marbar3778)
+
+- Blockchain Protocol
+
+- P2P Protocol
+
+### FEATURES:
+
+- [node] Refactor `NewNode` to use functional options to make it more flexible
```
ebuchman:
```suggestion
- [node] Add variadic argument to `NewNode` to support functional options, allowing the Node to be more easily customized. 
```

CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
...
+  and extensible in the future.
```
ebuchman:
```suggestion
```

CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
...
+
+### IMPROVEMENTS:
+
+- [rpc] [\#3700](https://github.com/tendermint/tendermint/issues/3700) Make possible to set absolute paths for TLS cert and key (@climber73)
+
+### BUG FIXES:
+
+- [p2p] [\#3338](https://github.com/tendermint/tendermint/issues/3338) Prevent "sent next PEX request too soon" errors by not calling
+  ensurePeers outside of ensurePeersRoutine
+- [behaviour] Return correct reason in MessageOutOfOrder (@jim380)
```
ebuchman: Missing an issue/PR. 

Also why isn't the `behaviour` pkg under `p2p` cc @brapse

version/version.go
ebuchman: Very good question. I don't think we've ever really specified what this version refers to. Probably it should be the protobuf files themselves, and any changes to the Go wrapper should be covered by the TMCoreSemVer. In that case, since we added the CheckTxType, this should probably change to `0.16.1`. My understanding is any implementations of v0.16.0 will just ignore the new field in RequestCheckTx.

ebuchman:
We should also fix the abci documentation about the recheck stuff and link to it from the changelog. See https://github.com/tendermint/tendermint/pull/3744#pullrequestreview-261052009

version/version.go
marbar3778: is TMCoreSemVar suppose to be changed or the automation is supposed to do it?

Issue #1937, opened 2018-07-10
tools: tm-monitor: docker containers change port on restart


zramsay:
original issue: https://github.com/tendermint/tools/issues/1

marbar3778:
tm-monitor has been removed

PR #3788, opened 2019-07-12
Add btcec to get_tools.sh

sabau:
Coming from a fresh environment, this package seems to be missing from the `pkg` folder to run `make` into `tools/tm-monitor`

I was looking into: https://github.com/tendermint/tendermint/issues/1937

* [ ] Wrote tests


ebuchman:
This seems weird - is there a binary in there we need? If not seems like this should be taking care of by `go mod` somehow

sabau:
You're right, in the container I was playing `GO111MODULE` was off

PR #3798, opened 2019-07-15
Remove go func from finishAsyncCall()

### LINKED ISSUE #357: Race condition due to abci-grpc finishAsyncCall
this probably shouldn't be in a go routine: https://github.com/tendermint/tmsp/blob/master/client/grpc_client.go#L228

 run the `bash test/app/test.sh` with a tendermint built using `-race` to replicate

marbar3778:
closes #357

- Remove go func(){}() that caused race condiditon

- To reproduce
	- add -race in make file to `install_abci`
	- Remove `CGO_ENABLED=0` & add -race to `install`

* [ ] Wrote tests

### DIFF
--- a/abci/client/grpc_client.go
+++ b/abci/client/grpc_client.go
@@ -228,18 +228,15 @@ func (cli *grpcClient) finishAsyncCall(req *types.Request, res *types.Response) 
 	reqres.Done()         // Release waiters
 	reqres.SetDone()      // so reqRes.SetCallback will run the callback
 
-	// go routine for callbacks
-	go func() {
-		// Notify reqRes listener if set
-		if cb := reqres.GetCallback(); cb != nil {
-			cb(res)
-		}
+	// Notify reqRes listener if set
+	if cb := reqres.GetCallback(); cb != nil {
+		cb(res)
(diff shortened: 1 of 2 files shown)


abci/client/grpc_client.go
```diff
@@ -228,18 +228,15 @@ func (cli *grpcClient) finishAsyncCall(req *types.Request, res *types.Response)
 	reqres.Done()         // Release waiters
 	reqres.SetDone()      // so reqRes.SetCallback will run the callback
 
-	// go routine for callbacks
-	go func() {
```
melekes: I am not 100% sure, but right thing to do seems to be adding a mutex, not blocking the async call.

melekes:
```
// GRPC calls are synchronous, but some callbacks expect to be called asynchronously
// (eg. the mempool expects to be able to lock to remove bad txs from cache).
// To accommodate, we finish each call in its own go-routine,
// which is expensive, but easy - if you want something better, use the socket protocol!
// maybe one day, if people really want it, we use grpc streams,
// but hopefully not :D
```

PR #3799, opened 2019-07-15
add tree-chunking.md

ebuchman:
Wasn't sure where to put this but opening for review/discussion.

I think I convinced myself that if we're going to use some form of chunking, we'll want to sync the chunks in order.

### DIFF
--- a/docs/architecture/tree-chunking.md
+++ b/docs/architecture/tree-chunking.md
@@ -1,0 +1,123 @@
+# Tree Chunking
+
+In order to securely sync a Merkle tree, we must have some algorithm for
+splitting it into chunks that remain verifiably part of the Merkle tree.
+Whether the chunks are computed eagerly ahead of time,
+or lazily in real time, is an orthogonal consideration.
+
+We refer to the node computing the chunks from a complete Merkle tree as the
+"chunker", and the node reassembling the Merkle tree from received chunks as the
+"chunkee".
+
+The main issue with various chunking strategies is how to verify that a given
+chunk corresponds to a given chunk index. Especially since we desire for
+chunkees to be able to apply chunks in any order, we need to ensure that
+applying an out-of-order chunk is fully verifiable - ie. the chunk is a valid
+part of the Merkle tree, and correctly corresponds to its chunk index.
+Turns out this may not be possible in general.
+
+## Arbitrary Chunking
+
+There are two general approaches here - breadth-first and depth-first.
+In the former case, we descend from the root one layer at a time, accumulating
+full internal nodes into chunks. In the later case, we traverse the tree from
+say left to right, accumulting full key-value pairs into chunks. In each case,
+we require additional proof data, which consists of internal nodes that prove
+all compnents of a chunk (whether nodes or key-value pairs) are valid members of
+the tree.
+
+Note that for AVL trees, which are insertion-order dependent, that the
+depth-first approach must include all internal nodes in order to properly
+communicate the structure of the tree, which cannot be derrived from key-value
+pairs alone. Such nodes would be part of the proof structure.
+
+The chunker can assign to each chunk a unique index, and every chunker will
+compute the same index for each chunk, assuming they follow the same
+strategy. However, it is not clear that, in general, a chunkee can verify the
+chunk index, as there is no explicit map from chunk index to chunk or vice
+versa.
+
+This raises two concerns for a state sync protocol:
+- how to handle mis-indexed chunks (ie. requesting chunk 3 and receiving chunk
+  4)
+- how to handle non-indexed chunks (ie. receiving a chunk that doesn't
+  correspond to any index from the given chunker strategy
+
+In each case, it appears that liveness is made much more difficult by the need
+to figure out what went wrong with the applied chunks. However, these problems
+can be eliminated by requiring chunks to be applied in-order.
+
+### MisIndexed Chunks
+
+Suppose an honest chunker follows a strategy creating chunks 1,2,3,4,5.
+Suppose an honest chunkee requests chunk 3 from a malicious chunker,
+who responds with chunk 4. The chunkee applies chunk 4 to its Merkle tree
+and verifies that it's correct. At this point the chunkee believes it
+successfully applied chunk 3. This raises two questions:
+
+- When the chunkee actually requests and receives chunk 4, and then discovers it already
+  applied that chunk, how will it know which peer was malicious, the original
+  one or the new one?
+- How will the chunkee discover that it is missing the real chunk 3?
+
+### NonIndexed Chunks
+
+Suppose an honest chunker follows a strategy creating chunks 1,2,3,4,5.
+Suppose a malicious chunker follows an alternative strategy, creating chunks
+Q,R,S,T,U, none of which correspond to chunks 1,2,3,4,5, but all of which are
+valid chunks in the tree.
+Suppose an honest chunkee requests chunk 3 from the malicious chunker,
+who responds with chunk R. The chunkee applies chunk R to its Merkle tree and
+verifies that it's correct. At this point the chunkee believes it
+successfully applied chunk 3. This raises two questions:
+
+- When the chunkee receives another chunk that overlaps with chunk R, how will
+  it know which peer was malicious, the original one or the new one?
+- How will the chunkee determine which real chunks it is missing?
+
+## Provably Indexed Chunking
+
+An alternative approach would be a strategy that chunks the tree in a manner
+such that all chunks have a verifiable index. Certain tree structures may be
+able to incorporate this requirement into their design. In the extreme case,
+the map from chunks to indices (or vice versa) could be committed into the tree
+itself. But generally, we would like to avoid such requirements on the tree and
+state and instead pursue a general strategy for chunking trees that preserves
+verifiable chunk indices.
+
+Consider an example strategy for illustration of how this would work:
+
+- the first chunk (chunk 0) contains all nodes in the top 10 layers of the tree
+- the nodes in the layer 10, which are part of chunk 0, form the roots for 1024
+  sub-trees that contain the rest of the tree
+- each of these sub-trees is its own chunk, numbered from 1 to 1024
+- so long as chunk 0 is received first, each additional chunk can be verified
+  with its index. For instance, if we receive chunk 5, the root hash of the
+  sub-tree contained therein should correspond to the 5th node in layer 10
+
+While this example provides the intuition for how the design might work,
+we now need to generalize it to handle arbitrary tree sizes and bounded
+chunk sizes. With the example above, we were only required to receive chunk 0 up
+front, and could then receive the other chunks in any order. However to
+generalize, we may be required to receive multiple chunks in order first, before
+we get to a point where we can receive them in any order.
+
+It may be the case that we need to receive all chunks in order until we get to a
+depth where the remaining chunks are complete sub-trees and can be received in
+any order. It is not clear how the chunkee can determine for itself when it
+reaches this state; while it can verify it has all nodes up to a certain depth,
+it does not necessarily know that each remaining sub-tree can fit in a chunk.
+
+Thus it may not be possible to achieve this sort of out-of-order chunking.
+
+## Conclusion
+
+The above considerations demonstrate the challenges with applying chunks of a
+Merkle tree out-of-order. While it may be possible with more complex protocol
+design, it may be worth first considering a protocol that proceeds in-order, as
+this should dramatically reduce the protocol complexity while maintaining the
+ability to verify each chunk fully. This would also be very similar to the
+blockchain reactor, which must process blocks in serial. Since any form of state sync
+is likely to be orders of magnitude faster than the blockchain sync, the
+performance loss from processing chunks in order is likely to be relatively small
+compared to a more optimized, out-of-order approach.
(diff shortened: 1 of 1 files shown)


docs/architecture/tree-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+In the former case, we descend from the root one layer at a time, accumulating
+full internal nodes into chunks. In the later case, we traverse the tree from
+say left to right, accumulting full key-value pairs into chunks. In each case,
+we require additional proof data, which consists of internal nodes that prove
+all compnents of a chunk (whether nodes or key-value pairs) are valid members of
+the tree.
+
+Note that for AVL trees, which are insertion-order dependent, that the
+depth-first approach must include all internal nodes in order to properly
+communicate the structure of the tree, which cannot be derrived from key-value
```
cwgoes:
```suggestion
communicate the structure of the tree, which cannot be derived from key-value
```

docs/architecture/tree-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+  with its index. For instance, if we receive chunk 5, the root hash of the
+  sub-tree contained therein should correspond to the 5th node in layer 10
+
+While this example provides the intuition for how the design might work,
+we now need to generalize it to handle arbitrary tree sizes and bounded
+chunk sizes. With the example above, we were only required to receive chunk 0 up
+front, and could then receive the other chunks in any order. However to
+generalize, we may be required to receive multiple chunks in order first, before
+we get to a point where we can receive them in any order.
+
```
cwgoes: For a depth-first traversal, consider the following:

- Chunkee & chunker negotiate a manifest `m` mapping chunk indices to contiguous subsets (ranges) of keyspace. They agree on `n` chunks, and they now both can map the integer range `0 - n - 1` to a subset of keyspace (start key, end key), where the end keys and start keys of contiguous chunks are identical (end is exclusive).
- Chunkee requests chunk `c_k` from chunker. Chunker sends some `c_l` where `k` might or might not be equal to `l`.
- Chunkee checks keys in `c_l` against known mapping from `k`. If keys are not within the range, chunkee rejects chunk & bans peer. Chunkee now knows that the keys do belong to chunk `c_k`, although it does not know if the values are correct.
- Chunkee now chooses random key in the chunk and requests Merkle proof of key, value pair (against known state root from light client) from chunker, or chooses random key range not in the included keys but in the chunk range and requests range proof of non-inclusion from chunker. If Merkle proof is not provided, chunkee rejects chunk and bans peer. If Merkle proof does not validate, chunkee rejects chunk and bans peer.
- Repeat random requests some `r` times with `r` a security parameter chosen according to expected peer behaviour, total number of chunks, etc.
- Chunkee now knows (in expectation) that this chunk is valid - that it has the correct keys and all the keys within this range. Chunkee applies chunk.
- Repeat for all chunks in range `0 - n - 1`.

- At the end, verify whole tree. If the state root does not match, binary search by requesting Merkle proofs of subtrees from peers and fetch any incorrect chunks. This step could be taken earlier if we're concerned about malicious peers.

For trees which can handle out-of-order-insertion, that should work in parallel as far as I can tell.

ackratos: It looks like this still doesn't solve all connected chunker are the malicious (but provide consistent chunks) case. 

If at first, the chunkee negotiates with 3 malicious peers, then it has no chance to connect to correct network.
If at first, the 2 peers send chunkee a set of chunk and keyspace mapping while other 2 peers send chunkee another kind of mapping, how the chunkee would behave (decide which 2 peers are malicious)?

If this is not the case merkle proof to resolve, I think negociate hash (which provides more information than indexes) of chunk (begin and end key can still be kept) would be easier. 
I cannot see the additional benefit merkle proof provides than hash of each chunk. Hash verification is:
1. much more efficient (without proof request-responde round trip)
2. rejecting unexpected chunks earlier.
3. doesn't need top layer chunks comes in order to achieve a verifiable tree, all chunks can be processed as long as it arrives

ackratos: > * Chunkee now chooses random key in the chunk and requests Merkle proof of key, value pair (against known state root from light client) from chunker

why choose random key for existence check? Wouldn't one whole range request be better (avoid `c_l` miss or add key but still in range)

ackratos:
> Also, what is the plan for the binance warp sync?

It's upgraded in production yesterday, you can try https://github.com/binance-chain/node-binary/tree/master/fullnode/prod/0.6.0

just turn on state_sync_reactor to `true` and state_sync_height to `0` https://github.com/binance-chain/node-binary/blob/master/fullnode/prod/0.6.0/config/config.toml#L19-L27

docs/architecture/tree-chunking.md
cwgoes: > If at first, the 2 peers send chunkee a set of chunk and keyspace mapping while other 2 peers send chunkee another kind of mapping, how the chunkee would behave (decide which 2 peers are malicious)?

Different keyspace mappings could be valid, there doesn't need to be a single canonical one - if the chunkee and chunker can't agree on the manifest (maybe the chunkee wants smaller or larger chunks), the chunkee can disconnect.

> If this is not the case merkle proof to resolve, I think negociate hash (which provides more information than indexes) of chunk (begin and end key can still be kept) would be easier.
I cannot see the additional benefit merkle proof provides than hash of each chunk.

The Merkle proof provides existence / non-existence proofs against a trusted state root - how would the chunkee know that a chunk with a particular hash was in fact the correct chunk for the keyspace?

> why choose random key for existence check? Wouldn't one whole range request be better (avoid c_l miss or add key but still in range)

Just so the proof construction, bandwidth, and verifications costs are lower. You could request proofs for the whole chunk, it would be more expensive.

melekes: > how would the chunkee know that a chunk with a particular hash was in fact the correct chunk for the keyspace?

later when the chunkee compares root hash with one in the block's header?

cwgoes: > later when the chunkee compares root hash with one in the block's header?

Sure, but then we can't validate chunks as we receive them, so we might waste time receiving & storing a bunch of chunks which later turn out to be invalid.

melekes: How hard it is to construct the chunk with **a particular hash** with invalid data inside?

ackratos: > Different keyspace mappings could be valid, there doesn't need to be a single canonical one - if the chunkee and chunker can't agree on the manifest (maybe the chunkee wants smaller or larger chunks), the chunkee can disconnect.

Then how chunkee knows keyspace at first?

cwgoes: > Then how chunkee knows keyspace at first? Without fixed-size chunk, you cannot support eager state sync right?

Chunker and chunkee negotiate on a manifest mapping integer chunk indices to bounded subsets of (the entire) keyspace. What do fixed-size chunks have to do with eager state sync? The chunker and chunkee would also need to pick a block height of the state to sync, if that's what you mean.

ackratos: > Chunker and chunkee negotiate on a manifest mapping integer chunk indices to bounded subsets of (the entire) keyspace.

I mean chunkee is brand new node, it doesn't have any information about how many keys in total. `For entire keyspace` did you mean whole possible 32 bytes of sha256 (hash of iavl tree node) i.e. 0x10..00 - 0xFF..FF?



> What do fixed-size chunks have to do with eager state sync? The chunker and chunkee would also need to pick a block height of the state to sync, if that's what you mean.

I mean think about there are 100 chunkee peers and 1 chunker. All 100 chunkee want negociate different chunk size of chunker. i.e. The first chunkee wants chunk size be 1M, the second one wants 2M, the third one wants 3M, ..., the 100th chunkee want 100M. To serve 100 kinds of clients, if the chunker want eagerly prepared the chunks (without traverse ival tree each time a request comes), does it need prepare 100 kinds of chunks on disk?

cwgoes: > I mean chunkee is brand new node, it doesn't have any information about how many keys in total. For entire keyspace did you mean whole possible 32 bytes of sha256 (hash of iavl tree node) i.e. 0x10..00 - 0xFF..FF?

Yes; the manifest maps integer indices to contiguous subsets of the entire keyspace.

> I mean think about there are 100 chunkee peers and 1 chunker. All 100 chunkee want negotiate different chunk sizes of chunker. i.e. The first chunkee wants chunk size to be 1M, the second one wants 2M, the third one wants 3M, ..., the 100th chunkee want 100M. To serve 100 kinds of clients, if the chunker wants eagerly prepared the chunks (without traverse ival tree each time a request comes), does it need to prepare 100 kinds of chunks on disk?

That's right, but there could be suggested chunk sizes (which could themselves change over time); it's all in the peer-to-peer protocol and altruism is required anyways.

melekes:
If I am not mistaken, @ebuchman have promised to expand on this.

docs/architecture/tree-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+applying an out-of-order chunk is fully verifiable - ie. the chunk is a valid
+part of the Merkle tree, and correctly corresponds to its chunk index.
+Turns out this may not be possible in general.
+
+## Arbitrary Chunking
+
+There are two general approaches here - breadth-first and depth-first.
+In the former case, we descend from the root one layer at a time, accumulating
+full internal nodes into chunks. In the later case, we traverse the tree from
+say left to right, accumulting full key-value pairs into chunks. In each case,
```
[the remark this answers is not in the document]
ebuchman: We do: 

> we require additional proof data, which consists of internal nodes that prove
all compnents of a chunk (whether nodes or key-value pairs) are valid members of the tree.

docs/architecture/tree-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+we require additional proof data, which consists of internal nodes that prove
+all compnents of a chunk (whether nodes or key-value pairs) are valid members of
+the tree.
+
+Note that for AVL trees, which are insertion-order dependent, that the
```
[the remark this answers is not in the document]
ebuchman: It's not just the key-value pairs, but also the  `internal nodes that prove all compnents of a chunk (whether nodes or key-value pairs) are valid members of the tree.`

docs/architecture/tree-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+depth-first approach must include all internal nodes in order to properly
+communicate the structure of the tree, which cannot be derrived from key-value
+pairs alone. Such nodes would be part of the proof structure.
+
+The chunker can assign to each chunk a unique index, and every chunker will
+compute the same index for each chunk, assuming they follow the same
+strategy. However, it is not clear that, in general, a chunkee can verify the
```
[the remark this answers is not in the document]
ebuchman: Right but that's an external convention, not something inherent to the tree structure. I'll change the language a bit.

docs/architecture/tree-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+
+Note that for AVL trees, which are insertion-order dependent, that the
+depth-first approach must include all internal nodes in order to properly
+communicate the structure of the tree, which cannot be derived from key-value
+pairs alone. Such nodes would be part of the proof structure.
+
+The chunker can assign to each chunk a unique index, and every chunker will
+compute the same index for each chunk, assuming they follow the same
+strategy. However, it is not clear that, in general, a chunkee can verify the
+chunk index, as there is no explicit map from chunk index to chunk or vice
```
ebuchman:
```suggestion
chunk index, as there is no inherent map from chunk index to chunk or vice
```

docs/architecture/tree-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+versa.
```
ebuchman:
```suggestion
versa. Such mappings may need to be provided externally and trusted, like in the case of Warp-Sync.
```

docs/architecture/tree-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+
+This raises two concerns for a state sync protocol:
+- how to handle mis-indexed chunks (ie. requesting chunk 3 and receiving chunk
+  4)
+- how to handle non-indexed chunks (ie. receiving a chunk that doesn't
+  correspond to any index from the given chunker strategy
+
+In each case, it appears that liveness is made much more difficult by the need
+to figure out what went wrong with the applied chunks. However, these problems
+can be eliminated by requiring chunks to be applied in-order.
```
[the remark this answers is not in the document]
ebuchman: > What we do is calculate a mapping from iavl nodes in chunk to its multistore.

Right but isn't your mapping only verifiable according to the manifest, which depends on majority honest peers? This write up is specifically about chunking in the context of full light client security.

docs/architecture/tree-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+
+### MisIndexed Chunks
```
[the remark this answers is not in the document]
ebuchman: Again, this doc is in the context of not having a manifest.

ebuchman:
> If I am not mistaken, @ebuchman have promised to expand on this.

I think we probably want to consolidate with ADR 042. Maybe ADR-042 should actually have its own folder and we can put multiple files in there as we work this all out. 

We still need to actually write up an ADR for the initial design - let's chat about it at next team meeting

PR #3809, opened 2019-07-16
abci/server: recover from app panics in socket server

### LINKED ISSUE #3800: The ability to cleanup when application panics in any of the routines spawned by s.acceptConnectionsRoutine().
Tested with tendermint 0.31.5.

The issue is that when something goes wrong in abci app, for example, in `InitChain` method - we are forced to panic, as there is no other way to return an error and stop the app.

When this happens we cannot recover, because these panics happen in a goroutine 

```
func (s *SocketServer) handleRequests(
```
which in turn is being spawned by
```
func (s *SocketServer) acceptConnectionsRoutine()
```

What this leads to is the socket not being closed and cleaned up, so when the application fails once it is not able to start again, there is a manual step required, which is basically `rm /tmp/app.socket`.

There are different options to fix that:
1. Recover from panic, log the error and allow main thread to exit gracefully, which will let the code looking like this:

```
	svr, err := server.NewServer(addr, "socket", app)
	if err != nil {
		return errors.Wrap(err, "failed to create a listener")
	}

	svr.SetLogger(logger.With("module", "abci-server"))

	done := make(chan bool)
	cleanupCallback := func() {
		// Cleanup
		_ = svr.Stop()
		done <- true
	}

	cmn.TrapSignal(logger, cleanupCallback)

	defer func() {
		if err := recover(); err != nil {
			logger.Error("recovered from panic", "err", err)
			cleanupCallback()
		}
	}()

	err = svr.Start()
	if err != nil {
		return errors.Wrap(err, "failed to start a server")
	}
	fmt.Println("123")
	// wait forever
	<-done

	return nil

```
to do proper cleanup. 

2. Alternatively allow passing a cleanup callback to the socket server, and call that on `recover()` and then propagate panic. In our case such callback would still be something like
```func() {svr.Stop()}```. I think the solution number one is preferred, because it would be easier to cleanup in the main thread as shown above. 

3. Yet another solution is to modify ABCI interface to be able to return errors in responses or alongside them. Then instead of panicking the app could just return an error.

### PR DESCRIPTION
**Pull request by ruseinov**
fixes #3800

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -18,6 +18,6 @@ program](https://hackerone.com/tendermint).
 ### FEATURES:
 
 ### IMPROVEMENTS:
-- [abci] Recover from application panics in `server/socket_server.go` to allow socket cleanup (@ruseinov)
+- [abci] \#3809 Recover from application panics in `server/socket_server.go` to allow socket cleanup (@ruseinov)
 
 ### BUG FIXES:
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -18,5 +18,6 @@ program](https://hackerone.com/tendermint).
 ### FEATURES:
 
 ### IMPROVEMENTS:
+- Recover from application panics in `abci/server/socket_server.go` to allow socket cleanup.
 
 ### BUG FIXES:
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -18,6 +18,6 @@ program](https://hackerone.com/tendermint).
 ### FEATURES:
 
 ### IMPROVEMENTS:
-- Recover from application panics in `abci/server/socket_server.go` to allow socket cleanup.
+- [abci] Recover from application panics in `server/socket_server.go` to allow socket cleanup (@ruseinov)
 
 ### BUG FIXES:
--- a/abci/server/socket_server.go
+++ b/abci/server/socket_server.go
@@ -172,6 +172,7 @@ func (s *SocketServer) handleRequests(closeConn chan error, conn net.Conn, respo
 
 		if panicVal != nil {
 			closeConn <- fmt.Errorf("recovered from panic: %v", panicVal)
+			return
 		}
 	}
 }
--- a/abci/server/socket_server.go
+++ b/abci/server/socket_server.go
@@ -144,8 +144,15 @@ func (s *SocketServer) waitForClose(closeConn chan error, connID int) {
 
 // Read requests from conn and deal with them
 func (s *SocketServer) handleRequests(closeConn chan error, conn net.Conn, responses chan<- *types.Response) {
+	var panicVal interface{}
 	var count int
 	var bufReader = bufio.NewReader(conn)
+
+	defer func() {
+		// make sure to recover from any app-related panics to allow proper socket cleanup
+		panicVal = recover()
+	}()
+
 	for {
 
 		var req = &types.Request{}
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: abci/server/socket_server.go [hunk 1]
```diff
@@ -154,14 +161,18 @@ func (s *SocketServer) handleRequests(closeConn chan error, conn net.Conn, respo
...
+				closeConn <- fmt.Errorf("error reading message: %v", err)
 			}
 			return
 		}
 		s.appMtx.Lock()
 		count++
 		s.handleRequest(req, responses)
 		s.appMtx.Unlock()
+
+		if panicVal != nil {
```
melekes: why not do it inside the defer?

abci/server/socket_server.go
```diff
@@ -154,14 +161,18 @@ func (s *SocketServer) handleRequests(closeConn chan error, conn net.Conn, respo
...
+			closeConn <- fmt.Errorf("recovered from panic: %v", panicVal)
```
melekes: we also need to unlock the application mutex, no? `s.appMtx.Unlock()`

CHANGELOG_PENDING.md
```diff
@@ -18,5 +18,6 @@ program](https://hackerone.com/tendermint).
 ### FEATURES:
 
 ### IMPROVEMENTS:
+- Recover from application panics in `abci/server/socket_server.go` to allow socket cleanup.
```
```suggestion

- [abci] Recover from application panics in `server/socket_server.go` to allow socket cleanup (@ruseinov)
```

abci/server/socket_server.go
ruseinov: Because returning from a defer does not break the execution, this is a good catch btw - I will add a return.

abci/server/socket_server.go
ruseinov: AFAIK the execution is going to continue on the next line after the panic happened, so this would be `s.appMtx.Unlock()`, so by the time we are at this condition the mutex has already been unlocked.

CHANGELOG_PENDING.md
ruseinov: this is addressed, I didn't realise I can commit it directly from here :/

ruseinov:
@melekes scratch my previous comments, review comments are addressed now!

abci/server/socket_server.go
```diff
@@ -151,6 +151,8 @@ func (s *SocketServer) handleRequests(closeConn chan error, conn net.Conn, respo
 	defer func() {
 		// make sure to recover from any app-related panics to allow proper socket cleanup
 		panicVal = recover()
+		closeConn <- fmt.Errorf("recovered from panic: %v", panicVal)
```
melekes: you forgot the `if panicVal != nil` part

ruseinov: yep, figured it out as I reviewed the code, so it's all good now!

CHANGELOG_PENDING.md
```diff
@@ -18,5 +18,6 @@ program](https://hackerone.com/tendermint).
 ### FEATURES:
 
 ### IMPROVEMENTS:
+- [abci] Recover from application panics in `server/socket_server.go` to allow socket cleanup (@ruseinov)
```
```suggestion
- [abci] \#3809  Recover from application panics in `server/socket_server.go` to allow socket cleanup (@ruseinov)
```

ruseinov:
@melekes feel free to merge this when appropriate as I'm not authorised to do so.

ruseinov:
@melekes curious if this can be backported to a `0.31.x` release for us to be able to use it without integrating new tendermint changes. 

I will, however, consider backporting this in our local repo otherwise.

melekes:
> curious if this can be backported to a 0.31.x release for us to be able to use it without integrating new tendermint changes.

can be done

PR #3824, opened 2019-07-22
p2p: Fix error logging for connection stop

melekes:
Original author and PR: @defunctzombie https://github.com/tendermint/tendermint/pull/3644

---

This changeset fixes two types of false-positive errors occurring during
connection shutdown.

The first occurs when the process invokes FlushStop() or Stop() on a
connection. While the previous behavior did properly wait for the sendRoutine
to finish, it did not notify the recvRoutine that the connection was shutting
down. This would cause the recvRouting to receive and error when reading and
log this error. The changeset fixes this by notifying the recvRoutine that
the connection is shutting down.

The second occurs when the connection is terminated (gracefully) by the other side.
The recvRoutine would get an EOF error during the read, log it, and stop the connection
with an error. The changeset detects EOF and gracefully shuts down the connection.

I am not intimately familiar with idiomatic go practices so if there is a better approach for performing the signaling I did please let me know and I will update the PR. I used a channel buffer size of 1 to ensure the signal is received and non-blocking messaging to avoid breaking existing logic.


melekes:
The p2p tests have been running for 20 min now. The node(s) must be blocked on select or smth.

/cc @defunctzombie

defunctzombie:
@melekes did the tests run successfully on your local machine?

melekes:
I haven't, but I'm pretty sure if I do it, I will get the same result.

Restarted the tests => test_p2p is running forever

melekes:
```
starting test client container with CMD=test/p2p/fast_sync/check_peer.sh 1
Other peer is on height 14 with state "0800000000000000"

Waiting for peer 1 to catch up

... 0

... 0

... 0
```

melekes:
@defunctzombie I think we have to call stopForError because otherwise peer will stay in the peer set. see https://github.com/tendermint/tendermint/pull/3824/commits/6ff48b68226f3f8dc29b70b007d0bb9e11dc03fc for details

brapse:
In the following snippet, isn't it the goal/conclusion to break out of the loop?
https://github.com/tendermint/tendermint/blob/7a9d76e307e83dbc01aa47d1e35ebd7351f70957/p2p/conn/connection.go#L566-L582

If so then wouldn't it be possible to remove the channel checking select as the result is to break out of the for loop on line 582?

One problem which might cause live locks seems to be the call to `stopServices` might cleanup up the tickers or it might not depending on the contents of the two channels. It would be preferable here to remove any variance to the cleanup process to remove any cases in which `stopService` is called but is unable to cleanup the channels, tickers and any other resources which would create a zombie process 💀

Would it be possible here to use the lifecycle management facilities provided BaseService?

https://github.com/tendermint/tendermint/blob/7a9d76e307e83dbc01aa47d1e35ebd7351f70957/libs/common/service.go#L156-L178

This way we could potentially unify lifecycle management and the call to `IsRunning` on line 574 would have the semantics we want and eliminate any need to additional channels. 🤔

melekes:
> If so then wouldn't it be possible to remove the channel checking select as the result is to break out of the for loop on line 582?

we can, but then we'll log the error - this PR fixes that

> Would it be possible here to use the lifecycle management facilities provided BaseService?

Ideally, yes. But since `FlushStop` exists and it does not call `Stop()`, we are forced to use 

1) mutex to coordinate between Stop and FlushStop 
2) channels (not the `Quit()` standard channel) to signal exiting

brapse:
Ah thanks that makes sense. I would agree that the lifecycle management cocnern is out scope so i made a follow up issue #3833. With the non-determinism fixed I think we are good to go 👍

PR #3834, opened 2019-07-25
Do not write 'Couldn't connect to any seeds' if there are no seeds

folex:
* [ ] Wrote tests

I'm setting up all peers dynamically by calling `dial_peers`, so `p2p.seeds` in configs is empty, and I'm seeing error log a lot in logs.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -20,7 +20,7 @@ program](https://hackerone.com/tendermint).
 
 ### IMPROVEMENTS:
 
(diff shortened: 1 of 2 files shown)


CHANGELOG_PENDING.md
```diff
@@ -20,6 +20,7 @@ program](https://hackerone.com/tendermint).
 
 ### IMPROVEMENTS:
 
+- [p2p] Do not write 'Couldn't connect to any seeds' error log if there are no seeds in config file
```
```suggestion
- [p2p] \#3834 Do not write 'Couldn't connect to any seeds' error log if there are no seeds in config file
```

PR #3840, opened 2019-07-29
Zm fork accountability

mikeygbooth99:
* [x] Wrote tests


docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+signing messages contributed to a forked chain although they are not part of the validator set on the main chain.
+This attack can also be used to attack full node during a period of time he is eclipsed. 
+
+### Lunatic validator
+
+Lunatic validator agrees to sign commit messages for arbitrary application state. It is used to attack lite clients. 
+Note that detecting this behavior require application knowledge. Detecting this behavior can probably be done by
+referring to the block before the one in which height happen.  
+
+Q: can we say that in this case a validator ignore to check if proposed value is valid before voting for it?
```
melekes: I think so, yes and we should punish for it

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+There are several scenarios in which fork might happen:
+
+1. Equivocation: faulty validators sign multiple vote messages (prevote and/or precommit) for different values during
+   the same round r at a given height h. 
+
+2. Amnesia: faulty validators precommit some value v in round r (value v is locked in round r) and then prevote for different value v' in higher round r' > r without previously correctly unlocking value v. In this case faulty processes "forget" that they have locked value v and prevote some other value in the following rounds.
+
+3. Back to the past: faulty validators precommit some value v in round r from the past in which they haven't voted before. In this case some other value has been decided in a round r' > r, but there are some precommit messages signed by (correct) processes for value v in the round r. 
+
+4. Fantom validators: faulty validators vote (sign prevote and precommit messages) in heights in which they are not part of the validator sets (at the main chain).
```
cwgoes:
```suggestion
4. Phantom validators: faulty validators vote (sign prevote and precommit messages) in heights in which they were not part of the validator set (in the canonical chain).
```

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+
+5. Lunatic validator: faulty validator that sign vote messages to support arbitrary application state.
```
cwgoes: Specifically, we mean application states other than those which resulted from valid state transitions, right?

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+## Types of victims
+
+We consider three types of potential attack victims: 
+
+- full node
+- lite client with sequential header verification (LCS) and 
+- lite client with bisection based header verification (LCB).
+
+In case of full node, attack surface is constrained to the blockchain as a correct full node executes 
+all transactions so it can verify if computed application state diverge from the proposed state. Therefore,
```
cwgoes:
```suggestion
all transactions so it can verify if the computed application state diverges from the proposed state. Therefore,
```

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+full node can be tricked to follow forked chain by committing different set of transactions to a blockchain.
```
cwgoes: I don't quite understand the final clause of this sentence - what is "tricked" and what is the "forked chain" in this case?

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+
+In case of lite clients, attack surface is bigger as lite client only verified headers without executing transactions.
```
cwgoes:
```suggestion
In case of lite clients, the attack surface is bigger as a lite client only verifies headers without executing transactions.
```

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+Therefore, faulty nodes can in some cases attack lite clients by arbitrarily changing state if there is a sufficient
+number of voting power that can commit such state.
```
cwgoes:
```suggestion
amount of voting power which commits such a state.
```

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+
+In case of full node, attack surface is constrained to the blockchain as a correct full node executes 
+all transactions so it can verify if computed application state diverge from the proposed state. Therefore,
+full node can be tricked to follow forked chain by committing different set of transactions to a blockchain.
+
+In case of lite clients, attack surface is bigger as lite client only verified headers without executing transactions.
+Therefore, faulty nodes can in some cases attack lite clients by arbitrarily changing state if there is a sufficient
+number of voting power that can commit such state. 
+
+Q: Is there an important difference between LCS and LCB, i.e., are there some attacks that are possible with LCB but not
```
cwgoes: I think this may depend on the particulars (e.g. thresholds) of the bisection algorithm used by LCB.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+precomit for both values A and B, so we have commits for both values A and B. This is the attack on the full node level,
+that of course extends also to the lite clients, but it is easily detected at the main chain level as we assume that 
+correct nodes are able to eventually talk to each other. Creating an evidence of misbehavior is simple in this case
+as we have multiple messages signed by the same processes for different values in the same round. 
+
+In case of lite clients, a variant of this attack is possible where faulty activity is visible only to a lite clients.
+In this case +2/3 of voting power is needed. Note that once equivocation is used to attack lite client it opens space
+for different kind of attacks as application state can be diverged in any direction. For example it can modify validator set such that it contains only validators that don't have any stake bonded. In order to detect such (equivocation based attack) lite client need to cross check it's state with some correct validator (or to obtain
+hash of the state from the main chain using out of band channels). In case of equivocation attack client would
```
cwgoes: I don't quite understand. Shouldn't the light client only need to broadcast the signatures it has received (which would be equivocations vis-a-vis the canonical chain) to a full node, which can then perform the evidence submission?

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+as we have multiple messages signed by the same processes for different values in the same round. 
+
+In case of lite clients, a variant of this attack is possible where faulty activity is visible only to a lite clients.
+In this case +2/3 of voting power is needed. Note that once equivocation is used to attack lite client it opens space
+for different kind of attacks as application state can be diverged in any direction. For example it can modify validator set such that it contains only validators that don't have any stake bonded. In order to detect such (equivocation based attack) lite client need to cross check it's state with some correct validator (or to obtain
+hash of the state from the main chain using out of band channels). In case of equivocation attack client would 
+be able to create evidence of misbehavior but this would require to pull potentially lot of data from 
+correct full nodes. Maybe we need to figure out different architecture where client that is attacked with push
+all it's data for the current unbonding period to a correct node that will inspect this data and submit corresponding
```
cwgoes: What about the algorithm at the moment prevents this from being possible?

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+evidence. There are also architectures that assumes a special role (sometimes called fisherman) whose goal is to collect
```
cwgoes: Specifically, "fisherman" (in Polkadot's parlance) are responsible for collecting & submitting evidence of signatures over *invalid state transitions*, which I don't think is what we're talking about here, so it might not be the right term.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+this document https://docs.google.com/document/d/11ZhMsCj3y7zIZz4udO9l25xqb0kl7gmWqNpGVRzOeyY/edit?usp=sharing. 
+
+If a lite client is attacked using this attack with +1/3 of voting power, attacker can not arbitrarily change application state. In case there is an attack with +2/3 of voting power, an attacker can arbitrarily change application state. Note that in this case at least +1/3 of faulty validators locked value A in round r, so if they sign different value in follow up rounds this will be detectable by the the fork accountability mechanisms. The remaining 1/3 of faulty processes might not locked value A in round r so they can't be detected using this mechanism. Only in case they signed something which conflicts with the application this can be used against them. Otherwise they don't do anything incorrect. Note that this case is not covered by the report https://docs.google.com/document/d/11ZhMsCj3y7zIZz4udO9l25xqb0kl7gmWqNpGVRzOeyY/edit?usp=sharing as it only assumes at most 2/3 of faulty validators.
+
+Q: do we need to define a special kind of attack for the case where a validator sign arbitrarily state? It seems that
```
cwgoes: Can we simply slash bonded or unbonding validators for signing (with their consensus keys) *anything* except the actual application state committed on the canonical chain - or are you talking about a case where the application state could have been valid (if it was signed in a different round from the canonical commit)?

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+detecting such attack requires different mechanism that would require as an evidence a sequence of blocks that lead to
+that state. This might be very tricky to implement. 
+
+### Back to the past
+
+In this kind of attacks faulty validators take advantage of the fact that he hasn't signed messages in some
```
cwgoes:
```suggestion
In this kind of attack a faulty validator takes advantage of the fact that he hasn't signed messages in some 
```

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+Together with precomit messages of correct processes this for a commit for value A. Note that only a faulty validator
+that previously precomit nil did equivocation, while other 1/3 of faulty validators actually executed an
+attack that has exactly the same sequence of messages as part of amnesia attack. Detecting this kind of attack boil down to mechanisms for equivocation and amnesia. 
+
+Q: should we keep this as a separate kind of attacks? It seems that equivocation, amnesia and fantom validators 
+are the only kind of attack we need to support and this gives us security also in other cases. This would not
+be surprising as equivocation and amnesia are attacks that followed from the protocol and fantom attack is 
+not really an attack to Tendermint but more to the Proof of stake module. 
+
+### Fantom validators
```
cwgoes:
```suggestion
### Phantom validators
```

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+
+In case of fantom validators, processes that are not part of the current validator set but are still bonded (
```
cwgoes:
```suggestion
In case of phantom validators, processes that are not part of the current validator set but are still bonded (
```

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+as attack happen during their unbonding period) can be part of the attack by signing vote messages. This attack 
+can be executed against both full nodes and lite clients.
```
cwgoes: How can it be used against full nodes which are verifying state transitions?

They should know what the actual validator set is at each block.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+In case of fantom validators, processes that are not part of the current validator set but are still bonded (
+as attack happen during their unbonding period) can be part of the attack by signing vote messages. This attack 
+can be executed against both full nodes and lite clients. 
+
+Example scenario: a lite client has a trust in a header for height h (and the corresponding validator set VS1). 
+As part of bisection header verification it verifies header at height h+k with new validator set VS2. 
+Some of the validators from VS2 signed the header at height h+k that is different from the corresponding header
+at the main chain. Detecting these kind of attacks is easy as it just requires verifying if processes are signing messages in heights in which they are not part of the validator set.  
+
+Note that we can have fantom validators based attacks as a follow up of equivocation or amnesia based attack where
```
cwgoes:
```suggestion
Note that we can have phantom validators based attacks as a follow up of equivocation or amnesia based attack where 
```

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+
+Example scenario: a lite client has a trust in a header for height h (and the corresponding validator set VS1). 
+As part of bisection header verification it verifies header at height h+k with new validator set VS2. 
+Some of the validators from VS2 signed the header at height h+k that is different from the corresponding header
+at the main chain. Detecting these kind of attacks is easy as it just requires verifying if processes are signing messages in heights in which they are not part of the validator set.  
+
+Note that we can have fantom validators based attacks as a follow up of equivocation or amnesia based attack where 
+forked state contains validators that are not part of the validator set at the main chain. In this case they keep
+signing messages contributed to a forked chain although they are not part of the validator set on the main chain.
+This attack can also be used to attack full node during a period of time he is eclipsed.
```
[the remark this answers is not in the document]
cwgoes: That is / was my understanding as well.

docs/spec/consensus/fork-accountability.md
cwgoes: Agreed, I wonder if this is complex if combined with differing rounds, it seems like we might need some sort of challenge-response where a validator which signed a state that didn't end up being part of the canonical chain must submit the sequence of transactions which led to it in order to avoid being slashed.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,142 @@
...
+such an attack and delayed message. This kind of attack can be used at both full nodes and lite clients.
+
+Example scenario: in a round r of height h we have 1/3 of correct validators precommitting a value A, 1/3 of correct
+validators precomit nil, 1/3 of faulty validators don't send any message and 1 faulty validator precommit nil.
+In some round r' > r, +1/3 of faulty validators together with 1/3 of correct that haven't locked any value commit some other value B. +1/3 of faulty validators now "go back to the past" and sign precomit message for value A in round r.
+Together with precomit messages of correct processes this for a commit for value A. Note that only a faulty validator
+that previously precomit nil did equivocation, while other 1/3 of faulty validators actually executed an
+attack that has exactly the same sequence of messages as part of amnesia attack. Detecting this kind of attack boil down to mechanisms for equivocation and amnesia. 
+
+Q: should we keep this as a separate kind of attacks? It seems that equivocation, amnesia and fantom validators
```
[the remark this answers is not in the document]
josef-widder: I think it would make sense to distinguish the case where 
(1) correct validators disagree on a block
(2) all correct validators agree, but some messages from the past are used to forge headers.

The back to the past, as I understand it, is of type (2). To me it seems less harmful, because it should be clear which is the correct block of a given height. If we have equivocation and two correct validators decide on different blocks, I think we need social consensus to decide which one was the right one (if any).

docs/spec/consensus/fork-accountability.md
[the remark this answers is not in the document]
josef-widder: Yes, I think this is what is meant by "as follow up of  equivocation or amnesia", that is, until the fork is detected we can have phantom validators creating blocks. However, if we have a fork with different blocks with different next validator sets for the same height, the question is whether, from a technical viewpoint, it makes a difference whether one is a (faulty) phantom validator or whether one is correct. At the end, we will need social consensus to resolve which block to maintain, or where to roll-back to?

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,297 @@
...
+* F2 does not lock A in round *r*
+
+Consequences:
+
+* The validators in F1 will be detectable by the the fork accountability mechanisms.
+* The validators in F2 cannot be detected using this mechanism.
+Only in case they signed something which conflicts with the application this can be used against them. Otherwise they do not do anything incorrect.
+* This case is not covered by the report https://docs.google.com/document/d/11ZhMsCj3y7zIZz4udO9l25xqb0kl7gmWqNpGVRzOeyY/edit?usp=sharing as it only assumes at most 2/3 of faulty validators.
+
+**Q:** do we need to define a special kind of attack for the case where a validator sign arbitrarily state? It seems that detecting such attack requires different mechanism that would require as an evidence a sequence of blocks that lead to that state. This might be very tricky to implement.
```
melekes: Do you mean a correct validator?

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,297 @@
...
+
+Tendermint consensus guarantees the specifications
+* agreement
+* validity, and
+* termination,
+if the
+faulty validators have at most 1/3 of voting power in the current validator set. In the case where this assumption
+does not hold, each of the specification may be violated.
+
+The agreement property says that for a given height, any to correct validators that decide on a block for that height decide on the same block. That the block was indeed generated by the blockchain, can be verified starting from a trusted (genesis) block, by checking that all subsequent blocks are properly signed.
```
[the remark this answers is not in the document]
josef-widder: I added a description

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
+# Fork accountability -- Problem statement and attacks
+
+## Problem Statement
+
+Tendermint consensus guarantees the following specifications for all heights:
+* agreement -- no two full nodes decide differently.
```
milosevic: Should probably be "no two correct full nodes".

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+
+if the
+faulty validators have at most 1/3 of voting power in the current validator set. In the case where this assumption
+does not hold, each of the specification may be violated.
+
+The agreement property says that for a given height, any two correct validators that decide on a block for that height decide on the same block. That the block was indeed generated by the blockchain, can be verified starting from a trusted (genesis) block, and checking that all subsequent blocks are properly signed.
+
+
+However, faulty nodes may forge blocks and try to convince users (lite clients) that the blocks had been correctly generated. In addition, Tendermint agreement might be violated in the case where more than 1/3 of the voting power belongs to faulty validators: Two correct validators decide on different blocks. The latter case motivates the term "fork": as Tendermint consensus also agrees on the next validator set, correct validators may have decided on disjoint next validator sets, and the chain branches into two or more partitions (possibly having faulty validators in common) and each branch continues to generate blocks independently of the other.
```
milosevic: Both full nodes and lite clients can be corrupted by forged blocks.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+
+We say that a fork is a case in which there are two commits for different block at the same height of the blockchain. The proplem is to ensure that in those cases we are able to detect faulty validators (and not mistakenly accuse correct validators), and incentivize therefore validators to behave according to the protocol specification.
```
milosevic: block -> blocks

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+
+**Conceptual Limit.** In order to prove misbehavior of a node, we have to show that the behavior deviates from correct behavior with respect to a given algorithm. Thus, an algorithm that detects misbehavior of nodes executing some algorithm *A* must be defined with respect to algorithm *A*. In our case, *A* is Tendermint consensus (+ other protocols in the infrastructure; e.g.,full nodes and the Lite Client). If the consensus algorithm is changed/updated/optimized in the future, we have to check whether changes to the accountability algorithm are also required. All the discussions in this document are thus inherently specific to Tendermint consensus and the Lite Client specification.
+
+**Q:** Should we distinguish agreement for validators and full nodes for agreement? The case where all correct validators agree on a block, but a correct full node decides on a different block seems to be slightly less severe that the case where two correct validators decide on different blocks. Still, if a contaminated full node becomes validator that may be problematic later on. Also it is not clear how gossiping is impaired if a contaminated full node is on a different branch.
```
milosevic: Recovering full node that is corrupted with a forked block is very different than the case where we have a fork on the main chain. We might want to support some kind of automatic recovery of corrupted full nodes in the future, although the simplest thing is just to do clean initialisation from trusted state/header.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+*Remark.* In the case more than 1/3 of the voting power belongs to faulty validators, also validity and termination can be broken. Termination can be broken if faulty processes just do not send the messages that are needed to make progress. Due to asynchrony, this is not punishable, because faulty validators can always claim they never received the messages that would have forced them to send messages.
+
+
+## The Misbehavior of Faulty Validators
+
+Forks are the result of faulty validators deviating from the protocol. In principle several such deviations can be detected without a fork actually occurring:
+
+1. double proposal: A faulty proposer proposes two different values (blocks) for the same height and the same round in Tendermint consensus.
+
+2. double signing: Tendermint consensus forces correct validators to prevote and precommit for at most one value per round. In case a faulty proposer proposed mutliple values, faulty validators can prevote and precommit multiple values.
```
milosevic: This sentence is a bit confusing. Maybe it's simpler just saying that a faulty validator send multiple different prevote or precommit messages in a round.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+
+3. lunatic validator: Tendermint consensus forces correct validators to prevote and precommit only for values *v* that satisfy *valid(v)*. If faulty validators prevote and precommit for *v* although *valid(v)=false* this is misbehavior.
+
+*Remark.* In isolation, Point 3 is an attack on validity (rather than agreement). However, the prevotes and precommits can then also be used to forge blocks
+
```
milosevic: Add . above.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+4. amnesia: Tendermint consensus has a locking mechanism. Validators may only prevote/precommit a value if they have not locked a different value before.
```
milosevic: It would be good adding description a bit here. If a validators has locked a value v in round r, only prevote/precommit for v or nil are possible. The current spec and implementation differ slightly here, as in the implementation requires to prevote for locked value always, while spec could prevote for nil even in case some value is locked.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+
+5. spurious messages: In Tendermint consensus most of the message send instructions are guarded by threshold guards, e.g., one needs to receive *2f + 1* prevote messages to send precommit. Faulty validators may send precommit without having received the prevote messages.
+
```
milosevic: Similarly, it could send proposal for some value v and round r without receiving *2f+1* prevote messages for v in round r.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+
+Independently of a fork happening, punishing this behavior might be important to prevent forks altogether. This should keep attackers from misbehaving: if at most 1/3 of the voting power is faulty, this misbehavior is detectable but will not lead to a safety violation. Thus, unless they have more than 1/3 (or in some cases more than 2/3) of the voting power attackers have the incentive to not misbehave. If attackers control too much voting power, we have to deal with forks, as discussed in this document.   
+
+
+## Two types of forks
+
+* Fork-Full. Two correct validators decide on different blocks for the same height. Since also the next validator sets are decided upon, the correct validators may be partitioned to participate in two distinct branches of the forked chain.
```
milosevic: Should we phrase this at the level of full nodes instead of validators? Or fork detected by the validator in case all correct validators decided the same block is different case?

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+### Equivocation (one round)
+
+There are several scenarios in which forks might happen. The first is double signing within a round.
+
+* F1. Equivocation: faulty validators sign multiple vote messages (prevote and/or precommit) for different values *during the same round r* at a given height h.
+
+
+### Flip-flopping
+
+Tendermint consensus implements a locking mechanism: If a correct validator *p* receives *2f + 1* prevotes for a value *v* in round *r*, it locks *v* and remembers *r*. In this case, *p* also sends a precommit message for *v*, which later may serve as proof that *p* signed *v*.
```
milosevic: To be more precise, it is necessary receiving proposal for value v in the current round r and *2f+1* prevote messages for id(v) in r in order to lock a value v in the round r.

milosevic: Sending precommit message for v is a sign that p locked value v in that round.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+In subsequent rounds, *p* only sends prevote messages for a value it had previously locked. However, it is possible to change the locked value if in a future round *r' > r*, if the process receives *2f + 1* prevotes for a different value *v'*. In this case, *p* sends a precommit for *v'*. This algorithmic feature can be exploited in two ways:
+
+
+
+* F2. Faulty Flip-flopping (Amnesia): faulty validators precommit some value *v* in round *r* (value *v* is locked in round *r*) and then prevote for different value *v'* in higher round *r' > r* without previously correctly unlocking value *v*. In this case faulty processes "forget" that they have locked value *v* and prevote some other value in the following rounds.
+Some correct validators might have decided on *v* in *r*, and other correct validators decide on *v'* in *r'*. Here we can have branching on the main chain (Fork-Full).
```
milosevic: Amnesia can also be used to attack lite clients without fork on the main chain. In that case, faulty validators after they have decided in round r, keep signing messages in higher rounds for different values, but send those messages only to lite clients.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+
+
+
+
+## Off-chain attacks
+
+F1-F3 may contaminate the state of full nodes (and even validators). Contaminated (but otherwise correct) full nodes may thus communicate faulty blocks to lite clients.
+Similarly, without actually interfering with the main chain, we can have the following:
+
+* F4. Phantom validators: faulty validators vote (sign prevote and precommit messages) in heights in which they are not part of the validator sets (at the main chain).
```
milosevic: Note that validity property of consensus does not cover this case, as by consensus specification processes that are not part of the validator set are not taken into account.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+
+* F5. Lunatic validator: faulty validator that sign vote messages to support (arbitrary) application state that is different from the application state that resulted from valid state transitions.
```
milosevic: This is a violation of consensus validity property.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+
+## Types of victims
+
+We consider four types of potential attack victims:
```
milosevic: three?

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+more than 2/3 of prevotes for A and B and precommit for A and B, respectively.
+* Faulty validators from the set F precommit both values A and B.
+* Thus, we have more than 2/3 commits for both A and B.
+
+Consequences:
+* Creating evidence of misbehavior is simple in this case as we have multiple messages signed by the same faulty processes for different values in the same round.
+
+* We have to ensure that these different messages reach a correct process (full node, monitor?), which can submit evidence.
+
+* If the underlying communication ensures that every message exchanged between correct full nodes is eventually received, this attack is easily detected at the main chain level
```
milosevic: Current gossip layer does not ensure such property.

docs/spec/consensus/fork-accountability.md
```diff
@@ -0,0 +1,320 @@
...
+
+Validators:
+
+* a set F of faulty validators with more than 1/3 but at most 2/3 of the voting power
+* a set C of correct validators
+
+Execution:
+
+* More than 2/3 of voting power (containing correct and faulty validators) commit a block A in round *r*.
+* All reach a round *r' > r*.
```
milosevic: Once a correct validator commit a block in some round, it does not proceed to higher rounds of that height. It starts new height with round 0.

PR #3841, opened 2019-07-29
Updated lite client spec 

### LINKED ISSUE #1413: Fixes to and for Light Client Spec

josef-widder:
closes #1413 

* [ ] Wrote tests


docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
+
+Lite Client Accuracy:
+- Assume by contradiction that *h2* was not generated correctly and the lite client sets trust to true because Bisection returns true.
+- Bisection returns true only if all calls to CheckSupport in the recursion return true.
+- Thus we have a sequence of headers that all satisfied the CheckSupport
+- again a contradiction
+
+Lite Client Completeness:
+
+This is only ensured if upon *Commit(pivot)* the lite client is always provided with a correctly generated header.
```
melekes: what if we fail to verify that commit was signed by 2/3 of initial validators? I.e.

```
votingpower_in(signers(header.Commit), h1.Header.NextV) < 2/3
```

Do we simply abort?

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
+      Commit        Commit
+    }
 ```
 
-The light client is initialised with the trusted validator set, for example based on the known validator set hash,
-validator set sequence number and the validator set init time.
-The core of the light client logic is captured by the VerifyAndUpdate function that is used to 1) verify if the given header is valid,
-and 2) update the validator set (when the given header is valid and it is more recent than the seen headers).
+  * *tp*: trusting period   
+  * for realtime *t*, the predicate *correct(v,t)* is true if the validator *v*
```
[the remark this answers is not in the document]
josef-widder: The write-up was not clear. Each bullet is supposed to be a definition we are using. Some of which actually are data structures (for which I only give the details we are using). I will clean this up. Thanks for spotting this!

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
+header. Trust can be obtained by (possibly) the combination of three methods.
+
+1. **Uninterrupted sequence of proof.** If a block is appended to the chain, where the last block
+is trusted (and properly committed by the old validator set in the next block),
+and the new block contains a new validator set, the new block is trusted if the lite client knows all headers in the prefix.
+Intuitively, a trusted validator set is assumed to not chose a new validator set
+that violates the Tendermint Failure Model.
+
+2. **Trusting period.** Based on a trusted block *h*, and the lite client
+invariant, which ensures the fault assumption during the trusting period, we can check whether at least one  validator, that has been continuously correct from *h.Header.bfttime* until now, has signed *snh*.
```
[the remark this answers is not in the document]
josef-widder: No, I am saying that I need one signature by a *correct* validator. This is checked by having signatures from more than 1/3 of the voting power. As long as we are within the Tendermint Failure Model, we do not have to revert. If the Tendermint Failure Model is violated, we will have to deal with forks and recover a good state.

docs/spec/consensus/light-client.md
josef-widder: The functions will return false. As I mentioned in one remark, in this case we have encountered a forged header, and can submit evidence. We are currently preparing a spec on fork accountability that will discuss these cases.

docs/spec/consensus/light-client.md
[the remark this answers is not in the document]
josef-widder: Also, by setting the trustlevel parameter to a value greater than 1/3 we can require signatures by validators with more voting power than 1/3.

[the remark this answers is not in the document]
melekes: Got it, thanks!

docs/spec/consensus/light-client.md
melekes: Yeah, that is what I thought, but in the pseudo code it looks like Commit always gets correct commit.

After we submit evidence, do we try another full node?

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
-type ResultValidators struct {
-    BlockHeight   int64
-    Validators    []Validator
-    // time the current validator set is initialised, i.e, time of the last validator change before header BlockHeight
-    ValSetTime    int64
-}
-```
+A lite client is a process that connects to Tendermint full nodes and then tries to verify application data using the Merkle proofs.
+
+We assume that the lite client knows a (base) header *inithead* it trusts (by social consensus or because the lite client has decided to trust the header before). Now the goal is to check whether another header *newhead* can be trusted based on the data in *inithead*.
```
cwgoes: Possibly worth clarifying what it means to trust a header in this context - are we referring to trusting some subset of the validator public keys in the header?

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
-valSet        []Validator    // current validator set (last known and verified validator set)
-valSetNumber  int64          // sequence number of the current validator set
-valSetHash    []byte         // hash of the current validator set
-valSetTime    int64          // time when the current validator set is initialised
+  * header fields
+    - *height*
+    - *bfttime*: the chain time when the header (block) was generated
+    - *V*: validator set containing validators.
+    - *NextV*: validator set for next block.
+    - *commit*: evidence that block with height *height* - 1 was committed by a set of validators (canonic commit). We will use ```signers(commit)``` to refer to the set of validators that committed the block.
```
cwgoes: canonical commit?

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
+\[
+\sum_{(v,p) \in h.Header.NextV \wedge correct(v,h.Header.bfttime + tp)} p >
+2/3 \sum_{(v,p) \in h.Header.NextV} p
+\]
+
+*Assumption*: "correct" and "faulty" are defined w.r.t. realtime (some Newtonian global notion of time, i.e., wall time), while *bfttime* corresponds to the reading of the local clock of a validator (how this time is computed may change when the Tendermint consensus is modified). In this note, we assume that all clocks are synchronized to realtime. We can make this more precise eventually (incorporating clock drift, accuracy, precision, etc.). Right now, we consider this assumption sufficient, as clock synchronization (under NTP) is in the order of milliseconds and *tp* is in the order of weeks.  
+
+*Remark*: This failure model might change to a hybrid version that takes heights into account in the future.
+
+The specification in this document considers an implementation of the lite client under this assumption. Issues like *counter-factual signing* and *fork accountability* and *evidence submission* are mechanisms to enforce this assumption, that is, incentivize validators to follow the protocol.
```
cwgoes:
```suggestion
The specification in this document considers an implementation of the lite client under this assumption. Issues like *counter-factual signing* and *fork accountability* and *evidence submission* are mechanisms to justify this assumption by incentivizing validators to follow the protocol.
```

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
+If they don't, and we have more that 1/3 faults, safety may be violated. Our approach then is to *detect* these cases (after the fact), and take suitable repair actions (automatic and social). This is discussed in an upcoming document on "Fork accountability". (These safety violations include the lite client wrongly trusting a header, a fork in the blockchain, etc.)
+
+
+## Lite Client Trusting Spec
+
+The lite client communicates with a full node and learns new headers. The goal is to locally decide whether to trust a header. Our implementation needs to ensure the following two properties:
+
+- Lite Client Completeness: If header *h* was correctly generated by the Tendermint blockchain (and its age is less than the trusting period), then the lite client should eventually set *trust(h)* to true.
```
cwgoes: I think "instance of Tendermint consensus" is a bit clearer than "blockchain"; there doesn't actually need to be a blockchain I believe.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
+
+- Lite Client Accuracy:  If header *h* was *not generated* by the Tendermint blockchain, then the lite client should never set *trust(h)* to true.
+
+*Remark*: If in the course of the computation, the lite client obtains certainty that some headers were forged by adversaries, it may submit (a subset of) the headers it has seen as evidence of misbehavior.
```
cwgoes: "forged" is unclear in this context; what is the difference between forging a header and signing one normally?

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
+
+*Remark*: In Completeness we use "eventually", while in practice *trust(h)* should be set to true before *h.Header.bfttime + tp*. If not, the block cannot be trusted because it is too old.
+
+*Assumption*: Initially, the lite client has a header *inithead* that it trusts correctly, that is, *inithead* was correctly generated by the Tendermint blockchain.
```
cwgoes: Same comment on "Tendermint blockchain"

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
+Upon initialization, the lite client is given a header *inithead* it trusts (by
+social consensus). It is assumed that *inithead* satisfies the lite client invariant. (If *inithead* has been correctly generated by the Tendermint blockchain, the invariant follows from the Tendermint Failure Model.)
+
+When a lite clients sees a signed new header *snh*, it has to decide whether to trust the new
+header. Trust can be obtained by (possibly) the combination of three methods.
+
+1. **Uninterrupted sequence of proof.** If a block is appended to the chain, where the last block
+is trusted (and properly committed by the old validator set in the next block),
+and the new block contains a new validator set, the new block is trusted if the lite client knows all headers in the prefix.
+Intuitively, a trusted validator set is assumed to not chose a new validator set
```
cwgoes: the double negative here is confusing, and what is the "Tendermint Failure Model"?

I think we mean that we assume a trusted validator set will choose only a new validator set for which we can make the same 2/3 protocol-following assumption.

docs/spec/consensus/light-client.md
josef-widder: Yes. I mentioned that further above in the text following "the lite client must retry to retrieve correct headers from another full node". Should we specify that in more detail?

mikeygbooth99:
```suggestion
This is only ensured if upon *Commit(pivot)* the lite client is always provided with a correctly generated header.
```

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
 ```
 
-The light client is initialised with the trusted validator set, for example based on the known validator set hash,
-validator set sequence number and the validator set init time.
-The core of the light client logic is captured by the VerifyAndUpdate function that is used to 1) verify if the given header is valid,
-and 2) update the validator set (when the given header is valid and it is more recent than the seen headers).
+  * *tp*: trusting period   
+  * for realtime *t*, the predicate *correct(v,t)* is true if the validator *v*
+    follows the protocol until time *t* (we will see about recovery later).
+    Similarly we define *faulty(v,t)*.
```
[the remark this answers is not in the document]
josef-widder: I guess I was trying to keep an option open for a more dynamic failure model with recovery. But you are right, this is not so clear at the moment. Given that we do not really use faulty(v,t) I will remove that.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
+    + *p* is its voting power
+
+
+### Tendermint Failure Model
+
+If a block *h* is generated at time *bfttime* (and this time is stored in the block), then validators that are correct until time *h.Header.bfttime + tp* have more than two thirds of the voting power in *h.Header.V*.
+
+Formally,
+\[
+\sum_{(v,p) \in h.Header.NextV \wedge correct(v,h.Header.bfttime + tp)} p >
```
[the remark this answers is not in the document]
josef-widder: To the best of my understanding it should be NextV also in the text above. I believe the proper definition is that the trusting period starts the moment a validator set is decided on in NextV.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
+2/3 \sum_{(v,p) \in h.Header.NextV} p
+\]
+
+*Assumption*: "correct" and "faulty" are defined w.r.t. realtime (some Newtonian global notion of time, i.e., wall time), while *bfttime* corresponds to the reading of the local clock of a validator (how this time is computed may change when the Tendermint consensus is modified). In this note, we assume that all clocks are synchronized to realtime. We can make this more precise eventually (incorporating clock drift, accuracy, precision, etc.). Right now, we consider this assumption sufficient, as clock synchronization (under NTP) is in the order of milliseconds and *tp* is in the order of weeks.  
+
+*Remark*: This failure model might change to a hybrid version that takes heights into account in the future.
```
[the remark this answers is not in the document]
josef-widder: We discussed this last week, when you were not there. Adding a height-based trusting period on-chain seems to be simple. 
Off-chain, for the lite client, it is less clear. If a lite client trusts a header with height h, and it wants to check a header with height h+k, the lite client must figure out what is the current height on the chain. If the height on the chain has exceeded h + tp_h, the validators that signed at height h are not trusted anymore, and could have forged an arbitrary header with height h+k in the meantime, without violating a height-based Tendermint failure model.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
+*Assumption*: Initially, the lite client has a header *inithead* that it trusts correctly, that is, *inithead* was correctly generated by the Tendermint blockchain.
+
+To reason about the correctness, we may prove the following invariant.
+
+*Lite Client Invariant.*
+ For each lite client *l* and each header *h*:
+if *l* has set *trust(h) = true*,
+  then validators that are correct until time *h.Header.bfttime + tp* have more than two thirds of the voting power in *h.Header.NextV*.
+
+  Formally,
```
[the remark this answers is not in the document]
josef-widder: Yes, the expression is the same. The difference is that here the precondition is that the lite client trusts the header, while before we talk about generating the header. I will try to emphasize this in the write-up.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
+
+**Auxiliary Functions.** We will use the  function ```votingpower_in(V1,V2)``` to compute the voting power the validators in set V1 have according to their voting power in set V2;
+ ```votingpower_in(V,V)``` returns the total voting power in V.
+We further use the function ```signers(Commit)``` that returns the set of validators that signed the Commit.
+
+**CheckSupport.** The following function checks whether we can trust the header h2 based on header h1 following the trusting period method.
+
+```go
+  func CheckSupport(h1,h2,trustlevel) bool {
+    if h1.Header.bfttime + unbonding_period < now { // Observation 1
```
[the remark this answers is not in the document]
josef-widder: should be tp. yes.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
+      return false // old header was once trusted but it is expired
+    }  
+    vp_all := votingpower_in(h1.Header.NextV,h1.Header.NextV)
+      // total sum of voting power in h1
+
+    if h2.Header.height == h1.Header.height + 1 {
+      // specific check for adjacent headers; everything must be
+      // properly signed.
+      // Plus the following check that 2/3 of the voting power
+      // in h1 signed h2
```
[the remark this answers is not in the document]
josef-widder: Yes. I guess there are some hashes, etc., that should be checked. I am not sure I know all the details what needs to be checked here. Are you saying that we should add the check "h2.Header.V == h1.Header.NextV"? Is this all?

docs/spec/consensus/light-client.md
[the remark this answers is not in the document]
cwgoes: Right; the only local reference the light client has is a local clock, so some synchrony assumptions on time are unavoidable.

I still think it's prudent to make the unbonding period a minimum of time & height though, it should give us a somewhat stronger property about when evidence would be valid.

docs/spec/consensus/light-client.md
josef-widder: double negation is a Germanic disease ;-)  I will fix that

docs/spec/consensus/light-client.md
josef-widder: I will add a clarification

docs/spec/consensus/light-client.md
josef-widder: I changed that throughout the text for the next version

docs/spec/consensus/light-client.md
josef-widder: That is true. For the next version, I added a comment  that this is an underlying assumption that the user has to make sure is justified.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,281 @@
...
+  * For each header *h* it has locally stored, the lite client stores whether
+    it trusts *h*. We write *trust(h) = true*, if this is the case.
+  * Validator fields. We will write a validator as a tuple *(v,p)* such that
+    + *v* is the identifier (we assume identifiers are unique in each validator set)
+    + *p* is its voting power
+
+
+### Tendermint Failure Model
+
+If a block *h* is generated at time *bfttime* (and this time is stored in the block), then validators that are correct until time *h.Header.bfttime + tp* have more than two thirds of the voting power in *h.Header.V*.
```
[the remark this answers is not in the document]
josef-widder: I do not require the *v* to stay in the validators. But they indeed need to be correct, that is, stop signing if they are not validators anymore. I suspect there is something to clarify about the relationship of tp and the unbonding period here?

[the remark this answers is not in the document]
josef-widder: ... and I rephreased it similar to Ethan's suggestion for the next version.

docs/spec/consensus/light-client.md
josef-widder: changed that.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,313 @@
...
+3. **Bisection.** If a check according to the trusting period fails, the lite client can try to obtain a header *hp* whose height lies between *h* and *snh* in order to check whether *h* can be used to get trust for *hp* and *hp* can be used to get trust for *snh*. If this is the case we can trust *snh*; if not, we may continue recursively.
+
+## How to use it
+
+We consider the following use case:
+ the lite client wants to verify a header for some given height *k*. Thus:
+  - it requests the signed header for height *k* from a full node
+  - it tries to verify this header with the methods described here.
+
+This can be used in several settings
```
ancazamfir:
```suggestion
This can be used in several settings:
```

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,313 @@
...
+1. *tp < unbonding period*.
+2. *snh.Header.bfttime < now*
+3. *snh.Header.bfttime < h.Header.bfttime+tp*
+4. *trust(h)=true*
+
+
+**Observation 1.** If *h.Header.bfttime + tp > now*, we trust the old
+validator set *h.Header.NextV*.
+
+When we say we trust *h.Header.NextV* we do *not* trust that each individual validator in *h.Header.NextV* is correct, but we only trust the fact that at most 1/3 of them is faulty (more precisely, the faulty ones have less than 1/3 of the total voting power).
```
ancazamfir:
```suggestion
When we say we trust *h.Header.NextV* we do *not* trust that each individual validator in *h.Header.NextV* is correct, but we only trust the fact that at most 1/3 of them are faulty (more precisely, the faulty ones have less than 1/3 of the total voting power).
```

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,313 @@
...
+
+**CheckSupport.** The following function checks whether we can trust the header h2 based on header h1 following the trusting period method.
+
+```go
+  func CheckSupport(h1,h2,trustlevel,trustlevel_adj) bool {
+    if h1.Header.bfttime + tp < now { // Observation 1
+      return false // old header was once trusted but it is expired
+    }  
+    vp_all := votingpower_in(h1.Header.NextV,h1.Header.NextV)
+      // total sum of voting power in h1
```
ancazamfir: This is the total voting power for the validator set for h2. The comment is a bit misleading.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,313 @@
...
+        // headers in Cosmos are typically signed by all validators, so one
+        // may request that all signed (liveness might be endangered)
+    }
+
+    s := intersection(h1.Header.NextV,signers(h2.Commit))
+      // get validators in h1 that signed h2
+    vps := votingpower_in(s,h1.Header.NextV)
+      // sum of voting powers in h1 of
+      // validators that signed h2
+    return (vps > max(1/3,trustlevel) * vp_all)
```
ancazamfir: I assume we do more verification here before we return true if `(vps > max(1/3,trustlevel) * vp_all)` holds.
Should still check signatures and more than 2/3rds signed, like in the sequential case. This could be done in the beginning for both cases.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,313 @@
...
+
+
+*Correctness arguments*
+
+Towards Lite Client Accuracy:
+- Assume by contradiction that *h2* was not generated correctly and the lite client sets trust to true because *CheckSupport* returns true.
+- h1 is trusted and sufficiently new
+- by Tendermint Fault Model, less than 1/3 of voting power held by faulty validators => at least one correct validator *v* has signed *h2*.
+- as *v* is correct up to now, it followed the Tendermint consensus protocol at least up to signing *h2* => *h2* was correctly generated, we arrive at the required contradiction.
+
```
ancazamfir: I wish we had more clear definitions on the type of faults and what this client proposal must/should/may catch. We talk about "conflicts", "forged headers", "...not generated correctly", etc.
What happens if two different clients with same h1 are presented with two different headers h2 (fork). It is possible that they make inconsistent decisions and will each accept their version and continue (??). And the way to deal with this is outside the scope (fork accountability)?

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,313 @@
...
+This is only ensured if upon *Commit(pivot)* the lite client is always provided with a correctly generated header.
+
+*Stalling*
+
+With Bisection, a faulty full node could stall a lite client by creating a long sequence of headers that are queried one-by-one by the lite client and look OK, before the lite client eventually detects a problem. There are two ways to address this:
+* Each call to ```Commit``` could be issued to a different full node
+* Instead of querying header by header, the lite client tells a full node which header it trusts, and the height of the header it needs. The full node responds with the header along with a proof consisting of intermediate headers that the light client can use to verify. Roughly, Bisection would then be executed at the full node.
+
+
+### The case *h2.Header.height < h1.Header.height*
```
ancazamfir: What is this section for? Could you add more comments/ explanations if it is used by the spec?

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,313 @@
...
+
+We iterate down the heights and check the hashes in each step.
+
+```go
+func Backwards(h1,h2) bool {
+  assert (h2.Header.height < h1.Header.height)
+  old := h1
+  for i := h1.Header.height - 1; h2.Header.height < i; i-- {
```
ancazamfir:
```suggestion
  for i := h1.Header.height - 1; i > h2.Header.height; i-- {
```

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,313 @@
...
+We further use the function ```signers(Commit)``` that returns the set of validators that signed the Commit.
+
+**CheckSupport.** The following function checks whether we can trust the header h2 based on header h1 following the trusting period method.
+
+```go
+  func CheckSupport(h1,h2,trustlevel,trustlevel_adj) bool {
+    if h1.Header.bfttime + tp < now { // Observation 1
+      return false // old header was once trusted but it is expired
+    }  
+    vp_all := votingpower_in(h1.Header.NextV,h1.Header.NextV)
```
melekes: Since it's a pseudo-code, wouldn't it be better to write smth like:

```suggestion
    vp_all := totalVotingPower(h1.Header.NextV)
```

?

docs/spec/consensus/light-client.md
melekes: https://github.com/tendermint/tendermint/pull/3840

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,313 @@
...
+- h1 is trusted and sufficiently new
+- by Tendermint Fault Model, less than 1/3 of voting power held by faulty validators => at least one correct validator *v* has signed *h2*.
+- as *v* is correct up to now, it followed the Tendermint consensus protocol at least up to signing *h2* => *h2* was correctly generated, we arrive at the required contradiction.
+
+
+Towards Lite Client Completeness:
+- The check is successful if sufficiently many validators of *h1* are still validators in *h2* and signed *h2*.
+- If *h2.Header.height = h1.Header.height + 1*, and both headers were generated correctly, the test passes
+
+*Assumption*: We may need a Tendermint invariant stating that if *h2.Header.height = h1.Header.height + 1* then *signers(h2.Commit) \subseteq h1.Header.NextV*.
```
melekes: Does this mean we should forbid changing validator set completely?

docs/spec/consensus/light-client.md
ancazamfir: Exactly what I was looking for, thanks!

docs/spec/consensus/light-client.md
josef-widder: I was thinking of a use case where the lite clients trusts a recent header but needs to verify an older one; someone told the user that there is useful information in the older block. In this case we need this method for verification. Does that make sense? I will add an explanation.

docs/spec/consensus/light-client.md
josef-widder: yes, makes sense. Thanks!

docs/spec/consensus/light-client.md
josef-widder: No, this is just a remark to make explicit that the validators who sign h2 must be in h1.Header.NextV. This is enforced by Tendermint consensus. I just collect properties that are important to capture for automated verification. But I guess I should mark them differently to make this clear in the write-up.

docs/spec/consensus/light-client.md
josef-widder: You are right. Good catch! Thanks!

docs/spec/consensus/light-client.md
josef-widder: Yes. When we receive a header we should do all checks to verify that the header is "well-formed". Is there a list of checks that we do in the sequential case that I could point to?

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,313 @@
...
-```
+A lite client is a process that connects to Tendermint full nodes and then tries to verify application data using the Merkle proofs.
+
+## Context of this document
+
+In order to make sure that full nodes have the incentive to follow the protocol, we have to address the following three Issues
+
+1) The lite clients needs a method to verify header it obtains from full nodes -- addressed in this document)
+2) Detect conflicts/forged headers -- when the verification method of this document fails, then a conflict is detected. However, this document does not address to do this in an efficient way, or with special attacks in mind. This will be treated in fork accountability.
+3) How can evidence for conflicts be reported, and to whom.
```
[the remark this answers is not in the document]
josef-widder: No. The plan is to have a third document for that, and eventually merge the spefications 1-3 to have the complete picture.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,313 @@
...
-valSetHash    []byte         // hash of the current validator set
-valSetTime    int64          // time when the current validator set is initialised
+In the following, I only give the details of the data structures needed for this specification.
+
+  * header fields
+    - *height*
+    - *bfttime*: the chain time when the header (block) was generated
+    - *V*: validator set containing validators.
+    - *NextV*: validator set for next block.
+    - *commit*: evidence that block with height *height* - 1 was committed by a set of validators (canonical commit). We will use ```signers(commit)``` to refer to the set of validators that committed the block.
```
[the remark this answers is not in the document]
josef-widder: In the specification we use the commit stored in a signed header returned by Commit(height). The commit h2.header.commit will be used for the specific checks of adjacent headers that are not describe in detail in this spec. These checks should be similar to the sequential case.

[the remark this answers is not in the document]
ancazamfir: So can we remove then the commit from the header or say it's for `height` and remove from `SignedHeader`?
The current datastructures are a bit confusing to me. 
In the TM current code the header includes only the first two fields from what you have there (`*height*` and `*bfttime*`). In addition it has a bunch of hashes for the validator sets and other info.
Then:
```go
type SignedHeader struct {
	*Header `json:"header"`
	Commit  *Commit `json:"commit"`
}
```
Not clear if `Commit` is filled by the "provider" with the canonical or seen commit. But it is definitely for the `Height` specified by `Header`.

```go
type FullCommit struct {
	SignedHeader   types.SignedHeader  `json:"signed_header"`
	Validators     *types.ValidatorSet `json:"validator_set"`
	NextValidators *types.ValidatorSet `json:"next_validator_set"`
}

type Provider interface {
...
	// LatestFullCommit returns the latest commit with minHeight <= height <= maxHeight.
	// If maxHeight is zero, returns the latest where minHeight <= height.
	LatestFullCommit(chainID string, minHeight, maxHeight int64) (FullCommit, error)
...
```
Given this what about using these structures in your specification (trying to keep the names you had before) if it makes sense. We can remove the vset hashes if we don't show any validations with them.
```go
type SignedHeader struct {
	height   int64   
	bfttime  time.Time
	VHash     []byte // vset hash for the current block
	nextVHash []byte // vset hash  for the next block 
        commit  Commit  // commit evidence for this height
}

type FullCommit struct {
	header  SignedHeader 
	V     ValidatorSet // vset for height
	NextV ValidatorSet // vset for height+1
}

// Commit returns a FullCommit
func Commit(height int64) (FullCommit, error)

```

docs/spec/consensus/light-client.md
ancazamfir: https://github.com/tendermint/tendermint/blob/master/types/validator_set.go#L591-L633

There are other checks outside this when we receive the `FullCommit` but I think it would be enough if we add a comment that for both sequential and non-sequential cases we do the same basic verification.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,315 @@
...
+        // headers in Cosmos are typically signed by all validators, so one
+        // may request that all signed (liveness might be endangered)
+    }
+
+    s := intersection(h1.Header.NextV,signers(h2.Commit))
+      // get validators in h1 that signed h2
+    vps := votingpower_in(s,h1.Header.NextV)
+      // sum of voting powers in h1 of
+      // validators that signed h2
+    return (vps > max(1/3,trustlevel) * vp_all)
```
ancazamfir: I feel like we are missing some explanations along the lines of the old ADR.

In theory the lite client should do:
1) get vps, the voting power of the intersection of V1 and V2 in V1
2) substract 1/3rd of total power of V1 (worst case where close to the 1/3rd of potentially faulty validators in V1 are part of the intersection) from vps, be that vps'
3) check vps' > 2/3rd of total power of V2 (guaranteed correct validators in V2 have more than 2/3rds)

The check in 3) fails most of the time, even for small validator set changes. Because of this we have relaxed the check in 3). 
Along these lines, I think trustlevel should be against V2 total power:
```suggestion
    return (vps - 1/3 * vp_all_v1 > trustlevel *vp_all_v2)
```

docs/spec/consensus/light-client.md
josef-widder: Thanks!

docs/spec/consensus/light-client.md
josef-widder: There is a difference to the intersection mechanism in the old version. We do not do intersection and check for two thirds. We check whether at least one *correct* validator from header h1 signed h2. That a correct validator exists, we check by requiring more than 1/3 of the old validators signing h2.

So we rely on the fact that a correct validator only signs "correct" blocks/headers, and that header h2 is well-formed, that is, all the checks are successful.

ancazamfir: Yes, I am just saying that we should add the justification for the difference.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,316 @@
...
+- as *v* is correct up to now, it followed the Tendermint consensus protocol at least up to signing *h2* => *h2* was correctly generated, we arrive at the required contradiction.
+
+
+Towards Lite Client Completeness:
+- The check is successful if sufficiently many validators of *h1* are still validators in *h2* and signed *h2*.
+- If *h2.Header.height = h1.Header.height + 1*, and both headers were generated correctly, the test passes
+
+*Verification Condition:* We may need a Tendermint invariant stating that if *h2.Header.height = h1.Header.height + 1* then *signers(h2.Commit) \subseteq h1.Header.NextV*.
+
+*Remark*: The variable *trustlevel* can be used if we believe that relying on one correct validator is not sufficient. However, in case of (frequent) changes in the validator set, the higher the *trustlevel* is chosen, the more unlikely it becomes that CheckSupport returns true for non-adjacent headers.
```
melekes: > f we believe that relying on one correct validator is not sufficient

who are we here? if relying on one correct validator is enough from security standpoint, then we don't need `trustlevel`

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,316 @@
...
-}
-```
+A lite client is a process that connects to Tendermint full nodes and then tries to verify application data using the Merkle proofs.
+
+## Context of this document
+
+In order to make sure that full nodes have the incentive to follow the protocol, we have to address the following three Issues
+
+1) The lite clients needs a method to verify header it obtains from full nodes -- addressed in this document)
+2) Detect conflicts/forged headers -- when the verification method of this document fails, then a conflict is detected. However, this document does not address to do this in an efficient way, or with special attacks in mind. This will be treated in fork accountability.
```
ebuchman: 2) The lite client must be able to connect to one correct full node to detect and report on failures in the trust assumptions (ie. conflicting headers) -- a future document.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,316 @@
...
+3) How can evidence for conflicts be reported, and to whom. This will be treated in a future specification.
```
ebuchman:
```suggestion
3) In the event the trust assumption fails (ie. a lite client is fooled by a conflicting header), the Tendermint fork accountability protocol must account for the evidence -- see #3840 
```

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,316 @@
...
+      // Plus the following check that 2/3 of the voting power
+      // in h1 signed h2
+      return (votingpower_in(signers(h2.Commit),h1.Header.NextV) >
+              max(2/3,trustlevel_adj) * vp_all)
+        // signing validators are more than two third in h1.
+        // headers in Cosmos are typically signed by all validators, so one
+        // may request that all signed (liveness might be endangered)
+    }
+
+    s := intersection(h1.Header.NextV,signers(h2.Commit))
```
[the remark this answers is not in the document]
josef-widder: yes. I will change that.

docs/spec/consensus/light-client.md
josef-widder: I incorporated the suggested text. I am not sure whether the addition was intended that way. Please check!

docs/spec/consensus/light-client.md
josef-widder: I tried to clarify that "we" is the user in the new revision...

melekes: > Does that make sense?

Makes sense, thanks!

docs/spec/consensus/light-client.md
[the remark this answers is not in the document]
josef-widder: In the Tendermint Core Dev meeting on Aug 19, we decided to stay with the current level of detail regarding the data structures.

docs/spec/consensus/light-client.md
josef-widder: OK. I added a comment on the old method and why we have chosen the new one.

docs/spec/consensus/light-client.md
```diff
@@ -1,113 +1,323 @@
...
+- again a contradiction
+
+Lite Client Completeness:
+
+This is only ensured if upon *Commit(pivot)* the lite client is always provided with a correctly generated header.
+
+*Stalling*
+
+With Bisection, a faulty full node could stall a lite client by creating a long sequence of headers that are queried one-by-one by the lite client and look OK, before the lite client eventually detects a problem. There are two ways to address this:
+* Each call to ```Commit``` could be issued to a different full node
```
melekes: as an alternative: can we set a time limit for how long the bisection can take?

josef-widder: I added timeout as another way of how we can address the problem.

PR #3846, opened 2019-07-30
 node: allow replacing existing p2p.Reactor(s) 

melekes:
using [`CustomReactors` option](https://godoc.org/github.com/tendermint/tendermint/node#CustomReactors).

Warning: beware of accidental name clashes. Here is the list of existing
reactors: MEMPOOL, BLOCKCHAIN, CONSENSUS, EVIDENCE, PEX.

* [x] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -17,6 +17,10 @@ program](https://hackerone.com/tendermint).
   - [libs] \#3811 Remove `db` from libs in favor of `https://github.com/tendermint/tm-cmn`
 
 ### FEATURES:
+- [node] Allow replacing existing p2p.Reactor(s) using [`CustomReactors`
+  option](https://godoc.org/github.com/tendermint/tendermint/node#CustomReactors).
+  Warning: beware of accidental name clashes. Here is the list of existing
+  reactors: MEMPOOL, BLOCKCHAIN, CONSENSUS, EVIDENCE, PEX.
 
 ### IMPROVEMENTS:
 
--- a/node/node.go
+++ b/node/node.go
@@ -48,10 +48,6 @@ import (
 	dbm "github.com/tendermint/tm-cmn/db"
 )
 
-// CustomReactorNamePrefix is a prefix for all custom reactors to prevent
-// clashes with built-in reactors.
-const CustomReactorNamePrefix = "CUSTOM_"
-
 //------------------------------------------------------------------------------
 
 // DBContext specifies config information for loading a new DB.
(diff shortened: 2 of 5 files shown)


node/node_test.go
```diff
@@ -309,6 +309,36 @@ func TestNodeNewNodeCustomReactors(t *testing.T) {
 	defer n.Stop()
 
 	assert.True(t, cr.IsRunning())
+	assert.Equal(t, cr, n.Switch().Reactor("FOO"))
+}
+
+func TestNodeNewNodeCustomReactorsReplacingExistingOne(t *testing.T) {
```
golangcibot: 315-342 lines are duplicate of `node/node_test.go:286-313` (from `dupl`)

node/node.go
```diff
@@ -144,11 +140,26 @@ func DefaultMetricsProvider(config *cfg.InstrumentationConfig) MetricsProvider {
...
+//
+// WARNING: using any name from the below list of the existing reactors will
+// result in replacing it with the custom one.
+//
+//  - MEMPOOL
+//  - BLOCKCHAIN
+//  - CONSENSUS
+//  - EVIDENCE
+//  - PEX
 func CustomReactors(reactors map[string]p2p.Reactor) Option {
```
brapse: Does it make sense to call this `SetCustomReactors`?

node/node.go
```diff
@@ -144,11 +140,26 @@ func DefaultMetricsProvider(config *cfg.InstrumentationConfig) MetricsProvider {
...
 	return func(n *Node) {
 		for name, reactor := range reactors {
-			n.sw.AddReactor(CustomReactorNamePrefix+name, reactor)
+			if existingReactor := n.sw.Reactor(name); existingReactor != nil {
+				n.sw.Logger.Info("Replacing existing reactor with a custom one",
```
brapse: nit, do we typically use `Infof` or is it preferable to concatenate strings?

melekes: We're currently using structured logging with https://github.com/go-kit/kit/tree/master/log under the hood. There're no such methods as `Infof`, `Errorf`, ...

node/node.go
melekes: I don't think `Set` prefix is needed 

This whole approach is largely based on Dave's article https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis. He's not using Set prefixes either.

CHANGELOG_PENDING.md
```diff
@@ -17,6 +17,10 @@ program](https://hackerone.com/tendermint).
   - [libs] \#3811 Remove `db` from libs in favor of `https://github.com/tendermint/tm-cmn`
 
 ### FEATURES:
+- [node] Allow replacing existing p2p.Reactor(s) using [`CustomReactors`
```
marbar3778: 👍

Issue #3211, opened 2019-01-25
tx indexer: can lose data if crash after block committed before indexing finishes


ebuchman:
Indexing of the txs in a block happens asynchronously so as to not block the next heights consensus. But if the block is fully committed and the process crashes before the indexing is complete, on restart, the old block will not be replayed, and so some txs will not have been properly indexed.

During the replay sequence on a restart, we must check to ensure that all txs from the previous block were indexed, and if not, re-index them. 

Related to https://github.com/tendermint/tendermint/issues/3186, but that's talking about a block that didn't even finish committing. In that case, the block does get replayed, we just didn't start the eventBus ahead of time. Here, we need to check explicitly if txs didnt finish indexing.

Stumble:
One question, is there setting or any ongoing proposal that can make tx_indexer runs synchronously?

marbar3778:
we recently implemented a re-indexing feature. Please reopen if you see this issue as still relevant.

Issue #3823, opened 2019-07-22
index service will lose data when node restart or start after being killed


guagualvcha:
As index service is running asynchronously, if block have been applied but tx index haven't write into db, if node crash at this point, we will lose index data. Any idea or plan to resolve this?

melekes:
Yeah, we need to store last indexed height and request missing blocks (`latestSavedHeight <= LatestBlockHeight`) upon startup.

guagualvcha:
> latestSavedHeight

great news. we will follow up if it comes soon.

ebuchman:
Is this a duplicate of https://github.com/tendermint/tendermint/issues/3211 ?

melekes:
Yep. Closing this for #3211

PR #3847, opened 2019-07-30
support Index service recovery and indexHeight in Status api

guagualvcha:
### Description
resolve . https://github.com/tendermint/tendermint/issues/3823

1. Will recover index data from restart or crash
2. Add field--'indexHeight'  to status API.
3. Add indexHegiht metrics. 
### Rationale
This pr have passed though following end to end test cases:

1. Upgrade from old binary, no recover will do.  checked
2. If previous do not use index and now change to use the index, no recover will do. checked
3. If index data lost, it will recover. checked 
4. If data have indexed, but didn't write the height, redo index has no side effect. checked
5. RPC API has changed.   checked
6. Metrics have changed.  checked
7. write to statedb will not cause consensus issues. checked

## Why indexHub.
For our tendermint fork, we have block index service, for other fork, may extend other index service. But all these should register in indexHub.

### Example

api of 'Status ' changed:

 ```json
 {
 "jsonrpc": "2.0",
 "id": "",
"result": {
...
 "sync_info": {
   		"latest_block_hash": "F51538DA498299F4C57AC8162AAFA0254CE08286",
   		"latest_app_hash": "0000000000000000",
   		"latest_block_height": "18",
   		"latest_block_time": "2018-09-17T11:42:19.149920551Z",
   		"catching_up": false,
   		"index_height": "18"
   	},
...
}
```

### DIFF
--- a/state/index.go
+++ b/state/index.go
@@ -1,0 +1,177 @@
+package state
+
+import (
+	"sync"
+
+	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/types"
+	dbm "github.com/tendermint/tm-cmn/db"
+)
+
+const (
+	// Actually the max lag is 2, use 10 for tolerance.
+	MaxIndexLag = 10
+)
+
+var indexHeight = []byte("indexHeight")
+
+type IndexService interface {
+	SetOnIndex(callback func(int64))
+}
+
+type IndexHub struct {
+	cmn.BaseService
+	mtx sync.Mutex
+
+	stateHeight  int64
+	expectHeight int64
+
+	// the total registered index service
+	numIdxSvc        int
+	indexTaskCounter map[int64]int
+	indexTaskEvents  chan int64
+
+	stateDB    dbm.DB
+	blockStore BlockStore
+	eventBus   types.BlockEventPublisher
+
+	metrics *Metrics
+}
+
+func NewIndexHub(initialHeight int64, stateDB dbm.DB, blockStore BlockStore, eventBus types.BlockEventPublisher, options ...IndexHubOption) *IndexHub {
+	ih := &IndexHub{
+		stateHeight:      initialHeight,
+		indexTaskCounter: make(map[int64]int),
+		indexTaskEvents:  make(chan int64, MaxIndexLag),
+		stateDB:          stateDB,
+		blockStore:       blockStore,
+		eventBus:         eventBus,
+		metrics:          NopMetrics(),
+	}
+	indexedHeight := ih.GetIndexedHeight()
+	if indexedHeight < 0 {
+		// no indexedHeight found, will do no recover
+		ih.expectHeight = ih.stateHeight + 1
+	} else {
+		ih.expectHeight = indexedHeight + 1
+	}
+	for _, option := range options {
+		option(ih)
+	}
+	ih.BaseService = *cmn.NewBaseService(nil, "indexHub", ih)
+	return ih
+}
+
+type IndexHubOption func(*IndexHub)
+
+func IndexHubWithMetrics(metrics *Metrics) IndexHubOption {
+	return func(ih *IndexHub) {
+		ih.metrics = metrics
+	}
+}
+
+func (ih *IndexHub) OnStart() error {
+	// start listen routine before recovering.
+	go ih.indexRoutine()
+	ih.recoverIndex()
+	return nil
+}
+
+func (ih *IndexHub) recoverIndex() {
+	for h := ih.expectHeight; h <= ih.stateHeight; h++ {
+		ih.Logger.Info("try to recover index", "height", h)
+		block := ih.blockStore.LoadBlock(h)
+		if block == nil {
+			ih.Logger.Error("index skip since the the block is missing", "height", h)
+		} else {
+			abciResponses, err := LoadABCIResponses(ih.stateDB, h)
+			if err != nil {
+				ih.Logger.Error("failed to load ABCIResponse, will use default")
+				abciResponses = NewABCIResponses(block)
+			}
+			abciValUpdates := abciResponses.EndBlock.ValidatorUpdates
+			validatorUpdates, err := types.PB2TM.ValidatorUpdates(abciValUpdates)
+			if err != nil {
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: state/index_test.go [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+			Header: types.Header{Height: h, NumTxs: numTxs},
+		})
+		for i := int64(0); i < numTxs; i++ {
+			txResult := &types.TxResult{
+				Height: h,
+				Index:  uint32(i),
+				Tx:     types.Tx("foo"),
+				Result: abci.ResponseDeliverTx{Code: 0},
+			}
+			eventBus.PublishEventTx(types.EventDataTx{*txResult})
```
golangcibot: composites: `github.com/tendermint/tendermint/types.EventDataTx` composite literal uses unkeyed fields (from `govet`)

state/index_test.go
```diff
@@ -0,0 +1,78 @@
+package state
+
+import (
+	"github.com/stretchr/testify/assert"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

state/index.go
```diff
@@ -0,0 +1,174 @@
...
+		if block == nil {
+			ih.Logger.Error("index skip since the the block is missing", "height", h)
+		} else {
+			abciResponses, err := LoadABCIResponses(ih.stateDB, h)
+			if err != nil {
+				ih.Logger.Error("failed to load ABCIResponse, will use default")
+				abciResponses = NewABCIResponses(block)
+			}
+			abciValUpdates := abciResponses.EndBlock.ValidatorUpdates
+			validatorUpdates, err := types.PB2TM.ValidatorUpdates(abciValUpdates)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

state/index.go
```diff
@@ -0,0 +1,174 @@
...
+	blockStore BlockStore
+	eventBus   types.BlockEventPublisher
+
+	metrics *Metrics
+}
+
+func NewIndexHub(initialHeight int64, stateDB dbm.DB, blockStore BlockStore, eventBus types.BlockEventPublisher, options ...IndexHubOption) *IndexHub {
+	ih := &IndexHub{
+		stateHeight:      initialHeight,
+		indexTaskCounter: make(map[int64]int, 0),
```
golangcibot: S1019: should use make(map[int64]int) instead (from `gosimple`)

state/index.go
```diff
@@ -0,0 +1,177 @@
...
+			ih.Logger.Error("index skip since the the block is missing", "height", h)
+		} else {
+			abciResponses, err := LoadABCIResponses(ih.stateDB, h)
+			if err != nil {
+				ih.Logger.Error("failed to load ABCIResponse, will use default")
+				abciResponses = NewABCIResponses(block)
+			}
+			abciValUpdates := abciResponses.EndBlock.ValidatorUpdates
+			validatorUpdates, err := types.PB2TM.ValidatorUpdates(abciValUpdates)
+			if err!=nil{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
			if err != nil {
```

guagualvcha:
@melekes  Hi, does CI failed because of test case coverage?

state/index.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func (ih *IndexHub) recoverIndex() {
+	for h := ih.expectHeight; h <= ih.stateHeight; h++ {
+		ih.Logger.Info("try to recover index", "height", h)
+		block := ih.blockStore.LoadBlock(h)
+		if block == nil {
+			ih.Logger.Error("index skip since the the block is missing", "height", h)
+		} else {
+			abciResponses, err := LoadABCIResponses(ih.stateDB, h)
```
melekes: Don't we already do that as a part of replay process?

state/index.go
```diff
@@ -0,0 +1,177 @@
...
+			if err != nil {
+				ih.Logger.Error("failed to load ABCIResponse, will use default")
+				abciResponses = NewABCIResponses(block)
+			}
+			abciValUpdates := abciResponses.EndBlock.ValidatorUpdates
+			validatorUpdates, err := types.PB2TM.ValidatorUpdates(abciValUpdates)
+			if err != nil {
+				ih.Logger.Error("failed to load validatorUpdates, will use nil by default")
+			}
+			fireEvents(ih.Logger, ih.eventBus, block, abciResponses, validatorUpdates)
```
melekes: This might shoot us in the leg one day. I think this function is only expected to be called from state/execution.

guagualvcha:
> Thank you for your work 👍 This does solve the task at hand. However, it seems a bit more complex than it has to be. I feel like we could simplify this a lot by
> 
> 1. getting `LastIndexedHeight()` from `TxIndexer`
> 2. in node.go, call ^, check if there's a lag. If there is, fetch the missing blocks from the blockstore and index them by calling `IndexerService#Index(blocks)`
> 
> As for `last_indexed_height` metric, is it really useful for validators? I mean if the indexer works correctly, there should be no need for this metric.
> 
> Thank you again! Really appreciate all the work you are doing 🙇

Because we add block index service in our fork(support query block by block hash), so there are about two index services, that is why  I choose a more common solution.  `last_indexed_height` is just for test, I can remove it, according to my test, the most lags only be 2, because index service subscribe event from event hub in unbuffer way.

state/index.go
guagualvcha: this part is before replay. replay process will only replay the saved block but still do not play the state, but index replay will refire the event of played block but unindexed block

state/index.go
guagualvcha: any suggestions?

guagualvcha:
Becasue tendermint only have tx index now,  there do have some easy way to simplify the solution, if that is want you expect, I will do in that way. @melekes

marbar3778:
believe this also closes #3211 
I dont see last_index_height as a necessary metric, can we remove it

melekes:
Closing due to https://github.com/tendermint/tendermint/pull/3847#pullrequestreview-272417749. If you're still willing to work on this @guagualvcha, please notify us and we will happily reopen the PR.

Issue #3076, opened 2019-01-01
[rpc] tx_search can take a long time


gianfelipe93:
Hi guys,

I am running Tendermint with 7 nodes on AWS using Kubernetes. Currently, I have 36.000 transactions in the blockchain but it is taking quite a long time to query transactions (around 10 seconds to query 100 transactions).

Is there any sort of production config that I can use that will make it faster?

Thank for your help guys.

Have a great 2019!

melekes:
What are you using to query transactions?

gianfelipe93:
@melekes I am using `TxQuery` like below:
```
resQuery, err := rpcClient.TxSearch(query, false, page, txPerPage)
```

is it the proper way to query transactions ?

melekes:
> is it the proper way to query transactions ?

- for individual transactions - use `/tx`
- for multiple transactions sharing a common tag - use `/tx_search`
- for complete blocks - use `/block` or /blockchain`

Could you also post a query string here? Thanks

gianfelipe93:
@melekes thanks for you help!

can I query by a tag using `/tx` or `/block`?

I am not sure if they are suitable for my use case:
A user uploads 500 transactions per day (average) and the user is able to see its own transactions:
`tx.UserID='abc'`

the user can also query by time:
`tx.UserID='abc AND tx.Timestamp>=*<TIMESTAMP>* AND tx.Timestamp<=*<TIMESTAMP>* `

And also by other tags but the that is pretty much the query structure.

I will always query multiple transactions so I think we can discart the `/tx` route and since the transactions might be in different block probably the `/block` route doesn't suit me.

is there anything you suggest I could do to make the query faster ? 

Additionally, since we are deploying on AWS, do you think a more powerful VM would speed up the query ?

Thanks for you help!

melekes:
Oh, I think I know how we can make it faster.

gianfelipe93:
@melekes how ?

melekes:
Right now we're fetching the records from the DB and perform an intersection. We should instead pass a first selection around which should shrink with a number of filters.

gianfelipe93:
@melekes is there something I can help you with ?

gianfelipe93:
This is becoming a real issue because it's taking more than 2 minutes to query transactions.

melekes:
[...]

PR #3851, opened 2019-07-30
Refactor Tx Search Aggregation

alexanderbez:
- Replace the previous `intersect` call, which was called at each query condition, with a map intersection.
- Replace `fmt.Sprintf` with `string()`

closes: #3076

----

### Old

```
goos: darwin
goarch: amd64
pkg: github.com/tendermint/tendermint/state/txindex/kv
BenchmarkTxSearch-4   	     200	 103641206 ns/op	 7998416 B/op	   71171 allocs/op
PASS
ok  	github.com/tendermint/tendermint/state/txindex/kv	26.019s
```

### New

```
goos: darwin
goarch: amd64
pkg: github.com/tendermint/tendermint/state/txindex/kv
BenchmarkTxSearch-4   	    1000	  38615024 ns/op	13515226 B/op	  166460 allocs/op
PASS
ok  	github.com/tendermint/tendermint/state/txindex/kv	53.618s
```

**~62% performance improvement**

### New

pprof002

At this point, it seems the only bottleneck is the DB itself.

----

* [x] Wrote tests


state/txindex/kv/kv.go
```diff
@@ -353,63 +365,115 @@ func isRangeOperation(op query.Operator) bool {
 	}
 }
 
-func (txi *TxIndex) match(c query.Condition, startKeyBz []byte) (hashes [][]byte) {
+// match returns all matching txs by hash that meet a given condition and start
+// key. An already filtered result (filteredHashes) is provided such that any
+// non-intersecting matches are removed.
+//
+// NOTE: filteredHashes may be empty if no previous condition has matched.
+func (txi *TxIndex) match(c query.Condition, startKeyBz []byte, filteredHashes map[string][]byte, init bool) map[string][]byte {
```
melekes:
```suggestion
func (txi *TxIndex) match(c query.Condition, startKeyBz []byte, filteredHashes map[string][]byte, firstRun bool) map[string][]byte {
```

state/txindex/kv/kv.go
```diff
@@ -353,63 +365,115 @@ func isRangeOperation(op query.Operator) bool {
...
-				hashes = append(hashes, it.Value())
+				tmpHashes[string(it.Value())] = it.Value()
 			}
 		}
 	} else {
 		panic("other operators should be handled already")
 	}
-	return
+
+	if len(tmpHashes) == 0 || (init && len(filteredHashes) == 0) {
```
melekes: do we really need to check if `len(filteredHashes) == 0` in the second condition?

state/txindex/kv/kv_bench_test.go
```diff
@@ -0,0 +1,70 @@
...
+				Type: "transfer",
+				Attributes: []cmn.KVPair{
+					{Key: []byte("address"), Value: []byte(fmt.Sprintf("address_%d", i%100))},
+					{Key: []byte("amount"), Value: []byte("50")},
+				},
+			},
+		}
+
+		txBz := make([]byte, 8)
+		_, _ = rand.Read(txBz)
```
melekes: why do we need _, _?

alexanderbez: Just to please any IDE. We can remove it if so desired.

state/txindex/kv/kv.go
alexanderbez: Mhmm, no I don't believe so!

state/txindex/kv/kv.go
```diff
@@ -353,63 +365,115 @@ func isRangeOperation(op query.Operator) bool {
...
-func (txi *TxIndex) match(c query.Condition, startKeyBz []byte) (hashes [][]byte) {
+// match returns all matching txs by hash that meet a given condition and start
+// key. An already filtered result (filteredHashes) is provided such that any
+// non-intersecting matches are removed.
+//
+// NOTE: filteredHashes may be empty if no previous condition has matched.
+func (txi *TxIndex) match(c query.Condition, startKeyBz []byte, filteredHashes map[string][]byte, firstRun bool) map[string][]byte {
+	// A previous match was attempted but resulted in no matches, so we return
+	// no matches (assuming AND operand).
+	if !init && len(filteredHashes) == 0 {
```
golangcibot: undeclared name: `init` (from `typecheck`)

state/txindex/kv/kv_bench_test.go
alexanderbez: Updated.

PR #3856, opened 2019-07-31
 Remove duplication of data in `ResultBlock `

### LINKED ISSUE #3188: rpc: remove header from block_meta in /block method response and add size in bytes
I am using the tendermint, I notice that  blockmeta and block both are generating same structure result, please let me know the difference between it.

marbar3778:
- Removed `BlockMeta` in `ResultBlock` in favor of `BlockId` for `/block`
- Added block_size to `BlockMeta` this is reflected in `/blockchain`

fixes #3188

added breaking as some clients may be using header from blockmeta instead of block in `/block`

* [ ] Wrote tests


rpc/client/rpc_test.go
```diff
@@ -226,9 +226,9 @@ func TestAppCalls(t *testing.T) {
 		if assert.Equal(1, len(info.BlockMetas)) {
 			lastMeta := info.BlockMetas[0]
 			assert.EqualValues(apph, lastMeta.Header.Height)
-			bMeta := block.BlockMeta
-			assert.Equal(bMeta.Header.AppHash, lastMeta.Header.AppHash)
-			assert.Equal(bMeta.BlockID, lastMeta.BlockID)
+			blockData := block.Block
+			assert.Equal(blockData.Header.AppHash, lastMeta.Header.AppHash)
+			assert.Equal(*block.BlockID, lastMeta.BlockID)
```
golangcibot: invalid operation: cannot indirect block.BlockID (variable of type types.BlockID) (from `typecheck`)

golangcibot: invalid operation: cannot indirect block.BlockID (variable of type types.BlockID) (from `typecheck`)

golangcibot: invalid operation: cannot indirect block.BlockID (variable of type types.BlockID) (from `typecheck`)

rpc/core/blocks.go
```diff
@@ -58,7 +58,8 @@ import (
 // 						"total": "1"
 // 					},
 // 					"hash": "96B1D2F2D201BA4BC383EB8224139DB1294944E5"
-// 				}
+// 				},
+// 				"block_size": "10000"
```
marbar3778: note to self: this is in the wrong place

types/block_meta.go
```diff
@@ -1,16 +1,18 @@
 package types
 
-// BlockMeta contains meta information about a block - namely, it's ID and Header.
+// BlockMeta contains meta information about a block - ID, Size, and Header.
 type BlockMeta struct {
-	BlockID BlockID `json:"block_id"` // the block hash and partsethash
-	Header  Header  `json:"header"`   // The block's Header
+	BlockID   BlockID `json:"block_id"`   // the block hash and partsethash
+	BlockSize int     `json:"block_size"` // The size of the block
```
melekes: I wonder if we should make this `uint32` or simply `uint`

melekes:
swagger.yml needs to be updated as well

types/block_meta.go
marbar3778: I am for this change, there isn't a case where this should ever be -, uint seems good enough

melekes:
contract tests failed:

```
error: API description parser error in /home/circleci/.go_workspace/src/github.com/tendermint/tendermint/docs/spec/rpc/swagger.yaml: Error resolving $ref pointer "#/definitions/BlockId". 
Token "BlockId" does not exist.
```

types/block.go
```diff
@@ -202,12 +202,12 @@ func (b *Block) HashesTo(hash []byte) bool {
 }
 
 // Size returns size of the block in bytes.
-func (b *Block) Size() int {
+func (b *Block) Size() uint {
```
melekes: Most of the Go "Size" stdlib functions return `int`

types/block_meta.go
```diff
@@ -2,17 +2,19 @@ package types
 
 // BlockMeta contains meta information.
 type BlockMeta struct {
-	BlockID BlockID `json:"block_id"`
-	Header  Header  `json:"header"`
-	NumTxs  int64   `json:"number_txs"`
+	BlockID   BlockID `json:"block_id"`
+	BlockSize int     `json:"block_size"`
```
ebuchman: Do we care about the ordering here re breaking? Ie could we consider this "less breaking" if we just add fields to the end instead of in the middle?

melekes: In the presence of other protocol breaking changes? Probably no

Issue #2949, opened 2018-11-29
Responses to "subscribe" sent over WebSocket don't conform to RPC 2.0 specification


clymbon:
**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):

0.25.0-0c9c3292

**ABCI app** (name for built-in, URL for self-written if it's publicly available):

kvstore

**Environment**:
- **OS** (e.g. from /etc/os-release):

NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"

- **Install tools**:
- **Others**:

**What happened**:

The responses to "subscribe" over websocket are not as expected.

I believe that they are not in conformance with the RPC 2.0 spec.  (I could be wrong about that.)

When the following "subscribe" message is sent over the websocket:

{"jsonrpc":"2.0","method":"subscribe","id":"1","params":["tm.event = 'NewBlock'"]} I get this response:

I get the following response:

{
  "jsonrpc": "2.0",
  "id": "1",
  "result": {}
}

This is as expected, indicating that the message was received and processed without error.  In the above, the "id" value is 1, indicating this is the first request sent.  (I send it immediately after connecting the socket.)

Subsequently, whenever a new block is created, I get a message like this:

{"jsonrpc":"2.0","id":"1#event","result":{"query":"tm.event = 'NewBlock'","data":{"type":"tendermint/event/NewBlock","value":{"block":  ..... (block contents deleted)

I believe this is NOT in accordance with the JSON RPC 2.0 spec.  The id value is set to "1#event', which chokes the javascript RPC library I'm using.  (https://github.com/jershell/simple-jsonrpc-js)

I think I get the idea behind this - sending the events responding to the subscription request with an id value that can be correlated back to the original request.  But I don't think this is conformant.  The spec says the ID value has to be identical to the request, and that a notification should not contain an ID value.

I was able to work around this by hacking the siimple-jsonrpc-js code to look for responses with id's ending in #event, and convert them into notifications.

Maybe this is how RPC 2.0 is supposed to work?  If so, I'm not seeing that in the spec, and the library I'm using doesn't recognize it.

I do see a related bug request #1099 

**What you expected to happen**:

The events resulting from a subscription request should be sent as NOTIFICATIONs.

**Have you tried the latest version**: yes/no

[...]

PR #3858, opened 2019-07-31
rpc/lib/client & server: try to conform to JSON-RPC 2.0 spec

melekes:
https://www.jsonrpc.org/specification

What is done in this PR:

- **JSONRPCClient**: validate that Response.ID matches Request.ID
   I wanted to do the same for the WSClient, but since we're sending events as responses, not notifications, checking IDs would require storing them in memory indefinitely (and we won't be able to remove them upon client unsubscribing because ID is different then).

- Request.ID is now optional. Notification is a Request without an ID. Previously "" or 0 were considered as notifications

- Remove `#event` suffix from ID from an event response (partially fixes https://github.com/tendermint/tendermint/issues/2949) ID must be either string, int or null AND must be equal to request's ID. Now, because we've implemented events as responses, WS clients are tripping when they see Response.ID("0#event") != Request.ID("0"). Implementing events as requests would require a lot of time (~ 2 days to completely rewrite WS client and server)

- generate unique ID for each request

- switch to integer IDs instead of "json-client-XYZ"
   id=0 method=/subscribe
   id=0 result=...
   id=1 method=/abci_query
   id=1 result=...

TODO:
- ~~[ ] send events (resulting from `/subscribe`) as notifications (not responses) https://github.com/ethereum/go-ethereum/wiki/RPC-PUB-SUB https://github.com/tendermint/tendermint/blob/2f32b9f65e26b29c1cfc7697fd06d742b1243d50/rpc/core/events.go#L180-L185 
From spec: "A Notification is a Request object"~~
- [ ] check tm-bench works

---

* [ ] Wrote tests


rpc/lib/types/types.go
```diff
@@ -22,12 +22,14 @@ type jsonrpcid interface {
 // JSONRPCStringID a wrapper for JSON-RPC string IDs
 type JSONRPCStringID string
 
-func (JSONRPCStringID) isJSONRPCID() {}
+func (JSONRPCStringID) isJSONRPCID()      {}
+func (id JSONRPCStringID) String() string { return fmt.Sprintf("%s", id) }
```
golangcibot: printf: Sprintf format %s with arg id causes recursive String method call (from `govet`)

rpc/lib/client/ws_client.go
```diff
@@ -86,6 +91,7 @@ func NewWSClient(remoteAddr, endpoint string, options ...func(*WSClient)) *WSCli
 	}
 
 	c := &WSClient{
+		sentIds:              make(map[types.JSONRPCIntID]bool, 0),
```
golangcibot: S1019: should use make(map[types.JSONRPCIntID]bool) instead (from `gosimple`)

rpc/lib/server/ws_handler_test.go
```diff
@@ -0,0 +1,53 @@
...
+	types "github.com/tendermint/tendermint/rpc/lib/types"
+)
+
+func TestWebsocketManagerHandler(t *testing.T) {
+	s := newWSServer()
+	defer s.Close()
+
+	// check upgrader works
+	d := websocket.Dialer{}
+	c, dialResp, err := d.Dial("ws://"+s.Listener.Addr().String()+"/websocket", nil)
```
golangcibot: bodyclose: response body must be closed (from `bodyclose`)

melekes:
> send events (resulting from /subscribe) as requests+notifications (not responses) 

this will require a lot of work. probably not worth it

rpc/lib/client/decode.go
```diff
@@ -0,0 +1,122 @@
...
+			return errors.Errorf("unsolicited response ID #%d: %v", i, id)
+		}
+	}
+
+	return nil
+}
+
+// From the JSON-RPC 2.0 spec:
+//  id: It MUST be the same as the value of the id member in the Request Object.
+func assertResponseIDEqual(res *types.RPCResponse, expectedID types.JSONRPCIntID) error {
```
melekes: revert

rpc/lib/client/ws_client.go
```diff
@@ -466,7 +480,21 @@ func (c *WSClient) readRoutine() {
...
+			c.Logger.Error("error in response ID", "id", response.ID, "err", err)
+			continue
+		}
+
+		c.mtx.RLock()
+		if _, ok := c.sentIDs[response.ID.(types.JSONRPCIntID)]; !ok {
+			c.Logger.Error("unsolicited response ID", "id", response.ID, "expected", c.sentIDs)
+			c.mtx.RUnlock()
+			continue
+		}
```
melekes:
```suggestion
		}
		delete(c.sentIDs, response.ID.(types.JSONRPCIntID))
```

rpc/lib/server/http_json_handler.go
```diff
@@ -0,0 +1,215 @@
...
+
+// jsonrpc calls grab the given method's function info and runs reflect.Call
+func makeJSONRPCHandler(funcMap map[string]*RPCFunc, cdc *amino.Codec, logger log.Logger) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		b, err := ioutil.ReadAll(r.Body)
+		if err != nil {
+			WriteRPCResponseHTTP(w, types.RPCInvalidRequestError(nil, errors.Wrap(err, "error reading request body")))
+			return
+		}
+		defer r.Body.Close() // nolint: errcheck
```
melekes:
```suggestion
		r.Body.Close() // nolint: errcheck
```

rpc/lib/server/ws_handler.go
```diff
@@ -0,0 +1,423 @@
...
+
+// NewWebsocketManager returns a new WebsocketManager that passes a map of
+// functions, connection options and logger to new WS connections.
+func NewWebsocketManager(funcMap map[string]*RPCFunc, cdc *amino.Codec, wsConnOptions ...func(*wsConnection)) *WebsocketManager {
+	return &WebsocketManager{
+		funcMap: funcMap,
+		cdc:     cdc,
+		Upgrader: websocket.Upgrader{
+			CheckOrigin: func(r *http.Request) bool {
+				// TODO ???
```
melekes: ???

thanethomson: The [default behaviour](https://github.com/gorilla/websocket/blob/483fb8d7c32fcb4b5636cd293a92e3935932e2f4/server.go#L87) would be relevant to browser-based clients, afaik. I suppose having a pass-through is a workaround for allowing for more complex security schemes, shifting the burden of AuthN/AuthZ outside the Tendermint RPC.

I can't think of other uses right now that would warrant a `TODO` though. The real backstory of this `TODO` shall remain shrouded in mystery 👻

tools/tm-bench/transacter.go
```diff
@@ -194,7 +194,7 @@ func (t *transacter) sendLoop(connIndex int) {
 				c.SetWriteDeadline(now.Add(sendTimeout))
 				err = c.WriteJSON(rpctypes.RPCRequest{
 					JSONRPC: "2.0",
-					ID:      rpctypes.JSONRPCStringID("tm-bench"),
+					ID:      rpctypes.JSONRPCIntID((connIndex * 10000) + i),
```
thanethomson: Given the current system, 10k here seems to allow for good "padding" around tx IDs, but one could just as easily use `t.Rate` instead of 10k, right?

rpc/lib/server/ws_handler.go
```diff
@@ -0,0 +1,423 @@
...
+
+// WebsocketHandler upgrades the request/response (via http.Hijack) and starts
+// the wsConnection.
+func (wm *WebsocketManager) WebsocketHandler(w http.ResponseWriter, r *http.Request) {
+	wsConn, err := wm.Upgrade(w, r, nil)
+	if err != nil {
+		// TODO - return http error
+		wm.logger.Error("Failed to upgrade to websocket connection", "err", err)
+		return
+	}
```
thanethomson: Is it not perhaps better to do a `defer wsConn.Close()` after this `if` statement than to try to manage the closing of the WebSockets connection from within `wsConnection`?

rpc/lib/server/ws_handler.go
```diff
@@ -0,0 +1,423 @@
...
+		if r := recover(); r != nil {
+			err, ok := r.(error)
+			if !ok {
+				err = fmt.Errorf("WSJSONRPC: %v", r)
+			}
+			wsc.Logger.Error("Panic in WSJSONRPC handler", "err", err, "stack", string(debug.Stack()))
+			wsc.WriteRPCResponse(types.RPCInternalError(request.ID, err))
+			go wsc.readRoutine()
+		} else {
+			wsc.baseConn.Close() // nolint: errcheck
```
thanethomson: As per other comment regarding closing connection

rpc/lib/server/ws_handler.go
```diff
@@ -0,0 +1,423 @@
...
+		}
+	}
+}
+
+// receives on a write channel and writes out on the socket
+func (wsc *wsConnection) writeRoutine() {
+	pingTicker := time.NewTicker(wsc.pingPeriod)
+	defer func() {
+		pingTicker.Stop()
+		if err := wsc.baseConn.Close(); err != nil {
```
thanethomson: As per other comment regarding closing connection

rpc/lib/server/ws_handler.go
melekes: Gold! Saved this comment for future generations 😄

marbar3778:
> > send events (resulting from /subscribe) as requests+notifications (not responses)
> 
> this will require a lot of work. probably not worth it

I do believe this is worth it, to bring us within the spec and what the wider community is using and used to. Do you think we could have it for 0.33? we can branch off from this branch then make the PR to be merged into this PR as it has already been reviewed?

melekes:
Impossible to merge. Will need to redo every step.

PR #3876, opened 2019-08-02
Add Timeout to `http.client`

### LINKED ISSUE #2010: rpc: allow setting a Timeout for client requests

marbar3778:
- added a configurable timeout to http.client
- fixes #2010

* [ ] Wrote tests

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -381,6 +381,7 @@ type RPCConfig struct {
 
 	// ClientTimeout specifies a time limit for requests made by the client.
 	// A ClientTimeout of zero means no timeout
+	// Define the timeout in seconds
 	ClientTimeout time.Duration `mapstrucutre:"client_timeout"`
 }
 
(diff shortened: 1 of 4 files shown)


melekes:
This should be done jointly with https://github.com/tendermint/tendermint/pull/3779.

Issue #2880, opened 2018-11-19
Improve Tendermint RPC docs (POST/websocket)


ethanfrey:
First of all, thank you for https://tendermint.com/rpc, it contains a very good source of information and defines the endpoints and return values quite well. One issue, however, is that all examples are in URI format, which works for quick curl tests, while most developers would prefer the json-rpc POST API, or better yet, the websocket endpoint that allows for subscriptions (which are a wonderful feature).

In order to make the docs more useful for people coding clients, I would suggest three improvements:

1. Alongside the `curl` examples, please show an example JSON-RPC POST body
2. Please specify if all byte fields are returned as hex or base64 (this seems more standard on hex these days, but I am unsure of eg. the broadcast_tx_commit data field, which was always base64 in the past)
3. Please provide documentation of valid subscription query types and expected response bodies. They are the hardest ones to produce example for, and very useful.

Thank you for this work, and I know the docs keep getting better, just want to push for a little more polish.

ethanfrey:
It seems that the docs are quite out-of-date as well.

https://tendermint.com/rpc/#abciinfo shows
```
    "result": {
        "response": {
            "log": "exists",
            "height": "0",
            "proof": "010114FED0DAD959F36091AD761C922ABA3CBF1D8349990101020103011406AA2262E2F448242DF2C2607C3CDC705313EE3B0001149D16177BC71E445476174622EA559715C293740C",
            "value": "61626364",
            "key": "61626364",
            "index": "-1",
            "code": "0"
        }
    }
```

When running against a local tendermint 0.25 kvstore, I get:
`curl 'localhost:12345/abci_query?path=""&data="abcd"&prove=false'`
```
  "result": {
    "response": {
      "log": "does not exist",
      "index": "-1",
      "key": "YWJjZA=="
    }
  }
```

key is base64, while example was hex

ethanfrey:
Move to swagger?
Then you remove all the custom rpc code with swagger autogen?

I have used swagger before for go apis and it can work, but will require a total overhaul of the rpc package. Is there an issue for that?

srmo:
Would love this. Or even thinking about jsonschema. 
Our use case would be, that we use such an API to autogenerate json mapping classes (java) when the protocol is updated. Versioning would also allow to easily recognize updated formats.

PR #3880, opened 2019-08-05
[RPC] Static swagger

sabau:
Describing statically the RPC implementation with swagger

Closes: #2880, #2419 

The initial goal was to introduce proper annotations:
- I tried `swaggo` but I hit a limitation in the usage of aliases https://github.com/sabau/tendermint/tree/2880-swagger-rpc-docs
- I tried `go-swagger` but they don't support go modules yet, even though `go-swagger` itself is using them https://github.com/sabau/tendermint/tree/2880-go-swagger-rpc

With both, workarounds to make it work are possible but may create a brittle solution that will soon need to be refactored once go modules are supported (or aliases in the case of `swaggo`)

The first part of the definitions is organized, the second half is left as it will look once auto-generated, so we can see the differences

* [ ] Wrote tests


safareli:
This is definition of ResponseInfo message from proto file but it's
https://github.com/tendermint/tendermint/blob/62f97a69e97262b5feb57b1c2498f0c1e0e297b3/abci/types/types.proto#L137-L145

But in swagger file you have:

https://github.com/tendermint/tendermint/blob/a9ae00b3f2a21eebb03daf0cd19e8851d94a5808/docs/spec/rpc/swagger.yaml#L2115-L2118

i.e only 3 fields are specified vs 5.

also general issue of having omitempty everywhere in generated types means non of the fileds of generated types are actually required #3882

sabau:
Thanks for cathing this! I will adapt the swagger as your comment.

marbar3778:
@mircea-c can we have this in place of https://tendermint.com/rpc/#introduction, when its merged?

sabau:
@mircea-c @marbar3778 To get this working we should deploy just once the index.html I added in the same folder of the swagger file (if it's an S3 bucket let's just be sure they are both reachable permission-wise) and every time we want to update the RPC docs, just replace `swagger.yaml` with the new version of it

mircea-c:
@marbar3778 @sabau you guys need to loop in the design team and check how the website is currently set up and what it would take to integrate this into it.

.circleci/config.yml
```diff
@@ -359,6 +359,26 @@ jobs:
...
+          at: /tmp/workspace
+      - restore_cache:
+          key: v4-pkg-cache
+      - restore_cache:
+          key: v3-tree-{{ .Environment.CIRCLE_SHA1 }}
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run:
+          name: Contract tests
```
melekes: Can we be more specific about what sort of contract?

melekes:
```suggestion
          name: Test Swagger
```

.circleci/config.yml
```diff
@@ -359,6 +359,26 @@ jobs:
...
+          command: |
+            go get github.com/snikch/goodman/cmd/goodman
+            export PATH=~/.local/bin:$PATH
+            ./scripts/get_nodejs.sh && make contract-tests
```
melekes: fetching node.js on every build seems like an overkill

melekes: maybe we can just use https://circleci.com/docs/2.0/language-javascript/ node.js container?

Makefile
```diff
@@ -308,7 +308,17 @@ sentry-stop:
...
 
+build-contract-tests-hooks:
+ifeq ($(OS),Windows_NT)
+	go build -mod=readonly $(BUILD_FLAGS) -o build/contract_tests.exe ./cmd/contract_tests
+else
+	go build -mod=readonly $(BUILD_FLAGS) -o build/contract_tests ./cmd/contract_tests
+endif
+
+contract-tests: build build-contract-tests-hooks
+	dredd
```
melekes: what's this? should we install it during `get_tools` https://github.com/tendermint/tendermint/blob/master/scripts/get_tools.sh ? if not, then there should be a link to it at least

melekes: also github.com/snikch/goodman needs to be installed via `get_tools` ^

.circleci/config.yml
sabau: It's basically a way to check the implementation against a document 'contract' instead of an integration test, usually more expensive
In theory, the documentation will describe the specs we want to follow, and we test the implementation against the spec.

Makefile
sabau: `dredd` is the tool that runs every endpoint he finds in the swagger file and check if the payload of the calls agree with the documented one

sabau: Right! I added only in the get_node.sh but I forgot to describe it here for local usage, will do it now

sabau: Just added some documentation under `CONTRIBUTING`

.circleci/config.yml
sabau: Here it's not too nice, but we have to use either our `go` container and add the `nodejs` binary ~13MB or the other way around, use the `node` container and add `go` to build the two binaries we use for this tool:
- tendermint
- hooks

melekes: right, so we can't use nodejs container because we need tendermint binary

CONTRIBUTING.md
```diff
@@ -183,3 +183,15 @@ If they have `.go` files in the root directory, they will be automatically
...
+
+If you contribute to the RPC endpoints it's important to document your changes in the [Swagger file](./docs/spec/rpc/swagger.yaml)
+To test your changes you should install `nodejs` version `v11.15.*` and run:
+
+```bash
+npm i -g dredd
+make contract-tests
+``` 
+
+This command will popup a network and check every endpoint against what has been documented
```
melekes: 👍 awesome, thanks

scripts/get_tools.sh
```diff
@@ -64,3 +64,4 @@ installFromGithub golangci/golangci-lint 7b2421d55194c9dc385eff7720a037aa9244ca3
 installFromGithub petermattis/goid b0b1615b78e5ee59739545bb38426383b2cda4c9
 installFromGithub sasha-s/go-deadlock d68e2bc52ae3291765881b9056f2c1527f245f1e
 go get golang.org/x/tools/cmd/goimports
+go get github.com/snikch/goodman/cmd/goodman
```
melekes:
```suggestion
installFromGithub snikch/goodman 10e37e294daa3c9a90abded60ff9924bafab3888 cmd/goodman
```

Makefile
```diff
@@ -315,7 +315,7 @@ else
 	go build -mod=readonly $(BUILD_FLAGS) -o build/contract_tests ./cmd/contract_tests
 endif
 
-contract-tests: build build-contract-tests-hooks
+contract-tests: build-linux build-contract-tests-hooks
```
melekes: why build-linux instead of just build?

sabau: I tried on my mac :P and it failed because the dockers where it runs need the Linux version of it. I've been working on my ubuntu machine before that's why I haven't noticed it until now

melekes: but what if I want to run it on my own machine?

sabau: Should work now

melekes: it failed because the dockers where it runs need the Linux version of it.

I don't understand. `build` target should build Linux version when run inside Linux

sabau: The containers are unable to build, I guess they were created as minimal as possible, just able to run a pre-made binary.

The past days I was working on a Linux machine, so this was all good.

Today I'm on my Mac and by purging the build folder before every run, the `make testnet-start` command failed, because I was building our `tendermint` binary for Mac and trying to use it on the Linux containers

networks/local/localnode/Dockerfile
```diff
@@ -14,3 +14,4 @@ STOPSIGNAL SIGTERM
 
 COPY wrapper.sh /usr/bin/wrapper.sh
 COPY config-template.toml /etc/tendermint/config-template.toml
+COPY ../../../build/tendermint /tendermint/tendermint
```
melekes: why do you think we need this?

fadeev:
> @marbar3778 @sabau you guys need to loop in the design team and check how the website is currently set up and what it would take to integrate this into it.

If I understand correctly, to switch to Swagger (from Slate) we would have to edit the `Makefile`:

https://github.com/tendermint/tendermint.com/blob/master/Makefile#L48

so that it would copy `docs/spec/rpc/swagger.yaml ` and `docs/spec/rpc/index.html ` from this repo into `dist/rpc/` in `tendermint.com` repo.

networks/local/localnode/Dockerfile
sabau: Not needed, it's just to debug why the tendermint executable is not in the mounted volume on CircleCI.

sabau:
A follow-up PR could make the network a bit more interesting to tests, adding addresses, transactions hashes to look at etc

networks/local/localnode/Dockerfile
sabau: just removed it, the experiment failed, I had to switch to `machine` in CircleCI

docs/spec/rpc/swagger.yaml
```diff
@@ -0,0 +1,2680 @@
+swagger: "2.0"
+info:
+  version: "0.32.1"
```
thanethomson: Will we have to manually update the version each time?

docs/spec/rpc/swagger.yaml
```diff
@@ -0,0 +1,2680 @@
...
+    description: Informations about the node APIs
+  - name: Tx
+    description: Transactions broadcast APIs
+  - name: ABCI
+    description: ABCI APIs
+  - name: Evidence
+    description: Evidence APIs
+schemes:
+  - https
+host: stargate.cosmos.network:26657
```
thanethomson: Should we be using this address here? Due to the more generic nature of Tendermint, I'd imagine we'd need to provide an address that's relevant to a broader user base. The default for Tendermint configuration at the moment is to bind to `127.0.0.1`, so should we perhaps not just use:
```suggestion
host: localhost:26657
```

docs/spec/rpc/swagger.yaml
```diff
@@ -0,0 +1,2680 @@
...
+  - name: Info
+    description: Informations about the node APIs
+  - name: Tx
+    description: Transactions broadcast APIs
+  - name: ABCI
+    description: ABCI APIs
+  - name: Evidence
+    description: Evidence APIs
+schemes:
+  - https
```
thanethomson: By default we expose an HTTP interface - could we add that in here?

docs/spec/rpc/swagger.yaml
```diff
@@ -0,0 +1,2680 @@
...
+                          type: "string"
+                          example: "2019-08-05T11:28:49.21730864Z"
+                        proposal:
+                          type: "boolean"
+                          example: false
+                        proposal_block_parts_header:
+                          required:
+                            - "total"
+                            - "hash"
+
```
thanethomson: Is there a way to indicate that this can be either a string or an integer? Right now it actually comes back as an integer. See #3898. I think we're starting to move towards having integers be represented as integers.

thanethomson: Right now this is actually an integer. See #3898 - I'm sure we're going to rather tend towards representing all integers as integers instead of strings in future.

thanethomson: I've tried adding this comment twice now and GitHub seems to consistently get the line number wrong somehow. I'm actually referring to `docs/spec/rpc/swagger.yaml#2666` 🤷‍♂

docs/spec/rpc/swagger.yaml
marbar3778: I do believe this is so when a user comes to the swagger docs they can hit the endpoint and get results back instead of having to spin a local node to be able to test it. similar to the try it out button here: https://cosmos.network/rpc/#/ICS0/get_node_info

thanethomson: Perhaps we should look into setting up a Tendermint/kvstore node here instead then, running the latest version of Tendermint? Something generic where users can follow along in the Tendermint documentation to interact with the server.

.circleci/config.yml
```diff
@@ -359,6 +359,35 @@ jobs:
       - store_artifacts:
           path: /go/src/github.com/tendermint/tendermint/tendermint-*.tar.gz
 
+  # Test RPC implementation against the swagger documented specs
+  contract_tests:
+    working_directory: /home/circleci/.go_workspace/src/github.com/tendermint/tendermint
+    machine:
+      image: circleci/classic:latest
```
melekes: why not default one or nodejs?

marbar3778: For this @sabau mentioned its either start in node and download go or start in go and download node, he chose node.

melekes: he chose neither!!

safareli:
When would this be hosted on tendermint website?

Issue #3882, opened 2019-08-06
Don't use `omitempty`


safareli:
Types which are generated from `proto` files in go have JSON annotations which contain `omitempty` for all keys, for example:
https://github.com/tendermint/tendermint/blob/62f97a69e97262b5feb57b1c2498f0c1e0e297b3/abci/types/types.proto#L137-L145

Results in this:
https://github.com/tendermint/tendermint/blob/62f97a69e97262b5feb57b1c2498f0c1e0e297b3/abci/types/types.pb.go#L1711-L1720

Which means if you just make empty `ResponseInfo{}` and serialize it to JSON you would get empty JSON object, or if you do not set some field that field will be omitted from resulting JSON.

On the other hand, types defined by hand in go don't use `omitempty` at all, all this results in inconsistency. As some fields are in json all the time and some aren't, it makes writing parsers of request/response types unnecessarily harder in other statically typed languages.

safareli:
Quick google gave me this article which seems to solve the issue
https://medium.com/@nevio/golang-dealing-with-protocol-buffer-omitempty-and-boolean-json-responses-ae1ba5f4e9c3

But I'm not a go developer and can't pick it up myself.

marbar3778:
>  As some fields are in json all the time and some aren't, it makes writing parsers of request/response types unnecessarily harder in other statically typed languages.
I agree with the reasoning herer, @melekes, thoughts?

safareli:
As a note, besides types from /abci/types/types.proto, other types from:
- /crypto/merkle/merkle.proto
- /types/proto3/block.proto
- /libs/common/types.proto
are also exposed as part of RPC

melekes:
Thanks for bringing this up @safareli. Based on what I know there are four options available:

1) create intermediate structs for RPC specifically (+) or 
2) set EmitDefaults to true when marshaling https://stackoverflow.com/a/50044963/820520 (++) or 
3) use gogo proto json.tag feature (+) or
4) post-processing script https://github.com/tendermint/tendermint/issues/3882#issuecomment-518770040

safareli:
No problem!, happy for seeing responce so fast 🙌

While we are at it, there is one type which I noticed in go side having no json annotations reslting in fields named `LikeThis` in json probably fixing it would be good too: https://github.com/tendermint/tendermint/blob/v0.32.2/types/tx.go#L85

blinky3713:
[...]

PR #3889, opened 2019-08-08
Remove omitempty from *pb.go

marbar3778:
- remove omitempty from *pb.go files
- added command to makefile for everytime `make protoc_all` is run

- open question:
  - Do we want to further remove omitempty from other places
	- https://github.com/tendermint/tendermint/blob/master/rpc/lib/types/types.go#L151
	- and other places

ref #3882

* [ ] Wrote tests


abci/types/result.go
```diff
@@ -49,7 +49,7 @@ func (r ResponseQuery) IsErr() bool {
 var (
```
melekes: above comment needs to be updated as well

Makefile
```diff
@@ -40,6 +40,7 @@ install_c:
 ### Protobuf
 
 protoc_all: protoc_libs protoc_merkle protoc_abci protoc_grpc protoc_proto3types
+	@ls **/**/**.pb.go | xargs -n1 -IX bash -c 'sed s/,omitempty// X > X.tmp && mv X{.tmp,}'
```
melekes: I'd rather leave omitempty property in pb.go files. It reduces the size of the end data when sent over p2p network. I'd prefer to use custom marshallers with EmitDefaults set to true for RPC specifically.

blinky3713:
I'm looking at your most recent commit and I'm not if we're on the same page, so I just want to lay out how we understand the problem. 

If you have `omitempty` in any field for JSON codecs, it causes the key to be omitted during serialization if the field is uninitialized. This makes it difficult for strongly typed languages (in our case Haskell) to parse the data idiomatically. For example this addition you've made https://github.com/tendermint/tendermint/pull/3889/commits/b50aa9a9ca0ee0b8d8ab08c4dfbac6c6702a12e3#diff-e41225101b540b7e08bbbcddb8f8ffdcR499 defines

```go
type RequestEcho struct {
	Message              string   `protobuf:"bytes,1,opt,name=message,proto3" json:"message,omitempty"`
	XXX_NoUnkeyedLiteral struct{} `json:"-"`
	XXX_unrecognized     []byte   `json:"-"`
	XXX_sizecache        int32    `json:"-"`
```

Doesn't this mean that in the event that the `Message` field wasn't initialized, then the JSON object will not include a `message` field?

Sorry if I'm wrong, but again just wanting to make sure we're on the same page. Long story short, if keys sometimes appear, sometimes don't in makes it really hard for us. If they don't have a value, you are still free to use JSON's `null` value, but it has to be there. Thanks!

marbar3778:
Yes, we are on the same page. @melekes requested to change to use jsonpb and not remove omitempty from the *pb.go files. So it would be handled by the marshler instead of the script to remove from the pb.go files. This is how I understood https://github.com/tendermint/tendermint/pull/3889#discussion_r311961963

blinky3713:
🙏

types/tx.go
```diff
@@ -83,9 +83,9 @@ func (txs Txs) Proof(i int) TxProof {
 
 // TxProof represents a Merkle proof of the presence of a transaction in the Merkle tree.
 type TxProof struct {
-	RootHash cmn.HexBytes
-	Data     Tx
-	Proof    merkle.SimpleProof
+	RootHash cmn.HexBytes       `json:"root_hash"`
+	Data     Tx                 `json:"Data"`
```
ebuchman:
```suggestion
	Data     Tx                 `json:"data"`
```

abci/types/types.pb.go
```diff
@@ -496,7 +496,7 @@ func _Request_OneofSizer(msg proto.Message) (n int) {
 }
 
 type RequestEcho struct {
-	Message              string   `protobuf:"bytes,1,opt,name=message,proto3" json:"message"`
+	Message              string   `protobuf:"bytes,1,opt,name=message,proto3" json:"message,omitempty"`
```
ebuchman: Maybe want to undo changes to this file too?

benchmarks/proto/test.pb.go
```diff
@@ -34,11 +34,11 @@ var _ = math.Inf
 const _ = proto.GoGoProtoPackageIsVersion2 // please upgrade the proto package
 
 type ResultStatus struct {
-	NodeInfo          *NodeInfo `protobuf:"bytes,1,opt,name=nodeInfo" json:"nodeInfo,omitempty"`
-	PubKey            *PubKey   `protobuf:"bytes,2,req,name=pubKey" json:"pubKey,omitempty"`
-	LatestBlockHash   []byte    `protobuf:"bytes,3,req,name=latestBlockHash" json:"latestBlockHash,omitempty"`
-	LatestBlockHeight *int64    `protobuf:"varint,4,req,name=latestBlockHeight" json:"latestBlockHeight,omitempty"`
-	LatestBlocktime   *int64    `protobuf:"varint,5,req,name=latestBlocktime" json:"latestBlocktime,omitempty"`
+	NodeInfo          *NodeInfo `protobuf:"bytes,1,opt,name=nodeInfo" json:"nodeInfo"`
```
ebuchman: don't we want to undo this one too?

crypto/merkle/result.go
```diff
@@ -0,0 +1,54 @@
+package merkle
```
golangcibot: 1-54 lines are duplicate of `libs/common/result.go:1-54` (from `dupl`)

libs/common/result.go
```diff
@@ -0,0 +1,54 @@
+package common
```
golangcibot: 1-54 lines are duplicate of `crypto/merkle/result.go:1-54` (from `dupl`)

abci/types/messages_test.go
```diff
@@ -13,10 +12,9 @@ import (
 )
 
 func TestMarshalJSON(t *testing.T) {
-	b, err := json.Marshal(&ResponseDeliverTx{})
+	_, err := json.Marshal(&ResponseDeliverTx{})
 	assert.Nil(t, err)
-	// Do not include empty fields.
-	assert.False(t, strings.Contains(string(b), "code"))
```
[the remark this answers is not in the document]
marbar3778: this now checks that values that are empty eg. `code` is there if its empty.

crypto/merkle/result.go
```diff
@@ -0,0 +1,55 @@
...
+	"bytes"
+	"encoding/json"
+
+	"github.com/gogo/protobuf/jsonpb"
+)
+
+//---------------------------------------------------------------------------
+// override JSON marshalling so we emit defaults (ie. disable omitempty)
+// note we need Unmarshal functions too because protobuf had the bright idea
+// to marshal int64->string. cool. cool, cool, cool: https://developers.google.com/protocol-buffers/docs/proto3#json
```
safareli: I might be misunderstanding something but AFAIK int64, fixed64 and uint64 should be encoded as strings?
https://github.com/tendermint/tendermint/issues/3898#issuecomment-521344376

abci/types/result.go
```diff
@@ -42,14 +42,14 @@ func (r ResponseQuery) IsErr() bool {
 }
 
 //---------------------------------------------------------------------------
-// override JSON marshalling so we dont emit defaults (ie. disable omitempty)
+// override JSON marshalling so we emit defaults (ie. disable omitempty)
```
melekes:
```suggestion
// override JSON marshalling so we emit defaults (ie. enable omitempty)
```

melekes:
```
=== RUN   TestTransportMultiplexValidateNodeInfo
panic: send on closed channel

goroutine 3027 [running]:
github.com/tendermint/tendermint/p2p.TestTransportMultiplexAcceptNonBlocking.func1(0xc000d1c100, 0xc000cca240, 0xc000cca300, 0xc000cca2a0)
	/go/src/github.com/tendermint/tendermint/p2p/transport_test.go:255 +0x55c
created by github.com/tendermint/tendermint/p2p.TestTransportMultiplexAcceptNonBlocking
	/go/src/github.com/tendermint/tendermint/p2p/transport_test.go:224 +0x271
```

not sure if it's related. probably not

marbar3778:
@blinky3713 you can give the branch a try now, sorry for the delay

crypto/merkle/result.go
melekes: Yep, we should probably remove the "note we need ..."

abci/types/result.go
```diff
@@ -42,9 +42,7 @@ func (r ResponseQuery) IsErr() bool {
 }
 
 //---------------------------------------------------------------------------
-// override JSON marshalling so we emit defaults (ie. disable omitempty)
-// note we need Unmarshal functions too because protobuf had the bright idea
-// to marshal int64->string. cool. cool, cool, cool: https://developers.google.com/protocol-buffers/docs/proto3#json
+// override JSON marshalling so we emit defaults (ie. enable omitempty)
```
melekes: "disable omitempty" was correct. sorry for confusion

melekes:
@marbar3778 can you also update the changelog?

Issue #3845, opened 2019-07-29
Add command to roll-back a single block


AdityaSripal:
Many times, a full node panics on Consensus failure due to a mismatch in the expected and actual `AppHash`. This is often caused by misconfiguration or running the wrong versions of software on the part of the full node, rather than a malicious validator set.

Currently the only solution I know of is to reset the data, fix the issue, then start over.

However it is possible for the node to get a mismatch after doing a lot of syncing. In this case, the node will panic on the first block for which it sees an app-hash mismatch.

Rather than resetting the data and undoing a lot of valid syncing, tendermint should allow the ability to rollback a single block's execution. This would allow the full node operator to diagnose the issue, fix it, and continue syncing from where it left off rather than having to start over.

The ability to Rollback a version is already available on IAVL, so rolling back app-state is not an issue. We would just need to add functionality to rollback rest of Tendermint State as well if that doesn't already exist, and create a cli command to do it.

If this feature already exists, it would be very useful to be pointed to the code where it is implemented.

ebuchman:
So it seems this might be simpler to fix than so far anticipated, assuming the latest committed block is correct, but the app hash computed from it is incorrect on less than 1/3 of the validators.

In this case, it should be possible to just rollback the app one height, fix things there, and restart tendermint. Tendermint will replay the latest block (which is itself correct), and give the app the chance to return the correct app hash.

Only now, the correct app hash returned from the app will differ from the one stored by Tendermint the last time the block was executed, and the replay mechanism will fail: https://github.com/tendermint/tendermint/blob/master/consensus/replay.go#L450

So it seems we might only need some way to let the app over ride the latest app hash stored by tendermint, rather than actually rolling anything back at the tendermint level.

Of course this also assumes the changes to the app won't impact the app hash from previous blocks.

Would this be sufficient to address this issue? Or is there some more pathological case ?

AdityaSripal:
Ah I see, yea i think that should solve the specific problem I have above. I'm not sure if there are any other situations where users might want to roll-back TM blocks.

Is there any danger in not having the check to make sure the new replayed app-hash equals the old app-hash?

tessr:
@AdityaSripal Is this something you're still interested in?

AdityaSripal:
No I believe Ethan's suggestion is sufficient. You may close this. Thanks!

sunnya97:
Is there an issue for @ebuchman's solution?

cmwaters:
[...]

PR #3901, opened 2019-08-15
Revert State Changes from Latest Block

AdityaSripal:
This PR aims to add a command to revert the latest block. This will undo any changes in Tendermint State and block store from the last time `ApplyBlock` was called. It should also rollback ApplicationState changes though that is currently unimplemented since further discussion is necessary on how that should be accomplished.

Reasons for why reverting a block would be useful are explained here: https://github.com/tendermint/tendermint/issues/3196 and https://github.com/tendermint/tendermint/issues/3845

Prior work: https://github.com/tendermint/tendermint/pull/3332

* [ ] Wrote tests

### DIFF
--- a/state/execution.go
+++ b/state/execution.go
@@ -179,6 +179,25 @@ func (blockExec *BlockExecutor) ApplyBlock(state State, blockID types.BlockID, b
 	return state, nil
 }
 
+// RevertBlock takes in a reverted block and the new latest block and returns the state
+// to its previous State before reverted block got applied
+// reverted block must be the LastBlockHeight of current state or will return error
+// Will not revert changes to mempool or evpool
+func (blockExec *BlockExecutor) RevertBlock(state State, reverted, newHead *types.Block) (State, error) {
+	if reverted.Header.Height != state.LastBlockHeight {
+		return state, fmt.Errorf("Can only revert latest block. Expected reverted block %d, got %d", state.LastBlockHeight, reverted.Header.Height)
+	}
+	prevState := blockExec.revertState(state, reverted, newHead)
+	// revert state to previous version
+	SaveState(blockExec.db, prevState)
+	// delete peripheral state info for reverted height
+	oldHeight := reverted.Header.Height
+	deleteABCIResponses(blockExec.db, oldHeight)
+	deleteConsensusParamsInfo(blockExec.db, oldHeight+1)
+	deleteValidatorsInfo(blockExec.db, oldHeight+2)
+	return prevState, nil
+}
+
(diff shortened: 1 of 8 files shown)


store/store.go
```diff
@@ -195,6 +195,46 @@ func (bs *BlockStore) saveBlockPart(height int64, index int, part *types.Part) {
...
+	blockID := blockMeta.BlockID
+	bs.db.Delete(calcBlockMetaKey(latest))
+
+	// Delete block parts
+	for i := 0; i < blockID.PartsHeader.Total; i++ {
+		bs.db.Delete(calcBlockPartKey(latest, i))
+	}
+
+	// Delete block commit
+	bs.db.Delete(calcBlockCommitKey(latest - 1))
```
AdityaSripal: Not sure if I should be deleting the `SeenCommit` for the same height as well

state/execution.go
```diff
@@ -179,6 +179,25 @@ func (blockExec *BlockExecutor) ApplyBlock(state State, blockID types.BlockID, b
 	return state, nil
 }
 
+// RevertBlock takes in a reverted block and the new latest block and returns the state
+// to its previous State before reverted block got applied
+// reverted block must be the LastBlockHeight of current state or will return error
+// Will not revert changes to mempool or evpool
+func (blockExec *BlockExecutor) RevertBlock(state State, reverted, newHead *types.Block) (State, error) {
```
AdityaSripal: How should `AppState` get rolled back? Should Tendermint call some ABCI Method to request proxy connection to rollback `AppState`?

alexanderbez: Great question. Maybe through ABCI we have the app `DeleteVersion` on every version up until the rollback height?

marbar3778:
@AdityaSripal can you confirm that this can be closed. I believe I remember you mentioning that it wasn't needed anymore.

AdityaSripal:
Yes, can you confirm @ebuchman ?

melekes:
How's this not an issue anymore?

melekes:
"I would double check the use case. The last time I weighed in I think we realized there was a much simpler way to do this: https://github.com/tendermint/tendermint/issues/3845#issuecomment-528388207. In any case, it may still be a useful dev tool to be able to undo blocks, but then it should be an independent binary: https://github.com/tendermint/tendermint/issues/3196#issuecomment-459872337" - @ebuchman

tessr:
Howdy folks. I'm going to close this PR--it looks like the motivations for it are captured elsewhere as well, and it sounds like this isn't the implementation strategy we'd want at this point.

Issue #3908, opened 2019-08-22
Creating 1 block per Tx when create_empty_block flag set to false


b00f:
**Tendermint version** : 0.32.0
**Environment**: Linux
- **OS**: Debian
- **Install tools**: go mod ...

**What happened**:
1- Set `create_empty_block` to false to not create empty block when there is no transaction in mempool
2- Try to send 10 txs simultaneously
3- Tendermint creates 10 (or 9) blocks very fast and there is only 1 (or maybe 2) txs in each block

**What you expected to happen**:
- Tendermint should wait for commit timeout after recieving first tx, Then notifying consensus engine
- Tendermint should create 1 block with 10 txs

**Have you tried the latest version**: no

**Logs (paste a small part showing an error (< 10 lines) or link a pastebin, gist, etc. containing more of the log file)**:
```
notifyTxsAvailable:  size: 1, notifiedTxsAvailable: false
I[2019-08-22|15:30:10.519] Executed block                               module=state height=12 validTxs=1 invalidTxs=0
I[2019-08-22|15:30:10.524] Committed state                              module=state height=12 txs=1 appHash=1F1FA71E94F5D63C667BAD9611440A1FBAEE5D6436970250EDAD3FD27A0CCC0B
notifyTxsAvailable:  size: 1, notifiedTxsAvailable: false
I[2019-08-22|15:30:10.577] Executed block                               module=state height=13 validTxs=1 invalidTxs=0
I[2019-08-22|15:30:10.583] Committed state                              module=state height=13 txs=1 appHash=3911B0BB1BF87887EB2AF66C5874F87D1888A501F8A8DE033107C31EB7D5FA0C
notifyTxsAvailable:  size: 1, notifiedTxsAvailable: false
I[2019-08-22|15:30:10.648] Executed block                               module=state height=14 validTxs=1 invalidTxs=0
notifyTxsAvailable:  size: 2, notifiedTxsAvailable: true
I[2019-08-22|15:30:10.654] Committed state                              module=state height=14 txs=1 appHash=13C195F3C26CC8E2B28DF661BAC2BCC6F73FEADEE888C252AC2F3AF887CDF212
resCbRecheck:  size: 1, notifiedTxsAvailable: false
notifyTxsAvailable:  size: 1, notifiedTxsAvailable: false
notifyTxsAvailable:  size: 2, notifiedTxsAvailable: true
[...]

PR #3909, opened 2019-08-22
Wait for commit timeout before notifying consensus engine for availab…

b00f:
Fixing #3908

### DIFF
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -29,7 +29,7 @@ import (
 // mempool uses a concurrent list structure for storing transactions that can
 // be efficiently accessed by multiple concurrent readers.
 type CListMempool struct {
-	config *cfg.MempoolConfig
+	config *cfg.Config
 
 	proxyMtx     sync.Mutex
 	proxyAppConn proxy.AppConnMempool
@@ -75,7 +75,7 @@ type CListMempoolOption func(*CListMempool)
 
 // NewCListMempool returns a new mempool with the given configuration and connection to an application.
 func NewCListMempool(
-	config *cfg.MempoolConfig,
+	config *cfg.Config,
 	proxyAppConn proxy.AppConnMempool,
 	height int64,
 	options ...CListMempoolOption,
@@ -91,8 +91,8 @@ func NewCListMempool(
 		logger:        log.NewNopLogger(),
 		metrics:       NopMetrics(),
 	}
-	if config.CacheSize > 0 {
-		mempool.cache = newMapTxCache(config.CacheSize)
+	if config.Mempool.CacheSize > 0 {
+		mempool.cache = newMapTxCache(config.Mempool.CacheSize)
 	} else {
 		mempool.cache = nopTxCache{}
 	}
@@ -133,7 +133,7 @@ func WithMetrics(metrics *Metrics) CListMempoolOption {
 // *panics* if can't create directory or open file.
 // *not thread safe*
 func (mem *CListMempool) InitWAL() {
-	walDir := mem.config.WalDir()
+	walDir := mem.config.Mempool.WalDir()
 	err := cmn.EnsureDir(walDir, 0700)
 	if err != nil {
 		panic(errors.Wrap(err, "Error ensuring WAL dir"))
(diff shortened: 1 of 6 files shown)


b00f:
The log after this fix, Compare with #3908 log

```
notifyTxsAvailable:  size: 1, notifiedTxsAvailable: false
notifyTxsAvailable:  size: 2, notifiedTxsAvailable: true
notifyTxsAvailable:  size: 3, notifiedTxsAvailable: true
notifyTxsAvailable:  size: 4, notifiedTxsAvailable: true
notifyTxsAvailable:  size: 5, notifiedTxsAvailable: true
notifyTxsAvailable:  size: 6, notifiedTxsAvailable: true
notifyTxsAvailable:  size: 7, notifiedTxsAvailable: true
notifyTxsAvailable:  size: 8, notifiedTxsAvailable: true
notifyTxsAvailable:  size: 9, notifiedTxsAvailable: true
notifyTxsAvailable:  size: 10, notifiedTxsAvailable: true
I[2019-08-22|15:28:48.660] Executed block                               module=state height=6 validTxs=10 invalidTxs=0
I[2019-08-22|15:28:48.670] Committed state                              module=state height=6 txs=10 appHash=57169ECA1BDDF8D1D527D3FF3D7B5274B4D01902020783965DADBBE2FAE1AEF4
I[2019-08-22|15:28:53.673] Executed block                               module=state height=7 validTxs=0 invalidTxs=0
I[2019-08-22|15:28:53.677] Committed state                              module=state height=7 txs=0 appHash=57169ECA1BDDF8D1D527D3FF3D7B5274B4D01902020783965DADBBE2FAE1AEF4
```

mempool/clist_mempool.go
```diff
@@ -29,7 +29,7 @@ import (
 // mempool uses a concurrent list structure for storing transactions that can
 // be efficiently accessed by multiple concurrent readers.
 type CListMempool struct {
-	config *cfg.MempoolConfig
+	config *cfg.Config
```
melekes: we must not require the full config here

mempool/clist_mempool.go
```diff
@@ -446,17 +446,22 @@ func (mem *CListMempool) TxsAvailable() <-chan struct{} {
...
-		panic("notified txs available but mempool is empty!")
-	}
-	if mem.txsAvailable != nil && !mem.notifiedTxsAvailable {
-		// channel cap is 1, so this will send once
-		mem.notifiedTxsAvailable = true
-		select {
-		case mem.txsAvailable <- struct{}{}:
-		default:
+	go func() {
+		if mem.Size() == 0 {
```
melekes: I think this check should be outside of the goroutine and it should really panic.

mempool/clist_mempool.go
```diff
@@ -446,17 +446,22 @@ func (mem *CListMempool) TxsAvailable() <-chan struct{} {
...
+			mem.logger.Error("notified txs available but mempool is empty!")
+			return
 		}
-	}
+		if mem.txsAvailable != nil && !mem.notifiedTxsAvailable {
+			// channel cap is 1, so this will send once
+			mem.notifiedTxsAvailable = true
+			time.Sleep(mem.config.Consensus.TimeoutCommit)
```
melekes: why `TimeoutCommit` though and not some configurable interval (say 1s)?

mempool/clist_mempool.go
b00f: we need to have timeout for mempool before sending notification. I think commit-timeout is fair enough. 
I didn't want to add new item to the mempool config. and I used the consensus config. It makes sense.

mempool/clist_mempool.go
b00f: Oh no panic please. Who knows what happens in threading apps!

mempool/clist_mempool.go
b00f: In my tests I never faced this condition.

mempool/clist_mempool.go
b00f: if 1 sec, then we create a block each second. But in consensus engine commit time out is set to 5 (default). I believe commit timeout should be same regardless of create_empty_block flag is set or not.

melekes: I am not talking about changing TimeoutCommit. But rather using another variable

b00f: You mean adding new item to mempool config?

melekes: Yeah

b00f: `no_empty_blocks_wait`, Is it good name for this config? Any suggestion?

b00f: Another way, What's your idea I pass `TimeoutCommit` an new param to the constructor? So I don't need to pass whole config here.

melekes:
Closing in favor of https://github.com/tendermint/tendermint/pull/3928

PR #3911, opened 2019-08-22
docs: write about implementation in Light Client ADR

melekes:
Refs https://github.com/tendermint/tendermint/issues/2133

* [ ] ~~Wrote tests~~


docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
```diff
@@ -121,6 +122,136 @@ network usage.
...
+
+```go
+func LinearVerification() Option {
+	return func(v *Verifier) {
+		v.mode = LINEAR
+	}
+}
+
+func BisectingVerification(trustLevel, trustLevelAdj float) Option {
+	if trustLevel > 1 || trustLevel < 1/3 || trustLevelAdj > 1 || trustLevelAdj < 1/3 {
```
zmanian: I don't see where trustLevel is defined

melekes: they will be defined as constants

```go
const (
	defaultBisectionTrustLevel = 1/3
	defaultBisectionTrustLevelAdj = 1/3
)
```

zmanian: But is this defined as the maximum amount the voting power can change between the HEAD header and the trusted header?

melekes: `defaultBisectionTrustLevel` - yes

`defaultBisectionTrustLevelAdj` is used when HEAD and the trusted (or pivot) headers are consecutive (i.e. HEAD height - 3, pivot height - 2)

melekes: https://github.com/tendermint/tendermint/pull/3911/commits/91f66592f3de67ec47b6621a30b2ee9c6a6be27b

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
```diff
@@ -121,6 +122,166 @@ network usage.
...
+}
+
+// trustLevel - maximum change between two headers in terms of validators &
+// their respective voting power, required to trust a new header (default:
+// 1/3).
+//
+// trustLevelAdj - maximum change between two consequitive headers in terms of
+// validators & their respective voting power, required to trust a new header
+// (default: 2/3).
+func BisectingVerification(trustLevel, trustLevelAdj float) Option {
```
melekes: TODO: remove trustLevelAdj. decided 2/3 is always enough

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
```diff
@@ -53,22 +54,25 @@ network or when a light client that has been offline for longer than the
...
-
-    // Option 2: Callback can be set to implement a confirmation
-    // step if the trust store is uninitialized, or expired.
-    Callback func(height int64, hash []byte) error
+	// Required: only trust commits up to this old.
+	// Should be equal to the unbonding period minus some delta for evidence reporting.
+	TrustPeriod time.Duration `json:"trust-period"`
+
+	// Required: validator whom we've got the TrustHeight/Hash from
+	ValidatorAddress types.Address `json:"validator-address"`
```
melekes: Had to add validator address for evidence handling & submission

jackzampolin: This means full nodes can't publish hashes?

melekes: Yes. We can't slash full nodes because they aren't bonded.

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
```diff
@@ -121,6 +124,176 @@ network usage.
...
+	// 0 - latest
+	SignedHeader(height int64) (SignedHeader, error)
+
+  // 0 - latest
+  Validators(height int64) (ValidatorSet, error)
+}
+```
+
+In case of the proxy it will be a `http` provider (wrapping RPC client). For
+IBC, it will be a `ibc` provider, receiving information from IBC transactions.
```
ebuchman: I'm not sure there will be something in the blockchain app that is actually going to be calling these functions though? It seems like that's the relayers job, and then the relayer submits all this stuff and once it's recieved by the app it just needs to call verify with whatever it received. As far as I know the app isn't going to be asking anything to "please send me signed header for block 5", it will just try to verify whatever it receives.

So that can/should probably be better clarified.

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
```diff
@@ -121,6 +124,176 @@ network usage.
...
+	chainID            string
+	options            TrustOptions
+	lastVerifiedHeight int64
+	logger             log.Logger
+
+	mode       mode
+	trustLevel float
+
+	// Source of new headers.
+	source Provider
```
ebuchman: Don't think the `source` or `verifiers` make sense in the context of IBC?

melekes: relayer can be a source. I will rename `verifiers` to `alternativeSources`, which in IBC case, can be other relayers.

melekes:
> I took a stab at implementing that in rust here - https://github.com/interchainio/tendermint-rs/pull/31/files . It may be instructive for this ADR as well

So you're talking about making `Verifier` a "pure function". Probably a good thing to do.

But bisection logic requires something providing intermediate headers and I thought Provider naturally fits there. Btw, I don't see bisection algo in your PR.

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
melekes: > As far as I know the app isn't going to be asking anything to "please send me signed header for block 5", it will just try to verify whatever it receives.

What about bisection in IBC?

> So that can/should probably be better clarified.

Agree

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
```diff
@@ -53,22 +56,22 @@ network or when a light client that has been offline for longer than the
...
+	// Option 1: TrustHeight and TrustHash can both be provided
+	// to force the trusting of a particular height and hash.
+	// If the latest trusted height/hash is more recent, then this option is
+	// ignored.
+	TrustHeight int64  `json:"trust-height"`
+	TrustHash   []byte `json:"trust-hash"`
+
+	// Option 2: Callback can be set to implement a confirmation
+	// step if the trust store is uninitialized, or expired.
+	Callback func(height int64, hash []byte) error
```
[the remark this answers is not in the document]
melekes: > What's the deal with this Callback? Can we just have a struct that contains the initial configuration parameters?

idk yet. I was going to deal with it later 😄 

> Also what Hash are we working with here? Is it the header hash or the validator set hash? Presumably it should be the former.

Header's hash, I think

melekes:
Thanks for taking a look at this @ebuchman 🙇

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
zmanian: The expected behavior during IBC is that a relayer will perform the bisection and submit all the headers downloaded during the bisection in the IBC packet transaction.

The light client verifier on chain should then sequentially validate the headers provided.

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
```diff
@@ -23,7 +25,8 @@ synchronize a lite client without downloading the intervening headers.
 
 Light clients (and full nodes) operating in the Proof Of Stake context need a
 trusted block height from a trusted source that is no older than 1 unbonding
-window plus a configurable evidence submission synchrony bound. This is called “weak subjectivity”.
+window plus a configurable evidence submission synchrony bound. This is called
```
milosevic: We assume normally smaller trusted period tp for the lite client such that tp + evidence submission bound < unbonding period.

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
milosevic: This is very interesting as it seems that part of bisection logic could (should) be implemented by the relayer.

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
```diff
@@ -121,6 +124,176 @@ network usage.
...
+IBC, it will be a `ibc` provider, receiving information from IBC transactions.
+
+Once we have the information, we need to verify it.
+
+```go
+type mode int
+
+const (
+	sequential mode = iota
+	bisecting
```
milosevic: Probably not the right place to raise this but isn't it better calling this skipping mode as bisection seems like implementation method for the skipping case (not the only option however).

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
```diff
@@ -121,6 +124,176 @@ network usage.
...
+
+	mode       mode
+	trustLevel float
+
+	// Source of new headers.
+	source Provider
+
+	// Alternative sources for checking the primary for misbehavior by comparing data.
+	// If the primary misbehaves, we report the evidence to them.
+	verifiers []ProviderAndEvidenceReporter
```
milosevic: The idea is to separate cross-checking logic in a separate failure detection module.

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
```diff
@@ -121,6 +124,176 @@ network usage.
...
+
+	// Where trusted headers are stored.
```
milosevic: Maybe better to say where verified headers are stored as trust is temporal property. Note that after trusted period of a header expires, we still want to keep it at least for some time so we can give time to failure detection module to find conflicting header in case of attack (fork).

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
```diff
@@ -121,6 +124,176 @@ network usage.
...
+require.NoError(t, err)
+
+err = c.Verify(height)
+require.NoError(t, err)
+```
+
+**Evidence Handling and Reporting**
+
+light client should also be able to submit evidence of malfeasance and handle
+evidence coming from a full node or another source.
```
milosevic: Why lite client need to handle evidence coming from a full node? What would be the use case?

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
```diff
@@ -121,6 +124,176 @@ network usage.
...
+
+We'll need to add `Evidence(height int64)` func to `Provider` interface.
+
+When/if evidence is received, client should check it and disconnect from the
+node if `evidence.Address == TrustOptions.ValidatorAddress`. It's unwise to
```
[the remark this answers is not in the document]
milosevic: It's probably cleaner if failure detection module just send signal to core lite client that it should disconnect from the full node it is connected to, and reinitialise its root of trust.

docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
```diff
@@ -129,13 +302,15 @@ Accepted.
...
 
 * complexity of bisection
 
 ### Neutral
 
-* social consensus can be prone to errors (for cases where a new light client
-  joins a network or it has been offline for too long)
+* if the light client has been offline for `duration > TrustPeriod`, a user
+  will have to reinitialize it
+* evidence propagation delay prevents the light client from immediately
```
milosevic: Not sure we can reason that way. We can think about user defined policy on when to consider header as trusted (for example waiting some time, receiving header from multiple full nodes, etc), but in any case this does not guarantee that a user is still seeing forged header. Therefore, we always act after the fact and by having these mechanisms incentivise correct behaviour so that user can trust the verified header.

marbar3778:
What is the status of this PR?

melekes:
Closing in favor of https://github.com/tendermint/tendermint/pull/4209

PR #3912, opened 2019-08-22
address PR comments and concerns

blinky3713:
I'm submitting this to address the comments and concerns raised by @melekes [here](https://github.com/tendermint/tendermint/pull/3889#pullrequestreview-273472120) so that we can merge PR #3889. Hopefully this takes care of it.


marbar3778:
I dont believe this addresses the comments as omitempty will still exist, will need to use jsonpb instead of json, correct?

marbar3778:
Will close this PR, commits here were handled in https://github.com/tendermint/tendermint/pull/3889

Issue #3903, opened 2019-08-16
Swagger Static Follow-up


marbar3778:
This issue is to follow up on swagger documentation PR #3880 .

- [x] Comment https://github.com/tendermint/tendermint/pull/3880#discussion_r313965087
- [x] Comment https://github.com/tendermint/tendermint/pull/3880#discussion_r313966400
- [ ] Comment https://github.com/tendermint/tendermint/pull/3880#discussion_r313966723
- [x] Comment https://github.com/tendermint/tendermint/pull/3880#discussion_r313972498
- [x] Comment https://github.com/tendermint/tendermint/pull/3880#discussion_r314636491
- [x] Comment https://github.com/tendermint/tendermint/pull/3880#discussion_r314636678

marbar3778:
closing this as all the needed updates have been made

PR #3921, opened 2019-08-26
Test nodejs image

marbar3778:
Testing Nodejs image
follow up to come comments from #3903 

* [ ] Wrote tests


.circleci/config.yml
```diff
@@ -373,12 +373,12 @@ jobs:
     steps:
```
marbar3778: the reason i had to revert the above lines is that this config is using a machine and not a docker image, in circle ci they only have machines that are `classic` and there is no well-maintained image of docker that contains docker and node, so its either we do an image for this use case or just keep it as is @melekes

PR #3927, opened 2019-08-28
checkTxLocal fix #3919

### LINKED ISSUE #3919: Add a checkTxLocal method in mempool interface
Our application is based on ethermint and Ethereum, each transaction is signed using secp256k and has three fields r,s,v, the signer of a transaction (More precisely, the signer's public key and the address) is computed from the three fields. I find the computation runs four times for each tx:
1. When the tx first arrive in Ethereum's txPool
2. When ethermint's receive tx routine call tendermint's mempool's checkTx, which will call back to ethermint's checkTx func 
3. When DeliverTx, the tx is rlp decoded, and is executed in etherem's EVM
4. When the commit of every block triggers recheckTx.   This is kind of a problem, I find 30000~40000 txs in mempool need about 8 seconds to finish the recheck.    We have extended the memPool size to 50000.

I see #3735 has added a type field in Request_CheckTx, there are two types:
CheckTx_New and CheckTx_Recheck.
So problem 4 is solved.

As for DeliverTx, the check and restore the signer is inevitable, unless the application adopt some cache mechanism, cache the txHash and the From it has already known. I think the cache mechanism is necessary because for recheck tx, the app needs to query the cache to skip the signature verification too.

As for problem 2, I think we can add a new type in  Request_CheckTx in order to let the application know this tx is triggered by itself  or ,    by receiving from peer or rpc interface called by client.
The application can just store the "From" of current handling tx. When it judges it is a checkTx from it own triggered, it can just fetch the "From", without checking the signature to restore the signer. And the "From" can be free from lock, because the receive tx routine waits for each response to start handling the next tx, there is no other routine reads or writes the "From".

james-ray:
* [ ] Wrote tests
* [y ] Updated CHANGELOG_PENDING.md

fix #3919 

I delete CheckTxType_New, add a CheckTxType_Remote.
I think The CheckTxType_New can be replaced with CheckTxType_Remote, because if the tx comes from local application, the application should have checked the tx signature at the very beginning when the tx arrives. The CheckTxType_Remote means receiving tx from peers, or from rpc call.
The local application should have certain way to call mempool.CheckTxLocal through the code, the checkTxLocal should not be exposed through rpc call.


melekes:
If anything, the flag should be CheckTxType_Local, so the app can skip signatures verification. Otherwise, if it's set to Remote, how would the app know if it should do so in case of the tx rechecking.

melekes:
We need to discuss the best way to go around the original problem. I will close this PR temporary. Let's discuss in https://github.com/tendermint/tendermint/issues/3919

PR #3930, opened 2019-08-30
Merging BlockMeta with Block.

yenkhoon:
Fixing https://github.com/tendermint/tendermint/issues/3188

* [ ] Wrote tests

### DIFF
--- a/types/block_meta.go
+++ b/types/block_meta.go
@@ -2,15 +2,17 @@ package types
 
 // BlockMeta contains meta information about a block - namely, it's ID and Header.
 type BlockMeta struct {
-	BlockID BlockID `json:"block_id"` // the block hash and partsethash
-	Block *Block `json:"block"`   // The block data
+	BlockID   BlockID `json:"block_id"`   // the block hash and partsethash
+	BlockSize int     `json:"block_size"` //The block size
+	Block     *Block  `json:"block"`      // The block data
 }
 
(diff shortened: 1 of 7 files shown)


types/block_meta.go
```diff
@@ -3,14 +3,14 @@ package types
 // BlockMeta contains meta information about a block - namely, it's ID and Header.
 type BlockMeta struct {
 	BlockID BlockID `json:"block_id"` // the block hash and partsethash
-	Header  Header  `json:"header"`   // The block's Header
+	Block *Block `json:"block"`   // The block data
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	Block   *Block  `json:"block"`    // The block data
```

marbar3778:
Thank you for your contribution, I'm opposed to including the block in blockmeta, if we are doing this then it doesn't make sense to have a blockmeta type. 

In this PR https://github.com/tendermint/tendermint/pull/3856 I approached it a bit differently, instead removed blockmeta from result block because it comes in https://tendermint.com/rpc/#/Info/blockchain already, and instead favored including just the block and blockID. 

let me know what you think of this approack

melekes:
@yenkhoon thank you for submitting a PR! But I believe https://github.com/tendermint/tendermint/pull/3856 is the right way to go here. Next time please tell us you're going to work on an issue (by commenting on it!) before starting to work on it. This will allow us to avoid any duplicate effort.

Issue #3484, opened 2019-03-26
types: test block commits with votes for the wrong blockID


ebuchman:
Current code allows Commit to contain votes for the wrong blockID to facilitate tracking validator liveness.

We should add tests for this, both for the case of nil-votes and of votes for the wrong blockID. Both are allowed as valid votes in a Commit, so long as the Commit contains +2/3 votes for the right block ID.

Note this was discovered in #3298 which aimed to be the last step in implementing #1648 (ie. removing redundancy from commit votes). 

Ultimately, we maybe don't want to allow votes for the wrong blockID, but just for nil.

jscode017:
@ebuchman , do you mean that a test that testing a commit that contain +2/3 right votes?

ebuchman:
Yes, a commit that contains +2/3 of the right votes, but also some nil votes, and/or some votes for the wrong block id

tessr:
Closed by #3936

PR #3936, opened 2019-09-03
types: add test for block commits with votes for the wrong blockID

tessr:
Addresses #3484. 

I'm not completely sure if I'm on the right track here, so let me know if I am, and I can refactor this a little bit to have less code duplication. 

* [x] Wrote tests

### DIFF
--- a/types/block_test.go
+++ b/types/block_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/tendermint/tendermint/crypto"
 	"github.com/tendermint/tendermint/crypto/tmhash"
 	cmn "github.com/tendermint/tendermint/libs/common"
+	tmtime "github.com/tendermint/tendermint/types/time"
 	"github.com/tendermint/tendermint/version"
 )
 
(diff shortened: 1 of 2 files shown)


types/block_test.go
```diff
@@ -367,6 +368,104 @@ func TestCommitToVoteSet(t *testing.T) {
 	}
 }
 
+func TestCommitToVoteSetWithBadVotes(t *testing.T) {
```
melekes: I think this should be called TestCommitWithVotesForAnotherBlock

types/block_test.go
```diff
@@ -367,6 +368,104 @@ func TestCommitToVoteSet(t *testing.T) {
...
+			Timestamp:        tmtime.Now(),
+		}
+
+		_, err := signAddVote(vals[i], vote, voteSet)
+		assert.NoError(t, err)
+	}
+	commit := voteSet.MakeCommit()
+
+	chainID := voteSet.ChainID()
+	voteSet2 := CommitToVoteSet(chainID, commit, valSet)
```
melekes: Do we have to test CommitToVoteSet func here? Should be enoth to call `MakeCommit` and maybe `commit.ValidateBasic()`

types/block_test.go
```diff
@@ -367,6 +368,104 @@ func TestCommitToVoteSet(t *testing.T) {
...
+	assert.NotNil(t, voteSet2)
+}
+
+func TestCommitToVoteSetWithNil(t *testing.T) {
```
melekes:
```suggestion
func TestCommitWithNilVotes(t *testing.T) {
```

types/block_test.go
```diff
@@ -367,6 +368,104 @@ func TestCommitToVoteSet(t *testing.T) {
...
+			Timestamp:        tmtime.Now(),
+		}
+
+		_, err := signAddVote(vals[i], vote, voteSet)
+		assert.NoError(t, err)
+	}
+	commit := voteSet.MakeCommit()
+
+	chainID := voteSet.ChainID()
+	voteSet2 := CommitToVoteSet(chainID, commit, valSet)
```
melekes: same here

types/block_test.go
tessr: I wasn't sure! It does seem like `CommitToVoteSet` does some validation on the vote, so I thought maybe we should test it, too. However, upon closer inspection, it seems like it's not doing any validation on the whole set of votes, so I think you're right.

tessr:
Thank you for the feedback! I think this is ready for another look 👀

types/block_test.go
```diff
@@ -367,6 +368,98 @@ func TestCommitToVoteSet(t *testing.T) {
...
+			Type:             PrecommitType,
+			BlockID:          lastID,
+			Timestamp:        tmtime.Now(),
+		}
+
+		_, err := signAddVote(vals[i], vote, voteSet)
+		assert.NoError(t, err)
+	}
+
+	// 1/3 vote for a different block
```
melekes:
```suggestion
	// 1/3 vote for different blocks
```

every call to `makeBlockIDRandom` produces different value imo

tessr: of course! good catch!

ebuchman:
This can be merged, but the issue shouldn't be closed probably until more of the above comment is addressed.

Also @tessr check out the updated ADR in https://github.com/tendermint/tendermint/pull/3958 for more details about changes coming here.

tessr:
@ebuchman Thanks for the pointers! I'll try to make this PR cover everything to address #3484.

tessr:
@ebuchman OK, so just to be clear--some of the upcoming changes (e.g., https://github.com/tendermint/tendermint/pull/3958/files#diff-fded621f112710b405b7ab1a26f0d6c0R113) should not be accommodated in this PR, right? That is, I should still be testing for votes for the wrong block ID, even though we won't allow them soon.

types/block_test.go
```diff
@@ -367,6 +368,66 @@ func TestCommitToVoteSet(t *testing.T) {
...
+
+	type commitVoteTest struct {
+		blockIDs      []BlockID
+		numVotes      []int // must sum to numValidators
+		numValidators int
+		valid         bool
+	}
+
+	testCases := []commitVoteTest{
+		commitVoteTest{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
		{
```

types/block_test.go
```diff
@@ -367,6 +368,66 @@ func TestCommitToVoteSet(t *testing.T) {
...
+		blockIDs      []BlockID
+		numVotes      []int // must sum to numValidators
+		numValidators int
+		valid         bool
+	}
+
+	testCases := []commitVoteTest{
+		{
+			[]BlockID{blockID, blockID2, blockID3}, []int{8, 1, 1}, 10, true,
+		}, commitVoteTest{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
		}, {
```

types/block_test.go
```diff
@@ -367,6 +368,66 @@ func TestCommitToVoteSet(t *testing.T) {
...
+		numValidators int
+		valid         bool
+	}
+
+	testCases := []commitVoteTest{
+		{
+			[]BlockID{blockID, blockID2, blockID3}, []int{8, 1, 1}, 10, true,
+		}, {
+			[]BlockID{blockID, blockID2, blockID3}, []int{67, 20, 13}, 100, true,
+		}, commitVoteTest{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

tessr:
This now:
✔️ uses a testing table
✔️ calls `VerifyCommit` in addition to `MakeCommit` 
✔️ is one test (with a long name) instead of two 

One note is that it's tricky to test the failure cases, since `MakeCommit` panics if there aren't > 2/3 valid votes. My inclination is to change that function to return an error instead (and then the caller can panic, if necessary). But let me know if I shouldn't do that. 

Please take another look!

melekes:
> One note is that it's tricky to test the failure cases, since MakeCommit panics

You can use `assert.Panics(t, func() { MakeCommit(...) }` func and a bool flag (say `mustPanic`) to determine if MakeCommit should panic for a given set of votes or not.

I am generally in favor of returning errors, but, in this case, those are really pre-condition checks. I.e. if there's no 2/3 or type != Precommit, then we must have done something terribly wrong and there's nothing we can do other than crash and signal to developers.

https://dave.cheney.net/2012/01/18/why-go-gets-exceptions-right good article! (an old one though)

tessr:
Ah, cool, I've never used much beyond the standard library for unit tests before, so I didn't know about `assert.Panics`. Thanks for the tip! 

Regarding panics more generally, I'm always wary of introducing them, but it does make sense if this is truly a "game over" situation. Thank you for walking me through the domain!

ebuchman:
> should not be accommodated in this PR, right? That is, I should still be testing for votes for the wrong block ID, even though we won't allow them soon.

I think so. The testing issue was opened before we decided definitively not to include incorrect block ids, so it might be silly to have the test, but since they were supported before and something is going to change, it would probably be nice if a test breaks when the code/behaviour changes :)

tessr:
OK, cool! I'm going to merge this unless there are any other changes I should make.

PR #3957, opened 2019-09-06
Add Vote to BeginBlock

### LINKED ISSUE #3956: Add Vote to BeginBlock
Hi! I'm working at FluenceLabs, and we're using Tendermint as a consensus engine to provide trusted execution environment for user applications. 

### In short
We upload Tendermint blocks to decentralized storage, and need to reconstruct blocks inside ABCI app from info from `BeginBlock` and `DeliverTx`. But we're signatures from `Vote`s since `VoteInfo` contains only a boolean flag and validator address.

#### Background
In addition to security guarantees given by Tendermint, we also add execution verification to detect Byzantine takeovers. It is essentially a delayed replay of all blocks and transactions, followed by a check that state produced by the Tendermint cluster equals to the state after delayed verification.

To be able to perform delayed verification, we aim to store Tendermint-produced blocks in a decentralized storage like Swarm or Filecoin, so verifiers (fishermen, essentially) are guaranteed to have access to all transaction history. 

#### The problem
To upload a block, we first need to collect it from Tendermint. This is currently done via RPC (both HTTP and Websocket). But there's a problem with collecting blocks only through RPC: it is not available during block replay (after Tendermint node restart). 

The reason this is a problem is that to guarantee all blocks are indeed stored in the decentralized storage, each block must include a storage link to the previous block. **So the block N could be committed only after block N-1 is uploaded.** Currently, our ABCI app waits in `commit` of block N until there's a storage link available for block N-1.

Because of that dependency the following is possible:
1. Block `N-1` was produced, committed, retrieved via RPC and started uploading
2. Tendermint has started block `N` (upload for `N-1` isn't finished yet)
3. Tendermint restarts (e.g., because system is rebooted), and block replay begins
4. ABCI app `commit` for block `N` hangs until there's a receipt for block `N-1`
5. RPC isn't available during replay, so the system can't proceed since block `N-1` isn't available

#### Solution
Currently we solve this problem with a dirty hack – we read blocks directly from `blockstore.db`. This is fragile since it is not a part of public Tendermint API, and overall a bad solution.

There are two possible solutions for that: 
1. Reconstruct blocks inside ABCI app
2. Make RPC available during block replay. 

After discussing that with members of Tendermint Core team in person, we decided to go with the former solution. And the only pieces of information that are lacking are validator's signatures encoded in `Vote` structure.

### Implementation
I will provide a PR that adds `Vote` to `VoteInfo` shortly, and would be glad to discuss all the details, and would be grateful to get some help and advice with that.

folex:
fixes #3956

I hope the issue describes the problem in enough detail. If not, let's discuss that :)

I tried to fix all the documentation and comments I could find, please point me if I missed something. Same with tests – I modified test on BeginBlock, but maybe I should write new test.

There are some TODOs in code – they are there to start discussion, I will delete them during the review.

Also, I was thinking about making `VoteInfo::Vote` an optional field, depending on configuration, so it is disabled by default to avoid transferring possibly big signatures over network, but I'm not sure if this is a good idea. If it is – let me know, I will do my best to implement that. Any help on how to better provide corresponding configuration value to the `state/execution` module would be much appreciated.

___
Also, I noticed that there's a little quirk in using `DOCKER/Makefile` – it expects tendermint binary by default in the current directory (which would be `DOCKER`), but binary usually is in `build` directory. I could add commands to build the project and copy binary so it is easier to use, but need approval from you since I'm not sure If I'm using it correctly:
```bash
cd tendermint
make build
cp build/tendermint DOCKER/tendermint
cd DOCKER
# build docker container
make build
```
___
* [ ] Updated all code comments where relevant <-- _not sure about this one, please point if I missed something_
* [x] Wrote tests


abci/types/types.proto
```diff
@@ -311,6 +311,22 @@ message ValidatorUpdate {
 message VoteInfo {
   Validator validator = 1 [(gogoproto.nullable)=false];
   bool signed_last_block = 2;
+  Vote full_vote = 3;
+}
+
+// Validator's vote for a block
+message Vote {
+    int32 type = 1;
```
marbar3778: we are working on restructuring the data in the vote here: https://github.com/tendermint/tendermint/pull/3958, you can comment with any further ideas.

types/protobuf.go
```diff
@@ -220,3 +220,22 @@ func (pb2tm) ValidatorUpdates(vals []abci.ValidatorUpdate) ([]*Validator, error)
 	}
 	return tmVals, nil
 }
+
+func (tm2pb) Vote(vote *CommitSig) *abci.Vote {
```
golangcibot: Vote not declared by package types (from `typecheck`)

state/execution.go
```diff
@@ -334,6 +334,8 @@ func getBeginBlockValidatorInfo(block *types.Block, stateDB dbm.DB) (abci.LastCo
 		voteInfo := abci.VoteInfo{
 			Validator:       types.TM2PB.Validator(val),
 			SignedLastBlock: vote != nil,
+			// TODO: maybe make it optional, set by config, disabled by default?
+			FullVote: types.TM2PB.Vote(vote),
```
golangcibot: unknown field `FullVote` in struct literal (from `typecheck`)

abci/types/typespb_test.go
```diff
@@ -4941,6 +5043,28 @@ func TestVoteInfoSize(t *testing.T) {
 	}
 }
 
+func TestVoteSize(t *testing.T) {
+	seed := time.Now().UnixNano()
+	popr := math_rand.New(math_rand.NewSource(seed))
+	p := NewPopulatedVote(popr, true)
```
golangcibot: undeclared name: `NewPopulatedVote` (from `typecheck`)

abci/types/typespb_test.go
```diff
@@ -4049,6 +4123,34 @@ func TestVoteInfoProtoCompactText(t *testing.T) {
...
+		t.Fatalf("seed = %d, %#v !Proto %#v", seed, msg, p)
+	}
+}
+
+func TestVoteProtoCompactText(t *testing.T) {
+	seed := time.Now().UnixNano()
+	popr := math_rand.New(math_rand.NewSource(seed))
+	p := NewPopulatedVote(popr, true)
+	dAtA := github_com_gogo_protobuf_proto.CompactTextString(p)
+	msg := &Vote{}
```
golangcibot: undeclared name: `Vote` (from `typecheck`)

state/execution_test.go
```diff
@@ -92,9 +93,12 @@ func TestBeginBlockValidators(t *testing.T) {
 				tc.expectedAbsentValidators[ctr] == i {
 
 				assert.False(t, v.SignedLastBlock)
+				assert.Nil(t, v.FullVote)
```
golangcibot: v.FullVote undefined (type types.VoteInfo has no field or method FullVote) (from `typecheck`)

state/execution.go
folex: I'm not sure why this error popped out – `FullVote` field is sure there in `types.pb.go`, and I ran `golangci-linter` myself – no mention of this.

types/protobuf.go
folex: Same here – `abci.Vote` is defined in `types.pb.go`, and local run of `golangci-lint` didn't say anything about that.

melekes:
Closing for discussion in https://github.com/tendermint/tendermint/issues/3956.

PR #3960, opened 2019-09-07
Fix linter errors thrown by `unconvert`, `goconst`, and `nakedret`

PSalant726:

This PR repairs linter errors seen when running the following commands:
`golangci-lint run --no-config --disable-all=true --enable=unconvert`
`golangci-lint run --no-config --disable-all=true --enable=goconst`
`golangci-lint run --no-config --disable-all=true --enable=nakedret`

Contributes to #3262

### DIFF
--- a/blockchain/v1/reactor_fsm_test.go
+++ b/blockchain/v1/reactor_fsm_test.go
@@ -140,7 +140,7 @@ func sBlockRespEv(current, expected string, peerID p2p.ID, height int64, prevBlo
 		data: bReactorEventData{
 			peerID: peerID,
 			height: height,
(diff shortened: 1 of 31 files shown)


blockchain/v1/reactor_fsm_test.go
```diff
@@ -769,7 +769,7 @@ forLoop:
 	for i := 0; i < int(numBlocks); i++ {
 
 		// Add the makeRequestEv step periodically.
-		if i%int(maxRequestsPerPeer) == 0 {
+		if maxRequestsPerPeer == 0 {
```
melekes: what happened to `i%`?

```suggestion
		if i%maxRequestsPerPeer == 0 {
```

PR #3963, opened 2019-09-08
Fix linter errors thrown by `scopelint`

PSalant726:

This PR repairs linter errors seen when running the following commands:
`golangci-lint run --no-config --disable-all=true --enable=scopelint`

Contributes to #3262


blockchain/v0/reactor_test.go
```diff
@@ -258,6 +258,7 @@ func TestBcBlockRequestMessageValidateBasic(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
+		tc := tc
```
melekes: first time I am seeing such construct, that's for sure :)

melekes: why do we need to do it?

PSalant726: In Go, the `tc` variable in the above loop iterations is actually a single variable, so in many cases (like the above), using it can create potential concurrency problems..

melekes:
Can you also remove `- scopelint` from `disable` list in `.golangci.yml`?

blockchain/v0/reactor_test.go
melekes: https://github.com/kyoh86/scopelint

PR #3966, opened 2019-09-09
Override Go's GOPROXY default value

alessio:
Go 1.13 uses http://proxy.golang.org as GOPROXY's default
value if the environment variable is unset or blank. This
patch makes tendermint's buildsystem override Go env's
default and fallback to 'direct' if and only if GOPROXY is
unset in the user's environment; else it honours users'
settings.

* [ ] Wrote tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,13 +6,17 @@ GOTOOLS = \
 GOBIN?=${GOPATH}/bin
 PACKAGES=$(shell go list ./...)
 OUTPUT?=build/tendermint
+GOPROXY ?= direct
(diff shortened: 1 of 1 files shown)


zmanian:
IMO this degrades security without clear privacy benefits.

This software is published on the open Internet where Google crawls it's dependencies. It hits google run source repositories to build itself.

The default GOPROXY service runs a verifiable log system that ensures view consistency between what software users when querying for a specific dependency and provides protection against software software supply chain attacks on Tendermint developers.

https://blog.golang.org/module-mirror-launch

nACK

Issue #3968, opened 2019-09-10
`mempool` Memory Loading Error on 32-bit Ubuntu 16.04 Machine


jon-certik:
**Tendermint version**: 0.32.3-22000c6a (Latest `master` to date)

**OS**:

```bash
Linux <MY_MACHINE> 4.15.0-58-generic #64~16.04.1-Ubuntu SMP Wed Aug 7 14:09:34 UTC 2019 i686 i686 i686 GNU/Linux
```

**What happened**: Running `go test .` in `mempool` module results in the following memory loading error.

**What you expected to happen**: All tests passed.

**Have you tried the latest version**: Yes. Latest version to date.

**How to reproduce it**: Run `go test .` in `memopool/` on a 32-bit Ubuntu 16.04 machine.

**Logs (paste a small part showing an error (< 10 lines) or link a pastebin, gist, etc. containing more of the log file)**:

```bash
--- FAIL: TestCacheAfterUpdate (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x804aa5b]

goroutine 12 [running]:
testing.tRunner.func1(0x9c9e1e0)
	/usr/local/go/src/testing/testing.go:830 +0x345
panic(0x86b1b00, 0x8e3b4c8)
	/usr/local/go/src/runtime/panic.go:522 +0x16e
runtime/internal/atomic.Load64(0xa3f25d4, 0x0, 0xa3f2584)
	/usr/local/go/src/runtime/internal/atomic/asm_386.s:194 +0xb
github.com/tendermint/tendermint/mempool.(*CListMempool).TxsBytes(...)
	/home/jun/Desktop/tendermint/mempool/clist_mempool.go:171
github.com/tendermint/tendermint/mempool.(*CListMempool).CheckTxWithInfo(0xa3f2580, 0x9c2bb5c, 0x1, 0x1, 0x0, 0x9c80000, 0x0, 0x0)
	/home/jun/Desktop/tendermint/mempool/clist_mempool.go:222 +0xa8
github.com/tendermint/tendermint/mempool.(*CListMempool).CheckTx(...)
	/home/jun/Desktop/tendermint/mempool/clist_mempool.go:212
github.com/tendermint/tendermint/mempool.TestCacheAfterUpdate(0x9c9e1e0)
[...]

PR #3969, opened 2019-09-10
Simple Patch for `atomic`'s Bug on 32-bit Machine

jon-certik:
## What For

The bug of `atomic` results in `go test` error in the `mempool` module. See #3968 for details.

## Bug Cause

This is related to `atomic` library's bug: [atomic](https://golang.org/pkg/sync/atomic/#pkg-note-BUG)

> On ARM, x86-32, and 32-bit MIPS, it is the caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically. The first word in a variable or in an allocated struct, array, or slice can be relied upon to be 64-bit aligned.

## Solution

In `clist_mempool.go`, the following variables are to be loaded by `atomic`'s `LoadInt64`:

```
// Atomic integers
height     int64 // the last block Update()'d to
rechecking int32 // for re-checking filtered txs on Update()
txsBytes   int64 // total size of mempool, in bytes
```

On 32-bit architecture, the memory alignment guarantee of type `int64` is 32-bit. However `atomic.LoadInt64` requires "64-bit alignment of 64-bit words".

The current alignment for `txsBytes` is not 64-bit aligned and thus would result in error upon being read by `atomic.LoadInt64` on 32-bit architecture.

Switching the `rechecking` and `txsBytes` fields of `CListMempool` makes the `txsBytes` to be 64-bit aligned and passed all the tests (`go test`).

## Result

All tests in `mempool` passed after the patch:

```bash
PASS
ok  	github.com/tendermint/tendermint/mempool	26.860s
```

## Checklist

* [x] Wrote tests: Already covered by the original tests.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -23,4 +23,4 @@ program](https://hackerone.com/tendermint).
 
 ### BUG FIXES:
 
-- [mempool] \#3968 `mempool` Memory Loading Error on 32-bit Ubuntu 16.04 Machine.
+- [mempool] \#3968 Fix memory loading error on 32-bit machines (@jon-certik)
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -22,3 +22,5 @@ program](https://hackerone.com/tendermint).
 - [rpc] \#2010 Add NewHTTPWithClient and NewJSONRPCClientWithHTTPClient (note these and NewHTTP, NewJSONRPCClient functions panic if remote is invalid) (@gracenoah)
 
 ### BUG FIXES:
+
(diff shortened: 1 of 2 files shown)


CHANGELOG_PENDING.md
```diff
@@ -22,3 +22,5 @@ program](https://hackerone.com/tendermint).
 - [rpc] \#2010 Add NewHTTPWithClient and NewJSONRPCClientWithHTTPClient (note these and NewHTTP, NewJSONRPCClient functions panic if remote is invalid) (@gracenoah)
 
 ### BUG FIXES:
+
+- [mempool] \#3968 `mempool` Memory Loading Error on 32-bit Ubuntu 16.04 Machine.
```
```suggestion
- [mempool] \#3968 Fix memory loading error on 32-bit machines (@jon-certik)
```

Issue #3900, opened 2019-08-15
6 Month Objectives (Start July, 2019)


marbar3778:
This issue is meant to layout a 6-month roadmap, we have `First Objectives` and `Secondary Objectives`. I will be updating this periodically with updates according to what is finished and being worked on.

The objectives are broken into three categories - Feature, Refactor & Research.

## First Objectives
_Note: the order of the list holds no value_

- [ ] Feature - (ABCI): DeliverBatchTx with configurable batch size. This provides an abstraction that supports both deliverTx and DeliverBlock, thereby enabling apps to use batching and parallel algorithms naturally across txs.
   - [ADR 037: Deliver Block](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-037-deliver-block.md)
- [ ] Feature/Refactor - (Mempool): Allow the application to define its own mempool. Many components of mempool of logic are app-dependent. It would be nice to keep the standard mempool in Tendermint, but allow the app to use a different one. This is needed for mimblewimble like protocols.
- [ ] Feature - (State-sync): StateSync is a feature which would allow a new node to receive a snapshot of the application state without downloading blocks or going through consensus.
  -  [ADR 042: State Sync Design](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-042-state-sync.md)
- [x] Feature/Refactor - (Light Client)
   - [ADR 044: Lite Client with Weak Subjectivity](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-044-lite-client-with-weak-subjectivity.md)
   - https://github.com/tendermint/tendermint/pull/3841
   - https://github.com/tendermint/tendermint/pull/3577
   - *Update*: Lite client has been merged, you can take a peek at within the lite2 pkg
- [ ] Refactor - (Blockchain Reactor): Refactor the Blockchain reactor for better testing.
   - [ADR 040: Blockchain Reactor Refactor](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-040-blockchain-reactor-refactor.md)
  - [ADR 043: Blockhchain Reactor Riri-Org](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-043-blockchain-riri-org.md)
  - *Update*: The reactor is being finalized and preparing to be merged. Currently extensive testing is taking place on testnets and mainnets 
- [ ] Refactor - (Amino): Make amino protobuf compatible
   - https://github.com/tendermint/go-amino
[...]

PR #3980, opened 2019-09-12
Adding adr-048

b00f:
* [ ] Wrote tests


b00f:
@ebuchman May I know your idea about this?

melekes:
This is an interesting idea!

> I fully support the TTL additions

can't you do it already via app & check_tx?

> Personally, I think transparency and auditability is a core concept to support.

This seems to be the culprit ☝️ 

@b00f what are the use-cases behind this?

b00f:
> can't you do it already via app & check_tx?

No. Since a normal tx hasn't TTL. but stamped_tx has. An stamped_tx can be checked if it's alive yet in check_tx.

Transaction will be individual entity by stamping it.

b00f:
> What are the use-cases behind this?

Many use-cases. For example:
* Reducing the size of blockchain. Blocks can only keep tx_hash not whole Tx data. 
* TTL can help to mange Txs in a better way. 
* stamped_tx binds to the main fork. So it helps security of blockchain.

melekes:
> No. Since a normal tx hasn't TTL

I don't understand. Tendermint does not know content of transaction. You can put whatever you want in it - TTL, version, anything else you desire.

ethanfrey:
The TTL stuff sounds like cosmos-sdk zone.
But pruning historical blocks (or not storing all tx data) would need a change in tendermint

b00f:
> But pruning historical blocks (or not storing all tx data) would need a change in tendermint

This ADR is very general and can be applied to any blockchain. But since I love Tendermint I suggested it here.

If you look carefully it doesn't necessary need to change anything in Tendermint. Two interesting facts:
1- Application validate transactions. 
2- Tendermint is agnostic about transactions data

It will be like this: Application receive a Tx. Validate it and save it to its own DB and probably gossip it. (There is no consensus about this DB). Application now can only pass the hash of TX to the Tendermint. Then Tendermint do its routine works. If an application has that Tx in its own DB, simply can get it ad verify it, otherwise request it from other applications. 

Note: Mempool can move to app layer for better implementation. 

In this scenario block id will be simply block hash. We can get rid of block_parts, which is a little bit complicated. block never get that big to be partitioned.

marbar3778:
>Interesting proposal, thanks for the writeup. It probably doesn't belong as an ADR yet, since it's more of a protocol change than an architecture document, but we don't have a better place for that yet. We're working on setting up a better process for RFCs to make protocol changes like this. cc @marbar3778

https://github.com/tendermint/spec/tree/master/rfc

melekes:
> But since I love Tendermint I suggested it here.

❤️ 💙 💚 

> Note: Mempool can move to app layer for better implementation.

that's the plan

b00f:
> How would you see this fitting in with Tendermint/ABCI though?

No big changes in Tendermint. Alos creating blocks will be more simple. Block structure in Tendermint is a bit complicated. ABCI also can be untouched. In better implementation mempool moves to the app side and Application will handle Txs. 

For coinbase blockchain like Bitcoin, this is very easy to implement and very useful. Transactions individually can be checked if it is unspent.
For state based blocckhain like Tendermint, we need to find a ways to validate a block without executing all transactions from genesis block. Maybe the perfect solution is a Coinbase blockchain with Tendermint core.

I wanted to implement it in Berlin hackathon, but unfortunately my visa get rejected. So sad ;(

melekes:
@b00f would you mind to submit this as an RFC to https://github.com/tendermint/spec/tree/master/rfc?

As for app-level mempool, it's in the pipeline https://github.com/tendermint/tendermint/issues/3900.

b00f:
@melekes Please check here. 
https://github.com/tendermint/spec/pull/58

PR #3981, opened 2019-09-12
Add adr-049

b00f:
* [ ] Wrote tests

### DIFF
--- a/docs/architecture/adr-049-using-cid.md
+++ b/docs/architecture/adr-049-using-cid.md
@@ -1,0 +1,26 @@
+# ADR 049: Using IPFS CIDs instead of Hash
+
+## Changelog
+
+* 11-09-2019: Initial draft
+
+## Context
+
+Currently Tendermint is using hash bytes array for referring a block. However IPFS introduced CID for pointing materials.
+Using CIDs will help to integrate blocks and transaction easier with IPFS network. At the same time we can use IPFS to query transactions or blocks with their CIDs.
+
+
+## Decision
+
+## Status
+
+## Consequences
+
+
+### Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
+
+
+## References
+* [Content Identifiers (CIDs)](https://docs.ipfs.io/guides/concepts/cid/)


marbar3778:
Sorry for the delay in our response, this isnt an interest of Tendermint at the current time. There is a team working on extending vulcanizeDB to cosmos and tendermint based chains and they do support this. I do believe this is a better option for now. Will close this for now, but if you want to open an issue to hold some discussions, it may be better.

melekes:
This could be implemented as IPFS indexer, which saves transactions to IPFS (https://github.com/tendermint/tendermint/tree/master/state/txindex). I don't see any reason to switch to IPFS standard though. Are there any other benefits except that it would be easier to fetch blocks/transactions from IPFs?

b00f:
>  but if you want to open an issue to hold some discussions, it may be better.

Why not discussing here?

> Are there any other benefits except that it would be easier to fetch blocks/transactions from IPFs?

Yes. Not only txs, blocks also can be indexed in IPFS. Hash is just a raw dara, however CID has better format and since it has version and type it's more unique. CID is a better way to point any piece of information compare to hash.

melekes:
> Yes. Not only txs, blocks also can be indexed in IPFS. 

They can be indexed right now too, don't they? You just need to wrap the hash into CID (i.e. add version, etc.)

> Hash is just a raw dara, however CID has better format and since it has version and type it's more unique.

Hash is unique too.

> CID is a better way to point any piece of information compare to hash.

for multiple kinds of information - agree, but if you only have one type (blocks), then I still don't see a strong enough reason to use CID.

b00f:
Using CID might help to work with IPFS easier. (i.e. libp2p integration). I checked FileCoin project, They preferred using CID to [refer a block](https://github.com/filecoin-project/go-filecoin/blob/c266d961029599d090926f7c90490e9b7dcf5fd2/internal/pkg/block/block.go#L65)

melekes:
> i.e. libp2p integration

The only place where cid is being used in go-libp2p is https://github.com/libp2p/go-libp2p/blob/c73d2b9c8ee65326fe6c1b76a05a6b6cdcf223b9/p2p/host/relay/autorelay_test.go#L65, which is a mock object used for testing. So not sure how it helps.

> I checked FileCoin project, They preferred using CID to refer a block

Interesting. I guess for them the major reason was that files (blocks) are usually stored on IPFS

melekes:
https://github.com/tendermint/tendermint/pull/3981#issuecomment-537149969

PR #3988, opened 2019-09-16
get_tools = tools

marbar3778:
- in the makefile update `get_tools` was changed to just `tools`

* [ ] Wrote tests


DOCKER/Dockerfile.abci
```diff
@@ -15,7 +15,7 @@ RUN apt-get update && apt-get install -y \
 
 COPY Gopkg.toml /go/src/github.com/tendermint/abci/
 COPY Gopkg.lock /go/src/github.com/tendermint/abci/
```
marbar3778: this could use some updating 😆

ebuchman:
Can we link the original source for this change in the opening comment please?

cc @greg-szabo @mircea-c for awareness though they may already be aware

PR #3989, opened 2019-09-16
lite: light client MVP

melekes:
Refs #1771 

ADR: https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-044-lite-client-with-weak-subjectivity.md (**incomplete** - https://github.com/tendermint/tendermint/issues/2133)

* [ ] Wrote tests


types/validator_set.go
```diff
@@ -717,6 +715,71 @@ func (vals *ValidatorSet) VerifyFutureCommit(newSet *ValidatorSet, chainID strin
 	return nil
 }
 
+// VerifyCommitTrusting verifies that trustLevel ([1/3, 1]) of the validator
+// set signed this commit.
+// NOTE the given validators do not necessarily correspond to the validator set
+// for this commit, but there may be some intersection.
+func (vals *ValidatorSet) VerifyCommitTrusting(chainID string, blockID BlockID,
+	height int64, commit *Commit, trustLevel float) error {
```
golangcibot: undeclared name: `float` (from `typecheck`)

types/validator_set.go
```diff
@@ -717,6 +715,71 @@ func (vals *ValidatorSet) VerifyFutureCommit(newSet *ValidatorSet, chainID strin
...
+		return err
+	}
+	if height != commit.Height() {
+		return NewErrInvalidCommitHeight(height, commit.Height())
+	}
+	if !blockID.Equals(commit.BlockID) {
+		return fmt.Errorf("invalid commit -- wrong block ID: want %v, got %v",
+			blockID, commit.BlockID)
+	}
+}
```
golangcibot: missing return (from `typecheck`)

lite2/verifier.go
```diff
@@ -0,0 +1,102 @@
...
+	LastHeader *types.Header       // height H-1
+	Validators *types.ValidatorSet // height H
+}
+
+// Verifier is the core of the light client performing validation of the
+// headers.
+type Verifier struct {
+	chainID        string
+	trustingPeriod time.Duration
+	trustLevel     float
```
golangcibot: undeclared name: `float` (from `typecheck`)

lite2/verifier.go
```diff
@@ -0,0 +1,102 @@
...
+func (v *Verifier) expired() error {
+	expired := v.state.LastHeader.Time.Add(v.trustingPeriod)
+	if expired.Before(now) {
+		return errors.Errorf("last header expired at %v and too old to be trusted now %v. Verifier must be reset subjectively", expired, now)
+	}
+	return nil
+}
+
+func verifyNewHeaderAndVals(newHeader *types.SignedHeader, vals *types.ValidatorSet) error {
+	if err := newHeader.ValidateBasic(v.chainID); err != nil {
```
golangcibot: undeclared name: `v` (from `typecheck`)

lite2/verifier.go
```diff
@@ -0,0 +1,102 @@
...
+		return errors.Wrap(err, "newHeader.ValidateBasic failed")
```
golangcibot: Wrap not declared by package errors (from `typecheck`)

lite2/verifier.go
```diff
@@ -0,0 +1,137 @@
...
+			v.state.LastHeader.Time)
+	}
+
+	if !bytes.Equal(newHeader.ValidatorsHash, vals.Hash()) {
+		return fmt.Errorf("expected validators (%X) to match those from new header (%X)",
+			vals.Hash(),
+			newHeader.NextValidatorsHash,
+		)
+	}
+}
```
golangcibot: missing return (from `typecheck`)

lite2/verifier.go
```diff
@@ -0,0 +1,106 @@
...
+	newHeader *types.SignedHeader,
+	newVals *types.ValidatorSet,
+	now time.Time) error {
+
+	if err := verifyNewHeaderAndVals(chainID, newHeader, newVals, lastHeader, now); err != nil {
+		return err
+	}
+
+	// Ensure that +1/3 of last trusted validators signed correctly.
+	err := lastVals.VerifyCommitTrusting(chainID, newHeader.Commit.BlockID, newHeader.Height, newHeader.Commit, v.trustLevel)
```
golangcibot: undeclared name: `v` (from `typecheck`)

lite2/verifier.go
```diff
@@ -0,0 +1,106 @@
...
+
+	if !bytes.Equal(newHeader.ValidatorsHash, lastVals.Hash()) {
+		return errors.Errorf("expected our validators (%X) to match those from new header (%X)",
+			lastVals.Hash(),
+			newHeader.ValidatorsHash,
+		)
+	}
+
+	// Ensure that +2/3 of current validators signed correctly.
+	err := vals.VerifyCommit(chainID, newHeader.Commit.BlockID, newHeader.Height, newHeader.Commit)
```
golangcibot: undeclared name: `vals` (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,67 @@
...
+	return VerifyTrusting(c.chainID, c.state.Header, c.state.Vals, newHeader, newVals, now)
+}
+
+func (c *Client) trustedHeaderExpired(now time.Time) error {
+	expirationTime := c.state.Header.Time.Add(c.trustingPeriod)
+	return expirationTime.Before(now)
+}
+
+func (c *Client) canNewHeaderBeTrusted(now time.Time) error {
+	trustedHeaderExpirationTime := c.trustedHeader.Time.Add(c.trustingPeriod)
```
golangcibot: c.trustedHeader undefined (type *Client has no field or method trustedHeader) (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,67 @@
...
+	return newHeader.Time.Before(trustedHeaderExpirationTime)
```
golangcibot: undeclared name: `newHeader` (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if newHeader.Height == c.state.Header.Height+1 {
+		return Verify(c.chainID, c.state.Header, c.state.Vals, newHeader, newVals, now)
+	}
+
+	return VerifyTrusting(c.chainID, c.state.Header, c.state.Vals, newHeader, newVals, now)
+}
+
+func (c *Client) trustedHeaderExpired(now time.Time) error {
+	expirationTime := c.state.Header.Time.Add(c.trustingPeriod)
```
golangcibot: c.state undefined (type *Client has no field or method state) (from `typecheck`)

lite2/verifier.go
```diff
@@ -0,0 +1,84 @@
...
+	lastVals *types.ValidatorSet,
+	newHeader *types.SignedHeader,
+	newVals *types.ValidatorSet,
+	now time.Time) error {
+
+	if err := verifyNewHeaderAndVals(chainID, newHeader, newVals, lastHeader, now); err != nil {
+		return err
+	}
+
+	if newHeader.Height == c.state.Header.Height+1 {
```
golangcibot: undeclared name: `c` (from `typecheck`)

lite2/verifier.go
```diff
@@ -0,0 +1,84 @@
...
+		if !bytes.Equal(newHeader.ValidatorsHash, lastVals.Hash()) {
+			return errors.Errorf("expected our validators (%X) to match those from new header (%X)",
+				lastVals.Hash(),
+				newHeader.ValidatorsHash,
+			)
+		}
+	} else {
+		// Ensure that +1/3 of last trusted validators signed correctly.
+		err := lastVals.VerifyCommitTrusting(chainID, newHeader.Commit.BlockID, newHeader.Height, newHeader.Commit, v.trustLevel)
```
golangcibot: undeclared name: `v` (from `typecheck`)

lite2/verifier.go
```diff
@@ -0,0 +1,84 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	// Ensure that +2/3 of current validators signed correctly.
+	err = newVals.VerifyCommit(v.chainID, newHeader.Commit.BlockID, newHeader.Height, newHeader.Commit)
```
golangcibot: undeclared name: `err` (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (c *Client) bisection(lastHeader *types.SignedHeader,
+	lastVals *types.ValidatorSet,
+	newHeader *types.SignedHeader,
+	newVals *types.ValidatorSet,
+	now time.Time) error {
+
+	err := Verify(c.chainID, lastHeader, lastVals, newHeader, newVals, now)
+	if err != nil && IsErrTooMuchChange(err) {
```
golangcibot: undeclared name: `IsErrTooMuchChange` (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,62 @@
...
+		return err
+	}
+
+	if newHeader.Height == v.state.LastHeader.Height+1 {
+		// TODO: submit evidence here
+		return errors.New("adjacent headers that are not matching")
+	}
+
+	pivot := (v.state.LastHeader.Height + newHeader.Header.Height) / 2
+	pivotHeader := c.signedHeader(pivot)
```
golangcibot: c.signedHeader undefined (type *Client has no field or method signedHeader) (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,62 @@
...
+
+	c.store(pivotHeader)
```
golangcibot: c.store undefined (type *Client has no field or method store) (from `typecheck`)

lite2/verifier.go
```diff
@@ -0,0 +1,84 @@
...
+	} else {
+		// Ensure that +1/3 of last trusted validators signed correctly.
+		err := lastVals.VerifyCommitTrusting(chainID, newHeader.Commit.BlockID, newHeader.Height, newHeader.Commit, v.trustLevel)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Ensure that +2/3 of current validators signed correctly.
+	err = newVals.VerifyCommit(chainID, newHeader.Commit.BlockID, newHeader.Height, newHeader.Commit)
```
golangcibot: undeclared name: `err` (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,83 @@
...
+
+	err := Verify(c.chainID, lastHeader, lastVals, newHeader, newVals, now)
+	switch {
+	case err == nil:
+		if !c.newHeaderWithinTrustingPeriod(newHeader) {
+			// TODO: continue bisection
+			// if adjused headers, fail?
+		}
+		return nil
+	case err != nil && !IsErrTooMuchChange(err):
```
golangcibot: undeclared name: `IsErrTooMuchChange` (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,83 @@
...
+		// TODO: submit evidence here
+		return errors.New("adjacent headers that are not matching")
+	}
+
+	pivot := (v.state.LastHeader.Height + newHeader.Header.Height) / 2
+	pivotHeader := c.signedHeader(pivot)
+
+	c.storeSignedHeader(pivotHeader)
+
+	if err := c.bisection(lastHeader, lastVals, pivotHeader, pivotVals, now); err != nil {
```
golangcibot: undeclared name: `pivotVals` (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if c.trustedHeaderExpired(now) {
+		return erros.New("trusted header expired. reset light client subjectvely")
+	}
+
+	return c.bisection(c.trustedHeader, c.trustedVals, newHeader, newVals, now)
+}
+
+func (c *Client) trustedHeaderExpired(now time.Time) bool {
+	expirationTime := c.state.Header.Time.Add(c.trustingPeriod)
```
golangcibot: c.state undefined (type *Client has no field or method state) (from `typecheck`)

lite2/verifier_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	now := time.Now()
+	trustingPeriod := 1 * time.Hour
+	for i, tc := range testCases {
+		t.Run(fmt.Sprintf("#%d", i), func(t *testing.T) {
+			err := Verify(chainID, header, vals, tc.newHeader, tc.newVals, trustingPeriod, now, 0)
+
+			if tc.expErr == nil {
+				assert.NoError(t, err)
+			} else {
```
golangcibot: elseif: can replace 'else {if cond {}}' with 'else if cond {}' (from `gocritic`)

lite2/verifier_test.go
```diff
@@ -0,0 +1,102 @@
...
+			keys.ToValidators(10, 1),
+			nil,
+		},
+	}
+
+	now := time.Now()
+	trustingPeriod := 1 * time.Hour
+	for i, tc := range testCases {
+		t.Run(fmt.Sprintf("#%d", i), func(t *testing.T) {
+			err := Verify(chainID, header, vals, tc.newHeader, tc.newVals, trustingPeriod, now, 0)
```
golangcibot: Using the variable on range scope `tc` in function literal (from `scopelint`)

lite2/verifier_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+			if tc.expErr == nil {
```
golangcibot: Using the variable on range scope `tc` in function literal (from `scopelint`)

lite2/verifier_test.go
```diff
@@ -0,0 +1,102 @@
...
+				assert.NoError(t, err)
+			} else {
+				if assert.Error(t, err) {
+					assert.Equal(t, tc.expErr, err)
```
golangcibot: Using the variable on range scope `tc` in function literal (from `scopelint`)

lite2/verifier_test.go
```diff
@@ -0,0 +1,135 @@
...
+			keys.ToValidators(10, 1),
+			3 * time.Hour,
+			bTime.Add(2 * time.Hour), // not relevant
+			nil,
+		},
+	}
+
+	for i, tc := range testCases {
+		t.Run(fmt.Sprintf("#%d", i), func(t *testing.T) {
+			err := Verify(chainID, header, vals, tc.newHeader, tc.newVals, tc.trustingPeriod, tc.now, 0)
```
golangcibot: Using the variable on range scope `tc` in function literal (from `scopelint`)

lite2/verifier_test.go
```diff
@@ -0,0 +1,146 @@
...
+			"to be within the trusting period",
+		},
+	}
+
+	for i, tc := range testCases {
+		t.Run(fmt.Sprintf("#%d", i), func(t *testing.T) {
+			tc := tc
+			err := Verify(chainID, header, vals, tc.newHeader, tc.newVals, tc.trustingPeriod, tc.now, 0)
+
+			if tc.expErr != nil && assert.Error(t, err) {
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

lite2/verifier_test.go
```diff
@@ -0,0 +1,146 @@
...
+			3 * time.Hour,
+			bTime.Add(2 * time.Hour), // not relevant
+			nil,
+			"to be within the trusting period",
+		},
+	}
+
+	for i, tc := range testCases {
+		t.Run(fmt.Sprintf("#%d", i), func(t *testing.T) {
+			tc := tc
```
golangcibot: Using the variable on range scope `tc` in function literal (from `scopelint`)

lite2/client.go
```diff
@@ -0,0 +1,246 @@
...
+	return c.bisection(c.trustedHeader, c.trustedVals, newHeader, newVals, now)
+}
+
+func (c *Client) bisection(lastHeader *types.SignedHeader,
+	lastVals *types.ValidatorSet,
+	newHeader *types.SignedHeader,
+	newVals *types.ValidatorSet,
+	now time.Time) error {
+
+	err := Verify(c.chainID, lastHeader, lastVals, newHeader, newVals, c.trustOptions.Period, now, c.trustLevel)
```
golangcibot: c.trustOptions undefined (type *Client has no field or method trustOptions) (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,246 @@
...
+	newHeader, err := c.primary.SignedHeader(height)
+	if err != nil {
+		return err
+	}
+	newVals, err := c.primary.ValidatorSet(height)
+	if err != nil {
+		return err
+	}
+
+	c.VerifyHeader(c.trustedHeader, c.trustedVals, newHeader, newVals, now)
```
golangcibot: cannot use newHeader (variable of type *types.SignedHeader) as time.Time value in argument to c.VerifyHeader (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,246 @@
...
+}
```
golangcibot: missing return (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,246 @@
...
+		c.alternatives = providers
+	}
+}
+
+type Client struct {
+	chainID            string
+	trustingPeriod     time.Duration
+	mode               mode
+	trustLevel         float32
+	lastVerifiedHeight int64
```
golangcibot: `lastVerifiedHeight` is unused (from `structcheck`)

lite2/client.go
```diff
@@ -0,0 +1,267 @@
...
+
+	if err := c.bisection(lastHeader, lastVals, pivotHeader, pivotVals, now); err != nil {
+		return c.bisection(pivotHeader, pivotVals, newHeader, newVals, now)
+	}
+
+	return errors.New("bisection failed. restart with different full-node?")
+}
+
+func (c *Client) saveTrustedHeaderAndVals(h *types.SignedHeader, vals *types.ValidatorSet) error {
+	err = c.trustedStore.SaveSignedHeader(h)
```
golangcibot: undeclared name: `err` (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,267 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to save trusted header")
+	}
+	err = c.trustedStore.SaveValidatorSet(vals, options.Height+1)
```
golangcibot: undeclared name: `options` (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,267 @@
...
+		return err
+	}
+	err = c.saveTrustedHeaderAndVals(newHeader, nextVals)
+	if err != nil {
+		return err
+	}
+
+	c.trustedHeader = newHeader
+	c.trustedVals = nextVals
+}
```
golangcibot: missing return (from `typecheck`)

lite2/client.go
```diff
@@ -0,0 +1,308 @@
...
+
+	err := c.initializeWithTrustOptions(trustOptions)
+	if err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
+
+func (c *Client) crossCheckPrimary() {
```
golangcibot: func `(*Client).crossCheckPrimary` is unused (from `unused`)

lite2/client.go
```diff
@@ -0,0 +1,304 @@
...
+	bisecting
+)
+
+// Option sets a parameter for the light client.
+type Option func(*Client)
+
+// SequentialVerification option can be used to instruct Verifier to
+// sequentially check the headers. Note this is much slower than
+// BisectingVerification, albeit more secure.
+func SequentialVerification() Option {
```
melekes: Do we even need this option?

lite2/client.go
```diff
@@ -0,0 +1,304 @@
...
+
+	return c, nil
+}
+
+// TODO: cross check primary with alternatives providers
+// func (c *Client) crossCheckPrimary() {
+// }
+
+func (c *Client) initializeWithTrustOptions(options TrustOptions) error {
+	h, err := c.primary.SignedHeader(options.Height)
```
melekes: TODO: check if we're ahead of options.Height. If so, revert?

lite2/client.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// SetLogger sets a logger.
+func (c *Client) SetLogger(l log.Logger) {
+	c.logger = l
+}
+
+func (c *Client) VerifyHeaderAtHeight(height int64, now time.Time) error {
+	// Request the header and the vals.
+	newHeader, err := c.primary.SignedHeader(height)
```
melekes: TODO: check if height is already verified/trusted.

zmanian:
Proposal: In verification, we call it `skipping` rather than bisecting. Bisecting is only referred to when fetching headers.

lite2/verifier.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	return nil
+}
+
+// ValidateTrustLevel checks that trustLevel is within the allowed range [1/3,
+// 1]. If not, it returns an error. 1/3 is the minimum amount of trust needed
+// which does not break the security model.
+func ValidateTrustLevel(lvl float32) error {
+	if lvl > 1 || lvl < float32(1)/float32(3) {
```
Shivani912: Shouldn't this be 1/3 < lvl < 1 ?
```suggestion
	if lvl < 1 || lvl > float32(1)/float32(3) {
```

melekes: not if we panic

Shivani912: Right right. Dont know what i was thinking.

lite2/client.go
```diff
@@ -0,0 +1,362 @@
...
+// header exist. It returns an error if there are some issues with the trusted
+// store, although that should not happen normally. TODO mention how many
+// headers will be kept by the light client.
+func (c *Client) TrustedHeader(height int64) (*types.SignedHeader, error) {
+	return c.trustedStore.SignedHeader(height)
+}
+
+// LastTrustedHeight returns a last trusted height.
+func (c *Client) LastTrustedHeight() (int64, error) {
+	h, err := c.trustedStore.LastSignedHeader()
```
Shivani912: Where is LastSignedHeader() defined?

lite2/client.go
```diff
@@ -0,0 +1,369 @@
...
+// TrustedHeader returns a trusted header at the given height or nil if no such
+// header exist. It returns an error if there are some issues with the trusted
+// store, although that should not happen normally. TODO mention how many
+// headers will be kept by the light client.
+//
+// 0 - the latest.
+// height must be >= 0.
+func (c *Client) TrustedHeader(height int64) (*types.SignedHeader, error) {
+	if height < 0 {
+		return errors.New("negative height")
```
Shivani912:
```suggestion
		return nil, errors.New("negative height")
```

lite2/client.go
melekes: In the `Store`

lite2/client.go
melekes: Thanks!

lite2/client.go
Shivani912: yep. Got the update!

lite2/client.go
```diff
@@ -0,0 +1,374 @@
...
+		err = c.sequence(newHeader, newVals, now)
+	case skipping:
+		err = c.bisection(c.trustedHeader, c.trustedNextVals, newHeader, newVals, now)
+	}
+	if err != nil {
+		return err
+	}
+
+	// Update trusted header and vals.
+	nextVals, err := c.primary.ValidatorSet(newHeader.Height + 1)
```
cwgoes: Does this function verify against the trusted validator set hash? Below it looks like we call it and then verify.

I think we do need to verify it here, right? Does `updateTrustedHeaderAndVals` do that?

I think functions which verify should include "verify" in their names.

lite2/client.go
```diff
@@ -0,0 +1,374 @@
...
+		// Update trusted header and vals.
+		if height == newHeader.Height-1 {
+			nextVals = newVals
+		} else {
+			nextVals, err = c.primary.ValidatorSet(height + 1)
+			if err != nil {
+				return errors.Wrapf(err, "failed to obtain the vals #%d", height+1)
+			}
+		}
+		if !bytes.Equal(interimHeader.NextValidatorsHash, nextVals.Hash()) {
```
cwgoes: ... see here we verify the next validator set hash explicitly, and then call `updateTrustedHeaderAndVals`.

lite2/client.go
```diff
@@ -0,0 +1,374 @@
...
+	switch err.(type) {
+	case nil:
+		return nil
+	case types.ErrTooMuchChange:
+		// continue bisection
+	default:
+		return errors.Wrapf(err, "failed to verify the header #%d ", newHeader.Height)
+	}
+
+	if newHeader.Height == c.trustedHeader.Height+1 {
```
cwgoes: I don't understand this. There *can* be more than 1/3 change between subsequent headers, but since we have the `NextValidatorSetHash` (signed over by the previous validator set), we can verify the subsequent header.

lite2/client.go
```diff
@@ -0,0 +1,374 @@
...
+		// TODO: submit evidence here
```
cwgoes: Evidence of what?

lite2/client.go
```diff
@@ -0,0 +1,374 @@
...
+		return errors.Errorf("adjacent headers (#%d and #%d) that are not matching", lastHeader.Height, newHeader.Height)
+	}
+
+	pivot := (c.trustedHeader.Height + newHeader.Header.Height) / 2
```
cwgoes: This will do floor division, right?

lite2/client.go
```diff
@@ -0,0 +1,374 @@
...
+	pivotHeader, pivotVals, err := c.fetchHeaderAndValsAtHeight(pivot)
+	if err != nil {
+		return err
+	}
+
+	if err := c.bisection(lastHeader, lastVals, pivotHeader, pivotVals, now); err == nil {
```
cwgoes: This is a confusing pattern and seems unlikely to be optimised by the compiler. Can we just do a standard tail call?

lite2/client.go
```diff
@@ -0,0 +1,374 @@
...
+		if !bytes.Equal(pivotHeader.NextValidatorsHash, nextVals.Hash()) {
+			return errors.Errorf("expected next validator's hash %X, but got %X (height #%d)",
+				pivotHeader.NextValidatorsHash,
+				nextVals.Hash(),
+				pivot)
+		}
+		return c.bisection(pivotHeader, nextVals, newHeader, newVals, now)
+	}
+
+	return errors.New("bisection failed. Restart with different full-node?")
```
cwgoes: Why would it have failed here?

lite2/client.go
```diff
@@ -0,0 +1,374 @@
...
+}
+
+func (c *Client) updateTrustedHeaderAndVals(h *types.SignedHeader, vals *types.ValidatorSet) error {
```
cwgoes: IMO this function should verify the validator set against the header's `NextValSetHash`, unless there is a reason we would ever not want to do that.

lite2/verifier.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	if h2.Height <= h1.Height {
+		return errors.Errorf("expected new header height %d to be greater than one of old header %d",
+			h2.Height,
+			h1.Height)
+	}
+
+	if !h2.Time.After(h1.Time) {
+		return errors.Errorf("expected new header time %v to be after old header time %v",
```
cwgoes: To confirm, do we enforce this in Tendermint?

lite2/verifier.go
```diff
@@ -0,0 +1,113 @@
...
+			h2.Time,
+			h1.Time)
+	}
+
+	if !h2.Time.Before(now) {
```
cwgoes: There could be false positives here due to local clock skew, just something to note.

lite2/verifier.go
```diff
@@ -0,0 +1,113 @@
...
+	} else {
+		// Ensure that +1/3 or more of last trusted validators signed correctly.
+		err := h1NextVals.VerifyCommitTrusting(chainID, h2.Commit.BlockID, h2.Height, h2.Commit, trustLevel)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Ensure that +2/3 of new validators signed correctly.
+	err := h2Vals.VerifyCommit(chainID, h2.Commit.BlockID, h2.Height, h2.Commit)
```
cwgoes: The light client may have reason to require more than 2/3 here. That can probably happen later though.

lite2/verifier.go
```diff
@@ -0,0 +1,113 @@
...
+
+	if h2.Height == h1.Height+1 {
+		if !bytes.Equal(h2.ValidatorsHash, h1NextVals.Hash()) {
+			return errors.Errorf("expected our validators (%X) to match those from new header (%X)",
+				h1NextVals.Hash(),
+				h2.ValidatorsHash,
+			)
+		}
+	} else {
+		// Ensure that +1/3 or more of last trusted validators signed correctly.
```
cwgoes:
```suggestion
		// Ensure that +`trustLevel` (default 1/3) or more of last trusted validators signed correctly.
```

types/validator_set.go
```diff
@@ -626,10 +624,11 @@ func (vals *ValidatorSet) VerifyCommit(chainID string, blockID BlockID, height i
 		// }
 	}
 
-	if talliedVotingPower > vals.TotalVotingPower()*2/3 {
-		return nil
+	if got, needed := float32(talliedVotingPower), float32(vals.TotalVotingPower()*2/3); got <= needed {
```
cwgoes: `float32` conversion seems like a recipe for disaster. Can we multiply by integers instead?

cwgoes: ... especially in a chain through IBC, I don't think we can do this, since it might be machine-dependent.

types/validator_set.go
```diff
@@ -711,27 +710,97 @@ func (vals *ValidatorSet) VerifyFutureCommit(newSet *ValidatorSet, chainID strin
...
+				talliedVotingPower += val.VotingPower
+			}
+			// else {
+			// It's OK that the BlockID doesn't match.  We include stray
+			// precommits to measure validator availability.
+			// }
+		}
+	}
+
+	if got, needed := float32(talliedVotingPower), float32(vals.TotalVotingPower())*trustLevel; got <= needed {
```
cwgoes: Same concern about `float32` here.

types/validator_set.go
```diff
@@ -711,27 +710,97 @@ func (vals *ValidatorSet) VerifyFutureCommit(newSet *ValidatorSet, chainID strin
...
+		return ErrTooMuchChange{Got: got, Needed: needed}
+	}
+	return nil
+}
+
+// VerifyCommitTrusting verifies that trustLevel ([1/3, 1]) of the validator
+// set signed this commit.
+// NOTE the given validators do not necessarily correspond to the validator set
+// for this commit, but there may be some intersection.
+func (vals *ValidatorSet) VerifyCommitTrusting(chainID string, blockID BlockID,
```
cwgoes: Can we combine this and the above function? They seem to duplicate a lot of code.

lite2/client.go
```diff
@@ -0,0 +1,374 @@
...
+		c.mode = sequential
+	}
+}
+
+// SkippingVerification option configures the light client to skip headers as
+// long as {trustLevel} of the old validator set signed the new header. The
+// bisection algorithm from the specification is used for finding the minimal
+// "trust path".
+//
+// trustLevel - maximum change between two not consecutive headers in terms of
```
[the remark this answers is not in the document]
melekes: mmm... faults have nothing to do with trust assumptions (I mean validator set can change because of faults, but not always).

> Given that this is the fraction of faults we will tolerate, not the fraction of the validator set we "trust" (assume is correct)

I think it's neither of those. It's the fraction of the trusted old validator set, which has signed the new commit, required to trust it. I think the existing name captures this pretty well. Although maybe we should rename to `trustThreshold`.

lite2/client.go
```diff
@@ -0,0 +1,374 @@
...
+	if !bytes.Equal(h.Hash(), options.Hash) {
+		return errors.Errorf("expected header's hash %X, but got %X", options.Hash, h.Hash())
+	}
+
+	// 2) Fetch and verify the next vals.
+	vals, err := c.primary.ValidatorSet(options.Height + 1)
+	if err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
melekes: My intuition was that the check will be performed in the background, inside a goroutine constantly verifying newly trusted headers AFTER they've been added to the trust store. 

Better to separate the actual verifying logic and checking with secondary providers imho

lite2/client.go
melekes:
```
  if h2.Header.height == h1.Header.height + 1 {
    // we have adjacent headers that are not matching (failed
    // the CheckSupport)
    // we could submit evidence here
    return false
  }
```

from the spec

melekes: > There can be more than 1/3 change between subsequent headers, but since we have the NextValidatorSetHash (signed over by the previous validator set), we can verify the subsequent header.

if > 2/3 of new vals signed h2 and h2.ValidatorHash == h1NextVals, then Verify will return nil (no error). See Verify

lite2/client.go
melekes: misbehavior? when a primary provider sends us an invalid header

lite2/client.go
melekes: Right

lite2/client.go
melekes: https://stackoverflow.com/a/12116644

afaik Go only does tail call optimization in rare cases

melekes: see 11e1e9015

lite2/client.go
melekes: removed in 11e1e9015

lite2/verifier.go
melekes: https://github.com/tendermint/tendermint/blob/98c595312af02037843b8fe74f0ee0625665448e/state/validation.go#L108

yes

types/validator_set.go
melekes: Yeah, but we can't multiply here because trustLevel is float32. We can probably add a new type 

```go
// Represents a fraction in terms of a numerator divided by a denominator. 
type Fraction struct {
	// The portion of the denominator in the faction, e.g. 2 in 2/3.
	Numerator in64
	// The value by which the numerator is divided, e.g. 3 in 2/3. Must be
	// positive.
	Denominator in64
}
```

types/validator_set.go
melekes: In a separate PR. I am not going to touch old lite client in this PR.

types/validator_set.go
cwgoes: I don't think `trustLevel` should be `float32` either. A `Fraction` type sounds OK, or I think there is a fixed-precision `Decimal` in the Cosmos SDK that we could use.

lite2/client.go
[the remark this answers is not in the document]
cwgoes: It's not 'the fraction of the trusted old validator set', since it's the maximum change (so it's `1 - the fraction of the trusted old validator set`, as it were), right?

lite2/client.go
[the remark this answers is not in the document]
cwgoes: OK!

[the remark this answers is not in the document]
cwgoes: We should add a comment to that effect though.

lite2/client.go
cwgoes: Ah I see, OK then!

cwgoes: Ah I see, you are right, that is the control flow.

lite2/client.go
cwgoes: But this would be `TooMuchChange` between two subsequent headers, right?

I'm still not sure when this case would actually arise. Header `h + 1` is either valid with respect to `h` or it is not, there is no notion of `TooMuchChange` in that case, since we have the `NextValSetHash`.

lite2/client.go
cwgoes: I still think this control flow could be clearer. Up to you though.

types/validator_set.go
cwgoes: OK.

lite2/client.go
melekes: good catch!

lite2/client.go
melekes: > I still think this control flow could be clearer.

if you think it can be improved, please do so! you can commit to this branch

lite2/client.go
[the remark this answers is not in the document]
melekes: updated comment to be `fraction of the old validator set (in terms of voting power),
// which must sign the new header in order for us to trust it.`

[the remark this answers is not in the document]
cwgoes: Great - thanks.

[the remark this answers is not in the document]
cwgoes: But I don't think that's what this is - it's `1/3`, not `2/3`, concretely.

See https://github.com/tendermint/tendermint/pull/3989/files#diff-f6c374f81d77f2ae4041a970bc378b89R16

Whereas the "fraction of the old validator set (in terms of voting power) which must sign the new header in order for us to trust it" is `2/3`.

lite2/client.go
melekes: > there is no notion of TooMuchChange in that case

isn't there? what if new validator set (suppose `NextValSetHash` matches and we trust it) did not sign the commit (i.e. < 2/3 signed). `VerifyCommit` will return `ErrTooMuchChange` in that case. I agree this is confusing and we should probably return different error, say `ErrNotEnoughValsSigned` in `VerifyCommit`. What do you think?

cwgoes: Ah I see - yes, I think that should be a different error then.

lite2/client.go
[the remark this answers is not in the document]
melekes:
```
      // get validators in h1 that signed h2
      // sum of voting powers in h1 of
      // validators that signed h2
      // is more than a third in h1
```

from the spec ^^

Are you trying to say we should stress out that `trustLevel` is only used for non-consecutive headers?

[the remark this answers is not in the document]
cwgoes: Right, you do say `non-consecutive`, I just think clearer wording would help - the idea (as I understand it) is that we assume that `> 2/3` are correct, so that if `> 1/3` sign a new header at least one of the signers will have been in the correct set.

melekes:
TODO:

- [ ] cross-checking primary using alternative providers
- [ ] how many headers to store? how much is enough for evidence purposes?
- [ ] trust a header only after receiving N headers? how many? default to immediate (N=0) trust?
- [ ] evidence handling and submission
- [ ] more docs & tests

lite2/rpc/client.go
```diff
@@ -0,0 +1,416 @@
...
+}
+
+func (c *Client) RegisterOpDecoder(typ string, dec merkle.OpDecoder) {
+	c.prt.RegisterOpDecoder(typ, dec)
+}
+
+// SubscribeWS subscribes for events using the given query and remote address as
+// a subscriber, but does not verify responses (UNSAFE)!
+// TODO: verify data
+func (c *Client) SubscribeWS(ctx *rpctypes.Context, query string) (*ctypes.ResultSubscribe, error) {
```
fedekunze: ditto. do we need to return the result if it's `nil`?

lite2/rpc/client.go
```diff
@@ -0,0 +1,416 @@
...
+	err := c.next.Unsubscribe(context.Background(), ctx.RemoteAddr(), query)
+	if err != nil {
+		return nil, err
+	}
+	return &ctypes.ResultUnsubscribe{}, nil
+}
+
+// UnsubscribeAllWS calls original client's UnsubscribeAll using remote address
+// as a subscriber.
+func (c *Client) UnsubscribeAllWS(ctx *rpctypes.Context) (*ctypes.ResultUnsubscribe, error) {
```
fedekunze: ditto

lite2/rpc/client.go
```diff
@@ -0,0 +1,416 @@
...
+
+func parseQueryStorePath(path string) (storeName string, err error) {
+	if !strings.HasPrefix(path, "/") {
+		return "", fmt.Errorf("expected path to start with /")
+	}
+
+	paths := strings.SplitN(path[1:], "/", 3)
+	switch {
+	case len(paths) != 3:
+		return "", fmt.Errorf("expected format like /store/<storeName>/key")
```
fedekunze: you can use `errors.New` if you're not formating the error

lite2/rpc/client.go
```diff
@@ -0,0 +1,416 @@
...
+	case paths[0] != "store":
+		return "", fmt.Errorf("expected format like /store/<storeName>/key")
```
fedekunze: ditto

lite2/rpc/client.go
```diff
@@ -0,0 +1,416 @@
...
+	case paths[2] != "key":
+		return "", fmt.Errorf("expected format like /store/<storeName>/key")
```
fedekunze: ditto

lite2/store/db/db.go
```diff
@@ -0,0 +1,132 @@
...
+
+// New returns a Store that wraps any DB (with an optional prefix in case you
+// want to use one DB with many light clients).
+func New(db dbm.DB, prefix string) store.Store {
+	cdc := amino.NewCodec()
+	cryptoAmino.RegisterAmino(cdc)
+	return &dbs{db: db, prefix: prefix, cdc: cdc}
+}
+
+func (s *dbs) SaveSignedHeader(sh *types.SignedHeader) error {
```
fedekunze: nit, missing godocs on the public funcs in this file

lite2/verifier.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/tendermint/tendermint/types"
+)
+
+var (
+	// DefaultTrustLevel - new header can be trusted if at least one correct old
+	// validator signed it.
+	DefaultTrustLevel = cmn.Fraction{Numerator: 1, Denominator: 3}
+)
+
+func Verify(
```
fedekunze: missing godoc

lite2/verifier.go
```diff
@@ -0,0 +1,115 @@
...
+		return ErrOldHeaderExpired{expirationTime, now}
+	}
+
+	if err := verifyNewHeaderAndVals(chainID, h2, h2Vals, h1, now); err != nil {
+		return err
+	}
+
+	if h2.Height == h1.Height+1 {
+		if !bytes.Equal(h2.ValidatorsHash, h1NextVals.Hash()) {
+			return errors.Errorf("expected our validators (%X) to match those from new header (%X)",
```
fedekunze:
```suggestion
			return errors.Errorf("expected header1 validators (%X) to match those from new header2 (%X)",
```

lite2/verifier.go
```diff
@@ -0,0 +1,115 @@
...
+		)
+	}
+
+	return nil
+}
+
+// ValidateTrustLevel checks that trustLevel is within the allowed range [1/3,
+// 1]. If not, it returns an error. 1/3 is the minimum amount of trust needed
+// which does not break the security model.
+func ValidateTrustLevel(lvl cmn.Fraction) error {
```
fedekunze: see prev comment. Consider moving
```go
func (fr Fraction) ValidateBasic() {
...
}
```

lite2/verifier.go
```diff
@@ -0,0 +1,115 @@
...
+	if lvl.Numerator*3 < lvl.Denominator || // < 1/3
```
fedekunze: NOTE: this returns `nil` if 0/0 is provided as input param

lite2/verifier_test.go
```diff
@@ -0,0 +1,277 @@
...
+				assert.Contains(t, err.Error(), tc.expErrText)
+			default:
+				assert.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestValidateTrustLevel(t *testing.T) {
+	testCases := []struct {
```
fedekunze: missing `0/0` and other `0` denominator cases

lite2/auto_client.go
```diff
@@ -0,0 +1,74 @@
...
+	err            chan error
+}
+
+// NewAutoClient creates a new client and starts a polling goroutine.
+func NewAutoClient(base *Client, updatePeriod time.Duration) *AutoClient {
+	c := &AutoClient{
+		base:           base,
+		updatePeriod:   updatePeriod,
+		quit:           make(chan struct{}),
+		trustedHeaders: make(chan *types.SignedHeader),
```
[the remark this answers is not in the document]
melekes: Is that what you do in cosmos-sdk? I always thought plural name is more appropriate because when you iterate over it looks like:

```
for h := trustedHeaders {

}
```

lite2/client.go
```diff
@@ -0,0 +1,387 @@
...
+	//
+	// More specifically, trusting period + time needed to check headers + time
+	// needed to report and punish misbehavior should be less than the unbonding
+	// period.
+	Period time.Duration
+
+	// Header's Height and Hash must both be provided to force the trusting of a
+	// particular header.
+	Height int64
+	Hash   []byte
```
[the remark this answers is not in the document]
melekes: `Header's Height and Hash` from the above comment

lite2/client.go
```diff
@@ -0,0 +1,387 @@
...
+// SkippingVerification option configures the light client to skip headers as
+// long as {trustLevel} of the old validator set signed the new header. The
+// bisection algorithm from the specification is used for finding the minimal
+// "trust path".
+//
+// trustLevel - fraction of the old validator set (in terms of voting power),
+// which must sign the new header in order for us to trust it. NOTE this only
+// applies to non-adjusted headers. For adjusted headers, sequential
+// verification is used.
+func SkippingVerification(trustLevel cmn.Fraction) Option {
```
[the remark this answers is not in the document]
melekes: last one wins :) i think it's okay

lite2/client.go
```diff
@@ -0,0 +1,387 @@
...
+	if err := ValidateTrustLevel(trustLevel); err != nil {
```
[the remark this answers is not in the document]
melekes: no. Fraction is not limited (can be -1/2). trust level is (1/3 - 1)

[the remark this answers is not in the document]
melekes: ValidateBasic could contain only a basic checks like Denominator != 0

lite2/client.go
```diff
@@ -0,0 +1,387 @@
...
+
+	if err := c.initializeWithTrustOptions(trustOptions); err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
+
+func (c *Client) initializeWithTrustOptions(options TrustOptions) error {
+	// 1) Fetch and verify the header.
```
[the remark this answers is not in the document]
melekes: I'd prefer them closer to the code (especially when func is private)

lite2/rpc/client.go
```diff
@@ -0,0 +1,416 @@
...
+				// We should have a switch here that performs a validation
+				// depending on the event's type.
+				ctx.WSConn.TryWriteRPCResponse(
+					rpctypes.NewRPCSuccessResponse(
+						ctx.WSConn.Codec(),
+						rpctypes.JSONRPCStringID(fmt.Sprintf("%v#event", ctx.JSONReq.ID)),
+						resultEvent,
+					))
+			case <-c.Quit():
+				return
```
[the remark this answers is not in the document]
melekes: it's a goroutine, not the main func

lite2/rpc/client.go
```diff
@@ -0,0 +1,416 @@
...
+			}
+		}
+	}()
+
+	return &ctypes.ResultSubscribe{}, nil
+}
+
+// UnsubscribeWS calls original client's Unsubscribe using remote address as a
+// subscriber.
+func (c *Client) UnsubscribeWS(ctx *rpctypes.Context, query string) (*ctypes.ResultUnsubscribe, error) {
```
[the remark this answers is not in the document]
melekes: yes

lite2/rpc/client.go
melekes: yes

lite2/verifier.go
```diff
@@ -0,0 +1,116 @@
...
+	// Ensure that +2/3 of new validators signed correctly.
+	err := h2Vals.VerifyCommit(chainID, h2.Commit.BlockID, h2.Height, h2.Commit)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func verifyNewHeaderAndVals(
```
Shivani912: We might want to check if the correct proposer committed the block. Currently, It passes this function with wrong proposer from the same validator set as it only compares the hash.

lite2/verifier.go
marbar3778: can there be a case where a private chain would want to set the lite client to trust 0?

lite2/verifier.go
melekes: But we don't know who should be the proposer?

lite2/client.go
cwgoes: It's better now!

cwgoes:
> TODO:
> 
> * [ ]  cross-checking primary using alternative providers
> * [ ]  how many headers to store? how much is enough for evidence purposes?
> * [ ]  trust a header only after receiving N headers? how many? default to immediate (N=0) trust?
> * [ ]  evidence handling and submission (what types of evidence we will have?)
> * [ ]  more docs & tests

Could we split out cross-checking and evidence handling to a follow-up PR?

I'd like to start integrating this on the IBC side - for which we mostly just need the `Verify` function - and evidence needs to be handled differently there as well. As soon as we merge this we can start on the ICS 2 integration in the SDK.

melekes:
> Could we split out cross-checking and evidence handling to a follow-up PR?

Absolutely. I can merge this PR tomorrow if someone approves.

marbar3778:
could you add a feature flag in the config for it to be used? then we can merge

melekes:
> could you add a feature flag in the config for it to be used?

it's not yet ready for general usage imo

cwgoes:
> > could you add a feature flag in the config for it to be used?
> 
> it's not yet ready for general usage imo

To be fair, neither is the current light client (which is not secure).

lite2/verifier.go
Shivani912: Well, we don't have the state but we do have next validator set from previous height!

Shivani912:
I feel it'll be great if we can have error types everywhere instead of error strings. That will help standardize error handling and so we can use the same for Tendermint in Rust. Let me know :)

lite2/client.go
```diff
@@ -0,0 +1,419 @@
...
+	// Primary provider of new headers.
+	primary provider.Provider
+
+	// Alternative providers for checking the primary for misbehavior by
+	// comparing data.
+	alternatives []provider.Provider
+
+	// Where trusted headers are stored.
+	trustedStore    store.Store
+	trustedHeader   *types.SignedHeader // H
```
ancazamfir: I believe this is the highest trusted header from the store. Some comments would help.
It is updated as we add more trusted headers, however we don't "expire" it. I assume the error ErrOldHeaderExpired raised in Verify() bubbles up to a caller that know what to do (e.g. restart the client with a new trusted header)

lite2/client.go
```diff
@@ -0,0 +1,419 @@
...
+}
+
+func (c *Client) initializeWithTrustOptions(options TrustOptions) error {
+	// 1) Fetch and verify the header.
+	h, err := c.primary.SignedHeader(options.Height)
+	if err != nil {
+		return err
+	}
+
+	// NOTE: Verify func will check if it's expired or not.
```
ancazamfir: True but maybe Client initialization should fail (? not sure myself)

lite2/client.go
```diff
@@ -0,0 +1,419 @@
...
+	if err := h.ValidateBasic(c.chainID); err != nil {
+		return errors.Wrap(err, "ValidateBasic failed")
+	}
+
```
ancazamfir: Should we fetch also c.primary.ValidatorSet(options.Height) and make sure 2/3rds signed?
Is there a way to have a function similar to verifyNewHeaderAndVals() that does not make the "h1" checks there? i.e. given a signed header and the valset for that header do all checks wether it's for the first header used to initialize the client, or for subsequent verifications.

lite2/client.go
```diff
@@ -0,0 +1,419 @@
...
+}
+
+// TrustedHeader returns a trusted header at the given height or nil if no such
+// header exist. It returns an error if there are some issues with the trusted
+// store, although that should not happen normally. TODO: mention how many
+// headers will be kept by the light client.
+//
+// 0 - the latest.
+// height must be >= 0.
+func (c *Client) TrustedHeader(height int64) (*types.SignedHeader, error) {
```
ancazamfir: Again, this trusted header may be outside the trusting period, correct?

lite2/client.go
```diff
@@ -0,0 +1,419 @@
...
+}
+
+// VerifyHeaderAtHeight fetches the header and validators at the given height
+// and calls VerifyHeader.
+//
+// If the trusted header is more recent than one here, an error is returned.
+func (c *Client) VerifyHeaderAtHeight(height int64, now time.Time) error {
+	if c.trustedHeader.Height >= height {
+		return errors.Errorf("height #%d is already trusted (last: #%d)", height, c.trustedHeader.Height)
+	}
```
ancazamfir: Is it true that it is "already trusted"? Maybe this is an older header that was never verified., (skipping case). Should we lookup the store instead?
Let's say we start with height A (trusted) and we successfully verify for height Z. So we store headers A and Z.
Now we get a request for height Q with A < Q < Z. 
Is this a valid request?  Or is there a requirement that VerifyHeaderAtHeight() must be called with increasing heights for both sequential and skipping case?
If it is a valid request then we should not trust Q automatically just because we trusted Z. It is possible that a verification of Q against A will fail (e.g. less than 1/3rds common validators between A and Q signed even though there were more than 1/3rds common between A and Z -> Note*)
Also, a proper solution for this "backward" case should check that A is still within trusting period and if not perform backwards-sequential verification (i.e. h, h-1, h-2, etc.) or return error if backwards-sequential is not supported.

So to summarize, if we do allow for random (not necessarily increasing in height) client header verification for skipping case then some changes are needed. If we do not, then maybe clarify this someplace.

Note* - if validators are required to change address when re-bonding then probably this is not an issue.

lite2/client.go
```diff
@@ -0,0 +1,419 @@
...
+		nextVals, err := c.primary.ValidatorSet(pivot + 1)
+		if err != nil {
+			return errors.Wrapf(err, "failed to obtain the vals #%d", pivot+1)
+		}
+		if !bytes.Equal(pivotHeader.NextValidatorsHash, nextVals.Hash()) {
+			return errors.Errorf("expected next validator's hash %X, but got %X (height #%d)",
+				pivotHeader.NextValidatorsHash,
+				nextVals.Hash(),
+				pivot)
+		}
```
ancazamfir: Shouldn't we store the pivots?

```suggestion
		}
		err = c.updateTrustedHeaderAndVals(pivotHeader, nextVals)
		if err != nil {
			return errors.Wrapf(err, "failed to update trusted state #%d", pivot)
		}```

File: lite2/client.go [hunk 85]
```diff
@@ -0,0 +1,419 @@
...
+
+	// Primary provider of new headers.
+	primary provider.Provider
+
+	// Alternative providers for checking the primary for misbehavior by
+	// comparing data.
+	alternatives []provider.Provider
+
+	// Where trusted headers are stored.
+	trustedStore    store.Store
```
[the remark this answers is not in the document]
melekes: > I believe headers are not removed from the store as they become "old", i.e. get out of the trusting period.

This is a TODO. We need to only store the amount of headers necessary for fork-accountability evidence. I don't know the exact number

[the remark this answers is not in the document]
melekes: I will add a comment once the requirements are known and implementation is adjusted accordingly

lite2/client.go
melekes: Good point re expiring. However, I don't think it makes sense to write about it here since "expired" is not inherent property of the Header. I will add a check and a comment to `TrustedHeader()` method

lite2/client.go
melekes: good question

lite2/client.go
melekes: > Should we fetch also c.primary.ValidatorSet(options.Height) and make sure 2/3rds signed?

what for? thought the whole point of trusting a header with HASH XYZ is that you trust it. Plus we will need to make 1 more request.

> Is there a way to have a function similar to verifyNewHeaderAndVals() that does not make the "h1" checks there?

see verifyNewHeaderAndVals in verifier.go

lite2/client.go
melekes: yes! added the check. thanks!

lite2/client.go
```diff
@@ -0,0 +1,442 @@
...
+	return c.updateTrustedHeaderAndVals(h, vals)
+}
+
+// SetLogger sets a logger.
+func (c *Client) SetLogger(l log.Logger) {
+	c.logger = l
+}
+
+// TrustedHeader returns a trusted header at the given height (0 - the latest)
+// or nil if no such header exist.
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
// or nil if no such header exist.
```

lite2/proxy/routes.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+}
+
+func makeHealthFunc(c *lrpc.Client) func(ctx *rpctypes.Context) (*ctypes.ResultHealth, error) {
+	return func(ctx *rpctypes.Context) (*ctypes.ResultHealth, error) {
+		return c.Health()
+	}
+}
+
+func makeStatusFunc(c *lrpc.Client) func(ctx *rpctypes.Context) (*ctypes.ResultStatus, error) {
```
golangcibot: `c` can be `github.com/tendermint/tendermint/rpc/client.StatusClient` (from `interfacer`)

lite2/proxy/routes.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+}
+
+func makeABCIInfoFunc(c *lrpc.Client) func(ctx *rpctypes.Context) (*ctypes.ResultABCIInfo, error) {
+	return func(ctx *rpctypes.Context) (*ctypes.ResultABCIInfo, error) {
+		return c.ABCIInfo()
+	}
+}
+
+func makeBroadcastEvidenceFunc(c *lrpc.Client) func(ctx *rpctypes.Context, ev types.Evidence) (*ctypes.ResultBroadcastEvidence, error) {
```
golangcibot: `c` can be `github.com/tendermint/tendermint/rpc/client.EvidenceClient` (from `interfacer`)

types/validator_set.go
```diff
@@ -717,27 +717,98 @@ func (vals *ValidatorSet) VerifyFutureCommit(newSet *ValidatorSet, chainID strin
...
+				talliedVotingPower += val.VotingPower
+			}
+			// else {
+			// It's OK that the BlockID doesn't match.  We include stray
+			// precommits to measure validator availability.
+			// }
+		}
+	}
+
+	if got, needed := talliedVotingPower, (vals.TotalVotingPower()*trustLevel.Numerator)/trustLevel.Denominator; got <= needed {
```
golangcibot: line is 125 characters (from `lll`)

lite2/client.go
```diff
@@ -0,0 +1,442 @@
...
+		nextVals      *types.ValidatorSet
+		err           error
+	)
+	for height := c.trustedHeader.Height + 1; height < newHeader.Height; height++ {
+		interimHeader, err = c.primary.SignedHeader(height)
+		if err != nil {
+			return errors.Wrapf(err, "failed to obtain the header #%d", height)
+		}
+
+		err = Verify(c.chainID, c.trustedHeader, c.trustedNextVals, interimHeader, c.trustedNextVals, c.trustingPeriod, now, c.trustLevel)
```
golangcibot: line is 132 characters (from `lll`)

lite2/rpc/client.go
```diff
@@ -0,0 +1,416 @@
...
+	return c.next.ABCIInfo()
+}
+
+func (c *Client) ABCIQuery(path string, data cmn.HexBytes) (*ctypes.ResultABCIQuery, error) {
+	return c.ABCIQueryWithOptions(path, data, rpcclient.DefaultABCIQueryOptions)
+}
+
+// GetWithProofOptions is useful if you want full access to the ABCIQueryOptions.
+// XXX Usage of path?  It's not used, and sometimes it's /, sometimes /key, sometimes /store.
+func (c *Client) ABCIQueryWithOptions(path string, data cmn.HexBytes, opts rpcclient.ABCIQueryOptions) (*ctypes.ResultABCIQuery, error) {
```
golangcibot: line is 137 characters (from `lll`)

lite2/proxy/routes.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+}
+
+func makeStatusFunc(c *lrpc.Client) func(ctx *rpctypes.Context) (*ctypes.ResultStatus, error) {
+	return func(ctx *rpctypes.Context) (*ctypes.ResultStatus, error) {
+		return c.Status()
+	}
+}
+
+func makeNetInfoFunc(c *lrpc.Client) func(ctx *rpctypes.Context, minHeight, maxHeight int64) (*ctypes.ResultNetInfo, error) {
```
golangcibot: line is 125 characters (from `lll`)

lite2/proxy/routes.go
```diff
@@ -0,0 +1,181 @@
...
+	return func(ctx *rpctypes.Context, minHeight, maxHeight int64) (*ctypes.ResultNetInfo, error) {
+		return c.NetInfo()
+	}
+}
+
+func makeBlockchainInfoFunc(c *lrpc.Client) func(ctx *rpctypes.Context, minHeight, maxHeight int64) (*ctypes.ResultBlockchainInfo, error) {
```
golangcibot: line is 139 characters (from `lll`)

lite2/proxy/routes.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+}
+
+func makeBlockFunc(c *lrpc.Client) func(ctx *rpctypes.Context, height *int64) (*ctypes.ResultBlock, error) {
+	return func(ctx *rpctypes.Context, height *int64) (*ctypes.ResultBlock, error) {
+		return c.Block(height)
+	}
+}
+
+func makeBlockResultsFunc(c *lrpc.Client) func(ctx *rpctypes.Context, height *int64) (*ctypes.ResultBlockResults, error) {
```
golangcibot: line is 122 characters (from `lll`)

lite2/proxy/routes.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+}
+
+func makeTxFunc(c *lrpc.Client) func(ctx *rpctypes.Context, hash []byte, prove bool) (*ctypes.ResultTx, error) {
+	return func(ctx *rpctypes.Context, hash []byte, prove bool) (*ctypes.ResultTx, error) {
+		return c.Tx(hash, prove)
+	}
+}
+
+func makeTxSearchFunc(c *lrpc.Client) func(ctx *rpctypes.Context, query string, prove bool, page, perPage int) (*ctypes.ResultTxSearch, error) {
```
golangcibot: line is 144 characters (from `lll`)

lite2/proxy/routes.go
```diff
@@ -0,0 +1,181 @@
...
+	return func(ctx *rpctypes.Context, query string, prove bool, page, perPage int) (*ctypes.ResultTxSearch, error) {
+		return c.TxSearch(query, prove, page, perPage)
+	}
+}
+
+func makeValidatorsFunc(c *lrpc.Client) func(ctx *rpctypes.Context, height *int64, page, perPage int) (*ctypes.ResultValidators, error) {
```
golangcibot: line is 137 characters (from `lll`)

lite2/proxy/routes.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+}
+
+func makeConsensusStateFunc(c *lrpc.Client) func(ctx *rpctypes.Context) (*ctypes.ResultConsensusState, error) {
+	return func(ctx *rpctypes.Context) (*ctypes.ResultConsensusState, error) {
+		return c.ConsensusState()
+	}
+}
+
+func makeConsensusParamsFunc(c *lrpc.Client) func(ctx *rpctypes.Context, height *int64) (*ctypes.ResultConsensusParams, error) {
```
golangcibot: line is 128 characters (from `lll`)

lite2/proxy/routes.go
```diff
@@ -0,0 +1,181 @@
...
+	return func(ctx *rpctypes.Context, height *int64) (*ctypes.ResultConsensusParams, error) {
+		return c.ConsensusParams(height)
+	}
+}
+
+func makeUnconfirmedTxsFunc(c *lrpc.Client) func(ctx *rpctypes.Context, limit int) (*ctypes.ResultUnconfirmedTxs, error) {
```
golangcibot: line is 122 characters (from `lll`)

lite2/proxy/routes.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+}
+
+func makeNumUnconfirmedTxsFunc(c *lrpc.Client) func(ctx *rpctypes.Context) (*ctypes.ResultUnconfirmedTxs, error) {
+	return func(ctx *rpctypes.Context) (*ctypes.ResultUnconfirmedTxs, error) {
+		return c.NumUnconfirmedTxs()
+	}
+}
+
+func makeBroadcastTxCommitFunc(c *lrpc.Client) func(ctx *rpctypes.Context, tx types.Tx) (*ctypes.ResultBroadcastTxCommit, error) {
```
golangcibot: line is 130 characters (from `lll`)

lite2/proxy/routes.go
```diff
@@ -0,0 +1,181 @@
...
+	return func(ctx *rpctypes.Context, tx types.Tx) (*ctypes.ResultBroadcastTxCommit, error) {
+		return c.BroadcastTxCommit(tx)
+	}
+}
+
+func makeBroadcastTxSyncFunc(c *lrpc.Client) func(ctx *rpctypes.Context, tx types.Tx) (*ctypes.ResultBroadcastTx, error) {
```
golangcibot: line is 122 characters (from `lll`)

lite2/proxy/routes.go
```diff
@@ -0,0 +1,181 @@
...
+	return func(ctx *rpctypes.Context, tx types.Tx) (*ctypes.ResultBroadcastTx, error) {
+		return c.BroadcastTxSync(tx)
+	}
+}
+
+func makeBroadcastTxAsyncFunc(c *lrpc.Client) func(ctx *rpctypes.Context, tx types.Tx) (*ctypes.ResultBroadcastTx, error) {
```
golangcibot: line is 123 characters (from `lll`)

lite2/proxy/routes.go
```diff
@@ -0,0 +1,181 @@
...
+	return func(ctx *rpctypes.Context, tx types.Tx) (*ctypes.ResultBroadcastTx, error) {
+		return c.BroadcastTxAsync(tx)
+	}
+}
+
+func makeABCIQueryFunc(c *lrpc.Client) func(ctx *rpctypes.Context, path string, data cmn.HexBytes) (*ctypes.ResultABCIQuery, error) {
```
golangcibot: line is 133 characters (from `lll`)

types/validator_set.go
```diff
@@ -717,27 +717,98 @@ func (vals *ValidatorSet) VerifyFutureCommit(newSet *ValidatorSet, chainID strin
...
+		}
+
+		// We don't know the validators that committed this block, so we have to
+		// check for each vote if its validator is already known.
+		_, val := vals.GetByAddress(precommit.ValidatorAddress)
+		if val != nil {
+			// Validate signature.
+			precommitSignBytes := commit.VoteSignBytes(chainID, idx)
+			if !val.PubKey.VerifyBytes(precommitSignBytes, precommit.Signature) {
+				return fmt.Errorf("Invalid commit -- invalid signature: %v", precommit)
```
golangcibot: ST1005: error strings should not be capitalized (from `stylecheck`)

lite2/client.go
ancazamfir: yes, but that takes h1 as parameter too.

ancazamfir: >> Should we fetch also c.primary.ValidatorSet(options.Height) and make sure 2/3rds signed?
>
>what for? thought the whole point of trusting a header with HASH XYZ is that you trust it. Plus we will need to make 1 more request.

We already do some checks on this "trusted" header. I was wondering why not the >2/3rds signed one. And what would happen if this would fail, vs. what happens when the basic checks fail. We definitely have to "blindly" trust the validator set that signed this header. 
I am also not entirely sure if getting the header from the primary provider is the way to initialize the client (vs. passing the header to the client init function and leave out of scope of this PR how it is obtained). I must admit I don't know the precise mechanism of "social consensus".

lite2/client.go
cwgoes: Even in bisection I don't think we should ever end up looking up a header that is older than one we've already decided to trust. We should never start bisection from a 'start header' that's before one we've already decided to trust, so presumably this won't happen (but good question & merits a comment).

ancazamfir: Ok. We do have (potentially) many trusted headers in the store. Any new bisection call for H could start with the highest trusted header <= H. I think the code can support this with minimal changes if we ever need it. 
But at least, the error message we currently have `height #%d is already trusted..` should be changed.

lite2/client.go
milosevic: We should probably not use term evidence in this context. This is error due to malformed header but not necessarily evidence of slashable behaviour. Evidence comes from cross-checking verified headers and it should probably be done as part of failure detection module.

lite2/verifier.go
```diff
@@ -0,0 +1,116 @@
...
+	h2Vals *types.ValidatorSet,
+	trustingPeriod time.Duration,
+	now time.Time,
+	trustLevel cmn.Fraction) error {
+
+	if err := ValidateTrustLevel(trustLevel); err != nil {
+		return err
+	}
+
+	// Ensure last header can still be trusted.
```
milosevic: This should probably be extracted into a function.

lite2/verifier.go
milosevic: We probably can't check proposer at the level of lite client as it requires access to Tendermint State.

lite2/verifier_test.go
```diff
@@ -0,0 +1,279 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/types"
+)
+
+func TestVerifyAdjustedHeaders(t *testing.T) {
```
milosevic: Should it be adjacent instead of adjusted?

types/validator_set.go
```diff
@@ -632,10 +631,11 @@ func (vals *ValidatorSet) VerifyCommit(chainID string, blockID BlockID, height i
 		// }
 	}
 
-	if talliedVotingPower > vals.TotalVotingPower()*2/3 {
-		return nil
+	if got, needed := talliedVotingPower, vals.TotalVotingPower()*2/3; got <= needed {
+		return ErrTooMuchChange{Got: got, Needed: needed}
```
milosevic: Probably different error should be returned in this case; it's not enough voting power not too much change.

types/validator_set.go
```diff
@@ -717,27 +717,100 @@ func (vals *ValidatorSet) VerifyFutureCommit(newSet *ValidatorSet, chainID strin
 		// }
 	}
 
-	if oldVotingPower <= oldVals.TotalVotingPower()*2/3 {
-		return errTooMuchChange{oldVotingPower, oldVals.TotalVotingPower()*2/3 + 1}
+	if got, needed := oldVotingPower, oldVals.TotalVotingPower()*2/3; got <= needed {
+		return ErrTooMuchChange{Got: got, Needed: needed}
```
milosevic: ditto

lite2/verifier.go
Shivani912: I was thinking of this in case of sequential verification, but never mind as there is no way in case of skipping.

lite2/verifier.go
melekes: why?

milosevic: Defining what is a trusted header will potentially used from several places. It is also important part of business logic that could be changed in the future so it is probably good idea if it's isolated in a function.

melekes: "will potentially used" argument rarely works in practice. better to make changes when A/B/C actually happens

lite2/client.go
melekes: > And what would happen if this would fail, vs. what happens when the basic checks fail

light client exits with error?

On the second thought, we should probably design light client for max security. at least initially and only after maybe relax conditions. I will add the check.

lite2/client.go
melekes: > We do have (potentially) many trusted headers in the store. Any new bisection call for H could start with the highest trusted header <= H.

I think it only makes sense with a different provider. Consider this:

a) light client performs bisection and decides to trust H#100
b) later it receives an evidence of attach (lunatic) and learns that H#100 is fabricated
c) light client exits (we should probably remove all "trusted" headers from DB at this point)
c) operator resets it with new primary source and new trusted header

lite2/client.go
melekes: https://github.com/tendermint/tendermint/pull/4209/commits/e757cff5a707718e2ebd6d9a7384738822f7edfd

lite2/client.go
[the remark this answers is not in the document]
melekes: we will need to store all headers that are within trusted period

lite2/client.go
melekes: > Is initializeWithTrustOptions intended to be called multiple times?

no, one time. but light client can be started/stopped multiple times (due to crash or other reasons)

PR #3990, opened 2019-09-17
checkTx trusted

### LINKED ISSUE #3919: Add a checkTxLocal method in mempool interface

james-ray:
fix #3919 
* [ ] Wrote tests


james-ray:
Is this possible to be merged?  Are the names CheckTx_Local and CheckTx_Remote confusing, to be modified to other names?

docs/spec/abci/apps.md
```diff
@@ -67,8 +67,8 @@ node's local mempool after filtering those included in the block. To prevent the
 mempool from rechecking all transactions every time a block is committed, set
 the configuration option `mempool.recheck=false`. As of Tendermint v0.32.1,
 an additional `Type` parameter is made available to the CheckTx function that
-indicates whether an incoming transaction is new (`CheckTxType_New`), or a
-recheck (`CheckTxType_Recheck`).
+indicates whether an incoming transaction is (`CheckTxType_Local`), (`CheckTxType_Remote`),
```
melekes: needs to be rewritten

```suggestion
indicates whether an incoming transaction is untrusted (`CheckTxType_Untrusted`), trusted (`CheckTxType_Trusted`),
```

melekes:
`CheckTx` ABCI docs need to be updated as well https://github.com/tendermint/tendermint/blob/master/docs/spec/abci/abci.md#checktx

consensus/state.go
```diff
@@ -1329,7 +1409,10 @@ func (cs *ConsensusState) finalizeCommit(height int64) {
 	// Either way, the ConsensusState should not be resumed until we
 	// successfully call ApplyBlock (ie. later here, or in Handshake after
 	// restart).
-	cs.wal.WriteSync(EndHeightMessage{height}) // NOTE: fsync
+	endMsg := EndHeightMessage{height}
+	if err := cs.wal.WriteSync(endMsg); err != nil { // NOTE: fsync
+		panic(fmt.Sprintf("Failed to write %v msg to consensus wal due to %v. Check your FS and restart the node", endMsg, err))
```
golangcibot: line is 122 characters (from `lll`)

consensus/wal.go
```diff
@@ -174,29 +176,39 @@ func (wal *baseWAL) Wait() {
...
+	}
+
+	if err := wal.Write(msg); err != nil {
+		return err
 	}
 
-	wal.Write(msg)
 	if err := wal.FlushAndSync(); err != nil {
-		panic(fmt.Sprintf("Error flushing consensus wal buf to file. Error: %v \n", err))
+		wal.Logger.Error("WriteSync failed to flush consensus wal. WARNING: may result in creating alternative proposals / votes for the current height iff the node restarted",
```
golangcibot: line is 170 characters (from `lll`)

abci/tests/test_app/main.go
```diff
@@ -53,7 +53,7 @@ func testCounter() {
 	}
 
 	fmt.Printf("Running %s test with abci=%s\n", abciApp, abciType)
-	cmd := exec.Command("bash", "-c", fmt.Sprintf("abci-cli %s", abciApp)) // nolint: gas
+	cmd := exec.Command("bash", "-c", fmt.Sprintf("abci-cli %s", abciApp))
```
golangcibot: G204: Subprocess launched with function call as argument or cmd arguments (from `gosec`)

james-ray:
@melekes  Seems like the decrease of the codecov/patch is not relative of this pr, but because I merge recent updates to my local branch.

melekes:
This PR needs a proper rebase.

PR #3999, opened 2019-09-19
remove benchmarks folder

melekes:
Afaik, nobody is using it. I've never used it myself.

1) atomic, os.Write, map, codec benchmarks all seem temporary and to be
of no use in the future.
2) syncing benchmark may be useful, but we're trying to move away from
Bash. Also, the blocks are empty there.

* [ ] Wrote tests


marbar3778:
changelog_pending entry?

melekes:
> changelog_pending entry?

no need

PR #4005, opened 2019-09-19
Remove -v from test_cover

marbar3778:
* [ ] Wrote tests


melekes:
Without `-v` we won't get any logs.

marbar3778:
I would be more in favor of setting test_cover without v and test_coverv with v but allow failure since it fails from the logs mainly

marbar3778:
closing this, when it happens again we can discuss options

Issue #3986, opened 2019-09-13
Conflicts on protobuf types


antho1404:
When including Tendermint in a code source that already uses proto and has similar types, some types are not registered because Tendermint already registers them.

More details here https://github.com/mesg-foundation/engine/issues/1236

To fix that all the proto definitions could be prefixed with a specific namespace, eg: `tendermint.types` instead of `types` to avoiding any conflicts.

melekes:
Can't you put generated code in `tendermint_abci` package? I don't understand where about conflict comes from. Like if you combine all .proto files together (which contain duplicate types), then yes, it will rightfully fail. But if you generate Tendermint .proto files independently from XYZ .proto files, then there should be no conflict.

antho1404:
I'm not doing anything with the proto, I'm using directly the go package (and cosmos sdk) so I don't build these proto.

My application already has its own types in gogoproto. We define these types on a type package and have one called `Event`. 
Because of that, gogoproto will register `types.Event` from ABCI [here](https://github.com/tendermint/tendermint/blob/7e0b64e8b0842d2f13c2b24feed13ec187be10f5/abci/types/types.pb.go#L3417) and `types.Event` from my application [here](https://github.com/mesg-foundation/engine/blob/3f4b289bad1482e7c298aa61ba949a75992dde56/event/event.pb.go#L66).

One of these proto messages cannot be registered as there is already one with the same name.

melekes:
Makes sense now, thanks! Does this mean we should prefix all our types?

antho1404:
That's the solution that we adopted to fix this issue here https://github.com/mesg-foundation/engine/pull/1337. 
I think that would be something really nice to have at the tendermint level too.

This might have some impacts on some libraries that generate some code based on the proto files so it might be a breaking change.

If that's fine, I can work on a pull request for that.

melekes:
> If that's fine, I can work on a pull request for that.

Please do 🙇

PR #4007, opened 2019-09-20
types: prefix abci proto types

antho1404:
Fix https://github.com/tendermint/tendermint/issues/3986

This pull request is prefixing all the types in proto to avoid conflict.
When a go application is using Tendermint as a library and also define similar types in gogo proto some conflicts might occur (as types is a common package in go).

By prefixing the types with `tendermint`, this highly reduces the risk of conflicts.

**BREAKING CHANGE.**

This modification breaks the ABCI Application endpoint.
What was accessible before with `/types.ABCIApplication/Flush` is now accessible with `/tendermint.types.ABCIApplication/Flush`.

This breaking change could be avoided by splitting the `abci/types/types.proto` into 2 different proto packages `types` (for the service) and `tendermint.types` (for the messages).


Makefile
```diff
@@ -7,7 +7,7 @@ GOBIN?=${GOPATH}/bin
 PACKAGES=$(shell go list ./...)
 OUTPUT?=build/tendermint
 
-INCLUDE = -I=. -I=${GOPATH}/src -I=${GOPATH}/src/github.com/gogo/protobuf/protobuf
+INCLUDE = -I=${GOPATH}/src/github.com/tendermint/tendermint -I=${GOPATH}/src -I=${GOPATH}/src/github.com/gogo/protobuf/protobuf
```
melekes: `.` does not work?

antho1404: It doesn't work because the compilation is now moved at the `GOPATH/src` level to keep the naming from the `go_package` option.

Generated files are generated using the full path of the go_package and will then generate a new `github.com/tendermint/tendermint/packageX` within the sources. 

That's why the compilation is not on the root of the project but the root of the `GOPATH/src`.

This can be changed by changing the `go_package` option from `github.com/tendermint/tendermint/packageX` to `packageX`. It doesn't follow go conventions but it's totally acceptable.

If it's done this way, the makefile will not change and the compilation will still be at the project level.

Let me know if you prefer this solution.

melekes: I see. Thanks for explaining it to me! We should probably play nice and respect the naming.

antho1404: Do you want me to revert the last commit?

melekes: Please do

antho1404: done :)

types/proto3/block.proto
```diff
@@ -1,7 +1,7 @@
 syntax = "proto3";
 
 package proto3;
```
melekes: why not `tendermint.proto3`?

rpc/grpc/types.proto
```diff
@@ -1,5 +1,6 @@
 syntax = "proto3";
 package core_grpc;
```
melekes: why not `tendermint.core_grpc`?

libs/common/types.proto
```diff
@@ -1,5 +1,6 @@
 syntax = "proto3";
 package common;
```
melekes: why not `tendermint.common`?

crypto/merkle/merkle.proto
```diff
@@ -1,5 +1,6 @@
 syntax = "proto3";
 package merkle;
```
melekes: why not `tendermint.merkle`?

abci/types/types.proto
```diff
@@ -1,5 +1,6 @@
 syntax = "proto3";
-package types;
+package tendermint.types;
+option go_package = "github.com/tendermint/tendermint/abci/types";
```
melekes: do you think we should also provide custom class names for other languages?

```
option csharp_namespace = "Tendermint.ABCI.Types";
option cc_enable_arenas = true;
option java_package = "com.tendermint.abci";
option java_outer_classname = "TypesProto";
option java_multiple_files = true;
option objc_class_prefix = "TMABCI";
option go_package = "github.com/tendermint/tendermint/abci/types;types";
```

similar to https://github.com/protocolbuffers/protobuf/blob/bb0c5439343a01ffbe9542f80f5297a7aedaba3b/src/google/protobuf/type.proto#L38

melekes: also, don't `package tendermint.abci.types;` make more sense to you?

types/proto3/block.proto
antho1404: I wanted not to break everything that's why I didn't prefix the other packages but yes it's definitely better to namespace everything, I can add this.

abci/types/types.proto
antho1404: For the namespace for other languages, I guess it's better, I'm not sure how it works in each case so I would rather keep the default but if if you have someone that manages the compatibility of your proto that would definitely be a good thing to do. 

I can prefix it with the package path. And maybe all of them, so something like that
- tendermint.abci.types
- tendermint.crypto.merkle
- tendermint.libs.common
- tendermint.rpc.grpc
- tendermint.types.proto3

We could have better names but at least this one is one decided by convention.

Let me know what you think.

melekes: love the new names

antho1404: update done ;)

CHANGELOG_PENDING.md
```diff
@@ -52,6 +52,8 @@ program](https://hackerone.com/tendermint).
 - P2P Protocol
   - [p2p] [\3668](https://github.com/tendermint/tendermint/pull/3668) Make `SecretConnection` non-malleable
 
+- \#3986 Prefix protobuf types. ABCI types are prefixed with `tendermint`. API endpoint changes from `/types.ABCIApplication/SERVICE` to `/tendermint.types.ABCIApplication/SERVICE`.
```
melekes: needs to be updated

antho1404: done

PR #4016, opened 2019-09-25
Update go-built-in.md

shiknian:
there is an error on function parameter .the abcitypes.Application--DeliverTx's parameter is []byte,not abcitypes.ReqyestDeliverTx


shiknian:
i'm sorry ,i found  my tendermint is not the latest . it's my wrong

PR #4024, opened 2019-09-30
Add Version to golangci

marbar3778:
- added version to golangci because of recent update

* [ ] Wrote tests

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,6 +14,10 @@ linters:
     - gochecknoglobals
     - gochecknoinits
     - stylecheck
+    - funlen
+    - godox
+    - dogsled
+    - whitespace
 # linters-settings:
 #   govet:
 #     check-shadowing: true
@@ -45,7 +49,3 @@ linters:
 #     disabled-checks:
 #       - wrapperFunc
 #       - commentFormatting # https://github.com/go-critic/go-critic/issues/755
-
-service:
-  golangci-lint-version: 1.17.x # use the fixed version to not introduce new linters unexpectedly
-  prepare:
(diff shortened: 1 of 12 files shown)


.golangci.yml
```diff
@@ -46,3 +46,6 @@ linters:
 #       - wrapperFunc
 #       - commentFormatting # https://github.com/go-critic/go-critic/issues/755
 
+service:
+  golangci-lint-version: 1.17.x # use the fixed version to not introduce new linters unexpectedly
+  prepare:
```
marbar3778: doing this now, will go through and update to most recent and clean up according to the added linters in the coming days.

rpc/core/pipe.go
```diff
@@ -70,7 +70,7 @@ var (
 	// objects
 	pubKey           crypto.PubKey
 	genDoc           *types.GenesisDoc // cache the genesis structure
-	addrBook         p2p.AddrBook
+	addrBook         p2p.AddrBook      // nolint: unused
```
melekes: unused? can be removed if so

.golangci.yml
```diff
@@ -46,3 +46,6 @@ linters:
 #       - wrapperFunc
 #       - commentFormatting # https://github.com/go-critic/go-critic/issues/755
 
+service:
+  golangci-lint-version: 1.17.x # use the fixed version to not introduce new linters unexpectedly
```
melekes: how often new linters are added? feel like we'll never be updating lint version if it will require us manually changing the version

marbar3778: i can also disable the new linters for now and work through them this week

PR #4037, opened 2019-10-04
Cleanup docs

marbar3778:
- There were some docs that needed to be removed as they were deprecated or moved

=
* [ ] Wrote tests


melekes:
where's `ecosystem.json` now? tendermint.com still using it I think

marbar3778:
theres a pr in the tendermint site to remove it from gitignore, so it lives in there now, also for the new one there wont be an ecosystem page...

PR #4042, opened 2019-10-07
Docs theme

fadeev:
This PR includes an updated documentation theme.

Screenshot 2019-10-07 at 19 04 21

To test it out run `cd docs && npm install && npm run serve`, or visit http://tendermint-docs.surge.sh/

Features such as search, definition tooltips and language switcher will be added incrementally.


fadeev:
Since sidebar is generated automatically, it is very possible that it looks not the way we want it to. Please, report what needs to be changed, and I will either modify the script or propose changes to the directory structure (sidebar displays only two levels of dirs).

melekes:
love it ❤️ 

One note: Help & Support section does not make sense to me. The riot channel for Tendermint is https://riot.im/app/#/room/#tendermint:matrix.org. Forum link: https://forum.cosmos.network/c/tendermint. Help & Support and footer both look like from cosmos project. Maybe we should tailor it for Tendermint?

fadeev:
@melekes thanks! Absolutely, i'll make it configurable.

docs/dev_sessions/DEV_SESSIONS.md
```diff
@@ -1,3 +1,9 @@
+---
+order: 1
+parent:
+  title: Developer Sessions
```
marbar3778: @fadeev I couldn't get this title to display in the sidebar, any ideas?

fadeev: `parent` is only parsed in `readme.md` at this point. Otherwise, if we have several files in a directory, and each has a `parent` property, the script wouldn't know which to use, hence, `readme.md`.

So, I've added readme.md:

```
---
order: false
parent:
  title: Developer Sessions
  order: 8
---

Comment (marbar3778) [reply docs/dev_sessions/DEV_SESSIONS.md]: ah okay, makes sense, thank you! 🔥 🚰 🚒 👨‍🚒 👩‍🚒

Comment (fadeev) [reply docs/dev_sessions/DEV_SESSIONS.md]: I promise to write a readme on how to use this theme 😅 after we settle on the details.

**Comment by fadeev** [comment]
Made changes to the script, so specs should show up in the sidebar (for example, ABCI spec). Only `{{spec_name}}/readme.md` shows up, because we have a two-level hierarchy. According to design each spec should have it's contents (files, other than readme) displayed as cards. I'm working on it.

**Comment by fadeev** [comment]
Added cards for direcotry contents. Useful for specs in `tendermint` and modules in `cosmos-sdk`. To enable add `cards: true`. Descriptions under title are taken from `description` property of the corresponding file's frontmatter.

Looks something like this:

Screenshot 2019-10-08 at 17 27 31

**Comment by fadeev** [comment]
Screenshot 2019-10-08 at 21 06 21

**Comment by fadeev** [comment]
Made footer customizable (see `config.js`), brought back language selector, added very basic search. This is a super naive search, it doesn't represent how a proper search should look or behave like.

Screenshot 2019-10-09 at 18 07 06

File: docs/.vuepress/config.js [hunk 2]
```diff
@@ -1,11 +1,120 @@
...
+    editLink: true,
     docsDir: "docs",
     logo: "/logo.svg",
     label: "core",
+    gutter: {
+      title: "Help & Support",
+      editLink: true,
+      children: [
+        {
+          title: "Riot Chat",
```
marbar3778: should these be links?

marbar3778:
docs/spec/rpc/index.html → docs/.vuepress/public/rpc/index.html 
 docs/spec/rpc/swagger.yaml → docs/.vuepress/public/rpc/swagger.yaml 

how come these were moved? I think these should live in the rpc pkg for the time being,

fadeev:
@marbar3778 re:rpc VuePress pretty much ignores HTML files outside of public and doesn't include them into the final build directory.

marbar3778:
>VuePress pretty much ignores HTML files outside of public and doesn't include them into the final build directory.

yes, simplest is to copy in before build

fadeev:
@marbar3778 Do you mean to store rpc in `docs/rpc` and copy it to public as part of build script (and delete afterwards) or to keep it in `docs/.vuepress/public/rpc`?

marbar3778:
just store the swagger files in `rpc/swagger` folder then before build just `cp -rT rpc/swagger docs/vue-press/public` and then its there on build but doesn't live there permanently

docs/.vuepress/config.js
marbar3778: sweet with that i think this is ready to be merged

docs/.vuepress/config.js
```diff
@@ -1,141 +1,140 @@
...
-  markdown: {
-    lineNumbers: true
-  },
+  theme: "cosmos",
+  // locales: {
+  //   "/": {
+  //     lang: "en-US"
+  //   },
+  //   "/ru/": {
+  //     lang: "ru"
```
melekes: ?

marbar3778: currently there is no russian translation so commented out the options

.circleci/config.yml
```diff
@@ -397,11 +377,7 @@ workflows:
   version: 2
   test-suite:
     jobs:
-      - deploy_docs:
```
marbar3778: can we keep this to master? I believe this is what the sdk team talked about.

PR #4056, opened 2019-10-14
New lint version upgrade

marbar3778:
- linter was upgraded

* [ ] Wrote tests

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -22,8 +22,8 @@ linters:
     - ineffassign
     - interfacer
     # - lll
-    - maligned
     - misspell
+    - maligned
     - nakedret
     # - prealloc
     - scopelint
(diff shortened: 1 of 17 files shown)


.golangci.yml
```diff
@@ -49,3 +77,6 @@ linters-settings:
 #     disabled-checks:
 #       - wrapperFunc
 #       - commentFormatting # https://github.com/go-critic/go-critic/issues/755
+
+service:
```
marbar3778: this is done for now, will test tomorrow if this can be removed, the bot was not upgraded in accordance with the breaking change

state/state_test.go
```diff
@@ -37,7 +37,6 @@ func setupTestCase(t *testing.T) (func(t *testing.T), dbm.DB, sm.State) {
 func TestStateCopy(t *testing.T) {
 	tearDown, _, state := setupTestCase(t)
 	defer tearDown(t)
-	// nolint: vetshadow
 	assert := assert.New(t)
```
melekes: note: the agreement here is to not use such a construct

PR #4057, opened 2019-10-14
Include peer ID when logging rejected txns

### LINKED ISSUE #3091: Mempool: Rejected bad transaction should log source (nodeid or RPC)
```
Jan  6 11:07:52 peering gaiad[10881]: I[6016-01-06|11:07:52.322] Rejected bad transaction                     module=mempool tx=5A3EEEA0890677800FB26CD31C1B1ECF7CFFB20F9E730DFD54169909DAB7ADE4 res="&{CheckTx:code:4 log:\"{\\\"codespace\\\":\\\"sdk\\\",\\\"code\\\":4,\\\"message\\\":\\\"signature verification failed\\\"}\" }" err=null
```

Log is missing important information, such as the source of the invalid TXs (p2p nodeid or RPC).

This is very useful for production to determine source of invalid TXs, especially in a "trusted" relay/peering setup.

erikgrinaker:
Fixes #3091. Adds `TxInfo.SenderAddress` for the `p2p.ID`, and logs it as `peerID=<p2p.ID>`, but keeps using `TxInfo.SenderID` internally to save memory.

* [x] ~Wrote tests~


mempool/clist_mempool.go
```diff
@@ -360,7 +366,12 @@ func (mem *CListMempool) removeTx(tx types.Tx, elem *clist.CElement, removeFromC
 //
 // The case where the app checks the tx for the second and subsequent times is
 // handled by the resCbRecheck callback.
-func (mem *CListMempool) resCbFirstTime(tx []byte, peerID uint16, res *abci.Response) {
+func (mem *CListMempool) resCbFirstTime(
+	tx []byte,
+	peerID uint16,
```
melekes: Does it make sense to rename these two variables to `peerShortID` and `peerID` resp.?

erikgrinaker: Yeah, the naming isn't ideal. I think we should keep the `peerID` name, since that's what it's called throughout the Mempool, but I changed `TxInfo.SenderAddress` and `peerAddress` to `PeerFullID`. What do you think?

melekes: so peerID in mempool and p2p would mean different things... I think It's fine, but we should probably rename `PeerFullID` to `peerP2PID`, so it's obvious that `peerP2PID` equals to `p2p.ID`. Thoughts?

erikgrinaker: Sure thing, done.

mempool/reactor.go
```diff
@@ -165,8 +165,11 @@ func (memR *Reactor) Receive(chID byte, src p2p.Peer, msgBytes []byte) {
 
 	switch msg := msg.(type) {
 	case *TxMessage:
-		peerID := memR.ids.GetForPeer(src)
-		err := memR.mempool.CheckTxWithInfo(msg.Tx, nil, TxInfo{SenderID: peerID})
+		txInfo := TxInfo{SenderID: memR.ids.GetForPeer(src)}
+		if src != nil {
+			txInfo.SenderFullID = src.ID()
```
golangcibot: txInfo.SenderFullID undefined (type TxInfo has no field or method SenderFullID) (from `typecheck`)

PR #4070, opened 2019-10-21
Relax Tx Querying

alexanderbez:
- Some linting/cleanup missed from the initial events refactor
- Don't panic; instead, return `false, error` when matching breaks unexpectedly 
- Strip non-numeric chars from values when attempting to match against query values
- Have the server log during `send` upon error 

----

* [x] Wrote tests

### DIFF
--- a/libs/pubsub/pubsub.go
+++ b/libs/pubsub/pubsub.go
@@ -37,6 +37,7 @@ package pubsub
 import (
 	"context"
 	"errors"
+	"fmt"
 	"sync"
 
 	cmn "github.com/tendermint/tendermint/libs/common"
@@ -68,7 +69,7 @@ var (
 // allows event types to repeat themselves with the same set of keys and
 // different values.
 type Query interface {
-	Matches(events map[string][]string) bool
+	Matches(events map[string][]string) (bool, error)
 	String() string
 }
 
@@ -334,7 +335,9 @@ loop:
 		case sub:
 			state.add(cmd.clientID, cmd.query, cmd.subscription)
 		case pub:
-			state.send(cmd.msg, cmd.events)
+			if err := state.send(cmd.msg, cmd.events); err != nil {
+				s.Logger.Error("Error querying for events", "err", err)
+			}
 		}
 	}
 }
@@ -401,10 +404,16 @@ func (state *state) removeAll(reason error) {
 	}
 }
 
-func (state *state) send(msg interface{}, events map[string][]string) {
+func (state *state) send(msg interface{}, events map[string][]string) error {
 	for qStr, clientSubscriptions := range state.subscriptions {
 		q := state.queries[qStr].q
-		if q.Matches(events) {
+
+		match, err := q.Matches(events)
+		if err != nil {
+			return fmt.Errorf("failed to match against query %s: %w", q.String(), err)
+		}
+
+		if match {
 			for clientID, subscription := range clientSubscriptions {
 				if cap(subscription.out) == 0 {
 					// block on unbuffered channel
@@ -420,4 +429,6 @@ func (state *state) send(msg interface{}, events map[string][]string) {
 			}
 		}
 	}
+
+	return nil
 }
--- a/libs/pubsub/pubsub.go
+++ b/libs/pubsub/pubsub.go
@@ -410,7 +410,7 @@ func (state *state) send(msg interface{}, events map[string][]string) error {
 
 		match, err := q.Matches(events)
 		if err != nil {
-			return fmt.Errorf("failed to match against query %s: %w", q.String(), err)
+			return errors.Wrapf(err, "failed to match against query %s", q.String())
 		}
 
 		if match {
--- a/libs/pubsub/pubsub.go
+++ b/libs/pubsub/pubsub.go
@@ -36,10 +36,9 @@ package pubsub
 
 import (
 	"context"
-	"errors"
-	"fmt"
 	"sync"
 
+	"github.com/pkg/errors"
 	cmn "github.com/tendermint/tendermint/libs/common"
 )
 
--- a/libs/pubsub/query/query.go
+++ b/libs/pubsub/query/query.go
@@ -84,74 +84,103 @@ const (
 
 // Conditions returns a list of conditions.
 func (q *Query) Conditions() []Condition {
-	conditions := make([]Condition, 0)
+	var (
+		eventAttr string
+		op        Operator
+	)
 
+	conditions := make([]Condition, 0)
 	buffer, begin, end := q.parser.Buffer, 0, 0
 
-	var tag string
-	var op Operator
-
-	// tokens must be in the following order: tag ("tx.gas") -> operator ("=") -> operand ("7")
+	// tokens must be in the following order: event attribute ("tx.gas") -> operator ("=") -> operand ("7")
 	for _, token := range q.parser.Tokens() {
 		switch token.pegRule {
-
 		case rulePegText:
 			begin, end = int(token.begin), int(token.end)
+
 		case ruletag:
-			tag = buffer[begin:end]
+			eventAttr = buffer[begin:end]
+
 		case rulele:
 			op = OpLessEqual
+
 		case rulege:
 			op = OpGreaterEqual
+
 		case rulel:
 			op = OpLess
+
 		case ruleg:
 			op = OpGreater
+
 		case ruleequal:
 			op = OpEqual
+
 		case rulecontains:
 			op = OpContains
+
 		case rulevalue:
 			// strip single quotes from value (i.e. "'NewBlock'" -> "NewBlock")
 			valueWithoutSingleQuotes := buffer[begin+1 : end-1]
-			conditions = append(conditions, Condition{tag, op, valueWithoutSingleQuotes})
+			conditions = append(conditions, Condition{eventAttr, op, valueWithoutSingleQuotes})
+
 		case rulenumber:
 			number := buffer[begin:end]
 			if strings.ContainsAny(number, ".") { // if it looks like a floating-point number
 				value, err := strconv.ParseFloat(number, 64)
 				if err != nil {
-					panic(fmt.Sprintf("got %v while trying to parse %s as float64 (should never happen if the grammar is correct)",
+					fmt.Printf(
+						"got %v while trying to parse %s as float64 (should never happen if the grammar is correct)\n",
 						err,
-						number))
+						number,
+					)
+
+					return []Condition{}
 				}
-				conditions = append(conditions, Condition{tag, op, value})
+
+				conditions = append(conditions, Condition{eventAttr, op, value})
 			} else {
 				value, err := strconv.ParseInt(number, 10, 64)
 				if err != nil {
-					panic(fmt.Sprintf("got %v while trying to parse %s as int64 (should never happen if the grammar is correct)",
+					fmt.Printf(
+						"got %v while trying to parse %s as int64 (should never happen if the grammar is correct)\n",
 						err,
-						number))
+						number,
+					)
+
+					return []Condition{}
 				}
-				conditions = append(conditions, Condition{tag, op, value})
+
+				conditions = append(conditions, Condition{eventAttr, op, value})
 			}
+
 		case ruletime:
 			value, err := time.Parse(TimeLayout, buffer[begin:end])
 			if err != nil {
-				panic(fmt.Sprintf(
-					"got %v while trying to parse %s as time.Time / RFC3339 (should never happen if the grammar is correct)",
+				fmt.Printf(
+					"got %v while trying to parse %s as time.Time / RFC3339 (should never happen if the grammar is correct)\n",
 					err,
-					buffer[begin:end]))
+					buffer[begin:end],
+				)
+
+				return []Condition{}
 			}
-			conditions = append(conditions, Condition{tag, op, value})
+
+			conditions = append(conditions, Condition{eventAttr, op, value})
+
 		case ruledate:
 			value, err := time.Parse("2006-01-02", buffer[begin:end])
 			if err != nil {
-				panic(fmt.Sprintf(
-					"got %v while trying to parse %s as time.Time / '2006-01-02' (should never happen if the grammar is correct)",
+				fmt.Printf(
+					"got %v while trying to parse %s as time.Time / '2006-01-02' (should never happen if the grammar is correct)\n",
 					err,
-					buffer[begin:end]))
+					buffer[begin:end],
+				)
+
+				return []Condition{}
 			}
-			conditions = append(conditions, Condition{tag, op, value})
+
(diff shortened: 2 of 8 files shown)


libs/pubsub/query/query.go
```diff
@@ -84,74 +84,103 @@ const (
...
+			conditions = append(conditions, Condition{eventAttr, op, valueWithoutSingleQuotes})
+
 		case rulenumber:
 			number := buffer[begin:end]
 			if strings.ContainsAny(number, ".") { // if it looks like a floating-point number
 				value, err := strconv.ParseFloat(number, 64)
 				if err != nil {
-					panic(fmt.Sprintf("got %v while trying to parse %s as float64 (should never happen if the grammar is correct)",
+					fmt.Printf(
+						"got %v while trying to parse %s as float64 (should never happen if the grammar is correct)\n",
```
melekes: I feel like we should leave panics in this function because they assert the grammar. I.e. the only way this code panics is if the grammar is wrong

alexanderbez: Ok, reverted.

libs/pubsub/query/query.go
```diff
@@ -169,86 +191,114 @@ func (q *Query) Matches(events map[string][]string) bool {
...
+			if !match(eventAttr, op, reflect.ValueOf(value), events) {
 				return false
 			}
+
 		case ruledate:
 			value, err := time.Parse("2006-01-02", buffer[begin:end])
 			if err != nil {
-				panic(fmt.Sprintf(
-					"got %v while trying to parse %s as time.Time / '2006-01-02' (should never happen if the grammar is correct)",
+				fmt.Printf(
```
melekes: same comment about panic here

libs/pubsub/query/query.go
```diff
@@ -169,86 +191,114 @@ func (q *Query) Matches(events map[string][]string) bool {
...
 					return false
 				}
 			}
+
 		case ruletime:
 			value, err := time.Parse(TimeLayout, buffer[begin:end])
 			if err != nil {
-				panic(fmt.Sprintf(
-					"got %v while trying to parse %s as time.Time / RFC3339 (should never happen if the grammar is correct)",
+				fmt.Printf(
```
melekes: same comment about panic here

libs/pubsub/query/query.go
```diff
@@ -169,86 +191,114 @@ func (q *Query) Matches(events map[string][]string) bool {
...
+
+				if !match(eventAttr, op, reflect.ValueOf(value), events) {
 					return false
 				}
 			} else {
 				value, err := strconv.ParseInt(number, 10, 64)
 				if err != nil {
-					panic(fmt.Sprintf(
-						"got %v while trying to parse %s as int64 (should never happen if the grammar is correct)",
+					fmt.Printf(
```
melekes: same comment about panic here

libs/pubsub/query/query.go
```diff
@@ -169,86 +191,114 @@ func (q *Query) Matches(events map[string][]string) bool {
...
 			}
+
 		case rulenumber:
 			number := buffer[begin:end]
 			if strings.ContainsAny(number, ".") { // if it looks like a floating-point number
 				value, err := strconv.ParseFloat(number, 64)
 				if err != nil {
-					panic(fmt.Sprintf(
-						"got %v while trying to parse %s as float64 (should never happen if the grammar is correct)",
+					fmt.Printf(
```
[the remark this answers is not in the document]
alexanderbez: Why panic here and not in `matchValue`? Why panic at all in query-handling logic?

libs/pubsub/query/query.go
alexanderbez: Check response :)

libs/pubsub/query/query.go
alexanderbez: Check response :)

libs/pubsub/query/query.go
alexanderbez: Check response :)

libs/pubsub/query/query.go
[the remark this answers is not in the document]
marbar3778: I am agreeing with bez on this one, I can think of a reason to panic here, the incorrect grammar will be noticed by the developer and fixed.

melekes:
> TODO: I would rather use an actual logger with a tagged module/context instead of fmt.Printf. We can use a private package-level logger if you're OK with this @melekes.

I very hesitant of adding logger to `Query` struct. What if we just use stdlib logger?

libs/pubsub/query/query.go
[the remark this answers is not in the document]
melekes: error here indicates a fundamental error with the grammar, which I'd rather be aware immediately. I.e. it's a corner case, not a normal situation.

libs/pubsub/query/query.go
```diff
@@ -318,13 +369,16 @@ func matchValue(value string, op Operator, operand reflect.Value) bool {
 		}
 
 	case reflect.Float64:
-		operandFloat64 := operand.Interface().(float64)
 		var v float64
 
+		operandFloat64 := operand.Interface().(float64)
+		matchedValue := numRegex.FindString(value)
```
melekes:
```suggestion
		filteredValue := numRegex.FindString(value)
```

?

libs/pubsub/query/query.go
```diff
@@ -318,13 +369,16 @@ func matchValue(value string, op Operator, operand reflect.Value) bool {
...
+
 		// try our best to convert value from tags to float64
-		v, err := strconv.ParseFloat(value, 64)
+		v, err := strconv.ParseFloat(matchedValue, 64)
 		if err != nil {
-			panic(fmt.Sprintf("failed to convert value %v from tag to float64: %v", value, err))
+			fmt.Printf("failed to convert value %v from event attribute to float64: %v\n", matchedValue, err)
```
melekes: use stdlib logger?

libs/pubsub/query/query.go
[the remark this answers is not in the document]
melekes: maybe we should not panic, but return an error...

libs/pubsub/query/query.go
alexanderbez: sure 🤷‍♂

libs/pubsub/query/query.go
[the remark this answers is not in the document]
melekes: note: it will only panic if Query.New("foo.bar > 50") returns OK, but Conditions suddenly discovers that "foo.bar > 50" is invalid, which should not happen NORMALLY if parser (Query.New step) works correctly

marbar3778:
libs/pubsub/query/query.go:326:16: multiple-value matchValue() in single-value context

libs/pubsub/query/query.go
```diff
@@ -160,131 +182,167 @@ func (q *Query) Conditions() []Condition {
...
 		case rulenumber:
 			number := buffer[begin:end]
 			if strings.ContainsAny(number, ".") { // if it looks like a floating-point number
 				value, err := strconv.ParseFloat(number, 64)
 				if err != nil {
-					panic(fmt.Sprintf(
-						"got %v while trying to parse %s as float64 (should never happen if the grammar is correct)",
-						err,
-						number))
+					return false, fmt.Errorf("got %v while trying to parse %s as float64 (should never happen if the grammar is correct)", err, number)
```
golangcibot: line is 136 characters (from `lll`)

libs/pubsub/query/query.go
```diff
@@ -160,131 +182,167 @@ func (q *Query) Conditions() []Condition {
...
+					return false, nil
 				}
 			} else {
 				value, err := strconv.ParseInt(number, 10, 64)
 				if err != nil {
-					panic(fmt.Sprintf(
-						"got %v while trying to parse %s as int64 (should never happen if the grammar is correct)",
-						err,
-						number))
+					return false, fmt.Errorf("got %v while trying to parse %s as int64 (should never happen if the grammar is correct)", err, number)
```
golangcibot: line is 134 characters (from `lll`)

libs/pubsub/query/query.go
```diff
@@ -160,131 +182,167 @@ func (q *Query) Conditions() []Condition {
...
 			}
+
 		case ruletime:
 			value, err := time.Parse(TimeLayout, buffer[begin:end])
 			if err != nil {
-				panic(fmt.Sprintf(
-					"got %v while trying to parse %s as time.Time / RFC3339 (should never happen if the grammar is correct)",
-					err,
-					buffer[begin:end]))
+				return false, fmt.Errorf("got %v while trying to parse %s as time.Time / RFC3339 (should never happen if the grammar is correct)", err, buffer[begin:end])
```
golangcibot: line is 158 characters (from `lll`)

libs/pubsub/query/query.go
```diff
@@ -160,131 +182,167 @@ func (q *Query) Conditions() []Condition {
...
 			}
+
 		case ruledate:
 			value, err := time.Parse("2006-01-02", buffer[begin:end])
 			if err != nil {
-				panic(fmt.Sprintf(
-					"got %v while trying to parse %s as time.Time / '2006-01-02' (should never happen if the grammar is correct)",
-					err,
-					buffer[begin:end]))
+				return false, fmt.Errorf("got %v while trying to parse %s as time.Time / '2006-01-02' (should never happen if the grammar is correct)", err, buffer[begin:end])
```
golangcibot: line is 163 characters (from `lll`)

libs/pubsub/pubsub.go
```diff
@@ -401,10 +404,16 @@ func (state *state) removeAll(reason error) {
...
 
-func (state *state) send(msg interface{}, events map[string][]string) {
+func (state *state) send(msg interface{}, events map[string][]string) error {
 	for qStr, clientSubscriptions := range state.subscriptions {
 		q := state.queries[qStr].q
-		if q.Matches(events) {
+
+		match, err := q.Matches(events)
+		if err != nil {
+			return fmt.Errorf("failed to match against query %s: %w", q.String(), err)
```
melekes:
```suggestion
			return errors.Wrapf(err, "failed to match against query %s", q.String())
```

libs/pubsub/query/query.go
```diff
@@ -301,87 +375,94 @@ func matchValue(value string, op Operator, operand reflect.Value) bool {
 			v, err = time.Parse(DateLayout, value)
 		}
 		if err != nil {
-			panic(fmt.Sprintf("failed to convert value %v from tag to time.Time: %v", value, err))
+			return false, fmt.Errorf("failed to convert value %v from event attribute to time.Time: %w", value, err)
```
melekes:
```suggestion
			return false, errors.Wrapf(err, "failed to convert value %v from event attribute to time.Time", value)
```

libs/pubsub/query/query.go
```diff
@@ -301,87 +375,94 @@ func matchValue(value string, op Operator, operand reflect.Value) bool {
...
 
+		operandFloat64 := operand.Interface().(float64)
+		filteredValue := numRegex.FindString(value)
+
 		// try our best to convert value from tags to float64
-		v, err := strconv.ParseFloat(value, 64)
+		v, err := strconv.ParseFloat(filteredValue, 64)
 		if err != nil {
-			panic(fmt.Sprintf("failed to convert value %v from tag to float64: %v", value, err))
+			return false, fmt.Errorf("failed to convert value %v from event attribute to float64: %w", filteredValue, err)
```
melekes:
```suggestion
			return false, errors.Wrapf(err, "failed to convert value %v from event attribute to float64", filteredValue)
```

libs/pubsub/query/query.go
```diff
@@ -301,87 +375,94 @@ func matchValue(value string, op Operator, operand reflect.Value) bool {
...
+		filteredValue := numRegex.FindString(value)
+
 		// if value looks like float, we try to parse it as float
-		if strings.ContainsAny(value, ".") {
-			v1, err := strconv.ParseFloat(value, 64)
+		if strings.ContainsAny(filteredValue, ".") {
+			v1, err := strconv.ParseFloat(filteredValue, 64)
 			if err != nil {
-				panic(fmt.Sprintf("failed to convert value %v from tag to float64: %v", value, err))
+				return false, fmt.Errorf("failed to convert value %v from event attribute to float64: %w", filteredValue, err)
```
melekes:
```suggestion
				return false, errors.Wrapf(err, "failed to convert value %v from event attribute to float64", filteredValue)
```

libs/pubsub/query/query.go
```diff
@@ -301,87 +375,94 @@ func matchValue(value string, op Operator, operand reflect.Value) bool {
...
+
 			v = int64(v1)
 		} else {
 			var err error
 			// try our best to convert value from tags to int64
-			v, err = strconv.ParseInt(value, 10, 64)
+			v, err = strconv.ParseInt(filteredValue, 10, 64)
 			if err != nil {
-				panic(fmt.Sprintf("failed to convert value %v from tag to int64: %v", value, err))
+				return false, fmt.Errorf("failed to convert value %v from event attribute to int64: %w", filteredValue, err)
```
melekes:
```suggestion
				return false, errors.Wrapf(err, "failed to convert value %v from event attribute to int64", filteredValue)
```

libs/pubsub/query/query.go
```diff
@@ -84,74 +89,91 @@ const (
 
 // Conditions returns a list of conditions.
 func (q *Query) Conditions() []Condition {
```
melekes: can we do the same here? return an error instead of panicking?

libs/pubsub/query/query.go
alexanderbez: Why this over what Go 1.13 recommends?

ref: https://blog.golang.org/go1.13-errors

alexanderbez:
R4R again :)

libs/pubsub/query/query.go
```diff
@@ -82,209 +89,283 @@ const (
...
+
 		case rulenumber:
 			number := buffer[begin:end]
 			if strings.ContainsAny(number, ".") { // if it looks like a floating-point number
 				value, err := strconv.ParseFloat(number, 64)
 				if err != nil {
-					panic(fmt.Sprintf("got %v while trying to parse %s as float64 (should never happen if the grammar is correct)",
-						err,
-						number))
+					err = fmt.Errorf(
```
melekes: may be a good idea to just write 

```
return nil, fmt.Errorf(...
```

to avoid variable shadowing

alexanderbez: Then you get linter errors. If you reformat the return to:

```go
return nil, fmt.Errorf(
  "got %v while trying to parse %s as int64 (should never happen if the grammar is correct)",
  err, number,
)
```

the linting is fixed, but then the return is harder to grok. Up to you.

melekes: let's stick with your version

Issue #4053, opened 2019-10-14
two p2p feature enhancement proposal


dlguddus:
I would like to suggest two p2p feature inhancement.

1. wildcard_peer_ids
- If id of a peer is in this list, the peer can try peering my node even if max_num_inbound(outbound)_peers is reached.
- benefit : nodes can allow peering with most trusted list of peers(wildcard peers) even if maximum is reached by other peers.

2. persistent_peers_dialing_max_period(in seconds)
- When exponential backoff, the dialing period will not increase more than this period
- benefit : Tendermint will continuesly try dialing all persistent peers indefinitely so that even though it disconnected for a while, it tries to reconnect.

Please share opinions(agreement, disagreement, additional feature related to these features, possible vulnerabilities, etc). Also if anyone wants to add other improvement on p2p peer management, we can add those to our job list.(such as blacklist_peer_ids)

These two ideas were one of the most pain-in-the-ass problems for real validator operation experience. It will greatly improve the stability of connection among nodes.

Based on feedbacks from community, we start hacking on these two issues. We will ask a grant of 2k atom from community fund by governance proposal "after" we PR this. We expect PR to be ready in October.

mdyring:
Good suggestions.

Not sure I understand the rationale for differentiating between persistent peers and "wildcard peers", is it because you don't want to provide a static IP for a peers that should always be allowed an inbound slot? 

I think the default behavior should be to allow explicitly configured peers a connection to slot (similar to what is suggested in #2041).

Instead of having both a "wildcard peers" and "persistent peers", perhaps something like this would work instead:

1) persistent_peers are used as today to establish outbound connections
2) if an inbound connection arrives with a nodeid listed in persistent_peers, and there is no existing outbound connection to that nodeid, it will be allowed an inbound slot
3) limits on in/outbound connections should be changed to "on-demand" connections and not apply to explicitly configured peers in the config.

persistent_peers_dialing_max_period sounds sensible!

Creamers158:
Agree, think you would get something like:
persistent_peers = "e4d19c423f4b2ft7e40905b320d54155907b6b54@23.34.567.34:26656,7f40e45c3e3fc1d4dc68edc977b8e3eeb4e50724,b2ft7e40905320d541423f4b2ft@75.34.564.67:26656"

This way you can also setup a connection direction without worrying about open peer slots.

[...]

PR #4072, opened 2019-10-22
Create adr-050-improved-trusted-peering.md

dlguddus:
Refs https://github.com/tendermint/tendermint/issues/4053

* [ ] Wrote tests


docs/architecture/adr-050-improved-trusted-peering.md
```diff
@@ -0,0 +1,59 @@
...
+
+This is happening because of two reasons, exponential backoff and absence of unconditional peering feature for trusted peers.
+
+
+## Decision
+
+We would like to suggest solving this problem by introducing two parameters in `config.toml`, `wildcard_peer_ids` and 
+`maximum_dial_period`. 
+
+1) `wildcard_peer_ids`
```
melekes: What about `private_peers_ids`? If we force the node to always connect to private peers, would that be sufficient?

docs/architecture/adr-050-improved-trusted-peering.md
```diff
@@ -0,0 +1,59 @@
...
+
+A node operator inputs list of ids of peers which are allowed to be connected by both inbound or outbound regardless of 
+`max_num_inbound_peers` or `max_num_outbound_peers` of user's node reached or not.
+
+2) `maximum_dial_period`
```
melekes: Shouldn't it be called `persistent_peers_max_dial_period` if it's only applicable to persistent peers?

docs/architecture/adr-050-improved-trusted-peering.md
```diff
@@ -0,0 +1,59 @@
...
+
+Terms between each dial to each persistent peer will not exceed `maximum_dial_period` during exponential backoff.
```
melekes: Maybe it's enough to extend the max from 24h to 48h? and lower the exponent?

docs/architecture/adr-050-improved-trusted-peering.md
dlguddus: Private and wildcard are having completely different definition. Private means not want to gossip, wildcard means allow connection anytime regardless of max connection.

Putting wildcard into private causes mix of two definitions.

docs/architecture/adr-050-improved-trusted-peering.md
dlguddus: Agree. We will fix the name. Thanks!

docs/architecture/adr-050-improved-trusted-peering.md
dlguddus: For validators, monitoring all connection everyday is not very trivial operation. From our experience, running tens of networks sometimes result in missing to alert relatively crucial connection for days. 

For persistent peers, we feel backoff until 30minute is more than enough, from our experience. Never want it to back off 24 or 48 hours.

docs/architecture/adr-050-improved-trusted-peering.md
dlguddus: Done

dlguddus:
@melekes @tessr @ebuchman 
Is the spec discussed enough for us to PR our implementation?
Waiting for confirmation.

docs/architecture/adr-050-improved-trusted-peering.md
```diff
@@ -0,0 +1,60 @@
+# ADR 50: Improved Trusted Peering
+
+## Changelog
+* 22-10-2019: Initial draft
+* 05-11-2019: Modify `maximum_dial_period` to `persistent_peers_max_dial_period`
+
+## Context
+
+When `max_num_inbound_peers` or `max_num_outbound_peers` of a node reached, the node cannot spare more slots to any peer
```
tessr:
```suggestion
When `max_num_inbound_peers` or `max_num_outbound_peers` of a node is reached, the node cannot spare more slots to any peer 
```

docs/architecture/adr-050-improved-trusted-peering.md
```diff
@@ -0,0 +1,60 @@
+# ADR 50: Improved Trusted Peering
+
+## Changelog
+* 22-10-2019: Initial draft
+* 05-11-2019: Modify `maximum_dial_period` to `persistent_peers_max_dial_period`
+
+## Context
+
+When `max_num_inbound_peers` or `max_num_outbound_peers` of a node reached, the node cannot spare more slots to any peer 
+by inbound or outbound. Therefore, after certain period of disconnection, any important peering can be lost indefinitely
```
tessr:
```suggestion
by inbound or outbound. Therefore, after a certain period of disconnection, any important peering can be lost indefinitely 
```

docs/architecture/adr-050-improved-trusted-peering.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Changelog
+* 22-10-2019: Initial draft
+* 05-11-2019: Modify `maximum_dial_period` to `persistent_peers_max_dial_period`
+
+## Context
+
+When `max_num_inbound_peers` or `max_num_outbound_peers` of a node reached, the node cannot spare more slots to any peer 
+by inbound or outbound. Therefore, after certain period of disconnection, any important peering can be lost indefinitely 
+because all slots consumed by other peers, and the node stop trying to dial the peer anymore.
```
tessr:
```suggestion
because all slots are consumed by other peers, and the node stops trying to dial the peer anymore.
```

docs/architecture/adr-050-improved-trusted-peering.md
tessr: I'm not sure wildcard is the best name here, either, since it often implies needing to match some pattern. I like "unconditional"--a word you used yourself a few paragraphs earlier!

docs/architecture/adr-050-improved-trusted-peering.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Decision
+
+We would like to suggest solving this problem by introducing two parameters in `config.toml`, `wildcard_peer_ids` and 
+`persistent_peers_max_dial_period`. 
+
+1) `wildcard_peer_ids`
+
+A node operator inputs list of ids of peers which are allowed to be connected by both inbound or outbound regardless of 
+`max_num_inbound_peers` or `max_num_outbound_peers` of user's node reached or not.
```
tessr: Are there limits on the number of trusted/wildcard peers a node can specify? And does adding a trusted/wildcard peer effectively increase the maximum number of peers by 1 (e.g. is the maximum number of inbound peers specified by `max_num_inbound_peers` + n, where n is the number of wildcard peers)?

docs/architecture/adr-050-improved-trusted-peering.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Status
+
+Proposed
+
+## Consequences
+
+### Positive
+
+A node operator can configure two new parameters in `config.toml` so that he/she can assure that tendermint will allow connection
```
tessr:
```suggestion
A node operator can configure two new parameters in `config.toml` so that he/she can assure that tendermint will allow connections
```

docs/architecture/adr-050-improved-trusted-peering.md
```diff
@@ -0,0 +1,60 @@
...
+from/to peers in `wildcard_peer_ids`. Also he/she can assure that every persistent peer will be dialed at least once in every 
+`persistent_peers_max_dial_period` term. It achieves more stable and persistent peering for trusted peers.
+
+### Negative
```
tessr: Are there any possible performance downsides?

docs/architecture/adr-050-improved-trusted-peering.md
dlguddus: 1. There is no limits for wildcard peers. I think it is OK without the limit. Do we need it?

2. The peers in wildcard_peer_ids is not objective to `max_num_inbound_peers`. So we don't add `n`. The state machine just does not check `max_num_inbound_peers` criteria for any wildcard_peer_ids. Also the state machine just does not count any peer in wildcard_peer_ids to compare `max_num_inbound_peers`

docs/architecture/adr-050-improved-trusted-peering.md
dlguddus: Additional calculation what we need for the new features
1) checking whether given peer is listed in `wildcard_peer_ids`
2) for `persistent_peers`, the state machine will dial indefinitely with given period

We think both calculation is not a significant factor for performance downsides.

docs/architecture/adr-050-improved-trusted-peering.md
dlguddus: unconditional it is! thanks for advice!

tessr:
I see that you checked the "referenced an issue" box, but I don't see the issue linked. Can you include it in the PR description and the ADR, please?

dlguddus:
I didn't realize I missed that. Thank you. Link for related issue added!

docs/architecture/adr-050-improved-trusted-peering.md
melekes: NOTE: we will have a problem with "wildcard_peer_ids is not objective to max_num_inbound_peers" because there's usually a limited number of connections per OS process (`ulimit`)

docs/architecture/adr-050-improved-trusted-peering.md
```diff
@@ -0,0 +1,60 @@
...
+
+The new feature introduces two new parameters in `config.toml` which needs explanation for node operators.
+
+### Neutral
+
+## References
+
+
+
+* two p2p feature inhancement proposal(https://github.com/tendermint/tendermint/issues/4053)
```
tessr: Can you clean up the whitespace here and also fix the spelling (no worries, but it should be "enhancement," thank you!).

tessr:
> Do you want me to also locate the link in another location? 

I think it would be helpful to put it in the PR description so that the Github UI links them together. Thanks!

docs/architecture/adr-050-improved-trusted-peering.md
tessr: It sounds like we should have some kind of limit here.

dlguddus: I think the OS limit is much far larger than what we need to connect practically. 200 peers are what we practically think maximum, and it is nowhere close to default limit on OS.

Also users can aware that maxpeernum + numofunconditionalpeers = actual possible number of peers. So it is not a unpredictable risk to manage.

For such reason, I think it can be stated in ADR, but we dont think it is needed to be implemented in this spec.

melekes:
> I think it would be helpful to put it in the PR description so that the Github UI links them together. Thanks!

Done.

docs/architecture/adr-050-improved-trusted-peering.md
melekes: Fixed.

PR #4077, opened 2019-10-24
Add EXISTS condition to query grammar

haasted:
This PR adds an "EXISTS" condition to the event query grammar. It enables querying for the occurrence of an event without having to provide a condition for one of its attributes.

As an example, someone interested in all slashing events might currently catch them with a query such as `slash.power > 0`. 

With this PR the event can be captured with `slash.power EXISTS` or just `slash EXISTS` to catch by event type. 

* [X] Wrote tests


libs/pubsub/query/query.peg.go
```diff
@@ -139,55 +188,180 @@ func (t *tokens32) AST() *node32 {
...
+		close(s)
+	}()
+	return s, ordered
+}
+
+func (t *tokens32) PrintSyntax() {
+	tokens, ordered := t.PreOrder()
+	max := -1
+	for token := range tokens {
+		if !token.leaf {
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

libs/pubsub/query/query.peg.go
```diff
@@ -139,55 +188,180 @@ func (t *tokens32) AST() *node32 {
...
 }
 
 func (t *tokens32) PrintSyntaxTree(buffer string) {
-	t.AST().Print(buffer)
+	tokens, _ := t.PreOrder()
+	for token := range tokens {
+		for c := 0; c < int(token.next); c++ {
+			fmt.Printf(" ")
+		}
+		fmt.Printf("\x1B[34m%v\x1B[m %v\n", rul3s[token.pegRule], strconv.Quote(string(([]rune(buffer)[token.begin:token.end]))))
```
golangcibot: line is 123 characters (from `lll`)

libs/pubsub/query/query.peg.go
```diff
@@ -409,53 +581,57 @@ func (p *QueryParser) Init() {
...
+		/* 1 condition <- <(tag ' '* ((le ' '* ((&('D' | 'd') date) | (&('T' | 't') time) | (&('0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9') number))) / (ge ' '* ((&('D' | 'd') date) | (&('T' | 't') time) | (&('0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9') number))) / ((&('E' | 'e') exists) | (&('=') (equal ' '* ((&('\'') value) | (&('D' | 'd') date) | (&('T' | 't') time) | (&('0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9') number)))) | (&('>') (g ' '* ((&('D' | 'd') date) | (&('T' | 't') time) | (&('0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9') number)))) | (&('<') (l ' '* ((&('D' | 'd') date) | (&('T' | 't') time) | (&('0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9') number)))) | (&('C' | 'c') (contains ' '* value)))))> */
```
golangcibot: line is 786 characters (from `lll`)

libs/pubsub/query/query.peg.go
```diff
@@ -54,64 +56,74 @@ var rul3s = [...]string{
...
+func (t *token32) isZero() bool {
+	return t.pegRule == ruleUnknown && t.begin == 0 && t.end == 0 && t.next == 0
+}
+
+func (t *token32) isParentOf(u token32) bool {
+	return t.begin <= u.begin && t.end >= u.end && t.next > u.next
+}
+
+func (t *token32) getToken32() token32 {
+	return token32{pegRule: t.pegRule, begin: uint32(t.begin), end: uint32(t.end), next: uint32(t.next)}
```
golangcibot: unnecessary conversion (from `unconvert`)

libs/pubsub/query/query.peg.go
```diff
@@ -139,55 +188,180 @@ func (t *tokens32) AST() *node32 {
...
+	for token := range tokens {
+		for c := 0; c < int(token.next); c++ {
+			fmt.Printf(" ")
+		}
+		fmt.Printf("\x1B[34m%v\x1B[m %v\n", rul3s[token.pegRule], strconv.Quote(string(([]rune(buffer)[token.begin:token.end]))))
+	}
+}
+
+func (t *tokens32) Add(rule pegRule, begin, end, depth uint32, index int) {
+	t.tree[index] = token32{pegRule: rule, begin: uint32(begin), end: uint32(end), next: uint32(depth)}
```
golangcibot: unnecessary conversion (from `unconvert`)

libs/pubsub/query/query.peg.go
```diff
@@ -741,28 +923,128 @@ func (p *QueryParser) Init() {
...
+									if buffer[position] != rune('S') {
+										goto l16
+									}
+									position++
+								}
+							l51:
+								depth--
+								add(ruleexists, position40)
+							}
+							break
```
golangcibot: S1023: redundant break statement (from `gosimple`)

libs/pubsub/query/query.peg.go
```diff
@@ -54,64 +56,74 @@ var rul3s = [...]string{
...
+	down *element
+}
+
+/* ${@} bit structure for abstract syntax tree */
+type token32 struct {
+	pegRule
+	begin, end, next uint32
+}
+
+func (t *token32) isZero() bool {
```
golangcibot: func `(*token32).isZero` is unused (from `unused`)

libs/pubsub/query/query.go
```diff
@@ -195,6 +197,28 @@ func (q *Query) Matches(events map[string][]string) bool {
 			op = OpEqual
 		case rulecontains:
 			op = OpContains
+		case ruleexists:
```
alexanderbez: While I can't speak for the rest of the changes, this makes sense to me.

melekes:
Alternative syntax proposal: `slash.power = *` (refs: https://github.com/tendermint/tendermint/issues/1421)

melekes:
I am not entirely sure js-query is a good idea, but extending current format without clear understanding how it will evolve is either bad.

melekes:
`key EXISTS` looks fine to me though

libs/pubsub/query/query.go
```diff
@@ -73,6 +73,8 @@ const (
 	OpEqual
 	// "CONTAINS"; used to check if a string contains a certain sub string.
 	OpContains
+	// "EXISTS"; used to check whether an event has taken place
```
melekes:
```suggestion
	// "EXISTS"; used to check if a certain event attribute is present.
```

libs/pubsub/query/query.go
```diff
@@ -195,6 +197,28 @@ func (q *Query) Matches(events map[string][]string) bool {
 			op = OpEqual
 		case rulecontains:
 			op = OpContains
+		case ruleexists:
+			op = OpExists
+			if strings.Contains(tag, ".") {
+				// Searching for a full "type.attribute" event.
+				_, ok := events[tag]
+				if !ok {
```
melekes:
```suggestion
				return ok
```

libs/pubsub/query/query.go
```diff
@@ -195,6 +197,28 @@ func (q *Query) Matches(events map[string][]string) bool {
...
+					return false
+				}
+			} else {
+				foundEvent := false
```
melekes:
```
for eventType := range events {
  if eventType == tag {
    return true
  }
}
return false
```

haasted: It can't actually be simplified that much. It's important that on a positive match, the evaluation continues. Consider the query `slash.reason EXISTS AND slash.power > 1000` on an event where `slash.power` is `500`. First part of the query matches, but the second won't.

melekes: Right, thanks!

libs/pubsub/query/query.go
```diff
@@ -195,6 +197,28 @@ func (q *Query) Matches(events map[string][]string) bool {
...
+
+			loop:
+				for k := range events {
+					if strings.Index(k, tag) == 0 {
```
melekes: why Index? Shouldn't we use strict comparison here ?

```suggestion
					if k == tag {
```

libs/pubsub/query/query.go
```diff
@@ -195,6 +197,28 @@ func (q *Query) Matches(events map[string][]string) bool {
...
+				// Searching for a full "type.attribute" event.
+				_, ok := events[tag]
+				if !ok {
+					return false
+				}
+			} else {
+				foundEvent := false
+
+			loop:
+				for k := range events {
```
melekes:
```suggestion
				for eventType := range events {
```

libs/pubsub/query/query.go
haasted: This is the case where the query is the "event type", e.g. just "slash". The contents of the `events` slice are the fully qualified event attributes, e.g. `slash.reason`. The index is used to verify that the event starts with the tag from the query.

libs/pubsub/query/query.go
haasted: I can see the point of picking a better variable name than just `k`. `eventType` is not the correct term, I believe. Perhaps a better name would be `compositeTag`, which seems to be the term used in `event_bus.go` line 124.

haasted: Changed variable name to `compositeTag`.

libs/pubsub/query/query.go
melekes: Won't it lead to false positives? I.e. `sl EXISTS` will work for `slash.reason`, but I doubt that it's what we want.

libs/pubsub/query/query.go
melekes: eventType is the correct term

libs/pubsub/query/query.go
haasted: `sl EXISTS` is valid syntax in the current implementation. Not sure whether I would consider it a problem.

libs/pubsub/query/query.go
haasted: Are you certain? The `events` slice will contain entries of the form `<eventType>.<eventAttribute>` as outlined in the `validateAndStringifyEvents` function of the event bus.

libs/pubsub/query/query.go
```diff
@@ -233,6 +240,28 @@ func (q *Query) Matches(events map[string][]string) (bool, error) {
...
+
+			loop:
+				for compositeTag := range events {
+					if strings.Index(compositeTag, eventAttr) == 0 {
+						foundEvent = true
+						break loop
+					}
+				}
+				if !foundEvent {
+					return false,nil
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
					return false, nil
```

libs/pubsub/query/query.go
melekes: oh, so it's a composite key (`type.attribute`). then `compositeTag` or better `compositeKey` make total sense!

libs/pubsub/query/query.go
melekes: This needs to be reflected both in the docs https://github.com/tendermint/tendermint/blob/2df4ca0f3924961dfbd3cd66e1b530fd20909640/rpc/swagger/swagger.yaml#L155 and in the tests. Thank you 🙇

melekes:
https://github.com/tendermint/tendermint/pull/4077#discussion_r341857177 needs to be addressed before we can merge.

haasted:
Addressed comments.

PR #4084, opened 2019-10-28
Added make command to install the tendermint binary

vaibhavsinha:
I noticed that the Getting Started doc was missing the step to install the tendermint binary

* [ ] Wrote tests

### DIFF
--- a/docs/app-dev/getting-started.md
+++ b/docs/app-dev/getting-started.md
@@ -33,6 +33,7 @@ go get github.com/tendermint/tendermint
 cd $GOPATH/src/github.com/tendermint/tendermint
 make tools
 make install_abci
+make install
 ```
 
 Now you should have the `abci-cli` installed; you'll see a couple of


melekes:
This section shows how to install abci-cli, not Tendermint. The user is supposed to already have Tendermint installed from previous chapters (https://github.com/tendermint/tendermint/blob/5b601dc2e273c30f393e396f8b52c18bc5781fc6/docs/introduction/install.md).

Maybe we need to add an "Assumptions" section at the top saying "You should have a Tendermint binary installed. If not, follow the steps from ../introduction/install.md".

vaibhavsinha:
@melekes: You're right. I was in the wrong part of the documentation. An assumption under the "Install" section should do the job for people in a hurry like me :-) I'll close this PR and send another one with your suggestion.

melekes:
Thank you 👍

PR #4095, opened 2019-10-31
Fix error handling of TxIndex.Search

hsyis:
This PR fixes error handling for performing a txindex search.

`TxIndex.Get` returns
* `(txresult, nil)` if the transaction is found.
* `(nil, nil)` if the transaction is not found.
* `(nil, error)` if error is occurred.

Therefore, if `res` is not `nil`, I think `TxIndex.Search` should return `(txresult, nil)`.

Previously, however, this was not a problem because `errors.Wrap` returns `nil` if its first argument `err` is `nil`.

* [ ] Wrote tests


state/txindex/kv/kv.go
```diff
@@ -180,9 +180,9 @@ func (txi *TxIndex) Search(q *query.Query) ([]*types.TxResult, error) {
 	} else if ok {
 		res, err := txi.Get(hash)
 		if res == nil {
```
melekes: what do you think about using a switch here?

```go
switch {
	case err != nil:
		return []*types.TxResult{}, errors.Wrapf(err, "can't get tx %X", hash)
	case res == nil:
		return []*types.TxResult{}, nil
	default:
		return []*types.TxResult{res}, nil
```

hsyis: Good comment :) I think it's nice to have explicit error handling for each state.
I'll change to use the switch statement.

melekes:
Could you add an entry to CHANGELOG_PENDING.md?

```
- [state] txindex/kv: Return an error if there's one when the user searches for a tx (hash=X) (@hsyis)
```

PR #4103, opened 2019-11-01
Remove unnecessary error

cosmostuba:
* [ ] Wrote tests

Currently, this is logging "Write::Error Resetting deadline" even when there is no error. 

We already enforce that `se.timeoutReadWrite` is of type time.Duration. The time.Add method should not return errors. https://golang.org/pkg/time/#example_Time_Add

Hence, we can remove it. This is a tiny one-line change, so I suppose it does not need the following checkmarks?


privval/signer_endpoint.go
```diff
@@ -122,8 +122,7 @@ func (se *signerEndpoint) WriteMessage(msg SignerMessage) (err error) {
 
 	// Reset read deadline
 	deadline := time.Now().Add(se.timeoutReadWrite)
-	se.Logger.Debug("Write::Error Resetting deadline", "obj", se)
-
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

cosmostuba:
Hi, I'm not sure why this is failing the deploy_docs CI. It should be a trivial one-line change.

PR #4106, opened 2019-11-04
Fix trivial typo

corollari:
PR is trivial


corollari:
Just wanted to chime in to say that this has been the best experience I've had submitting a PR, kudos!

Issue #3333, opened 2019-02-20
rpc: add possibility of sorting /tx_search results


gianfelipe93:
Hi guys,

I am using the function TxSearch to query transactions from the blockchain and now I need to get the latest 5 transactions to populate a list. However, I couldn't find a way to sort the transactions so I am here to ask if there's a way to that currently.

I am running the version 0.30.0

Thanks for your help,
Gian.

melekes:
There's no way currently to sort txs returned by /tx_search.

gianfelipe93:
Hi Melekes,

Thank you for helping me.

Since I want to get the latest 5 transactions, sorting by `asc` should do the job.

For now I will calculate what the last page is and get the transactions using `TxSearch` .

I am closing the issue since the question has been answered.

Thanks you again Melekes,
Gian.

melekes:
We still need to implement the sorting. I'll update the title.

thith:
Is there a progress on this?
It is necessary feature because apps like wallet and scan need to show transactions in the latest-first order. It is not only '5 first transactions', we need paging in that order as well.
How does cosmos scan/wallet handle this?

PR #4109, opened 2019-11-05
RPC:tx_search show txs in the latest-first order

preminem:
This PR is to solve the  TxSearch problem #3333 .In most cases, we need to show transactions in the latest-first order.
* [ ] Wrote tests

### DIFF
--- a/rpc/core/tx.go
+++ b/rpc/core/tx.go
@@ -212,7 +212,8 @@ func TxSearch(ctx *rpctypes.Context, query string, prove bool, page, perPage int
 	var proof types.TxProof
 	// if there's no tx in the results array, we don't need to loop through the apiResults array
 	for i := 0; i < len(apiResults); i++ {
-		r := results[skipCount+i]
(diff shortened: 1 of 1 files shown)


melekes:
Thank you for the PR, but the correct way to address the issue I think is to introduce the `sort` flag https://github.com/tendermint/tendermint/issues/3333#issuecomment-465545500, not to change the default order.

melekes:
Plus, if we've wanted to change the default order, the right way to do it is to change Sort func here https://github.com/tendermint/tendermint/blob/274447e2b0a5caf6d4b285e2604cd3e76c54c547/state/txindex/kv/kv.go#L251, not to revert the resulting slice.

preminem:
@melekes Thank you for your reply. I agree with you.

melekes:
Feel free to send another PR implementing https://github.com/tendermint/tendermint/pull/4109#issuecomment-549739829. I'll close this PR in the meantime.

Issue #2358, opened 2018-09-07
Unable to inject own Key Type 


bgrieder:
Tendermint version 0.23.1

I would like to inject my own Public and Private key types because I do not want to use the ED25519 curve.

The embedded doc mentions it is possible to inject its own private validator as long as it implements the `types.PrivValidator` interface.

Unfortunately, this is currently impossible. 

The `typesGenesisDoc`requires an array of validators which can be bootsrapped using something like

```go
genDoc.Validators = []types.GenesisValidator{{
			PubKey: privValidator.GetPubKey(),  // <-- this pulls the "new" Public Key
			Power:  10,
		}}
```

The `Node` is instantiated using a (massive) `NewNode(...)` function that attempts to serialize the `GenesisDoc` in the state DB using a piece of code that does

```go
	genDoc, err := loadGenesisDoc(stateDB)
	if err != nil {
		genDoc, err = genesisDocProvider()
		if err != nil {
			return nil, err
		}
		// save genesis doc to prevent a certain class of user errors (e.g. when it
		// was changed, accidentally or not). Also good for audit trail.
		saveGenesisDoc(stateDB, genDoc)  // <---- THE PROBLEMATIC CALL
	}
```

To serialize the doc, `amino`is called. 
Unfortunately `amino`requires all serializable "concrete" types to be pre-registered and the piece of code that does this is, is located in the `wire.go` file in the `node` package in an `init()` call, namely

```go
var cdc = amino.NewCodec()

func init() {
	cryptoAmino.RegisterAmino(cdc) // <--- this registers all the types
}
```
As currently written, there is no way to inject custom types either by getting hold of `cdc`, which is private to the package, or via a provided callback.

I believe this static initialization should be revisited and `cdc` should be instantiated early and injected in the new node as part of the `NewNode` function parameters.

bgrieder:
The problem is pretty widespread. There are many files, usually named `wire.go` that register types in the `init()` function.
This causes two problems 1) a lot of duplicated calls to the types registration 2) the inability to inject your own types.

To fix 1) I guess the only robust solution is to used injection (much like is down for `StateDB` rather than using static initialization. I have tried but the impact on existing function signatures is very significant, especially around `State`. This would be worth the effort though to avoid bad unmarshalling surprises in the future by re-centralizing his code around a single function and instance.

[...]

PR #4112, opened 2019-11-05
Function to modify key codec

austinabell:
Hey, not sure if this is disallowed for any reason specifically, but it would be very beneficial to define additional types to decode tendermint key implementations from bytes, since it uses a static codec. If this is okay, let me know and I will add documentation.

Context: For Ethermint to switch to using Cosmos' keybase, decoding the keys requires this codec to be updated

* [ ] Wrote tests


melekes:
this needs a CHANGELOG_PENDING.md entry

```
- [crypto] Add `RegisterKeyType` to amino to allow external key types registration (@austinabell)
```

melekes:
Refs https://github.com/tendermint/tendermint/issues/2358

austinabell:
Just to document, I did experiment with creating a mapping from string to objects to be able to keep track of the key types added to be able to be used in the `RegisterAmino(..)` call, but because of how go is compiled, cosmos would just use the base types. This may be a useful feature for someone just building on top of Tendermint and not going through Cosmos, but to not add confusion or unnecessary complexity, I left it out.

crypto/encoding/amino/amino.go
```diff
@@ -60,11 +60,20 @@ func RegisterAmino(cdc *amino.Codec) {
 		secp256k1.PrivKeyAminoName, nil)
 }
 
+// RegisterKeyType registers an external key type to allow decoding it from bytes
+func RegisterKeyType(o interface{}, name string) {
+	cdc.RegisterConcrete(o, name, nil)
+
+	nameTable[reflect.TypeOf(o)] = name
```
melekes: Should we check if an entry with the same name already exists before possibly overriding it?

crypto/encoding/amino/amino.go
```diff
@@ -60,11 +60,20 @@ func RegisterAmino(cdc *amino.Codec) {
 		secp256k1.PrivKeyAminoName, nil)
 }
 
+// RegisterKeyType registers an external key type to allow decoding it from bytes
+func RegisterKeyType(o interface{}, name string) {
```
melekes: Would be great to have a test case for it.

crypto/encoding/amino/amino.go
austinabell: Yeah, I'll do that now. Do you want me to use the added types to be used in `RegisterAmino`?

crypto/encoding/amino/amino.go
austinabell: Will do

austinabell:
Sorry that took a little longer than expected, test setup was a little finicky and the print tests required me to make sure the codec was reset after the test. Alternatively I could have changed the expected print output but I didn't want to change any existing tests

crypto/encoding/amino/amino.go
```diff
@@ -62,8 +62,11 @@ func RegisterAmino(cdc *amino.Codec) {
 
 // RegisterKeyType registers an external key type to allow decoding it from bytes
 func RegisterKeyType(o interface{}, name string) {
-	cdc.RegisterConcrete(o, name, nil)
+	if _, exists := nameTable[reflect.TypeOf(o)]; exists {
+		return
```
melekes: I think we should panic, not silently return

crypto/encoding/amino/encode_test.go
```diff
@@ -147,3 +149,75 @@ func TestPubkeyAminoName(t *testing.T) {
...
+}
+func (key TestPub) VerifyBytes(msg []byte, sig []byte) bool { return true }
+func (key TestPub) Equals(other crypto.PubKey) bool         { return true }
+
+var (
+	privAminoName = "registerTest/Priv"
+	pubAminoName  = "registerTest/Pub"
+)
+
+func TestRegisterKeyType(t *testing.T) {
```
melekes: this is great!

crypto/encoding/amino/amino.go
austinabell: Yeah, I can do that, but isn't that basically the same panic that occurs in `RegisterConcrete` when you try to register an existing type?

melekes: Did not know we panic in Register concrete. If that's the case, we can get rid of the extra check. Thank you

crypto/encoding/amino/encode_test.go
```diff
@@ -2,10 +2,12 @@ package cryptoAmino
 
 import (
 	"os"
+	"reflect"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	amino "github.com/tendermint/go-amino"
```
melekes:
```suggestion

	amino "github.com/tendermint/go-amino"

```

ebuchman:
Does this actually work in a real case? I thought the codecs are declared privately on a per package basis so it would be hard to get this effect to thread through ?

austinabell:
> Does this actually work in a real case? I thought the codecs are declared privately on a per package basis so it would be hard to get this effect to thread through ?

Does for what I needed it for in Ethermint, can't say for other uses. If defined in the main() function of the cli and daemon binary as such:
```
	tmamino.RegisterKeyType(emintcrypto.PubKeySecp256k1{}, emintcrypto.PubKeyAminoName)
	tmamino.RegisterKeyType(emintcrypto.PrivKeySecp256k1{}, emintcrypto.PrivKeyAminoName)
```

It is able to be overriden for when a public key is decoded through the `..FromBytes` functions using that codec.

I have noticed other interactions haven't threaded through based on how dependencies of dependencies are compiled in go which made other things more annoying, but this PR allowed the Ethermint key type to be decoded from the Cosmos keybase.

PR #4121, opened 2019-11-08
Bump grpc dep

marbar3778:
- bump grpc dep as listed here: https://github.com/tendermint/tendermint/pull/4114

* [ ] Wrote tests


tessr:
What's the difference between this PR and #4114? Just trying to understand how Dependabot works and how we use it

marbar3778:
dependabot wasnt able to update go.sum correctly, so all ci was failing, had to jump in manually

tessr:
Got it. Thanks for the explanation!

PR #4126, opened 2019-11-11
rpc: remove godoc comments in favor of swagger docs

melekes:
No need to duplicate information in this case. It a) requires extra
efforts to keep both in sync b) nobody reads godoc documentation
anyways.

* [ ] ~~Referenced an issue explaining the need for the change~~
* [ ] ~~Wrote tests~~


tessr:
How did we decide to use Swagger over Godoc?

tessr:
Ah, @marbar3778 just explained that this is _only_ for the RPC endpoints (which I would've known if I'd read your PR title 🤪).

Issue #4051, opened 2019-10-11
`rpcclient` resolve https protocol to http


xiaoxiaff:
**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):
github.com/tendermint/tendermint v0.32.5

**ABCI app** (name for built-in, URL for self-written if it's publicly available):

**Environment**:
- **OS** (e.g. from /etc/os-release):
- **Install tools**:
- **Others**:

**What happened**:
When we set up a rpc client, we set node url to "https://fullnode.lino.network:443" then call NewHTTP function:
```
import (
	rpcclient "github.com/tendermint/tendermint/rpc/client"
)

rpc := rpcclient.NewHTTP("https://fullnode.lino.network:443", "/websocket")
```

In tendermint rpc lib it will be parsed to `http://fullnode.lino.network:443` which will fail the rpc call with error `ABCIQuery: error unmarshalling rpc response: invalid character '<' looking for beginning of value`

**What you expected to happen**:
Protocol should be parsed to https.

**Have you tried the latest version**: yes/no
yes

**How to reproduce it** (as minimally and precisely as possible):

**Logs (paste a small part showing an error (< 10 lines) or link a pastebin, gist, etc. containing more of the log file)**:

**Config (you can paste only the changes you've made)**:

**node command runtime flags**:

**`/dump_consensus_state` output for consensus bugs**

**Anything else we need to know**:
Code causes the issue: https://github.com/tendermint/tendermint/blob/v0.32.6/rpc/lib/client/http_client.go#L87

melekes:
We should add `rpcclient.NewHTTPS` and return an error if HTTPS is passed [here](https://github.com/tendermint/tendermint/blob/v0.32.6/rpc/lib/client/http_client.go#L87)

@xiaoxiaff thanks for pointing this out!

githubsands:
Hey guys can you assign me to this issue?

Thanks

melekes:
We can only assign people from Tendermint org, but you can help us fix ^ nevertheless.

githubsands:
Sounds good

pasDamola:
hi @melekes, could you point me to where the **NewHTTP** function is?

pasDamola:
hi @xiaoxiaff @melekes please could you point me to where the **NewHTTP** function is?

melekes:
[...]

PR #4131, opened 2019-11-12
fix issue #4051

yenkhoon:
This fix the issue #4051 .
function parseRemoteAddr is forcing protocolHTTP and protocolHTTPs to tcp. this causes the bug in the issue #4051 .

I find that the tcp is only needed where net.Dial. So I moved the switch case in to makeHTTPDialer . 

* [ ] Wrote tests


melekes:
Have you tested it? Do you want to try and write unit tests?

yenkhoon:
I have tested it. I just push a simple unit test for it.

rpc/lib/client/http_client_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestHTTPClientMakeHTTPDialer(t *testing.T) {
+	remote := []string{"https://foo-bar.com:80", "http://foo-bar.net:80"}
+
+	for _, f := range remote {
+		protocol, address, err := parseRemoteAddr(f)
+		require.NoError(t, err)
+		conn := makeHTTPDialer(f)
```
melekes:
```suggestion
		dialer := makeHTTPDialer(f)
```

melekes:
Could you please add 

```
- [rpc/lib] \#4051 Fix RPC client, which was previously resolving https protocol to http (@yenkhoon)
```

to CHANGELOG_PENDING.md under BUGFIXES?

yenkhoon:
> Could you please add
> 
> ```
> - [rpc/lib] \#4051 Fix RPC client, which was previously resolving https protocol to http (@yenkhoon)
> ```
> 
> to CHANGELOG_PENDING.md under BUGFIXES?

added

yenkhoon:
> Thanks for contributing to Tendermint 👍

no problem :)

PR #4141, opened 2019-11-14
rpc/lib/client & server: try to conform to JSON-RPC 2.0 spec

### LINKED ISSUE #2949: Responses to "subscribe" sent over WebSocket don't conform to RPC 2.0 specification

melekes:
Redo of https://github.com/tendermint/tendermint/pull/3858

https://www.jsonrpc.org/specification

What is done in this PR:

    JSONRPCClient: validate that Response.ID matches Request.ID I wanted
    to do the same for the WSClient, but since we're sending events as
    responses, not notifications, checking IDs would require storing
    them in memory indefinitely (and we won't be able to remove them
    upon client unsubscribing because ID is different then).

    Request.ID is now optional. Notification is a Request without an ID.
    Previously "" or 0 were considered as notifications

    Remove #event suffix from ID from an event response (partially fixes
    #2949) ID must be either string, int or null AND must be equal to
    request's ID. Now, because we've implemented events as responses, WS
    clients are tripping when they see Response.ID("0#event") !=
    Request.ID("0"). Implementing events as requests would require a lot
    of time (~ 2 days to completely rewrite WS client and server)

    generate unique ID for each request

    switch to integer IDs instead of "json-client-XYZ"
    id=0 method=/subscribe
    id=0 result=...
    id=1 method=/abci_query
    id=1 result=...

> send events (resulting from /subscribe) as requests+notifications (not
responses)

this will require a lot of work. probably not worth it

* rpc: generate an unique ID for each request
in conformance with JSON-RPC spec

* WSClient: check for unsolicited responses

* fix golangci warnings

* save commit

* fix errors

* remove ID from responses from subscribe
Fixes #2949

* clients are safe for concurrent access

* tm-bench: switch to int ID

* fixes after my own review

* comment out sentIDs in WSClient
see commit body for the reason

* remove body.Close
it will be closed automatically

* stop ws connection outside of write/read routines
also, use t.Rate in tm-bench indexer when calculating ID

* fix gocritic issues

* [x] Wrote tests


rpc/lib/server/ws_handler.go
```diff
@@ -0,0 +1,448 @@
...
+					wsc.Logger.Info("Client closed the connection")
+				} else {
+					wsc.Logger.Error("Failed to read request", "err", err)
+				}
+				wsc.Stop()
+				close(wsc.readRoutineQuit)
+				return
+			}
+
+			err = json.Unmarshal(in, request)
```
golangcibot: SA1014: json.Unmarshal expects to unmarshal into a pointer, but the provided value is not a pointer (from `staticcheck`)

rpc/lib/server/http_uri_handler.go
```diff
@@ -0,0 +1,196 @@
...
+		expectingInt = true
+	case reflect.String:
+		expectingString = true
+	case reflect.Slice:
+		expectingByteSlice = rt.Elem().Kind() == reflect.Uint8
+	}
+
+	if isIntString && expectingInt {
+		qarg := `"` + arg + `"`
+		// jsonStringToArg
```
melekes:
```suggestion
```

rpc/lib/server/ws_handler.go
```diff
@@ -0,0 +1,448 @@
...
+	if wsc.ctx != nil {
+		return wsc.ctx
+	}
+	wsc.ctx, wsc.cancel = context.WithCancel(context.Background())
+	return wsc.ctx
+}
+
+// Read from the socket and subscribe to or unsubscribe from events
+func (wsc *wsConnection) readRoutine() {
+	var request types.RPCRequest
```
NicolasMahe: a bug has been introduced by moving this variable outside of the for loop (because it is used in the defer function to display the `request.ID` in the error message).
When one ws client (one connection) wants to subscribe to multiple events (multiple call to `Subscribe` method), the ws messages always contains the ID for the last subscriber.
previous place where this variable was declared: https://github.com/tendermint/tendermint/pull/4141/files#diff-692f910080346cdb0d29bfba93c21569L701

@antho1404 is working on a fix and should create a PR soon.

melekes: Good catch! Thanks

NicolasMahe: PR in progress: https://github.com/tendermint/tendermint/pull/4406

Issue #4143, opened 2019-11-14
Can submit DuplicateVoteEvidence twice by Swapping Vote order


AdityaSripal:
**Tendermint version** `master`

From my understanding of how the evidence pool/reactors work: 
It is possible to submit two valid DoubleVoteEvidence to tendermint for the same double-sign event by swapping the order of conflicting votes
Since they hash to different values, both evidences will get added to evidence pool where they can be reaped by proposer into a block. The block executor will then convert both evidences into abci.Evidence and send them in RequestBeginBlock

Links to helpful parts of code:
1. DuplicateVoteEvidence: 
https://github.com/tendermint/tendermint/blob/master/types/evidence.go#L98
2. AddEvidence: 
https://github.com/tendermint/tendermint/blob/master/evidence/pool.go#L97
https://github.com/tendermint/tendermint/blob/b43da179583ef005acab32166a2922cb88ee3504/evidence/store.go#L139
3. Block Reaping from EvidencePool:
https://github.com/tendermint/tendermint/blob/b43da179583ef005acab32166a2922cb88ee3504/state/execution.go#L100
4. Sending Evidence to App Server:
https://github.com/tendermint/tendermint/blob/b43da179583ef005acab32166a2922cb88ee3504/state/execution.go#L276
https://github.com/tendermint/tendermint/blob/b43da179583ef005acab32166a2922cb88ee3504/state/execution.go#L348
https://github.com/tendermint/tendermint/blob/b43da179583ef005acab32166a2922cb88ee3504/types/protobuf.go#L142

### Proposed Solution

Lexicographically sort votes in the `DuplicateVoteEvidence` on hash of vote and enforce this property in `ValidateBasic`

This enforces a single valid ordering of votes in `DuplicateVoteEvidence` and removes this attack, since wrong ordering will be rejected from `EvidencePool`

AdityaSripal:
cc: @alexanderbez

PR #4151, opened 2019-11-16
Enforce Ordering in DuplicateVoteEvidence

AdityaSripal:
Fixes issue described in: #4143 

* [x] Wrote tests


types/evidence.go
```diff
@@ -209,6 +231,10 @@ func (dve *DuplicateVoteEvidence) ValidateBasic() error {
 	if err := dve.VoteB.ValidateBasic(); err != nil {
 		return fmt.Errorf("Invalid VoteB: %v", err)
 	}
+	// Enforce Votes are lexicographically sorted on blockID
+	if strings.Compare(dve.VoteA.BlockID.Key(), dve.VoteB.BlockID.Key()) >= 0 {
+		return errors.New("Duplicate Votes in invalid order")
```
melekes:
```suggestion
		return errors.New("duplicate votes in invalid order")
```

melekes:
This also needs a CHANGELOG_PENDING.md entry.

AdityaSripal:
@melekes Would this be considered a breaking change? I don't believe it is, since the submission of evidence is taken care of completely inside tendermint

melekes:
> @melekes Would this be considered a breaking change? I don't believe it is, since the submission of evidence is taken care of completely inside tendermint

no, just a bugfix

types/evidence.go
```diff
@@ -209,6 +231,10 @@ func (dve *DuplicateVoteEvidence) ValidateBasic() error {
 	if err := dve.VoteB.ValidateBasic(); err != nil {
 		return fmt.Errorf("Invalid VoteB: %v", err)
 	}
+	// Enforce Votes are lexicographically sorted on blockID
+	if strings.Compare(dve.VoteA.BlockID.Key(), dve.VoteB.BlockID.Key()) >= 0 {
```
melekes: This is actually breaking since old evidence may not be in correct order.

ebuchman:
We should update the spec for this here: https://github.com/tendermint/tendermint/blob/master/docs/spec/blockchain/blockchain.md#evidence-1

melekes:
@ebuchman thanks for pointing this out!

PR #4163, opened 2019-11-19
consensus: don't panic when block is not found in store

### LINKED ISSUE #4069: Don't panic when block is not found in store
Tendermint currently crashes when the consensus reactor wants to help a peer catch up and send block parts which it does not hold in the store.

At Certus One we wipe the blockstore of some of our nodes to create small backups and allow quicker redeployments which is important to quickly react to events like DDoS or major outages. 

If a block is not found, the consensus reactor should not crash since this does not affect the ability to participate in consensus. Instead it should wait for the gossipSleepDuration and continue.

https://github.com/tendermint/tendermint/blob/288f1a0bc66499b71d688c80768a9e91173e0853/consensus/reactor.go#L503-L506

gregzaitsev:
* [X] Referenced an issue explaining the need for the change
* [ ] Wrote tests

Fixes #4069


consensus/reactor.go
```diff
@@ -501,10 +501,14 @@ OUTER_LOOP:
...
 				blockMeta := conR.conS.blockStore.LoadBlockMeta(prs.Height)
 				if blockMeta == nil {
-					panic(fmt.Sprintf("Failed to load block %d when blockStore is at %d",
-						prs.Height, conR.conS.blockStore.Height()))
+					// Per issue 4069 comment: If a block is not found, the consensus reactor should not crash since 
+					// this does not affect the ability to participate in consensus. Instead it should wait for the 
+					// gossipSleepDuration and continue.
+					logger.Debug("Block not found in store", "height", prs.Height)
+					time.Sleep(conR.conS.config.PeerGossipSleepDuration)
+				} else {
```
marbar3778: I believe you can drop the else statement, otherwise LGTM

consensus/reactor.go
```diff
@@ -501,10 +501,14 @@ OUTER_LOOP:
 			if prs.ProposalBlockParts == nil {
 				blockMeta := conR.conS.blockStore.LoadBlockMeta(prs.Height)
 				if blockMeta == nil {
-					panic(fmt.Sprintf("Failed to load block %d when blockStore is at %d",
-						prs.Height, conR.conS.blockStore.Height()))
+					// Per issue 4069 comment: If a block is not found, the consensus reactor should not crash since
```
marbar3778: This comment won't really help in the future, can it be reworded or removed

consensus/reactor.go
melekes: no, we don't want to call InitProposalBlockParts with empty blockMeta

melekes:
Could you please add an entry to `CHANGELOG_PENDING.md`?

```
BUGFIXES:
- [cs] \#4069 Don't panic when block meta is not found in store
```

consensus/reactor.go
```diff
@@ -501,10 +501,12 @@ OUTER_LOOP:
 			if prs.ProposalBlockParts == nil {
 				blockMeta := conR.conS.blockStore.LoadBlockMeta(prs.Height)
 				if blockMeta == nil {
-					panic(fmt.Sprintf("Failed to load block %d when blockStore is at %d",
-						prs.Height, conR.conS.blockStore.Height()))
+					logger.Error("Failed to load block meta",
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
					logger.Error("Failed to load block meta",
```

hendrikhofstadt:
Awesome!

Any chance this could be backported to the tendermint version for the hub-3 (potential future patch release) so everyone can immediately profit from this?

Issue #4166, opened 2019-11-19
Follow up to #3783


ebuchman:
See https://github.com/tendermint/tendermint/pull/3783#pullrequestreview-319319081

The new `number_txs` fields should be `num_txs` for consistency with the rest of the codebase.

Also, the breaking changes should be better documented since they're multifold:
- Apps: breaks the ABCI header field numbering
- State: removed TotalTxs field breaks the format of State on disk (I think this normally goes under the CLI/RPC/Config section)
- RPC: since the header broke, all rpc requests which expose the header broke
- Go API: the header broke
- P2P: since blocks go over the wire, technically the P2P protocol broke, but I'm not sure if that's implied, or if there are ways for the block protocol to break without the p2p protocol to break (ie changing how a merkle root is computed is block breaking but not p2p breaking, but changing fields in a data structure breaks both!)

Issue #4136, opened 2019-11-14
Tendermint install script not working


Tendermeat:
**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source):
Last version (i think)

**ABCI app** (name for built-in, URL for self-written if it's publicly available):

**Environment**: Ubuntu 18.04
- **OS** (e.g. from /etc/os-release):
- **Install tools**:
- **Others**:

**What happened**:
run tendermint script https://github.com/tendermint/tendermint/blob/master/scripts/install/install_tendermint_ubuntu.sh

and this happened

go/test/zerodivide.go
package github.com/tendermint/tendermint: no Go files in /home/erg/go/src/github.com/tendermint/tendermint
Already on 'master'
Your branch is up to date with 'origin/master'.
Get Certstrap
go get: warning: modules disabled by GO111MODULE=auto in GOPATH/src;
	ignoring go.mod;
	see 'go help modules'
go: cannot use path@version syntax in GOPATH mode
tools.mk:61: recipe for target '/home/erg/go/bin/certstrap' failed
make: *** [/home/erg/go/bin/certstrap] Error 1
CGO_ENABLED=0 go install -mod=readonly -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short=8 HEAD` -s -w" -tags 'tendermint' ./cmd/tendermint
build flag -mod=readonly only valid when using modules
Makefile:34: recipe for target 'install' failed
make: *** [install] Error 1

**What you expected to happen**:
Installation succeed

**Have you tried the latest version**: yes/no
i think yes

**How to reproduce it** (as minimally and precisely as possible): 

**Logs (paste a small part showing an error (< 10 lines) or link a pastebin, gist, etc. containing more of the log file)**:

**Config (you can paste only the changes you've made)**:

**node command runtime flags**:

**`/dump_consensus_state` output for consensus bugs**

**Anything else we need to know**:

Tendermeat:
up

marbar3778:
can you try `go get ./...`. It is working, so not sure where the go.mod is wrong that is causing you to do the `go get` command

Tendermeat:
hey thx for the response, the go get command is working

melekes:
If you have `GO111MODULE=auto`, you need Go >= `1.13`. What Go version do you have?

Tendermeat:
I deleted everything and rerun the script with and modified go version to GO_VERSION=1.13
[...]

PR #4170, opened 2019-11-20
Various fixes

melekes:
Fixes https://github.com/tendermint/tendermint/issues/4136
Fixes https://github.com/tendermint/tendermint/issues/4166

* [ ] Wrote tests

### DIFF
--- a/docs/spec/blockchain/blockchain.md
+++ b/docs/spec/blockchain/blockchain.md
@@ -187,6 +187,8 @@ type DuplicateVoteEvidence struct {
 }
 ```
 
+Votes are lexicographically sorted on `BlockID`.
+
 See the [pubkey spec](./encoding.md#key-types) for more.
 
 ## Validation
--- a/tools.mk
+++ b/tools.mk
@@ -59,20 +59,18 @@ check_tools:
 certstrap: $(CERTSTRAP)
 $(CERTSTRAP):
 	@echo "Get Certstrap"
-	@go get github.com/square/certstrap@338204a88c4349b1c135eac1e8c14c693ad007da
+	@go get github.com/square/certstrap@v1.2.0
 
 protobuf: $(PROTOBUF)
 $(PROTOBUF):
 	@echo "Get Protobuf"
-	## protobuf v1.3.0
-	@go get github.com/gogo/protobuf/protoc-gen-gogo@0ca988a254f991240804bf9821f3450d87ccbb1b
+	@go get github.com/gogo/protobuf/protoc-gen-gogo@v1.3.1
 
+# used to build tm-monitor binaries
 gox: $(GOX)
 $(GOX):
 	@echo "Get Gox"
-# used to build tm-monitor binaries
-	## gox v1.0.1
-	@go get github.com/mitchellh/gox@d8caaff5a9dc98f4cfa1fcce6e7265a04689f641
+	@go get github.com/mitchellh/gox@v1.0.1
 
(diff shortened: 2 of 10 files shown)


marbar3778:
do you want to add some extra changelog entries with what https://github.com/tendermint/tendermint/issues/4166 recommends? then we can close that one as well

melekes:
> do you want to add some extra changelog entries with what #4166 recommends? then we can close that one as well

will do

tessr:
This PR looks like it addresses a number of separate and unrelated issues. Can they be split into multiple PRs/commits, please? This will make debugging easier in the future, if we ever have to go git history spelunking 😇

melekes:
What about merge commit?

tools.mk
```diff
@@ -59,28 +59,45 @@ check_tools:
 certstrap: $(CERTSTRAP)
 $(CERTSTRAP):
 	@echo "Get Certstrap"
-	@go get github.com/square/certstrap@338204a88c4349b1c135eac1e8c14c693ad007da
+	@go get github.com/square/certstrap@v1.2.0
+
+# Choose protobuf binary based on OS.
+PROTOC_ZIP=""
+ifneq ($(OS),Windows_NT)
```
melekes: should we try to make this work for Windows || 32 bit systems || ARM?

tessr:
> What about merge commit?

I'm not sure we're talking about the same thing--I mean that each commit should represent an independent change, so that we can easily identify the motivation for each change. And the right way to do that, given our git flow, is probably to open a separate PR for each one.

docs/spec/blockchain/blockchain.md
```diff
@@ -187,6 +187,8 @@ type DuplicateVoteEvidence struct {
 }
 ```
 
+Votes are lexicographically sorted on `BlockID`.
```
marbar3778: this should be added to the spec repo. the specs in this repo are going to be removed so we dont have duplicates

melekes:
> I mean that each commit should represent an independent change, so that we can easily identify the motivation for each change

https://github.com/tendermint/tendermint/pull/4170/commits look pretty good to me

> I'm not sure we're talking about the same thing

Sorry. I was trying to say that we can "Merge" this PR with merge commit (instead of "Squash and merge"). That way we will a) preserve independent commit history b) spare my time. What do you think?

tessr:
Oh yeah, that's fine. But let's rebase onto master so we don't leave a merge commit. (And speaking of which, can you rebase this branch to remove the merge commit that's already on it?)

docs/spec/blockchain/blockchain.md
marbar3778: added to the spec repo

tools.mk
```diff
@@ -59,28 +59,54 @@ check_tools:
...
+# Choose protobuf binary based on OS (only works for 64bit Linux and Mac).
+# NOTE: On Mac, installation via brew (brew install protoc) might be favorable.
+PROTOC_ZIP=""
+ifneq ($(OS),Windows_NT)
+		UNAME_S := $(shell uname -s)
+		ifeq ($(UNAME_S),Linux)
+			PROTOC_ZIP="protoc-3.10.1-linux-x86_64.zip"
+		endif
+		ifeq ($(UNAME_S),Darwin)
+				PROTOC_ZIP="protoc-3.10.1-osx-x86_64.zip"
```
melekes: formatting is off

PR #4175, opened 2019-11-20
docs/introduction: add notes on architecture intro

tessr:
Over the next few weeks, @brapse, @marbar3778 and I are having a series of meetings to go over the architecture of Tendermint Core. These are my notes from these meetings, which will either serve as an artifact for onboarding future engineers; or will provide the basis for such a document.

If these notes should live somewhere else, let me know (another repo? Notion?).


docs/introduction/architecture.md
```diff
@@ -0,0 +1,55 @@
...
+- Intranode communication: Happens within the node itself (i.e., between reactors or other components). These are typically function or method calls, or occasionally happen through an event bus. 
+- Client communiation: Happens between a client (like a wallet or a browser) and a node on the network.
+
+### Internode Communication
+
+Internode communication can happen in two ways:
+1. TCP connections through the p2p package
+    - Most common form of internode communication
+    - Connections between nodes are persisted and shared across reactors, facilitated by the switch. (More on the switch below.)
+2. RPC over HTTP
```
ebuchman: Why isn't this considered Client Communication?

docs/introduction/architecture.md
```diff
@@ -0,0 +1,55 @@
...
+
+TODO: What is the PEX and the Address Book? 
+
+## node.go 
+
+node.go is the entrypoint for running a node. It sets up reactors, sets up the switch, and registers all the RPC endpoints for a node.
+
+TODO: Enumerate reactors and (maybe) endpoints.
+
+## Types of Nodes
```
ebuchman: See node type documentation here: https://github.com/tendermint/tendermint/blob/master/docs/spec/p2p/node.md

Might be good source for this, but also possibly redundant

melekes:
I think this can be turned into a blog post

tessr:
> Nice initiative! Just beware of possible duplication with https://github.com/tendermint/tendermint/tree/master/docs/spec ... would be awesome to have a compressed/ more intro friendly version of that content!

Yes! Thanks for pointing this out. My hope is that this will be much more high-level than a spec and will focus on things like what the reactors are and how they are connected, rather than data structures or types.

> I think this can be turned into a blog post

That would be awesome, eventually! For now, I want to try to get all these notes together mostly as documentation of the "onboarding" sessions with Sean.

docs/introduction/architecture.md
tessr: This link is broken 😬

docs/introduction/architecture.md
tessr: Oh! I guess I assumed that client communication _also_ happens this way, but my understanding is that nodes can also talk to one another over HTTP. Is that true?

tessr:
Added more notes from two more sessions. Any other questions/should I merge this?

tessr:
Yes, that's the plan! I just needed an approval.

Issue #4164, opened 2019-11-19
Temporal Overflows when Updating validator set


gchaincl:
**Tendermint version**: `97222eb90fc4b2882fe1859a665f2c84764c4b58`

**What happened**: When adding a new validator that temporarily exceedes the `TotalMaxVotingPower` limit, Tendermint crashes

**What you expected to happen**: The validator should be added to the set

**Have you tried the latest version**: yes

**How to reproduce it** (as minimally and precisely as possible):
Say we have the following:
`validator_1` with power = `TotalMaxVotingPower - 1` and
`validator_2` with power = `1`

If you want to update the powers so that:
`validator_2` with power = `TotalMaxVotingPower - 1` and
`validator_1` with power = `1`

Tendermint might end up panicking due to Power Overflow

ancazamfir:
Could you clarify if you saw a panic or an error message? I believe the solution in https://github.com/tendermint/tendermint/pull/4165 is not correct and in fact we will see crashes here.

ancazamfir:
@gchaincl @tessr @melekes 
Just committed 3 tests at:
https://github.com/tendermint/tendermint/blob/5e4cbd2ca5716bf11bd7777f080b85c128bb3130/types/validator_set_test.go#L1264

- 1 - no false overflow error messages for updates
- 2 - no false overflow error messages for deletes
- 3 - check panic on overflow is prevented: update 8 validators with power int64(math.MaxInt64)/8

tests 1 and 2 fail before fix https://github.com/tendermint/tendermint/pull/4165, they return an error when they shouldn't
tests 2 and 3 fail after fix https://github.com/tendermint/tendermint/pull/4165, with 3 causing a panic.

Will work on a fix next.

melekes:
Should be fixed now with https://github.com/tendermint/tendermint/pull/4183. Thanks all!

PR #4183, opened 2019-11-22
types: prevent spurious validator power overflow warnings when changing the validator set

ancazamfir:
Fix for https://github.com/tendermint/tendermint/issues/4164
The general problem is that in certain conditions an overflow warning is issued when attempting to update a validator set even if the final set's total voting power is not over the maximum allowed.
Root cause is that in `verifyUpdates()`, updates are verified wrt to total voting power  in the order of validator address. It is then possible that a low address validator may increase its power such that the temporary total voting power count goes over `MaxTotalVotingPower`.

Scenarios where removing and adding/ updating validators with high voting power, in the same update operation, cause the same false warning and the updates are not applied.

Main changes to fix this are in `verifyUpdate()` that now does the verification starting with the decreases in power. It also takes into account the removals that are part of the update.

* [x] Wrote tests


types/validator_set_test.go
```diff
@@ -1260,6 +1246,74 @@ func verifyValSetUpdatePriorityOrder(t *testing.T, valSet *ValidatorSet, cfg tes
...
+				{"v1", 1}, {"v2", 1}, {"v3", 1}, {"v4", 1}, {"v5", 1},
+				{"v6", 1}, {"v7", 1}, {"v8", 1}, {"v9", 1}},
+			wantErr: true,
+		},
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
+			// create a new validator set
+			valSet := createNewValidatorSet(tt.startVals)
```
golangcibot: Using the variable on range scope `tt` in function literal (from `scopelint`)

types/validator_set_test.go
```diff
@@ -1260,6 +1246,74 @@ func verifyValSetUpdatePriorityOrder(t *testing.T, valSet *ValidatorSet, cfg tes
...
+			verifyValidatorSet(t, valSet)
+
+			// execute update and verify returned error is as expected
+			applyChangesToValSet(t, tt.wantErr, valSet, tt.addedVals, tt.updatedVals, tt.deletedVals)
```
golangcibot: Using the variable on range scope `tt` in function literal (from `scopelint`)

types/validator_set_test.go
```diff
@@ -1260,6 +1246,74 @@ func verifyValSetUpdatePriorityOrder(t *testing.T, valSet *ValidatorSet, cfg tes
...
+
+			// verify updated validator set is as expected
+			assert.Equal(t, tt.expectedVals, toTestValList(valSet.Validators))
```
golangcibot: Using the variable on range scope `tt` in function literal (from `scopelint`)

types/validator_set.go
```diff
@@ -370,52 +371,87 @@ func processChanges(origChanges []*Validator) (updates, removals []*Validator, e
...
 	updates []*Validator,
 	vals *ValidatorSet,
-) (updatedTotalVotingPower int64, numNewValidators int, err error) {
+	removedPower int64,
+) (tvpAfterUpdatesBeforeRemovals int64, numNewValidators int, err error) {
 
-	updatedTotalVotingPower = vals.TotalVotingPower()
+	// powerChanges is a temporary slice that eventually stores the changes, starting with the
+	// decreases in validator powers.
+	powerChanges := make([]powerChange, 0, len(updates))
```
melekes: can't we just sort `updates` by `VotingPower` and do "compute the total voting power after updates starting with the power decreases" over them instead of introducing two new slices? This would radically simplify the code

types/validator_set.go
```diff
@@ -370,52 +371,87 @@ func processChanges(origChanges []*Validator) (updates, removals []*Validator, e
...
+	for _, powerChange := range powerChanges {
+		tvpAfterRemovals += powerChange.delta
+		if tvpAfterRemovals > MaxTotalVotingPower {
+			err = fmt.Errorf(
+				"failed to add/update validator %v, total voting power would exceed the max allowed %v",
+				powerChange.address, MaxTotalVotingPower)
+			return 0, 0, err
+		}
+	}
+	tvpAfterUpdatesBeforeRemovals = tvpAfterRemovals + removedPower
```
melekes: shouldn't we subtract here? `before = after - removedDelta` makes more sense to me rather than `before = after + removedDelta`

types/validator_set.go
```diff
@@ -481,19 +517,21 @@ func (vals *ValidatorSet) applyUpdates(updates []*Validator) {
 
 // Checks that the validators to be removed are part of the validator set.
 // No changes are made to the validator set 'vals'.
-func verifyRemovals(deletes []*Validator, vals *ValidatorSet) error {
+func verifyRemovals(deletes []*Validator, vals *ValidatorSet) (int64, error) {
```
melekes:
```suggestion
func verifyRemovals(deletes []*Validator, vals *ValidatorSet) (votingPower int64, err error) {
```

types/validator_set.go
```diff
@@ -370,52 +371,87 @@ func processChanges(origChanges []*Validator) (updates, removals []*Validator, e
...
 //
-// 'updates' should be a list of proper validator changes, i.e. they have been verified
-// by processChanges for duplicates and invalid values.
-// No changes are made to the validator set 'vals'.
 func verifyUpdates(
 	updates []*Validator,
 	vals *ValidatorSet,
-) (updatedTotalVotingPower int64, numNewValidators int, err error) {
+	removedPower int64,
+) (tvpAfterUpdatesBeforeRemovals int64, numNewValidators int, err error) {
```
melekes:
```suggestion
) (votingPower int64, numNewValidators int, err error) {
```

Also, I feel like this function does 2 separate things - calculates voting power and num new validators. Would it make sense to split this into 2 funcs - `verifyUpdates(updates, vals) (votingPower int64, err error)` and `numNewValidators(updates, vals) int` ?

types/validator_set.go
```diff
@@ -562,13 +603,13 @@ func (vals *ValidatorSet) updateWithChangeSet(changes []*Validator, allowDeletes
...
 	// Compute the priorities for updates.
-	computeNewPriorities(updates, vals, updatedTotalVotingPower)
+	computeNewPriorities(updates, vals, tvpAfterUpdatesBeforeRemovals)
 
 	// Apply updates and removals.
 	vals.applyUpdates(updates)
 	vals.applyRemovals(deletes)
 
-	vals.updateTotalVotingPower()
+	vals.updateTotalVotingPower() // will panic if total voting power > MaxTotalVotingPower
```
melekes: comment should be in the func godoc, not here

types/validator_set_test.go
```diff
@@ -1260,6 +1246,75 @@ func verifyValSetUpdatePriorityOrder(t *testing.T, valSet *ValidatorSet, cfg tes
 	}
 }
 
+func TestValSetUpdateOverflowRelated(t *testing.T) {
+	// short name for test readability
```
melekes: what are you referring to? testCase.name below?

types/validator_set_test.go
```diff
@@ -1260,6 +1246,75 @@ func verifyValSetUpdatePriorityOrder(t *testing.T, valSet *ValidatorSet, cfg tes
...
+	big := MaxTotalVotingPower
```
melekes: why not use MaxTotalVotingPower directly? I think big obscures the meaning and can potentially bring more confusion rather than clarity

types/validator_set_test.go
```diff
@@ -1260,6 +1246,75 @@ func verifyValSetUpdatePriorityOrder(t *testing.T, valSet *ValidatorSet, cfg tes
...
+				{"v1", 1}, {"v2", 1}, {"v3", 1}, {"v4", 1}, {"v5", 1},
+				{"v6", 1}, {"v7", 1}, {"v8", 1}, {"v9", 1}},
+			wantErr: true,
+		},
+	}
+
+	for _, tt := range testCases {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			// create a new validator set
```
melekes: comment does not bring any value imho

melekes:
Thank you for fixing this ❤️

types/validator_set.go
ancazamfir: We want the power before removals so we add the power that is removed. removed was already substracted from "after".

types/validator_set.go
ancazamfir: Yes we could split but then we would do the loop twice. Alternatively I could change the name.

types/validator_set_test.go
ancazamfir: I found it easier to read/ write the tests.

types/validator_set.go
ancazamfir: I had a version earlier doing that, needs the new sorting function and has higher computational complexity. If you feel strongly about it I can change it back, let me know.

types/validator_set.go
melekes: > the loop twice

nanoseconds

types/validator_set_test.go
melekes: Maybe call it max then? Why big?

types/validator_set.go
melekes: Is there a separate commit with those changes I can see?

ancazamfir: I just did a commit to use sort to see how it looks.

types/validator_set.go
ancazamfir: changed

types/validator_set_test.go
ancazamfir: Because it doesn't have to be the max really, any value close enough to max would cause a problem in the earlier versions. Anyways, I changed it to `MaxTotalVotingPower` as per your earlier comment.

ancazamfir:
> Thank you for fixing this ❤️

Thanks for the review!

types/validator_set.go
melekes: Maybe I wasn't clear enough. I was talking about `updates`, not `powerChanges`. Why not sort `updates` and avoid having `powerChanges` altogether?

types/validator_set.go
```diff
@@ -370,52 +372,88 @@ func processChanges(origChanges []*Validator) (updates, removals []*Validator, e
...
+		tvpAfterRemovals += powerChange.delta
+		if tvpAfterRemovals > MaxTotalVotingPower {
+			err = fmt.Errorf(
+				"failed to add/update validator %v, total voting power would exceed the max allowed %v",
+				powerChange.address, MaxTotalVotingPower)
+			return 0, err
+		}
 	}
+	tvpAfterUpdatesBeforeRemovals = tvpAfterRemovals + removedPower
+	return tvpAfterUpdatesBeforeRemovals, nil
```
melekes:
```suggestion
	return tvpAfterRemovals + removedPower, nil
```

types/validator_set.go
```diff
@@ -370,52 +372,88 @@ func processChanges(origChanges []*Validator) (updates, removals []*Validator, e
...
+}
 
-	return updatedTotalVotingPower, numNewValidators, nil
+// Computes the number of new validators in the update.
+func numNewValidators(updates []*Validator, vals *ValidatorSet) int {
+	numNewValidators := 0
+	for _, valUpdate := range updates {
+		address := valUpdate.Address
+		_, val := vals.GetByAddress(address)
```
melekes:
```suggestion
		if vals.HasAddress(valUpdate.Address) {
```

types/validator_set.go
ancazamfir: Ah, sorry I missed that. Because `updates` need to be sorted by address for `applyUpdates()` that does the merge. So here we can sort updates by change but then sort them again by address.

melekes:
```
func verifyUpdates(
	updates []*Validator,
	vals *ValidatorSet,
	removedPower int64,
) (tvpAfterUpdatesBeforeRemovals int64, err error) {
	delta := func(update *Validator, vals *ValidatorSet) int64 {
		_, val := vals.GetByAddress(update.Address)
		if val != nil {
			return update.VotingPower - val.VotingPower
		}
		return update.VotingPower
	}

	updatesCopy := validatorListCopy(updates)
	sort.Slice(updatesCopy, func(i, j int) bool {
		return delta(updatesCopy[i], vals) < delta(updatesCopy[j], vals)
	})

	tvpAfterRemovals := vals.TotalVotingPower() - removedPower
	for _, upd := range updatesCopy {
		tvpAfterRemovals += delta(upd, vals)
		if tvpAfterRemovals > MaxTotalVotingPower {
			err = fmt.Errorf(
				"failed to add/update validator %v, total voting power would exceed the max allowed %v",
				upd.Address, MaxTotalVotingPower)
			return 0, err
		}
	}
	return tvpAfterRemovals + removedPower, nil
}
```

melekes: The above function will benefit even more if we sort updates by delta from the beginning

types/validator_set.go
```diff
@@ -788,6 +823,26 @@ func (valz ValidatorsByAddress) Swap(i, j int) {
 	valz[j] = it
 }
 
+//-------------------------------------
+// Implements sort for sorting updates by voting power changes.
+
+// Sort validators by voting power.
+type updatesByPowerChange []powerChange
```
golangcibot: type `updatesByPowerChange` is unused (from `unused`)

types/validator_set.go
```diff
@@ -351,71 +354,97 @@ func processChanges(origChanges []*Validator) (updates, removals []*Validator, e
...
 			updates = append(updates, valUpdate)
 		}
+
 		prevAddr = valUpdate.Address
 	}
+
 	return updates, removals, err
 }
 
+type powerChange struct {
```
golangcibot: type `powerChange` is unused (from `unused`)

types/validator_set.go
```diff
@@ -351,71 +354,97 @@ func processChanges(origChanges []*Validator) (updates, removals []*Validator, e
...
+	address Address
+	delta   int64
+}
+
 // Verifies a list of updates against a validator set, making sure the allowed
```
tessr: This is a nitpick but since you updated the format of other comments, do you want to format this one, too? (e.g. `verifyUpdates verifies a list...`)

types/validator_set.go
```diff
@@ -351,71 +354,97 @@ func processChanges(origChanges []*Validator) (updates, removals []*Validator, e
...
-// by processChanges for duplicates and invalid values.
-// No changes are made to the validator set 'vals'.
+// err - non-nil if the maximum allowed total voting power would be exceeded
 func verifyUpdates(
 	updates []*Validator,
 	vals *ValidatorSet,
-) (updatedTotalVotingPower int64, numNewValidators int, err error) {
+	removedPower int64,
+) (tvpAfterUpdatesBeforeRemovals int64, err error) {
+	delta := func(update *Validator, vals *ValidatorSet) int64 {
```
tessr: This function gets the delta between a given validator's current power and updated power, right? The delta will be positive if the validator has increased in power and negative if the validator has decreased in power?

types/validator_set.go
```diff
@@ -351,71 +354,97 @@ func processChanges(origChanges []*Validator) (updates, removals []*Validator, e
...
+}
+
 // Verifies a list of updates against a validator set, making sure the allowed
 // total voting power would not be exceeded if these updates would be applied to the set.
 //
+// Inputs:
+// updates - a list of proper validator changes, i.e. they have been verified by processChanges for duplicates
+//   and invalid values.
+// vals - the original validator set. Note that vals is NOT modified by this function.
+// removedPower - the total voting power that will be removed after the updates are verified and applied.
```
tessr: Checking my understanding here--

We get this value as the return value of `verifyRemovals`, which checks that the newly-removed validators were all part of the previous validator set, and _also_ calculates the total power that they had. In other words, this value is the total voting power held by the validators that are getting removed outright, correct? And it's possible that other validators will also have their power reduced, even if they remain in the validator set, right?

types/validator_set.go
```diff
@@ -545,30 +576,33 @@ func (vals *ValidatorSet) updateWithChangeSet(changes []*Validator, allowDeletes
...
 	}
 
-	// Check that the resulting set will not be empty.
-	if numNewValidators == 0 && len(vals.Validators) == len(deletes) {
-		return errors.New("applying the validator changes would result in empty set")
-	}
-
 	// Compute the priorities for updates.
-	computeNewPriorities(updates, vals, updatedTotalVotingPower)
+	computeNewPriorities(updates, vals, tvpAfterUpdatesBeforeRemovals)
```
tessr: I'm confused about why we're passing in the tvp before the removals now, especially since it looks like the implementation of computeNewPriorities hasn't changed?

types/validator_set.go
```diff
@@ -351,71 +354,97 @@ func processChanges(origChanges []*Validator) (updates, removals []*Validator, e
...
+		if !vals.HasAddress(valUpdate.Address) {
+			numNewValidators++
+		}
 	}
-
-	return updatedTotalVotingPower, numNewValidators, nil
+	return numNewValidators
 }
 
 // Computes the proposer priority for the validators not present in the set based on 'updatedTotalVotingPower'.
```
tessr: nitpick: comment format here too!

types/validator_set.go
melekes: I think the code there will benefit from refactoring (example: https://github.com/tendermint/tendermint/pull/4183/commits/02a89fca9457738550173efe0517c9a139a41f69)

tessr: Can you elaborate more on that example? Are you saying we should use a switch here too? Or just giving that as an example of some effective refactoring in the same part of the codebase?

melekes: The latter 😄

types/validator_set.go
ancazamfir: I just updated the public ones to make golint happy :)

types/validator_set.go
ancazamfir: Right.

types/validator_set.go
ancazamfir: correct.

types/validator_set.go
ancazamfir: > I'm confused about why we're passing in the tvp before the removals now, especially since it looks like the implementation of computeNewPriorities hasn't changed?

It was the same before, just changed the variable name.

types/validator_set_test.go
```diff
@@ -1260,6 +1246,77 @@ func verifyValSetUpdatePriorityOrder(t *testing.T, valSet *ValidatorSet, cfg tes
...
 }
 
+func TestValSetUpdateOverflowRelated(t *testing.T) {
+	testCases := []testVSetCfg{
+		{
+			name:         "1 no false overflow error messages for updates",
+			startVals:    []testVal{{"v1", 1}, {"v2", MaxTotalVotingPower - 1}},
+			updatedVals:  []testVal{{"v1", MaxTotalVotingPower - 1}, {"v2", 1}},
+			expectedVals: []testVal{{"v1", MaxTotalVotingPower - 1}, {"v2", 1}},
+			wantErr:      false,
```
melekes: would be nice to assert for specific error, not just some error!

types/validator_set.go
tessr: Maybe a comment to that effect would be helpful here!

types/validator_set.go
tessr: Ahh, I see. I couldn't tell at first from reading the guts of `verifyUpdates`, but this makes sense. Thanks!

PR #4185, opened 2019-11-24
p2p/conn: simplify secret connection handshake malleability fix with merlin

### LINKED ISSUE #4140: p2p/secret_connection: use the transcript solely for authentication (i.e. MAC)
https://github.com/tendermint/tendermint/pull/3668#issuecomment-553860663

Refs #3668

zmanian:
This simplifies transcript hashing using the merlin protocol rather than HKDF.
It's a cleaner API
This also only uses the transcript hash as a MAC. Fixes #4140 as well.

* [ ] Wrote tests


CHANGELOG_PENDING.md
```diff
@@ -69,7 +69,7 @@ program](https://hackerone.com/tendermint).
 
 - P2P Protocol
 
-  - [p2p][\3668](https://github.com/tendermint/tendermint/pull/3668) Make `SecretConnection` non-malleable
+  - [p2p][\4185](https://github.com/tendermint/tendermint/pull/4185) Make `SecretConnection` non-malleable
```
melekes: please create a separate entry and not modify existing one

p2p/conn/secret_connection.go
```diff
@@ -20,6 +20,7 @@ import (
 	"golang.org/x/crypto/hkdf"
 	"golang.org/x/crypto/nacl/box"
 
+	"github.com/gtank/merlin"
```
```suggestion
	"github.com/gtank/merlin"

```

p2p/conn/secret_connection.go
```diff
@@ -20,6 +20,7 @@ import (
 	"golang.org/x/crypto/hkdf"
 	"golang.org/x/crypto/nacl/box"
```
```suggestion
```

p2p/conn/secret_connection.go
```diff
@@ -97,17 +96,11 @@ func MakeSecretConnection(conn io.ReadWriteCloser, locPrivKey crypto.PrivKey) (*
...
+	transcript := merlin.NewTranscript("TENDERMINT_SECRET_CONNECTION_TRANSCRIPT_HASH")
 
-	hkdfLoEphPub := hkdf.New(hash, loEphPub[:], hdkfState[:], []byte("TENDERMINT_SECRET_CONNECTION_TRANSCRIPT_HASH"))
-	if _, err := io.ReadFull(hkdfLoEphPub, hdkfState[:]); err != nil {
-		return nil, err
-	}
+	transcript.AppendMessage([]byte("EPHEMERAL_LOWER_PUBLIC_KEYS"), loEphPub[:])
 
-	hkdfHiEphPub := hkdf.New(hash, hiEphPub[:], hdkfState[:], []byte("TENDERMINT_SECRET_CONNECTION_TRANSCRIPT_HASH"))
+	transcript.AppendMessage([]byte("EPHEMERAL_UPPER_PUBLIC_KEYS"), hiEphPub[:])
```
```suggestion
	transcript.AppendMessage([]byte("EPHEMERAL_UPPER_PUBLIC_KEY"), hiEphPub[:])
```

?

p2p/conn/secret_connection.go
```diff
@@ -118,19 +111,18 @@ func MakeSecretConnection(conn io.ReadWriteCloser, locPrivKey crypto.PrivKey) (*
...
-	}
+	transcript.AppendMessage([]byte("DH_SECRET"), dhSecret[:])
 
 	// Generate the secret used for receiving, sending, challenge via HKDF-SHA2
 	// on the transcript state (which itself also uses HKDF-SHA2 to derive a key
 	// from the dhSecret).
-	recvSecret, sendSecret, challenge := deriveSecretAndChallenge(hdkfState, locIsLeast)
+	recvSecret, sendSecret := deriveSecrets(dhSecret, locIsLeast)
+
+	var challenge [32]byte
```
melekes: please avoid using magic numbers.

Is there a constant for this? `ChallengeSize` or something

if not, could you add one

```suggestion
	const challengeSize = 32
	var challenge [challengeSize]byte
```

p2p/conn/secret_connection.go
```diff
@@ -97,17 +96,11 @@ func MakeSecretConnection(conn io.ReadWriteCloser, locPrivKey crypto.PrivKey) (*
...
-	if _, err := io.ReadFull(hkdfInit, hdkfState[:]); err != nil {
-		return nil, err
-	}
+	transcript := merlin.NewTranscript("TENDERMINT_SECRET_CONNECTION_TRANSCRIPT_HASH")
 
-	hkdfLoEphPub := hkdf.New(hash, loEphPub[:], hdkfState[:], []byte("TENDERMINT_SECRET_CONNECTION_TRANSCRIPT_HASH"))
-	if _, err := io.ReadFull(hkdfLoEphPub, hdkfState[:]); err != nil {
-		return nil, err
-	}
+	transcript.AppendMessage([]byte("EPHEMERAL_LOWER_PUBLIC_KEYS"), loEphPub[:])
```
```suggestion
	transcript.AppendMessage([]byte("EPHEMERAL_LOWER_PUBLIC_KEY"), loEphPub[:])
```

?

tessr: Should we make this byte slice a variable as well?

melekes: can be a package level const

melekes: probably okay to leave them inside []byte until they are used somewhere else

marbar3778: or are they: https://github.com/uber-go/guide/blob/master/style.md#avoid-string-to-byte-conversion.

melekes: good point @marbar3778 !

tessr:
Nice. I'm going to merge this, but I'd also love to read more about Merlin vs. HKDF, and it could be a useful thing to include in the comments of this PR. @zmanian do you have any resources handy? 

(I personally have some understanding of Merlin--thanks Henry!--but don't know anything about HKDF, and I imagine others would find it useful too.)

marbar3778:
dev session?

tessr:
That would be fun. Maybe we could get Henry or George to come, too.

zmanian:
Transcript hashing and exchanging a transcript derived authenticator 

Merlin is kind of the ideal transcript hashing API. It's safe accumulate and extract state in the function.

TLS 1.3 uses a strategy of concatenating all messages together.
https://tools.ietf.org/id/draft-ietf-tls-tls13-21.html#rfc.section.4.4.1

Noise uses HKDF to derive both encryption and MAC keys from the transcript. 
https://noiseprotocol.org/noise.html#hash-functions

IMO Merlin is the nicest and easiest to understand API for doing thing.  Our needs are basically like TLS 1.3. We don't really need to ability to accumulate and extract.

One downside of this is that it would make alternative language implementations harder. Merlin is available in go and rust. And likely c soonish.

ebuchman:
@zmanian can we get this set of changes written up in an RFC in https://github.com/tendermint/spec so we have a permanent place for the rationale to live. A bit backwards this time since the changes have already been merged, but we should follow the process. Also need to update the specs for this in that repo. Thanks!

zmanian:
Yes will do.

PR #4203, opened 2019-11-28
networks/remote: turn on GO111MODULE and use git clone instead of go get

w1t:
To execute the integration.sh, you need to run  "source ./integration.sh".  just execute command "./integration.sh" will not work. 
It is because "source ~/.profile" executes in a separate shell than the integration.sh script 
# **turn on the go module, default is auto. The value is off, if tendermint source code
#is downloaded under $GOPATH/src directory

echo "export GO111MODULE=on" >> ~/.profile
# ** use git clone instead of go get.
# once go module is on, go get will download source code to
# specific version directory under $GOPATH/pkg/mod the make
# script will not work
git clone https://github.com/tendermint/tendermint.git

#** need to install the package, otherwise tendermint testnet will not execute
make install

* [ x] Referenced an issue explaining the need for the change
* [x ] Updated all code comments where relevant
* [ ] Wrote tests


networks/remote/integration.sh
```diff
@@ -21,17 +21,31 @@ echo "export PATH=\$PATH:/usr/local/go/bin" >> ~/.profile
...
+echo "export GO111MODULE=on" >> ~/.profile
 
 source ~/.profile
 
 ## get the code and move into repo
-REPO=github.com/tendermint/tendermint
-go get $REPO
-cd $GOPATH/src/$REPO
-
+#REPO=github.com/tendermint/tendermint
```
melekes: could you remove commented lines?

networks/remote/integration.sh
```diff
@@ -21,17 +21,31 @@ echo "export PATH=\$PATH:/usr/local/go/bin" >> ~/.profile
 mkdir goApps
 echo "export GOPATH=/root/goApps" >> ~/.profile
 echo "export PATH=\$PATH:\$GOPATH/bin" >> ~/.profile
+# **turn on the go module, default is auto. The value is off, if tendermint source code
+#is downloaded under $GOPATH/src directory
+
```
melekes:
```suggestion
```

melekes:
Thanks for contributing to Tendermint Core 👍

PR #4209, opened 2019-12-02
lite: Follow up from #3989

melekes:
Addresses comments in #3989 

- [x] https://github.com/tendermint/tendermint/pull/3989#discussion_r332751051
- [x] https://github.com/tendermint/tendermint/pull/3989#discussion_r352125455
- [x] https://github.com/tendermint/tendermint/pull/3989#pullrequestreview-315454506
- [x] cross-checking primary using alternative providers
- [x] https://github.com/tendermint/tendermint/pull/3989#discussion_r353117610 store only headers within the trusting period
- [x] more docs & tests

---

* [x] Wrote tests


melekes:
> https://github.com/tendermint/tendermint/pull/3989#discussion_r332751051

if options.Height:

1) ahead of trustedHeader.Height => perform Backward verification
2) equals trustedHeader.Height => check Hash is the same (if not, remove trustedHeader and all the headers before)
3) behind trustedHeader.Height => remove all the headers between options.Height and trustedHeader.Height, then check the Hash is the same (if not, remove trustedHeader and all the headers before)

The intuition here is the user always right. I.e. if she decides to reset the light client with an older header, there must be a reason for it.

lite2/verifier_test.go
```diff
@@ -247,6 +268,25 @@ func TestVerifyNonAdjustedHeaders(t *testing.T) {
...
+	var (
+		keys = genPrivKeys(4)
+		// 20, 30, 40, 50 - the first 3 don't have 2/3, the last 3 do!
+		vals     = keys.ToValidators(20, 10)
+		bTime, _ = time.Parse(time.RFC3339, "2006-01-02T15:04:05Z")
+		header   = keys.GenSignedHeader(chainID, lastHeight, bTime, nil, vals, vals,
+			[]byte("app_hash"), []byte("cons_hash"), []byte("results_hash"), 0, len(keys))
+	)
+
+	err := Verify(chainID, header, vals, header, vals, 2*time.Hour, time.Now(), tmmath.Fraction{Numerator: 2, Denominator: 1})
```
golangcibot: line is 123 characters (from `lll`)

lite2/verifier.go
```diff
@@ -16,6 +16,15 @@ var (
 	DefaultTrustLevel = tmmath.Fraction{Numerator: 1, Denominator: 3}
 )
 
+// Verify verifies the new header (h2) against the old header (h1). It ensures that:
+//
+//	a) h1 can still be trusted (if not, ErrOldHeaderExpired is returned);
+//	b) h2 is valid;
+//	c) either h2.ValidatorsHash equals h1NextVals.Hash()
+//		 OR trustLevel ([1/3, 1]) of last trusted validators (h1NextVals) signed
+//		 correctly  (if not, ErrNotEnoughVotingPowerSigned is returned);
```
ancazamfir: IMO there should be two different errors for the two cases:
- the validator set of a skipped header cannot be trusted, i.e. <1/3rd of h1 validator set has signed (new error, something like `ErrValidatorSetCannotBeTrusted`)
- the validator set is trusted but < 2/3rds has signed (`ErrNotEnoughVotingPowerSigned`)

lite2/client.go
```diff
@@ -442,3 +613,62 @@ func (c *Client) compareNewHeaderWithRandomAlternative(h *types.SignedHeader) er
...
+	}
+
+	// 2) Get the latest height.
+	latestHeight, err := c.LastTrustedHeight()
+	if err != nil {
+		c.logger.Error("can't get last trusted height", "err", err)
+		return
+	}
+
+	// 3) Remove all headers (except the very last one) that are outside of
```
ancazamfir: So here we remove all expired headers in the store except the last one. This is also the one referred to by the client's `trustedHeader *types.SignedHeader`.
It is possible that there is a single expired header in the store, same as the client's `trustedHeader`. I believe any header verification from  this point on will fail as we do check for this in `Verify()`. 
But why do we need to keep it and not clean this up as well since we know it is not trusted anymore? If the client restarts with this store, it will restore this header only to be overwritten if `c.trustedHeader.Height != trustOptions.Height`. Also, shouldn't a client fail right away if initialized with an already expired header?

lite2/client.go
```diff
@@ -129,26 +144,111 @@ func NewClient(
...
+//     options.Height and trustedHeader.Height, then check the Hash is the same
+//     (if not, remove trustedHeader and all the headers before)
+//
+// The intuition here is the user always right. I.e. if she decides to reset
+// the light client with an older header, there must be a reason for it.
+func (c *Client) checkTrustedHeaderUsingOptions(options TrustOptions) error {
+	var primaryHash []byte
+	switch {
+	case options.Height > c.trustedHeader.Height:
+		h, err := c.primary.SignedHeader(c.trustedHeader.Height)
```
cwgoes: This seems to mismatch the comment - shouldn't we fetch for `options.Height` ?

Also, why are we re-fetching a header we've already decided to trust - what's the point?

lite2/client.go
```diff
@@ -129,26 +144,111 @@ func NewClient(
...
+	case options.Height < c.trustedHeader.Height:
+		// remove all the headers ( options.Height, trustedHeader.Height ]
+		c.cleanup(options.Height + 1)
+		// set c.trustedHeader to one at options.Height
+		c.restoreTrustedHeaderAndNextVals()
+		primaryHash = options.Hash
+	}
+
+	if !bytes.Equal(primaryHash, c.trustedHeader.Hash()) {
+		c.logger.Info("Prev. trusted header's hash %X doesn't match hash %X from primary provider",
```
cwgoes: Here is one place I think we might want the user to explicitly confirm that they wanted to do this.

lite2/client.go
```diff
@@ -129,26 +144,111 @@ func NewClient(
...
+	case options.Height > c.trustedHeader.Height:
+		h, err := c.primary.SignedHeader(c.trustedHeader.Height)
+		if err != nil {
+			return err
+		}
+		primaryHash = h.Hash()
+	case options.Height == c.trustedHeader.Height:
+		primaryHash = options.Hash
+	case options.Height < c.trustedHeader.Height:
+		// remove all the headers ( options.Height, trustedHeader.Height ]
```
cwgoes: Here is one place I think we might want the user to explicitly confirm that they wanted to do this.

lite2/client.go
```diff
@@ -208,22 +333,25 @@ func (c *Client) TrustedHeader(height int64, now time.Time) (*types.SignedHeader
 	if err != nil {
 		return nil, err
 	}
+	if h == nil {
```
cwgoes: When would this happen - should we return an error?

lite2/client.go
```diff
@@ -442,3 +613,62 @@ func (c *Client) compareNewHeaderWithRandomAlternative(h *types.SignedHeader) er
...
+		return
+	}
+
+	// 3) Remove all headers (except the very last one) that are outside of
+	// the trusting period.
+	for height := oldestHeight; height < latestHeight; height++ {
+		h, err := c.trustedStore.SignedHeader(height)
+		if err != nil {
+			c.logger.Error("can't get a trusted header", "err", err, "height", height)
+			continue
```
cwgoes: Should we panic? When would this happen?

lite2/client.go
```diff
@@ -442,3 +613,62 @@ func (c *Client) compareNewHeaderWithRandomAlternative(h *types.SignedHeader) er
...
+		}
+		if h == nil {
```
cwgoes: Presumably, this will happen if we skipped headers?

Can we iterate through the actual keyspace instead perhaps?

lite2/client.go
```diff
@@ -442,3 +613,62 @@ func (c *Client) compareNewHeaderWithRandomAlternative(h *types.SignedHeader) er
...
+
+		// Stop if the header is within the trusting period.
+		if !HeaderExpired(h, c.trustingPeriod, now) {
+			break
+		}
+
+		err = c.trustedStore.DeleteSignedHeaderAndNextValidatorSet(height)
+		if err != nil {
+			c.logger.Error("can't remove a trusted header & validator set", "err", err, "height", height)
+			continue
```
cwgoes: Seems like we should panic here?

lite2/verifier.go
cwgoes: I agree; I think errors should be as specific as possible, if we are checking a different conditional it should be a different error.

ancazamfir:
I believe we should change a bit the code such that:
- bisection is called with a `newHeader` that has been verified with `verifyNewHeaderAndVals()` and  `newVals.VerifyCommit()`. These two can be combined in some new function `verifyUntrustedSignedHeader()`
- bisection calls `lastVals.VerifyCommitTrusting()` to check that `newVals` can be trusted
- If not, it retrieves the pivot and verifies it with `verifyUntrustedSignedHeader()`. If all good it goes into recursion.

The reasons are:
- `Verify()` can return early with `ErrTooMuchChange` (before `VerifyCommit()` is called), will trigger bisection only to discover when returning from the (potentially deep) recursive call that `Verify()` fails.
- we call `Verify()` multiple times with same `newHeader` and `newVals`. Even though this happens with different `lastHeader` and `lastVals`, there are checks in `Verify()` that are independent of the `last..` values.

Note that the above proposed changes are as described in the english spec.

ancazamfir:
We just had a discussion along the same lines for the rust version of lite client and @ebuchman  explained that it is in fact better to do the `newVals.VerifyCommit()` after bisection (the way it is done now in Go code) in order to prevent a lite client to perform a CPU-intensive full commit verification for a commit with thousands signatures it got from a bad node only to fail on `VerifyCommitTrusting()` later. 

As a follow up to this, we will update the spec.

lite2/client.go
melekes: > Also, shouldn't a client fail right away if initialized with an already expired header?

Yes.

> But why do we need to keep it and not clean this up as well since we know it is not trusted anymore?

because c.trustedHeader is used in VerifyHeader (both in bisection and linear verification). Removing it (setting c.trustedHeader to `nil`) can crash the light client process. We can remove the header from the store however. It should not affect the light client work.

lite2/client.go
```diff
@@ -88,6 +90,15 @@ func AlternativeSources(providers []provider.Provider) Option {
 	}
 }
 
+// RemoveNoLongerTrustedHeadersPeriod option can be used to define how often
```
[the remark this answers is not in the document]
melekes: > I'm a little confused - do we expect the light client to be running as a daemon?

Yeah. Constantly running on people's servers or mobile phones or smth else

> Can this just be a flag or command to gaiacli [xyz]?

You mean making RemoveNoLongerTrustedHeadersPeriod optional. Totally possible, but most would probably enroll since it will be maintaining constant disk usage (opposite to storing all headers).

lite2/client.go
melekes: > This seems to mismatch the comment - shouldn't we fetch for options.Height ?

We're fetching the header, which is trusted by the light client to ensure it's hash matches one from full node. If not, it (along with previous headers) can't be trusted, hence removed. 

That said, maybe we don't need them when the light client is reset with a newer header. Note I expose `Cleanup()` function to remove all the existing headers.

lite2/client.go
```diff
@@ -129,26 +144,111 @@ func NewClient(
...
+//     1) ahead of trustedHeader.Height => fetch header from primary provider
+//     and check it matches trustedHeader (if not, remove trustedHeader and all
+//     the headers before)
+//     2) equals trustedHeader.Height => check Hash is the same (if not, remove
+//     trustedHeader and all the headers before)
+//     3) behind trustedHeader.Height => remove all the headers between
+//     options.Height and trustedHeader.Height, then check the Hash is the same
+//     (if not, remove trustedHeader and all the headers before)
+//
+// The intuition here is the user always right. I.e. if she decides to reset
```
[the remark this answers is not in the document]
melekes: I'd have to think about the right amount of callbacks. Don't want to add too many to avoid callback hell.

lite2/client.go
melekes: Someone requests outdated header (outside of trusting period - removed) OR header which is not yet verified (saved to `trustedStore`)? There's no error per se - light client just does not have the header.

lite2/client.go
melekes: > When would this happen?

problem with underlying database. no panics please. header will be removed during next iteration.

lite2/client.go
melekes: > Can we iterate through the actual keyspace instead perhaps?

Sure. But we'd have to add `Iterator` like function to `Store` interface.

lite2/client.go
melekes: Panics are reserved for exceptional cases. This is not one of them. We'll retry delete once again after `c.removeNoLongerTrustedHeadersPeriod`.

melekes:
@ancazamfir thanks for the update

melekes:
@cwgoes @ancazamfir I believe I've addressed most of your comments. Please review again. Note I'm planning to do https://github.com/tendermint/tendermint/pull/4209#issuecomment-570583895 in a follow up PR.

lite2/verifier.go
```diff
@@ -31,34 +40,34 @@ func Verify(
...
 	}
 
 	if h2.Height == h1.Height+1 {
 		if !bytes.Equal(h2.ValidatorsHash, h1NextVals.Hash()) {
-			return errors.Errorf("expected old header validators (%X) to match those from new header (%X)",
+			err := errors.Errorf("expected old header next validators (%X) to match those from new header (%X)",
 				h1NextVals.Hash(),
 				h2.ValidatorsHash,
 			)
+			return ErrNewValSetCantBeTrusted{err}
```
ancazamfir: Shouldn't we return a different error here?
```suggestion
			return ErrInvalidNewHeader{err}
```
I think the caller deals with this by checking if these are sequential headers and returning an error (vs for example continuing bisection in the skipping header case). But I don't see an advantage of returning the `ErrNewValSetCantBeTrusted` error type. Or maybe just add a comment.

lite2/verifier.go
```diff
@@ -31,34 +40,34 @@ func Verify(
...
 		}
 	} else {
 		// Ensure that +`trustLevel` (default 1/3) or more of last trusted validators signed correctly.
 		err := h1NextVals.VerifyCommitTrusting(chainID, h2.Commit.BlockID, h2.Height, h2.Commit, trustLevel)
 		if err != nil {
-			return err
+			return ErrNewValSetCantBeTrusted{err}
```
ancazamfir: Looks like `VerifyCommitTrusting()` can return some errors that are fatal (the ones different than `ErrNotEnoughVotingPowerSigned`) and should not result in returning `ErrNewValSetCantBeTrusted`.

lite2/verifier.go
```diff
@@ -31,34 +40,34 @@ func Verify(
...
 		}
 	}
 
 	// Ensure that +2/3 of new validators signed correctly.
 	err := h2Vals.VerifyCommit(chainID, h2.Commit.BlockID, h2.Height, h2.Commit)
 	if err != nil {
-		return err
+		return ErrNewHeaderCantBeTrusted{err}
```
ancazamfir: Should we just return err here? Again assimilating with `ErrNewValSetCantBeTrusted` doesn't seem to be the right thing to do. IMO `ErrNewValSetCantBeTrusted` should only show on the skipping header case where we verify that > 1/3rd power of old validators have signed the new header, and should be the only case where bisection continues.

lite2/client.go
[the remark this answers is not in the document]
cwgoes: Ah, more I mean "running every time gaiacli [xyz] is called" (if there are headers which are old).

lite2/client.go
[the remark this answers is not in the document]
cwgoes: Yeah, that's fair. This could be split up into multiple functions perhaps.

lite2/client.go
cwgoes: Hmm, is that not available? It's not a big deal I suppose.

lite2/client.go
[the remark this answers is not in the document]
melekes: Got it. Will make this optional

lite2/client.go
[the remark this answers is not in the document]
melekes: Added `ConfirmationFunction` option.

lite2/verifier.go
melekes: Good catch! Thank you

lite2/verifier.go
melekes: > Again assimilating with ErrNewValSetCantBeTrusted doesn't seem to be the right thing to do.

NOTE these are 2 different errors - `ErrNewValSetCantBeTrusted` VS `ErrNewHeaderCantBeTrusted`

melekes: Removed all new errors except ErrNewValSetCantBeTrusted

lite2/verifier.go
melekes: Removed all the new errors except ErrNewValSetCantBeTrusted

lite2/client.go
```diff
@@ -547,7 +547,7 @@ func (c *Client) bisection(
 	case ErrNewValSetCantBeTrusted:
 		// continue bisection
 	default:
-		return errors.Wrapf(err, "failed to verify the header #%d ", newHeader.Height)
+		return errors.Wrapf(err, "failed to verify the header #%d", newHeader.Height)
 	}
 
 	if newHeader.Height == lastHeader.Height+1 {
```
ancazamfir: Is this check still needed? Since it cannot happen that we get `ErrNewValSetCantBeTrusted` with adjacent headers this could be handled by the `default:` above.

lite2/verifier.go
```diff
@@ -54,20 +54,25 @@ func Verify(
...
+			return err
 		}
 	} else {
 		// Ensure that +`trustLevel` (default 1/3) or more of last trusted validators signed correctly.
 		err := h1NextVals.VerifyCommitTrusting(chainID, h2.Commit.BlockID, h2.Height, h2.Commit, trustLevel)
 		if err != nil {
-			return ErrNewValSetCantBeTrusted{err}
+			switch e := err.(type) {
+			case types.ErrNotEnoughVotingPowerSigned:
+				return ErrNewValSetCantBeTrusted{e}
```
ancazamfir: I might be missing something but I don't understand why not just return err, why do we need translation for this case? Couldn't `VerifyCommitTrusting()` return `ErrNewValSetCantBeTrusted` if `trustLevel` was not achieved?

lite2/verifier.go
ancazamfir: > NOTE these are 2 different errors - ErrNewValSetCantBeTrusted VS ErrNewHeaderCantBeTrusted

Ah! Sorry, my bad.

lite2/verifier.go
melekes: > Couldn't VerifyCommitTrusting() return ErrNewValSetCantBeTrusted if trustLevel was not achieved?

Nope because it does not know that we're checking new validator set. It only knows that vals presence in commit is some number.

lite2/client.go
melekes: Nope. Removed.

ancazamfir:
Unrelated to your changes, why does `validator_set.go:verifyCommitBasic()` have `ValidatorSet` as target? It never uses 'vals'.

melekes:
> Unrelated to your changes, why does validator_set.go:verifyCommitBasic() have ValidatorSet as target? It never uses 'vals'.

another good catch. will make it pure function

lite2/client.go
cwgoes: Is there some way for the user to provide options such that the client will take the provided height and provided hash as *the* trusted header, regardless of whether it's in the future or not?

lite2/client.go
cwgoes: Hmm OK, as long as the caller handles `nil` correctly in all cases.

lite2/client.go
melekes: > Is there some way for the user to provide options such that the client will take the provided height and provided hash as the trusted header, regardless of whether it's in the future or not?

this is the default behavior.

Issue #3007, opened 2018-12-12
Make dynamic changing appVersion possible


james-ray:
Currently , we set state.Version when we get ResponseInfo in handshake.
This upgrade strategy has an obvious flaw that it requires all the validator nodes stopped at the same height, and they upgrade, deploy new version tm and app, then restart.
I suggest we can add an AppVersion in the ResponseEndBlock Message.
```
message ResponseEndBlock {
  repeated ValidatorUpdate validator_updates = 1 [(gogoproto.nullable)=false];
  ConsensusParams consensus_param_updates = 2;
  repeated common.KVPair tags = 3 [(gogoproto.nullable)=false, (gogoproto.jsontag)="tags,omitempty"];
  repeated string blsKeyString = 4;
  uint64 app_version = 5;    //new added field
}
```
When we decide to upgrade to a new version, we set an upgrade height in the future. In a long time period before the upgrade height, the validator nodes can deploy the new version whenever he likes.

In the upgrade height, the app responds this field AppVersion to tell tm whether the upgrade is successful, then tm can update the version in state.

Whether the new code logic in the new version be executed or not can be based on the version, if the upgrade fails, then new logic won't be executed, the program still run the old version logic.

ebuchman:
What do you think about https://github.com/tendermint/tendermint/issues/2671 ? Should we just add this to ConsensusParams?

PR #4225, opened 2019-12-09
abci: update app version on EndBlock

james-ray:
I remember there was a time the issue of updating the app version of State was discussed, but I cannot find the exact issue that it was discussed.
I just find the code that updates the app version of State on handshake is not so reasonable.
We should have a deterministic logic about updating the app version in the State.

Refs #3007

* [ ] Wrote tests


consensus/replay.go
```diff
@@ -259,12 +259,9 @@ func (h *Handshaker) Handshake(proxyApp proxy.AppConns) error {
 		"protocol-version", res.AppVersion,
 	)
 
-	// Set AppVersion on the state.
-	if h.initialState.Version.Consensus.App != version.Protocol(res.AppVersion) {
+	if h.initialState.LastBlockHeight == 0 { //we only update state when we are in initial state
 		h.initialState.Version.Consensus.App = version.Protocol(res.AppVersion)
-		sm.SaveState(h.stateDB, h.initialState)
```
melekes: why we no longer need to save the state?

james-ray:
I find the relavant issue. It is about #3007  
@ebuchman  suggests that we add it in Consensus Param, while I add a new field "AppVersion" in ResponseEndBlock
@melekes  I've updated the code, I think it is logical as the comment in the code:
```
if h.initialState.LastBlockHeight == 0 { //we only update state when we are in initial state
```
because the state is the last state when Tendermint node shut down, the app should not edit it arbitrarily because it will affect the last block on replay, in another word, if the app modifies the app version of state to another version, the check of the last block on replay will fail.

melekes:
> @ebuchman suggests that we add it in Consensus Param, while I add a new field "AppVersion" in ResponseEndBlock

I agree with @ebuchman.

> and inside this if, the state is saved.

what if Tendermint crashes before?

james-ray:
OK, I'll change the code, move appVersion into ConsensusParam.

> what if Tendermint crashes before?

> 
When the node restarts, the appHeight and stateHeight should still be 0,
so they just handshake again.

melekes:
Could you please add an entry to `CHANGELOG_PENDING.md`?

james-ray:
```
#!/bin/bash -eo pipefail
make install install_abci
CGO_ENABLED=0 go install -mod=readonly -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short=8 HEAD` -s -w" -tags 'tendermint' ./cmd/tendermint
build github.com/tendermint/tendermint/cmd/tendermint: cannot load github.com/tendermint/tendermint/libs/common: import lookup disabled by -mod=readonly
make: *** [Makefile:28: install] Error 1
```
why is this?

marbar3778:
libs common was recently removed in favor of having pkgs for the files.

james-ray:
@melekes  I have merged the code, please review it.

james-ray:
Is this possible to merge into 0.33?

CHANGELOG_PENDING.md
```diff
@@ -19,6 +19,8 @@ program](https://hackerone.com/tendermint).
 
 ### IMPROVEMENTS:
 
+- [abci] [\#4225](https://github.com/tendermint/tendermint/issues/3007) Add VersionParams in ResponseEndBlock to allow the app inform Tendermint its updated version
```
```suggestion
- [abci] \#2671 Add Version in ConsensusParams to allow the app inform Tendermint its updated version
```

abci/types/types.proto
```diff
@@ -239,6 +240,11 @@ message ValidatorParams {
   repeated string pub_key_types = 1;
 }
 
+// VersionParams contains the app version of this height
+message VersionParams {
+  uint64 app_version = 1;
```
melekes: uint16 would've been enough imo

consensus/replay.go
```diff
@@ -259,12 +259,9 @@ func (h *Handshaker) Handshake(proxyApp proxy.AppConns) error {
 		"protocol-version", res.AppVersion,
 	)
 
-	// Set AppVersion on the state.
-	if h.initialState.Version.Consensus.App != version.Protocol(res.AppVersion) {
+	if h.initialState.LastBlockHeight == 0 { //we only update state when we are in initial state
```
melekes: why we only update state if LastBlockHeight == 0? say version changes at block 10. Don't we want to update the state in that case?

types/params.go
```diff
@@ -194,5 +207,8 @@ func (params ConsensusParams) Update(params2 *abci.ConsensusParams) ConsensusPar
 		// This avoids having to initialize the slice to 0 values, and then write to it again.
 		res.Validator.PubKeyTypes = append([]string{}, params2.Validator.PubKeyTypes...)
 	}
+	if params2.Version != nil {
+		res.Version.AppVersion = params2.Version.AppVersion
```
melekes: need to write unit tests

consensus/replay.go
```diff
@@ -259,12 +259,9 @@ func (h *Handshaker) Handshake(proxyApp proxy.AppConns) error {
...
 		h.initialState.Version.Consensus.App = version.Protocol(res.AppVersion)
```
melekes: Why do we need to update version here? can't we do it in ReplayBlocks when `if stateBlockHeight == 0`

james-ray: If we don't, the version in the state mismatches with the app version, the app version is supposed to be returned in EndBlock

tessr:
What do we need to do to get this PR across the finish line? Is it just a matter of resolving merge conflicts? Also, can these changes be included in a point release, or should we wait for 0.34?

melekes:
https://github.com/tendermint/tendermint/pull/4225#discussion_r389486841 and double check the logic

james-ray:
> [#4225 (comment)](https://github.com/tendermint/tendermint/pull/4225#discussion_r389486841) and double check the logic

In state/validation.go   line 24,  
```
// Validate basic info.
	if block.Version != state.Version.Consensus {
		return fmt.Errorf("wrong Block.Header.Version. Expected %v, got %v",
			state.Version.Consensus,
			block.Version,
		)
	}
```
If we do not set the initial state.Version.Consensus, this check fails at height 1.

james-ray:
we can use the code below in execution.go
```
	nextParams := state.ConsensusParams
	lastHeightParamsChanged := state.LastHeightConsensusParamsChanged
	if abciResponses.EndBlock.ConsensusParamUpdates != nil {
		// NOTE: must not mutate s.ConsensusParams
		nextParams = state.ConsensusParams.Update(abciResponses.EndBlock.ConsensusParamUpdates)
		err := nextParams.Validate()
		if err != nil {
			return state, fmt.Errorf("Error updating consensus params: %v", err)
		}
		// Change results from this height but only applies to the next height.
		lastHeightParamsChanged = header.Height + 1
	}

	state.Version.Consensus.App = version.Protocol(abciResponses.EndBlock.AppVersion)
	nextVersion := state.Version
	// NOTE: the AppHash has not been populated.
	// It will be filled on state.Save.
	return State{
		Version:                          nextVersion,
```
It is a simple assign value here, we only need to update the AppVersion every height to the state.

melekes:
Closing in favor of https://github.com/tendermint/tendermint/pull/5031

PR #4190, opened 2019-11-25
crypto:  add sr25519 signature scheme

sunnya97:
[sr25519](https://github.com/w3f/schnorrkel) is the schnorrs on ristretto signature scheme that's primarily used by [Parity Substrate](https://substrate.io/).  Adding to Tendermint will be useful for cross-chain compatibility purposes like IBC and hard spoons of Substrate-based chains.  This makes use of the [go-schnorrkel](https://github.com/ChainSafe/go-schnorrkel) library written by Chainsafe.

* [ ] Referenced an issue explaining the need for the change
* [ ] Updated all relevant documentation in docs
* [ ] Updated all code comments where relevant
* [ ] Wrote tests
* [ ] Updated CHANGELOG_PENDING.md

PR #4228, opened 2019-12-10
Update docs and scripts to golang/1.13

dongsam:
`github.com/ChainSafe/go-schnorrkel v0.0.0-20191204112926-74e22bf42a85` is required from https://github.com/tendermint/tendermint/pull/4190, but it dependencies `go 1.13`, so we need to update minimum requirements of go version to 1.13

```
# github.com/ChainSafe/go-schnorrkel
../../go/pkg/mod/github.com/!chain!safe/go-schnorrkel@v0.0.0-20191204112926-74e22bf42a85/helpers.go:14:18: syntax error: unexpected b00000111 at end of statement
note: module requires Go 1.13
make: *** [build] Error 2
```

- update README.md (Minimum requirements of Go version and svg badge)
- update relevant docs and install, make scripts

* [ ] Wrote tests

### DIFF
--- a/tools/build/Makefile
+++ b/tools/build/Makefile
@@ -4,7 +4,7 @@
 
 requirements_check = true
 gpg_check = false
-go_min_version = 1.12.0
+go_min_version = 1.13
 gpg_key = 2122CBE9
 
 ifeq ($(requirements_check),true)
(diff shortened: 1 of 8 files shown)


tools/build/Makefile
```diff
@@ -4,7 +4,7 @@
 
 requirements_check = true
 gpg_check = false
-go_min_version = 1.12.0
+go_min_version = 1.13
```
melekes: shouldn't it be

```suggestion
go_min_version = 1.13.0
```

?

dongsam: there are no 1.13.0 version ends with `.0`
so if ends with `.0` we cannot pass `ifneq ($(go_min_version),$(go_version_check))`
anyway I think `tools/build/Makefile` is out of date and not using now

Issue #4147, opened 2019-11-15
refactor `libs/common` package 


tessr:
Ideally, in idiomatic Go, we wouldn't have a `common` package. Instead, we'd have packages that more closely map to our domain model. [Dave Cheney has written more on this.](https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common)

It's possible that our current package structure can't be immediately reorganized to achieve this. And that's OK--but we should keep this in mind as a goal as we refactor the various reactors and other components of Tendermint. If we can get to a place where our packages are labeled by what they _provide_, that's likely an indicator that we've done a good job with our refactoring.

jackzampolin:
I like adopting Dave Cheney's views on project structure. Incorporating more "idiomatic go" practices :ok_hand: :man_dancing:

marbar3778:
We have quite a few issues related to the common package, maybe we should use this issue as a meta-issue and track where we are with that stuff

melekes:
Refs https://github.com/tendermint/tendermint/issues/2191

marbar3778:
For pkgs that could share a similar name to the std lib, should we name the pkg tm<pkg>, @melekes @tessr thoughts?

melekes:
I'd say yes mainly to avoid the confusion. Like if I say to you imagine apple, and then give you a banana. You will be surprised.

PR #4230, opened 2019-12-10
libs/common: refactor libs/common 01

marbar3778:
- move kv files out of common into its own pkg, updated proto 
- move `IntInSlice()` to where its being used (only in one place)

ref #4147 

* [ ] Wrote tests


state/txindex/kv/kv.go
```diff
@@ -220,7 +220,7 @@ func (txi *TxIndex) Search(q *query.Query) ([]*types.TxResult, error) {
 
 	// for all other conditions
 	for i, c := range conditions {
-		if cmn.IntInSlice(i, skipIndexes) {
+		if IntInSlice(i, skipIndexes) {
```
marbar3778: this is only being used here, so i moved it into the pkg as a utils folder

state/txindex/kv/utils.go
```diff
@@ -1,4 +1,4 @@
-package common
+package kv
 
 // IntInSlice returns true if a is found in the list.
 func IntInSlice(a int, list []int) bool {
```
melekes: Do you think it makes sense to a) make the func private, because it's not core API of this package b) move it closer to where it's used ?

marbar3778: YES!

marbar3778: went with A

PR #4231, opened 2019-12-10
libs/common: refactor libs/common 2

marbar3778:
- move random function to there own pkg

ref #4147 

* [ ] Wrote tests


proxy/app_conn_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/tendermint/tendermint/abci/example/kvstore"
 	"github.com/tendermint/tendermint/abci/server"
 	"github.com/tendermint/tendermint/abci/types"
-	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/libs/rand"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

blockchain/v1/reactor_fsm_test.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	cmn "github.com/tendermint/tendermint/libs/common"
+	"github.com/tendermint/tendermint/libs/rand"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

crypto/multisig/bitarray/compact_bit_array_test.go
```diff
@@ -7,12 +7,12 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	cmn "github.com/tendermint/tendermint/libs/common"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
```
melekes: better to be consistent I think. i.e. use 1 name everywhere - `tmrand`

marbar3778: I tend to disagree because there isn't a need to alias if nothing in the import section isn't conflicting. There are places we are using `mrand` for `math/rand` and `crand` for `crypto/rand` and places we are not aliasing them.  If this isn't a good practice to follow then I can change it and do this going forward.

melekes: okay, let's leave it then

Issue #3983, opened 2019-09-12
Start block and highest block syncing info on status rpc call


austinabell:
Currently, the data that is returned by the `/status` endpoint regarding sync info is:

```
    "sync_info": {
      "latest_block_hash": "string",
      "latest_app_hash": "string",
      "latest_block_height": "string",
      "latest_block_time": "string",
      "catching_up": true
    },
```

But it would be beneficial to also see the block the sync started at and also the highest block the node is syncing toward, like the [eth api](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_syncing), instead of just outputting the current block.

I assume this part is assumed from the feature request, but also the syncing info through the RPC client interface to be updated to match this change.

PR #4251, opened 2019-12-15
rpc: Added highest block in status RPC method

princesinha19:
I have added `highest_block` in `status` RPC method

Fixes (Feature): #3983 (partially)

### DIFF
--- a/rpc/core/status.go
+++ b/rpc/core/status.go
@@ -45,11 +45,12 @@ func Status(ctx *rpctypes.Context) (*ctypes.ResultStatus, error) {
 	result := &ctypes.ResultStatus{
 		NodeInfo: p2pTransport.NodeInfo().(p2p.DefaultNodeInfo),
 		SyncInfo: ctypes.SyncInfo{
-			LatestBlockHash:   latestBlockHash,
-			LatestAppHash:     latestAppHash,
-			LatestBlockHeight: latestHeight,
-			LatestBlockTime:   latestBlockTime,
-			CatchingUp:        consensusReactor.FastSync(),
+			LatestBlockHash:    latestBlockHash,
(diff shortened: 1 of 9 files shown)


rpc/core/status.go
```diff
@@ -45,11 +45,12 @@ func Status(ctx *rpctypes.Context) (*ctypes.ResultStatus, error) {
...
 		NodeInfo: p2pTransport.NodeInfo().(p2p.DefaultNodeInfo),
 		SyncInfo: ctypes.SyncInfo{
-			LatestBlockHash:   latestBlockHash,
-			LatestAppHash:     latestAppHash,
-			LatestBlockHeight: latestHeight,
-			LatestBlockTime:   latestBlockTime,
-			CatchingUp:        consensusReactor.FastSync(),
+			LatestBlockHash:    latestBlockHash,
+			LatestAppHash:      latestAppHash,
+			CurrentBlockHeight: blockStore.Height(),
```
melekes: why rename `LatestBlockHeight` to `CurrentBlockHeight`?

a) `Latest` prefix assumes current
b) we would need to rename `LatestBlockTime` to `CurrentBlockTime` (extra breaking change)

rpc/core/status.go
```diff
@@ -45,11 +45,12 @@ func Status(ctx *rpctypes.Context) (*ctypes.ResultStatus, error) {
...
+			HighestBlockHeight: consensusState.GetLastHeight(),
```
melekes: Won't HighestBlockHeight be the same as CurrentBlockHeight?

princesinha19: Yes, correct @melekes. For now, we can close this PR. Later, I will expose internal info in the reactor and then, I will create separate PR for this.

Issue #2956, opened 2018-12-01
Remove usage of rand.Rand


ValarDragon:
Currently all randomness happening in tendermint is going through cmn.Rand. This is done under the guise of cmn.Rand being "thread-safe randomness". cmn.Rand is just a global random instance, but with a mutex lock on every single call.

We should instead have each of these distinct threads getting its own randomness instance. This avoids the mutex lock, and more fundamentally is getting its randomness via that thread owning its randomness instance. Having each thread have its own randomness is inherently safer (any issue one thread has won't cascade into other threads), and is much faster due to avoiding contention on that mutex. 

We _don't_ need a global randomness instance to seed our per-thread PRG's. This is because as stated in the [godoc](https://golang.org/pkg/math/rand/), the default randomness instance is thread safe.

marbar3778:
I believe this "proposal accepted"

usamahamid:
@marbar3778, Is this still valid/open? Is there any other arguments for retaining the Mutex Lock? 
I noticed you are refactoring the whole of cmn package (to tendermint), will issue be fixed as part of that?

Just asking since I think this would be good issue to get started.

marbar3778:
I recently removed the `common` pkg so this issue would now relate to the `rand` pkg.  This issue is still valid and available. It was not fixed in the removal of the common pkg.

melekes:
Some work was done in https://github.com/tendermint/tendermint/pull/4252 (removing grand)

liamsi:
Although this is only used in tests etc, this library now often called tmrand.Rand instead of cmn.Rand really should just be deleted. If I have some time during the weekends, I'll try to replace it with "math/rand" or a super lightweight wrapper around math/rand.

PR #4252, opened 2019-12-16
libs: remove mutex from lib

AyushKaul:
Removed mutex from libs/rand, for threads with independent randomness.

Fixes: #2956 

* [x] Wrote tests


libs/rand/random_test.go
```diff
@@ -74,21 +71,6 @@ func testThemAll() string {
 	return out.String()
 }
 
-func TestRngConcurrencySafety(t *testing.T) {
```
melekes: why remove?

libs/rand/random.go
```diff
@@ -17,9 +16,8 @@ const (
 // They all utilize math/rand's prng internally.
 //
 // All of the methods here are suitable for concurrent use.
-// This is achieved by using a mutex lock on all of the provided methods.
+// Different randomness each instance
```
melekes: ? not sure what are trying to say with this sentence

libs/rand/random.go
```diff
@@ -17,9 +16,8 @@ const (
...
 type Rand struct {
-	sync.Mutex
 	rand *mrand.Rand
 }
```
melekes: what do you think about removing grand completely?

libs/rand/random.go
```diff
@@ -141,9 +139,7 @@ func RandPerm(n int) []int {
 // Rand methods
 
 func (r *Rand) Seed(seed int64) {
```
melekes: we'll need to make sure we seed each instance of Rand throughout the code

libs/rand/random.go
pushcodeveryday: rand is thread-safe by default. In the case of `NewSource()`, we can still use grand. But if we don't need global randomness then it's fine.

libs/rand/random.go
pushcodeveryday: this is done for seeding:
```
for i := 0; i < 8; i++ {                                                     
         seed |= uint64(bz[i])                                                    
         seed <<= 8                                                               
}
```
you want to replace this?

melekes: we only need OS random in one / two places, in other we can just seed with the timestamp I believe

libs/rand/random_test.go
pushcodeveryday: I have removed the mutex so this test did not make sense.

libs/rand/random.go
pushcodeveryday: I'll have to re-phrase it.

libs/rand/random.go
pushcodeveryday: I don't see how that can be of use. Do you have something on your mind?

libs/rand/random.go
pushcodeveryday: okay. Let me see if I can help with that.

libs/rand/random.go
pushcodeveryday: I have removed the comment.

pushcodeveryday:
> what do you think about removing grand completely?

@melekes , I have tried to remove grand. Not sure if this is correct.

libs/rand/random.go
```diff
@@ -54,96 +42,115 @@ func (r *Rand) reset(seed int64) {
 // Global functions
 
 func Seed(seed int64) {
```
melekes: I think the idea in the original issue was to get rid of these global functions completely. Instead, each "thread of execution" (i.e. caller) would have a `rand.New()` instance (probably as part of the main struct) and call `rand.XXX` (e.g. rand.Intn) each time it needs a random unit

pushcodeveryday: ok, got it.

melekes:
Could you please rebase your PR against the latest master? Also, do you think you will have time to work on https://github.com/tendermint/tendermint/pull/4252#discussion_r362435151 ?

pushcodeveryday:
I will be able to make the changes in a few days.

melekes:
Closing due to inactivity. @AyushKaul please drop a comment here if you still want to finish this. Thank you!

Issue #3985, opened 2019-09-12
Query block by hash


austinabell:
Query option or endpoint to allow querying block by hash? 

Feature would have a benefit for the project I'm working on, but not sure if this feature would be beneficial and desired for you guys as I can't think of many usable cases for this.

Let me know if this is intentionally left out or unwanted and I can just close this issue and do a workaround to achieve what I need.

melekes:
Done in https://github.com/tendermint/tendermint/pull/4257

PR #4257, opened 2019-12-17
rpc: add method block_by_hash

princesinha19:
I have added the RPC method `block_by_hash`. Which will be used to query a block by hash. It takes parameter as a block hash.

Fixes (Feature): #3985

* [x] Wrote tests


rpc/core/blocks.go
```diff
@@ -81,6 +82,14 @@ func Block(ctx *rpctypes.Context, heightPtr *int64) (*ctypes.ResultBlock, error)
 	return &ctypes.ResultBlock{BlockID: blockMeta.BlockID, Block: block}, nil
 }
 
+// BlockByHash gets block of given hash.
+// More: https://tendermint.com/rpc/#/Info/block_by_hash
+func BlockByHash(ctx *rpctypes.Context, hash bytes.HexBytes) (*ctypes.ResultBlock, error) {
+	blockMeta := blockStore.LoadBlockMetaByHash(hash)
```
melekes: I think it's cheaper to contruct blockmeta from loaded block than loading block meta from the DB.

rpc/core/blocks.go
```diff
@@ -81,6 +82,14 @@ func Block(ctx *rpctypes.Context, heightPtr *int64) (*ctypes.ResultBlock, error)
 	return &ctypes.ResultBlock{BlockID: blockMeta.BlockID, Block: block}, nil
 }
 
+// BlockByHash gets block of given hash.
+// More: https://tendermint.com/rpc/#/Info/block_by_hash
+func BlockByHash(ctx *rpctypes.Context, hash bytes.HexBytes) (*ctypes.ResultBlock, error) {
```
melekes:
```suggestion
func BlockByHash(ctx *rpctypes.Context, hash []byte) (*ctypes.ResultBlock, error) {
```

What is the purpose of hash being `bytes.HexBytes`?

consensus/replay_test.go
```diff
@@ -1062,13 +1063,23 @@ func newMockBlockStore(config *cfg.Config, params types.ConsensusParams) *mockBl
 
 func (bs *mockBlockStore) Height() int64                       { return int64(len(bs.chain)) }
 func (bs *mockBlockStore) LoadBlock(height int64) *types.Block { return bs.chain[height-1] }
+func (bs *mockBlockStore) LoadBlockByHash(hash libBytes.HexBytes) *types.Block {
```
melekes: same question here

rpc/swagger/swagger.yaml
```diff
@@ -550,6 +550,35 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorResponse"
+  /block_by_hash:
+    get:
+      summary: Get block of specified hash
```
melekes:
```suggestion
      summary: Get block by hash
```

rpc/swagger/swagger.yaml
```diff
@@ -550,6 +550,35 @@ paths:
...
+      operationId: block_by_hash
+      parameters:
+        - in: query
+          name: hash
+          description: block Hash to retrive
```
melekes:
```suggestion
          description: block hash
```

store/store.go
```diff
@@ -73,6 +74,29 @@ func (bs *BlockStore) LoadBlock(height int64) *types.Block {
 	return block
 }
 
+// LoadBlockByHash returns the block with the given height.
+// If no block is found for that height, it returns nil.
+func (bs *BlockStore) LoadBlockByHash(hash bytes.HexBytes) *types.Block {
+	var blockMeta = bs.LoadBlockMetaByHash(hash)
```
melekes: I think what we need here is to store `hash => height` relation and use it to retrieve the block

```
bz := bs.db.Get(calcBlockHashKey(hash))
	if len(bz) == 0 {
		return nil
	}
height := parseHeight(string(bz))
return LoadBlock(height)
```

rpc/core/blocks.go
princesinha19: Everywhere is bytes.HexBytes.

princesinha19: @melekes Internally the type of bock hash is `bytes.HexBytes`. That's why I have used this.

consensus/replay_test.go
princesinha19: If I will change it to [ ]bytes, which I already did earlier to create PR this gives me an error that, the type is not correct. Maybe I will change type wherever necessary to make this happen.

princesinha19:
@melekes Thanks for your suggestions. I have made the changes accordingly. Please have a look. 
@marbar3778 Thanks for adding tags. Can you please add a `feature`  tag instead of an `enhancement` because it's a feature/

rpc/core/blocks.go
```diff
@@ -81,6 +81,16 @@ func Block(ctx *rpctypes.Context, heightPtr *int64) (*ctypes.ResultBlock, error)
 	return &ctypes.ResultBlock{BlockID: blockMeta.BlockID, Block: block}, nil
 }
 
+// BlockByHash gets block of given hash.
```
melekes:
```suggestion
// BlockByHash gets block by hash.
```

rpc/core/blocks.go
```diff
@@ -81,6 +81,16 @@ func Block(ctx *rpctypes.Context, heightPtr *int64) (*ctypes.ResultBlock, error)
...
+// More: https://tendermint.com/rpc/#/Info/block_by_hash
+func BlockByHash(ctx *rpctypes.Context, hash []byte) (*ctypes.ResultBlock, error) {
+	block := blockStore.LoadBlockByHash(hash)
+	height := block.Height
+
+	blockMeta := blockStore.LoadBlockMeta(height)
```
melekes: easier to construct BlockMeta from block than to load it.

store/store.go
```diff
@@ -73,6 +74,23 @@ func (bs *BlockStore) LoadBlock(height int64) *types.Block {
...
+	bz := bs.db.Get(calcBlockHashKey(hash))
+	if len(bz) == 0 {
+		return nil
+	}
+
+	s := string(bz)
+	height, err := strconv.ParseInt(s, 10, 64)
+
+	if err != nil {
+		panic(errors.Wrap(err, "Error converting string to int64"))
```
melekes:
```suggestion
		panic(errors.Wrapf(err, "failed to extract height from %s", s))
```

store/store.go
```diff
@@ -159,11 +180,12 @@ func (bs *BlockStore) SaveBlock(block *types.Block, blockParts *types.PartSet, s
 	blockMeta := types.NewBlockMeta(block, blockParts)
 	metaBytes := cdc.MustMarshalBinaryBare(blockMeta)
 	bs.db.Set(calcBlockMetaKey(height), metaBytes)
+	bs.db.Set(calcBlockHashKey(hash), []byte(strconv.FormatInt(height, 10)))
 
 	// Save block parts
 	for i := 0; i < blockParts.Total(); i++ {
 		part := blockParts.GetPart(i)
-		bs.saveBlockPart(height, i, part)
+		bs.saveBlockPart(height, hash, i, part)
```
melekes: why this is needed?

store/store.go
```diff
@@ -213,6 +235,10 @@ func calcSeenCommitKey(height int64) []byte {
 	return []byte(fmt.Sprintf("SC:%v", height))
 }
 
+func calcBlockHashKey(hash []byte) []byte {
+	return []byte(fmt.Sprintf("H:%v", hash))
```
melekes:
```suggestion
	return []byte(fmt.Sprintf("BH:%v", hash))
```

prefix must be unique!

store/store.go
princesinha19: Yes, correct. Thanks.

rpc/core/blocks.go
princesinha19: @melekes Everywhere in Block and Commit, we are using LoadBlockMeta to get the BlockMeta. So, Should I change this? I thought it will make consistent code.
Also, To construct blockMeta we also need BlockParts along with blocks, I guess. Correct me if I am wrong here.

melekes: Right, thanks!

store/store.go
```diff
@@ -73,6 +74,23 @@ func (bs *BlockStore) LoadBlock(height int64) *types.Block {
 	return block
 }
 
+// LoadBlockByHash returns the block with the given height.
```
melekes:
```suggestion
// LoadBlockByHash returns the block with the given hash.
```

store/store.go
```diff
@@ -73,6 +74,23 @@ func (bs *BlockStore) LoadBlock(height int64) *types.Block {
...
+// If no block is found for that height, it returns nil.
```
melekes:
```suggestion
// If no block is found for that hash, it returns nil.
// Panics if it fails to parse height associated with the given hash.
```

store/store.go
```diff
@@ -159,6 +180,7 @@ func (bs *BlockStore) SaveBlock(block *types.Block, blockParts *types.PartSet, s
 	blockMeta := types.NewBlockMeta(block, blockParts)
 	metaBytes := cdc.MustMarshalBinaryBare(blockMeta)
 	bs.db.Set(calcBlockMetaKey(height), metaBytes)
+	bs.db.Set(calcBlockHashKey(hash), []byte(strconv.FormatInt(height, 10)))
```
melekes:
```suggestion
	bs.db.Set(calcBlockHashKey(hash), []byte(fmt.Sprintf("%d", height)))
```

Sprintf is easier imho

store/store.go
```diff
@@ -213,6 +236,10 @@ func calcSeenCommitKey(height int64) []byte {
 	return []byte(fmt.Sprintf("SC:%v", height))
 }
 
+func calcBlockHashKey(hash []byte) []byte {
+	return []byte(fmt.Sprintf("BH:%v", hash))
```
melekes:
```suggestion
	return []byte(fmt.Sprintf("BH:%x", hash))
```

princesinha19: Done @melekes.

PR #4258, opened 2019-12-17
lint: golint issue fixes

marbar3778:
- on my local machine golint seems to be stricter than the bot 

* [ ] Wrote tests

### DIFF
--- a/p2p/switch.go
+++ b/p2p/switch.go
@@ -604,7 +604,7 @@ func (sw *Switch) acceptRoutine() {
 			onPeerError:  sw.StopPeerForError,
 			reactorsByCh: sw.reactorsByCh,
 			metrics:      sw.metrics,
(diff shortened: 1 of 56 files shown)


p2p/switch.go
```diff
@@ -585,9 +585,7 @@ func (sw *Switch) AddUnconditionalPeerIDs(ids []string) error {
 }
 
 func (sw *Switch) isPeerPersistentFn() func(*NetAddress) bool {
```
melekes: do we need this func at all?

marbar3778: I can't see a reason to keep it, but I also can't speak why its there. If I remove it then everything works as intended, so if want to remove we can

melekes: Let's remove

Issue #4059, opened 2019-10-15
double-signing protection 


dlguddus:
1. problem definition
- validators often mistakenly run duplicated validators to cause double-signing incident
- this proposed feature is to reduce the risk of mistaken double-signing incident
- when we think of such serious impact on double-signing incident, it is very reasonable to have multiple protection algorithm built in gaiad

2. new feature
- "gaiad start" has default setting to not participate on consensus, aka "fullnode as default"
- after starting gaiad, user enter command "gaiad activate-validation --from {key} --pubkey {valconspub}" with the signature from the {key}
- when gaiad receive this command, gaiad will look up recent(predifined block heights) consensus result and search for any consensus votes by given validator operator of the {key} AND also by given {valconspub}
- if there exists any vote from given validator operator or {valconspub}, gaiad will reject(or reconfirm) the activate-validation request and stays in fullnode
- if there exists no vote from given {key}, gaiad will transform this node from fullnode to validator which only use the private key paired with given {valconspub}
- when priv_validator_key changes while node is running as validator, the given {valconspub} and the new pubkey of priv_validator_key does not match therefore the node will not sign any consensus vote.
- all default setting can be customized by config.toml

3. B-Harvest wants to tackle this when we confirm enough support from community.

marbar3778:
I believe this can be considered a duplicate of https://github.com/tendermint/tendermint/issues/2237#issuecomment-520477659. We have discussed this on a few calls, just not gotten around to working on it.

dlguddus:
> I believe this can be considered a duplicate of [#2237 (comment)](https://github.com/tendermint/tendermint/issues/2237#issuecomment-520477659). We have discussed this on a few calls, just not gotten around to working on it.

We will review #2237 and provide updated spec so that it can include the "mode" abstraction! Thank you for connecting this!

alexanderbez:
Seems reasonable. One question I have surrounds the following:

```
when gaiad receive this command, gaiad will look up recent(predifined block heights) consensus result and search for any consensus votes by given validator operator of the {key} AND also by given {valconspub}
```

Will it only need to look past "unbonding period" # of blocks?

dlguddus:
@alexanderbez no. Practically it will be definitely much less than that. I am imagining like 50 blocks or less. 

The main purpose of this feature is to double check that another node is "currently" validating with same valoper or valconspub on the chain.

I think maximum look up height might be around several hundred blocks because this usecase does not need more.

[...]

PR #4262, opened 2019-12-18
[ADR] ADR-051: Double Signing Risk Reduction

dongsam:
### ADR-051: Double Signing Risk Reduction
To provide a risk reduction method for double signing incidents mistakenly executed by validators

### Refs
- double-signing protection(https://github.com/tendermint/tendermint/issues/4059)

* [ ] Wrote tests


docs/architecture/adr-051-double-signing-protection-with-tendermint-mode.md
```diff
@@ -0,0 +1,104 @@
...
+    - `host:26657/status`
+        - return empty `validator_info` when fullnode mode
+        - add return `mode` of this node
+    - no rpc server in seed mode
+- Where to modify in codebase
+    - Add  switch for `config.Mode` on `node/node.go:DefaultNewNode`
+    - If `config.Mode==validator`, call default `NewNode` (current logic)
+    - If `config.Mode==fullnode`, call `NewNode` with `nil` `privValidator` (do not load or generation)
+        - Need to add exception routine for `nil` `privValidator` to related functions
+    - If `config.Mode==seed`, call `NewSeedNode` (seed version of `node/node.go:NewNode`)
```
melekes: I think we should extract seed mode completely into its own package

docs/architecture/adr-051-double-signing-protection-with-tendermint-mode.md
```diff
@@ -0,0 +1,104 @@
...
+
+### 2 ) Double signing prevention method
+
+- Methodology : query recent consensus results to find out whether node's consensus key is used on consensus recently or not
+- When to check
+    - When the state machine starts `ConsensusReactor` after fully synced
+    - When the node is validator ( with privValidator )
+    - When `cs.config.DoubleSignCheckHeight > 0`
+- How to check
+    1. When a validator is transformed from syncing status to fully synced status, the state machine check recent N blocks to find out whether there exists consensus votes using the validator's consensus key
```
melekes:
```suggestion
    1. When a validator is transformed from syncing status to fully synced status, the state machine check recent N blocks (`latest_height - double_sign_check_height`) to find out whether there exists consensus votes using the validator's consensus key
```

docs/architecture/adr-051-double-signing-protection-with-tendermint-mode.md
```diff
@@ -0,0 +1,104 @@
...
+    - If `config.Mode==seed`, call `NewSeedNode` (seed version of `node/node.go:NewNode`)
+        - Need to add exception routine for `nil` `reactor`, `component` to related functions
+
+### 2 ) Double signing prevention method
+
+- Methodology : query recent consensus results to find out whether node's consensus key is used on consensus recently or not
+- When to check
+    - When the state machine starts `ConsensusReactor` after fully synced
+    - When the node is validator ( with privValidator )
+    - When `cs.config.DoubleSignCheckHeight > 0`
```
melekes: Can you explain why it has to be a parameter? I.e. why don't just use some constant like check 10 recent blocks?

dongsam: When the prevention method is on, restarting a validator node immediately will panic because the node itself voted on consensus with the same consensus key. So, validators should stop the state machine, wait for DoubleSignCheckHeight blocks and then restart the state machine to avoid panic stop.
I think some users don't want this side effect, so they want to turn it off or adjust the checking height as parameter,  they'll be able to turn it off by setting DoubleSignCheckHeight = 0

docs/architecture/adr-051-double-signing-protection-with-tendermint-mode.md
dongsam: you mean it should be independent binary for seed mode, not flag option?

dongsam:
> Proposed modes refactoring has little to do with double signing protection (although it would help). I wonder if it makes sense to separate this into 2 ADR because modes refactoring will take much bigger time than "Double signing prevention method".

Sure, Let me seperate into 2 ADR, This ADR will only cover Double signing Protection and I'll PR another ADR about Tendermint Mode

docs/architecture/adr-051-double-signing-protection-with-tendermint-mode.md
melekes: > I think some users don't want this side effect

if it's only ON in `validator` mode, then choosing safety is fine in my opinion.

dongsam: yeah that could be good option, do you want specify that in ADR ?  or we could derive a conclusion on implementation PR process

docs/architecture/adr-051-double-signing-protection-with-tendermint-mode.md
melekes: no, I meant `Go package` ~ extract seed mode code from p2p/pex/pex_reactor.go into its own Go package.

dongsam: oh I see, that would be better

marbar3778: was it decided how this will be handled? in adr-052 it says under one binary but here there is a conversation to have it separated.

dongsam: I think what melekes said means not a binary separation but a refactoring of the go package directory structure, right?

marbar3778: ah whoops sorry misread this

docs/architecture/adr-051-double-signing-protection.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Positive
+
+- Validators can avoid double signing incident by mistakes. (eg. If another validator node is voting on consensus, starting new validator node with same consensus key will cause panic stop of the state machine because consensus votes with the consensus key are found in recent blocks)
+- We expect this method will prevent majority of double signing incident by mistakes.
+
+### Negative
+
+Double signing prevention method
+- When the prevention method is on, restarting a validator node will panic because the node itself voted on consensus with the same consensus key. So, validators should stop the state machine, wait for some blocks, and then restart the state machine to avoid panic stop.
```
marbar3778: is there a way that when the node comes back online and it sees a signed block from the same key at x, it waits for y blocks to see if there are any signatures if not it inserts itself to sign?

docs/architecture/adr-051-double-signing-protection-with-tendermint-mode.md
leoluk: Yes, there definitely needs to be an option to disable this (like when doing enable active-active node coordination at the signer level).

leoluk:
As stated in #4059, the mechanism proposed only reduces the risk of double signing, but cannot eliminate it - for instance, two nodes can still _start_ signing at the same height. This can happen easily with automation (running a playbook on all nodes, mistakenly scaling up a deployment...).

It's good to have defense in depth, but perhaps don't call it "double signing protection"? Maybe "double signing risk reduction"? Otherwise, it might lead to a false sense of security.

docs/architecture/adr-051-double-signing-protection.md
dlguddus: I think it is easily doable with an external script if we want to. But we are little bit reluctant to put any feature for any convinience needs because it might increase operational mistake.

I think this should be rather little bit inconvenient but simple and not complicated

dlguddus:
I think the word "protection" includes the existence of failure risk even with the protection. It generally means reducing risk from my sense.

But maybe I am not aware of more accurate general perception of the word? How others think?

melekes:
> I think the word "protection" includes the existence of failure risk even with the protection. It generally means reducing risk from my sense.

That's not how I understand it. I like "double signing risk reduction". Or we can name it "double signing prevention measures".

dlguddus:
> > I think the word "protection" includes the existence of failure risk even with the protection. It generally means reducing risk from my sense.
> 
> That's not how I understand it. I like "double signing risk reduction". Or we can name it "double signing prevention measures".

Then, I think both are good options for new name. Let's fix it with "double signing risk reduction" then? How do you think @dongsam ?

dlguddus:
One improvement suggestion

Problem) like leo stated, if two validator nodes start at the same time, this risk reduction method will not prevent double signing.

One suggested solution) each validator nodes wait n(a random number from 0 to 9) blocks before starting validation. Then the double signing risk will be reduced 90% because the method will fail only if two nodes randomly pick same n.

Although it is not completely preventing, I guess 90% is a good enough statistical improvement. We can make it 99% by ranging n to 99 but I am not sure it worth it over inconvenience of long wait time for validator operators.

Of course operator can manually turn off this feature if he/she is confident enough.

Issue #3773, opened 2019-07-04
Add Prometheus observability of last signed height per validator


mdyring:
We'd like to observe individual last signed height for validators  via Prometheus by requesting a tendermint_consensus_last_signed_height metric with the following labels:

* address (example: "E33032D801DA4D124FBF696DEF7832E415EA74EE")
* operator (example: "cosmosvaloper1lc4jjw8zveuuj3njhy70zl23vgmpds2xg830hn")
* chain_id (example: cosmoshub-2)

As some might object to the cardinality of this (100+ validators on cosmoshub-2), a config option to enable this could be beneficial as well. Request is related to per-validator voting power in #3083, so a common toggle for per-validator observability would be suitable.

Thanks for your consideration. :-)

ebuchman:
Note `operator` isn't a concept native to Tendermint. Tendermint could provide the validator's consensus pubkey, but the operator is a Cosmos-SDK concept. I wonder if the SDK is a better place for per-validator metrics?

mdyring:
Hm, seems nice to provide for any Tendermint-based chain I think it makes sense to provide it here.

Providing address, consensus pubkey and chain-id labels would be neat!

mdyring:
On second thought, would be messy to have both Tendermint and cosmos-sdk endpoints for Prometheus. I suspect cosmos-sdk is doing all the Prometheus-related work today, so if that it the case it might make better sense to do it from there. @jackzampolin any thoughts?

mdyring:
This metric is useful for detecting external signing errors [such as this](https://twitter.com/validator_net/status/1156386709648662528?s=20).

Currently we're using [our own tool](https://github.com/validator-network/votewatcher) for monitoring the last signed height, but would be nice to offer out of the box IMHO.

Issue #3083, opened 2019-01-04
Add Prometheus metric tendermint_consensus_validator_power


mdyring:
We'd like to observe individual validator power via Prometheus by requesting a *tendermint_consensus_validator_power* metric with the following labels:
* address (example: "E33032D801DA4D124FBF696DEF7832E415EA74EE")
* operator (example: "cosmosvaloper1lc4jjw8zveuuj3njhy70zl23vgmpds2xg830hn")
* chain_id (see #3082)

Thanks for your consideration. :-)

PR #4263, opened 2019-12-18
prometheus/metrics: three new metrics for consensus

marbar3778:
- add consensus_validator_power metric so if a node is a validator it can see its own power in prometheus
- add last_signed_height metric so if a node is a validator it can be aware of which was the most recent height it signed.

- closes: #3773
- closes: #3083

@mdyring 

* [ ] Wrote tests


consensus/state.go
```diff
@@ -1470,6 +1470,11 @@ func (cs *State) recordMetrics(height int64, block *types.Block) {
 		if i >= len(block.LastCommit.Signatures) {
 			break
 		}
+
+		if cs.privValidator != nil || cs.Validators.HasAddress(cs.privValidator.GetPubKey().Address()) && val.Address.String() == cs.privValidator.GetPubKey().Address().String() {
```
golangcibot: line is 173 characters (from `lll`)

consensus/state.go
```diff
@@ -1486,6 +1486,13 @@ func (cs *State) recordMetrics(height int64, block *types.Block) {
 
 		for i, val := range cs.LastValidators.Validators {
 			commitSig := block.LastCommit.Signatures[i]
+			if cs.privValidator != nil && val.Address.String() == cs.privValidator.GetPubKey().Address().String() {
```
```suggestion
			if cs.privValidator != nil && bytes.Equal(val.Address, cs.privValidator.GetPubKey().Address()) {
```

melekes:
This also needs a changelog_pending entry.

CHANGELOG_PENDING.md
```diff
@@ -113,6 +113,10 @@ program](https://hackerone.com/tendermint).
 - [cli] \#4234 Add `--db_backend and --db_dir` flags (@princesinha19)
 - [cli] \#4113 Add optional `--genesis_hash` flag to check genesis hash upon startup
 - [config] \#3831 Add support for [RocksDB](https://rocksdb.org/) (@Stumble)
+- [metrics] \#4263 Added
```
```suggestion
- [metrics] \#4263 Add
```

CHANGELOG_PENDING.md
```diff
@@ -113,6 +113,10 @@ program](https://hackerone.com/tendermint).
 - [cli] \#4234 Add `--db_backend and --db_dir` flags (@princesinha19)
 - [cli] \#4113 Add optional `--genesis_hash` flag to check genesis hash upon startup
 - [config] \#3831 Add support for [RocksDB](https://rocksdb.org/) (@Stumble)
+- [metrics] \#4263 Add
+  - `consensus_validator_power`: track your validators validator power
```
```suggestion
  - `consensus_validator_power`: track your validators power
```

melekes:
What makes these metrics validator specific? I do not see any additional labels

marbar3778:
add a label with the validator address to the added metrics

consensus/metrics.go
```diff
@@ -81,12 +86,24 @@ func PrometheusMetrics(namespace string, labelsAndValues ...string) *Metrics {
...
 			Subsystem: MetricsSubsystem,
 			Name:      "validators_power",
 			Help:      "Total power of all validators.",
 		}, labels).With(labelsAndValues...),
+		ValidatorPower: prometheus.NewGaugeFrom(stdprometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: MetricsSubsystem,
+			Name:      "validator_power",
+			Help:      "Power of a validator.",
+		}, append(labels, "address")).With(labelsAndValues...),
```
melekes: don't you think address is too abstract? maybe

```suggestion
		}, append(labels, "validator_address")).With(labelsAndValues...),
```

marbar3778: thought of this right when I pushed and made the change locally then switched branches and deleted the change

mdyring:
Thanks  @marbar3778 for your work on this.

@leoluk has previously suggested having a "total missed block" counter per validator, I think that could be a valuable addition to detect missed blocks that occurs sporadically.

Also worth considering is providing these metrics for all validators in the set, not just the local one found in priv_validator.

Con: Consumes more resources to store/query 125 samples instead of 1. 

Pro: Sentries can collect samples so there is a higher degree of certainty of getting accurate data, even if there is a network partition or similar restricting Prometheus from scraping the validator.

Perhaps a config option is the most sensible approach to this, but personally I'd be comfortable with collecting this data for the current 125 validators.

marbar3778:
I can add the missing of blocks to this pr for individual validators but to keep the scope smaller, it would be good to add the second part, adding these metrics for the entire set, to a followup PR.

CHANGELOG_PENDING.md
```diff
@@ -113,6 +113,11 @@ program](https://hackerone.com/tendermint).
 - [cli] \#4234 Add `--db_backend and --db_dir` flags (@princesinha19)
 - [cli] \#4113 Add optional `--genesis_hash` flag to check genesis hash upon startup
 - [config] \#3831 Add support for [RocksDB](https://rocksdb.org/) (@Stumble)
+- [metrics] \#4263 Add
+  - `consensus_validator_power`: track your validators power
+  - `last_signed_height`: track at which height the validator last signed
```
```suggestion
  - `consensus_last_signed_height`: track at which height the validator last signed
```

CHANGELOG_PENDING.md
```diff
@@ -113,6 +113,11 @@ program](https://hackerone.com/tendermint).
...
+  - `missed_blocks`: total amount of missed blocks for a validator
```
```suggestion
  - `consensus_missed_blocks`: total amount of missed blocks for a validator
```

consensus/metrics.go
```diff
@@ -81,12 +88,30 @@ func PrometheusMetrics(namespace string, labelsAndValues ...string) *Metrics {
 			Name:      "validators",
 			Help:      "Number of validators.",
 		}, labels).With(labelsAndValues...),
+		LastSignedHeight: prometheus.NewGaugeFrom(stdprometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: MetricsSubsystem,
+			Name:      "last_signed_height",
```
melekes: I think `validator_` prefix is necessary as consensus scope does not necessary mean that. See https://docs.tendermint.com/master/tendermint-core/metrics.html

consensus/metrics.go
```diff
@@ -81,12 +88,30 @@ func PrometheusMetrics(namespace string, labelsAndValues ...string) *Metrics {
...
+			Help:      "Last signed height for a validator",
+		}, append(labels, "validator_address")).With(labelsAndValues...),
+		MissedBlocks: prometheus.NewGaugeFrom(stdprometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: MetricsSubsystem,
+			Name:      "missed_blocks",
```
melekes: same here

consensus/metrics.go
```diff
@@ -81,12 +88,30 @@ func PrometheusMetrics(namespace string, labelsAndValues ...string) *Metrics {
...
 			Namespace: namespace,
 			Subsystem: MetricsSubsystem,
 			Name:      "validators_power",
 			Help:      "Total power of all validators.",
 		}, labels).With(labelsAndValues...),
+		ValidatorPower: prometheus.NewGaugeFrom(stdprometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: MetricsSubsystem,
+			Name:      "validator_power",
+			Help:      "Power of a validator.",
```
melekes:
```suggestion
			Help:      "Power of a validator",
```

consensus/state.go
```diff
@@ -1486,6 +1486,19 @@ func (cs *State) recordMetrics(height int64, block *types.Block) {
 
 		for i, val := range cs.LastValidators.Validators {
 			commitSig := block.LastCommit.Signatures[i]
+			privValAddress := cs.privValidator.GetPubKey().Address()
+			if cs.privValidator != nil && bytes.Equal(val.Address, privValAddress) {
+				label := []string{
+					"validator_address", privValAddress.String(),
+				}
+				cs.metrics.ValidatorPower.With(label...).Set(float64(val.VotingPower))
+				if !commitSig.Absent() {
```
leoluk: This checks for BlockIDFlagAbsent. Wouldn't BlockIDFlagNil (nil vote) also count as a miss?

marbar3778: https://github.com/tendermint/tendermint/pull/4283 did the follow up here

PR #4266, opened 2019-12-19
test: branch for fix of ci

marbar3778:
- testing replacing syscall with os.FindProces

* [ ] Wrote tests

### DIFF
--- a/cmd/tendermint/commands/debug/kill.go
+++ b/cmd/tendermint/commands/debug/kill.go
@@ -121,7 +121,12 @@ func killProc(pid uint64, dir string) error {
 	go func() {
 		// Killing the Tendermint process with the '-ABRT|-6' signal will result in
 		// a goroutine stacktrace.
-		if err := syscall.Kill(int(pid), syscall.SIGABRT); err != nil {
+		p, err := os.FindProcess(os.Getpid())
+		if err != nil {
+			fmt.Printf("failed to find PID to kill Tendermint process: %s", err)
+		}
+		if err = p.Signal(syscall.SIGABRT); err != nil {
+			// if err := syscall.Kill(int(pid), syscall.SIGABRT); err != nil {
 			fmt.Fprintf(os.Stderr, "failed to kill Tendermint process: %s", err)
 		}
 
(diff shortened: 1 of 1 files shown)


cmd/tendermint/commands/debug/kill.go
```diff
@@ -121,7 +121,12 @@ func killProc(pid uint64, dir string) error {
 	go func() {
 		// Killing the Tendermint process with the '-ABRT|-6' signal will result in
 		// a goroutine stacktrace.
-		if err := syscall.Kill(int(pid), syscall.SIGABRT); err != nil {
+		p, err := os.FindProcess(os.Getpid())
+		if err != nil {
+			fmt.Printf("failed to find PID to kill Tendermint process: %s", err)
+		}
+		if err = p.Signal(syscall.SIGABRT); err != nil {
```
alessio: If err is not nil here, p is likely nil

cmd/tendermint/commands/debug/kill.go
```diff
@@ -121,7 +121,12 @@ func killProc(pid uint64, dir string) error {
 	go func() {
 		// Killing the Tendermint process with the '-ABRT|-6' signal will result in
 		// a goroutine stacktrace.
-		if err := syscall.Kill(int(pid), syscall.SIGABRT); err != nil {
+		p, err := os.FindProcess(os.Getpid())
+		if err != nil {
+			fmt.Printf("failed to find PID to kill Tendermint process: %s", err)
```
melekes:
```suggestion
			fmt.Fprintf(os.Stderr, "failed to find PID to kill Tendermint process: %s", err)
```

cmd/tendermint/commands/debug/kill.go
```diff
@@ -121,7 +121,12 @@ func killProc(pid uint64, dir string) error {
...
+		}
+		if err = p.Signal(syscall.SIGABRT); err != nil {
+			// if err := syscall.Kill(int(pid), syscall.SIGABRT); err != nil {
```
melekes: remove?

cmd/tendermint/commands/debug/kill.go
alessio: Any comment on this?

melekes: Oops. Looks like we need an `else` branch. Thanks @alessio!

melekes: https://github.com/tendermint/tendermint/pull/4271

PR #4272, opened 2019-12-21
[crypto] fix sr25519 from raw import

sunnya97:
Fixes sr25519 pubkey generation and signing when importing from raw bytes

* [ ] Wrote tests


marbar3778:
does this need a bump in the dependency? @sunnya97 If not could we do it anyways?

marbar3778:
also in the last PR to add sr25519 we didnt add it to `crypto/multisig` codec. Should we add it in this PR or this was on purpose?

sunnya97:
@marbar3778 You mean bump go-schnorrkel?  Yeah we can do that.

Also, good catch.  SR25519 should really have its own custom multisig eventually (because it can be an aggregate signature),  but there's no reason it can't also be in a normal multisig with other key types as well.  I'll add it to the multisig codec this PR as well.

marbar3778:
yea,  bump go-schnorrkel if possible

sunnya97:
Bumped `go-schnorrkel` 👍

PR #4274, opened 2019-12-28
refactor(rpc): return an error on invalid remote (#3953)

### LINKED ISSUE #3953: Modify NewHTTP, NewXXXClient funcs to return errors
if the given remote address is invalid.

See https://github.com/tendermint/tendermint/pull/3779#issuecomment-528332719

mrekucci:
The `New*` client functions return an error instead
of panicking when the remote address is invalid.

Fixes #3953

### DIFF
--- a/cmd/tendermint/commands/lite.go
+++ b/cmd/tendermint/commands/lite.go
@@ -80,7 +80,10 @@ func runProxy(cmd *cobra.Command, args []string) error {
 
 	// First, connect a client
 	logger.Info("Connecting to source HTTP client...")
-	node := rpcclient.NewHTTP(nodeAddr, "/websocket")
+	node, err := rpcclient.NewHTTP(nodeAddr, "/websocket")
+	if err != nil {
+		return errors.Wrap(err, "new HTTP client")
+	}
 
 	logger.Info("Constructing Verifier...")
 	cert, err := proxy.NewVerifier(chainID, home, node, logger, cacheSize)
--- a/rpc/client/httpclient.go
+++ b/rpc/client/httpclient.go
@@ -87,29 +87,38 @@ var _ rpcClient = (*baseRPCClient)(nil)
 
 // NewHTTP takes a remote endpoint in the form <protocol>://<host>:<port> and
 // the websocket path (which always seems to be "/websocket")
-// The function panics if the provided remote is invalid.<Paste>
-func NewHTTP(remote, wsEndpoint string) *HTTP {
-	httpClient := rpcclient.DefaultHTTPClient(remote)
+// An error is returned on invalid remote. The function panics when remote is nil.
+func NewHTTP(remote, wsEndpoint string) (*HTTP, error) {
+	httpClient, err := rpcclient.DefaultHTTPClient(remote)
+	if err != nil {
+		return nil, err
+	}
 	return NewHTTPWithClient(remote, wsEndpoint, httpClient)
 }
 
-// NewHTTPWithClient allows for setting a custom http client. See NewHTTP
-// The function panics if the provided client is nil or remote is invalid.
-func NewHTTPWithClient(remote, wsEndpoint string, client *http.Client) *HTTP {
+// NewHTTPWithClient allows for setting a custom http client (See NewHTTP).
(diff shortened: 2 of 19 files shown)


rpc/client/httpclient.go
```diff
@@ -87,29 +87,38 @@ var _ rpcClient = (*baseRPCClient)(nil)
...
+	rc, err := rpcclient.NewJSONRPCClientWithHTTPClient(remote, client)
+	if err != nil {
+		return nil, err
+	}
 	cdc := rc.Codec()
 	ctypes.RegisterAmino(cdc)
 	rc.SetCodec(cdc)
 
-	return &HTTP{
+	httpClient := &HTTP{
```
melekes: why not

```
return &HTTP{...}, nil
```

?

rpc/lib/client/http_json_client.go
```diff
@@ -67,27 +67,35 @@ var _ JSONRPCCaller = (*JSONRPCClient)(nil)
...
 	}
 
 	clientAddress, err := toClientAddress(remote)
 	if err != nil {
-		panic(fmt.Sprintf("invalid remote %s: %s", remote, err))
+		return nil, fmt.Errorf("invalid remote %s: %s", remote, err)
 	}
 
-	return &JSONRPCClient{
+	rpcClient := &JSONRPCClient{
```
melekes: same question

rpc/lib/client/http_json_client.go
```diff
@@ -298,16 +306,10 @@ func parseRemoteAddr(remoteAddr string) (network string, s string, err error) {
...
-		return nil, err
-	}
-}
-
-func makeHTTPDialer(remoteAddr string) func(string, string) (net.Conn, error) {
+func makeHTTPDialer(remoteAddr string) (func(string, string) (net.Conn, error), error) {
 	protocol, address, err := parseRemoteAddr(remoteAddr)
 	if err != nil {
-		return makeErrorDialer(err)
+		return nil, err
```
melekes: wouldn't this result in panic when someone tries to call `nil` dialFn?

rpc/client/httpclient.go
mrekucci: It is a matter of taste, but I personally consider it more readable than:
```
return &HTTP{
	rpc:           rc,
	remote:        remote,
	baseRPCClient: &baseRPCClient{caller: rc},
	WSEvents:      newWSEvents(cdc, remote, wsEndpoint),
}, nil
```
Other than that, I don't have any problem with your suggested version and can change the code if this style is what the project is using.

rpc/lib/client/http_json_client.go
mrekucci: See the comment above.

cmd/tendermint/commands/lite.go
```diff
@@ -80,7 +80,10 @@ func runProxy(cmd *cobra.Command, args []string) error {
 
 	// First, connect a client
 	logger.Info("Connecting to source HTTP client...")
-	node := rpcclient.NewHTTP(nodeAddr, "/websocket")
+	node, err := rpcclient.NewHTTP(nodeAddr, "/websocket")
+	if err != nil {
+		return errors.Wrap(err, "connection failed")
```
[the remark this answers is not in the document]
mrekucci: 👍

rpc/lib/client/http_json_client.go
mrekucci: Good catch, yes it would if the client using the code wouldn't follow the Golang convention and won't check the error before using the `dialFn`. Also, another Golang convention is to set each non-error returning value to a zero value when an error is returned. The zero value for an interface is nil. See the [Returning Errors Alongside Values](https://www.digitalocean.com/community/tutorials/handling-errors-in-go) section.

I removed the `makeErrorDialer` for clarity, it's better to have less code. But it's also possible that maybe I missed something (some other purpose why the `makeErrorDialer` was returned)...

melekes: > some other purpose why the makeErrorDialer was returned

I doubt that

PR #4283, opened 2020-01-06
metrics: add metric for when commit is nil

marbar3778:
- follow up to #4263
- when a commit is nil, then it should be counted as a missed commit

* [ ] Wrote tests

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -1492,7 +1492,7 @@ func (cs *State) recordMetrics(height int64, block *types.Block) {
 					"validator_address", privValAddress.String(),
 				}
 				cs.metrics.ValidatorPower.With(label...).Set(float64(val.VotingPower))
-				if !commitSig.Absent() {
+				if !commitSig.Absent() || !commitSig.Nil() {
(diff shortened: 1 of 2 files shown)


consensus/state.go
```diff
@@ -1492,7 +1492,7 @@ func (cs *State) recordMetrics(height int64, block *types.Block) {
 					"validator_address", privValAddress.String(),
 				}
 				cs.metrics.ValidatorPower.With(label...).Set(float64(val.VotingPower))
-				if !commitSig.Absent() {
+				if !commitSig.Absent() || !commitSig.Nil() {
```
```suggestion
				if commitSig.ForBlock() {
```

types/block.go
```diff
@@ -479,6 +479,15 @@ func (cs CommitSig) Absent() bool {
 	return cs.BlockIDFlag == BlockIDFlagAbsent
 }
 
+func (cs CommitSig) ForBlock() bool {
```
marbar3778: @melekes is this what you meant?

melekes: yes

Issue #4248, opened 2019-12-13
Enhance RPCClient Dialer with basic authentication / HTTP headers


greg-szabo:
User story:
I'm using a CLI to connect to a Tendermint-based blockchain RPC. I would like to lock down the RPC connection with HTTP Basic Authentication. This requires the JSONRPCClient.Call to add headers to the connection request.

greg-szabo:
What happens in PR 4285:

Changes in http_client.go:
1. Extended JSONRPCClient struct with username and password fields.
2. Modified the JSONRPCClient.Call function to add the username and password headers if they are not empty in the struct.
3. Modified the JSONRPCClient.sendBatch function to add the username and password headers if they are not empty in the struct.
4. Removed the parseRemoteAddr, toClientAddress, toClientAddrAndParse functions and replaced their functionality with a natively parsed "parsedURL" struct and associated functions. This allows the username and password to be extracted from the URL string during parsing and be placed in the struct. The previous method of returning strings was not flexible enough for this.

Changes in ws_client.go:
1. Modified NewWSClient to use the new parsedURL struct and functions.

Test functions were modified so "//" is prefixed when a testing HTTP server address is handed over. The net/url package adheres to some rfc which defines valid address as `//192.168.0.1:443` or such, instead of `192.168.0.1:443`. (not a problem with domain names)

Additional testing was done by building a GaiaCLI with the modified tendermint vendor package and running it on a live network with an nginx server providing the server-side of TLS encryption and user authentication. Worked fine.

greg-szabo:
I've opened the change against master. It's the same functionality that was implemented for the backported version.

greg-szabo:
Anton requested to make the default protocol HTTP, instead of TCP in `JSONRPClient`, when a protocol is not defined. This is a breaking change in some obscure cases.

In the current test implementations, addresses seem to be written with the `tcp://` prefix (or no prefix which defaults to tcp) and the underlying code changes it to `http://` where it is necessary. All test code rely on this assumption.

I'm expecting other developers and other chains to rely on this assumption too. If we change the default to HTTP, we might break other developers' work for not much benefit.

The benefit of defaulting to HTTP would be at the end-user, who wouldn't need to type out `http://` but currently, it is mandatory to add a protocol in the URL at a higher level of the code - so end-users will not benefit from this change. (Also, end-users should use TLS and the `https://` prefix.)

I can implement this change in an additional commit in this same PR but currently, I'm leaning toward opposing this change.

[...]

PR #4285, opened 2020-01-07
RPC client basic authentication with URL parsing refactored

greg-szabo:
Enable basic authentication in the RPC client using the https://username:password@node:port format.

Issue #4248 has details about what was refactored/enhanced (it's not as bad as it looks.)

I'm open to suggestions on where/how the documentation should be updated.

Please note that PR #4284 is superseded with this PR. The reason for this is because both PR makes changes to the same code.

* [X] Referenced an issue explaining the need for the change
* [X] Wrote tests

### DIFF
--- a/rpc/lib/client/http_client.go
+++ b/rpc/lib/client/http_client.go
@@ -28,61 +28,60 @@ const (
 	protoTCP   = "tcp"
 )
 
+// Parsed URL structure
+type parsedURL struct {
+	url.URL
+}
+
+// Parse URL and set defaults (a flexible implementation for 'parseRemoteAddr')
+func newParsedURL(remoteAddr string) (*parsedURL, error) {
+	u, err := url.Parse(remoteAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	// default to tcp if nothing specified
+	if u.Scheme == "" {
+		u.Scheme = protoTCP
+	}
+
+	return &parsedURL{*u}, nil
+}
+
+// Change protocol to HTTP for unknown protocols and TCP protocol (flexible HTTPS downgrade fix) - useful for RPC connections
+func (u *parsedURL) SetDefaultSchemeHTTP() {
+	// protocol to use for http operations, to support both http and https
+	switch u.Scheme {
+	case protoHTTP, protoHTTPS, protoWS, protoWSS:
+		// known protocols not changed
+	default:
+		// default to http for unknown protocols (ex. tcp)
+		u.Scheme = protoHTTP
+	}
+}
+
+// Get full address without the protocol - useful for Dialer connections
+func (u parsedURL) GetHostWithPath() string {
+	// Remove protocol, userinfo and # fragment, assume opaque is empty
+	return u.Host + u.EscapedPath()
+}
+
+// Get a trimmed address (Implements previous 'toClientAddrAndParse') - useful for WS connections
+func (u parsedURL) GetTrimmedHostWithPath() string {
+	// replace / with . for http requests (kvstore domain)
+	return strings.Replace(u.GetHostWithPath(), "/", ".", -1)
+}
+
+// Get a trimmed address with protocol (implements previous 'toClientAddress') - useful as address in RPC connections
+func (u parsedURL) GetTrimmedURL() string {
+	return u.Scheme + "://" + u.GetTrimmedHostWithPath()
+}
+
 // HTTPClient is a common interface for JSONRPCClient and URIClient.
 type HTTPClient interface {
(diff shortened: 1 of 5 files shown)


rpc/lib/client/http_client.go
```diff
@@ -28,67 +28,60 @@ const (
...
+	// default to tcp if nothing specified
+	if u.Scheme == "" {
+		u.Scheme = protoTCP
+	}
+
+	return &parsedURL{*u}, nil
+}
+
+// Change protocol to HTTP for unknown protocols and TCP protocol (flexible HTTPS downgrade fix) - useful for RPC connections
+func (u *parsedURL) SetDefaultSchemeHTTP() {
```
melekes: why not do this in `newParsedURL`?

rpc/lib/client/http_client.go
```diff
@@ -28,67 +28,60 @@ const (
...
-	switch {
-	case len(parts) == 1:
-		// default to tcp if nothing specified
-		protocol, address = protoTCP, remoteAddr
-	case len(parts) == 2:
-		protocol, address = parts[0], parts[1]
-	default:
-		return "", "", fmt.Errorf("invalid addr: %s", remoteAddr)
-	}
+// Get a trimmed address (Implements previous 'toClientAddrAndParse') - useful for WS connections
```
melekes: > (Implements previous 'toClientAddrAndParse')

do we really need to mention funcs that were removed

rpc/lib/client/http_client.go
```diff
@@ -195,7 +206,15 @@ func (c *JSONRPCClient) Call(method string, params map[string]interface{}, resul
 		return nil, err
 	}
 	requestBuf := bytes.NewBuffer(requestBytes)
-	httpResponse, err := c.client.Post(c.address, "text/json", requestBuf)
+	httpRequest, err := http.NewRequest(http.MethodPost, c.address, requestBuf)
+	if err != nil {
+		return nil, err
+	}
+	httpRequest.Header.Set("Content-Type", "text/json")
+	if c.username != "" || c.password != "" {
```
```suggestion
	if c.username != "" && c.password != "" {
```

shouldn't we use &&?

rpc/lib/client/http_client.go
```diff
@@ -228,7 +247,15 @@ func (c *JSONRPCClient) sendBatch(requests []*jsonRPCBufferedRequest) ([]interfa
 	if err != nil {
 		return nil, err
 	}
-	httpResponse, err := c.client.Post(c.address, "text/json", bytes.NewBuffer(requestBytes))
+	httpRequest, err := http.NewRequest(http.MethodPost, c.address, bytes.NewBuffer(requestBytes))
+	if err != nil {
+		return nil, err
+	}
+	httpRequest.Header.Set("Content-Type", "text/json")
+	if c.username != "" || c.password != "" {
```
```suggestion
	if c.username != "" && c.password != "" {
```

shouldn't we use &&?

melekes:
Could you also open a PR with the same changes against master?

rpc/lib/client/http_client.go
greg-szabo: Good question, it makes the code even easier. The reason I kept it this way is because I wanted to keep compatibility with the superseded code which originally defaulted to TCP.

I'm not sure if there's any impact if we change this. I propose we change it in the PR for master (I think this will have to be reimplemented there because of other changes). For backward compatibility reasons, I'd leave it like this for this release.

But if you want I'll squash the feature into `newParsedURL`.

rpc/lib/client/http_client.go
greg-szabo: Not really, this is again a backward-compatibility thing. I actually added these here so if someone wants to check backward compatibility, they have an easier time going through the code.

I'll remove it.

rpc/lib/client/http_client.go
greg-szabo: I also contemplated that. Apparently, the description of how a URL should look like allows empty passwords - or empty users as well. With `&&` we would limit the code to accept URLs with both username and password filled. This way, we allow empty passwords.

rpc/lib/client/http_client.go
greg-szabo: Same as above: empty passwords or usernames should be allowed too.

rpc/lib/client/http_client.go
melekes: Thanks for explaining.

rpc/lib/client/http_client.go
melekes: > I'm not sure if there's any impact if we change this

not sure either lol :) let's not do this here, but do in a separate PR on master. How does that sound?

greg-szabo: Agreed

PR #4294, opened 2020-01-08
metrics: add metrics for specific validators

marbar3778:
* [ ] Wrote tests

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -1459,8 +1459,11 @@ func (cs *ConsensusState) finalizeCommit(height int64) {
 func (cs *ConsensusState) recordMetrics(height int64, block *types.Block) {
 	cs.metrics.Validators.Set(float64(cs.Validators.Size()))
 	cs.metrics.ValidatorsPower.Set(float64(cs.Validators.TotalVotingPower()))
-	missingValidators := 0
-	missingValidatorsPower := int64(0)
+
+	var (
+		missingValidators      = 0
+		missingValidatorsPower int64
+	)
 	for i, val := range cs.Validators.Validators {
 		var vote *types.CommitSig
 		if i < len(block.LastCommit.Precommits) {
@@ -1471,10 +1474,9 @@ func (cs *ConsensusState) recordMetrics(height int64, block *types.Block) {
 			missingValidatorsPower += val.VotingPower
 		}
 
-		_, privVal := cs.LastValidators.GetByAddress(cs.privValidator.GetPubKey().Address())
-		if privVal != nil && bytes.Equal(val.Address, privVal.Address) {
+		if cs.privValidator != nil && bytes.Equal(val.Address, cs.privValidator.GetPubKey().Address()) {
 			label := []string{
-				"validator_address", privVal.Address.String(),
+				"validator_address", val.Address.String(),
 			}
 			cs.metrics.ValidatorPower.With(label...).Set(float64(val.VotingPower))
 			if vote != nil {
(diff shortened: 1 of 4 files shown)


consensus/state.go
```diff
@@ -1470,7 +1470,23 @@ func (cs *ConsensusState) recordMetrics(height int64, block *types.Block) {
...
+			if cs.privValidator != nil && bytes.Equal(val.Address, privValAddress) {
+				label := []string{
+					"validator_address", privValAddress.String(),
+				}
+				cs.metrics.ValidatorPower.With(label...).Set(float64(val.VotingPower))
+				if vote != nil {
+					cs.metrics.ValidatorLastSignedHeight.With(label...).Set(float64(height))
+				} else {
+					cs.metrics.ValidatorMissedBlocks.With(label...).Add(float64(1))
+				}
```
```suggestion
				}
				break
```

melekes: should be done on master too

consensus/state.go
```diff
@@ -1470,7 +1470,23 @@ func (cs *ConsensusState) recordMetrics(height int64, block *types.Block) {
 			missingValidators++
 			missingValidatorsPower += val.VotingPower
 		}
+
+		for _, val := range cs.LastValidators.Validators {
+			privValAddress := cs.privValidator.GetPubKey().Address()
```
melekes: this will panic if cs.privValidator is nil!

melekes: we should use `cs.LastValidators.GetByAddress(cs.privValidator.GetPubKey().Address())` function to get validator address instead of `for` loop.

consensus/state.go
```diff
@@ -1470,7 +1470,22 @@ func (cs *ConsensusState) recordMetrics(height int64, block *types.Block) {
...
+			label := []string{
+				"validator_address", privVal.Address.String(),
+			}
+			cs.metrics.ValidatorPower.With(label...).Set(float64(val.VotingPower))
+			if vote != nil {
+				cs.metrics.ValidatorLastSignedHeight.With(label...).Set(float64(height))
+			} else {
+				cs.metrics.ValidatorMissedBlocks.With(label...).Add(float64(1))
+			}
+			break
```
```suggestion
			break
```

break only works with `for` loop

consensus/state.go
```diff
@@ -1470,7 +1470,22 @@ func (cs *ConsensusState) recordMetrics(height int64, block *types.Block) {
 			missingValidators++
 			missingValidatorsPower += val.VotingPower
 		}
+
+		_, privVal := cs.LastValidators.GetByAddress(cs.privValidator.GetPubKey().Address())
```
melekes: I believe we need to wrap the whole thing in `if cs.privValidator != nil { .... }`

consensus/state.go
```diff
@@ -1470,9 +1473,22 @@ func (cs *ConsensusState) recordMetrics(height int64, block *types.Block) {
...
+
+		if cs.privValidator != nil && bytes.Equal(val.Address, cs.privValidator.GetPubKey().Address()) {
+			label := []string{
+				"validator_address", val.Address.String(),
+			}
+			cs.metrics.ValidatorPower.With(label...).Set(float64(val.VotingPower))
+			if vote != nil {
+				cs.metrics.ValidatorLastSignedHeight.With(label...).Set(float64(height))
+			} else {
+				cs.metrics.ValidatorMissedBlocks.With(label...).Add(float64(1))
```
leoluk: I would set this to zero during initialization, otherwise, the metric only shows up after the first miss.

https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics

PR #4297, opened 2020-01-09
Release/v0.32.9

marbar3778:
back port of: 

- [rpc/lib] [\#4248](https://github.com/tendermint/tendermint/issues/4248) RPC client basic authentication support (@greg-szabo)

- [metrics] \#4263 Add
  - `consensus_validator_power`: track your validators power
  - `consensus_validator_last_signed_height`: track at which height the validator last signed
  - `consensus_validator_missed_blocks`: total amount of missed blocks for a validator
    as gauges in prometheus for validator specific metrics

- [rpc/lib] [\#4051](https://github.com/tendermint/tendermint/pull/4131) Fix RPC client, which was previously resolving https protocol to http (@yenkhoon)
- [cs] \#4069 Don't panic when block meta is not found in store (@gregzaitsev)

* [ ] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -16,6 +16,7 @@ program](https://hackerone.com/tendermint).
 - Go API
 
 ### FEATURES:
+- [rpc/lib] [\#4248](https://github.com/tendermint/tendermint/issues/4248) RPC client basic authentication support (@greg-szabo)
 
 ### IMPROVEMENTS:
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,4 +1,4 @@
-## v0.32.9
+## v0.32.10
 
 \*\*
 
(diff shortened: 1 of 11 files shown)


marbar3778:
https://github.com/tendermint/tendermint/pull/4294#discussion_r364768118 there is this comment that we can add @melekes thoughts?

CHANGELOG_PENDING.md
```diff
@@ -17,6 +17,17 @@ program](https://hackerone.com/tendermint).
 
 ### FEATURES:
 
+- [rpc/lib] [\#4248](https://github.com/tendermint/tendermint/issues/4248) RPC client basic authentication support (@greg-szabo)
```
melekes: move to `changelog.md`?

melekes: also don't see version update

marbar3778: https://github.com/tendermint/tendermint/pull/4298, sorry a bit brain dead right now

melekes:
> #4294 (comment) there is this comment that we can add @melekes thoughts?

should be easy to do. but it does not have to be in this release imo

Issue #4288, opened 2020-01-07
proto: Use linting,  breakage checking & gogoslick or gogofaster


marbar3778:
1. Use [buf](https://github.com/bufbuild/buf) https://buf.build/ to add linting and breaking change detection in ci.
   - ref #4078 
   - this would further deprecate GOPATH as we would use go mod vendor instead

2. move from `gogo` to `gogoslick` or `gogofaster` in the proto stub generation
   - protoc-gen-gogofaster (same as gogofast, without XXX_unrecognized, less pointer fields)
   - protoc-gen-gogoslick (same as gogofaster, but with generated string, gostring and equal methods)
the second proposal is aimed at starting to create types in proto. 

If we get rid of the XXX fields then they should be amino compatible and would not be breaking changes. 

@ebuchman @melekes thoughts on this?

melekes:
Strongly in favor 👍 However we may be relaying on some features of `gogo`, which might prevent us from switching to `gogoslick` / `gogofaster`.

marbar3778:
#4304 is ready to merge and I have a different branch with breaking changes. I was thinking of using one branch to merge all the breaking proto work into. @melekes thoughts?

melekes:
fine by me

marbar3778:
part 1 of this issue is completed, part 2 is ongoing

marbar3778:
In the last tendermint dev call we decided to move away from gogoproto and use vanilla go protobuf. This issue can be closed. Will open a new one to track the move to vanilla

PR #4304, opened 2020-01-13
proto: add linting & breakage check

marbar3778:
- adding in buf as a proto tool to help with linting and breakage checking

ref #4288

* [ ] Wrote tests


crypto/merkle/merkle.proto
```diff
@@ -4,7 +4,7 @@ option go_package = "github.com/tendermint/tendermint/crypto/merkle";
 
 // For more information on gogo.proto, see:
 // https://github.com/gogo/protobuf/blob/master/extensions.md
-import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+import "third_party/proto/gogoproto/gogo.proto";
```
marbar3778: this would be considered breaking as when users fetch these files they will also need to fetch third_party dir. Are we fine with this breakage in a minor release @melekes @tessr you can see how it changes things here: https://github.com/tendermint/rust-abci/pull/116

marbar3778:
breakage will always fail here since it is comparing it to master in which the imports are different than what are expected.

Makefile
```diff
@@ -1,7 +1,7 @@
 PACKAGES=$(shell go list ./...)
 OUTPUT?=build/tendermint
 
-INCLUDE = -I=${GOPATH}/src/github.com/tendermint/tendermint -I=${GOPATH}/src -I=${GOPATH}/src/github.com/gogo/protobuf/protobuf
+INCLUDE = -I=.
```
marbar3778: this can be removed

crypto/merkle/merkle.proto
melekes: it only breaks developer experience & our tutorials (java, kotlin) which rely on protobuf, correct?

third_party/proto/gogoproto/gogo.proto
```diff
@@ -0,0 +1,147 @@
+// Protocol Buffers for Go with Gadgets
+//
+// Copied from https://github.com/gogo/protobuf/blob/master/gogoproto/gogo.proto
+//
+// Copyright (c) 2013, The GoGo Authors. All rights reserved.
```
melekes: Why do we have to copy this file?

marbar3778: Since we are not using GOPATH anymore then it was required to use `vendor` dir, but if we were to go with vendor then we would need devs that are fetching these files to also fetch the vendor file, at least the gogoproto dep & the buf file would be something along the lines of
```
build:
  roots:
    - .
lint:
  ignore:
    - conformance
    - gogoproto
    - jsonpb
    - proto
    - protobuf
    - protoc-gen-gogo
    - test
    - vanity
breaking:
  ignore:
    - conformance
    - gogoproto
    - jsonpb
    - proto
    - protobuf
    - protoc-gen-gogo
    - test
    - vanity
``` 
or
```
build:
  roots:
    - .
  excludes:
    - vendor/github.com/gogo/protobuf/conformance
    - vendor/github.com/gogo/protobuf/jsonpb
    - vendor/github.com/gogo/protobuf/proto
    - vendor/github.com/gogo/protobuf/protobuf
    - vendor/github.com/gogo/protobuf/protoc-gen-gogo
    - vendor/github.com/gogo/protobuf/test
    - vendor/github.com/gogo/protobuf/vanity
lint:
  ignore:
    - gogoproto
breaking:
  ignore:
    - gogoproto
```
instead of what it is currently. This is because buf lists out the all the proto files located in the repo and with vendor this includes all the gogo proto ones.

crypto/merkle/merkle.proto
marbar3778: yes that is the only thing it changes. To merge this I will add a section to the upgrading.md file and see about updating the tutorials.

third_party/proto/gogoproto/gogo.proto
melekes: Okay, thanks!

Issue #4286, opened 2020-01-07
Crash on startup: error decoding 'consensus.create_empty_blocks_interval': time: invalid duration 


Ekliptor:
tendermint version0.32.8-7d7b47a3 compiled from github source

**ABCI app**:
Using the go Sample from here: https://docs.tendermint.com/master/guides/go.html

**Environment**:
Mac OS 10.15.2 (19C57)
go version go1.13.5 darwin/amd64
abci-cli version 0.16.1
go.mod:
```
module github.com/tendermint/tendermint

go 1.13

require (
	github.com/ChainSafe/go-schnorrkel v0.0.0-20191204112926-74e22bf42a85
	github.com/VividCortex/gohistogram v1.0.0 // indirect
	github.com/Workiva/go-datastructures v1.0.50
	github.com/btcsuite/btcd v0.0.0-20190115013929-ed77733ec07d
	github.com/btcsuite/btcutil v0.0.0-20180706230648-ab6388e0c60a
	github.com/fortytw2/leaktest v1.3.0
	github.com/go-kit/kit v0.9.0
	github.com/go-logfmt/logfmt v0.4.0
	github.com/gogo/protobuf v1.3.1
	github.com/golang/protobuf v1.3.2
	github.com/google/gofuzz v1.0.0 // indirect
	github.com/gorilla/websocket v1.4.1
	github.com/gtank/merlin v0.1.1-0.20191105220539-8318aed1a79f
	github.com/inconshreveable/mousetrap v1.0.0 // indirect
	github.com/libp2p/go-buffer-pool v0.0.2
	github.com/magiconair/properties v1.8.1
	github.com/mitchellh/gox v1.0.1 // indirect
	github.com/pkg/errors v0.8.1
	github.com/prometheus/client_golang v0.9.3
	github.com/rcrowley/go-metrics v0.0.0-20180503174638-e2704e165165
	github.com/rs/cors v1.7.0
	github.com/snikch/goodman v0.0.0-20171125024755-10e37e294daa
	github.com/spf13/cobra v0.0.1
	github.com/spf13/viper v1.6.1
	github.com/square/certstrap v1.2.0 // indirect
	github.com/stretchr/testify v1.4.0
	github.com/stumble/gorocksdb v0.0.3 // indirect
	github.com/tendermint/go-amino v0.14.1
	github.com/tendermint/tm-db v0.3.0
	golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2
[...]

PR #4307, opened 2020-01-14
docs: add link for installing Tendermint

melekes:
to Go guides

Refs #4286

* [ ] Wrote tests


docs/guides/go-built-in.md
```diff
@@ -40,7 +40,7 @@ Verify that you have the latest version of Go installed:
 
 ```sh
 $ go version
-go version go1.13 darwin/amd64
+go version go1.13.1 darwin/amd64
```
marbar3778:
```suggestion
go version go1.13.* darwin/amd64
```

since we support all versions of 1.13, just so the user knows they can ues any version?

melekes: But it's the output of go version command. Let's maybe add a sentence like `Tendermint requires Go versions >= 1.12`

melekes: `Verify that you have the latest version of Go installed` should be enough imo

Issue #4316, opened 2020-01-17
Commit#ValidateBasic should check for duplicate votes


melekes:
It's incorrect for a commit https://godoc.org/github.com/tendermint/tendermint/types#Commit to contain duplicate votes for the same value.

After we do that, https://github.com/tendermint/tendermint/blob/5aa209a15617b91f19e18bf286e4cfccc68a229b/types/validator_set.go#L776-L779 can be removed.

melekes:
https://github.com/tendermint/tendermint/pull/4321#issuecomment-576150073

PR #4321, opened 2020-01-17
check for duplicate signatures in Commit#ValidateBasic

cmwaters:
Addresses #4316 

Checks for duplicate votes in a commit by checking for matching signatures (not matching validator addresses) 

* [ ] Wrote tests


melekes:
I thought about it a little more and now think it's unnecessary overhead for `VerifyCommit`. As for `VerifyCommitTrusting`, we check for duplicate validators there. I am going to close this PR & issue. Sorry for this.

PR #4322, opened 2020-01-18
tests: bind test servers to 127.0.0.1

erikgrinaker:
* [x] Wrote tests
* [x] ~Referenced an issue explaining the need for the change~

The test servers bind to public interfaces by default, triggering about a dozen macOS firewall dialogs that have to be rejected each time tests are run. Binding to the loopback interface instead prevents these warnings.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -19,8 +19,6 @@ program](https://hackerone.com/tendermint).
 
 ### IMPROVEMENTS:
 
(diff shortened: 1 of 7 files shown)


CHANGELOG_PENDING.md
```diff
@@ -19,6 +19,8 @@ program](https://hackerone.com/tendermint).
 
 ### IMPROVEMENTS:
 
+- [tests] [#\4322](https://github.com/tendermint/tendermint/pull/4322) Bind test servers to 127.0.0.1 (@erikgrinaker)
```
melekes: no need to add a changelog entry for every change. if it becomes too big, no one will read it.

PR #4342, opened 2020-01-27
rpc: add sort_order option to tx_search

melekes:
I have added order_by which can be "asc" or "desc" (should be in string format) in the tx_search RPC method.

Fixes: #3333

Author: @princesinha19 

* [x] Wrote tests


rpc/core/tx.go
```diff
@@ -100,5 +102,24 @@ func TxSearch(ctx *rpctypes.Context, query string, prove bool, page, perPage int
...
+	if len(apiResults) > 1 {
+		switch orderBy {
+		case "desc":
+			sort.Slice(apiResults, func(i, j int) bool {
+				if apiResults[i].Height == apiResults[j].Height {
+					return apiResults[i].Index > apiResults[j].Index
+				}
+				return apiResults[i].Height > apiResults[j].Height
+			})
+		default: // asc
```
erikgrinaker: I think this should explicitly handle `"asc"` and `""`, otherwise return error. As it is, I believe e.g. `"DESC"` would silently be interpreted as `"asc"´.

melekes:
> Is the idea to eventually support arbitrary expressions here, so for example "height asc, time desc"?

Eventually, yes.

> If so, it might be better to take an array of field/direction pairs, to avoid having to do any expression parsing and stuff.

Parsing is not that hard :)

rpc/core/tx.go
melekes: Great catch!

rpc/core/tx.go
```diff
@@ -111,13 +112,17 @@ func TxSearch(ctx *rpctypes.Context, query string, prove bool, page, perPage int
 				}
 				return apiResults[i].Height > apiResults[j].Height
 			})
-		default: // asc
+		case "": // defaults to asc
+			fallthrough
```
erikgrinaker: Nit, but why not just `case "asc", "":`?

state/txindex/kv/kv.go
```diff
@@ -160,12 +159,14 @@ func (txi *TxIndex) indexEvents(result *types.TxResult, hash []byte, store dbm.S
 	}
 }
 
-// Search performs a search using the given query. It breaks the query into
-// conditions (like "tx.height > 5"). For each condition, it queries the DB
-// index. One special use cases here: (1) if "tx.hash" is found, it returns tx
-// result for it (2) for range queries it is better for the client to provide
-// both lower and upper bounds, so we are not performing a full scan. Results
-// from querying indexes are then intersected and returned to the caller.
+// Search performs a search using the given query.
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
// Search performs a search using the given query.
```

PR #4343, opened 2020-01-27
lite2: refactor cleanup()

cmwaters:
Refactored lite2/client.go#cleanup to handle cases when height is greater than 0 but less than the last trusted height so that the lite client doesn't attempt to delete headers and validator sets it doesn't have. This is done by setting the inputted `stopHeight` to the height of the first header if the it receives a `stopHeight` that is less than the `FirstSignedHeaderHeight`.

In addition if no header or validator set is left in memory, the function will set `trustedHeader` and `trustedNextVals` to `nil`, else it restores the values to the latest trusted ones it can find by running `restoreTrustedHeaderAndNextVals()`.

* [ ] Wrote tests


tessr:
It looks like you might've picked up some of @melekes's commits here too--do you need to rebase? Or maybe they're meant to be included and I misunderstood the PR message!

cmwaters:
No you're right, I branched of his branch which he's now merged with master so I believe I will need to rebase

lite2/client.go
```diff
@@ -234,6 +234,10 @@ func (c *Client) restoreTrustedHeaderAndNextVals() error {
 		c.trustedNextVals = trustedNextVals
 
 		c.logger.Debug("Restored trusted header and next vals", lastHeight)
+	} else {
+		c.trustedHeader = nil
```
melekes: I don't think this is the right place for nullifying the header & vals. The name of this func implies restoration, not setting trustedHeader to nil.

lite2/client.go
```diff
@@ -552,7 +556,7 @@ func (c *Client) Cleanup() error {
 	return c.cleanup(0)
 }
 
-// stopHeight=0 -> remove all data
+// cleanup removes data from
```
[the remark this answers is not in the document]
melekes: how about

```suggestion
cleanup deletes all headers & validator sets between +stopHeight+ and latest height included
```

lite2/client.go
```diff
@@ -578,8 +582,10 @@ func (c *Client) cleanup(stopHeight int64) error {
 		}
 	}
 
-	c.trustedHeader = nil
```
melekes: let's keep this code and call c.restoreTrustedHeaderAndNextVals right after

PR #4344, opened 2020-01-28
lite2: rename alternative providers to witnesses

### LINKED ISSUE #4341: lite2: rename Alternatives to Witnesses
The reason is alternative provider may be viewed as something not important. But in reality we are using alternative providers for cross-checking headers and they are super important.

Other solution may be requiring at least 5 (don't know what the good number is) alternative providers.

melekes:
Closes #4341

* [x] Wrote tests

### DIFF
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -84,11 +84,12 @@ func SkippingVerification(trustLevel tmmath.Fraction) Option {
 	}
 }
 
-// AlternativeSources option can be used to supply alternative providers, which
-// will be used for cross-checking the primary provider.
-func AlternativeSources(providers []provider.Provider) Option {
+// Witnesses option can be used to supply providers, which will be used for
+// cross-checking the primary provider. A witness can become a primary iff the
+// current primary is unavailable.
+func Witnesses(providers []provider.Provider) Option {
 	return func(c *Client) {
-		c.alternatives = providers
+		c.witnesses = providers
 	}
 }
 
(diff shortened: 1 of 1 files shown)


lite2/client.go
```diff
@@ -84,11 +84,12 @@ func SkippingVerification(trustLevel tmmath.Fraction) Option {
 	}
 }
 
-// AlternativeSources option can be used to supply alternative providers, which
-// will be used for cross-checking the primary provider.
-func AlternativeSources(providers []provider.Provider) Option {
+// Witnesses option can be used to supply providers, which will be used for
+// cross-checking the primary provider. A witness can become a primary iff the
```
cmwaters: *picky:* A witness can become a primary iff -> A witness can become a primary if

lite2/client.go
```diff
@@ -84,11 +84,12 @@ func SkippingVerification(trustLevel tmmath.Fraction) Option {
...
+// current primary is unavailable.
+func Witnesses(providers []provider.Provider) Option {
 	return func(c *Client) {
-		c.alternatives = providers
```
cmwaters: Just as a thought. Do we want to check if these are on the same chain - i.e. `c.chainID == provider.ChainID()`. Also could be a way of initially checking if the provider is responding. Could be unnecessary thought.

lite2/client.go
melekes: iff - if and only if

cmwaters: ah I see. Thanks :)

lite2/client.go
melekes: Good idea! We should do this upon startup. Want to do it? https://github.com/tendermint/tendermint/issues/4350

PR #4345, opened 2020-01-28
lite2: batch save & delete operations in DB store

### LINKED ISSUE #4330: lite2: batch save & delete operations in DB store
https://github.com/tendermint/tendermint/blob/5f2f97548f8714e05f5dae877ea2972d3e074187/lite2/store/db/db.go#L40-L51

https://github.com/tendermint/tendermint/blob/5f2f97548f8714e05f5dae877ea2972d3e074187/lite2/store/db/db.go#L63-L65

melekes:
Closes #4330

* [ ] ~~Wrote tests~~ already covered

### DIFF
--- a/lite2/store/db/db.go
+++ b/lite2/store/db/db.go
@@ -1,11 +1,11 @@
 package db
 
 import (
-	"errors"
 	"fmt"
 	"regexp"
 	"strconv"
 
+	"github.com/pkg/errors"
 	"github.com/tendermint/go-amino"
 	dbm "github.com/tendermint/tm-db"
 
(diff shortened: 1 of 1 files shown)


lite2/store/db/db.go
```diff
@@ -38,20 +38,21 @@ func (s *dbs) SaveSignedHeaderAndNextValidatorSet(sh *types.SignedHeader, valSet
...
 	}
-	s.db.Set(s.vsKey(sh.Height+1), bz)
 
-	return nil
+	b := s.db.NewBatch()
+	b.Set(s.shKey(sh.Height), shBz)
+	b.Set(s.vsKey(sh.Height+1), valSetBz)
+	err = b.WriteSync()
+	b.Close()
+	return err
```
marbar3778: nit: wrap error so IF this fails we can pinpoint it a bit easier. if this were to fail there would be a larger problem at hand I believe.

PR #4352, opened 2020-01-28
ADR-053: state sync prototype

erikgrinaker:
Draft ADR describing a prototype for state sync (see #828).

* [x] ~Wrote tests~

### DIFF
--- a/docs/architecture/adr-053-state-sync-prototype.md
+++ b/docs/architecture/adr-053-state-sync-prototype.md
@@ -1,6 +1,6 @@
 # ADR 053: State Sync Prototype
 
-This ADR describes an initial state sync prototype, and is subject to change as we gain feedback and experience.
+This ADR describes an initial state sync prototype, and is subject to change as we gain feedback and experience. It builds on discussions and findings in [ADR-042](./adr-042-state-sync.md), see there for background information.
 
 ## Changelog
 
@@ -28,23 +28,23 @@ This describes the snapshot/restore process seen from Tendermint. The interface 
 
 ### Snapshot Data Structure
 
+A node can have multiple snapshots taken at various heights. Snapshots can be taken in different application-specified formats (e.g. JSON as format `1` and Protobuf as format `2`, or similarly with schema versioning). Each snapshot consists of multiple chunks containing the actual state data, allowing parallel downloads and reduced memory usage during restores.
+
 ```proto
 message Snapshot {
     uint64 height = 1;   // The height at which the snapshot was taken
-    uint64 format = 2;   // The application-specific snapshot format version
+    uint64 format = 2;   // The application-specific snapshot format
     uint64 chunks = 3;   // The number of chunks in the snapshot
     bytes metadata = 4;  // Arbitrary application metadata
 }
 
 message SnapshotChunk {
     uint64 height = 1;   // The height of the corresponding snapshot
-    uint64 format = 2;   // The snapshot format version
+    uint64 format = 2;   // The application-specific snapshot format
     uint64 chunk = 3;    // The chunk index (zero-based)
     bytes data = 4;      // Serialized application state in an arbitrary format
 }
 ```
-
-A node can have multiple snapshots taken at various heights, in different application-specific format versions. Each snapshot consists of multiple chunks containing the actual state data.
 
 Chunk verification data must be encoded along with the state data in the `data` field.
 
--- a/docs/architecture/adr-053-state-sync-prototype.md
+++ b/docs/architecture/adr-053-state-sync-prototype.md
@@ -82,7 +82,6 @@ message ResponseGetSnapshotChunk {
 message RequestApplySnapshotChunk {
     SnapshotChunk chunk = 1;
     bytes chain_hash = 2;
-    bool final = 3;
 }
 
 message ResponseApplySnapshotChunk {}
@@ -126,7 +125,7 @@ When starting an empty node with state sync and fast sync enabled, snapshots are
 
 6. The node downloads chunks in parallel from multiple peers via `RequestGetSnapshotChunk`.
 
-7. The node passes chunks sequentially to the app via `RequestApplySnapshotChunk`, along with the root chain hash at the snapshot height for verification. If the chunk cannot be verified or applied, the application returns `ResponseException` and Tendermint tries refetching the chunk. The final chunk has `final: true`.
+7. The node passes chunks sequentially to the app via `RequestApplySnapshotChunk`, along with the root chain hash at the snapshot height for verification. If the chunk cannot be verified or applied, the application returns `ResponseException` and Tendermint tries refetching the chunk.
 
 8. Once all chunks have been applied, the node compares the app hash to the chain hash, and if they do not match it either errors or discards the state and starts over.
 
--- a/docs/architecture/adr-053-state-sync-prototype.md
+++ b/docs/architecture/adr-053-state-sync-prototype.md
@@ -31,7 +31,7 @@ This describes the snapshot/restore process seen from Tendermint. The interface 
 ```proto
 message Snapshot {
     uint64 height = 1;   // The height at which the snapshot was taken
-    uint64 format = 2;   // The application-specific snapshot format
+    uint64 format = 2;   // The application-specific snapshot format version
     uint64 chunks = 3;   // The number of chunks in the snapshot
     bytes metadata = 4;  // Arbitrary application metadata
 }
@@ -44,7 +44,7 @@ message SnapshotChunk {
 }
 ```
 
-A node can have multiple snapshots taken at various heights, in different application-specific formats (e.g. for format versioning). Each snapshot consists of multiple chunks containing the actual state data.
+A node can have multiple snapshots taken at various heights, in different application-specific format versions. Each snapshot consists of multiple chunks containing the actual state data.
 
 Chunk verification data must be encoded along with the state data in the `data` field.
 
@@ -92,7 +92,7 @@ message ResponseApplySnapshotChunk {}
 
 Tendermint is not aware of the snapshotting process at all, it is entirely an application concern. The following guarantees must be provided:
 
-* **Periodic:** snapshots must be taken periodically, not on-demand - for faster restores, lower load, and less DoS risk.
+* **Periodic:** snapshots must be taken periodically, not on-demand, for faster restores, lower load, and less DoS risk.
 
 * **Deterministic:** snapshots must be deterministic, and identical across all nodes - typically by taking a snapshot at given height intervals.
 
@@ -106,7 +106,7 @@ Tendermint is not aware of the snapshotting process at all, it is entirely an ap
 
 ### Restoring Snapshots
 
-Nodes should have options for enabling state sync and/or fast sync, and a way to provide a trusted header hash for the light client.
+Nodes should have options for enabling state sync and/or fast sync, and be provided a trusted header hash for the light client.
 
 When starting an empty node with state sync and fast sync enabled, snapshots are restored as follows:
 
@@ -118,13 +118,13 @@ When starting an empty node with state sync and fast sync enabled, snapshots are
 
 4. The node requests available snapshots via `RequestListSnapshots`.
 
-5. The node iterates over all snapshots in reverse order by height and format until it finds one that satisfies the following conditions (otherwise error out or switch to fast sync):
+5. The node iterates over all snapshots in reverse order by height and format until it finds one that satisfies the following conditions:
 
-  * The snapshot height's block is considered trustworthy by the light client (i.e. snapshot height is greater than trusted header and within unbonding period of the latest trustworthy block).
+    * The snapshot height's block is considered trustworthy by the light client (i.e. snapshot height is greater than trusted header and within unbonding period of the latest trustworthy block).
 
-  * The application accepts the `RequestOfferSnapshot` call.
+    * The application accepts the `RequestOfferSnapshot` call.
 
-6. The node downloads chunks in parallel from multiple peers, via `RequestGetSnapshotChunk`.
+6. The node downloads chunks in parallel from multiple peers via `RequestGetSnapshotChunk`.
 
 7. The node passes chunks sequentially to the app via `RequestApplySnapshotChunk`, along with the root chain hash at the snapshot height for verification. If the chunk cannot be verified or applied, the application returns `ResponseException` and Tendermint tries refetching the chunk. The final chunk has `final: true`.
 
@@ -156,14 +156,10 @@ A chunk corresponds to a subtree key range within an IAVL tree, in insertion ord
 
 ```go
 struct SnapshotChunk {
-    // Store is the name (key) of the IAVL store in the surrounding MultiStore
-    Store   string
-    // Keys contains snapshotted keys in insertion order
-    Keys    [][]byte
-    // Values contains snapshotted values corresponding to Keys
-    Values  [][]byte
-    // Proof is the Merkle proof from the subtree root up to the MultiStore root
-    Proof   []merkle.ProofOp
+    Store   string            // Name (key) of IAVL store in outer MultiStore
+    Keys    [][]byte          // Snapshotted keys in insertion order
+    Values  [][]byte          // Snapshotted values corresponding to Keys
+    Proof   []merkle.ProofOp  // Merkle proof from subtree root to MultiStore root
 }
 ```
 
@@ -171,13 +167,13 @@ This chunk structure is believed to be sufficient to reconstruct an identical IA
 
 We do not use IAVL RangeProofs, since these include redundant data such as proofs for intermediate and leaf nodes that can be derived from the above data.
 
-Chunks should be built greedily by collecting key/value pairs constituting a complete subtree in order up until some size limit, e.g. 16 MB, then serialized and stored in the file system as `snapshots/<height>/<format>/<chunk>/data` and served via `RequestGetSnapshotChunk`.
+Chunks should be built greedily by collecting key/value pairs constituting a complete subtree up to some size limit (e.g. 16 MB), then serialized and stored in the file system as `snapshots/<height>/<format>/<chunk>/data` and served via `RequestGetSnapshotChunk`.
 
 ### Snapshot Scheduling
 
 Snapshots should be taken at some configurable height interval, e.g. every 1000 blocks. All nodes should preferably have the same snapshot schedule, such that all nodes can serve chunks for a given snapshot.
 
-Taking consistent snapshots of IAVL trees is greatly simplified by IAVL trees being versioned: we simply snapshot the version that corresponds to the snapshot height, while concurrent writes can continue creating new versions. IAVL pruning must make sure it does not prune a version that is being snapshotted.
+Taking consistent snapshots of IAVL trees is greatly simplified by them being versioned: simply snapshot the version that corresponds to the snapshot height, while concurrent writes create new versions. IAVL pruning must not prune a version that is being snapshotted.
 
 Snapshots must also be garbage collected after some configurable time, e.g. by keeping the latest `n` snapshots.
 
--- a/docs/architecture/adr-053-state-sync-prototype.md
+++ b/docs/architecture/adr-053-state-sync-prototype.md
@@ -189,8 +189,6 @@ Snapshots must also be garbage collected after some configurable time, e.g. by k
 
 * Is it OK for state-synced nodes to not have historical blocks nor historical IAVL versions?
 
-* Do we need to reject chunk sizes beyond a configurable limit, or will this be handled implicitly by e.g. gRPC request size limits?
-
 ## Status
 
 Proposed
--- a/docs/architecture/adr-053-state-sync-prototype.md
+++ b/docs/architecture/adr-053-state-sync-prototype.md
@@ -43,6 +43,7 @@ message SnapshotChunk {
     uint16 format = 2;   // The application-specific snapshot format
     uint64 chunk = 3;    // The chunk index (zero-based)
     bytes data = 4;      // Serialized application state in an arbitrary format
+    bytes checksum = 5;  // SHA-1 checksum of data
 }
 ```
 
@@ -123,7 +124,7 @@ When starting an empty node with state sync and fast sync enabled, snapshots are
 
     * The application accepts the `RequestOfferSnapshot` call.
 
-6. The node downloads chunks in parallel from multiple peers via `RequestGetSnapshotChunk`.
+6. The node downloads chunks in parallel from multiple peers via `RequestGetSnapshotChunk`, and both the sender and receiver verifies their checksums.
 
 7. The node passes chunks sequentially to the app via `RequestApplySnapshotChunk`, along with the root chain hash at the snapshot height for verification. If the chunk cannot be verified or applied, the application returns `ResponseException` and Tendermint tries refetching the chunk.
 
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,196 @@
+# ADR 053: State Sync Prototype
+
+This ADR describes an initial state sync prototype, and is subject to change as we gain feedback and experience.
```
tessr: You might mention right off the bat how this is different from ADR 42.

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,196 @@
...
+message Snapshot {
+    uint64 height = 1;   // The height at which the snapshot was taken
+    uint64 format = 2;   // The application-specific snapshot format version
+    uint64 chunks = 3;   // The number of chunks in the snapshot
+    bytes metadata = 4;  // Arbitrary application metadata
+}
+
+message SnapshotChunk {
+    uint64 height = 1;   // The height of the corresponding snapshot
+    uint64 format = 2;   // The snapshot format version
```
tessr: This describes the version of the application? Or something else?

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,196 @@
...
+
+A node can have multiple snapshots taken at various heights, in different application-specific format versions. Each snapshot consists of multiple chunks containing the actual state data.
+
+Chunk verification data must be encoded along with the state data in the `data` field.
+
+### ABCI Interface
+
+```proto
+// Lists available snapshots
+message RequestListSnapshots {}
```
tessr: This question reveals some naiveté about ABCI, but do we need empty protos for requests with no bodies, or can we just submit the request along the appropriate RPC?

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,196 @@
...
+
+This ADR describes an initial state sync prototype, and is subject to change as we gain feedback and experience.
+
+## Changelog
+
+* 2020-01-28: Initial draft (Erik Grinaker)
+
+## Context
+
+State sync will allow a new node to receive a snapshot of the application state without downloading blocks or going through consensus. This bootstraps the node significantly faster than the current fast sync system, which replays all historical blocks.
```
fedekunze: > State sync will allow a new node to receive a snapshot of the application state without downloading blocks or going through consensus.

I assume this is how it's done now, right? It'd be cool to outline this and the current problems in a separate paragraph.

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,196 @@
...
+
+```proto
+message Snapshot {
+    uint64 height = 1;   // The height at which the snapshot was taken
+    uint64 format = 2;   // The application-specific snapshot format version
+    uint64 chunks = 3;   // The number of chunks in the snapshot
+    bytes metadata = 4;  // Arbitrary application metadata
+}
+
+message SnapshotChunk {
```
fedekunze: can you explain what a chunk is in the context of a snapshot?

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,196 @@
...
+    uint64 height = 1;   // The height of the corresponding snapshot
+    uint64 format = 2;   // The snapshot format version
+    uint64 chunk = 3;    // The chunk index (zero-based)
+    bytes data = 4;      // Serialized application state in an arbitrary format
+}
+```
+
+A node can have multiple snapshots taken at various heights, in different application-specific format versions. Each snapshot consists of multiple chunks containing the actual state data.
```
fedekunze: Ohh I see. Maybe move this paragraph between the section subtitle and the proto definitions

docs/architecture/adr-053-state-sync-prototype.md
fedekunze: EDIT: see comment below

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: The snapshot format version. Let's say we serialize chunks as Amino. Then in SDK version Z we change to Protobuf. Since nodes do not necessarily upgrade at the same time, now some nodes will host Amino snapshots, while others will host Protobuf snapshots. We need a way for the application to distinguish between them, for example by calling Amino format `1` and Protobuf format `2`.

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,196 @@
...
+
+7. The node passes chunks sequentially to the app via `RequestApplySnapshotChunk`, along with the root chain hash at the snapshot height for verification. If the chunk cannot be verified or applied, the application returns `ResponseException` and Tendermint tries refetching the chunk. The final chunk has `final: true`.
+
+8. Once all chunks have been applied, the node compares the app hash to the chain hash, and if they do not match it either errors or discards the state and starts over.
+
+9. The node switches to fast sync to catch up blocks that were committed while restoring the snapshot.
+
+10. The node switches to normal consensus mode.
+
+## Gaia Proposal
```
marbar3778: this part of the proposal may be better suited to live in the Cosmos-SDK.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: This is outlined in ADR-042, which has a ton of background material on why this solution is chosen. This is mentioned below, but I'll add a line on it at the beginning.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: Done

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: Good idea, done.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: I dropped "version", and just called it "format", since it's not necessarily only used for versioning. However, Tendermint will consider higher format numbers to be "better", an attempt to apply those first.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: We could just have a common `RequestEmpty` message, but the current Protobufs seem to have separate empty messages for various requests. I'm fine with either, as long as we're being consistent.

marbar3778: to not break things I would lean towards following the standard we have now.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: I don't think we need to overcomplicate things for an initial _prototype_, but once we converge on a final approach the final specification should definitely be in the Cosmos SDK repos.

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,196 @@
...
+
+Tendermint is not aware of the snapshotting process at all, it is entirely an application concern. The following guarantees must be provided:
+
+* **Periodic:** snapshots must be taken periodically, not on-demand, for faster restores, lower load, and less DoS risk.
+
+* **Deterministic:** snapshots must be deterministic, and identical across all nodes - typically by taking a snapshot at given height intervals.
+
+* **Consistent:** snapshots must be consistent, i.e. not affected by concurrent writes - typically by using a data store that supports versioning and/or snapshot isolation.
+
+* **Asynchronous:** snapshots must be asynchronous, i.e. not halt block processing and state transitions.
```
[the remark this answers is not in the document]
erikgrinaker: Technically they are _should_, not _must_ - there is no way for us to enforce these, and for some applications that might be fine. But in Friday's meeting it was argued that we should use _must_ in order to strongly recommend these, in the same way that we say "state machines must be deterministic".

docs/architecture/adr-053-state-sync-prototype.md
marbar3778: ah okay, I misunderstood

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,196 @@
...
+
+Nodes should have options for enabling state sync and/or fast sync, and be provided a trusted header hash for the light client.
+
+When starting an empty node with state sync and fast sync enabled, snapshots are restored as follows:
+
+1. The node checks that it is empty, i.e. that it has no state nor blocks.
+
+2. The node contacts the given seeds to discover peers.
+
+3. The node contacts a set of full nodes, and verifies the trusted block header using the given hash via the light client.
```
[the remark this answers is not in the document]
erikgrinaker: That is deliberately underspecified, we'll need further testing while prototyping to determine the optimal number.

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,196 @@
...
+
+    * The snapshot height's block is considered trustworthy by the light client (i.e. snapshot height is greater than trusted header and within unbonding period of the latest trustworthy block).
+
+    * The application accepts the `RequestOfferSnapshot` call.
+
+6. The node downloads chunks in parallel from multiple peers via `RequestGetSnapshotChunk`.
+
+7. The node passes chunks sequentially to the app via `RequestApplySnapshotChunk`, along with the root chain hash at the snapshot height for verification. If the chunk cannot be verified or applied, the application returns `ResponseException` and Tendermint tries refetching the chunk. The final chunk has `final: true`.
+
+8. Once all chunks have been applied, the node compares the app hash to the chain hash, and if they do not match it either errors or discards the state and starts over.
```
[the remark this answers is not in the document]
erikgrinaker: Good point, I'll add an open question.

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Gaia Proposal
+
+This describes the snapshot process seen from Gaia, using format version `1`. The serialization format is unspecified, but likely to be compressed Amino or Protobuf.
+
+### Snapshot Metadata
+
+In the initial version there is no snapshot metadata, so it is set to an empty byte buffer.
+
+Once all chunks have been successfully built, snapshot metadata should be serialized and stored in the file system as e.g. `snapshots/<height>/<format>/metadata`, and served via `RequestListSnapshots`.
```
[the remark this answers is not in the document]
erikgrinaker: Correct.

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,196 @@
...
+
+1. The node checks that it is empty, i.e. that it has no state nor blocks.
+
+2. The node contacts the given seeds to discover peers.
+
+3. The node contacts a set of full nodes, and verifies the trusted block header using the given hash via the light client.
+
+4. The node requests available snapshots via `RequestListSnapshots`.
+
+5. The node iterates over all snapshots in reverse order by height and format until it finds one that satisfies the following conditions:
```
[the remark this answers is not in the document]
erikgrinaker: Sure, done

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,198 @@
...
+
+message ResponseGetSnapshotChunk {
+    SnapshotChunk chunk = 1;
+}
+
+// Applies a snapshot chunk
+message RequestApplySnapshotChunk {
+    SnapshotChunk chunk = 1;
+    bytes chain_hash = 2;
+    bool final = 3;
```
melekes: why do we need final field if we know the number of chunks?

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,198 @@
...
+
+5. The node iterates over all snapshots in reverse order by height and format until it finds one that satisfies both of the following conditions:
+
+    * The snapshot height's block is considered trustworthy by the light client (i.e. snapshot height is greater than trusted header and within unbonding period of the latest trustworthy block).
+
+    * The application accepts the `RequestOfferSnapshot` call.
+
+6. The node downloads chunks in parallel from multiple peers via `RequestGetSnapshotChunk`.
+
+7. The node passes chunks sequentially to the app via `RequestApplySnapshotChunk`, along with the root chain hash at the snapshot height for verification. If the chunk cannot be verified or applied, the application returns `ResponseException` and Tendermint tries refetching the chunk. The final chunk has `final: true`.
```
melekes: Could you explain why chunks have to be applied sequentially? Because iAVL tree is order dependent?

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,198 @@
...
+
+8. Once all chunks have been applied, the node compares the app hash to the chain hash, and if they do not match it either errors or discards the state and starts over.
```
melekes: Wait. I thought chunks are immediately verifiable, no?

docs/architecture/adr-053-state-sync-prototype.md
melekes: What is the root chain hash?

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,198 @@
...
+    Values  [][]byte          // Snapshotted values corresponding to Keys
+    Proof   []merkle.ProofOp  // Merkle proof from subtree root to MultiStore root
+}
+```
+
+This chunk structure is believed to be sufficient to reconstruct an identical IAVL tree by applying separate chunks, but this may require the chunks to be ordered in a certain way; further research is needed.
+
+We do not use IAVL RangeProofs, since these include redundant data such as proofs for intermediate and leaf nodes that can be derived from the above data.
+
+Chunks should be built greedily by collecting key/value pairs constituting a complete subtree up to some size limit (e.g. 16 MB), then serialized and stored in the file system as `snapshots/<height>/<format>/<chunk>/data` and served via `RequestGetSnapshotChunk`.
```
melekes: I would encourage you to checksum all the data you're going to store. E.g.:

```go
struct SnapshotChunk {
    ...
    Checksum [md5.Size]byte
}

This will allow you to detect broken snapshots before you send them to a peer requesting data.

File: docs/architecture/adr-053-state-sync-prototype.md [hunk 17]
```diff
@@ -0,0 +1,198 @@
...
+
+Taking consistent snapshots of IAVL trees is greatly simplified by them being versioned: simply snapshot the version that corresponds to the snapshot height, while concurrent writes create new versions. IAVL pruning must not prune a version that is being snapshotted.
+
+Snapshots must also be garbage collected after some configurable time, e.g. by keeping the latest `n` snapshots.
+
+## Open Questions
+
+* Is it possible to reconstruct an identical IAVL tree given separate subtrees in an appropriate order, or is more data needed about the branch structure?
+
+* Should we punish nodes that provide invalid snapshots?
```
melekes: What do you mean by punishing? Disconnecting from them should be enough for now.

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// Applies a snapshot chunk
+message RequestApplySnapshotChunk {
+    SnapshotChunk chunk = 1;
+    bytes chain_hash = 2;
+    bool final = 3;
+}
+
+message ResponseApplySnapshotChunk {}
```
melekes: shouldn't it contain `ok bool = 1;` field to indicate that snapshot chunk is accepted?

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,198 @@
...
+
+Snapshots must also be garbage collected after some configurable time, e.g. by keeping the latest `n` snapshots.
+
+## Open Questions
+
+* Is it possible to reconstruct an identical IAVL tree given separate subtrees in an appropriate order, or is more data needed about the branch structure?
+
+* Should we punish nodes that provide invalid snapshots?
+
+* Should `ResponseOfferSnapshot` and `ResponseApplySnapshotChunk` return specific codes for e.g. invalid heights, unknown formats, and so on?
```
melekes: probably not if we assume logging (ability to figure out why certain things are not happening) on the app side

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,198 @@
...
+
+* Is it OK for state-synced nodes to not have historical blocks nor historical IAVL versions?
+
+* Do we need to reject chunk sizes beyond a configurable limit, or will this be handled implicitly by e.g. gRPC request size limits?
```
melekes: Explicit limit is better than implicit

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,198 @@
...
+This describes the snapshot/restore process seen from Tendermint. The interface is kept as small and general as possible to give applications maximum flexibility.
+
+### Snapshot Data Structure
+
+A node can have multiple snapshots taken at various heights. Snapshots can be taken in different application-specified formats (e.g. MessagePack as format `1` and Protobuf as format `2`, or similarly for schema versioning). Each snapshot consists of multiple chunks containing the actual state data, allowing parallel downloads and reduced memory usage.
+
+```proto
+message Snapshot {
+    uint64 height = 1;   // The height at which the snapshot was taken
+    uint64 format = 2;   // The application-specific snapshot format
```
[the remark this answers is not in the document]
erikgrinaker: True. Let's say uint16 though - 256 is most probably enough for most, but no point in placing arbitrary restrictions on it to save a byte.

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -0,0 +1,198 @@
...
+    uint64 chunks = 3;   // The number of chunks in the snapshot
+    bytes metadata = 4;  // Arbitrary application metadata
+}
+
+message SnapshotChunk {
+    uint64 height = 1;   // The height of the corresponding snapshot
+    uint64 format = 2;   // The application-specific snapshot format
```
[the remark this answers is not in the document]
erikgrinaker: Snapshots are uniquely identified by height and format (since there can be multiple formats at a given height, e.g. if two nodes are running different versions), so the chunks need both as well to determine which snapshot they belong to.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: Good point. An earlier proposal didn't have `RequestOfferSnapshot` at all, so the application wouldn't know how many chunks were being restored - `final` is just a vestige from that. Will remove.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: If the chunk fails we return `ResponseException`, so this implicitly is a success. But I am leaning towards returning specific status fields here instead, saying whether the chunk was applied/verified/rejected or some such - then Tendermint also knows what the problem was (internal error vs verification failure) and can take appropriate action.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: Yes, and also because I believe there were some security/DoS implications with nodes giving wrong indexes for chunks (e.g. chunk 3 being sent as chunk 1) and this not being verifiable until the entire tree had been rebuilt unless we verify them in order. I can't remember the details (see https://github.com/tendermint/tendermint/pull/3799), but it seems like a reasonable starting point anyway.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: Yes. But that assumes that the application does that correctly. This is an additional sanity check in case the application either doesn't verify at all, or does so incorrectly.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: Sorry, I'm probably using incorrect terminology here. The block at the snapshotted height should contain the root hash of the state machine confirmed by consensus - what would be the correct term for that hash?

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: That seems like a good idea, I'll add a field.

docs/architecture/adr-053-state-sync-prototype.md
tessr: If the node happens to be a validators, it could be slashed for this, right?

docs/architecture/adr-053-state-sync-prototype.md
tessr: Sounds good to me.

docs/architecture/adr-053-state-sync-prototype.md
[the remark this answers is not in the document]
tessr: Yeah, I guess I can get behind that. Programmers are not great at observing "should" (as in "you _should_ not use this in production") 😛

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: Actually, the checksum should be in the outer envelope, not the serialized payload, so that we can verify checksums without deserializing the payload. Have added this now.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: Sure, but Tendermint can do different things in response. `ResponseOfferSnapshot` returns an internal error, it may make sense to either abort or retry, while a verification error should probably indicate that the sender can't be trusted. I'll be adding status codes for these now.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: True. I'll add a 100 MB chunk size limit. However, since Gaia should use chunk compression, it also needs to be careful when decompressing, since an attacker could send heavily compressed chunks (e.g. 99% compression rate) and cause an OOM DoS.

docs/architecture/adr-053-state-sync-prototype.md
melekes: AppHash? https://github.com/tendermint/tendermint/blob/master/types/block.go#L341

docs/architecture/adr-053-state-sync-prototype.md
melekes: If you can prove it, sure

docs/architecture/adr-053-state-sync-prototype.md
melekes: > If ResponseOfferSnapshot returns an internal error, it may make sense to either abort or retry, while a verification error should probably indicate that the sender can't be trusted

that makes sense

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: Yes, thank you.

Issue #4338, opened 2020-01-24
lite2: swap primary with alternative when the former is unavailable


melekes:
In cases where primary provider crashed (i.e. not responding for some period of time), we should swap it with one of the alternative providers.

melekes:
More info on desired implementation:

after the light client is started, every time we try to access the primary, we should 

a) retry say 5 times (exponential retry with jitter)
b) if it fails, we swap the primary with one of the witnesses (alternative providers) and retry again

cmwaters:
Q: If we swap out the primary, should we append it to the end of the alternatives slice or just remove it completely?

melekes:
Remove because it's not responding imho.

cmwaters:
Yes, okay, I would agree.

PR #4354, opened 2020-01-28
lite2: replace primary provider with alternative when unavailable

cmwaters:
Closes issue #4338 

Uses a wrapper function around both the signedHeader and validatorSet calls to the primary provider which attempts to retrieve the information 5 times before deeming the provider unavailable and replacing the primary provider with the first alternative before trying recursively again (until all alternatives are depleted)

Employs a mutex lock for any operations involving the providers of the light client to ensure no operations occurs whilst the new primary is chosen.

* [x] Wrote tests

### DIFF
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -209,16 +209,6 @@ func NewClient(
 		}
 	}
 
-	if c.removeNoLongerTrustedHeadersPeriod > 0 {
-		c.routinesWaitGroup.Add(1)
-		go c.removeNoLongerTrustedHeadersRoutine()
-	}
-
-	if c.updatePeriod > 0 {
-		c.routinesWaitGroup.Add(1)
-		go c.autoUpdateRoutine()
-	}
-
 	return c, nil
 }
 
@@ -367,10 +357,30 @@ func (c *Client) initializeWithTrustOptions(options TrustOptions) error {
 	return c.updateTrustedHeaderAndVals(h, nextVals)
 }
 
-// Stop stops all the goroutines. If you wish to remove all the data, call
-// Cleanup.
+// Start starts two processes: 1) auto updating 2) removing outdated headers.
+func (c *Client) Start() error {
+	c.logger.Info("Starting light client")
+
+	if c.removeNoLongerTrustedHeadersPeriod > 0 {
+		c.routinesWaitGroup.Add(1)
+		go c.removeNoLongerTrustedHeadersRoutine()
+	}
+
+	if c.updatePeriod > 0 {
+		c.routinesWaitGroup.Add(1)
+		go c.autoUpdateRoutine()
+	}
+
+	return nil
+}
+
+// Stop stops two processes: 1) auto updating 2) removing outdated headers.
+// Stop only returns after both of them are finished running. If you wish to
+// remove all the data, call Cleanup.
 func (c *Client) Stop() {
+	c.logger.Info("Stopping light client")
 	close(c.quit)
+	c.routinesWaitGroup.Wait()
 }
 
 // TrustedHeader returns a trusted header at the given height (0 - the latest)
@@ -413,6 +423,30 @@ func (c *Client) TrustedHeader(height int64, now time.Time) (*types.SignedHeader
 	}
 
 	return h, nil
+}
+
+// TrustedValidatorSet returns a trusted validator set at the given height (0 -
+// the latest) or nil if no such validator set exist. The second return
+// parameter is height validator set corresponds to (useful when you pass 0).
+//
+// height must be >= 0.
+//
+// It returns an error if:
+//	- header signed by that validator set expired (ErrOldHeaderExpired)
+//  - there are some issues with the trusted store, although that should not
+//  happen normally;
+//  - negative height is passed;
+//  - validator set is not found.
+//
+// Safe for concurrent use by multiple goroutines.
+func (c *Client) TrustedValidatorSet(height int64, now time.Time) (*types.ValidatorSet, error) {
+	// Checks height is positive and header is not expired.
+	_, err := c.TrustedHeader(height, now)
+	if err != nil {
+		return nil, err
+	}
+
+	return c.trustedStore.ValidatorSet(height)
 }
 
 // LastTrustedHeight returns a last trusted height. -1 and nil are returned if
@@ -511,12 +545,10 @@ func (c *Client) VerifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
 	return c.updateTrustedHeaderAndVals(newHeader, nextVals)
 }
 
-// Cleanup removes all the data (headers and validator sets) stored. It blocks
-// until internal routines are finished. Note: the client must be stopped at
-// this point.
+// Cleanup removes all the data (headers and validator sets) stored. Note: the
+// client must be stopped at this point.
 func (c *Client) Cleanup() error {
-	c.routinesWaitGroup.Wait()
-	c.logger.Info("Cleanup everything")
+	c.logger.Info("Removing all the data")
 	return c.cleanup(0)
 }
 
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -52,7 +52,7 @@ const (
 
 	defaultUpdatePeriod                       = 5 * time.Second
 	defaultRemoveNoLongerTrustedHeadersPeriod = 24 * time.Hour
-	maxAttempts                               = 5
+	maxRetryAttempts                          = 10
 )
 
 // Option sets a parameter for the light client.
@@ -876,7 +876,7 @@ func (c *Client) replacePrimaryProvider() error {
 	c.providerMutex.Lock()
 	defer c.providerMutex.Unlock()
 	if len(c.witnesses) == 0 {
-		return errors.Errorf("no witnesses left.")
+		return errors.Errorf("no witnesses left")
 	}
 	c.primary = c.witnesses[0]
 	c.witnesses = c.witnesses[1:]
@@ -887,14 +887,14 @@ func (c *Client) replacePrimaryProvider() error {
 // signedHeaderFromPrimary retrieves the SignedHeader from the primary provider at the specified height.
 // Handles dropout by the primary provider by swapping with an alternative provider
 func (c *Client) signedHeaderFromPrimary(height int64) (*types.SignedHeader, error) {
-	for attempt := 0; attempt < maxAttempts; attempt++ {
+	for attempt := 1; attempt <= maxRetryAttempts; attempt++ {
 		c.providerMutex.Lock()
 		h, err := c.primary.SignedHeader(height)
 		c.providerMutex.Unlock()
 		if err == nil || err == provider.ErrSignedHeaderNotFound {
 			return h, err
 		}
-		time.Sleep(backoffAndJitterTime(attempt+1, 1, 5))
+		time.Sleep(time.Duration(1<<attempt)*time.Second + time.Duration(rand.Intn(1000))*time.Millisecond)
 	}
 	c.logger.Info("Primary is unavailable. Replacing with the first witness")
 	err := c.replacePrimaryProvider()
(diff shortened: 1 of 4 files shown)


lite2/client.go
```diff
@@ -146,6 +146,9 @@ type Client struct {
 	// See Witnesses option
 	witnesses []provider.Provider
 
+	// Mutex for locking during changes of the lite clients providers
+	providerMutex sync.Mutex
```
melekes: the convention we're using in Tendermint - fields protected by the mutex should be declared after it. E.g.

```go
mtx      sync.Mutex
A typeA
B typeB

V typeC // fields, which are not protected by mutex, get separated by a new line
```

lite2/client.go
```diff
@@ -520,12 +523,14 @@ func (c *Client) VerifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
 		return errors.Errorf("header at more recent height #%d exists", c.trustedHeader.Height)
 	}
 
+	c.providerMutex.Lock()
 	if len(c.witnesses) > 0 {
 		if err := c.compareNewHeaderWithRandomWitness(newHeader); err != nil {
 			c.logger.Error("Error when comparing new header with one from a witness", "err", err)
```
```suggestion
			c.providerMutex.Unlock()
			c.logger.Error("Error when comparing new header with one from a witness", "err", err)
```

otherwise mutex will stay locked after you `return err`

lite2/client.go
```diff
@@ -855,3 +860,55 @@ func (c *Client) Update(now time.Time) error {
 
 	return nil
 }
+
+// replaceProvider takes the first alternative provider and promotes it as the primary provider
+func (c *Client) replacePrimaryProvider() error {
+	c.providerMutex.Lock()
+	if len(c.witnesses) == 0 {
```
```suggestion
	defer c.providerMutex.Unlock()
	if len(c.witnesses) == 0 {
```

lite2/client.go
```diff
@@ -855,3 +860,55 @@ func (c *Client) Update(now time.Time) error {
...
+		return errors.Errorf("unable to replace provider. No other alternatives exist.")
+	}
+	c.primary = c.witnesses[0]
+	c.witnesses = c.witnesses[1:]
+	c.providerMutex.Unlock()
```
```suggestion
```

lite2/client.go
```diff
@@ -855,3 +860,55 @@ func (c *Client) Update(now time.Time) error {
 
 	return nil
 }
+
+// replaceProvider takes the first alternative provider and promotes it as the primary provider
+func (c *Client) replacePrimaryProvider() error {
+	c.providerMutex.Lock()
+	if len(c.witnesses) == 0 {
+		return errors.Errorf("unable to replace provider. No other alternatives exist.")
```
```suggestion
		return errors.New("no witnesses left")
```

lite2/client.go
```diff
@@ -855,3 +860,55 @@ func (c *Client) Update(now time.Time) error {
...
+	c.primary = c.witnesses[0]
+	c.witnesses = c.witnesses[1:]
+	c.providerMutex.Unlock()
+	return nil
+}
+
+// signedHeaderFromPrimary retrieves the SignedHeader from the primary provider at the specified height.
+// Handles dropout by the primary provider by swapping with an alternative provider
+func (c *Client) signedHeaderFromPrimary(height int64) (*types.SignedHeader, error) {
+	for attempts := 0; attempts < 5; attempts++ {
```
melekes: const maxAttempts = 5

lite2/client.go
```diff
@@ -855,3 +860,55 @@ func (c *Client) Update(now time.Time) error {
...
+		c.providerMutex.Lock()
+		h, err := c.primary.SignedHeader(height)
+		c.providerMutex.Unlock()
+		if err == nil || err == provider.ErrSignedHeaderNotFound {
+			return h, err
+		}
```
melekes: we need to add backoff (time.Sleep) + jitter (grep "jitter")

lite2/client.go
```diff
@@ -855,3 +860,55 @@ func (c *Client) Update(now time.Time) error {
...
+	}
+	c.logger.Info("Unable to communicate with primary provider. Replacing with an alternative.")
```
```suggestion
	c.logger.Info("Primary is unavailable. Replacing with a first witness")
```

lite2/client.go
```diff
@@ -855,3 +860,55 @@ func (c *Client) Update(now time.Time) error {
...
 	return nil
 }
+
+// replaceProvider takes the first alternative provider and promotes it as the primary provider
+func (c *Client) replacePrimaryProvider() error {
+	c.providerMutex.Lock()
+	if len(c.witnesses) == 0 {
+		return errors.Errorf("unable to replace provider. No other alternatives exist.")
+	}
+	c.primary = c.witnesses[0]
```
melekes: c.logger.Info("New primary", "p", c.primary)

lite2/client.go
```diff
@@ -855,3 +860,55 @@ func (c *Client) Update(now time.Time) error {
...
+		return nil, err
+	}
+	return c.signedHeaderFromPrimary(height)
+
+}
+
+// validatorSetFromPrimary retrieves the ValidatorSet from the primary provider at the specified height.
+// Handles dropout by the primary provider after 5 attempts by replacing it with an alternative provider
+func (c *Client) validatorSetFromPrimary(height int64) (*types.ValidatorSet, error) {
+	for attempts := 0; attempts < 5; attempts++ {
```
melekes: same comments

lite2/provider/mock/mock.go
```diff
@@ -45,3 +46,23 @@ func (p *mock) ValidatorSet(height int64) (*types.ValidatorSet, error) {
...
+func NewDeadMock(chainID string) provider.Provider {
+	return &deadMock{chainID: chainID}
+}
+
+func (p *deadMock) ChainID() string {
+	return p.chainID
+}
+
+func (p *deadMock) SignedHeader(height int64) (*types.SignedHeader, error) {
+	return nil, errors.New("No response from provider")
```
melekes:
```suggestion
	return nil, errors.New("no response from provider")
```

lite2/provider/mock/mock.go
```diff
@@ -45,3 +46,23 @@ func (p *mock) ValidatorSet(height int64) (*types.ValidatorSet, error) {
...
+}
+
+func (p *deadMock) ValidatorSet(height int64) (*types.ValidatorSet, error) {
+	return nil, errors.New("No response from provider")
```
melekes:
```suggestion
	return nil, errors.New("no response from provider")
```

melekes:
I think we should also add `Primary() provider.Provider` and `Witnesses() []provider.Provider` functions for introspection since we're now mutating both of them. Then we can use those in tests instead of accessing private fields directly.

lite2/client.go
```diff
@@ -51,6 +52,9 @@ const (
 
 	defaultUpdatePeriod                       = 5 * time.Second
 	defaultRemoveNoLongerTrustedHeadersPeriod = 24 * time.Hour
+	maxAttempts                               = 5
+	// max backOff in seconds
+	backOffCap = 1000
```
melekes: don't think we need that

lite2/client.go
```diff
@@ -903,19 +908,40 @@ func (c *Client) signedHeaderFromPrimary(height int64) (*types.SignedHeader, err
...
 	if err != nil {
 		return nil, err
 	}
 	return c.validatorSetFromPrimary(height)
 
 }
+
+// generates a backoff time between operations that is calculated as a random duration between 0 and a set cap
+func backoffAndJitterTime(attempt int) time.Duration {
+	rand.Seed(time.Now().UnixNano())
```
melekes: seeding should be performed only once

lite2/client.go
```diff
@@ -862,3 +870,77 @@ func (c *Client) Update(now time.Time) error {
...
+// Handles dropout by the primary provider by swapping with an alternative provider
+func (c *Client) signedHeaderFromPrimary(height int64) (*types.SignedHeader, error) {
+	for attempt := 0; attempt < maxAttempts; attempt++ {
+		c.providerMutex.Lock()
+		h, err := c.primary.SignedHeader(height)
+		c.providerMutex.Unlock()
+		if err == nil || err == provider.ErrSignedHeaderNotFound {
+			return h, err
+		}
+		time.Sleep(backoffAndJitterTime(attempt+1, 1, 5))
```
```suggestion
		time.Sleep(time.Duration(1 << attempt)*time.Second + time.Duration(rand.Intn(1000)) * time.Millisecond)
```

lite2/client.go
```diff
@@ -862,3 +870,77 @@ func (c *Client) Update(now time.Time) error {
...
+	c.primary = c.witnesses[0]
+	c.witnesses = c.witnesses[1:]
+	c.logger.Info("New primary", "p", c.primary)
+	return nil
+}
+
+// signedHeaderFromPrimary retrieves the SignedHeader from the primary provider at the specified height.
+// Handles dropout by the primary provider by swapping with an alternative provider
+func (c *Client) signedHeaderFromPrimary(height int64) (*types.SignedHeader, error) {
+	for attempt := 0; attempt < maxAttempts; attempt++ {
```
```suggestion
	for attempt := 1; attempt <= maxRetryAttempts; attempt++ {
```

lite2/client.go
```diff
@@ -862,3 +870,77 @@ func (c *Client) Update(now time.Time) error {
...
+	return c.validatorSetFromPrimary(height)
+
+}
+
+// generates a backoff time between operations that is calculated as a random duration between 0 and a set cap
+func backoffAndJitterTime(attempt, backOffBase, jitter int) time.Duration {
+	return time.Duration((backOffBase*attempt*attempt)+rand.Intn(jitter)) * time.Second
+}
+
+// Primary returns the primary provider for the lite client
```
```suggestion
// Primary returns the primary provider.
```

lite2/client.go
```diff
@@ -862,3 +870,77 @@ func (c *Client) Update(now time.Time) error {
...
+func (c *Client) Primary() provider.Provider {
+	c.providerMutex.Lock()
+	defer c.providerMutex.Unlock()
+	return c.primary
+}
+
+// Witnesses returns the witness providers for the lite client
```
```suggestion
// Witnesses returns the witness providers.
```

cmwaters:
I've implemented the suggested changes. I do notice that the exponential backoff increases test time when we test for provider dropout by a significant margin

lite2/client.go
```diff
@@ -862,3 +870,72 @@ func (c *Client) Update(now time.Time) error {
...
+		}
+		time.Sleep(time.Duration(500*attempt*attempt)*time.Millisecond + time.Duration(rand.Intn(1000))*time.Millisecond)
+	}
+	c.logger.Info("Primary is unavailable. Replacing with the first witness")
+	err := c.replacePrimaryProvider()
+	if err != nil {
+		return nil, err
+	}
+	return c.validatorSetFromPrimary(height)
+
```
marbar3778:
```suggestion
```
spacing nit

lite2/client.go
```diff
@@ -197,6 +200,8 @@ func NewClient(
 		logger:                             log.NewNopLogger(),
 	}
 
+	rand.Seed(time.Now().UnixNano())
```
melekes: not sure we need to do this at all...

lite2/client.go
```diff
@@ -862,3 +870,72 @@ func (c *Client) Update(now time.Time) error {
...
+// Handles dropout by the primary provider by swapping with an alternative provider
+func (c *Client) signedHeaderFromPrimary(height int64) (*types.SignedHeader, error) {
+	for attempt := 1; attempt <= maxRetryAttempts; attempt++ {
+		c.providerMutex.Lock()
+		h, err := c.primary.SignedHeader(height)
+		c.providerMutex.Unlock()
+		if err == nil || err == provider.ErrSignedHeaderNotFound {
+			return h, err
+		}
+		time.Sleep(time.Duration(500*attempt*attempt)*time.Millisecond + time.Duration(rand.Intn(1000))*time.Millisecond)
```
melekes: could you add a comment like `// 500ms -> 2s -> ...` showing the sequence (at least 5 first number and say last number)?

lite2/client.go
```diff
@@ -862,3 +870,72 @@ func (c *Client) Update(now time.Time) error {
...
+// Handles dropout by the primary provider after 5 attempts by replacing it with an alternative provider
+func (c *Client) validatorSetFromPrimary(height int64) (*types.ValidatorSet, error) {
+	for attempt := 1; attempt <= maxRetryAttempts; attempt++ {
+		c.providerMutex.Lock()
+		h, err := c.primary.ValidatorSet(height)
+		c.providerMutex.Unlock()
+		if err == nil || err == provider.ErrValidatorSetNotFound {
+			return h, err
+		}
+		time.Sleep(time.Duration(500*attempt*attempt)*time.Millisecond + time.Duration(rand.Intn(1000))*time.Millisecond)
```
melekes: and you were right about extracting this into the separate function. sorry for messing it up

lite2/client.go
```diff
@@ -862,3 +870,72 @@ func (c *Client) Update(now time.Time) error {
...
+	err := c.replacePrimaryProvider()
+	if err != nil {
+		return nil, err
+	}
+	return c.validatorSetFromPrimary(height)
+
+}
+
+// Primary returns the primary provider.
+func (c *Client) Primary() provider.Provider {
```
melekes: maybe we don't need to return actual providers here, but only `.String()` for introspection

lite2/client_test.go
```diff
@@ -827,3 +828,59 @@ func TestClient_Concurrency(t *testing.T) {
...
+		UpdatePeriod(0),
+		Logger(log.TestingLogger()),
+		Witnesses([]provider.Provider{witness}),
+	)
+	require.NoError(t, err)
+	err = c.Start()
+	require.NoError(t, err)
+	defer c.Stop()
+	assert.NotEqual(t, c.Primary(), primary)
+	assert.Equal(t, 0, len(c.witnesses))
```
melekes:
```suggestion
	assert.Equal(t, 0, len(c.Witnesses()))
```

lite2/provider/mock/mock.go
```diff
@@ -1,6 +1,7 @@
 package mock
 
 import (
+	"github.com/pkg/errors"
```
melekes:
```suggestion
	"github.com/pkg/errors"

```

PR #4356, opened 2020-01-29
lite2: filter valid witnesses

### LINKED ISSUE #4350: lite2: check witnesses are on the same chain as primary upon startup
If not, the light client should return an error.

cmwaters:
Closes #4350

Checks that the chain ID of the witness and that of the lite client are the same before updating the witness list.

* [ ] Wrote tests

### DIFF
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -89,16 +89,12 @@ func SkippingVerification(trustLevel tmmath.Fraction) Option {
 // current primary is unavailable.
 func Witnesses(providers []provider.Provider) Option {
 	return func(c *Client) {
-		validWitnesses := make([]provider.Provider, 0)
 		for _, witness := range providers {
-			if witness.ChainID() == c.ChainID() {
-				validWitnesses = append(validWitnesses, witness)
-			} else {
-				c.logger.Info("Witness possessed a different chain ID and was excluded from lite client witnesses",
-					"witness chain id", witness.ChainID())
+			if witness.ChainID() != c.ChainID() {
+				panic("Witness chainID is not equal to the Lite Client chainID")
 			}
 		}
-		c.witnesses = validWitnesses
+		c.witnesses = providers
 	}
 }
 
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -89,11 +89,16 @@ func SkippingVerification(trustLevel tmmath.Fraction) Option {
 // current primary is unavailable.
 func Witnesses(providers []provider.Provider) Option {
 	return func(c *Client) {
+		validWitnesses := make([]provider.Provider, 0)
 		for _, witness := range providers {
 			if witness.ChainID() == c.ChainID() {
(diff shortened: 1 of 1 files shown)


lite2/client.go
```diff
@@ -89,7 +89,16 @@ func SkippingVerification(trustLevel tmmath.Fraction) Option {
 // current primary is unavailable.
 func Witnesses(providers []provider.Provider) Option {
 	return func(c *Client) {
-		c.witnesses = providers
+		validWitnesses := make([]provider.Provider, 0)
```
tessr: nit: more idiomatic to say `var validWitness []provider.Provider` if you don't know the size of the slice you want to allocate

tessr:
Nice. Really nice PR title and message, too! One more question: #4350 says this function should return an error if the witnesses are on the wrong chain. Do you want to do that, or do you think logging is sufficient?

cmwaters:
> Nice. Really nice PR title and message, too! One more question: #4350 says this function should return an error if the witnesses are on the wrong chain. Do you want to do that, or do you think logging is sufficient?

I was thinking to return an error only if there were zero witnesses because as long as there is at least one witness it should be sufficient for the lite client - but this relates to a more important question that I raised in the actual issue #4350

lite2/client.go
```diff
@@ -89,7 +89,16 @@ func SkippingVerification(trustLevel tmmath.Fraction) Option {
...
+		for _, witness := range providers {
+			if witness.ChainID() == c.ChainID() {
```
marbar3778: nit: maybe a bit nicer if you check if witness.ChainID() != c.ChainID() {c.logger...} then you can drop the else statement.

melekes:
> I was thinking to return an error only if there were zero witnesses because as long as there is at least one witness it should be sufficient for the lite client - but this relates to a more important question that I raised in the actual issue #4350

I am leaning towards returning an error. The number of witnesses is important. If one witness is on another chain, we should return an error due to incorrect configuration.

lite2/client.go
```diff
@@ -89,6 +89,11 @@ func SkippingVerification(trustLevel tmmath.Fraction) Option {
 // current primary is unavailable.
 func Witnesses(providers []provider.Provider) Option {
 	return func(c *Client) {
+		for _, witness := range providers {
+			if witness.ChainID() != c.ChainID() {
+				panic("Witness chainID is not equal to the Lite Client chainID")
```
tessr: I think we probably want this function to add an error as a return type, rather than panic, no?

cmwaters: Yeah that was my initial thought but I noticed that another function that creates an option uses panic. It would also mean changing the function type `Option` to return possible errors for all other options. Let me know if you want me to make the change and I can do it.

marbar3778: does it recover somewhere or causes the node to panic and halt?

PR #4067, opened 2019-10-18
Blockchain reactor v2 reactor

brapse:
Final stage of in the implementation of [ADR-43](https://github.com/tendermint/tendermint/blob/brapse/blockchain-v2-riri-processor-rebase/docs/architecture/adr-043-blockchain-riri-org.md)

Work in progress

* [ ] Referenced an issue explaining the need for the change
* [ ] Updated all relevant documentation in docs
* [ ] Updated all code comments where relevant
* [ ] Wrote tests
* [ ] Updated CHANGELOG_PENDING.md

Issue #4206, opened 2019-11-29
[blockchain] Routine queue configuration in v2


brapse:
The Routine abstraction used in the blockchain v2 package is initialized with a `bufferSize` parameter which is supposed to limit the number of in-flight messages being processed by the corresponding routine. However, the underlying queue data structure only uses that parameter as an initial size of the queue and not the limit. The Routine abstraction should be updated to apply the `bufferSize` parameter to an enforced limit on the in-flight messages such that when `bufferSize` messages are in-flight, `routine.send` calls will return false.

Additionally
- [ ] Queuing behaviour should be documented, when does it load shed?
- [ ] Metrics and logging when messages are dropped
- [ ] Understanding of termination condition if dropping messages
- [ ] Understanding and documenting blocking output channel from routines

ebuchman:
Oh yes, I commented on the three distinct uses of the `bufferSize` in https://github.com/tendermint/tendermint/pull/4361#discussion_r376735278. While it does serve as a real bound for the reactor `events` channel and each routine's next channel, it's just a hint for the priority queues, which are unbounded.

But after [reviewing the PR](https://github.com/tendermint/tendermint/pull/4361#pullrequestreview-355560820) and thinking about this more, I'm not sure we want to bound those priority queues and just drop msgs, since then we'll have a more difficult problem of accounting for dropped msgs. So the unbounded queues may be ok, but we should do some analysis and testing to gain confidence that they will stay within a reasonable size (ie. they should be bounded by things like the max peers and send/recv rates, and by reactor specific mitigations like `targetPending`). The existing metrics should help to monitor this, and we can possibly instrument further by logging errors if the queues get too big ...

brapse:
Agreed! I think we want to load shed at the edge so `blockchainReactor.events`, while internal events can be more liberal.  We need to take a look at the queues better after we wrap up the more minor issues in #4361

marbar3778:
V2 has been removed

PR #4361, opened 2020-01-31
Blockchain-v2 reactor second PR

brapse:
The work includes the reactor which ties together all the seperate routines involved in the design of the blockchain v2 refactor. This PR replaces #4067 which got far too large and messy after a failed attempt to rebase.

* [ ] Wrote tests


blockchain/v2/reactor.go
```diff
@@ -1,118 +1,521 @@
...
+	reporter behaviour.Reporter
+	io       iIo
+	store    blockStore
+}
+
+type blockApplier interface {
+	ApplyBlock(state state.State, blockID types.BlockID, block *types.Block) (state.State, error)
+}
+
+type blockVerifier interface {
```
golangcibot: type `blockVerifier` is unused (from `unused`)

blockchain/v2/processor_context.go
```diff
@@ -4,37 +4,42 @@ import (
...
+	tdState() state.State
 }
 
-// nolint:unused
 type pContext struct {
-	store    *store.BlockStore
-	executor *state.BlockExecutor
-	state    *state.State
+	store    blockStore
+	verifier blockVerifier
```
golangcibot: field `verifier` is unused (from `unused`)

blockchain/v2/reactor_test.go
```diff
@@ -1,22 +1,559 @@
...
+func (mp mockPeer) Status() conn.ConnectionStatus { return conn.ConnectionStatus{} }
+func (mp mockPeer) SocketAddr() *p2p.NetAddress   { return &p2p.NetAddress{} }
+
+func (mp mockPeer) Send(byte, []byte) bool    { return true }
+func (mp mockPeer) TrySend(byte, []byte) bool { return true }
+
+func (mp mockPeer) Set(string, interface{}) {}
+func (mp mockPeer) Get(string) interface{}  { return struct{}{} }
+
+type mockBlockStore struct {
```
golangcibot: type `mockBlockStore` is unused (from `unused`)

blockchain/v2/reactor_test.go
```diff
@@ -1,22 +1,559 @@
...
+type mockBlockApplier struct {
+}
+
+// XXX: Add whitelist/blacklist?
+func (mba *mockBlockApplier) ApplyBlock(state sm.State, blockID types.BlockID, block *types.Block) (sm.State, error) {
+	state.LastBlockHeight++
+	return state, nil
+}
+
+type mockBlockVerifier struct{}
```
golangcibot: type `mockBlockVerifier` is unused (from `unused`)

blockchain/v2/reactor_test.go
```diff
@@ -1,22 +1,564 @@
...
+func makeBlock(height int64, state sm.State, lastCommit *types.Commit) *types.Block {
+	block, _ := state.MakeBlock(height, makeTxs(height), lastCommit, nil, state.Validators.GetProposer().Address)
+	return block
+}
+
+type testApp struct {
+	abci.BaseApplication
+}
+
+func makeVote(header *types.Header, blockID types.BlockID, valset *types.ValidatorSet, privVal types.PrivValidator) *types.Vote {
```
golangcibot: func `makeVote` is unused (from `unused`)

ebuchman:
Few more thoughts from reviewing this and the new TLA+.

## Removing Peers

Right now we remove blocks from the processor's internal queue if the peer was removed. This makes sense for peers that send us bad data, but for peers that only timedout, it might be aggressive to throw away their blocks. Should we consider checking the reason for the peer to be removed before forwarding it to the processor?

## Tickers

There are 4 tickers in the reactor and I wonder if we can/should try to reduce it to 1? It seems some functionality that is currently driven by ticks can be driven by real events instead, though FSMs may then need to be able to output multiple events at a time. Probably the only thing we should really need ticks for are detecting timeouts. In principle I think it means each routine (possibly the whole reactor) can/should have max 1 ticker. For instance:

`doProcessBlock` can be removed by just checking if we can process a block every time we receive one from the scheduler. Then the processor wouldn't need input from any tickers.

`doSchedule` can probably be eliminated if other calls to the scheduler check if we should schedule more requests, and return scBlockRequest as necessary

`doStatus` can probably be similarly emitted by letting the scheduler decide when it needs updated statuses and outputing a new scStatusRequest rather than just leaving this to the reactor.

`doPrunePeer` is then the only thing that's left, and it's only there to detect peer timeouts.

 Does it make sense to consider this? Or is there stronger justification for the FSMs to only output one event at a time?

blockchain/v2/scheduler.go
```diff
@@ -608,35 +570,45 @@ func (sc *scheduler) handleBlockProcessError(event pcBlockVerificationFailure) (
...
 	}
 
 	return noOp, nil
 }
 
-func (sc *scheduler) handleAddNewPeer(event addNewPeer) (Event, error) {
+func (sc *scheduler) handleAddNewPeer(event bcAddNewPeer) (Event, error) {
 	err := sc.addPeer(event.peerID)
 	if err != nil {
 		return scSchedulerFail{reason: err}, nil
```
ebuchman: So it seems these failure events don't trigger anything right? Note some of them seem like genuine internal errors (ie. the ones in `handleTrySchedule`), while others seem highly probable in practice. This one especially because the peer layer may try to re-add a peer we disconnected from in the past. Other fails are due to race conditions. So it certainly doesn't seem like we can consider these fatal without more work, and we should double check that its ok that these will happen.

blockchain/v2/reactor.go
```diff
@@ -1,118 +1,521 @@
...
 	go r.processor.start()
 	go r.demux()
+	return nil
+}
 
-	<-r.scheduler.ready()
-	<-r.processor.ready()
+// reactor generated ticker events:
+// ticker for cleaning peers
+type rTryPrunePeer struct {
```
melekes: why do we need `r` prefix? because events are used outside of this file?

blockchain/v2/reactor.go
```diff
@@ -1,118 +1,521 @@
...
+			}
+		case <-doStatusTk.C:
+			select {
+			case doStatusCh <- struct{}{}:
+			default:
+			}
+
+		// Tickers: perform tasks periodically
+		case <-doScheduleCh:
+			r.scheduler.send(rTrySchedule{time: time.Now()})
```
melekes: can't we merge this with above cases? i.e.

```go
case <-doScheduleTk.C:
  r.scheduler.send(rTrySchedule{time: time.Now()})
```

call scheduler directly?

blockchain/v2/reactor.go
```diff
@@ -1,118 +1,521 @@
...
+	case *bcNoBlockResponseMessage:
+		r.events <- bcNoBlockResponse{peerID: src.ID(), height: msg.Height, time: time.Now()}
+	}
+}
+
+// AddPeer implements Reactor interface
+func (r *BlockchainReactor) AddPeer(peer p2p.Peer) {
+	err := r.io.sendStatusResponse(r.store.Height(), peer.ID())
+	if err != nil {
+		r.logger.Error("Could not send status message to peer new", "src", peer.ID, "height", r.SyncHeight())
```
melekes: is it ok to continue if we can't send something to a peer? just want to double check

blockchain/v2/reactor.go
```diff
@@ -1,118 +1,521 @@
...
 	processor *Routine
-	ticker    *time.Ticker
 	logger    log.Logger
+
+	mtx           sync.RWMutex
+	maxPeerHeight int64
+	syncHeight    int64
+
+	reporter behaviour.Reporter
+	io       iIo
```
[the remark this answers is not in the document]
brapse: Yeah  think we roughly tried to isolate the `input-output` for testing but maybe didn't do such a. good job. Totally open to recommendations ☺️

blockchain/v2/reactor.go
brapse: We prefix the event with `r` to indicate that this is a reactor event.

blockchain/v2/reactor_test.go
```diff
@@ -1,22 +1,564 @@
...
+	"github.com/tendermint/tendermint/mock"
+	"github.com/tendermint/tendermint/p2p"
+	"github.com/tendermint/tendermint/p2p/conn"
+	"github.com/tendermint/tendermint/proxy"
+	sm "github.com/tendermint/tendermint/state"
+	"github.com/tendermint/tendermint/store"
+	"github.com/tendermint/tendermint/types"
+	tmtime "github.com/tendermint/tendermint/types/time"
+	dbm "github.com/tendermint/tm-db"
+	"net"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

blockchain/v2/reactor.go
```diff
@@ -1,118 +1,521 @@
...
+// NewBlockchainReactor creates a new reactor instance.
+func NewBlockchainReactor(state state.State, blockApplier blockApplier, store blockStore, fastSync bool) *BlockchainReactor {
+	reporter := behaviour.NewMockReporter()
+	return newReactor(state, store, reporter, blockApplier, 1000)
+}
+
+// SetSwitch implements Reactor interface.
+func (r *BlockchainReactor) SetSwitch(sw *p2p.Switch) {
+	if sw == nil {
+		panic("Initalized BlockchainReactor with a nil switch")
```
golangcibot: `Initalized` is a misspelling of `Initialized` (from `misspell`)
```suggestion

```

blockchain/v2/processor_context.go
```diff
@@ -45,22 +50,25 @@ func (pc *pContext) saveBlock(block *types.Block, blockParts *types.PartSet, see
 type mockPContext struct {
 	applicationBL  []int64
 	verificationBL []int64
+	state          state.State
 }
 
-func newMockProcessorContext(verificationBlackList []int64, applicationBlackList []int64) *mockPContext {
+func newMockProcessorContext(state state.State, verificationBlackList []int64, applicationBlackList []int64) *mockPContext {
```
golangcibot: line is 124 characters (from `lll`)

blockchain/v2/reactor.go
```diff
@@ -1,118 +1,521 @@
...
+		store:     store,
+		reporter:  reporter,
 		logger:    log.NewNopLogger(),
 	}
 }
 
-// nolint:unused
-func (r *Reactor) setLogger(logger log.Logger) {
+// NewBlockchainReactor creates a new reactor instance.
+func NewBlockchainReactor(state state.State, blockApplier blockApplier, store blockStore, fastSync bool) *BlockchainReactor {
```
golangcibot: line is 125 characters (from `lll`)

blockchain/v2/processor_test.go
```diff
@@ -256,9 +242,8 @@ func TestPcProcessBlockFailures(t *testing.T) {
 			name: "blocks H+1 and H+2 present from same peers - H+1 verification fails ",
 			steps: []pcFsmMakeStateValues{
 				{
-					currentState:  &params{items: []pcBlock{{"P1", 1}, {"P1", 2}, {"P2", 3}}, verBL: []int64{1}},
-					event:         pcProcessBlock{},
-					wantState:     &params{items: []pcBlock{{"P2", 3}}, verBL: []int64{1}},
+					currentState: &params{height: 0, items: []pcBlock{{"P1", 1}, {"P1", 2}, {"P2", 3}}, verBL: []int64{1}}, event: rProcessBlock{},
```
golangcibot: line is 132 characters (from `lll`)

blockchain/v2/processor_test.go
```diff
@@ -267,46 +252,8 @@ func TestPcProcessBlockFailures(t *testing.T) {
...
-			},
-		},
-		{
-			name: "all blocks are from errored peer",
-			steps: []pcFsmMakeStateValues{
-				{
-					currentState: &params{items: []pcBlock{{"P1", 100}, {"P1", 99}}}, event: &peerError{peerID: "P1"},
-					wantState:     &params{},
-					wantNextEvent: noOp,
+					currentState: &params{items: []pcBlock{{"P1", 1}, {"P2", 2}, {"P2", 3}}, appBL: []int64{1}}, event: rProcessBlock{},
```
golangcibot: line is 121 characters (from `lll`)

blockchain/v2/processor_test.go
```diff
@@ -315,35 +262,35 @@ func TestPcPeerError(t *testing.T) {
...
+					wantErr:       pcFinished{tmState: tmState.State{LastBlockHeight: 100}, blocksSynced: 100},
 				},
 			},
 		},
 		{
 			name: "more blocks present",
 			steps: []pcFsmMakeStateValues{
 				{
-					currentState: &params{height: 100, items: []pcBlock{{"P1", 101}, {"P1", 102}}, blocksSynced: 100}, event: pcStop{},
+					currentState: &params{height: 100, items: []pcBlock{{"P1", 101}, {"P1", 102}}, blocksSynced: 100}, event: scFinishedEv{},
```
golangcibot: line is 126 characters (from `lll`)

blockchain/v2/reactor_test.go
```diff
@@ -1,22 +1,564 @@
...
+
+// XXX: Add whitelist/blacklist?
+func (mba *mockBlockApplier) ApplyBlock(state sm.State, blockID types.BlockID, block *types.Block) (sm.State, error) {
+	state.LastBlockHeight++
+	return state, nil
+}
+
+type mockBlockVerifier struct{}
+
+func (mbv *mockBlockVerifier) VerifyCommit(chainID string, blockID types.BlockID, height int64, commit *types.Commit) error {
```
golangcibot: line is 125 characters (from `lll`)

blockchain/v2/reactor_test.go
```diff
@@ -1,22 +1,564 @@
...
+		Type:             types.PrecommitType,
+		BlockID:          blockID,
+	}
+
+	_ = privVal.SignVote(header.ChainID, vote)
+
+	return vote
+}
+
+func randGenesisDoc(chainID string, numValidators int, randPower bool, minPower int64) (*types.GenesisDoc, []types.PrivValidator) {
```
golangcibot: line is 131 characters (from `lll`)

blockchain/v2/scheduler_test.go
```diff
@@ -268,11 +280,10 @@ func TestScTouchPeer(t *testing.T) {
...
 		{
-			name:   "touch peer in state Ready",
-			fields: scTestParams{peers: map[string]*scPeer{"P1": {state: peerStateReady, lastTouched: now}}},
-			args:   args{peerID: "P1", time: now.Add(3 * time.Second)},
-			wantFields: scTestParams{peers: map[string]*scPeer{"P1": {state: peerStateReady,
-				lastTouched: now.Add(3 * time.Second)}}},
+			name:       "touch peer in state Ready",
+			fields:     scTestParams{peers: map[string]*scPeer{"P1": {state: peerStateReady, lastTouched: now}}},
+			args:       args{peerID: "P1", time: now.Add(3 * time.Second)},
+			wantFields: scTestParams{peers: map[string]*scPeer{"P1": {state: peerStateReady, lastTouched: now.Add(3 * time.Second)}}},
```
golangcibot: line is 125 characters (from `lll`)

blockchain/v2/scheduler_test.go
```diff
@@ -1082,8 +904,7 @@ func TestScMarkReceived(t *testing.T) {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			sc := newTestScheduler(tt.fields)
-			if err := sc.markReceived(tt.args.peerID,
-				tt.args.height, tt.args.size, now.Add(time.Second)); (err != nil) != tt.wantErr {
+			if err := sc.markReceived(tt.args.peerID, tt.args.height, tt.args.size, now.Add(time.Second)); (err != nil) != tt.wantErr {
```
golangcibot: line is 126 characters (from `lll`)

blockchain/v2/scheduler.go
```diff
@@ -608,35 +570,45 @@ func (sc *scheduler) handleBlockProcessError(event pcBlockVerificationFailure) (
...
 	}
 	return noOp, nil
 }
 
-func (sc *scheduler) handleTryPrunePeer(event tryPrunePeer) (Event, error) {
+func (sc *scheduler) handleTryPrunePeer(event rTryPrunePeer) (Event, error) {
+
+	// Check behavior of peer responsible to deliver block at sc.height.
+	timeHeightAsked, ok := sc.pendingTime[sc.height]
+	if ok && time.Now().Sub(timeHeightAsked) > sc.peerTimeout {
```
golangcibot: S1012: should use `time.Since` instead of `time.Now().Sub` (from `gosimple`)

blockchain/v2/scheduler.go
```diff
@@ -649,17 +621,20 @@ func (sc *scheduler) handleTryPrunePeer(event tryPrunePeer) (Event, error) {
...
 
 	return scPeersPruned{peers: prunablePeers}, nil
 
 }
 
-// TODO - Schedule multiple block requests
-func (sc *scheduler) handleTrySchedule(event trySchedule) (Event, error) {
+func (sc *scheduler) handleTrySchedule(event rTrySchedule) (Event, error) {
+
+	if time.Now().Sub(sc.lastAdvance) > sc.syncTimeout {
```
golangcibot: S1012: should use `time.Since` instead of `time.Now().Sub` (from `gosimple`)

blockchain/v2/reactor.go
marbar3778: could this be added to the `doc.go` file so someone doesn't remove it in the future

blockchain/v2/processor.go
```diff
@@ -4,48 +4,32 @@ import (
...
-type pcStop struct {
-	priorityNormal
-}
-
+// processor has finished
 type pcFinished struct {
 	priorityNormal
 	height       int64
-	blocksSynced int64
+	blocksSynced int
```
marbar3778: do we want to be more strict with int types? is there a reason why this was transitioned from int64 to int?

blockchain/v2/reactor.go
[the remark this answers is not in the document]
melekes: Yeah, I just did not understand what `i` prefix stands for. interface? so like iIo - IO interface?

blockchain/v2/reactor.go
```diff
@@ -1,118 +1,520 @@
...
+	reporter := behaviour.NewMockReporter()
+	return newReactor(state, store, reporter, blockApplier, 1000)
+}
+
+// SetSwitch implements Reactor interface.
+func (r *BlockchainReactor) SetSwitch(sw *p2p.Switch) {
+	if sw == nil {
```
golangcibot: SA9003: empty branch (from `staticcheck`)

blockchain/v2/processor.go
```diff
@@ -129,23 +101,34 @@ func (state *pcState) purgePeer(peerID p2p.ID) {
...
+
+	case scBlockReceived:
+		if event.block == nil {
+			return noOp, nil
 		}
-		err := state.enqueue(event.peerID, event.block, event.block.Height)
-		if err != nil {
-			return pcDuplicateBlock{}, nil
+
+		// enqueue block if height is higher than state height, else ignore it
```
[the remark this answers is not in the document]
ancazamfir: It is possible that the scheduler detected an error for the peer that sent us block at height H and re-requested from another peer. Examples are when this peer sends us a block we haven't asked for or lowered its height. Meantime the processor validated H, sent a `pcProcessedBlock` towards the scheduler and moved to H+1. Before the scheduler processes this, it receives and forwards `scBlockReceived` for H from a different peer towards the processor. 
processor and scheduler are two hops away wrt communication and there may be inflight messages in both directions.

blockchain/v2/processor.go
```diff
@@ -129,23 +101,34 @@ func (state *pcState) purgePeer(peerID p2p.ID) {
...
+		return noOp, nil
 
-	case pcProcessBlock:
+	case rProcessBlock:
+		tdState := state.context.tdState()
 		firstItem, secondItem, err := state.nextTwo()
 		if err != nil {
 			if state.draining {
-				return noOp, pcFinished{height: state.height}
+				return noOp, pcFinished{tdState: tdState, blocksSynced: state.blocksSynced}
```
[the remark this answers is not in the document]
ancazamfir: ouch, yes!

ancazamfir:
> Few more thoughts from reviewing this and the new TLA+.
> 
> ## Removing Peers
> Right now we remove blocks from the processor's internal queue if the peer was removed. This makes sense for peers that send us bad data, but for peers that only timedout, it might be aggressive to throw away their blocks. Should we consider checking the reason for the peer to be removed before forwarding it to the processor?

There isn't a `peerRemove` event going to the processor, only `scPeerError`. This is sent by the scheduler to the processor when a peer is deemed faulty (lowered height, sent us block we haven't asked for). A peer that is timed-out by the scheduler is only removed from the scheduler. The `scPeersPruned` event is sinked in the reactor.  While the scheduler removes all blocks from those peers, any unprocessed received blocks are in the processor's queue and may be processed or later removed or replaced by other blocks. 
Then there is the peerRemove event that comes from the Switch and that event is accompanied by a `reason` field. We ignore this as the peer is gone anyway and we need to remove it. Again, any blocks from this peer are left in the processor's queue. Maybe here we should remove them if the reason is not `nil`.
> 
> ## Tickers
> There are 4 tickers in the reactor and I wonder if we can/should try to reduce it to 1? It seems some functionality that is currently driven by ticks can be driven by real events instead, though FSMs may then need to be able to output multiple events at a time. Probably the only thing we should really need ticks for are detecting timeouts. In principle I think it means each routine (possibly the whole reactor) can/should have max 1 ticker. For instance:
> 
> `doProcessBlock` can be removed by just checking if we can process a block every time we receive one from the scheduler. Then the processor wouldn't need input from any tickers.
> 
> `doSchedule` can probably be eliminated if other calls to the scheduler check if we should schedule more requests, and return scBlockRequest as necessary
> 
> `doStatus` can probably be similarly emitted by letting the scheduler decide when it needs updated statuses and outputing a new scStatusRequest rather than just leaving this to the reactor.
> 
> `doPrunePeer` is then the only thing that's left, and it's only there to detect peer timeouts. In principle we could rename it to `doTick` and it could trigger scheduling block or status requests as well.
> 
> Does it make sense to consider this? Or is there stronger justification for the FSMs to only output one event at a time?

Yes, I was thinking about this. But on the other hand I believe we should isolate the time related handling from the FSMs and there should be explicit events as if we would have separate timers. This is very convenient for verification (as we discovered in the TLA+ specs for both v1 and v2).
I will look at having one ticker that generates multiple events for the scheduler and one for the processor.
I also found it cleaner to make the block requests on a ticker base vs. triggered by other events...but we can discuss...
Should we cover this change under a different PR? It is a bit more involved.

blockchain/v2/reactor.go
ancazamfir: I think it's ok. The peer will send a `statusRequest` later.

blockchain/v2/processor.go
```diff
@@ -102,18 +80,12 @@ func (state *pcState) synced() bool {
 	return len(state.queue) <= 1
 }
 
-func (state *pcState) advance() {
-	state.height++
-	delete(state.queue, state.height)
-	state.blocksSynced++
+func (state *pcState) enqueue(peerID p2p.ID, block *types.Block, height int64) {
+	state.queue[height] = queueItem{block: block, peerID: peerID}
```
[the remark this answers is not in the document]
ancazamfir: I believe this will will overwrite the existing block.

blockchain/v2/scheduler.go
ancazamfir: yeah, we need to comb through the cases and make sure we are not in a corrupt state.

blockchain/v2/processor.go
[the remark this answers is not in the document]
ancazamfir: fixed in https://github.com/tendermint/tendermint/pull/4361/commits/27aae37158593f7d307247ecc2a1b482f7dfd507

blockchain/v2/reactor.go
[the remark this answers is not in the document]
brapse: Yes, not the best naming i'm sure.

blockchain/v2/reactor.go
```diff
@@ -1,118 +1,521 @@
...
+	"github.com/tendermint/tendermint/state"
+	"github.com/tendermint/tendermint/types"
 )
 
-type timeCheck struct {
-	priorityHigh
-	time time.Time
+//-------------------------------------
+
+type bcBlockRequestMessage struct {
```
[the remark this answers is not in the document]
ancazamfir: at this point the messages/ events are owned and defined by the entities that generate/ send them.

blockchain/v2/reactor.go
brapse: Indeed, will address in a follow up PR #4410.

blockchain/v2/processor.go
```diff
@@ -60,37 +44,38 @@ type queueItem struct {
 type blockQueue map[int64]queueItem
 
 type pcState struct {
-	height       int64      // height of the last synced block
-	queue        blockQueue // blocks waiting to be processed
-	chainID      string
-	blocksSynced int64
+	// blocks waiting to be processed
+	queue        blockQueue
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	queue blockQueue
```

blockchain/v2/processor_context.go
```diff
@@ -4,37 +4,41 @@ import (
...
 	saveBlock(block *types.Block, blockParts *types.PartSet, seenCommit *types.Commit)
+	tmState() state.State
 }
 
-// nolint:unused
 type pContext struct {
-	store    *store.BlockStore
-	executor *state.BlockExecutor
-	state    *state.State
+	store    blockStore
```
golangcibot: File is not `goimports`-ed (from `goimports`)

blockchain/v2/reactor.go
```diff
@@ -1,118 +1,522 @@
...
+		reporter:  reporter,
 		logger:    log.NewNopLogger(),
 	}
 }
 
-// nolint:unused
-func (r *Reactor) setLogger(logger log.Logger) {
+// NewBlockchainReactor creates a new reactor instance.
+func NewBlockchainReactor(
+		state state.State,
```
golangcibot: File is not `goimports`-ed (from `goimports`)

blockchain/v2/scheduler_test.go
```diff
@@ -2123,24 +1985,26 @@ func TestScHandle(t *testing.T) {
 			name: "single peer, sync 3 blocks",
 			steps: []scStep{
 				{ // add P1
-					currentSc: &scTestParams{peers: map[string]*scPeer{}, height: 1},
-					args:      args{event: addNewPeer{peerID: "P1"}},
+					currentSc: &scTestParams{startTime: now, peers: map[string]*scPeer{}, height: 1},
+					args:      args{event: bcAddNewPeer{peerID: "P1"}},
 					wantEvent: noOpEvent{},
-					wantSc:    &scTestParams{peers: map[string]*scPeer{"P1": {height: -1, state: peerStateNew}}, height: 1},
+					wantSc:    &scTestParams{startTime: now, peers: map[string]*scPeer{"P1": {height: -1, state: peerStateNew}}, height: 1},
```
golangcibot: line is 125 characters (from `lll`)

blockchain/v2/io.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/tendermint/tendermint/state"
+	"github.com/tendermint/tendermint/types"
+)
+
+type iIo interface {
+	sendBlockRequest(peerID p2p.ID, height int64) error
+	sendBlockToPeer(block *types.Block, peerID p2p.ID) error
+	sendBlockNotFound(height int64, peerID p2p.ID) error
+	sendStatusResponse(height int64, peerID p2p.ID) error
+	switchToConsensus(state state.State, blocksSynced int)
```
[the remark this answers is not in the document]
brapse: Yes all of this does sound right and i tried to capture it in a follow up issue #4206

blockchain/v2/reactor.go
```diff
@@ -1,118 +1,521 @@
...
 		logger:    log.NewNopLogger(),
 	}
 }
 
-// nolint:unused
-func (r *Reactor) setLogger(logger log.Logger) {
+// NewBlockchainReactor creates a new reactor instance.
+func NewBlockchainReactor(state state.State, blockApplier blockApplier, store blockStore, fastSync bool) *BlockchainReactor {
+	reporter := behaviour.NewMockReporter()
+	return newReactor(state, store, reporter, blockApplier, 1000)
```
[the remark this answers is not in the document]
brapse: Tried to capture the queuing concerns in #4206

blockchain/v2/reactor.go
```diff
@@ -1,118 +1,521 @@
...
 		case event := <-r.scheduler.next():
-			r.processor.send(event)
+			switch event := event.(type) {
+			case scBlockReceived:
+				r.processor.send(event)
+			case scPeerError:
+				r.processor.send(event)
+				r.reporter.Report(behaviour.BadMessage(event.peerID, "scPeerError"))
+			case scBlockRequest:
+				r.io.sendBlockRequest(event.peerID, event.height)
```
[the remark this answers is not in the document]
brapse: Created a follow up issue #4414.

blockchain/v2/reactor.go
```diff
@@ -1,118 +1,521 @@
...
+			}
+		} else {
+			r.logger.Info("peer asking for a block we don't have", "src", src, "height", msg.Height)
+			peerID := src.ID()
+			if err = r.io.sendBlockNotFound(msg.Height, peerID); err != nil {
+				r.logger.Error("Couldn't send block not found: ", err)
+			}
+		}
+
+	case *bcStatusResponseMessage:
```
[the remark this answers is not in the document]
brapse: Make sense, should be covered in follow up issue #4206

blockchain/v2/scheduler.go
brapse: Made a follow up issue. #4415

blockchain/v2/processor.go
brapse: Agreed. Made an issue here #4416

blockchain/v2/processor.go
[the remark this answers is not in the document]
brapse: Fixed!

melekes:
@brapse do you want me to fix https://golangci.com/r/github.com/tendermint/tendermint/pulls/4361?

brapse:
Some of the mocks here are not used but probably should be. Created a follow up issue #4430.

melekes:
Merged. For some reason github did not close the PR. The commit is on master.

PR #4362, opened 2020-01-31
lite2: fetch missing headers

### LINKED ISSUE #4328: lite2: TrustedHeader method should be able to fetch missing headers
if they are within the trusting period. Some headers might be skipped during bisection, but still required say to verify a transaction(s) from associated blocks.

What to do:
1) find closest trusted header in the store 
2) perform either backwards verification https://github.com/tendermint/spec/blob/master/spec/consensus/light-client/verification.md#the-case-untrustedheaderheight--trustedheaderheight or ~~forward verification (what `VerifyHeader` does).~~
3) If header can be trusted, save it (if no, return an error)

cmwaters:
Closes #4328 

When `TrustedHeader(height)` is called, if the height is less than the trusted height but the header is not in the trusted store then a function finds the previous lowest height with a trusted header and performs a forwards sequential verification to the header of the height that was given. If no error is found it updates the trusted store with the header and validator set for that height and can then return them to the user. 

* [x] Wrote tests

### DIFF
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -980,20 +980,6 @@ func (c *Client) Update(now time.Time) error {
 	return nil
 }
 
-func (c *Client) previousTrustedHeight(height int64) (int64, error) {
-	firstTrustedHeight, err := c.FirstTrustedHeight()
-	if err != nil {
-		return -1, err
-	}
-	for i := height - 1; i > firstTrustedHeight; i-- {
-		previousTrustedHeader, err := c.trustedStore.SignedHeader(i)
-		if err == nil {
-			return previousTrustedHeader.Height, nil
-		}
-	}
-	return firstTrustedHeight, nil
-}
-
 // replaceProvider takes the first alternative provider and promotes it as the primary provider
 func (c *Client) replacePrimaryProvider() error {
 	c.providerMutex.Lock()
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -787,13 +787,43 @@ func (c *Client) fetchHeaderAndValsAtHeight(height int64) (*types.SignedHeader, 
 func (c *Client) fetchMissingTrustedHeader(height int64, now time.Time) (*types.SignedHeader, error) {
 	c.logger.Info("Fetching missing header", "height", height)
 
-	trustedHeader, err := c.trustedStore.SignedHeaderAfter(height)
+	closestHeader, err := c.trustedStore.SignedHeaderAfter(height)
 	if err != nil {
 		return nil, errors.Wrapf(err, "can't get signed header after %d", height)
 	}
 
-	var untrustedHeader *types.SignedHeader
-	for i := trustedHeader.Height - 1; i >= height; i-- {
+	// Perform backwards verification from closestHeader to header at the given
+	// height.
+	h, err := c.backwards(height, closestHeader, now)
+	if err != nil {
+		return nil, err
+	}
+
+	// Fetch next validator set from primary and persist it.
+	nextVals, err := c.validatorSetFromPrimary(height + 1)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to obtain the vals #%d", height)
+	}
+	if !bytes.Equal(h.NextValidatorsHash, nextVals.Hash()) {
+		return nil, errors.Errorf("expected next validator's hash %X, but got %X",
+			h.NextValidatorsHash, nextVals.Hash())
+	}
+	if err := c.trustedStore.SaveSignedHeaderAndNextValidatorSet(h, nextVals); err != nil {
+		return nil, errors.Wrap(err, "failed to save trusted header")
+	}
+
+	return h, nil
+}
+
+// Backwards verification (see VerifyHeaderBackwards func in the spec)
+func (c *Client) backwards(toHeight int64, fromHeader *types.SignedHeader, now time.Time) (*types.SignedHeader, error) {
+	var (
+		trustedHeader   = fromHeader
+		untrustedHeader *types.SignedHeader
+		err             error
+	)
+
+	for i := trustedHeader.Height - 1; i >= toHeight; i-- {
 		untrustedHeader, err = c.signedHeaderFromPrimary(i)
 		if err != nil {
 			return nil, errors.Wrapf(err, "failed to obtain the header #%d", i)
@@ -801,12 +831,6 @@ func (c *Client) fetchMissingTrustedHeader(height int64, now time.Time) (*types.
 
 		if err := untrustedHeader.ValidateBasic(c.chainID); err != nil {
 			return nil, errors.Wrap(err, "untrustedHeader.ValidateBasic failed")
-		}
-
-		if untrustedHeader.Height >= trustedHeader.Height {
-			return nil, errors.Errorf("expected older header height %d to be less than one of newer header %d",
-				untrustedHeader.Height,
-				trustedHeader.Height)
 		}
 
 		if !untrustedHeader.Time.Before(trustedHeader.Time) {
@@ -824,21 +848,11 @@ func (c *Client) fetchMissingTrustedHeader(height int64, now time.Time) (*types.
 				untrustedHeader.Hash(),
 				trustedHeader.LastBlockID.Hash)
 		}
+
+		trustedHeader = untrustedHeader
 	}
 
-	nextVals, err := c.validatorSetFromPrimary(height + 1)
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to obtain the vals #%d", height)
-	}
-	if !bytes.Equal(untrustedHeader.NextValidatorsHash, nextVals.Hash()) {
-		return nil, errors.Errorf("expected next validator's hash %X, but got %X",
-			untrustedHeader.NextValidatorsHash, nextVals.Hash())
-	}
-	if err := c.trustedStore.SaveSignedHeaderAndNextValidatorSet(untrustedHeader, nextVals); err != nil {
-		return nil, errors.Wrap(err, "failed to save trusted header")
-	}
-
-	return untrustedHeader, nil
+	return trustedHeader, nil
 }
 
 // compare header with one from a random witness.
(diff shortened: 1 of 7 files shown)


lite2/client.go
```diff
@@ -454,6 +459,42 @@ func (c *Client) TrustedValidatorSet(height int64, now time.Time) (*types.Valida
...
+	}
+	c.trustedNextVals, err = c.trustedStore.ValidatorSet(prevTrustedHeight)
+	if err != nil {
+		return nil, nil, err
+	}
+	newHeader, err = c.primary.SignedHeader(height)
+	if err != nil {
+		return nil, nil, err
+	}
+	newVals, err = c.primary.ValidatorSet(height)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

lite2/client.go
```diff
@@ -862,3 +903,20 @@ func (c *Client) Update(now time.Time) error {
...
 	return nil
 }
+
+func (c *Client) previousTrustedHeight(height int64) (int64, error) {
+	FirstTrustedHeight, err := c.FirstTrustedHeight()
+	if err != nil {
+		return -1, err
+	}
+	if height < FirstTrustedHeight {
+		return -1, errors.New("height is less than first trusted height.")
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

lite2/client.go
```diff
@@ -408,25 +408,30 @@ func (c *Client) TrustedHeader(height int64, now time.Time) (*types.SignedHeader
...
+	if latestHeight <= 0 {
+		return nil, errors.New("no headers exist")
+	}
 	if height == 0 {
-		var err error
-		height, err = c.LastTrustedHeight()
+		height = latestHeight
+	}
+	h, err := c.trustedStore.SignedHeader(height)
+	if err == provider.ErrSignedHeaderNotFound {
```
marbar3778:
```suggestion
	if errors.Is(err, provider.ErrSignedHeaderNotFound) {
```

lite2/client.go
```diff
@@ -408,25 +408,30 @@ func (c *Client) TrustedHeader(height int64, now time.Time) (*types.SignedHeader
...
+		//fetch header from primary and verify
+		h, _, err = c.updateMissingTrustedHeaderAndVals(height)
```
```suggestion
		h, _, err = c.fetchMissingTrustedHeaderAndVals(height)
```

lite2/client.go
```diff
@@ -454,6 +459,45 @@ func (c *Client) TrustedValidatorSet(height int64, now time.Time) (*types.Valida
...
+	}
+	c.trustedHeader, err = c.trustedStore.SignedHeader(prevTrustedHeight)
+	if err != nil {
+		return nil, nil, err
+	}
+	c.trustedNextVals, err = c.trustedStore.ValidatorSet(prevTrustedHeight)
+	if err != nil {
+		return nil, nil, err
+	}
+	newHeader, err = c.primary.SignedHeader(height)
```
melekes: we have a function for that - `c.fetchHeaderAndValsAtHeight(height)`

melekes: immediately after we should check if it's not expired I think. before checking if we can trust it

lite2/client.go
```diff
@@ -454,6 +459,45 @@ func (c *Client) TrustedValidatorSet(height int64, now time.Time) (*types.Valida
...
+	if err != nil {
+		return nil, nil, err
+	}
+	newVals, err = c.primary.ValidatorSet(height)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// updates store if valid
+	err = c.sequence(newHeader, newVals, time.Now())
```
melekes: this seems incomplete. we should do either a) bisection/sequence depending on verificationMode IF we're going forward b) `backwards` verification IF we're going backwards

lite2/client.go
```diff
@@ -862,3 +906,20 @@ func (c *Client) Update(now time.Time) error {
 
 	return nil
 }
+
+func (c *Client) previousTrustedHeight(height int64) (int64, error) {
+	FirstTrustedHeight, err := c.FirstTrustedHeight()
```
```suggestion
	firstTrustedHeight, err := c.FirstTrustedHeight()
```

lite2/client.go
```diff
@@ -862,3 +906,20 @@ func (c *Client) Update(now time.Time) error {
...
+	if err != nil {
+		return -1, err
+	}
+	if height < FirstTrustedHeight {
```
melekes: we won't need to check this if we do (immediately after we should check if it's not expired I think. before checking if we can trust it)

lite2/client.go
cmwaters: Checking if it is expired is done in the `Verify` function which is part of the `sequence` functions

lite2/client.go
cmwaters: What is the criteria for deciding whether we are going forwards or backwards?

lite2/client.go
```diff
@@ -415,25 +415,30 @@ func (c *Client) TrustedHeader(height int64, now time.Time) (*types.SignedHeader
 	if height < 0 {
 		return nil, errors.New("negative height")
 	}
-
+	latestHeight, err := c.LastTrustedHeight()
```
jackzampolin: So just to clarify here, this will fetch a header from a provider if there are no headers in the DB?

cmwaters: Yes but only if the requested header is between the first and last trusted header heights in the db. i.e. if you call one that is more recent you will need to call an update first

melekes:
TODO:

- [ ] write `Test_SignedHeaderAfter` in db_test.go (see `TODO` in the code)
- [ ] test fetching a missing header in client_test.go

lite2/client.go
```diff
@@ -455,8 +478,10 @@ func (c *Client) TrustedHeader(height int64, now time.Time) (*types.SignedHeader
 //
 // Safe for concurrent use by multiple goroutines.
 func (c *Client) TrustedValidatorSet(height int64, now time.Time) (*types.ValidatorSet, error) {
-	// Checks height is positive and header is not expired.
-	_, err := c.TrustedHeader(height, now)
+	// Checks height is positive and header (note: height - 1) is not expired.
+	// Additionaly, it fetches validator set from primary if it's missing in
```
golangcibot: `Additionaly` is a misspelling of `Additionally` (from `misspell`)
```suggestion

```

lite2/test_helpers.go
```diff
@@ -147,3 +147,15 @@ func (pkz privKeys) GenSignedHeader(chainID string, height int64, bTime time.Tim
 		Commit: pkz.signHeader(header, first, last),
 	}
 }
+
+// GenSignedHeaderLastBlockID calls genHeader and signHeader and combines them into a SignedHeader.
+func (pkz privKeys) GenSignedHeaderLastBlockID(chainID string, height int64, bTime time.Time, txs types.Txs,
+	valset, nextValset *types.ValidatorSet, appHash, consHash, resHash []byte, first, last int, lastBlockID types.BlockID) *types.SignedHeader {
```
golangcibot: line is 141 characters (from `lll`)

lite2/client_test.go
```diff
@@ -919,7 +919,69 @@ func TestProvider_Replacement(t *testing.T) {
...
+
+	// 1) header is missing => expect no error
+	h, err := c.TrustedHeader(2, bTime.Add(1*time.Hour).Add(1*time.Second))
+	require.NoError(t, err)
+	if assert.NotNil(t, h) {
+		assert.EqualValues(t, 2, h.Height)
+	}
+
+	// 2) header is missing, but it's expired => expect error
+	h, err = c.TrustedHeader(1, bTime.Add(1*time.Hour).Add(1*time.Second))
```

jackzampolin:
Needs rebase

lite2/client.go
```diff
@@ -757,6 +782,65 @@ func (c *Client) fetchHeaderAndValsAtHeight(height int64) (*types.SignedHeader,
...
 }
 
+// fetchMissingTrustedHeader finds the closest height after the
+// requested height and does backwards verification.
+func (c *Client) fetchMissingTrustedHeader(height int64, now time.Time) (*types.SignedHeader, error) {
+	c.logger.Info("Fetching missing header", "height", height)
+
+	trustedHeader, err := c.trustedStore.SignedHeaderAfter(height)
+	if err != nil {
+		return nil, errors.Wrapf(err, "can't get signed header after %d", height)
```
cmwaters: Suggestion: Instead of throwing back an error could we just use c.trustedHeader i.e. the latest trusted header of the lite client and use back verification from there.

lite2/client.go
```diff
@@ -757,6 +782,65 @@ func (c *Client) fetchHeaderAndValsAtHeight(height int64) (*types.SignedHeader,
...
+
+		if err := untrustedHeader.ValidateBasic(c.chainID); err != nil {
+			return nil, errors.Wrap(err, "untrustedHeader.ValidateBasic failed")
+		}
+
+		if untrustedHeader.Height >= trustedHeader.Height {
+			return nil, errors.Errorf("expected older header height %d to be less than one of newer header %d",
+				untrustedHeader.Height,
+				trustedHeader.Height)
+		}
```
cmwaters: `if untrustedHeader.Height >= trustedHeader.Height` - would this ever happen that the primary returns a different height then the one you requested for. In this case shouldn't the `signedHeaderFromPrimary` function filter for that.

lite2/client.go
```diff
@@ -757,6 +782,65 @@ func (c *Client) fetchHeaderAndValsAtHeight(height int64) (*types.SignedHeader,
...
+func (c *Client) fetchMissingTrustedHeader(height int64, now time.Time) (*types.SignedHeader, error) {
+	c.logger.Info("Fetching missing header", "height", height)
+
+	trustedHeader, err := c.trustedStore.SignedHeaderAfter(height)
+	if err != nil {
+		return nil, errors.Wrapf(err, "can't get signed header after %d", height)
+	}
+
+	var untrustedHeader *types.SignedHeader
+	for i := trustedHeader.Height - 1; i >= height; i-- {
```
cmwaters: Would you want the actual backwards verification process to be encapsulated as a separate function. (perhaps in the `verifier.go` file)

lite2/client.go
```diff
@@ -757,6 +782,65 @@ func (c *Client) fetchHeaderAndValsAtHeight(height int64) (*types.SignedHeader,
...
+
+		if HeaderExpired(untrustedHeader, c.trustingPeriod, now) {
+			return nil, ErrOldHeaderExpired{untrustedHeader.Time.Add(c.trustingPeriod), now}
+		}
+
+		if !bytes.Equal(untrustedHeader.Hash(), trustedHeader.LastBlockID.Hash) {
+			return nil, errors.Errorf("older header hash %X does not match trusted header's last block %X",
+				untrustedHeader.Hash(),
+				trustedHeader.LastBlockID.Hash)
+		}
```
cmwaters: `trustedHeader` remains the same throughout the entire for loop yet `untrustedHeader` changes iteratively - shouldn't trustedHeader be updated to the lastUntrustedHeader in order for the LastBlockID check to be correctly used?

lite2/client.go
melekes: Don't think so. c.trustedHeader should also exist in the database. If it's not there, it means something is horribly wrong

lite2/client.go
melekes: Good idea. Let's add this check

PR #4369, opened 2020-02-05
proto: add buf and protogen script

marbar3778:
- add buf with minimal changes
- add protogen script to easier generate proto files

* [ ] Wrote tests


marbar3778:
proto-breakage will fail here because master does not contain the changes nor the buf file

.circleci/config.yml
```diff
@@ -357,13 +373,10 @@ jobs:
   #         command: |
   #           set -x
   #           export PATH=~/.local/bin:$PATH
-
```
tessr: super nitpick but maybe add these whitespaces back in so this comment remains easier to read? with `#` in front if a linter complains

Makefile
```diff
@@ -30,35 +30,36 @@ install:
...
-### Protobuf
+###############################################################################
+###                                Protobuf                                 ###
+###############################################################################
 
-protoc_all: protoc_libs protoc_merkle protoc_abci protoc_grpc protoc_proto3types
+proto-all: proto-gen proto-lint proto-check-breaking
 
-%.pb.go: %.proto
-	## If you get the following error,
```
tessr: should we preserve this instruction somewhere? (I mostly found it annoying and never actually ran into this issue, but I assume it's here for a reason.)

UPGRADING.md
```diff
@@ -3,6 +3,14 @@
 This guide provides steps to be followed when you upgrade your applications to
 a newer version of Tendermint Core.
 
+## Unreleased
+
+<Overview>
+
+### Protobuf Changes
+
+When upgrading to version <version #> you will have to fetch the third_party directory along with the updated proto files.
```
tessr:
```suggestion
When upgrading to version <version #> you will have to fetch the `third_party` directory along with the updated proto files.
```

tessr: also, when you say "fetch," you literally mean getting them from the same repo as the proto files, right? just want to make sure we don't need to include more instructions.

tessr:
Nice. I'm very excited to try this out. Do you want to also write down the considerations re: making this part of a minor release?

UPGRADING.md
marbar3778: yes that is correct. I can link to rust-abci pr I have to do this for more context if you think it would help.

Makefile
marbar3778: left it where it is for now.

tessr:
Woo hoo, I just rebased off this branch and used it to generate protos for state sync! 

One note: it looks like this is regressing to an older version of the proto package (see https://github.com/tendermint/tendermint/commit/c53b8bdc43288b306bf947fff247313878978b7d#diff-e41225101b540b7e08bbbcddb8f8ffdcR40). Is that something to do with buf?

marbar3778:
what version of protoc do you have locally? i didnt get these changes on my machine so makes me think we have different versions locally

tessr:
Ah, I see. I have libprotoc 3.10.1; I can update and try again. (But I also thought these changes were going to standardize our proto gen environment? Maybe I misunderstand how/what Buf is.)

marbar3778:
we do have the option to do this. https://buf.build/docs/build-overview & https://buf.build/docs/build-images but seems you still require protoc to be used for local generation of stubs

tessr:
Hmm. I updated my proto compiler to 3.11.3 but it produced the same code as the compiler at 3.10.1. So I think there's something else going on here, too

marbar3778:
I think this would be fine to merge for a minor release as it is not changing the logic in which something is done. It is only changing if a project fetches the proto files. @melekes thoughts?

melekes:
> I think this would be fine to merge for a minor release as it is not changing the logic in which something is done. It is only changing if a project fetches the proto files. @melekes thoughts?

So it will only break ABCI servers in other languages, correct? I guess it's fine since we need it.

.circleci/config.yml
```diff
@@ -14,6 +14,9 @@ executors:
       - image: tendermintdev/docker-website-deployment
     environment:
       AWS_REGION: us-east-1
+  protoc:
+    docker:
+      - image: tendermintdev/docker-protoc
```
jackzampolin: Is there a more generic docker image to use here? Also adding a hash or version here would be helpful.

marbar3778: not sure what you mean by more generic. this is the simplest way to get buf into ci instead of installing it wach time

melekes: I think @jackzampolin was asking if there's an image supported by buf team.

Makefile
```diff
@@ -12,8 +11,9 @@ all: check build test install
 include tools.mk
 include tests.mk
 
-########################################
-### Build Tendermint
+###############################################################################
+###                                Build Tendermint                              ###
```
melekes:
```suggestion
###                                Build Tendermint                        ###
```

Makefile
```diff
@@ -176,8 +180,9 @@ localnet-start: localnet-stop build-docker-localnode
...
 	docker-compose down
 
-###########################################################
-### Remote full-nodes (sentry) using terraform and ansible
+###############################################################################
+###          Remote full-nodes (sentry) using terraform and ansible         ###
+###############################################################################
 
 # Server management
 sentry-start:
```
melekes: this is unsupported imo. not sure if it's working now. should we consider removing?

.circleci/config.yml
marbar3778: testing one

marbar3778: the issue is that the image is barebones and it seems buf requires go if you are using go extensions. reverted the change for now, but will look into it

Makefile
marbar3778: maybe we don't do it within this PR just to keep the scope narrow, but we can have a follow-up issue and discussion about what and IF we want to support these kind of commands.

marbar3778:
@melekes you good to merge as is?

melekes:
> you good to merge as is?

proto-breakage task has failed:

```
abci/types/types.proto@against:7:8:github.com/gogo/protobuf/gogoproto/gogo.proto: does not exist
crypto/merkle/merkle.proto@against:7:8:github.com/gogo/protobuf/gogoproto/gogo.proto: does not exist
libs/kv/types.proto@against:5:8:github.com/gogo/protobuf/gogoproto/gogo.proto: does not exist
rpc/grpc/types.proto@against:5:8:github.com/gogo/protobuf/gogoproto/gogo.proto: does not exist
Makefile:51: recipe for target 'proto-check-breaking' failed
```

marbar3778:
> proto-breakage task has failed:


https://github.com/tendermint/tendermint/pull/4369#issuecomment-582356692

tools.mk
```diff
@@ -65,6 +71,27 @@ $(PROTOBUF):
...
 	@go get github.com/gogo/protobuf/protoc-gen-gogo@v1.3.1
 
+buf: protoc-gen-buf-check-breaking protoc-gen-buf-check-lint
+	@echo "Installing buf..."
+	@curl -sSL \
+    "https://github.com/bufbuild/buf/releases/download/v${BUF_VERSION}/buf-${UNAME_S}-${UNAME_M}" \
+    -o "${BIN}/buf" && \
+	chmod +x "${BIN}/buf"
+
+protoc-gen-buf-check-breaking:
```
bufdev: FYI @marbar3778 there's no need for you to install `protoc-gen-buf-check-breaking` or `protoc-gen-buf-check-lint` for your usage of `buf check lint` and `buf check breaking` above - these are for if you want to do i.e. `protoc --buf-check-breaking_out=.`, which is primarily for use cases such as Bazel.

tools.mk
```diff
@@ -65,6 +71,27 @@ $(PROTOBUF):
 	@echo "Get GoGo Protobuf"
 	@go get github.com/gogo/protobuf/protoc-gen-gogo@v1.3.1
 
+buf: protoc-gen-buf-check-breaking protoc-gen-buf-check-lint
+	@echo "Installing buf..."
+	@curl -sSL \
+    "https://github.com/bufbuild/buf/releases/download/v${BUF_VERSION}/buf-${UNAME_S}-${UNAME_M}" \
```
bufdev: @marbar3778 nit: I would make the variables in Makefile style i.e. `$(UNAME_S)` instead of `${UNAME_S}`

crypto/merkle/merkle.proto
```diff
@@ -4,7 +4,7 @@ option go_package = "github.com/tendermint/tendermint/crypto/merkle";
 
 // For more information on gogo.proto, see:
 // https://github.com/gogo/protobuf/blob/master/extensions.md
-import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+import "third_party/proto/gogoproto/gogo.proto";
```
bufdev: I would include `third_party/proto` as a root (and also include for `protoc` with `-I third_party/proto`), and remove that part from these imports @marbar3778

buf.yaml
```diff
@@ -0,0 +1,16 @@
...
+    - MINIMAL
+    - FILE_LOWER_SNAKE_CASE
+    - UNARY_RPC
+  except:
+    - PACKAGE_DIRECTORY_MATCH
+  ignore:
+    - third_party
+breaking:
+  use:
+    - FILE
```
bufdev: @marbar3778 `FILE` implies `PACKAGE`, so there's no need for both. See https://buf.build/docs/breaking-checkers#choosing-a-category

crypto/merkle/merkle.proto
marbar3778: added `-I third_party/proto` but not sure what you mean by add it to root. If in the buf.yml file then it overlaps with `.`

PR #4374, opened 2020-02-06
lite2: validate TrustOptions, add NewClientFromTrustedStore

melekes:
and make maxRetryAttempts an option.

* [ ]~~ Referenced an issue explaining the need for the change~~
* [x] Wrote tests

### DIFF
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -43,6 +43,20 @@ type TrustOptions struct {
 	Hash   []byte
 }
 
+// ValidateBasic performs basic validation.
+func (opts TrustOptions) ValidateBasic() error {
+	if opts.Period <= 0 {
+		return errors.New("negative or zero period")
+	}
+	if opts.Height <= 0 {
+		return errors.New("negative or zero height")
+	}
+	if len(opts.Hash) == 0 {
+		return errors.New("empty hash")
+	}
+	return nil
+}
+
 type mode byte
 
 const (
@@ -172,6 +186,10 @@ func NewClient(
 	witnesses []provider.Provider,
 	trustedStore store.Store,
 	options ...Option) (*Client, error) {
+
+	if err := trustOptions.ValidateBasic(); err != nil {
+		return nil, errors.Wrap(err, "invalid TrustOptions")
+	}
 
 	c := &Client{
 		chainID:                            chainID,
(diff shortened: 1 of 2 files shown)


lite2/client.go
```diff
@@ -43,6 +43,20 @@ type TrustOptions struct {
...
 
+// ValidateBasic performs basic validation.
+func (opts TrustOptions) ValidateBasic() error {
+	if opts.Period <= 0 {
+		return errors.New("negative or zero period")
+	}
+	if opts.Height <= 0 {
+		return errors.New("negative or zero height")
+	}
+	if len(opts.Hash) == 0 {
```
```suggestion
	if len(opts.Hash) != tmhash.Size {
```

lite2/client.go
```diff
@@ -128,6 +150,7 @@ type Client struct {
 	trustingPeriod   time.Duration // see TrustOptions.Period
 	verificationMode mode
 	trustLevel       tmmath.Fraction
+	maxRetryAttempts int // see MaxRetryAttempts option
```
marbar3778: should we start being more strict on int types? this may be better suited as an int32 or uint32 possibly, maybe even smaller.

melekes: good idea

lite2/client_test.go
```diff
@@ -987,3 +987,49 @@ func TestProvider_TrustedHeaderFetchesMissingHeader(t *testing.T) {
...
+		chainID,
+		1*time.Hour,
+		primary,
+		[]provider.Provider{primary},
+		db,
+	)
+	require.NoError(t, err)
+
+	// 3) Check header exists through the lite clients eyes
+	h, err := c.TrustedHeader(1, bTime.Add(1*time.Second))
```
melekes: how do you know it wasn't fetched from primary?

cmwaters: does the lite client automatically update from the primary on start up?

melekes: no, but TrustedHeader can fetch missing headers, so how do you now they are not coming from primary?

cmwaters: but trusted header would return an error if the database was empty: "no headers exist"

melekes: https://github.com/tendermint/tendermint/pull/4374/commits/3bc2b1c5e20078e6b7a00c14dee6e720fbbc3ab9

melekes: You're right. Anyway, it never a bad idea to double check 😄

PR #4378, opened 2020-02-07
lite2: return if there are no headers in RemoveNoLongerTrustedHeaders

melekes:
* [ ] Wrote tests

### DIFF
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -965,11 +965,17 @@ func (c *Client) RemoveNoLongerTrustedHeaders(now time.Time) {
 		c.logger.Error("can't get first trusted height", "err", err)
 		return
 	}
+	if oldestHeight == -1 {
+		return
+	}
 
 	// 2) Get the latest height.
 	latestHeight, err := c.LastTrustedHeight()
 	if err != nil {
 		c.logger.Error("can't get last trusted height", "err", err)
+		return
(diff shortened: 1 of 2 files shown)


lite2/client.go
```diff
@@ -965,13 +965,19 @@ func (c *Client) RemoveNoLongerTrustedHeaders(now time.Time) {
 		c.logger.Error("can't get first trusted height", "err", err)
```
cmwaters: maybe update `c.trustedStore.FirstSignedHeaderHeight()` to `c.FirstTrustedHeight()`

lite2/client.go
```diff
@@ -965,13 +965,19 @@ func (c *Client) RemoveNoLongerTrustedHeaders(now time.Time) {
...
 
 	// 2) Get the latest height.
 	latestHeight, err := c.LastTrustedHeight()
 	if err != nil {
 		c.logger.Error("can't get last trusted height", "err", err)
 		return
 	}
+	if latestHeight == -1 { // no headers yet => wait
+		return
+	}
```
cmwaters: Are both if statements necessary? Is there a circumstance where if there are no headers that asking for the first header doesn't return -1

melekes: no, but it's okay to have both I think

PR #4380, opened 2020-02-07
lite2: manage witness dropout

### LINKED ISSUE #4346: lite2: remove a witness if it misses N headers
We'll be cross-checking each header with all witnesses. If some witness refuses to give us a header, we should increment some counter, so when the number reaches say 50 failed attempts (meaning a witness did not give us 50 headers in a row), we remove it from the list completely and forget about it.

cmwaters:
Closes #4346

As opposed to what was requested, the current implementation just removes the witness if there is no response after the `maxRetryAttempts`. We can discuss further implementing a counter for each witness and removing them after a set maxCounter is reached but this would be rather verbose. 
 

* [x] Wrote tests


lite2/errors.go
```diff
@@ -2,6 +2,7 @@ package lite
 
 import (
 	"fmt"
+	"github.com/tendermint/tendermint/lite2/provider"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

lite2/client.go
```diff
@@ -916,11 +916,25 @@ func (c *Client) compareNewHeaderWithWitnesses(h *types.SignedHeader) error {
...
+	for i := 0; i < len(c.witnesses); i++ {
+		witness := c.witnesses[i]
 
 		// 2. Fetch the header.
-		altH, err := witness.SignedHeader(h.Height)
-		if err != nil {
+		altH, err := c.signedHeaderFromWitness(witness, h.Height)
+		switch err.(type) {
+		case nil:
+			break
```
melekes: why do we break? shouldn't we continue and compare hashes?

lite2/client.go
```diff
@@ -916,11 +916,25 @@ func (c *Client) compareNewHeaderWithWitnesses(h *types.SignedHeader) error {
...
+		case ErrWitnessNonResponsive: // remove witness
```
melekes: not clear to me why we need this block? if witness is not responsive, we should just remove it and continue with the rest of the witnesses

lite2/client.go
```diff
@@ -1110,6 +1124,26 @@ func (c *Client) validatorSetFromPrimary(height int64) (*types.ValidatorSet, err
 	return c.validatorSetFromPrimary(height)
 }
 
+func (c *Client) signedHeaderFromWitness(witness provider.Provider, height int64) (*types.SignedHeader, error) {
+	for attempt := uint16(1); attempt <= c.maxRetryAttempts; attempt++ {
```
melekes: can we extract this code into a function (smth like `signedHeaderFrom(p Provider, ...)`)?

cmwaters: Yes, we can try. You mean to also merge it with the signedHeaderFromPrimary. I thought about earlier but just decided to implement it like this first

lite2/client.go
cmwaters: Do we need to make the distinction between errors or regardless of the errors should we then remove it and continue

lite2/client.go
melekes: > You mean to also merge it with the signedHeaderFromPrimary

yes

cmwaters: Okay yes I can do that

melekes:
> no response after the maxRetryAttempts

this will drastically increase the new header verification time. not sure we want that

lite2/client.go
melekes: regardless

cmwaters:
> > no response after the maxRetryAttempts
> 
> this will drastically increase the new header verification time. not sure we want that

This depends on what we set the maxRetryAttempts at. But I agree. If it is 10 than that is a couple of minutes per witness. I can only suggest that we don't retry but move on to the next witness immediately.

lite2/client.go
cmwaters: In hindsight, I don't think that these two should be merged because they are from distinct data structures and they have different processes surrounding how non responsiveness should be dealt with

PR #4386, opened 2020-02-11
proto: minor linting to proto files

### LINKED ISSUE #4381: Broken spec links
Lots of previous documentation made reference to the specs in this repo. Since they've been moved to the spec repo and deleted here, all those links are now broken. They can be found with `ack "docs/spec"`

We should either fix all those links to point to the spec repo and/or add back the old spec directory structure with place holder files telling folks to redirect. Note there's probably other documentation out there that makes reference to those old spec links, so doing both for now might not be a bad idea.

marbar3778:
- proto linting
this came about after i fetched the proto files for use in the sdk.

Co-authored-by: Anton Kaliaev <anton.kalyaev@gmail.com>

* [ ] Wrote tests

### DIFF
--- a/abci/types/types.proto
+++ b/abci/types/types.proto
@@ -46,8 +46,7 @@ message RequestEcho {
   string message = 1;
 }
 
-message RequestFlush {
-}
+message RequestFlush {}
 
 message RequestInfo {
   string version = 1;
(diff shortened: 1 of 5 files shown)


melekes:
there's a duplicate commit: docs: fix spec links (#4384). could you please rebase and force push?

PR #4390, opened 2020-02-11
autofile: resolve relative paths

### LINKED ISSUE #2649: Make autofile resolve absolute paths
When sharing a runtime with Tendermint and in testing it can be convenient to start Tendermint in a test directory. However after initialising a subsequent `os.Chdir` can make autofile fail if it was initialised with a relative path. Although this is somewhat 'as designed' I can't see a time when it wouldn't be preferable to have it resolve the path at the time the `Group` or `Autofile` is created: https://github.com/tendermint/tendermint/blob/master/libs/autofile/group.go#L78

`filepath.Abs()` would do the trick.

This comes up with the WAL in particular, where if you change directory in an unrelated area you get a consensus failure because the wal cannot be written. It would be useful not to have to sanitise tests of `os.Chdir`

erikgrinaker:
Fixes #2649. Does this warrant a changelog entry?

* [x] Wrote tests

### DIFF
--- a/libs/autofile/autofile_test.go
+++ b/libs/autofile/autofile_test.go
@@ -30,6 +30,7 @@ func TestSIGHUP(t *testing.T) {
 	name := "sighup_test"
 	af, err := OpenAutoFile(name)
 	require.NoError(t, err)
+	require.True(t, path.IsAbs(af.Path))
 
 	// Write to the file.
 	_, err = af.Write([]byte("Line 1\n"))
--- a/libs/autofile/autofile_test.go
+++ b/libs/autofile/autofile_test.go
@@ -3,24 +3,31 @@ package autofile
 import (
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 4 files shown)


melekes:
> Does this warrant a changelog entry?

Of course!

libs/autofile/autofile_test.go
```diff
@@ -3,26 +3,34 @@ package autofile
 import (
 	"io/ioutil"
 	"os"
+	"path"
```
melekes: better to use filepath - it's more cross-platform

erikgrinaker:
> > Does this warrant a changelog entry?
> 
> Of course!

Looks like this is only used for the WAL, and the WAL path is explicitly resolved on startup anyway, so this is sort of moot. May as well merge this, but don't see a point in adding it to the changelog, since there should not be any user-visible changes.

PR #4397, opened 2020-02-13
adr: light client implementation

### LINKED ISSUE #2133: Create an ADR for lite/provider

melekes:
📝 [RENDERED](https://github.com/tendermint/tendermint/blob/4faa12f5978590d7fe50e7c061e06f335e941186/docs/architecture/adr-046-light-client-implementation.md)

Closes #2133

* [ ] ~~Wrote tests~~

### DIFF
--- a/docs/architecture/README.md
+++ b/docs/architecture/README.md
@@ -64,5 +64,8 @@ Note the context/background should be written in the present tense.
 - [ADR-039-Peer-Behaviour](./adr-039-peer-behaviour.md)
 - [ADR-041-Proposer-Selection-via-ABCI](./adr-041-proposer-selection-via-abci.md)
 - [ADR-043-Blockchain-RiRi-Org](./adr-043-blockchain-riri-org.md)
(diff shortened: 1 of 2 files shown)


melekes:
> Do we need to add anything about the sequential or bisection methods?

don't see a need to duplicate https://pkg.go.dev/github.com/tendermint/tendermint/lite2?tab=doc#SequentialVerification

cmwaters:
Okay, perhaps reference it? Or I guess it's fine if you don’t

melekes:
> Okay, perhaps reference it?

as per our discussion, I don't think ADR should be a reference to API, but merely an explanation why certain choices were made (like having >= 1 witness, etc.)

cmwaters:
Ah okay, I guess I’m still understanding the exact purpose of adr‘s and the level of context that is required.

PR #4400, opened 2020-02-13
lite: modified bisection to loop

### LINKED ISSUE #4329: lite2: align with the newer version of the spec
once https://github.com/tendermint/spec/pull/61 is merged.

cmwaters:
closes #4329

As opposed to using recursion to implement the bisection method of verifying a header, which could have problems with memory allocation (especially for smaller devices), the bisection algorithm now uses a for loop. 

* [ ] Wrote tests


lite2/client.go
```diff
@@ -740,72 +740,66 @@ func (c *Client) sequence(newHeader *types.SignedHeader, newVals *types.Validato
...
 }
 
-// see VerifyHeader
+// see VerifyHeader  NB: although newVals is entered as input actually updates client database with newNextVals
 func (c *Client) bisection(
-	lastHeader *types.SignedHeader,
-	lastVals *types.ValidatorSet,
-	newHeader *types.SignedHeader,
-	newVals *types.ValidatorSet,
+	lastHeader *types.SignedHeader, // height h
```
melekes: let's rename to `trustedHeader` here too. same for vals

lite2/client.go
```diff
@@ -740,72 +740,66 @@ func (c *Client) sequence(newHeader *types.SignedHeader, newVals *types.Validato
 	return Verify(c.chainID, c.trustedHeader, c.trustedNextVals, newHeader, newVals, c.trustingPeriod, now, c.trustLevel)
 }
 
-// see VerifyHeader
+// see VerifyHeader  NB: although newVals is entered as input actually updates client database with newNextVals
```
melekes: NB?

lite2/client.go
```diff
@@ -740,72 +740,66 @@ func (c *Client) sequence(newHeader *types.SignedHeader, newVals *types.Validato
...
-			return errors.Wrapf(err, "failed to update trusted state #%d", pivot)
-		}
+			//update the upper bound to the untrustedHeader
+			interimHeader, interimVals = newHeader, newVals
+		case ErrNewValSetCantBeTrusted:
 
-		err = c.bisection(pivotHeader, nextVals, newHeader, newVals, now)
-		if err != nil {
-			return errors.Wrapf(err, "bisection of #%d and #%d", pivot, newHeader.Height)
+			// find the new pivot height
```
melekes: comment does not bring any value imo

lite2/client.go
```diff
@@ -740,72 +740,66 @@ func (c *Client) sequence(newHeader *types.SignedHeader, newVals *types.Validato
...
+			err = c.updateTrustedHeaderAndNextVals(lastHeader, lastNextVals)
+			if err != nil {
+				return err
+			}
 
-		err = c.updateTrustedHeaderAndVals(pivotHeader, nextVals)
-		if err != nil {
-			return errors.Wrapf(err, "failed to update trusted state #%d", pivot)
-		}
+			//update the upper bound to the untrustedHeader
```
```suggestion
			// Update the upper bound to the newHeader and repeat (if necessary).
```

lite2/client.go
```diff
@@ -740,72 +740,66 @@ func (c *Client) sequence(newHeader *types.SignedHeader, newVals *types.Validato
...
-		if err != nil {
-			return errors.Wrapf(err, "failed to obtain the vals #%d", pivot+1)
-		}
-		if !bytes.Equal(pivotHeader.NextValidatorsHash, nextVals.Hash()) {
-			return errors.Errorf("expected next validator's hash %X, but got %X (height #%d)",
-				pivotHeader.NextValidatorsHash,
-				nextVals.Hash(),
-				pivot)
-		}
+			// update the trusted store with trusted header and next vals
```
melekes: comment does not bring any value imo

lite2/client.go
```diff
@@ -740,72 +740,66 @@ func (c *Client) sequence(newHeader *types.SignedHeader, newVals *types.Validato
...
+		c.logger.Debug("Verify newHeader against lastHeader",
+			"lastHeight", lastHeader.Height,
+			"lastHash", lastHeader.Hash(),
+			"newHeight", newHeader.Height,
+			"newHash", newHeader.Hash())
+		err := Verify(c.chainID, lastHeader, lastNextVals, interimHeader, interimVals, c.trustingPeriod, now, c.trustLevel)
+		switch err.(type) {
+		case nil:
+
+			// update the lower bound to the previous upper bound
```
```suggestion
			// Update the lower bound to the previous upper bound.
```

lite2/verifier.go
```diff
@@ -49,7 +49,7 @@ func Verify(
 	}
 
 	if h2.Height == h1.Height+1 {
-		if !bytes.Equal(h2.ValidatorsHash, h1NextVals.Hash()) {
+		if !bytes.Equal(h2.ValidatorsHash, h1NextVals.Hash()) { //h1.nextValidatorHash
```
melekes: ???

cmwaters:
> looks good from the first glance 💛
> 
> have you seen the spec https://github.com/tendermint/spec/blob/zm_non-recursive-verification/spec/consensus/light-client/non-recursive-verification.md btw?

Yes, I've had a look at it. The code there seems a little bit bloated but I think my implementation matches it

lite2/client.go
cmwaters: Note well / Nota Bene

lite2/verifier.go
cmwaters: In the spec they use h1.nextValidatorHash instead of h1NextVals.Hash. I have actually made the changes in another pull request

lite2/client.go
```diff
@@ -708,72 +708,63 @@ func (c *Client) sequence(
...
-				nextVals.Hash(),
-				pivot)
-		}
+	for trustedHeader.Height < newHeader.Height {
+		c.logger.Debug("Verify newHeader against trustedHeader",
+			"lastHeight", trustedHeader.Height,
+			"lastHash", trustedHeader.Hash(),
+			"newHeight", newHeader.Height,
+			"newHash", newHeader.Hash())
+		err := Verify(c.chainID, trustedHeader, trustedNextVals, interimHeader, interimVals, c.trustingPeriod, now, c.trustLevel)
```
golangcibot: line is 123 characters (from `lll`)

lite2/client.go
```diff
@@ -708,72 +708,64 @@ func (c *Client) sequence(
...
+				return err
+			}
 
-		err = c.bisection(pivotHeader, nextVals, newHeader, newVals, now)
-		if err != nil {
-			return errors.Wrapf(err, "bisection of #%d and #%d", pivot, newHeader.Height)
+			// Update the upper bound to the untrustedHeader
+			interimHeader, interimVals = newHeader, newVals
+		case ErrNewValSetCantBeTrusted:
+			pivotHeight := (interimHeader.Height + trustedHeader.Height) / 2
```
melekes: what if interimHeader.Height = 2 and trustedHeader.Height = 1

=> pivotHeight = 1 => Infinite loop?

1) should we use `math.Ceil` function
2) should we check that headers are adjacent and exit if header cannot be trusted? I think there should be another condition (`if interimHeader.Height == trustedHeader.Height+1`) besides trustedHeader.Height < newHeader.Height for bisection to exit.

cmwaters: Yes I believe you are correct. I will need to change the code and write tests accordingly

cmwaters: I actually think the changes to the other pr #4412 with the addition of the generalised `Verify` function will solve this

melekes: Even if that is true, we still need to write more tests since current suite clearly did not cover ^ use-case.

cmwaters: I will write some tests regarding the non recursive bisection. Can we first review and merge #4412 and then I will adjust the bisection method accordingly

melekes: https://github.com/tendermint/tendermint/pull/4412#pullrequestreview-359887880 my review

lite2/client.go
melekes: let's stick with the commonly accepted codetags (https://en.wikipedia.org/wiki/Comment_(computer_programming)#Tags, https://docs.microsoft.com/en-us/visualstudio/ide/using-the-task-list?view=vs-2015&redirectedfrom=MSDN#customTokens)

melekes:
I am going to write tests now.

lite2/client.go
cmwaters: sure, thanks for pointing them out.

lite2/client_test.go
```diff
@@ -138,39 +141,42 @@ func TestClient_SequentialVerification(t *testing.T) {
...
-				tc.otherHeaders,
-				tc.vals,
-			)},
-			dbs.New(dbm.NewMemDB(), chainID),
-			SequentialVerification(),
-		)
+				trustOptions,
+				mockp.New(
+					chainID,
+					tc.otherHeaders,
```
golangcibot: Using the variable on range scope `tc` in function literal (from `scopelint`)

lite2/client_test.go
```diff
@@ -138,39 +141,42 @@ func TestClient_SequentialVerification(t *testing.T) {
...
+					tc.vals,
```
golangcibot: Using the variable on range scope `tc` in function literal (from `scopelint`)

lite2/client_test.go
```diff
@@ -138,39 +141,42 @@ func TestClient_SequentialVerification(t *testing.T) {
...
+				),
+				[]provider.Provider{mockp.New(
+					chainID,
+					tc.otherHeaders,
```
golangcibot: Using the variable on range scope `tc` in function literal (from `scopelint`)

PR #4403, opened 2020-02-13
lite2: modified sequence method to match bisection

cmwaters:
**Minor Edit**

Currently the sequence function always starts from the `trustedHeader` and `trustedNextVals` stored in the lite client. Whereas the bisection one allows the method to be started from any combination of header and validator set. I opened up the sequence verification method to do the same

### DIFF
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -698,8 +698,8 @@ func (c *Client) cleanup(stopHeight int64) error {
 
 // see VerifyHeader
 func (c *Client) sequence(
-	lastHeader *types.SignedHeader,
-	lastNextVals *types.ValidatorSet,
+	trustedHeader *types.SignedHeader,
+	trustedNextVals *types.ValidatorSet,
 	newHeader *types.SignedHeader,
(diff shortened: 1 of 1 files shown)


lite2/client.go
```diff
@@ -697,14 +697,19 @@ func (c *Client) cleanup(stopHeight int64) error {
 }
 
 // see VerifyHeader
-func (c *Client) sequence(newHeader *types.SignedHeader, newVals *types.ValidatorSet, now time.Time) error {
+func (c *Client) sequence(
+	lastHeader *types.SignedHeader,
```
```suggestion
	trustedHeader *types.SignedHeader,
```

lite2/client.go
```diff
@@ -697,14 +697,19 @@ func (c *Client) cleanup(stopHeight int64) error {
...
+	lastNextVals *types.ValidatorSet,
```
```suggestion
	trustedNextVals *types.ValidatorSet,
```

cmwaters:
> so the main change here seems to be aligning sequence and bisection func declarations, correct? there's no "breaking" changes?

yes correct

PR #4407, opened 2020-02-14
Docs staging

fadeev:
Updated to the latest version of the theme.

`<span v-pre>` is added to the following files to prevent docs build errors.

https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-051-double-signing-risk-reduction.md

https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-052-tendermint-mode.md

Also removed links to Riot.


docs/.vuepress/config.js
```diff
@@ -19,12 +19,6 @@ module.exports = {
     gutter: {
       title: "Help & Support",
       editLink: true,
-      chat: {
```
marbar3778: can we put in the discord instead of removing this?

docs/.vuepress/config.js
```diff
@@ -95,10 +90,6 @@ module.exports = {
             {
               title: "Forum",
               url: "https://forum.cosmos.network/c/tendermint"
-            },
-            {
-              title: "Chat",
```
marbar3778: we should have some chat, this is quite bad that we are getting rid of a way for people to talk to us

marbar3778:
the fix to this would be when this issue is closed: https://github.com/vuejs/vuepress/issues/1558

marbar3778:
would like to have some chat and not remove it entirely. we are closing off a gateway into the ecosystem by doing

docs/.vuepress/config.js
fadeev: Added link to Discord: https://discordapp.com/channels/669268347736686612.

Issue #2883, opened 2018-11-19
Don't interface encode mempool messages


ValarDragon:
Currently we interface encode mempool messages: https://github.com/tendermint/tendermint/blob/master/mempool/reactor.go#L177

This is an extra 4 bytes on every tx gossip, which is unnecessary and not really negligible. Were also not really getting anything by doing this, since there is only one message that goes to/from the mempool reactor, and in the future it will still be a small number of things. (This is also a good candidate for a protobuf oneof, the 4 byte prefix is a bit silly)

Liamsi:
>  and in the future it will still be a small number of things. 

How will it be more than "one thing"? Can't find anything about this.

ValarDragon:
It could potentially be more with future mempool models. Mostly meant even if that is the case, it will be a small constant, nothing on the order of `256^4` options that we are allowing to choose between.

PR #4409, opened 2020-02-14
 proto: move mempool to proto

marbar3778:
ref #2883

* [ ] Wrote tests


mempool/reactor.go
```diff
@@ -233,9 +226,9 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
 
 		// ensure peer hasn't already sent us this tx
 		if _, ok := memTx.senders.Load(peerID); !ok {
-			// send memTx
-			msg := &TxMessage{Tx: memTx.tx}
-			success := peer.Send(MempoolChannel, cdc.MustMarshalBinaryBare(msg))
+			msg := gogotypes.BytesValue{Value: []byte(memTx.tx)}
+			bz, _ := msg.Marshal()
```
marbar3778: should handle error here

mempool/reactor.go
```diff
@@ -19,7 +17,7 @@ import (
 const (
 	MempoolChannel = byte(0x30)
 
-	aminoOverheadForTxMessage = 8
+	protoOverheadForTxMessage = 4
```
Liamsi: Interesting, how is this even shorter? Ah, nvm, I see, it's because the prefix bytes that are gone now. Forgot that this was an interface.

mempool/clist_mempool_test.go
```diff
@@ -121,11 +121,11 @@ func TestReapMaxBytesMaxGas(t *testing.T) {
...
-		{20, 22, 10, 1},
-		{20, 220, -1, 10},
-		{20, 220, 5, 5},
-		{20, 220, 10, 10},
-		{20, 220, 15, 10},
+		{20, 20, 10, 1},
+		{20, 200, -1, 10},
+		{20, 200, 5, 5},
+		{20, 200, 10, 10},
+		{20, 200, 15, 10},
```
Liamsi: The changes in tests are a bit confusing (one `maxBytes` changes 2 bytes and the others 20).

mempool/clist_mempool_test.go
```diff
@@ -458,9 +450,9 @@ func TestMempoolMaxMsgSize(t *testing.T) {
 
 		tx := tmrand.Bytes(testCase.len)
 		err := mempl.CheckTx(tx, nil, TxInfo{})
-		msg := &TxMessage{tx}
-		encoded := cdc.MustMarshalBinaryBare(msg)
-		require.Equal(t, len(encoded), txMessageSize(tx), caseString)
+		bz := gogotypes.BytesValue{Value: tx}
```
Liamsi: Nit on consistency: `bz` usually is used for the resulting bytes for encoding (e.g. what is called `encoded` below).

rpc/grpc/types.pb.go
```diff
@@ -7,17 +7,16 @@ import (
...
 	_ "github.com/gogo/protobuf/gogoproto"
 	proto "github.com/gogo/protobuf/proto"
 	golang_proto "github.com/golang/protobuf/proto"
 	types "github.com/tendermint/tendermint/abci/types"
 	grpc "google.golang.org/grpc"
 	codes "google.golang.org/grpc/codes"
 	status "google.golang.org/grpc/status"
+	io "io"
+	math "math"
+	math_bits "math/bits"
```
Liamsi: Looks like these were moved by accident?

mempool/reactor.go
```diff
@@ -257,21 +253,18 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
...
 	maxMsgSize := calcMaxMsgSize(memR.config.MaxTxBytes)
 	if l := len(bz); l > maxMsgSize {
-		return msg, ErrTxTooLarge{maxMsgSize, l}
+		return TxMessage{}, ErrTxTooLarge{maxMsgSize, l}
+	}
+
+	msg := gogotypes.BytesValue{}
+	err := msg.Unmarshal(bz)
+	txMsg := TxMessage{
+		Tx: types.Tx(msg.Value),
```
Liamsi: Looks like `gogotypes.BytesValue` and `TxMessage` are basically equivalent. Only difference is the field name. I'm wondering if this cast/conversion to `Tx` copies the value and if yes, if this is desired.

mempool/reactor.go
```diff
@@ -233,9 +226,12 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
 
 		// ensure peer hasn't already sent us this tx
 		if _, ok := memTx.senders.Load(peerID); !ok {
-			// send memTx
-			msg := &TxMessage{Tx: memTx.tx}
-			success := peer.Send(MempoolChannel, cdc.MustMarshalBinaryBare(msg))
+			msg := gogotypes.BytesValue{Value: []byte(memTx.tx)}
+			bz, err := msg.Marshal()
+			if err != nil {
+				// TODO: question bubble up the error or ignore it?
```
Liamsi: The previous behaviour was to panic in case of an error `MustMarshalBinaryBare`.

mempool/clist_mempool_test.go
```diff
@@ -458,9 +450,9 @@ func TestMempoolMaxMsgSize(t *testing.T) {
...
+		encoded, _ := bz.Marshal() // ignore error
```
Liamsi: better: `require.NoError` here instead.

mempool/reactor.go
marbar3778: I was thinking about this as well but my reason for not doing this is that if we were to add a new type then we would like to add logic to accompany it, this would result in more code needing to be written so it would not be hard to do it then, but also we do have a refactor planned for the mempool so I cant see new functionality being added to this version.

rpc/grpc/types.pb.go
marbar3778: this happened automatically, I tried to regenerate but I am getting it consistently

mempool/clist_mempool_test.go
marbar3778: My understanding of this is that since {20, 20, 10, 1}, is expecting one tx then the bytes will be, but for the others, since they are expecting 10txs then the maxbytes would be 200, but seems the tests previously was overcompensating by allowing 220 because I believe it would be 210 bytes with 10txs, not 220

mempool/reactor.go
Liamsi: Conversion doesn't do copies apparently (only changes the type not the represntation), so this shouldn't be a concern: https://golang.org/ref/spec#Conversions

mempool/mempool.go
```diff
@@ -100,19 +100,12 @@ type TxInfo struct {
...
-		// NOTE: fieldNum = 1 as types.Block.Data contains Txs []Tx as first field.
-		// If this field order ever changes this needs to updated here accordingly.
-		// NOTE: if some []Tx are encoded without a parenting struct, the
-		// fieldNum is also equal to 1.
-		aminoOverhead := types.ComputeAminoOverhead(tx, 1)
-		txSize := int64(len(tx)) + aminoOverhead
+		txSize := int64(len(tx))
 		if txSize > maxBytes {
-			return fmt.Errorf("tx size (including amino overhead) is too big: %d, max: %d",
+			return fmt.Errorf("tx size (including proto overhead) is too big: %d, max: %d",
```
```suggestion
			return fmt.Errorf("tx size is too big: %d, max: %d",
```

looks like there's no overhead. at least, you're no accounting for it in this function

should we use protoOverheadForTxMessage here?

mempool/reactor.go
```diff
@@ -163,20 +161,15 @@ func (memR *Reactor) Receive(chID byte, src p2p.Peer, msgBytes []byte) {
...
-			txInfo.SenderP2PID = src.ID()
-		}
-		err := memR.mempool.CheckTx(msg.Tx, nil, txInfo)
-		if err != nil {
-			memR.Logger.Info("Could not check tx", "tx", txID(msg.Tx), "err", err)
-		}
-		// broadcasting happens from go routines per peer
-	default:
-		memR.Logger.Error(fmt.Sprintf("Unknown message type %v", reflect.TypeOf(msg)))
+	txInfo := TxInfo{SenderID: memR.ids.GetForPeer(src)}
```
melekes: won't this crash without the switch if I send something other than `TxMessage` to mempool reactor?

mempool/reactor.go
```diff
@@ -233,9 +226,12 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
 
 		// ensure peer hasn't already sent us this tx
 		if _, ok := memTx.senders.Load(peerID); !ok {
-			// send memTx
-			msg := &TxMessage{Tx: memTx.tx}
-			success := peer.Send(MempoolChannel, cdc.MustMarshalBinaryBare(msg))
+			msg := gogotypes.BytesValue{Value: []byte(memTx.tx)}
+			bz, err := msg.Marshal()
+			if err != nil {
+				panic(err)
```
melekes: maybe just log the error? is this critical enough to crash Tendermint?

mempool/reactor.go
```diff
@@ -257,21 +253,18 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
...
+func (memR *Reactor) decodeMsg(bz []byte) (TxMessage, error) {
 	maxMsgSize := calcMaxMsgSize(memR.config.MaxTxBytes)
 	if l := len(bz); l > maxMsgSize {
-		return msg, ErrTxTooLarge{maxMsgSize, l}
+		return TxMessage{}, ErrTxTooLarge{maxMsgSize, l}
+	}
+
+	msg := gogotypes.BytesValue{}
+	err := msg.Unmarshal(bz)
+	txMsg := TxMessage{
```
melekes: shouldn't we check for err before accessing any of its fields?

mempool/reactor.go
marbar3778: well now the msg is not an interface like it was before. Before we were returning a msg: https://github.com/tendermint/tendermint/pull/4409/files#diff-faa5920fe04b5bb0d81cec8ad247ccdcL268 which was an unneeded interface so if it is not able to decode the message then it would panic in the previous times. https://github.com/tendermint/tendermint/pull/4409/files#diff-faa5920fe04b5bb0d81cec8ad247ccdcL268

I can add a switch statement but since there is only 1 msg type then it seems a bit redundant

mempool/reactor.go
marbar3778: to keep with the previous behavior a panic would be needed. if we arent able to marshal the type then there could be a larger issue at play. I can add a log here on top of the panic

melekes: > I can add a log here on top of the panic

no need

mempool/reactor.go
melekes: > since there is only 1 msg type then it seems a bit redundant

Agree

marbar3778: also panic would accure here: https://github.com/tendermint/tendermint/pull/4409/files#diff-faa5920fe04b5bb0d81cec8ad247ccdcR156 if it was not of type txmessage

ebuchman:
I'm not sure this was a good idea since I think this now means we can't add more message types to the mempool in a backwards compatible way which is something we may want to do in the future. Why not use a proto Any encoding here or at least preserve some way to support multiple message types in the future?

Also is there really no more overhead from these messages, they're just the raw tx bytes? Doesn't proto encoding them still add some overhead that we need to account for? Not accounting for this overhead can cause very subtle bugs for proposers ...

marbar3778:
> Why not use a proto Any encoding here or at least preserve some way to support multiple message types in the future?

We can make this a oneof and only support one message for now. There was a issue to not interface encode messages so I went with the path of only encoding one message. Will make the change.

PR #4412, opened 2020-02-17
lite2: divide verify functions

### LINKED ISSUE #4398: lite2: divide verify functions proposal
The current `Verify` function in `lite2/verifier.go` uses two methods of differing security to verify a new header against an old header. The first checks, in the case that the headers are adjacent, that the nextValidatorHash of the trustedHeader (@ h) is equal to the validatorHash of the untrustedHeader (@ h + 1). The second method, in the case that the blocks are not adjacent, uses `VerifyCommitTrusting` to check that the validator set that signed the trustedHeader has at least a 1/3 overlap with the validator set that signed the untrusted Header. As these two methods corresponds to two distinct methods that the lite client uses: sequence and bisection, respectively, it might make sense to spit them into two distinct functions to avoid overhead and increase clarity.

cmwaters:
Closes #4398 

Refactored current `Verify` function to `VerifyViaVals` and `VerifyViaSigs`

* [x] Wrote tests

### DIFF
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -617,7 +617,7 @@ func (c *Client) VerifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
 	var err error
 	switch c.verificationMode {
 	case sequential:
-		err = c.sequence(newHeader, newVals, now)
+		err = c.sequence(c.trustedHeader, c.trustedNextVals, newHeader, newVals, now)
 	case skipping:
 		err = c.bisection(c.trustedHeader, c.trustedNextVals, newHeader, newVals, now)
 	default:
@@ -697,25 +697,30 @@ func (c *Client) cleanup(stopHeight int64) error {
 }
 
 // see VerifyHeader
-func (c *Client) sequence(newHeader *types.SignedHeader, newVals *types.ValidatorSet, now time.Time) error {
+func (c *Client) sequence(
+	trustedHeader *types.SignedHeader,
+	trustedNextVals *types.ValidatorSet,
+	newHeader *types.SignedHeader,
+	newVals *types.ValidatorSet,
+	now time.Time) error {
 	// 1) Verify any intermediate headers.
 	var (
-		interimHeader *types.SignedHeader
-		nextVals      *types.ValidatorSet
-		err           error
+		interimHeader   *types.SignedHeader
+		interimNextVals *types.ValidatorSet
+		err             error
 	)
-	for height := c.trustedHeader.Height + 1; height < newHeader.Height; height++ {
+	for height := trustedHeader.Height + 1; height < newHeader.Height; height++ {
 		interimHeader, err = c.signedHeaderFromPrimary(height)
 		if err != nil {
 			return errors.Wrapf(err, "failed to obtain the header #%d", height)
 		}
 
-		c.logger.Debug("Verify newHeader against lastHeader",
+		c.logger.Debug("Verify newHeader against trustedHeader",
 			"lastHeight", c.trustedHeader.Height,
 			"lastHash", c.trustedHeader.Hash(),
 			"newHeight", interimHeader.Height,
 			"newHash", interimHeader.Hash())
-		err = Verify(c.chainID, c.trustedHeader, c.trustedNextVals, interimHeader, c.trustedNextVals,
+		err = Verify(c.chainID, trustedHeader, trustedNextVals, interimHeader, trustedNextVals,
 			c.trustingPeriod, now, c.trustLevel)
 		if err != nil {
 			return errors.Wrapf(err, "failed to verify the header #%d", height)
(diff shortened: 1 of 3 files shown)


cmwaters:
> We should've established a common vocab. prior to making these changes. Right now, we have : h1, h2, trustedHeader, verifiedHeader, untrustedHeader, unverifiedHeader, interimHeader.

I agree. My personal opinion is that any variables present in exported functions should be very clearly named. i.e. don't use h1 or h2. I think we should choose either verified/trusted and stay with it. I used `verified` in alignment with the function name which is to verify but I would be happy to have the terminology as: `trusted`

cmwaters:
> Also, re VerifyViaVals / VerifyViaSigs - not sure these are the best names. May I suggest VerifySequential and VerifySkipping (or VerifyAdjacent & VerifyNonAdjacent) to underline the fact that the former func verifies adjacent headers, the latter - non-adjacent headers.

Yeah I chose to name based on how it does the verification - the actual method itself - not the conditions of the verification. You could theoretically use the `VerifyViaVals` to verify adjacent headers and the `sequence` function uses `VerifyViaSigs` to verify non adjacent headers. This is just my opinion and I would nevertheless be happy to rename

melekes:
`trusted` is more preferable since (when we pass the very first header to Verify and it's not `verified` header per se, we just trust it because of weak subjectivity)

lite2/verifier.go
```diff
@@ -16,101 +16,150 @@ var (
 	DefaultTrustLevel = tmmath.Fraction{Numerator: 1, Denominator: 3}
 )
 
-// Verify verifies the new header (h2) against the old header (h1). It ensures that:
+// VerifyViaVals verifies the new header (untrustedHeader) against the old header (trustedHeader). It ensures that:
```
melekes:
```suggestion
// VerifyNonAdjacent verifies the new header (untrustedHeader) against the old header (trustedHeader). It ensures that:
```

lite2/verifier.go
```diff
@@ -16,101 +16,150 @@ var (
...
-	h1NextVals *types.ValidatorSet,
-	h2 *types.SignedHeader,
-	h2Vals *types.ValidatorSet,
+	trustedHeader *types.SignedHeader,
+	trustedNextVals *types.ValidatorSet,
+	untrustedHeader *types.SignedHeader,
+	untrustedVals *types.ValidatorSet,
 	trustingPeriod time.Duration,
 	now time.Time,
 	trustLevel tmmath.Fraction) error {
```
melekes: we should check that headers are not adjacent

lite2/verifier.go
```diff
@@ -16,101 +16,150 @@ var (
...
+		case types.ErrNotEnoughVotingPowerSigned:
+			return ErrNewValSetCantBeTrusted{e}
+		default:
+			return e
+		}
+	}
+	return nil
+}
+
+// VerifyViaSigs verifies the directly adjacent new header (untrustedHeader) against the old header (trustedHeader).
```
melekes:
```suggestion
// VerifyAdjacent verifies the directly adjacent new header (untrustedHeader) against the old header (trustedHeader).
```

lite2/verifier.go
```diff
@@ -16,101 +16,150 @@ var (
...
+	if HeaderExpired(trustedHeader, trustingPeriod, now) {
+		return ErrOldHeaderExpired{trustedHeader.Time.Add(trustingPeriod), now}
+	}
+
+	if err := verifyNewHeaderAndVals(chainID, untrustedHeader, untrustedVals, trustedHeader, now); err != nil {
 		return err
 	}
 
+	// Ensure that +2/3 of new validators signed correctly.
+	if err := untrustedVals.VerifyCommit(chainID, untrustedHeader.Commit.BlockID, untrustedHeader.Height,
```
melekes: this should really go last - it's quite expensive - you can easily DDOS the light client

cmwaters: Okay I tried to combine the common functions called by both verifydajacent and verifynonadjacent, but I can separate them out again

lite2/client.go
```diff
@@ -682,8 +682,8 @@ func (c *Client) sequence(
 			"lastHash", c.trustedHeader.Hash(),
 			"newHeight", interimHeader.Height,
 			"newHash", interimHeader.Hash())
-		err = Verify(c.chainID, trustedHeader, trustedNextVals, interimHeader, trustedNextVals,
-			c.trustingPeriod, now, c.trustLevel)
+		err = VerifyAdjacent(c.chainID, trustedHeader, interimHeader, trustedNextVals,
```
melekes: ❤️ this

lite2/verifier.go
```diff
@@ -16,101 +16,156 @@ var (
...
+	untrustedHeader *types.SignedHeader,
+	untrustedVals *types.ValidatorSet,
+	trustingPeriod time.Duration,
+	now time.Time,
+	trustLevel tmmath.Fraction) error {
+
+	if untrustedHeader.Height != trustedHeader.Height+1 {
+		return VerifyNonAdjacent(chainID, trustedHeader, trustedNextVals, untrustedHeader, untrustedVals,
+			trustingPeriod, now, trustLevel)
+	} else {
```
golangcibot: `if` block ends with a `return` statement, so drop this `else` and outdent its block (from `golint`)

PR #4418, opened 2020-02-18
Stop txSearch result processing if context is done

### LINKED ISSUE #4375: [tx_search] goroutine for tx search does not stop even if the client timeout
**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source): v0.32.9

**ABCI app** (name for built-in, URL for self-written if it's publicly available): https://github.com/terra-project/core

**Environment**:
- **OS** (e.g. from /etc/os-release): mac-os
- **Install tools**: 
- **Others**:

**What happened**:
It seems tendermint does not stop the search goroutine after the client timeout. In Terra columbus chain, there are lots of txs can't be hold on memory so it cause out-of-memory crash. For test purpose, I applied timeout enabled rpcclient(custom) to prevent this situation. But after the client timeout,  the memory keeps increasing until it finish the search task or meet crash.

Stop search goroutine right after the client timeout.

sync columbus-3 chain & do try search with message.action='send' or message.action='multisend'

gterzian:
FIX #4375

* [ ] Wrote tests


gterzian:
@YunSuk-Yeo  could you try with this branch and see if it helps?

I'm not sure if the bottleneck is at the call to `txIndexer.Search` or in the loop handling the results...

@melekes Any ideas on how to best test this? Could it be done at a layer similar to [`TestTxSearch`](https://github.com/tendermint/tendermint/blob/5ea1ff94d1317f9ea130a36333c27a7dc7a61186/rpc/client/rpc_test.go#L416)?

melekes:
Thank you for the PR @gterzian! However, I don't think it quite archives the end goal, which is to stop searching for transactions when context gets cancelled. To do that, we will need to modify `txIndexer.Search(q)` function to accept the `context.Context` and use it to stop the search.

state/txindex/kv/kv.go
```diff
@@ -167,7 +167,14 @@ func (txi *TxIndex) indexEvents(result *types.TxResult, hash []byte, store dbm.S
 // better for the client to provide both lower and upper bounds, so we are not
 // performing a full scan. Results from querying indexes are then intersected
 // and returned to the caller, in no particular order.
-func (txi *TxIndex) Search(q *query.Query) ([]*types.TxResult, error) {
+func (txi *TxIndex) Search(q *query.Query, done <-chan struct{}) ([]*types.TxResult, error) {
+	select {
+	case <-done:
+		results := make([]*types.TxResult, 0, 0)
```
golangcibot: S1019: should use make([]*types.TxResult, 0) instead (from `gosimple`)

gterzian:
@melekes do you want to take another review pass at this one? I'm wondering how to make codecov happy, would that require an actual RPC test that would timeout? Perhaps in a follow-up we can do that by adding  client that times-out? I can't see an easy way to make the current clients timeout, for example for the purpose of a test.

state/txindex/kv/kv.go
```diff
@@ -417,6 +436,13 @@ func (txi *TxIndex) match(
 			if strings.Contains(extractValueFromKey(it.Key()), c.Operand.(string)) {
 				tmpHashes[string(it.Value())] = it.Value()
 			}
+
+			// Potentially exit early.
+			select {
```
YunSuk-Yeo: Shouldn't we need to add this escape feature to for loop in `c.Op == query.OpEqual`and in below intersection? If not, it will keep loading all tx index to memory and doing unnecessary intersection work.

https://github.com/tendermint/tendermint/blob/0c7827a226cc0928d4903ea5db60db8a3c5787a1/state/txindex/kv/kv.go#L418

https://github.com/tendermint/tendermint/blob/0c7827a226cc0928d4903ea5db60db8a3c5787a1/state/txindex/kv/kv.go#L464

rpc/client/rpc_test.go
```diff
@@ -413,6 +423,18 @@ func TestTx(t *testing.T) {
 	}
 }
 
+func TestTxSearchWithTimeout(t *testing.T) {
+	// Get a client with a time-out of 10 secs.
+	timeout_client := getHTTPClientWithTimeout(10)
```
golangcibot: don't use underscores in Go names; var `timeout_client` should be `timeoutClient` (from `golint`)

gterzian:
@YunSuk-Yeo I have addressed your comments, and also added an timeout client based on the code you linked to, thanks!

gterzian:
@melekes This one is ready for review.

gterzian:
```
could not clone .git: unexpected client error: reference not found
Makefile:51: recipe for target 'proto-check-breaking' failed
make: *** [proto-check-breaking] Error 1
```

The CI error appears unrelated.

gterzian:
@melekes  Looking at your initial comment again, so currently the `Indexed.Search` is expecting a done chan, not a Context as you initially suggested. Do you think a done chan is ok? It would seem to actually make it easier to test the Indexer without requiring a Context, and to overall tie it less into a specific type of client(one that would have a Context).

erikgrinaker:
Hi @gterzian, thanks for contributing this!

> currently the Indexed.Search is expecting a done chan, not a Context as you initially suggested. Do you think a done chan is ok?

I think passing a context is the idiomatic way to do this in Go, and it should always be the first argument to the function (as specified in the [context docs](https://golang.org/pkg/context/)). Creating a context is as easy as creating a channel. Would you mind updating with this?

gterzian:
@erikgrinaker thanks for the pointer, ready for another review, still wondering how, and whether, to make codecov happy...

state/txindex/kv/kv_test.go
```diff
@@ -132,6 +135,73 @@ func TestTxSearch(t *testing.T) {
...
+		{"account.owner CONTAINS 'an'", 0},
+		// search for non existing value using CONTAINS
+		{"account.owner CONTAINS 'Vlad'", 0},
+		// search using the wrong key (of numeric type) using CONTAINS
+		{"account.number CONTAINS 'Iv'", 0},
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	cancel()
```
melekes: By the looks of it, we only test the very first termination clause here, not others. My question then is why do we need all the queries above? Isn't one sufficient?

rpc/client/rpc_test.go
```diff
@@ -414,6 +424,18 @@ func TestTx(t *testing.T) {
 	}
 }
 
+func TestTxSearchWithTimeout(t *testing.T) {
+	// Get a client with a time-out of 10 secs.
+	timeoutClient := getHTTPClientWithTimeout(10)
```
melekes: not sure what are we testing here. we certainly don't check that when timeout expires, TxSearch func exits.

gterzian: I found it pretty hard to actually test for the timeout, because if it's too early its the rpc request that fails, and if it's too late you just get the query. So just testing that it basically works with a timeout client, but not testing the timeout...

state/txindex/kv/kv_test.go
gterzian: Perhaps, I just copy/pasted the queries from the other test, setting the expected results to 0 in all cases. Again I think it would be very hard to test specific timeout scenarios.

PR #4427, opened 2020-02-18
ADR-053: update after experimental prototype

erikgrinaker:
Updates ADR-053 after experimental prototype. Let's discuss open questions and implementation plan before merging.


docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -64,11 +72,13 @@ message ResponseListSnapshots {
...
 message RequestOfferSnapshot {
     Snapshot snapshot = 1;
+    bytes app_hash = 2;
 }
 
 message ResponseOfferSnapshot {
     bool accepted = 1;
-    enum reason {             // Reason why snapshot was rejected
+    Reason reason = 2;        // Reason why snapshot was rejected
+    enum Reason {
```
marbar3778: minor nit: the scoping rules of protobuf dictate that enum constant values must be unique within a protobuf file, hence calling it UNKNOWN_CHANGE_TYPE and not UNKNOWN.

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -192,14 +191,96 @@ Taking consistent snapshots of IAVL trees is greatly simplified by them being ve
...
 
 * Is it OK for state-synced nodes to not have historical blocks nor historical IAVL versions?
 
+* Do we need incremental chunk verification for first version?
+
+* Should we call these snapshots? IAVL already has a different concept called snapshots.
+
+* Should the snapshot ABCI interface be a separate optional ABCI service, or mandatory?
+
+* How can we make sure `ListSnapshots` data is valid? An adversary can provide a large number of invalid snapshots and make the node stuck in discovery mode. Perhaps collect snapshot list from many peers and pick ones that are available on a majority.
```
melekes: limit to some number (say 10). assume nodes send latest snapshots first

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -192,14 +191,96 @@ Taking consistent snapshots of IAVL trees is greatly simplified by them being ve
...
+
+* Should we punish nodes that provide invalid snapshots? How?
```
melekes: disconnecting is fine for now

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -192,14 +191,96 @@ Taking consistent snapshots of IAVL trees is greatly simplified by them being ve
...
+
+* Should we store snapshot and chunk metadata in a database? Can we use the database for chunks?
```
melekes: isn't this a responsibility of the app?

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -192,14 +191,96 @@ Taking consistent snapshots of IAVL trees is greatly simplified by them being ve
...
+
+* How large chunks can we send without getting memory issues? Can we stream the chunk contents?
```
melekes: We will probably need to cap chunk size

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -192,14 +191,96 @@ Taking consistent snapshots of IAVL trees is greatly simplified by them being ve
...
+
+* Should a snapshot at height H be taken before or after the block at H is processed? E.g. RPC `/commit` returns app_hash after _previous_ height, i.e. _before_  current height.
```
melekes: after the `Commit` is finished?

docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -192,14 +191,96 @@ Taking consistent snapshots of IAVL trees is greatly simplified by them being ve
...
+
+* Do we need to support all versions of blockchain reactor (i.e. fast sync)?
```
melekes: what work needs to be done to support each reactor version?

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: I believe this only applies to sibling enums. So the `unknown` enum values in `ResponseOfferSnapshot` and `ResponseApplySnapshotChunk` are not in conflict, because they are not siblings. Or were you thinking about possible conflicts with future additional enums within the same response classes?

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: Yes. But we will implement this in Gaia, so how will Gaia store snapshots? I think storing metadata in a database and chunks in the filesystem is the best option, but I see that the BlockStore is actually storing blocks in a database, so I'd like to discuss further tomorrow.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: Yes, it's capped at 64 MB. But if we're downloading e.g. 8 chunks in parallel, that's 0.5 GB memory usage. And I think this is initially limited by the gRPC response buffer size, which is probably on the order of 100 MB.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: I think that makes sense. But it seems to me like certain parts of Tendermint refer to "height H" as _before_ that block is applied, while other refer to it as _after_ it's applied, so I just want to make sure we're somewhat consistent.

docs/architecture/adr-053-state-sync-prototype.md
erikgrinaker: Depends on how we decide to go forward with implementation. If we do a staged startup with dynamic reactor startup, then no additional work per version. Otherwise, we will need to implement switching from state sync to fast sync by manipulating reactor internals, which I think should be about 1-3 days for all versions - this is also somewhat more brittle, since we must support several combinations of reactor state transitions. Personally I think we should support all versions regardless, or possibly remove some reactor versions.

melekes: > Personally I think we should support all versions regardless, or possibly remove some reactor versions.

agree. we should probably remove v1 since it's an intermediate version afaik

docs/architecture/adr-053-state-sync-prototype.md
melekes: > gRPC response buffer

at what stage we're using gRPC?

erikgrinaker: Sorry, by habit I just automatically think gRPC whenever I'm sending Protobuf requests. Does the p2p layer buffer entire messages in memory, or can it also do streaming processing?

docs/architecture/adr-053-state-sync-prototype.md
melekes: > BlockStore is actually storing blocks in a database

I think that's because it was super easy.

> storing metadata in a database and chunks in the filesystem is the best option

strongly agree

docs/architecture/adr-053-state-sync-prototype.md
melekes: the answer is somewhere in this file https://github.com/tendermint/tendermint/blob/master/p2p/conn/connection.go#L1

Issue #4213, opened 2019-12-03
Formalise definition, detection and processing of individual evidence


milosevic:
https://github.com/tendermint/tendermint/issues/4189 introduces two kind of evidences: individual and global. We need to formalise the notion of individual evidence and explain how it is being detected and created from a fork (detected by the lite client). Furthermore, the specification of required data structures and individual evidence processing on chain should be done.

melekes:
https://github.com/tendermint/tendermint/issues/4182 InvalidHeaderEvidence to punish "Lunatic" Validators

Issue #4182, opened 2019-11-21
Introduce InvalidHeaderEvidence to punish "Lunatic" Validators


ebuchman:
A previous discussion (#3244) surfaced an attack on light clients that can be conducted by a cabal of validators that once controlled +1/3 of the validator set (call them VC, no pun intended ;) ) but are no longer part of the validator set. Since the attacking validators are no longer part of the validator set, this was called "counterfactual". The attack works against the "skipping" verification method (aka "bisection"), where VC gives a light client a future header that is incorrect (ie. not a result of the state transition function on committed blocks) but that the light client trusts because it is signed by VC (+1/3 of a validator set it trusts). In this new header, the full validator set is VC, so they can now do whatever they want. The unique thing about the attack is that at no point do the validators in VC commit an equivocation or a violation of the Tendermint consensus rules (just application-level invalid state transitions). 

It was [also discovered](https://github.com/tendermint/tendermint/issues/3244#issuecomment-518379429) that the attack works even if the VC are still an active part of the validator set, but are now less than +1/3. Thus the attack is more general than the counterfactual case. Since the attack always involves signing an invalid header, we call it instead the "lunatic" attack, since the participating validators are clearly nuts :).

Note if VC were still +1/3 of the validator set, the attack would necessarily involve some form of violating the consensus rules (ie. equivocation or amnesia, according to [parlance](https://github.com/tendermint/spec/blob/master/spec/consensus/fork-accountability.md#attack-scenarios)), and would also allow them to censor any evidence. Since the attack only works if they become less than +1/3 of the validator set, it's possible to avoid censorship and submit evidence and punish them within the unbonding period. One caveat: after conducting the attack, if VC can recover +1/3 of the active validator set, they can censor evidence. Presumably, they cannot guarantee this for themselves, and thus it is worth trying to punish them on chain for this.

To punish this attack, we need support for a new Evidence type, call it `InvalidHeaderEvidence`. This type includes a vote and a header. The header must contain fields that are invalid with respect to the previous block, and a vote for that header by a validator that was in a validator set within the unbonding period. While the attack is only possible if +1/3 of some validator set colludes, the evidence should be verifiable independently for each individual validator. This means the total evidence can be split into one piece of evidence per attacking validator and gossipped to nodes to be verified one piece at a time, reducing the DoS attack surface at the peer layer.

[...]

PR #4429, opened 2020-02-18
adr-047: evidence handling

melekes:
[RENDERED](https://github.com/tendermint/tendermint/blob/611c5510837b1e661f6e7f9684514f7c337f7a04/docs/architecture/adr-047-handling-evidence-from-light-client.md)

Closes https://github.com/tendermint/tendermint/issues/4213 and https://github.com/tendermint/tendermint/issues/4182

* [ ] ~~Wrote tests~~

### DIFF
--- a/docs/architecture/adr-047-handling-evidence-from-light-client.md
+++ b/docs/architecture/adr-047-handling-evidence-from-light-client.md
@@ -1,0 +1,177 @@
+# ADR 047: Handling evidence from light client
+
+## Changelog
+* 18-02-2020: Initial draft
+* 24-02-2020: Second version
+
+## Context
+
+If the light client is under attack (either directly - lunatic / phantom
+validators or indirectly - fork on the main chain), it's supposed to halt and
+send an evidence of misbehavior to the correct full node. Upon receiving an
+evidence, full node should punish malicious validators (if possible).
+
+## Decision
+
+When a light client sees two conflicting headers (`H1.Hash() != H2.Hash()`,
+`H1.Height == H2.Height`), both having 1/3+ of voting power of the currently
+trusted validator set, it will submit a `ConflictingHeadersEvidence` to all
+full nodes it's connected to.
+
+```go
+type ConflictingHeadersEvidence struct {
+  H1 types.SignedHeader
+  H2 types.SignedHeader
+}
+```
+
+When a full node receives the `ConflictingHeadersEvidence` evidence, it should
+a) validate it b) figure out if malicious behaviour is obvious (immediately
+slashable) or the fork accountability protocol needs to be started.
+
+### Validating headers
+
+Check both headers are valid (`ValidateBasic`), have
+the same height and signed by 1/3+ of known validator set.
+
+- Q: What if light client validator set is not equal to full node's validator
+  set (i.e. from full node's point of view both headers are not properly signed;
+  this includes the case where none of the two headers were committed on the
+  main chain)
+
+  Reject the evidence. It means light client is following a fork, but, hey, at
+  least it will halt.
+
+- Q: Don't we want to punish validators who signed something else even if they
+  have less or equal than 1/3?
+
+  No consensus so far. Ethan said no, Zarko said yes.
+  https://github.com/tendermint/spec/pull/71#discussion_r374210533
+
+### Figuring out if malicious behaviour is immediately slashable
+
+Let's say H1 was committed on the main chain. Intersect validator sets of H1
+and H2.
+
+* if there are signers(H2) that are not part of validators(H1), they misbehaved as
+they are signing protocol messages in heights they are not validators =>
+immediately slashable (#F4).
+
+* if `H1.Round == H2.Round`, and some signers signed different precommit
+messages in both commits, then it is an equivocation misbehavior => immediately
+slashable (#F1).
+
+* if `H1.Round != H2.Round` we need to run full detection procedure => not
+immediately slashable.
+
+* if `ValidatorsHash`, `NextValidatorsHash`, `ConsensusHash`,
+`AppHash`, and `LastResultsHash` in H2 are different (incorrect application
+state transition), then it is a lunatic misbehavior => immediately slashable (#F5).
+
+If evidence is not immediately slashable, fork accountability needs to invoked
+(ADR does not yet exist).
+
+It's unclear if we should further break up `ConflictingHeadersEvidence` or
+gossip and commit it directly. See
+https://github.com/tendermint/tendermint/issues/4182#issuecomment-590339233
+
+If we'd go without breaking evidence, all we'll need to do is to strip the
+committed header from `ConflictingHeadersEvidence` (H1) and leave only the
+uncommitted header (H2):
+
+```go
+type ConflictingHeaderEvidence struct {
+  H types.SignedHeader
+}
+```
+
+If we'd go with breaking evidence, here are the types we'll need:
+
+### F1. Equivocation
+
+Existing `DuplicateVoteEvidence` needs to be created and gossiped.
+
+### F4. Phantom validators
+
+A new type of evidence needs to be created:
+
+```go
+type PhantomValidatorEvidence struct {
+  PubKey crypto.PubKey
+  Vote types.Vote
+}
+```
+
+It contains a validator's public key and a vote for a block, where this
+validator is not part of the validator set.
+
+### F5. Lunatic validator
+
+```go
+type LunaticValidatorEvidence struct {
+  Header types.Header
+  Vote types.Vote
+}
+```
+
+To punish this attack, we need support for a new Evidence type -
+`LunaticValidatorEvidence`. This type includes a vote and a header. The header
+must contain fields that are invalid with respect to the previous block, and a
+vote for that header by a validator that was in a validator set within the
+unbonding period. While the attack is only possible if +1/3 of some validator
+set colludes, the evidence should be verifiable independently for each
+individual validator. This means the total evidence can be split into one piece
+of evidence per attacking validator and gossipped to nodes to be verified one
+piece at a time, reducing the DoS attack surface at the peer layer.
+
+Note it is not sufficient to simply compare this header with that committed for
+the corresponding height, as an honest node may vote for a header that is not
+ultimately committed. Certain fields may also be variable, for instance the
+`LastCommitHash` and the `Time` may depend on which votes the proposer includes.
+Thus, the header must be explicitly checked for invalid data.
+
+For the attack to succeed, VC must sign a header that changes the validator set
+to consist of something they control. Without doing this, they can not
+otherwise attack the light client, since the client verifies commits according
+to validator sets. Thus, it should be sufficient to check only that
+`ValidatorsHash` and `NextValidatorsHash` are correct with respect to the
+header that was committed at the corresponding height.
+
+That said, if the attack is conducted by +2/3 of the validator set, they don't
+need to make an invalid change to the validator set, since they already control
+it. Instead they would make invalid changes to the `AppHash`, or possibly other
+fields. In order to punish them, then, we would have to check all header
+fields.
+
+Note some header fields require the block itself to verify, which the light
+client, by definition, does not possess, so it may not be possible to check
+these fields. For now, then, `LunaticValidatorEvidence` must be checked against
+all header fields which are a function of the application at previous blocks.
+This includes `ValidatorsHash`, `NextValidatorsHash`, `ConsensusHash`,
+`AppHash`, and `LastResultsHash`. These should all match what's in the header
+for the block that was actually committed at the corresponding height, and
+should thus be easy to check.
+
+## Status
+
+Proposed.
+
+## Consequences
+
+### Positive
+
+* Tendermint will be able to detect & punish new types of misbehavior
+* light clients connected to multiple full nodes can help full nodes notice a
+  fork faster
+
+### Negative
+
+* Accepting `ConflictingHeadersEvidence` from light clients opens up a DDOS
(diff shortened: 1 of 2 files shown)


docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+When a full node receives the `ConflictingHeadersEvidence` evidence, it should
+a) validate it b) figure out if malicious behaviour is obvious (immediately
+slashable) or the fork accountability protocol needs to be started.
+
+(a): check both headers are valid (`ValidateBasic`), have
+the same height and signed by 1/3+ of known validator set.
+
+- What to do if neither H1 nor H2 was committed on the main chain?
```
cmwaters: Is that possible if both headers are each signed by more than 1/3 of the known validator set and there are no double signings

melekes: yes, see https://github.com/tendermint/tendermint/issues/4182

docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Changelog
+* 18-02-2020: Initial draft
+* 24-02-2020: Second version
+
+## Context
+
+If the light client is under attack (either directly - lunatic / phantom
+validators or indirectly - fork on the main chain), it's supposed to halt and
+send an evidence of misbehavior to the correct full node. Upon receiving an
```
marbar3778:
```suggestion
send evidence of misbehavior to a full node. Upon receiving an
```
? Not sure what correct full node means, maybe I'm missing some context

docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -0,0 +1,165 @@
...
+Let's say H1 was committed on the main chain. Intersect validator sets of H1
+and H2.
+
+* if there are signers(H2) that are not part of validators(H1), they misbehaved as
+they are signing protocol messages in heights they are not validators =>
+immediately slashable (#F4).
+
+* if `H1.Round == H2.Round`, and some signers signed different precommit
+messages in both commits, then it is an equivocation misbehavior => immediately
+slashable (#F1).
```
marbar3778: link to what F1 is?

docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -0,0 +1,165 @@
...
+
+* if `H1.Round != H2.Round` we need to run full detection procedure => not
+immediately slashable.
+
+* if `ValidatorsHash`, `NextValidatorsHash`, `ConsensusHash`,
+`AppHash`, and `LastResultsHash` in H2 are different (incorrect application
+state transition), then it is a lunatic misbehavior => immediately slashable.
```
marbar3778:
```suggestion
state transition), then it is a lunatic misbehavior => immediately slashable (#F5).
```

docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -0,0 +1,165 @@
...
+
+Existing `DuplicateVoteEvidence` needs to be created and gossiped.
+
+### F4. Phantom validators
+
+A new type of evidence needs to be created:
+
+```go
+type PhantomValidatorEvidence struct {
+	PubKey crypto.PubKey
```
marbar3778:
```suggestion
  PubKey crypto.PubKey
```

docs/architecture/adr-047-handling-evidence-from-light-client.md
melekes: correct - not malicious

docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -0,0 +1,165 @@
...
+## Status
+
+Proposed.
+
+## Consequences
+
+### Positive
+
+* Tendermint will be able to detect & punish new types of misbehavior
+
```
melekes: light clients connected to multiple full nodes can help full nodes notice a fork faster?

docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -0,0 +1,177 @@
...
+## Changelog
+* 18-02-2020: Initial draft
+* 24-02-2020: Second version
+
+## Context
+
+If the light client is under attack (either directly - lunatic / phantom
+validators or indirectly - fork on the main chain), it's supposed to halt and
+send an evidence of misbehavior to the correct full node. Upon receiving an
```
marbar3778:
```suggestion
send evidence of misbehavior to the correct full node. Upon receiving an
```

docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -0,0 +1,177 @@
...
+state transition), then it is a lunatic misbehavior => immediately slashable (#F5).
+
+If evidence is not immediately slashable, fork accountability needs to invoked
+(ADR does not yet exist).
+
+It's unclear if we should further break up `ConflictingHeadersEvidence` or
+gossip and commit it directly. See
+https://github.com/tendermint/tendermint/issues/4182#issuecomment-590339233
+
+If we'd go without breaking evidence, all we'll need to do is to strip the
```
marbar3778: are there any benefits other than some simplicity to this approach? Seems like breaking evidence into different types may provide more benefits

melekes: > Seems like breaking evidence into different types may provide more benefits

I don't see any benefits in doing so.

erikgrinaker: From my very limited understanding the evidence receiver should be able to determine the type of misbehaviour from `ConflictingHeadersEvidence` alone. If the sender were to break the evidence up, the receiver would still have to verify the evidence, thus breaking it up does not simplify the implementation. On the other hand, keeping `ConflictingHeadersEvidence` does simplify the protocol.

It appears the only reason for breaking it up is to reduce a DoS vector, i.e. for performance reasons. I believe the typical amount of data is small, on the order of kilobytes, and can be managed by request size limits. What is the cost of verifying the evidence and slashing, in terms of CPU time and real time? How much is saved by submitting evidence for a single validator instead?

cmwaters: But in the case of breaking up the evidence, the receiver does not need to examine every type of attack, only the one being passed to it. For example, the node knows that it just needs to look for duplicate votes and not for a phantom validator. 

It could also be possible to batch them all at once i.e.
```
type PhantomValidatorsEvidence struct {
    PubKeys []crypto.PubKey
    Vote  types.Vote
}
```

erikgrinaker: Does that actually save anything though? The logic for evidence detection has to be in there anyway, and I'd guess running it is not very costly?

cmwaters: Yes the evidence detection is already there and maybe you're right in that it wouldn't be too costly to run

docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -0,0 +1,177 @@
+# ADR 047: Handling evidence from light client
+
+## Changelog
+* 18-02-2020: Initial draft
+* 24-02-2020: Second version
+
+## Context
+
+If the light client is under attack (either directly - lunatic / phantom
+validators or indirectly - fork on the main chain), it's supposed to halt and
```
ancazamfir: Could you add here a reference to the fork accountability spec where these are defined?

docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -0,0 +1,177 @@
...
+send evidence of misbehavior to the correct full node. Upon receiving an
```
ancazamfir: How does the client know which is the correct node?

docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+When a full node receives the `ConflictingHeadersEvidence` evidence, it should
+a) validate it b) figure out if malicious behaviour is obvious (immediately
+slashable) or the fork accountability protocol needs to be started.
+
+### Validating headers
+
+Check both headers are valid (`ValidateBasic`), have
+the same height and signed by 1/3+ of known validator set.
```
ancazamfir: what is the "known validator set" exactly? Is the valset that the full node has for its block at same height as H1 and H2? 
Also, are the validator sets as seen by the client for H1 and H2 part of the Evidence?

docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -0,0 +1,177 @@
...
+
+- Q: Don't we want to punish validators who signed something else even if they
+  have less or equal than 1/3?
+
+  No consensus so far. Ethan said no, Zarko said yes.
+  https://github.com/tendermint/spec/pull/71#discussion_r374210533
+
+### Figuring out if malicious behaviour is immediately slashable
+
+Let's say H1 was committed on the main chain. Intersect validator sets of H1
```
ancazamfir: What is "main chain"? Will a full node know that H1 was committed on the "main" chain? The full node sees two headers at same height, H1 and H2. It could check its state and see that it has one, let's say H1. H2 may indicate either a full-fork or a light-fork but I don't think the full node can distinguish. 
Maybe "the full node's chain" or something along those lines.

docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -0,0 +1,177 @@
...
+and H2.
+
+* if there are signers(H2) that are not part of validators(H1), they misbehaved as
+they are signing protocol messages in heights they are not validators =>
+immediately slashable (#F4).
```
ancazamfir: Here also I don't understand. I looked back at the fork accountability spec a bit and it is not clear there to me either. What if there is a case of full-fork on chain where validator set diverged and some honest validators un-bonded on one branch. I think the phantom scenario is for the case where we can tell with certainty that H1 is from the canonical chain and H2 is from a soft-fork...but not clear to me how a full node can determine this by just looking at H1 and H2.

docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -0,0 +1,177 @@
...
+If evidence is not immediately slashable, fork accountability needs to invoked
+(ADR does not yet exist).
+
+It's unclear if we should further break up `ConflictingHeadersEvidence` or
+gossip and commit it directly. See
+https://github.com/tendermint/tendermint/issues/4182#issuecomment-590339233
+
+If we'd go without breaking evidence, all we'll need to do is to strip the
+committed header from `ConflictingHeadersEvidence` (H1) and leave only the
+uncommitted header (H2):
```
ancazamfir: Does the light client "remember" where it got the headers from? If H1 came from FN1 and H2 from FN2 the client submits evidence to both?

docs/architecture/adr-047-handling-evidence-from-light-client.md
melekes: It's in the footer (see ## References)

docs/architecture/adr-047-handling-evidence-from-light-client.md
melekes: It doesn't, which is why it sends evidence to all connected full nodes.

docs/architecture/adr-047-handling-evidence-from-light-client.md
melekes: > Is the valset that the full node has for its block at same height as H1 and H2?

good question! it's the valset that the full node has for its block at height H1 - 1.

> Also, are the validator sets as seen by the client for H1 and H2 part of the Evidence?

no, only commit signatures (votes).

docs/architecture/adr-047-handling-evidence-from-light-client.md
melekes: > What is "main chain"

the longest branch in Bitcoin, the branch 2/3+ of vals voted for in Tendermint (note: can be a fork if >= 1/3+ are malicious).

> Will a full node know that H1 was committed on the "main" chain?

if there is an actual fork (chain split), full node may follow either one or another branch. It's supposed to halt if it notices the full fork. So both H1 or H2 can be considered committed on the main chain depending on which branch the full node is following.

> H2 may indicate either a full-fork or a light-fork but I don't think the full node can distinguish.

It can because it knows more than the light client (app state for example).

docs/architecture/adr-047-handling-evidence-from-light-client.md
melekes: to both

melekes: here we're talking about a full node, not the light client! a full node needs to strip committed header (from it's perspective) and send only "malicious" header to other nodes.

docs/architecture/adr-047-handling-evidence-from-light-client.md
melekes: > What if there is a case of full-fork on chain where validator set diverged and some honest validators un-bonded on one branch.

Then we'll punish them on our chain for voting on the forked chain.

melekes: > validator set diverged

note this can only happen with 2/3+ of malicious validators

PR #4433, opened 2020-02-19
Use uint64 for consensus.Reactor.SwitchToConsensus() blocksSynced

### LINKED ISSUE #4416: [consensus] switchToConsensus should take blocksSynced int64
In anticipation of of creating many blocks, the consensus reactor switchToConsensus method should accept an int64 instead of an int.

erikgrinaker:
Fixes #4416. Since we officially only support 64-bit archs, `int` already implies `int64`, but explicit is better than implicit. Also, we don't need nor want negative numbers, so made it `uint64`.

* [x] ~Wrote tests~

### DIFF
--- a/blockchain/v0/reactor.go
+++ b/blockchain/v0/reactor.go
@@ -41,7 +41,7 @@ const (
 type consensusReactor interface {
 	// for when we switch from blockchain reactor and fast sync to
 	// the consensus machine
-	SwitchToConsensus(sm.State, int)
+	SwitchToConsensus(sm.State, uint64)
 }
 
 type peerError struct {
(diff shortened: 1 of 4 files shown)


erikgrinaker:
I think this change was a result of v2 discussions, so I suggest we merge this to `master` and @brapse can merge it into the v2 branch.

marbar3778:
that works 😄 

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FguJZ84aiImq5i%2Fgiphy.gif&token=goal" data-gitmeme-token="goal"></a>

Issue #2998, opened 2018-12-10
change pex addressbooks bucket selection hash


ValarDragon:
Currently the pex address book uses double sha256. However this is essentially distributing buckets for a hashmap. It should instead be like typical hashmaps and use AESHash (this is in the stdlib so can just be copy pasta'd) or use high way hash #2187, it definitely doesn't need the overhead of two cryptographic hashes. (If we really wanted a cryptographic hash, then lets use a fast hash here, like blake2{b,s})

tessr:
In the issue that's linked (#2187), @ebuchman expressed some healthy hesitation about adopting highway hash quickly:

> That said, HighwayHash is young, and while collisions are unlikely and not consensus critical, if they ever do happen they may be difficult to detect and hard to debug. So at least I wouldn't think it's worth considering until its proven to be a significant resource hog in the profiler, and there's probably lots of other things to address in the mempool first.

I generally agree with this kind of thinking. Obviously Bucky is talking about the mempool cache in that case rather than the pex addr book, but I think the same logic applies here--do we know that sha256 is actually creating a bottleneck in the addr book? And if not, is it worth taking on the (small) risk of using a relatively new hash function?

ValarDragon:
This hash is for bucketing, not for uniqueness. Collisions are fine (and in fact should be parameterized such that they occur with non-negligible probability) for efficiency.

PR #4439, opened 2020-02-19
pex: use highwayhash for pex bucket

tau3:
Hello! I'd like to submit resolution for #2998 

* [ ] Wrote tests


p2p/pex/addrbook.go
```diff
@@ -100,6 +99,15 @@ type addrBook struct {
 	wg sync.WaitGroup
 }
 
+var hashKey []byte
+
+func init() {
+	highwayKeyLength := 32
+	hashKey = make([]byte, highwayKeyLength)
+	rand.Seed(time.Now().UnixNano())
+	rand.Read(hashKey)
```
golangcibot: G404: Use of weak random number generator (math/rand instead of crypto/rand) (from `gosec`)

p2p/pex/addrbook.go
```diff
@@ -5,20 +5,19 @@
...
-	"sync"
-	"time"
-
+	"github.com/minio/highwayhash"
 	"github.com/tendermint/tendermint/crypto"
 	tmmath "github.com/tendermint/tendermint/libs/math"
 	tmrand "github.com/tendermint/tendermint/libs/rand"
 	"github.com/tendermint/tendermint/libs/service"
 	"github.com/tendermint/tendermint/p2p"
+	"math"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

zmanian:
Seems like a good use of highwayhash to me.

tau3:
@marbar3778 

ok, it's done

p2p/pex/addrbook.go
```diff
@@ -100,6 +101,14 @@ type addrBook struct {
 	wg sync.WaitGroup
 }
 
+var hashKey []byte
```
melekes: is it possible to avoid creating a global var? I understand we need to seed hashKey with OS randomness, but maybe we can do it in the constructor? What do you think?

tau3:
@melekes i think both suggested changes are worth implementing :)

p2p/pex/addrbook.go
```diff
@@ -745,17 +768,24 @@ func (a *addrBook) calcNewBucket(addr, src *p2p.NetAddress) int {
...
+	}
+	result := int(binary.BigEndian.Uint64(hash2) % newBucketCount)
+	return result, nil
 }
 
-// doublesha256(  key + group +
-//                int64(doublesha256(key + addr))%buckets_per_group  ) % num_old_buckets
-func (a *addrBook) calcOldBucket(addr *p2p.NetAddress) int {
+// hash(  key + group +
+//                hash(doublesha256(key + addr))%buckets_per_group  ) % num_old_buckets
```
erikgrinaker: I think you meant to change `doublesha256` here, not `int64`.

p2p/pex/addrbook.go
```diff
@@ -96,10 +97,18 @@ type addrBook struct {
 	filePath          string
 	key               string // random prefix for bucket placement
 	routabilityStrict bool
+	hashKey           []byte
 
 	wg sync.WaitGroup
 }
 
+func newHashKey() []byte {
+	highwayKeyLength := highwayhash.Size
```
erikgrinaker: This variable isn't really necessary.

tau3:
done.

erikgrinaker:
Is this really breaking?

marbar3778:
I'm not sure if an existing addrbook will be compatible with these changes, I can test it out later today.

tessr:
I have a [question](https://github.com/tendermint/tendermint/issues/2998#issuecomment-597906074) about the issue (#2998) motivating this change. (Didn't repost it here just because I don't want to duplicate discussion.)

ValarDragon:
This would only be breaking if the address book's buckets / hashmap structure is serialized. 

My memory of the code is that only the actual peer's entries are serialized into the hashmap file, and the bucketing is done after deserialization. (In which case this is not breaking)

I could be wrong, I didn't double check the code, and its been a long time since I've looked at the Tendermint p2p stack.

tessr:
Hi @tau3, 

Can you update this branch and fix the conflicts, and then I'll merge it to master? You'll also have to update `ReinstateBadPeers` to handle the error that `calcNewBucket1` returns now--you might want to do something like this: https://github.com/tendermint/tendermint/commit/6eb379206e5a9a9157244e7d5562a030b0c9c0c5

melekes:
Closed for https://github.com/tendermint/tendermint/pull/4810

Issue #4442, opened 2020-02-20
blockchain v2: TestReactorTerminationScenarios fails on master


melekes:
```
--- FAIL: TestReactorTerminationScenarios (0.55s)
    --- PASS: TestReactorTerminationScenarios/simple_termination_on_max_peer_height_-_one_peer (0.09s)
    --- FAIL: TestReactorTerminationScenarios/simple_termination_on_max_peer_height_-_two_peers (0.12s)
        reactor_test.go:337: 
            	Error Trace:	reactor_test.go:337
            	Error:      	Should be true
            	Test:       	TestReactorTerminationScenarios/simple_termination_on_max_peer_height_-_two_peers
```

cc @brapse

PR #4443, opened 2020-02-20
blockchain (v2): comment out TestReactorTerminationScenarios

melekes:
Refs #4442 

Commented out the test to unblock development. See the issue for details.


tessr:
tbh I'd rather delete this test than comment it out. or at least leave a comment explaining why its commented out and how its failing?

melekes:
removed

marbar3778:
This works locally but fails in ci? I haven't run into this failing test locally. Maybe we figure out if its a nondeterministic failure or its is failing because of something else

marbar3778:
this now passes on master: https://circleci.com/gh/tendermint/tendermint/93510?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

melekes:
I believe this is a non-deterministic failure.

ancazamfir:
> I believe this is a non-deterministic failure.

Yes indeed. Never saw this failure, neither locally nor in circleci until merge preparations. They are very likely caused by the sleeps (1ms) that seem now to be too short (ci machines more loaded?). The solutions to make the tests without sleeps was a bit too involved and we did not pursue. So we can remove the tests or try with higher values (2-3 msec). Any preference?

marbar3778:
we do run into issues where the ci produces non-deterministic failures because they get overloaded. This isn't as present anymore but could be the issue.

melekes:
> with higher values (2-3 msec)

should work, I guess

Issue #4447, opened 2020-02-20
Fix golangci-lint warnings


erikgrinaker:
`master` currently throws lots of linter warnings with the latest `golangci-lint`, we should fix these.

erikgrinaker:
The remaining linter warnings are all caused by returning unexported types, e.g. this:

```
lite/multiprovider.go:19:56: exported func NewMultiProvider returns unexported type *github.com/tendermint/tendermint/lite.multiProvider, which can be annoying to use (golint)
func NewMultiProvider(providers ...PersistentProvider) *multiProvider {
```

I believe these are mostly used to force callers to use constructors instead of instantiating raw structs, but this prevents callers from e.g. placing the returned value in a struct or returning it themselves. We should either return interfaces, or export the structs.

erikgrinaker:
Fixed by 8f48c49, b0683b7, and fe11219.

PR #4449, opened 2020-02-20
p2p: use curve25519.X25519() instead of ScalarMult()

erikgrinaker:
Replaces a call to `curve25519.ScalarMult()` with `X25519()`, which fixes the following linter warning towards #4447:

```
p2p/conn/secret_connection.go:365:2: SA1019: curve25519.ScalarMult is deprecated: when provided a low-order point, ScalarMult will set dst to all zeroes, irrespective of the scalar. Instead, use the X25519 function, which will return an error.  (staticcheck)
	curve25519.ScalarMult(shrKey, locPrivKey, remPubKey)
	^
```

The zero check is now handled by `X25519()`, and can safely be removed.

Technically a breaking change, since we're removing `ErrSharedSecretIsZero`.

* [x] ~Wrote tests~

### DIFF
--- a/p2p/conn/secret_connection.go
+++ b/p2p/conn/secret_connection.go
@@ -361,9 +361,9 @@ func computeDHSecret(remPubKey, locPrivKey *[32]byte) (*[32]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	var array [32]byte
-	copy(array[:], shrKey)
-	return &array, nil
+	var shrKeyArray [32]byte
+	copy(shrKeyArray[:], shrKey)
+	return &shrKeyArray, nil
 }
 
 func sort32(foo, bar *[32]byte) (lo, hi *[32]byte) {
--- a/p2p/conn/secret_connection.go
+++ b/p2p/conn/secret_connection.go
@@ -5,7 +5,6 @@ import (
 	"crypto/cipher"
 	crand "crypto/rand"
 	"crypto/sha256"
-	"crypto/subtle"
 	"encoding/binary"
 	"io"
 	"math"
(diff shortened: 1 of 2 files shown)


p2p/conn/secret_connection.go
```diff
@@ -358,19 +356,14 @@ func deriveSecrets(
...
-	zero := new([32]byte)
-	if subtle.ConstantTimeCompare(shrKey[:], zero[:]) == 1 {
-		return nil, ErrSharedSecretIsZero
+func computeDHSecret(remPubKey, locPrivKey *[32]byte) (*[32]byte, error) {
+	shrKey, err := curve25519.X25519(locPrivKey[:], remPubKey[:])
+	if err != nil {
+		return nil, err
 	}
-	return
+	var array [32]byte
```
```suggestion
	var shrKeyCopy [32]byte
```

?

btw, why do we need to make a copy at all?

erikgrinaker: To return an array instead of a byte slice, so that we keep the same API as before. We could rewrite the entire module to use `[]byte` instead, which seems to be the direction that the `curve25519` package is heading anyway, but that's a bigger change.

melekes:
> Technically a breaking change, since we're removing ErrSharedSecretIsZero.

new code is compatible with the old version, right?

erikgrinaker:
>> Technically a breaking change, since we're removing ErrSharedSecretIsZero.
>
>new code is compatible with the old version, right?

Yes. But third-party apps who import the Tendermint package may, however unlikely, depend on `ErrSharedSecretIsZero`. I'm not sure what API stability guarantees we have for the Go API. See also #4451.

melekes:
> I'm not sure what API stability guarantees we have for the Go API.

https://github.com/tendermint/tendermint/blob/master/README.md#semantic-versioning

zmanian:
Good work from the Go Crypto team

PR #4452, opened 2020-02-20
lite2: don't save intermediate headers

### LINKED ISSUE #4426: lite2: make saving intermediate headers optional
Do we want to make saving intermediate headers optional? The intermediate headers just act as cache but if the user of the lite client doesn't query headers often then perhaps it's better to not save intermediate headers as this will decrease the execution speed.

cmwaters:
closes #4426

The sequence and bisection methods no longer save the intermediate headers and validator sets that they require to verify a currently untrusted header. 

* [ ] Wrote tests

### DIFF
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -705,11 +705,11 @@ func (c *Client) sequence(
 	now time.Time) error {
 	// 1) Verify any intermediate headers.
 	var (
-		interimHeader *types.SignedHeader
-		err           error
+		interimNextVals *types.ValidatorSet
+		interimHeader   *types.SignedHeader
+		err             error
 	)
-	for trustedHeader.Height < newHeader.Height {
-		height := trustedHeader.Height + 1
+	for height := trustedHeader.Height + 1; height < newHeader.Height; height++ {
 		interimHeader, err = c.signedHeaderFromPrimary(height)
 		if err != nil {
 			return errors.Wrapf(err, "failed to obtain the header #%d", height)
@@ -729,19 +729,20 @@ func (c *Client) sequence(
 		// Update trusted header and vals.
 		trustedHeader = interimHeader
 		if height == newHeader.Height-1 {
-			trustedNextVals = newVals
+			interimNextVals = newVals
 		} else {
-			trustedNextVals, err = c.validatorSetFromPrimary(height + 1)
+			interimNextVals, err = c.validatorSetFromPrimary(height + 1)
 			if err != nil {
 				return errors.Wrapf(err, "failed to obtain the vals #%d", height+1)
 			}
-			if !bytes.Equal(trustedHeader.NextValidatorsHash, trustedNextVals.Hash()) {
+			if !bytes.Equal(trustedHeader.NextValidatorsHash, interimNextVals.Hash()) {
 				return errors.Errorf("expected next validator's hash %X, but got %X (height #%d)",
 					trustedHeader.NextValidatorsHash,
-					trustedNextVals.Hash(),
+					interimNextVals.Hash(),
 					trustedHeader.Height)
 			}
 		}
+		trustedNextVals = interimNextVals
 	}
 
 	return nil
@@ -768,7 +769,6 @@ func (c *Client) bisection(
 			c.trustLevel)
 		switch err.(type) {
 		case nil:
-			// termination clause
 			if interimHeader.Height == newHeader.Height {
 				return nil
 			}
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -727,7 +727,6 @@ func (c *Client) sequence(
 		}
 
 		// Update trusted header and vals.
-		trustedHeader = interimHeader
 		if height == newHeader.Height-1 {
 			interimNextVals = newVals
 		} else {
@@ -742,7 +741,7 @@ func (c *Client) sequence(
 					trustedHeader.Height)
 			}
 		}
-		trustedNextVals = interimNextVals
+		trustedHeader, trustedNextVals = interimHeader, interimNextVals
 	}
 
 	return nil
@@ -774,7 +773,6 @@ func (c *Client) bisection(
 			}
 
 			// Update the lower bound to the previous upper bound
-			trustedHeader = interimHeader
 			interimVals, err = c.validatorSetFromPrimary(interimHeader.Height + 1)
 			if err != nil {
 				return err
@@ -785,8 +783,7 @@ func (c *Client) bisection(
 					interimVals.Hash(),
 					trustedHeader.Height)
 			}
-			trustedNextVals = interimVals
-
+			trustedNextVals, trustedHeader = interimVals, interimHeader
 			// Update the upper bound to the untrustedHeader
 			interimHeader, interimVals = newHeader, newVals
 
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -675,7 +675,8 @@ func (c *Client) sequence(
 		interimNextVals *types.ValidatorSet
 		err             error
 	)
-	for height := trustedHeader.Height + 1; height < newHeader.Height; height++ {
+	for trustedHeader.Height < newHeader.Height {
+		height := trustedHeader.Height + 1
 		interimHeader, err = c.signedHeaderFromPrimary(height)
 		if err != nil {
 			return errors.Wrapf(err, "failed to obtain the header #%d", height)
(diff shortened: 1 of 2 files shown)


lite2/client_test.go
```diff
@@ -762,6 +760,15 @@ func TestClient_BackwardsVerification(t *testing.T) {
...
 	assert.Error(t, err)
 	assert.NotNil(t, h)
+
+	// 3) already stored headers should return the header without error
+	h, err = c.VerifyHeaderAtHeight(3, bTime.Add(1*time.Hour).Add(1*time.Second))
+	assert.NoError(t, err)
+	assert.NotNil(t, h)
+
+	// 4) cannot verify a header in the future
+	h, err = c.VerifyHeaderAtHeight(4, bTime.Add(1*time.Hour).Add(1*time.Second))
```

lite2/client.go
```diff
@@ -728,16 +733,21 @@ func (c *Client) sequence(
 		}
 
 		// Update trusted header and vals.
+		trustedHeader = interimHeader
```
melekes: will this modify c.latestTrustedHeader each time?

lite2/client.go
```diff
@@ -728,16 +733,21 @@ func (c *Client) sequence(
...
 		if height == newHeader.Height-1 {
-			interimNextVals = newVals
+			trustedNextVals = newVals
 		} else {
-			interimNextVals, err = c.validatorSetFromPrimary(height + 1)
+			trustedNextVals, err = c.validatorSetFromPrimary(height + 1)
 			if err != nil {
 				return errors.Wrapf(err, "failed to obtain the vals #%d", height+1)
 			}
+			if !bytes.Equal(trustedHeader.NextValidatorsHash, trustedNextVals.Hash()) {
```
melekes: probably a bad idea to set something as trusted `trustedNextVals` unless we can trust it

lite2/client.go
```diff
@@ -764,6 +774,11 @@ func (c *Client) bisection(
 			c.trustLevel)
 		switch err.(type) {
 		case nil:
+			// termination clause
+			if interimHeader.Height == newHeader.Height {
```
melekes: why it's better than for trustedHeader.Height < newHeader.Height { ?

cmwaters: it saves having to finish the rest of the loop before termination

lite2/client.go
cmwaters: nope. Only at the end is c.lastTrustedHeader modified. Perhaps the naming should be changed if you think it is confusing. I just did it to save the need of declaring another variable

melekes: > I just did it to save the need of declaring another variable

not sure. it's a pointer. when you modify it's value, you modify the pointer

lite2/client.go
```diff
@@ -767,6 +768,11 @@ func (c *Client) bisection(
 			c.trustLevel)
 		switch err.(type) {
 		case nil:
+			// termination clause
```
```suggestion
```

lite2/client.go
cmwaters: Yeah it will update it (as well as `latestTrustedNextVals` ). But the headers should be trusted by then

lite2/client.go
```diff
@@ -705,11 +705,11 @@ func (c *Client) sequence(
...
 	// 1) Verify any intermediate headers.
 	var (
-		interimHeader   *types.SignedHeader
-		interimNextVals *types.ValidatorSet
-		err             error
+		interimHeader *types.SignedHeader
+		err           error
 	)
-	for height := trustedHeader.Height + 1; height < newHeader.Height; height++ {
+	for trustedHeader.Height < newHeader.Height {
```
melekes: how;s this better than before?

cmwaters: it's not - I tried to make it cleaner but there's not really a difference

melekes: 2 lines instead of one

cmwaters: Yeah I think reverting it is fine, in my defense, my line of thought was that we might not need to declare height in the for loop but we use it so often so I sort of declared it afterwards 😅

lite2/client.go
```diff
@@ -705,11 +705,11 @@ func (c *Client) sequence(
 	now time.Time) error {
 	// 1) Verify any intermediate headers.
 	var (
-		interimHeader   *types.SignedHeader
 		interimNextVals *types.ValidatorSet
+		interimHeader   *types.SignedHeader
```
melekes: why move down?

cmwaters: what do you mean?

lite2/client.go
```diff
@@ -727,23 +727,25 @@ func (c *Client) sequence(
 		}
 
 		// Update trusted header and vals.
+		trustedHeader = interimHeader
```
melekes: we need to set them both together. otherwise, if valset is invalid, trustedHeader will be new, trustedNextValSet - old

lite2/client.go
```diff
@@ -767,23 +769,23 @@ func (c *Client) bisection(
 			c.trustLevel)
 		switch err.(type) {
 		case nil:
+			if interimHeader.Height == newHeader.Height {
+				return nil
+			}
+
 			// Update the lower bound to the previous upper bound
 			trustedHeader = interimHeader
```
melekes: same

lite2/client.go
melekes: Screenshot 2020-02-21 at 12 14 14

cmwaters: Ah I see. I removed it at first and then I added it back later. Does it make a difference

melekes: consistency makes a great difference when other developers reading your code

melekes:
```
    --- FAIL: TestClient_SkippingVerification/good,_but_val_set_changes_100%_at_height_2 (0.03s)
        client_test.go:310: 
            	Error Trace:	client_test.go:310
            	Error:      	Received unexpected error:
            	            	expected next validator's hash 7B9032E2725E5B864B5AEE02B47EA1FE30808272A2701365DE6E3061F40F5693, but got FFC48803418AF8E3B22581141F66D1CEC7F7A7D286093FDA00E9A9AB4A70ECBF (height #1)
            	            	github.com/tendermint/tendermint/lite2.(*Client).bisection
            	            		/go/src/github.com/tendermint/tendermint/lite2/client.go:781
            	            	github.com/tendermint/tendermint/lite2.(*Client).verifyHeader
            	            		/go/src/github.com/tendermint/tendermint/lite2/client.go:602
            	            	github.com/tendermint/tendermint/lite2.(*Client).VerifyHeaderAtHeight
            	            		/go/src/github.com/tendermint/tendermint/lite2/client.go:546
            	            	github.com/tendermint/tendermint/lite2.TestClient_SkippingVerification.func1
            	            		/go/src/github.com/tendermint/tendermint/lite2/client_test.go:306
            	            	testing.tRunner
            	            		/usr/local/go/src/testing/testing.go:909
            	            	runtime.goexit
            	            		/usr/local/go/src/runtime/asm_amd64.s:1357
            	Test:       	TestClient_SkippingVerification/good,_but_val_set_changes_100%_at_height_2
```

cmwaters:
Yeah I'm looking into it.

cmwaters:
> Yeah I'm looking into it.

I know what it is, I made a simple change but missed changing another variable

lite2/client.go
```diff
@@ -734,16 +734,17 @@ func (c *Client) sequence(
 			if err != nil {
 				return errors.Wrapf(err, "failed to obtain the vals #%d", height+1)
 			}
-		}
-		err = c.updateTrustedHeaderAndNextVals(interimHeader, interimNextVals)
-		if err != nil {
-			return errors.Wrapf(err, "failed to update trusted state #%d", height)
+			if !bytes.Equal(trustedHeader.NextValidatorsHash, interimNextVals.Hash()) {
```
```suggestion
			if !bytes.Equal(interimHeader.NextValidatorsHash, interimNextVals.Hash()) {
```

?

melekes: we should have a test for it!

cmwaters: true the test only picked up the bisection method (skipping)

lite2/client_test.go
```diff
@@ -138,6 +141,25 @@ func TestClient_SequentialVerification(t *testing.T) {
 			false,
 			true,
 		},
+		{
+			"bad: different validator set",
```
melekes:
```suggestion
			"bad: different validator set at height 3",
```

lite2/client.go
```diff
@@ -767,24 +768,22 @@ func (c *Client) bisection(
...
+					interimVals.Hash(),
+					interimHeader.Height)
 			}
-
-			err = c.updateTrustedHeaderAndNextVals(trustedHeader, trustedNextVals)
-			if err != nil {
-				return err
-			}
-
+			trustedNextVals, trustedHeader = interimVals, interimHeader
```
melekes: same thing - try to be consistent

cmwaters: Can you further explain, what am I not being consistent with

melekes: In the above sequence func we have `trustedHeader, trustedNextVals =` while here we have `trustedNextVals, trustedHeader =`

Issue #4208, opened 2019-11-30
Public Key encoding  in protobuf compatible serialization world


zmanian:
We want to adopt https://github.com/tendermint/go-amino/pull/276.

If we adopt full protobuf compatibility, this changes the byte encodings of validator consensus keys. These manifest in human and developer-facing interfaces such as the Cosmos SDK cosmosvalpub encoded interfaces and in the Tendermint RPC.

It is desirable to have a prefix that identifies the cryptographic system that a given public key belongs to prevent key confusion attack. 

The legacy system is described in [here](https://github.com/tendermint/spec/blob/master/spec/blockchain/encoding.md#public-key-cryptography).  These encodings are created by the amino format but are trivial to implement as a customer serializer.

My recommendation is that we special case the serialization of public keys in Tendermint.

`<PrefixBytes> <Length> <ByteArray>` seems like generally reasonable format. For compatibility reasons, ed25519 and secp256k1 keys can keep their amino prefix bytes. New key types can select any prefix bytes they desire that don't collide with a key type in github.com/tendermint/spec repo.

A strawman alternative is drop legacy compatibility and just do `<PrefixBytes> <ByteArray>` and reset the prefix for ed25519 to `0x001` and sepc256k1 to `0x0021`.

jackzampolin:
This seems like something we could easily do in a `switch` statement when decoding the key. With strong documentation/reproducibility on the legacy for others who need it. I would imagine we would also need a rust impl here for `signatory`

zmanian:
By switch statement you mean basically a look up table?

I need to check but signatory might just be doing this manually. We were doing that at one point.

The length value is most basically superfluous. It does potentially allow you to send both an x,y coordinate instead of just x or y. This usually called "compressed" form.

jackzampolin:
> By switch statement you mean basically a look up table?

exactly

marbar3778:
@zmanian is this still applicable since we now adopted amino flavored JSON?

privval json encoding stays the same. remote signer encoding will change to oneof, but this change will be handled on the hsm (kms) side correct?

zmanian:
My overall impression is that we don't need to do this and can close this.

Issue #3739, opened 2019-06-20
Wrong implementation of Bytes method for interfaces crypto.PubKey and crypto.PrivKey


ironman0x7b2:
**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source): d2eab536ac1071b1b5f64283572397606326eb73

**ABCI app** (name for built-in, URL for self-written if it's publicly available): None

**Environment**:
- **OS** (e.g. from /etc/os-release): Ubuntu 16.04
- **Install tools**:
- **Others**:

**What happened**:
- It seems like the implementation of crypto.PrivKey.Bytes and crypto.PubKey.Bytes methods in edd25519 and secp256k1 are wrong. Why because the defined types PrivKeyEd25519, PubKeyEd25519, PrivKeySecp256k1 and PubKeySecp256k1 are basically byte arrays so we can directly return the byte slice instead of amino marshalled byte slice.

- The Bytes method returning byte slices with lengths 69, 37, 37 and 38 respectively.

**What you expected to happen**:
- Bytes method should return the original byte slice instead of amino marshalled byte slice.

- The Bytes method should return byte slices with lengths 64, 32, 32 and 33 respectively.

**Have you tried the latest version**: yes/no
Yes

**How to reproduce it** (as minimally and precisely as possible):

**Logs (paste a small part showing an error (< 10 lines) or link a pastebin, gist, etc. containing more of the log file)**:

**Config (you can paste only the changes you've made)**:

**node command runtime flags**:

**`/dump_consensus_state` output for consensus bugs**

**Anything else we need to know**:
Because of this issue, the Bech32 encoded public keys which are generated through Javascript are different than Golang.

https://github.com/tendermint/tendermint/blob/master/crypto/ed25519/ed25519.go#L44
https://github.com/tendermint/tendermint/blob/master/crypto/ed25519/ed25519.go#L143
https://github.com/tendermint/tendermint/blob/master/crypto/secp256k1/secp256k1.go#L46
https://github.com/tendermint/tendermint/blob/master/crypto/secp256k1/secp256k1.go#L154

ebuchman:
One reason not to return the raw bytes is to preserve the type data, so the Bytes() representation also encodes which kind of public key it is. There's also the case of mutlisig pubkeys, which need to encode more than one pubkey and their types.

So I think we might just need to update the JS for this.

ironman0x7b2:
[...]

Issue #2228, opened 2018-08-14
crypto: drop redundant suffixes like ed25519.PubKeyEd25519


ebuchman:
Since we now have nice namespacing via the package name, we don't need to keep mentioning the algorithm name in the package. So we should be able to just use `ed25519.PubKey`.

A related change happened in https://github.com/tendermint/tendermint/pull/2205/, but it didn't go all the way because it might make it harder to debug (https://github.com/tendermint/tendermint/pull/2205/#issuecomment-413030471) - we should consider fixing that in Amino

Issue #2204, opened 2018-08-10
crypto: Switch pubkeys to protobuf oneof


ValarDragon:
Issue is described here: https://github.com/tendermint/tendermint/pull/2164

Basically, we don't expose a way for people to add pubkeys with our current crypto API. Therefore its not the convenient permisionless addition system, so we should consider using the protobuf `oneof` for this, rather than amino interface types.

/cc @ebuchman

ebuchman:
I would go a step further and consider making tendermint-core completely free of amino-interface-types since there is no known use case for them. eg. all reactor messages could/should be done with one-of since it's unclear how/why someone would need the global name spacing here (reactor msgs are already namespaced by the reactors channel bytes). Having amino here just gets in the way of alternative implementations

PR #4460, opened 2020-02-21
proto: move crypto keys to proto

marbar3778:
- make keys `oneof` proto
- remove redundant suffixes from keys
- move keys to []byte instead [size]byte
- bytes method returns []byte(key) instead of amino encoding

ref #2204
ref #2228
ref #4208 
ref #3739

* [ ] Wrote tests


crypto/encoding/amino/encode_test.go
```diff
@@ -53,6 +53,7 @@ func checkAminoJSON(t *testing.T, src interface{}, dst interface{}, isNil bool)
 
 // ExamplePrintRegisteredTypes refers to unknown identifier: PrintRegisteredTypes
 //nolint:govet
+//TODO: this test will break, but keep it as is so we can work towards backwards compatability
```
golangcibot: `compatability` is a misspelling of `compatibility` (from `misspell`)
```suggestion

```

privval/socket_listeners.go
```diff
@@ -45,15 +45,15 @@ var _ net.Listener = (*tcpListener)(nil)
...
+	secretConnKey ed25519.PrivKey
 
 	timeoutAccept    time.Duration
 	timeoutReadWrite time.Duration
 }
 
 // NewTCPListener returns a listener that accepts authenticated encrypted connections
 // using the given secretConnKey and the default timeout values.
-func NewTCPListener(ln net.Listener, secretConnKey ed25519.PrivKeyEd25519) *tcpListener {
+func NewTCPListener(ln net.Listener, secretConnKey ed25519.PrivKey) *tcpListener {
```
golangcibot: exported func NewTCPListener returns unexported type *github.com/tendermint/tendermint/privval.tcpListener, which can be annoying to use (from `golint`)

marbar3778:
will write an adr as the final steps of this work. It will affect the entire ecosystem and should have more people input

crypto/crypto.go
```diff
@@ -26,11 +26,19 @@ type PubKey interface {
 	Equals(PubKey) bool
 }
 
+// this interface will be phased out, this is here for backwards compatibility where needed.
+// Bytes in tendermint 0.33 and before uses marshalbinarybare from amino, this function should provide the
+// bytes as return and not marshal.
+type AminoBytes interface {
+	AminoBytes() []byte
+}
+
```
marbar3778: @alexanderbez you won't need this if I just return []byte of the key and not marshal anything, correct?

alexanderbez: We'd just use the raw type `[]byte`. We wouldn't call any methods. But this might be moot as we're discussing move key types to the SDK.

types/protobuf.go
```diff
@@ -101,17 +101,17 @@ func (tm2pb) ValidatorUpdate(val *Validator) abci.ValidatorUpdate {
 // TODO: add cases when new pubkey types are added to crypto
 func (tm2pb) PubKey(pubKey crypto.PubKey) abci.PubKey {
```
marbar3778: I'll remove these two functions later on as this pr is already growing large and this would cause it to grow greatly

marbar3778:
this should be merged after #4332 

@Liamsi & @ValarDragon could i get your reviews on this as it touches the keys heavily

crypto/ed25519/ed25519.go
```diff
@@ -15,34 +14,24 @@ import (
...
 
 const (
 	PrivKeyAminoName = "tendermint/PrivKeyEd25519"
 	PubKeyAminoName  = "tendermint/PubKeyEd25519"
+	// PubKeySize is the number of bytes in an Ed25519 signature.
+	PubKeySize = 32
 	// Size of an Edwards25519 signature. Namely the size of a compressed
 	// Edwards25519 point, and a field element. Both of which are 32 bytes.
-	SignatureSize = 64
+	PrivateKeySize = 64
```
ValarDragon: Err somethings wrong with the comments here.

Pubkey size = 32 bytes is currently listed, but the comment says its the signature size.

Similar the PrivateKeySize's comment is for SignatureSize as well

consensus/types/round_state_test.go
```diff
@@ -25,7 +25,7 @@ func BenchmarkRoundStateDeepCopy(b *testing.B) {
 			Total: 1000,
 		},
 	}
-	sig := make([]byte, ed25519.SignatureSize)
+	sig := make([]byte, ed25519.PrivateKeySize)
```
melekes: ??? we should use SignatureSize for signatures

crypto/crypto.go
```diff
@@ -31,6 +31,7 @@ type PrivKey interface {
 	Sign(msg []byte) ([]byte, error)
 	PubKey() PubKey
 	Equals(PrivKey) bool
+	// ProtoBytes() []byte
```
melekes: leftover?

crypto/ed25519/ed25519.go
```diff
@@ -75,15 +63,13 @@ func (privKey PrivKeyEd25519) PubKey() crypto.PubKey {
 		panic("Expected PrivKeyEd25519 to include concatenated pubkey bytes")
 	}
 
-	var pubkeyBytes [PubKeyEd25519Size]byte
-	copy(pubkeyBytes[:], privKeyBytes[32:])
```
melekes: not an expert in cryptography, but sometimes copying is necessary because you don't want to refer to existing Go slice, you want an actual copy (privKey[32:] is a pointer!)

rpc/client/rpc_test.go
```diff
@@ -672,8 +673,8 @@ func TestBroadcastEvidenceDuplicateVote(t *testing.T) {
 		require.NoError(t, err)
 		client.WaitForHeight(c, status.SyncInfo.LatestBlockHeight+2, nil)
 
-		ed25519pub := ev.PubKey.(ed25519.PubKeyEd25519)
-		rawpub := ed25519pub[:]
+		ed25519pub := ev.PubKey.(ed25519.PubKey)
+		rawpub := tmbytes.HexBytes(ed25519pub)
```
ValarDragon: I'm a bit confused w.r.t. why this changed. Is it just a general RPC improvement?

types/signable.go
```diff
@@ -9,7 +9,7 @@ var (
 	// MaxSignatureSize is a maximum allowed signature size for the Proposal
 	// and Vote.
 	// XXX: secp256k1 does not have Size nor MaxSize defined.
-	MaxSignatureSize = tmmath.MaxInt(ed25519.SignatureSize, 64)
+	MaxSignatureSize = tmmath.MaxInt(ed25519.PrivateKeySize, 64)
```
ValarDragon: This seems wrong, it should be a variable named SignatureSize?

melekes:
> LGTM (untested)
> this seems wrong, it should be a variable named SignatureSize?

I am confused as to why approve PR when it's clearly wrong 😕

crypto/ed25519/ed25519.go
marbar3778: thank you, reverted these removals

crypto/ed25519/ed25519.go
```diff
@@ -74,16 +64,15 @@ func (privKey PrivKeyEd25519) PubKey() crypto.PubKey {
 	if !initialized {
 		panic("Expected PrivKeyEd25519 to include concatenated pubkey bytes")
 	}
-
-	var pubkeyBytes [PubKeyEd25519Size]byte
-	copy(pubkeyBytes[:], privKeyBytes[32:])
-	return PubKeyEd25519(pubkeyBytes)
+	var pubkeyBytes []byte
+	copy(pubkeyBytes[:], privKey[32:])
```
golangcibot: unslice: could simplify pubkeyBytes[:] to pubkeyBytes (from `gocritic`)

crypto/secp256k1/secp256k1.go
```diff
@@ -115,54 +96,54 @@ func GenPrivKeySecp256k1(secret []byte) PrivKeySecp256k1 {
 	fe.Add(fe, one)
 
 	feB := fe.Bytes()
-	var privKey32 [32]byte
+	privKey32 := make([]byte, PrivKeySize)
 	// copy feB over to fixed 32 byte privKey32 and pad (if necessary)
 	copy(privKey32[32-len(feB):32], feB)
 
-	return PrivKeySecp256k1(privKey32)
+	return PrivKey(privKey32[:])
```
golangcibot: unslice: could simplify privKey32[:] to privKey32 (from `gocritic`)

ValarDragon:
You're right. My rational at the time was that it seemed like a fairly small change, and I didn't want to block the entire PR on it. (I wasn't planning on doing a second pass over the whole thing)

In hindsight, that shouldn't have meant an approval

abci/example/kvstore/persistent_kvstore.go
```diff
@@ -211,8 +211,7 @@ func (app *PersistentKVStoreApplication) execValidatorTx(tx []byte) types.Respon
 func (app *PersistentKVStoreApplication) updateValidator(v types.ValidatorUpdate) types.ResponseDeliverTx {
 	key := []byte("val:" + string(v.PubKey.Data))
 
-	pubkey := ed25519.PubKeyEd25519{}
-	copy(pubkey[:], v.PubKey.Data)
```
marbar3778: do you think the copy is necessary here as well? cc @melekes

crypto/sr25519/privkey.go
```diff
@@ -10,20 +10,22 @@ import (
...
+func (privKey PrivKey) Bytes() []byte {
+	return []byte(privKey)
 }
 
 // Sign produces a signature on the provided message.
-func (privKey PrivKeySr25519) Sign(msg []byte) ([]byte, error) {
-	miniSecretKey, err := schnorrkel.NewMiniSecretKeyFromRaw(privKey)
+func (privKey PrivKey) Sign(msg []byte) ([]byte, error) {
+	var p [PrivKeySize]byte
+	copy(p[:], privKey)
```
melekes: we should avoid extra copy here.

crypto/sr25519/privkey.go
```diff
@@ -41,8 +43,10 @@ func (privKey PrivKeySr25519) Sign(msg []byte) ([]byte, error) {
 }
 
 // PubKey gets the corresponding public key from the private key.
-func (privKey PrivKeySr25519) PubKey() crypto.PubKey {
-	miniSecretKey, err := schnorrkel.NewMiniSecretKeyFromRaw(privKey)
+func (privKey PrivKey) PubKey() crypto.PubKey {
+	var p [PrivKeySize]byte
+	copy(p[:], privKey)
```
melekes: same

crypto/sr25519/pubkey.go
```diff
@@ -41,7 +37,9 @@ func (pubKey PubKeySr25519) VerifyBytes(msg []byte, sig []byte) bool {
 	copy(sig64[:], sig)
 
 	publicKey := &(schnorrkel.PublicKey{})
-	err := publicKey.Decode(pubKey)
+	var p [PubKeySize]byte
+	copy(p[:], pubKey)
```
melekes: same

crypto/sr25519/privkey.go
melekes: looks like there's no easy way to do that. please ignore

marbar3778:
will wait for @Liamsi or @ValarDragon to review before merge

crypto/encoding/proto/types.proto
```diff
@@ -0,0 +1,45 @@
...
+message PrivKey {
+  option (cosmos_proto.interface_type) = "*github.com/tendermint/tendermint/crypto.PrivKey";
+  oneof key {
+    bytes ed25519 = 1
+        [(gogoproto.casttype) = "github.com/tendermint/tendermint/crypto/ed25519.PrivKey"];
+    bytes secp256k1 = 2
+        [(gogoproto.casttype) = "github.com/tendermint/tendermint/crypto/secp256k1.PrivKey"];
+    bytes sr25519 = 3
+        [(gogoproto.casttype) = "github.com/tendermint/tendermint/crypto/sr25519.PrivKey"];
+  }
```
Liamsi: I'm wondering why we'd ever need to send this (any private key material) over the wire. Having a proto message for this seems wrong to me. If it's kept here it should come with a noisy warning, why it's here, how it's used and that you should never send it over the wire.

crypto/encoding/proto/types.proto
```diff
@@ -0,0 +1,45 @@
...
+    bytes ed25519 = 1
+        [(gogoproto.casttype) = "github.com/tendermint/tendermint/crypto/ed25519.PubKey"];
+    bytes secp256k1 = 2
+        [(gogoproto.casttype) = "github.com/tendermint/tendermint/crypto/secp256k1.PubKey"];
+    bytes sr25519 = 3
+        [(gogoproto.casttype) = "github.com/tendermint/tendermint/crypto/sr25519.PubKey"];
+    // TODO: implement
+    // PubKeyMultiSigThreshold multisig = 4;
+    // [(gogoproto.casttype) =
+    //      "github.com/tendermint/tendermint/crypto/multisig.PubKeyMultisigThreshold"];
```
Liamsi: Question: What is holding back this variant? It seems that there is a `PubKeyMultiSigThreshold` message.

crypto/encoding/amino/encode_test.go
```diff
@@ -53,18 +56,34 @@ func checkAminoJSON(t *testing.T, src interface{}, dst interface{}, isNil bool)
...
-	//| PubKeyEd25519 | tendermint/PubKeyEd25519 | 0x1624DE64 | 0x20 |  |
-	//| PubKeySr25519 | tendermint/PubKeySr25519 | 0x0DFB1005 | 0x20 |  |
-	//| PubKeySecp256k1 | tendermint/PubKeySecp256k1 | 0xEB5AE987 | 0x21 |  |
-	//| PubKeyMultisigThreshold | tendermint/PubKeyMultisigThreshold | 0x22C1F7E2 | variable |  |
-	//| PrivKeyEd25519 | tendermint/PrivKeyEd25519 | 0xA3288910 | 0x40 |  |
-	//| PrivKeySr25519 | tendermint/PrivKeySr25519 | 0x2F82D78B | 0x20 |  |
-	//| PrivKeySecp256k1 | tendermint/PrivKeySecp256k1 | 0xE1B0F79B | 0x20 |  |
-}
+//TODO: Fix this with the new proto encoding of keys which are []byte
+// TODO: make a issue for this or add it to the meta-issue
```
Liamsi: Was there an issue opened for this?

crypto/secp256k1/encoding.go
```diff
@@ -0,0 +1,122 @@
...
+var (
+	prefixPrivKeySecp256k1      = []byte{0xE1, 0xB0, 0xF7, 0x9B}
+	lengthPrivKeySecp256k1 byte = 0x20
+
+	prefixPubKeySecp256k1      = []byte{0xEB, 0x5A, 0xE9, 0x87}
+	lengthPubKeySecp256k1 byte = 0x21
+)
+
+// MarshalBinary attempts to marshal a PrivKeySecp256k1 type that is backwards
+// compatible with Amino.
```
Liamsi: Oh I see, this is used to store the private keys locally, right?

crypto/encoding/amino/encode_test.go
```diff
@@ -18,6 +17,8 @@ import (
 
 type byter interface {
 	Bytes() []byte
+	AminoMarshal() ([]byte, error)
+	AminoUnmarshal([]byte) error
```
Liamsi: Nit: I think @ValarDragon introduced this byter interface. It would be cleaner if it still had only one method namely `Bytes()`. Couldn't the other live in a separate Iface (e.g. AminoMarshaler?).

crypto/secp256k1/encoding.go
marbar3778: these are here for now, as I go through the codebase to find where we actually depend on `MarshalLengthPrefixed` (currently can't find anything), but if there isn't a need for them then they will be removed. These can also be helper functions if any apps are relying on `LengthPrefixed` bytes

crypto/encoding/proto/types.proto
marbar3778: running into quite a few cyclical imports here. Not sure how to tackle this type. If you have time in the coming days could we set up a call?

crypto/encoding/amino/encode_test.go
marbar3778: removed this as it will not apply anymore

crypto/encoding/proto/types.proto
marbar3778: After trying multiple strategies, the one that I have come to that is using the customtype extension which means we will need these methods on the multisig: https://github.com/gogo/protobuf/blob/master/custom_types.md#custom-type-method-signatures

marbar3778: there was also discussion around the cosmos-sdk moving keys to the app, the ones that they use. If this is the case then multisig would be removed.

crypto/multisig/threshold_pubkey.go
```diff
@@ -94,3 +94,39 @@ func (pk PubKeyMultisigThreshold) Equals(other crypto.PubKey) bool {
 	}
 	return true
 }
+
+//------------------------------
+
+// Methods needed for useing gogoproto.customtype
```
golangcibot: `useing` is a misspelling of `using` (from `misspell`)
```suggestion

```

crypto/multisig/threshold_pubkey.go
```diff
@@ -94,3 +94,39 @@ func (pk PubKeyMultisigThreshold) Equals(other crypto.PubKey) bool {
...
+// These is no clean way of getting multisg type into the types.proto
+// have two representations of the type is needed
+
+func (t PubKeyMultisigThreshold) Marshal() ([]byte, error) {
```
golangcibot: receiver name t should be consistent with previous receiver name pk for PubKeyMultisigThreshold (from `golint`)

crypto/multisig/threshold_pubkey.go
```diff
@@ -94,3 +94,39 @@ func (pk PubKeyMultisigThreshold) Equals(other crypto.PubKey) bool {
...
+	panic("Not Implemented!")
+}
+func (t *PubKeyMultisigThreshold) MarshalTo(data []byte) (n int, err error) {
```
golangcibot: receiver name t should be consistent with previous receiver name pk for PubKeyMultisigThreshold (from `golint`)

crypto/multisig/threshold_pubkey.go
```diff
@@ -94,3 +94,39 @@ func (pk PubKeyMultisigThreshold) Equals(other crypto.PubKey) bool {
...
+	panic("Not Implemented!")
+}
+func (t *PubKeyMultisigThreshold) Unmarshal(data []byte) error {
```
golangcibot: receiver name t should be consistent with previous receiver name pk for PubKeyMultisigThreshold (from `golint`)

marbar3778:
I would like to propose dropping secp, sr25519 and multisig from proto types in tendermint. this means the pubkey and privkey proto types would only house ed25519. This would allow the cosmos-sdk to define their own proto types for there keys and reduce overhead for supporting things in tendermint.

tessr:
> dropping secp, sr25519 and multisig from proto types in tendermint

Tendermint does not use any of these key types, right?

marbar3778:
we do not use these keys for messaging, or Tendermint state. We do use secp256k1 for testing as it has a larger key size, but this would still be possible.

crypto/ed25519/ed25519.go
```diff
@@ -52,19 +46,19 @@ func (privKey PrivKeyEd25519) Bytes() []byte {
...
-func (privKey PrivKeyEd25519) PubKey() crypto.PubKey {
-	privKeyBytes := [64]byte(privKey)
+//
+// Panics if the private key is not initialized.
+func (privKey PrivKey) PubKey() crypto.PubKey {
+	// If the latter 32 bytes of the privkey are all zero, privkey is not
+	// initialized.
 	initialized := false
-	// If the latter 32 bytes of the privkey are all zero, compute the pubkey
-	// otherwise privkey is initialized and we can use the cached value inside
```
tessr, line 65: Is the "cached value" part of this change still true?

marbar3778:
going to break this pr into smaller chunks, closing this for now

Issue #4455, opened 2020-02-21
lite2: backwards should verify headers outside the trustedPeriod


cmwaters:
As long as a `trustedHeader` remains within the `trustedPeriod`, backwards verification using header hashes should be able to verify headers that are outside or earlier in height than the trustedPeriod.

As backwards could potentially verify thousands of blocks, `trustedHeader` should be checked to not have expired at the start and end of the algorithm.

XXX: if it is outside the RemoveNoLongerTrustedHeadersPeriod then the saved header will be removed immediately after saving it

NOTE: could RemoveNoLongerTrustedHeadersPeriod be renamed if its purpose is more about pruning then not trusting the headers anymore

Issue #4425, opened 2020-02-18
lite2: store returns all trusted heights


cmwaters:
With the bisection method and a relative stable validator set you can expect to see large porosity in the trusted store. Many headers are missed and therefore not validated. At the moment `VerifyHeader()` runs the single check of making sure that the height of the `untrustedHeader` is greater than the `trustedHeader` stored in the client. This means that `untrustedHeaders` that are not greater in height to the one stored in the lite client (usually the latest) must be verified via `fetchMissingTrustedHeader()` which then runs the `backwards` verification. I would propose therefore that we use a single method that returns a trusted Header by checking if it is in the store and then if not, using a single verifyHeader function to verify the untrusted Header from the nearest trusted one (either sequence, bisection or backwards). To do so, I would replace these functions of the store interface:

    // LastSignedHeaderHeight returns the last (newest) SignedHeader height.
    //
	// If the store is empty, -1 and nil error are returned.
	LastSignedHeaderHeight() (int64, error

	// FirstSignedHeaderHeight returns the first (oldest) SignedHeader height.
	//
	// If the store is empty, -1 and nil error are returned.
	FirstSignedHeaderHeight() (int64, error)

	// SignedHeaderAfter returns the SignedHeader after the certain height.
	//
	// height must be > 0 && <= LastSignedHeaderHeight.
	SignedHeaderAfter(height int64) (*types.SignedHeader, error)

with 

    // Heights returns all the heights of the headers that the store has. 
    Heights() ([]int64, error)

which in a key-value database should just be all the keys

That way we simplify the interface and we can very quickly check if a header is in the store and if not, which is the closest header we can use to verify the untrusted header from.

PR #4463, opened 2020-02-24
lite2: prune headers

### LINKED ISSUE #4469: lite2: pruning headers and vals
Previously, the light client removed headers based on whether they resided outside the trustPeriod, however it is still possible to use verified headers outside of the trustPeriod, hence we need to modify the function to prune headers.

The pruning will need to occur whenever a new item is added to the database. We will check to see if the size of headers in the database has reached a prescribed height and if so delete the oldest headers and validator sets

cmwaters:
closes #4469 

* [ ] Wrote tests

### DIFF
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -194,19 +194,20 @@ func NewClientFromTrustedStore(
 	options ...Option) (*Client, error) {
 
 	c := &Client{
-		chainID:          chainID,
-		trustingPeriod:   trustingPeriod,
-		verificationMode: skipping,
-		trustLevel:       DefaultTrustLevel,
-		maxRetryAttempts: defaultMaxRetryAttempts,
-		primary:          primary,
-		witnesses:        witnesses,
-		trustedStore:     trustedStore,
-		updatePeriod:     defaultUpdatePeriod,
-		pruningSize:      defaultPruningSize,
-		confirmationFn:   func(action string) bool { return true },
-		quit:             make(chan struct{}),
-		logger:           log.NewNopLogger(),
+		chainID:            chainID,
+		trustingPeriod:     trustingPeriod,
+		verificationMode:   skipping,
+		trustLevel:         DefaultTrustLevel,
+		maxRetryAttempts:   defaultMaxRetryAttempts,
+		primary:            primary,
+		witnesses:          witnesses,
+		trustedStore:       trustedStore,
+		storedHeadersCount: 0,
+		updatePeriod:       defaultUpdatePeriod,
+		pruningSize:        defaultPruningSize,
+		confirmationFn:     func(action string) bool { return true },
+		quit:               make(chan struct{}),
+		logger:             log.NewNopLogger(),
 	}
 
 	for _, o := range options {
@@ -684,7 +685,9 @@ func (c *Client) cleanup(startHeight, stopHeight int64) error {
 			c.logger.Error("can't remove a trusted header & validator set", "err", err, "height", height)
 			continue
 		}
-		c.storedHeadersCount--
+		if c.storedHeadersCount > 0 {
+			c.storedHeadersCount--
+		}
 	}
 
 	c.latestTrustedHeader = nil
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -125,6 +125,8 @@ type Client struct {
 	latestTrustedHeader *types.SignedHeader
 	// Highest next validator set from the store (height=H+1).
 	latestTrustedNextVals *types.ValidatorSet
+	// Amount of headers and vals stored
+	storedHeadersCount uint64
 
 	// See UpdatePeriod option
 	updatePeriod time.Duration
@@ -682,6 +684,7 @@ func (c *Client) cleanup(startHeight, stopHeight int64) error {
 			c.logger.Error("can't remove a trusted header & validator set", "err", err, "height", height)
 			continue
 		}
+		c.storedHeadersCount--
 	}
 
 	c.latestTrustedHeader = nil
@@ -809,14 +812,23 @@ func (c *Client) updateTrustedHeaderAndNextVals(h *types.SignedHeader, nextVals 
 		return errors.Wrap(err, "failed to save trusted header")
 	}
 
+	// prune headers
+	c.storedHeadersCount++
+	if c.storedHeadersCount > c.pruningSize {
+		height, err := c.FirstTrustedHeight()
+		if err != nil {
+			return err
+		}
+		err = c.trustedStore.DeleteSignedHeaderAndNextValidatorSet(height)
+		if err != nil {
+			return err
+		}
+		c.storedHeadersCount--
+	}
+
 	if c.latestTrustedHeader == nil || h.Height > c.latestTrustedHeader.Height {
 		c.latestTrustedHeader = h
 		c.latestTrustedNextVals = nextVals
-	}
-
-	err := c.Prune()
-	if err != nil {
-		return err
 	}
 
 	return nil
@@ -954,37 +966,6 @@ func (c *Client) removeWitness(idx int) {
 		c.witnesses[idx] = c.witnesses[len(c.witnesses)-1]
 		c.witnesses = c.witnesses[:len(c.witnesses)-1]
 	}
-}
-
-// RemoveNoLongerTrustedHeaders removes no longer trusted headers (due to
-// expiration).
-//
-// Exposed for testing.
-func (c *Client) Prune() error {
-	// 1) find the height of the latest header in the store
-	latestHeight, err := c.LastTrustedHeight()
-	if err != nil {
-		return errors.Wrap(err, "unable to prune")
-	}
-
-	// 2) calculate the pruneHeight
-	pruneHeight := latestHeight - int64(c.pruningSize)
-	firstHeight, err := c.FirstTrustedHeight()
-	if err != nil {
-		return errors.Wrap(err, "unable to prune")
-	}
-	// 3) check to see if pruning is necessary
-	if pruneHeight < firstHeight {
-		return nil
-	}
-
-	// 4) prune from firstHeight to pruneHeight
-	err = c.cleanup(firstHeight, pruneHeight)
-	if err != nil {
-		return errors.Wrap(err, "unable to prune")
-	}
-
-	return nil
 }
 
 func (c *Client) autoUpdateRoutine() {
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -698,13 +698,14 @@ func (c *Client) sequence(
 	newHeader *types.SignedHeader,
 	newVals *types.ValidatorSet,
 	now time.Time) error {
+
 	// 1) Verify any intermediate headers.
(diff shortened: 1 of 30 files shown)


lite2/client.go
```diff
@@ -427,7 +427,7 @@ func (c *Client) Stop() {
 //  - header has not been verified yet
```
melekes: //  - header expired, therefore can't be trusted (ErrOldHeaderExpired);

needs to be removed

lite2/client.go
```diff
@@ -870,6 +870,10 @@ func (c *Client) backwards(trustedHeader *types.SignedHeader, newHeader *types.S
 		trustedHeader = interimHeader
```
melekes: do you think it's possible to get rid of `trustedHeader` variable and just use `interimHeader`?

cmwaters: you need at least two headers that change incrementally towards the target newHeader so I don't think that is possible

lite2/client_test.go
```diff
@@ -756,41 +757,135 @@ func TestClientReplacesPrimaryWithWitnessIfPrimaryIsUnavailable(t *testing.T) {
...
 
-	// 2) header is missing, but it's expired => expect error
-	h, err = c.VerifyHeaderAtHeight(1, bTime.Add(1*time.Hour).Add(1*time.Second))
-	assert.Error(t, err)
-	assert.NotNil(t, h)
+		// 5) trusted header has expired => expect error
+		_, err = c.VerifyHeaderAtHeight(1, bTime.Add(4*time.Hour).Add(1*time.Second))
+		assert.Error(t, err)
+	}
+	{
```
golangcibot: 817-852 lines are duplicate of `lite2/client_test.go:853-888` (from `dupl`)

lite2/client_test.go
```diff
@@ -756,41 +757,135 @@ func TestClientReplacesPrimaryWithWitnessIfPrimaryIsUnavailable(t *testing.T) {
...
+		require.NoError(t, err)
 
-	// 4) cannot verify a header in the future
-	_, err = c.VerifyHeaderAtHeight(4, bTime.Add(1*time.Hour).Add(1*time.Second))
-	assert.Error(t, err)
+		// 5) provided header is of a different height => expect error
+		_, err = c.VerifyHeaderAtHeight(2, bTime.Add(1*time.Hour).Add(1*time.Second))
+		assert.Error(t, err)
+	}
+	{
```
golangcibot: 853-888 lines are duplicate of `lite2/client_test.go:817-852` (from `dupl`)

cmwaters:
At the moment, `TrustedHeader` returns the latest header in the store if you pass 0 as the height. As this might not be the actual latest header (one must use `Update` to get the latest) I think we should remove this.

Also from #4455 

NOTE: should we rename RemoveNoLongerTrustedHeadersPeriod as its purpose is more about pruning then not trusting the headers anymore

melekes:
> At the moment, TrustedHeader returns the latest header in the store if you pass 0 as the height. As this might not be the actual latest header (one must use Update to get the latest) I think we should remove this.

I think it's convenient. Otherwise, you're forcing users to call LastTrustedHeight(). Imagine you're running the light client and want to see which is the latest trusted header. What do you do?

> NOTE: should we rename RemoveNoLongerTrustedHeadersPeriod as its purpose is more about pruning then not trusting the headers anymore

Agree.

cmwaters:
> > NOTE: should we rename RemoveNoLongerTrustedHeadersPeriod as its purpose is more about pruning then not trusting the headers anymore
> 
> Agree.

In this case, I would just modify the cleanup to have a `startHeight` and a `stopHeight` and use it for both `Cleanup()` and to prune headers. What do you think?

cmwaters:
Also, I noticed that I have merged a few general verification checks with the backwards verification test function, do you think it's best to extract them out or just leave them there?

melekes:
> In this case, I would just modify the cleanup to have a startHeight and a stopHeight and use it for both Cleanup() and to prune headers. What do you think?

Do what you think is best.

> Also, I noticed that I have merged a few general verification checks with the backwards verification test function, do you think it's best to extract them out or just leave them there?

it's okay to leave them if it's the only place in client.go we have this sequence

lite2/client_test.go
```diff
@@ -335,48 +310,38 @@ func TestClient_SkippingVerification(t *testing.T) {
...
 
-	// Remove expired headers.
-	c.RemoveNoLongerTrustedHeaders(now)
+	err = c.Prune()
+	assert.NoError(t, err)
 
-	// Check expired headers are no longer available.
-	h, err := c.TrustedHeader(1, now)
+	// Header should no longer exist => return Error
+	h, err = c.TrustedHeader(1)
```

lite2/client.go
```diff
@@ -22,9 +22,9 @@ const (
 	sequential mode = iota + 1
 	skipping
 
-	defaultUpdatePeriod                       = 5 * time.Second
-	defaultRemoveNoLongerTrustedHeadersPeriod = 24 * time.Hour
-	defaultMaxRetryAttempts                   = 10
+	defaultUpdatePeriod     = 5 * time.Second
+	defaultPruningAmount    = 1000
```
```suggestion
	defaultPruningSize    = 1000
```

lite2/client.go
```diff
@@ -62,13 +62,14 @@ func UpdatePeriod(d time.Duration) Option {
 	}
 }
 
-// RemoveNoLongerTrustedHeadersPeriod option can be used to define how often
-// the routine, which cleans up no longer trusted headers (outside of trusting
-// period), is run. Default: once a day. When set to zero, the routine won't be
-// started.
-func RemoveNoLongerTrustedHeadersPeriod(d time.Duration) Option {
+// PruningSize option sets the maximum amount of Headers and Validator sets that
```
```suggestion
// PruningSize option sets the maximum amount of headers and validator sets that
```

lite2/client.go
```diff
@@ -128,7 +129,7 @@ type Client struct {
 	// See UpdatePeriod option
 	updatePeriod time.Duration
 	// See RemoveNoLongerTrustedHeadersPeriod option
-	removeNoLongerTrustedHeadersPeriod time.Duration
+	pruningSize int64
```
melekes: uint32 is probably enough

lite2/client.go
```diff
@@ -387,9 +391,9 @@ func (c *Client) initializeWithTrustOptions(options TrustOptions) error {
 func (c *Client) Start() error {
 	c.logger.Info("Starting light client")
 
-	if c.removeNoLongerTrustedHeadersPeriod > 0 {
+	if c.pruningSize > 0 && c.updatePeriod > 0 {
```
melekes: don't understand why pruning should depend on autoUpdate

lite2/client.go
```diff
@@ -570,18 +559,17 @@ func (c *Client) VerifyHeaderAtHeight(height int64, now time.Time) (*types.Signe
 // validator set at height newHeader.Height+1 (i.e.
 // newHeader.NextValidatorsHash).
 func (c *Client) VerifyHeader(newHeader *types.SignedHeader, newVals *types.ValidatorSet, now time.Time) error {
-	h, err := c.TrustedHeader(newHeader.Height, now)
-	switch err.(type) {
-	case nil: // Return already trusted header
-		// Make sure it's the same header.
+	if newHeader.Height <= 0 {
+		return errors.New("negative or zero height")
+	}
```
```suggestion
	}

```

lite2/client.go
```diff
@@ -674,11 +663,23 @@ func (c *Client) cleanup(stopHeight int64) error {
...
+	if startHeight < oldestHeight {
+		startHeight = oldestHeight
 	}
-	for height := stopHeight; height <= latestHeight; height++ {
+	if startHeight > stopHeight {
+		return errors.Errorf("startHeight of %d must be less than stopHeight %d", startHeight, stopHeight)
+	}
+
+	for height := startHeight; height <= stopHeight; height++ {
+		// 4a)
```

lite2/client.go
```diff
@@ -655,13 +643,14 @@ func (c *Client) Witnesses() []provider.Provider {
...
 }
 
-// cleanup deletes all headers & validator sets between +stopHeight+ and latest
-// height included. It also sets trustedHeader (vals) to the latest header
-// (vals) if such exists.
-func (c *Client) cleanup(stopHeight int64) error {
+// cleanup deletes all headers & validator sets between startHeight and stopHeight (inclusive).
+// Using a height of 0 will default to the oldest and newest heights respectively.
+// It also resets latestTrustedHeader to the latest header.
+func (c *Client) cleanup(startHeight, stopHeight int64) error {
```
melekes: does stopHeight ever change? looks like it's 0 in both cases

lite2/client.go
```diff
@@ -944,16 +954,19 @@ func (c *Client) removeWitness(idx int) {
 	}
 }
 
-func (c *Client) removeNoLongerTrustedHeadersRoutine() {
+func (c *Client) pruneRoutine() {
 	defer c.routinesWaitGroup.Done()
 
-	ticker := time.NewTicker(c.removeNoLongerTrustedHeadersPeriod)
+	ticker := time.NewTicker(c.updatePeriod)
```
melekes: maybe instead of the routine we can use some kind of a counter in VerifyHeader, which we'll use to detect when we need to call Prune

lite2/client.go
cmwaters: the `Prune()` func has a different stopHeight

lite2/client.go
cmwaters: Yes I would agree with this implementation. Should I rewrite it then?

melekes: Yes. It would've been better to do it in a separate PR (leaving only expiration checks here), but I guess it's fine

cmwaters: I could create a separate issue and link it to this PR so we better evidence of what this PR did?

lite2/client.go
cmwaters: but I have to convert it to int64 when using it:

`pruneHeight := latestHeight - c.pruningSize`

melekes: `uint64` then. converting is fine

cmwaters:
@melekes I think that the lite client should keep a count of how many headers have been stored and if it reaches a defined level then every new header we delete the oldest header in the database. 

At the moment we delete based on height not on the pure amount of headers in the database which I think is more relevant.

cmwaters:
The lite client has a counter and uses this to ensure that the amount of headers it has doesn't reach a defined amount else it prunes the oldest headers. The problem with this is that if you initiate a lite client with a non-empty database it currently isn't aware that there are already headers there. I could add a count function to see how many headers are already stored but that could be computationally costly

cmwaters:
^ It might be better to have a method from the store interface that could tell us how many headers there are. It would be cleaner, more accurate and not involve checking for negatives. I motioned towards something like this here: #4425

melekes:
> I could add a count() function to see how many headers are already stored but that could be computationally costly

What about storing a counter in a database?

lite2/client.go
melekes: Please do so 🙏

PR #4466, opened 2020-02-24
indexer: allow to index an event at runtime

### LINKED ISSUE #3437: Allow setting tags to index at runtime
I am developing an Tendermint application that supports user-uploaded smart contracts. A smart contract is allowed to emit tags, specifying some to be indexed, some not. Indexed tags would cost more gas but useful for that smart contract's clients.

Currently, AFAIK, the only way to allow this is config to index _all_ tags, which is not optimal. I would be great if Tendermint let applications specify additional tags to index on deliver_tx.

favadi:
The PR added a new field `index` to event attribute, that will cause indexer service to index the event if set to true.

Fixes #3437.

* [x] Wrote tests


rpc/client/rpc_test.go
```diff
@@ -494,6 +494,13 @@ func TestTxSearch(t *testing.T) {
 			t.Fatal("expected a lot of transactions")
 		}
 
+		// query using an index key
+		result, err = c.TxSearch("app.index_key='index is working'", false, 1, 30, "asc")
```
erikgrinaker: It would be nice to test the negative as well, i.e. that a field with `Index: false` does not return results.

favadi: added a test case for that.

favadi:
> LGTM. Could you add a changelog entry please?

Changelog added.

libs/kv/types.proto
```diff
@@ -20,6 +20,7 @@ option (gogoproto.testgen_all)  = true;
 message Pair {
   bytes key   = 1;
   bytes value = 2;
+  bool  index = 3;
```
melekes: Pair is supposed to be an abstract type representing key-value pair. Adding the third field makes it a specific type. Therefore, we need to

a) make sure we don't use Pair anywhere else
b) rename it to IndexPair or something like that

melekes: Also, what about `KI64Pair`? Do we need to change it too?

favadi: > a) make sure we don't use Pair anywhere else
>  b) rename it to IndexPair or something like that

I did a reviewed and Pair is used outside of the index service. Therefore I created a new IndexPair instead of renaming it.

> Also, what about KI64Pair? Do we need to change it too?

I have no idea what it is and deleting it doesn't seem to break the test suite, so it is probably used nowhere.

melekes: > Also, what about KI64Pair?

It was removed.

melekes: > I did a reviewed and Pair is used outside of the index service.

where?

docs/app-dev/indexing-transactions.md
```diff
@@ -80,6 +80,7 @@ func (app *KVStoreApplication) DeliverTx(req types.RequestDeliverTx) types.Resul
                 kv.Pair{Key: []byte("sender"), Value: []byte("Bob")},
                 kv.Pair{Key: []byte("recipient"), Value: []byte("Alice")},
                 kv.Pair{Key: []byte("balance"), Value: []byte("100")},
+                kv.Pair{Key: []byte("note"), Value: []byte("nothing"), Index: true},
```
melekes: IndexPair

melekes:
`EventAttribute` name makes more sense to me than `IndexPair`. Pair was an abstract concept. That's why it was located in libs. Now we're adding implementation details to this type. That's why it makes sense to move it to `abci/types.proto`.

docs/app-dev/indexing-transactions.md
favadi: fixed in latest commit.

favadi:
@melekes I moved the `IndexPair` to abci/types.proto and renamed to `EventAttribute`.  The `kv.Pair` is still required as it is (at least) being used in [crypto/merkle](https://github.com/tendermint/tendermint/blob/dfebac86f7120818648e784b96c31ae9c1673b3c/crypto/merkle/simple_map.go#L36).

libs/kv/types.proto
```diff
@@ -17,6 +17,6 @@ option (gogoproto.testgen_all)  = true;
 // Abstract types
 
 message Pair {
-  bytes key   = 1;
```
melekes: is this due to you running `make proto-lint`?

favadi: Yes, it is.

libs/kv/types.proto
marbar3778: Do we need kv.Pair, it was primarily used for indexing? Looking through the repo it seems we should remove it and have the new type implement the sorting interface.

favadi:
PR rebased.

marbar3778:
@melekes @erikgrinaker what do you guys think about renaming `libs/kv` to `libs/event` and bringing back the type to that directory? Otherwise we should remove `kv.Pair` as it is not used anywhere and the new type needs to implment the sorting interface as it is here: https://github.com/tendermint/tendermint/blob/master/libs/kv/kvpair.go

melekes:
I like that `EventAttribute` is located in the same place where `Event` is.

erikgrinaker:
Now that 0.33.4 is out the door I think we can finally get this merged. Would you mind updating it with `master` @favadi?

favadi:
@erikgrinaker I rebased the PR on master, tests are passing.

erikgrinaker:
Fantastic, thank you!

PR #4467, opened 2020-02-24
In seed mode, sleep for 1 prior to flush stop after sending addrs

### LINKED ISSUE #3778: p2p: Possible issue with seed mode
**Tendermint version**: v0.32.0

**ABCI app**: `kvstore`

**Environment**:
- **OS** (e.g. from /etc/os-release): CentOS
- **Install tools**:
- **Others**:

**What happened**:
I tried setting up a single seed node (non-validator) with 4 validators on AWS (2 nodes in `us-east-1`, 1 in `us-east-2` and 2 in `us-west-1`). No `persistent_peers` configured, and all validators were pointed to the single seed node.

Even after several minutes, none of the validators could connect to each other.

Setting `seed_mode = false` on the seed node resolved the issue 🤔 

I would've expected the seed node to facilitate the validators connecting to each other.

Using the [tmtestnet.py](https://github.com/interchainio/testnets) script/tool I've been working on with [this testnet configuration file](https://gist.github.com/thanethomson/8700eb9821f189ff5b0fd00fb461c495#file-seeded-network-yaml), [`my-seeds-config.toml`](https://gist.github.com/thanethomson/8700eb9821f189ff5b0fd00fb461c495#file-my-seeds-config-toml) and [`my-validators-config.toml`](https://gist.github.com/thanethomson/8700eb9821f189ff5b0fd00fb461c495#file-my-validators-config-toml), run:

```bash
# Clone https://github.com/interchainio/testnets and run from root of that repo
./tmtestnet.py -c seeded-network.yaml -v network deploy
```

See [seed node logs](https://gist.github.com/thanethomson/8700eb9821f189ff5b0fd00fb461c495#file-seed-node-logs) and [validator 0 logs](https://gist.github.com/thanethomson/8700eb9821f189ff5b0fd00fb461c495#file-validator0-logs)

See the above gist.

From what I can see, it seems as though this is happening:
1. The validators hit the seed node and find no peers, so they disconnect immediately (to retry later)
2. The seed node adds a peer as it connects, but an error's triggered (use of closed network connection) and it removes the peer (keeping the address book empty).

gterzian:
FIX #3778

* [ ] Wrote tests

### DIFF
--- a/p2p/pex/pex_reactor.go
+++ b/p2p/pex/pex_reactor.go
@@ -264,6 +264,8 @@ func (r *Reactor) Receive(chID byte, src Peer, msgBytes []byte) {
 			r.SendAddrs(src, r.book.GetSelectionWithBias(biasToSelectNewPeers))
 			go func() {
 				// In a go-routine so it doesn't block .Receive.
+				// Sleep for one min to allow the peer to handle the addrs.
+				time.Sleep(1 * time.Minute)
 				src.FlushStop()
 				r.Switch.StopPeerGracefully(src)
 			}()


gterzian:
@thanethomson I think this might be a solution, since it appears like the seed node disconnects before the validator has a chance to handle the addr response. 

The seed node does a `FlushStop`, however I think that doesn't mean the other peer has had a chance to actually handle the message, and the `EOF` error in the `Connection.recvRoutine` might then disconnect the seednode before the validator has had a chance to handle the message. 

Is it very easy for you to try out this branch with your deploy?

p2p/pex/pex_reactor.go
```diff
@@ -264,6 +264,8 @@ func (r *Reactor) Receive(chID byte, src Peer, msgBytes []byte) {
 			r.SendAddrs(src, r.book.GetSelectionWithBias(biasToSelectNewPeers))
 			go func() {
 				// In a go-routine so it doesn't block .Receive.
+				// Sleep for one min to allow the peer to handle the addrs.
+				time.Sleep(1 * time.Minute)
```
thanethomson: Will this effectively allow peers to send multiple `pexRequestMessage`s during the 1 minute time interval, potentially spawning multiple goroutines that will attempt to stop the same peer? If so, what's the impact of that on the system?

thanethomson:
> Is it very easy for you to try out this branch with your deploy?

It's unfortunately not a simple task to test this branch out. The setup and experimentation time is significant.

p2p/pex/pex_reactor.go
gterzian: Thanks for asking, now that I've looked into it, I think the situation is the same as before, and that only on "stop goroutine" will be spawned per peer, see the check a bit above at https://github.com/tendermint/tendermint/pull/4467/files#diff-8c6b1a350a161ced01d68514ccda7484R256

melekes:
> the situation is the same as before

can you think of any other ideas on how to fix the issue?

gterzian:
> can you think of any other ideas on how to fix the issue?

What do you mean? When I wrote "the situation is the same as before", I meant it positively, as in the sleep in the goroutine wouldn't seem to allow a peer to send many `pexRequestMessage` messages during that period.

melekes:
> When I wrote "the situation is the same as before", I meant it positively

Oh, sorry then. I thought you mean the solution does not work.

tessr:
I'd like to change the name of this PR (and by extension, the resulting commit) to better match the [Go commit message standard](https://tip.golang.org/doc/contribute.html#commit_messages). Would it be accurate to say something like `p2p/pex: sleep to ensure peers can connect through seed node`?

tessr:
(Also, I see that @cmwaters was just assigned - Callum, is your plan to try deploying this and testing it out?)

p2p/pex/pex_reactor.go
```diff
@@ -264,6 +264,8 @@ func (r *Reactor) Receive(chID byte, src Peer, msgBytes []byte) {
 			r.SendAddrs(src, r.book.GetSelectionWithBias(biasToSelectNewPeers))
 			go func() {
 				// In a go-routine so it doesn't block .Receive.
+				// Sleep for one min to allow the peer to handle the addrs.
```
melekes: could you describe what/how are we handling the addrs exactly?

cmwaters:
> (Also, I see that @cmwaters was just assigned - Callum, is your plan to try deploying this and testing it out?)

@tessr I looked into deploying a testnet to test the changes but a) I don't have a AWS account (not sure if that's critical though) and b) I don't have a InfluxDB account which is, I believe, required. Is it possible to test this all locally?

melekes:
@gterzian see https://github.com/tendermint/tendermint/issues/3778#issuecomment-598181291.  thank you again for attempting to fix the issue! And feel free to add your thoughts there. We appreciate your input.

PR #4472, opened 2020-02-25
lite2: store current validator set

### LINKED ISSUE #4399: lite2: inconsistent header / validator set pairing
When using the lite client there are two types of pairings that functions use. The header at height h and the validator set at height h _(seen in green)_ and the header at height h and the validator set at height h + 1 _(seen in red)_

header-validator-dilemma

This can be confusing as for example, saving or deleting a header and validator set to the lite clients database using the pairing in red, whilst retrieving a header and validator set (`fetchHeaderAndValsAtHeight`) uses the green pairing. The `bisection` method requires the red pairing for the trusted header but requires the green pairing for the untrusted header. Is it possible to have a more consistent pairing between the validator set and the header?

### LINKED ISSUE #4470: lite2: consider storing current valset
"We need the current ValidatorSet of the header so that the connecting chain can verify the header was committed by the current ValidatorSet. The lite.Verify function takes in the current ValidatorSet which gets used in the IBC update function. The relayer needs some way to pull the current validator set from the liteclient db and then pass it to the connecting chain
https://github.com/cosmos/cosmos-sdk/blob/bb7e803cb9291ff408a2d8ceab29a6292537012b/x/ibc/07-tendermint/update.go#L87"

before: {Header} + {NextValset}
after: {Header} + {Valset} + {NextValset}

melekes:
Before we were storing trustedHeader (height=1) and trustedNextVals
(height=2).

After this change, we will be storing trustedHeader (height=1) and
trustedVals (height=1). This a) simplifies the code b) fixes #4399
inconsistent pairing issue c) gives a relayer access to the current
validator set #4470.

The only downside is more jumps during bisection. If validator set
changes between trustedHeader and the next header (by 2/3 or more), the
light client will be forced to download the next header and check that
2/3+ signed the transition. But we don't expect validator set change too
much and too often, so it's an acceptable compromise.

Closes #4470 and #4399

* [ ] ~~Wrote tests~~

### DIFF
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -122,8 +122,8 @@ type Client struct {
 	trustedStore store.Store
 	// Highest trusted header from the store (height=H).
 	latestTrustedHeader *types.SignedHeader
-	// Highest next validator set from the store (height=H+1).
-	latestTrustedNextVals *types.ValidatorSet
+	// Highest validator set from the store (height=H).
+	latestTrustedVals *types.ValidatorSet
 
 	// See UpdatePeriod option
 	updatePeriod time.Duration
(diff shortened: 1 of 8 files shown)


lite2/client.go
```diff
@@ -699,73 +686,71 @@ func (c *Client) cleanup(stopHeight int64) error {
...
-					interimHeader.NextValidatorsHash,
-					interimNextVals.Hash(),
-					interimHeader.Height)
-			}
-		}
-		trustedHeader, trustedNextVals = interimHeader, interimNextVals
+		// 3) Update trustedHeader and trustedVals
+		// NOTE: we're updating the c.latestTrustedHeader and c.latestTrustedVals
+		// indirectly here because trustedHeader/Vals are pointers.
+		trustedHeader, trustedVals = interimHeader, interimVals
```
golangcibot: ineffectual assignment to `trustedVals` (from `ineffassign`)

lite2/client.go
```diff
@@ -699,73 +686,71 @@ func (c *Client) cleanup(stopHeight int64) error {
 // see VerifyHeader
 func (c *Client) sequence(
 	trustedHeader *types.SignedHeader,
-	trustedNextVals *types.ValidatorSet,
+	trustedVals *types.ValidatorSet,
```
golangcibot: SA4009: argument trustedVals is overwritten before first use (from `staticcheck`)

PR #4478, opened 2020-02-26
lite2: prune-headers

### LINKED ISSUE #4469: lite2: pruning headers and vals

cmwaters:
closes #4469

Improved speed of cleanup by using `SignedHeaderAfter` instead of `TrustedHeader` to jump from header to header.

`Prune()` is now called when a new header and validator set are saved and is a function dealt by the database itself

### DIFF
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -22,9 +22,9 @@ const (
 	sequential mode = iota + 1
 	skipping
 
-	defaultUpdatePeriod                       = 5 * time.Second
-	defaultRemoveNoLongerTrustedHeadersPeriod = 24 * time.Hour
-	defaultMaxRetryAttempts                   = 10
+	defaultUpdatePeriod     = 5 * time.Second
+	defaultPruningSize      = 1000
+	defaultMaxRetryAttempts = 10
 )
 
 // Option sets a parameter for the light client.
@@ -62,13 +62,14 @@ func UpdatePeriod(d time.Duration) Option {
 	}
 }
 
-// RemoveNoLongerTrustedHeadersPeriod option can be used to define how often
-// the routine, which cleans up no longer trusted headers (outside of trusting
-// period), is run. Default: once a day. When set to zero, the routine won't be
-// started.
-func RemoveNoLongerTrustedHeadersPeriod(d time.Duration) Option {
+// PruningSize option sets the maximum amount of headers and validator sets that
+// the light client stores. When Prune() is run, all headers that are earlier than
+// the h amount of headers will be removed from the store. If UpdatePeriod is used,
+// prune will automatically run. Default: 1000. A pruning size of 0 will not prune
+// the lite client at all.
+func PruningSize(h uint64) Option {
 	return func(c *Client) {
-		c.removeNoLongerTrustedHeadersPeriod = d
+		c.pruningSize = h
 	}
 }
 
@@ -128,7 +129,7 @@ type Client struct {
 	// See UpdatePeriod option
 	updatePeriod time.Duration
 	// See RemoveNoLongerTrustedHeadersPeriod option
-	removeNoLongerTrustedHeadersPeriod time.Duration
+	pruningSize uint64
 	// See ConfirmationFunction option
 	confirmationFn func(action string) bool
 
@@ -191,19 +192,19 @@ func NewClientFromTrustedStore(
 	options ...Option) (*Client, error) {
 
 	c := &Client{
-		chainID:                            chainID,
-		trustingPeriod:                     trustingPeriod,
-		verificationMode:                   skipping,
-		trustLevel:                         DefaultTrustLevel,
-		maxRetryAttempts:                   defaultMaxRetryAttempts,
-		primary:                            primary,
-		witnesses:                          witnesses,
-		trustedStore:                       trustedStore,
-		updatePeriod:                       defaultUpdatePeriod,
-		removeNoLongerTrustedHeadersPeriod: defaultRemoveNoLongerTrustedHeadersPeriod,
-		confirmationFn:                     func(action string) bool { return true },
-		quit:                               make(chan struct{}),
-		logger:                             log.NewNopLogger(),
+		chainID:          chainID,
+		trustingPeriod:   trustingPeriod,
+		verificationMode: skipping,
+		trustLevel:       DefaultTrustLevel,
+		maxRetryAttempts: defaultMaxRetryAttempts,
+		primary:          primary,
+		witnesses:        witnesses,
+		trustedStore:     trustedStore,
+		updatePeriod:     defaultUpdatePeriod,
+		pruningSize:      defaultPruningSize,
+		confirmationFn:   func(action string) bool { return true },
+		quit:             make(chan struct{}),
+		logger:           log.NewNopLogger(),
 	}
 
 	for _, o := range options {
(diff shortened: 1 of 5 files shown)


lite2/client.go
```diff
@@ -62,13 +62,14 @@ func UpdatePeriod(d time.Duration) Option {
 	}
 }
 
-// RemoveNoLongerTrustedHeadersPeriod option can be used to define how often
-// the routine, which cleans up no longer trusted headers (outside of trusting
-// period), is run. Default: once a day. When set to zero, the routine won't be
-// started.
-func RemoveNoLongerTrustedHeadersPeriod(d time.Duration) Option {
+// PruningSize option sets the maximum amount of headers and validator sets that
```
melekes: we should be clear that header and validator set count for 1

lite2/client.go
```diff
@@ -62,13 +62,14 @@ func UpdatePeriod(d time.Duration) Option {
...
+// the light client stores. When Prune() is run, all headers that are earlier than
+// the h amount of headers will be removed from the store. If UpdatePeriod is used,
```
melekes: I don't think we should combine these two. I mean it totally makes sense to have pruning when you have auto update (default should be: pruning + auto update), but maybe leave the option to not have pruning even if auto update is enabled

lite2/client.go
```diff
@@ -304,7 +305,10 @@ func (c *Client) checkTrustedHeaderUsingOptions(options TrustOptions) error {
 			c.latestTrustedHeader.Height, c.latestTrustedHeader.Hash())
 		if c.confirmationFn(action) {
 			// remove all the headers (options.Height, trustedHeader.Height]
-			c.cleanup(options.Height + 1)
+			err := c.cleanup(options.Height+1, 0)
+			if err != nil {
+				return errors.Wrap(err, "unable to check trusted header using options")
```

lite2/client.go
```diff
@@ -649,15 +650,34 @@ func (c *Client) cleanup(stopHeight int64) error {
...
 	}
 
-	// 3) Remove all headers and validator sets.
-	if stopHeight < oldestHeight {
-		stopHeight = oldestHeight
+	// 3) Ensure the start height and stop height are within the range of headers stored.
+	if stopHeight == 0 || stopHeight > latestHeight {
+		stopHeight = latestHeight
+	}
+	if startHeight < oldestHeight {
```
melekes: what about startHeight == 0?

lite2/client.go
```diff
@@ -649,15 +650,34 @@ func (c *Client) cleanup(stopHeight int64) error {
...
+				startHeight)
+		}
+		// 4b) Jump to the next height
+		if startHeight < stopHeight {
+			h, err := c.trustedStore.SignedHeaderAfter(startHeight)
+			if err != nil {
+				return errors.Wrap(err, "tried to clean up headers")
+			}
+			startHeight = h.Height
+		} else {
```
melekes: does this mean there will be only one header left == stopHeight?

lite2/client.go
```diff
@@ -776,7 +796,10 @@ func (c *Client) updateTrustedHeaderAndVals(h *types.SignedHeader, vals *types.V
 		return errors.Wrap(err, "failed to save trusted header")
 	}
 
-	// Only update latestTrustedHeader if we move forward (not backwards).
+	if err := c.trustedStore.Prune(c.pruningSize); err != nil {
+		return errors.Wrap(err, "failed to save trusted header")
```

lite2/client.go
cmwaters: if startHeight is 0 than implicitly it is less than oldestHeight

lite2/client.go
cmwaters: lol, did not see that

lite2/client.go
cmwaters: yup

lite2/client.go
cmwaters: what's wrong with this one

lite2/client.go
cmwaters: sure, a header and validator set set

lite2/client.go
cmwaters: That's possible if you have a pruning size of 0 and an updatePeriod > 0

lite2/client.go
cmwaters: what's wrong with this?

lite2/client.go
```diff
@@ -62,13 +62,14 @@ func UpdatePeriod(d time.Duration) Option {
...
 }
 
-// RemoveNoLongerTrustedHeadersPeriod option can be used to define how often
-// the routine, which cleans up no longer trusted headers (outside of trusting
-// period), is run. Default: once a day. When set to zero, the routine won't be
-// started.
-func RemoveNoLongerTrustedHeadersPeriod(d time.Duration) Option {
+// PruningSize option sets the maximum amount of headers & validator set pairs
+// that the light client stores. When Prune() is run, all headers (along with
+// the associated validator set) that are earlier than the h amount of headers
```
```suggestion
// the associated validator sets) that are earlier than the h amount of headers
```

lite2/client.go
```diff
@@ -649,15 +650,34 @@ func (c *Client) cleanup(stopHeight int64) error {
...
+	if stopHeight == 0 || stopHeight > latestHeight {
+		stopHeight = latestHeight
+	}
+	if startHeight < oldestHeight {
+		startHeight = oldestHeight
 	}
-	for height := stopHeight; height <= latestHeight; height++ {
-		err = c.trustedStore.DeleteSignedHeaderAndValidatorSet(height)
+	if startHeight > stopHeight {
+		return errors.Errorf("startHeight of %d must be less than stopHeight %d", startHeight, stopHeight)
```
melekes: Is this a developer error?

lite2/client.go
```diff
@@ -630,13 +630,14 @@ func (c *Client) Witnesses() []provider.Provider {
 // client must be stopped at this point.
 func (c *Client) Cleanup() error {
 	c.logger.Info("Removing all the data")
-	return c.cleanup(0)
+	return c.cleanup(0, 0)
```
melekes: we can now use `trustedStore.Prune(0)` function

lite2/client.go
```diff
@@ -649,15 +650,34 @@ func (c *Client) cleanup(stopHeight int64) error {
...
+	if startHeight < oldestHeight {
+		startHeight = oldestHeight
 	}
-	for height := stopHeight; height <= latestHeight; height++ {
-		err = c.trustedStore.DeleteSignedHeaderAndValidatorSet(height)
+	if startHeight > stopHeight {
+		return errors.Errorf("startHeight of %d must be less than stopHeight %d", startHeight, stopHeight)
+	}
+
+	for startHeight <= stopHeight {
```
melekes: I think this is a bad patter of calling something start and then mutating it

lite2/client.go
cmwaters: Yup we could

lite2/client.go
cmwaters: yes, the function could work without it - perhaps log it as an error or we can also remove it

lite2/client.go
cmwaters: sure the name and the purpose aren't aligned. Perhaps you could create another variable just called height and assign startHeight to it initially before it begins to be mutated

PR #4479, opened 2020-02-26
lite2: return height as 2nd return param in TrustedValidatorSet

### LINKED ISSUE #4473: lite2: modify TrustedValidatorSet func to return height
If we're going to allow 0 height as an argument, returning height as a second param seems reasonable.

```
TrustedValidatorSet(0) => (*ValidatorSet, 10, nil)
TrustedValidatorSet(9) => (*ValidatorSet, 9, nil)
```

melekes:
Closes #4473

* [x] Wrote tests

### DIFF
--- a/lite2/client_test.go
+++ b/lite2/client_test.go
@@ -395,7 +395,7 @@ func TestClientRestoresTrustedHeaderAfterStartup1(t *testing.T) {
 		assert.NoError(t, err)
 		if assert.NotNil(t, h) {
 			assert.Equal(t, h.Hash(), header1.Hash())
-    }
+		}
 
 		valSet, _, err := c.TrustedValidatorSet(1)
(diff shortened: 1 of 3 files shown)


lite2/client_test.go
```diff
@@ -904,3 +963,28 @@ func TestClientRemovesWitnessIfItSendsUsIncorrectHeader(t *testing.T) {
...
+		[]provider.Provider{fullNode},
+		dbs.New(dbm.NewMemDB(), chainID),
+		UpdatePeriod(0),
+		Logger(log.TestingLogger()),
+	)
+
+	require.NoError(t, err)
+	err = c.Start()
+	require.NoError(t, err)
+	defer c.Stop()
```
cmwaters: why do you use start and stop funcs in this test?

PR #4481, opened 2020-02-27
adr: crypto encoding for proto

marbar3778:
- this adr is meant to help with deciding on how to move forward with keys in tendermint.

ref #2204
ref #4208

* [ ] Wrote tests


docs/architecture/adr-054-crypto-encoding-2.md
```diff
@@ -0,0 +1,67 @@
...
+1. Remove backwards compatibility and go with `oneof` proto encoding of keys.
+
+- This will be a major breaking change in the ecosystem.
+- Will need a migration script to be released in conjunction with the release which will have this change.
+
+> I am leaning towards the first option and go through the with headache in order to entirely remove the amino overhead
+
+2. Keep the amino encoding format `<PrefixBytes> <Length> <ByteArray>` and protobuf encode the key after appending the prefix bytes.
+
+- This may cause issues with bech32 as it will exceed the maximum length permitted by the libraries
```
Liamsi: I'm pretty sure we've already experienced problems with this approach (at least for bechified pub keys).

docs/architecture/adr-054-crypto-encoding-2.md
```diff
@@ -0,0 +1,67 @@
...
+> With keys becoming a [oneof type](https://developers.google.com/protocol-buffers/docs/proto3#oneof) we have the option of removing the prefix bytes.
+
+## Options
+
+If there are more options please leave a comment and we can discuss it.
+
+1. Remove backwards compatibility and go with `oneof` proto encoding of keys.
+
+- This will be a major breaking change in the ecosystem.
+- Will need a migration script to be released in conjunction with the release which will have this change.
```
Liamsi: IMHO this is the cleanest option. To go this way we'd need to make sure that we do not use the amino encoded bytes anywhere user facing (e.g. when bechifying: are we using the raw bytes or the prefixed bytes?).  Even if there are cases where this is user facing, it might be worth exploring.

docs/architecture/adr-054-crypto-encoding-2.md
```diff
@@ -0,0 +1,67 @@
...
+
+2. Keep the amino encoding format `<PrefixBytes> <Length> <ByteArray>` and protobuf encode the key after appending the prefix bytes.
+
+- This may cause issues with bech32 as it will exceed the maximum length permitted by the libraries
+- Backwards compatible, will not keep amino around but use custom marshlers in order to prefix the keys, less of a headache as no migration script would be needed and less coordination in the community.
+
+3. Use proto encoding for over the wire communication. Where backwards compatibility is needed use the amino encoding format: `<PrefixBytes> <Length> <ByteArray>`.
+
+- This is a good proposal but would be a bit hacky. 
+- Less headache, users will be happy for not needing to do a migration.
```
Liamsi: This is the way to go, if option 1 does not work because backwards compatibility is critical.

docs/architecture/adr-054-crypto-encoding-2.md
```diff
@@ -0,0 +1,67 @@
+# ADR 054: Crypto encoding (part 2)
+
+## Changelog
+
+\*2020-2-27: Created
+
+## Context
+
+Amino has been a pain point of many users in the ecosystem. While Tendermint does not suffer greatly from the performance degradation introduced by amino, we are making an effort in moving the encoding format to a widely adopt format, [Protocol Buffers](https://developers.google.com/protocol-buffers). With this migration a new standard is needed for the encoding of keys. This will cause ecosystem wide breaking changes.
```
tarcieri: nit: `adopt` -> `adopted`

docs/architecture/adr-054-crypto-encoding-2.md
tarcieri: I'm +1 on this approach. My past experiences with doing this sort of migration in Protos are all based on additive changes and work more or less like the following:

1. Add a new field with the proto encoding, supporting both the proto encoding and amino encoding as you just described
2. Have new code use the new proto encoding as canonical, only producing the old amino encoding for legacy message consumers
3. When you are sure that all message consumers using the legacy amino format around, then it can be phased out

docs/architecture/adr-054-crypto-encoding-2.md
```diff
@@ -0,0 +1,85 @@
...
+
+\*2020-2-27: Created
+
+## Context
+
+Amino has been a pain point of many users in the ecosystem. While Tendermint does not suffer greatly from the performance degradation introduced by amino, we are making an effort in moving the encoding format to a widely adopted format, [Protocol Buffers](https://developers.google.com/protocol-buffers). With this migration a new standard is needed for the encoding of keys. This will cause ecosystem wide breaking changes.
+
+Currently amino encodes keys as `<PrefixBytes> <Length> <ByteArray>`.
+
+> With keys becoming a [oneof type](https://developers.google.com/protocol-buffers/docs/proto3#oneof) we have the option of removing the prefix bytes.
```
tessr: What _are_ the prefix bytes?

docs/architecture/adr-054-crypto-encoding-2.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Decision
+
+Transitioning from a fixed size byte array to bytes would be the first step. This will enable usage of [cosmos-proto](https://github.com/regen-network/cosmos-proto) interface type. This removes boiler plate needed for oneof types.
```
tessr: When you say "bytes," do you mean a byte slice?

docs/architecture/adr-054-crypto-encoding-2.md
```diff
@@ -0,0 +1,85 @@
...
+
+The approach that will be taken to minimize headaches for users is one where all encoding of keys will shift to protobuf and where amino encoding is relied on, there will be custom marshal and unmarshal functions.
```
tessr: Maybe out of scope for this document, but do we have a list of places where we'll still need to use amino?

tessr: Ah, I see the comment on this below 😅

docs/architecture/adr-054-crypto-encoding-2.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Positive
+
+- Protocol Buffer encoding will not change going forward.
+- Removing amino overhead from keys will help with the KSM.
+- Have a large ecosystem of supported languages.
+
+### Negative
+
+-
```
tessr: The main negative point seems to be that it will require a hardfork upgrade, right?

go.sum
```diff
@@ -36,6 +36,7 @@ github.com/beorn7/perks v1.0.0 h1:HWo1m869IqiPhD389kmkxeTalrjNbbJTC8LXupb+sl0=
 github.com/beorn7/perks v1.0.0/go.mod h1:KWe93zE9D1o94FZ5RNwFwVgaQK1VOXiVxmqh+CedLV8=
 github.com/beorn7/perks v1.0.1 h1:VlbKKnNfV8bJzeqoa4cOKqO6bYr3WgKZxO8Z16+hsOM=
 github.com/beorn7/perks v1.0.1/go.mod h1:G2ZrVWU2WbWT9wwq4/hrbKbnv/1ERSJQ0ibhJ6rlkpw=
+github.com/bgentry/speakeasy v0.1.0 h1:ByYyxL9InA1OWqxJqqp2A5pYHUrCiAL6K3J+LKSsQkY=
```
tessr: why are these modules updates also in this PR?

marbar3778: dependabot doesn't run `go mod tidy` so when I run it these changes occur. I automatically do it at this point, need to break that habit

docs/architecture/adr-054-crypto-encoding-2.md
marbar3778: yes will detail that in the paragraph, thanks

docs/architecture/adr-054-crypto-encoding-2.md
marbar3778: i did a dive into the code where we encode keys and how we do it. Currently I have not found any reliance on amino encoded bytes but would like to spend sometime doublechecking as i go through the work

docs/architecture/adr-054-crypto-encoding-2.md
marbar3778: removed this comment as it was relevant before when I was debating on keeping the prefix bytes, but that was based on speaking with people about encoded keys. Looking through the code, this didn't seem like the case.

docs/architecture/adr-054-crypto-encoding-2.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Context
+
+Amino has been a pain point of many users in the ecosystem. While Tendermint does not suffer greatly from the performance degradation introduced by amino, we are making an effort in moving the encoding format to a widely adopted format, [Protocol Buffers](https://developers.google.com/protocol-buffers). With this migration a new standard is needed for the encoding of keys. This will cause ecosystem wide breaking changes.
+
+Currently amino encodes keys as `<PrefixBytes> <Length> <ByteArray>`.
+
+## Decision
+
+Transitioning from a fixed size byte array to byte slice's would be the first step. This will enable usage of [cosmos-proto](https://github.com/regen-network/cosmos-proto) interface type. This removes boiler plate needed for oneof types.
```
melekes: > Transitioning from a fixed size byte array to byte slice's would be the first step.

Can you explain what you mean by this?

NOTE: with bytes array, there's no difference between amino and protobuf (only prefix bytes are different).

amino: (prefix bytes) + varint length + bytes
proto3: (field number + wire_type) + varint length + bytes

https://developers.google.com/protocol-buffers/docs/encoding#strings

docs/architecture/adr-054-crypto-encoding-2.md
melekes: I'd also like for us to think about the storage format (currently, amino-encoded objects).

marbar3778: this will change the proto encoding. that is why a hardfork is necessary to integrate this into chains, unless you have other ideas on how to handle state encoding?

docs/architecture/adr-054-crypto-encoding-2.md
marbar3778: this is just going from [keySize]byte to []byte.

marbar3778: put a small explanation

Issue #4492, opened 2020-02-27
WebSocket subscriptions change ID to latest RPC request on the same connection


michaelfig:
**Tendermint version** (use `tendermint version` or `git rev-parse --verify HEAD` if installed from source): v0.33.1-dev2

**ABCI app** (name for built-in, URL for self-written if it's publicly available):

An Node.js FFI-based integration to cosmos-sdk.

https://github.com/Agoric/agoric-sdk/tree/master/packages/cosmic-swingset

**Environment**:
- **OS** (e.g. from /etc/os-release): MacOS
- **Install tools**:
- **Others**: js-tendermint with patches for tendermint v0.33.x

**What happened**:

When my lite client sends RPCs on the same connection that has a current subscription, the subscription's "id" field is changed to the latest RPC request's.

**What you expected to happen**:

Subscription "id" field remains constant for a given subscription.

**Have you tried the latest version**: yes/no

yes

**How to reproduce it** (as minimally and precisely as possible):

1. Open an RPC websocket
2. Subscribe to events, like `tm.event = \'NewBlockHeader\'`
3. Note the returned '"id"' field for a number of blocks
4. Send an RPC request on the same connection, with a different ID
5. Note that the returned '"id"' field for the subscription has changed to the one in 4

**Logs (paste a small part showing an error (< 10 lines) or link a pastebin, gist, etc. containing more of the log file)**:

The following is the WebSocket traffic that exhibits the problem

https://gist.github.com/michaelfig/47909aa73407193e7487b5b3aa1485f8

**Config (you can paste only the changes you've made)**:

**node command runtime flags**: `--pruning=nothing`

**`/dump_consensus_state` output for consensus bugs**

**Anything else we need to know**:

PR #4493, opened 2020-02-27
Keep the original subscription "id" field when new RPCs come in

michaelfig:
Closes: #4492

## Description

Capture the subscription ID that is set when the subscription begins, so that the emitted events use that same ID.

______

For contributor use:

- [ ] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -23,6 +23,8 @@ program](https://hackerone.com/tendermint).
 
 ### BUG FIXES:
 
+- [rpc] [\#4493](https://github.com/tendermint/tendermint/pull/4493) Keep the original subscription "id" field when new RPCs come in (@michaelfig)
+
 - [rpc] [\#4437](https://github.com/tendermint/tendermint/pull/4437) Fix tx_search pagination with ordered results (@erikgrinaker)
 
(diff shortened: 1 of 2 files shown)


michaelfig:
@jackzampolin can you help shepherd this through?

erikgrinaker:
Thanks! Can you add a regression test for this please?

michaelfig:
@melekes, @erikgrinaker 
> I think this is a correct fix, but better to write a test for it.

I would love to, but am having a hard time.  The Go client doesn't use the subscription id at all (it matches based on the query instead of id).  It doesn't reflect the ID field to the caller of Subscribe, so I don't know how to test it without punching some serious holes in the abstractions.

Technically, the subscriber is not supposed to rely on the JSONRPC id, so even though this fix works for me, it might be better to fix the clients for the eventuality when the publisher doesn't send ids at all.

erikgrinaker:
Yeah, it looks like we don't have a whole lot of existing tests and fixtures around this either. Let's drop it for now and aim to improve overall testing of that module later. Could you add a changelog entry and rebase/update with `master`?

michaelfig:
> Could you add a changelog entry and rebase/update with `master`?

@erikgrinaker done!  Not sure if `circleci/proto-breakage` is a flake.

erikgrinaker:
Thanks @michaelfig, much appreciated! `proto-breakage` is faulty, see #4506.

PR #4496, opened 2020-02-28
docs: adr-046 add bisection algorithm details

### LINKED ISSUE #4329: lite2: align with the newer version of the spec

melekes:
Closes #4329

## Description

Update ADR-046 with reasons for choosing non-recursive version of the bisection algorithm.

[RENDERED 📜 ](https://github.com/tendermint/tendermint/blob/b0f5283be3d936ca6e699a1fe6a1a593a2984e57/docs/architecture/adr-046-light-client-implementation.md#bisection-algorithm-details)
______

For contributor use:

- [ ] ~~Wrote tests~~
- [ ] ~~Updated CHANGELOG_PENDING.md~~


cmwaters:
As a side note:

> If the light client does not have a header for height `X` and `TrustedHeader(X)` or
> `TrustedValidatorSet(X)` methods are called, it will download the header from
> primary provider and perform a backwards verification.

This is (to my knowledge) not currently true

docs/architecture/adr-046-light-client-implementation.md
```diff
@@ -127,6 +128,21 @@ cases of adjacent and non-adjacent headers. In the former case, it compares the
 hashes directly (2/3+ signed transition). Otherwise, it verifies 1/3+
 (`trustLevel`) of trusted validators are still present in new validators.
 
+### Bisection algorithm details
+
+Non-recursive bisection algorithm was implemented despite the spec containing
+the recursive version. There are two major reasons:
+
+1) Constant memory consumption => no risk of getting OOM (Out-Of-Memory) exceptions;
+2) Faster finality (see Fig. 1).
```
ancazamfir: while this is true for the example you show in Fig.1, not sure this is always the case. For example, in another case you can have something like this:

- start with th=1, ih=256
- the +1/3rd check fails a few times and interim headers are retrieved and checked, let's say: [1], 8, 16, 32, 64, 128, [256]
- 8 verifies against 1 so now we have th=8, ih=256
- again the +1/3rd check fails and other headers are retrieved: [8], 16, 24, 40, 72, 136, [256]
- 16 verifies against 8 and at this point th=16, ih=256
-...

The downloaded headers are not kept around (so we save on memory) but more headers and vals need to be retrieved.
Was the OOM something you saw? If we do the math for th and ih being 3wk apart (trusting period) and 1block/sec we probably have a ~20 depth in recursion with ~20blocks retrieved.

melekes: > The downloaded headers are not kept around (so we save on memory) but more headers and vals need to be retrieved.

Hmm.. You're right! Looks like we want to save those interim headers.

Thanks!

> Was the OOM something you saw? 

no, theoretical concern

cmwaters: I am not too sure if caching the interim headers will work, because the range changes so therefore the headers to be requested by the provider will be all different.

melekes: Can you give an example?

Won't we continue with same headers [8], 16, 24, 40, 72, 136, [256]?

cmwaters: No I thought we would find the halfway point between 8 and 256

ancazamfir: it also defeats a bit the purpose of the current non-recursive version. You can still do a non-recursive version by simulating the recursive stack and just store the headers (instead of full stack frames) and keep the same range.  
But maybe before you make more changes, it would be interesting to get some header/vals data from mainnet, and run both the version you have and the old recursive one to see differences in memory usage and performance. Or just use generated data.

cmwaters: yes I would be interested in checking that out

melekes: created https://github.com/tendermint/tendermint/issues/4504

ancazamfir: Also, it may be likely that a +1/3rd power change is a rare event in the 3-wk sliding window worth of blocks and the current non-recursive version may turn out to be the best. But again, this is something worth investigating.

PR #4500, opened 2020-02-28
test: simplify txsearch cancellation test

erikgrinaker:
Fixes feedback in https://github.com/tendermint/tendermint/pull/4418#discussion_r385629506. But mostly tests Mergify.


erikgrinaker:
@Mergifyio refresh

PR #4506, opened 2020-02-28
 fix: proto-breakage

marbar3778:
- this is amed to fix proto breakage for consumers

Closes: #XXX

## Description

______

For contributor use:

- [ ] Wrote tests


marbar3778:
mhhh i need to clone the repo while in ci and check against that. 

do we want to do that or just avoid it? @erikgrinaker @melekes

erikgrinaker:
Isn't the repo already cloned in CI? Not really sure what this does, but we should be able to use that somehow, right?

marbar3778:
now it should be fixed

PR #4508, opened 2020-03-01
lite: add P2P provider and corresponding P2P reactor

### LINKED ISSUE #4456: lite2: p2p transport
For [state sync](https://github.com/tendermint/tendermint/issues/828), it would be nice if the light client could communicate with P2P peers instead of RPC servers, so that users won't have to supply RPC endpoints in addition to P2P seeds when bootstrapping a node via state sync.

erikgrinaker:
Fixes #4456.

@melekes @brapse Could you have a quick look and see if you agree with the overall approach? If so, I'll tighten it up a bit and add some tests.

______

For contributor use:

- [ ] Wrote tests


lite2/provider/p2p/p2p.go
```diff
@@ -0,0 +1,60 @@
...
+// p2p provider uses P2P to obtain necessary light client information.
+type p2p struct {
+	dispatcher *reactor.Dispatcher
+	chainID    string
+	peer       p2pm.Peer
+}
+
+// New creates a new P2P provider. It takes a dispatcher from Reactor.Dispatcher() and an
+// appropriate peer to communicate with.
+func New(dispatcher *reactor.Dispatcher, chainID string, peer p2pm.Peer) (provider.Provider, error) {
```
melekes: no need to return an error if it's not possible for the func to return an error

lite2/reactor/codec.go
```diff
@@ -0,0 +1,13 @@
+package reactor
+
+import (
+	amino "github.com/tendermint/go-amino"
+	"github.com/tendermint/tendermint/types"
+)
+
+var cdc = amino.NewCodec()
```
melekes: wonder if it makes sense to switch to proto right away here (cc @marbar3778 )

lite2/reactor/dispatcher.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// respond provides a call response.
+func (d *Dispatcher) respond(src p2p.Peer, msg Message) {
+	d.Lock()
+	defer d.Unlock()
+	callID := msg.GetCallID()
+	call, ok := d.calls[callID]
+	if !ok {
+		d.logger.Error("Received call response for unknown call %q", callID)
```
melekes:
```suggestion
		d.logger.Error("Received call response for unknown call", "callID", callID)
```

lite2/reactor/dispatcher.go
```diff
@@ -0,0 +1,122 @@
...
+		return
+	}
+	if call.peerID != src.ID() {
+		d.logger.Error("Received call response from wrong peer %q, expected %q",
```
melekes:
```suggestion
		d.logger.Error("Received call response from wrong peer",
```

lite2/reactor/dispatcher.go
```diff
@@ -0,0 +1,122 @@
...
+			src.ID(), call.peerID)
```
melekes:
```suggestion
			"got", src.ID(), "expected", call.peerID)
```

lite2/reactor/dispatcher.go
```diff
@@ -0,0 +1,122 @@
...
+		return
+	}
+	call.ch <- msg
+	close(call.ch)
```
melekes: do we even need to close it?

lite2/reactor/codec.go
marbar3778: emhh, not right now as there are missing proto types for what is needed here. To get the types you would have to change quite a bit.

brapse:
Interesting way of doing things. From what I understand it looks like we’ve created a reactor for the light client to fetch commits and validator sets in a request/response pattern. Although I’m certain this works and I recall the motivation for using p2p here instead of RPC I’m concerned that this is a false economy. 

On one side we eliminate the need for a configuration parameter and on the other we commit to maintaining a straight forward and yet whole new reactor which provides an alternative way of reading data already accessible. I suspect that while clearly possible this might not be the best tradeoff in the long run as we are increasing the amount of code we need to maintain. Are there other benefits that I'm missing?

I’m not sure we should be doing this but would like to hear from @xla, @ebuchman

erikgrinaker:
>On one side we eliminate the need for a configuration parameter

It's not so much about saving a configuration parameter, and more about reducing the operational complexity of setting up Tendermint, particularly for first-time users.

If you just want to set up a full node to play around with Tendermint, I'd imagine it's already challenging enough to figure out where to get a set of seed nodes or persistent peers to connect to. Once you've finally obtained a few, you now discover that only a small minority of these actually expose RPC, and so you have to go off and find a different set of full nodes that serve RPC.

I'm not heavily invested in this though, so I'm fine with scrapping it if it's considered unnecessary. It might be a good idea to ask a couple of actual node operators though.

erikgrinaker:
Rejected in today's Tendermint call, we'll use RPC nodes for the first iteration.

PR #4521, opened 2020-03-03
rpc: create buffered subscriptions on /subscribe

### LINKED ISSUE #3935: RPC subscriptions: client is not pulling messages fast enough
**Tendermint version**
0.32.3

**ABCI app**
- kvstore
- InProc implementation

**Environment**:
- **OS** linux-5.2.10 (Archlinux)
- **Install tools**: git

**What happened**:
When I subscribe to events via RPC API, after getting 0 or more events (this is completely random) I get the following error:
```json
{
  "jsonrpc": "2.0",
  "id": "1#event",
  "error": {
    "code": -32000,
    "message": "Server error",
    "data": "subscription was cancelled (reason: client is not pulling messages fast enough)"
  }
}
```

I should keep getting the events

- step 1
Start tendermint:
`go run ./cmd/tendermint/main.go node --proxy_app kvstore`

- step 2
Subscribe to events, I'm using [ws](https://github.com/yhat/ws) but any WS client should be sufficient:
```bash
$  ws ws://localhost:26657/websocket
> {"jsonrpc":"2.0", "id": 1, "method": "subscribe", "params": {"query": "tm.event='Tx'"}}
< {
  "jsonrpc": "2.0",
  "id": 1,
  "result": {}
}

- step 3
Broadcast tons of Tx:
```bash
WAIT=0.1
for i in $(seq 10000)
do
  echo $i
  sleep $WAIT
  curl http://localhost:26657/broadcast_tx_sync\?tx\=\"$i\"
done
```

You can increment `$WAIT` to 1 or more seconds until the error dissapears.

**Config**:
Generated by `tendermint init`

melekes:
buffer size: 100

Closes #3935

______

For contributor use:

- [ ] ~~Wrote tests~~ tested manually


rpc/core/events.go
```diff
@@ -12,6 +12,11 @@ import (
 	rpctypes "github.com/tendermint/tendermint/rpc/lib/types"
 )
 
+const (
+	// Buffer on the Tendermint (server) side to allow some slowness in clients.
+	subCap = 100
```
erikgrinaker: This name is a bit unclear. Maybe `subBufferSize`?

CHANGELOG_PENDING.md
```diff
@@ -24,6 +24,7 @@ Friendly reminder, we have a [bug bounty program](https://hackerone.com/tendermi
 
 ### BUG FIXES:
 
+- [rpc] \#3935 Create buffered subscriptions on `/subscribe` (@melekes)
```
tessr: nit, but do you want to make this a link like the other items in this list?

marbar3778: technically this is the correct way according to the contributing.md and then there is a script to add all the links prior to release. People have been putting links recently, I guess we should decide on one way or the other?

melekes: yeah, there's a script for that

Issue #4420, opened 2020-02-18
lite2: question: should we replace primary if it gives us invalid headers


melekes:
for example malformed header, a header that is not signed with +2/3 of voting power, or in the case there are two consecutive headers h1 and h2 that are not consistent (for example h2.V != h1.NextV, i.e., the current voting set of h2 is not equal to the next voting set field of h1)

NOTE: we do replace primary if it's unavailable for a long period of time.

melekes:
@cmwaters what do you think? I remember we've talked about this, but forgot why we have decided to leave the burden of removing primary (by resetting the light client) to the caller.

cmwaters:
I think we decided to replace the primary. However we don't give the lite client the opportunity to add witnesses. Once all witnesses are exhausted, the lite clients returns an error and closes down and it is up to the caller to reset the lite client with a new (or potentially the same) list of providers

cmwaters:
Are we still looking to implement this?

melekes:
I think so

PR #4523, opened 2020-03-03
lite2: replace primary when providing invalid header

cmwaters:
Closes: #4420

## Description

Created a new error `ErrInvalidHeader`which can be formed during the verification process `verifier.go` and will result in the replacement of the primary provider with a witness by executing: `replacePrimaryProvider()`

______

For contributor use:

- [ ] Wrote tests


lite2/errors.go
```diff
@@ -28,3 +28,11 @@ type ErrNewValSetCantBeTrusted struct {
 func (e ErrNewValSetCantBeTrusted) Error() string {
 	return fmt.Sprintf("cant trust new val set: %v", e.Reason)
 }
+
+type ErrInvalidHeader struct {
+	Reason string
```
melekes:
```suggestion
	Reason error
```

lite2/client.go
```diff
@@ -711,7 +711,19 @@ func (c *Client) sequence(
 		err = VerifyAdjacent(c.chainID, trustedHeader, interimHeader, interimVals,
 			c.trustingPeriod, now)
 		if err != nil {
-			return errors.Wrapf(err, "failed to verify the header #%d", height)
+			switch err.(type) {
+			case ErrInvalidHeader:
+				c.logger.Debug("replacing primary because: ", "err", err.Error())
```
```suggestion
				c.logger.Error("primary sent us invalid header", "err", err, "primary", primary)
```

lite2/client.go
```diff
@@ -711,7 +711,19 @@ func (c *Client) sequence(
...
+				err = c.replacePrimaryProvider()
```
```suggestion
				err1 := c.replacePrimaryProvider()
```

lite2/client.go
```diff
@@ -711,7 +711,19 @@ func (c *Client) sequence(
...
+				if err != nil {
+					return errors.Wrapf(err, "tried to verify adjacent header")
```
```suggestion
					err = errors.Wrapf(err, "failed to verify the header #%d", height)
					return errors.Wrapf(err1, "can't replace primary (original error: %v)", err)
```

lite2/client.go
```diff
@@ -830,19 +851,36 @@ func (c *Client) backwards(
 		}
 
 		if err := interimHeader.ValidateBasic(c.chainID); err != nil {
```
melekes: the following 3 if conditions need to be extracted into a function. then we won't have to repeat replacePrimaryProvider 3 times

cmwaters: I'm going to put it in `verifier.go` unless you disagree

lite2/client.go
```diff
@@ -711,7 +711,20 @@ func (c *Client) sequence(
 		err = VerifyAdjacent(c.chainID, trustedHeader, interimHeader, interimVals,
 			c.trustingPeriod, now)
 		if err != nil {
-			return errors.Wrapf(err, "failed to verify the header #%d", height)
+			switch err.(type) {
+			case ErrInvalidHeader:
+				c.logger.Error("replacing primary because: ", "err", err.Error())
```
```suggestion
				c.logger.Error("primary sent us invalid header -> replacing", "err", err)
```

lite2/client.go
```diff
@@ -764,6 +777,16 @@ func (c *Client) bisection(
 				return err
 			}
 
+		case ErrInvalidHeader:
+			c.logger.Error("replacing primary because: ", "err", err.Error())
```
```suggestion
			c.logger.Error("primary sent us invalid header -> replacing", "err", err)
```

lite2/verifier.go
```diff
@@ -118,7 +118,27 @@ func VerifyAdjacent(
...
 	if err := untrustedVals.VerifyCommit(chainID, untrustedHeader.Commit.BlockID, untrustedHeader.Height,
 		untrustedHeader.Commit); err != nil {
-		return err
+		return ErrInvalidHeader{err}
+	}
+
+	return nil
+}
+
+func VerifyBackwards(chainID string, untrustedHeader, trustedHeader *types.SignedHeader) error {
```
melekes: godoc please

melekes:
Should we write tests for this?

cmwaters:
> Should we write tests for this?

All the verification tests are already covered - the only thing we would need to check is that the providers are replaced and that the verification process resumes as normal

cmwaters:
I wrote some more helper functions in the benchmarking PR that we could use if we want to construct more elaborate providers

melekes:
> I wrote some more helper functions in the benchmarking PR that we could use if we want to construct more elaborate providers

works for me

Issue #4396, opened 2020-02-13
lite2: forbid to have both auto-update and concurrent calls to VerifyX


melekes:
this will lead to unexpected races. we should probably return an error in VerifyX funcs if auto update is running (i.e. update_period > 0 && Start was called).

The alternative is to make VerifyX functions concurrently safe.

cmwaters:
Shouldn't a manual operation such as VerifyX cause an automatic operation like auto update to block and not the other way around?

melekes:
> Shouldn't a manual operation such as VerifyX cause an automatic operation like auto update to block

"The alternative is to make VerifyX functions concurrently safe." as I said ^, this is an option, yes.

melekes:
I'm just hesitant to add yet another mutex.

cmwaters:
I'm going to add a mutex, unless you have a better suggestion for resolving this

cmwaters:
It should be safe to run sequence/bisection (i.e. Autoupdate) alongside backwards verification (verifying a header that is before the latestTrustedHeader) right? In that case I should just put the mutex on the forward verification functions

melekes:
> It should be safe to run sequence/bisection (i.e. Autoupdate) alongside backwards verification (verifying a header that is before the latestTrustedHeader) right?

Correct

melekes:
Removed auto-update here: https://github.com/tendermint/tendermint/pull/4535

As for Update / VerifyX funcs,

Let's delay making Update / VerifyX funcs concurrency safe until there's a real need. Sorry about that 🙏

melekes:
https://github.com/tendermint/tendermint/pull/4518 left the branch as we might reconsider in the future.

PR #4527, opened 2020-03-04
lite: simple concurrent verification

cmwaters:
Closes: #4396

## Description

Instead of using locks, this PR implements a simple bool value which allows for the running of one `autoUpdateRoutine` and the blocking of any manual forward verification.

NOTE: that backwards verification still operates irrelevant of autoUpdate

NOTE: There is a second PR that uses locks to allow for concurrent verification

______

For contributor use:

- [ ] Wrote tests

### DIFF
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -128,6 +128,8 @@ type Client struct {
 
 	// See UpdatePeriod option
 	updatePeriod time.Duration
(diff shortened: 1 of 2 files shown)


melekes:
I am going to add a mutex. Can't find these changes here.

Issue #4530, opened 2020-03-05
Handle evidence from light client(s)


melekes:
https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-047-handling-evidence-from-light-client.md

tessr:
This is an issue for tracking the implementation of evidence handling from light clients, yeah?

melekes:
> This is an issue for tracking the implementation of evidence handling from light clients, yeah?

correct

Issue #4214, opened 2019-12-03
Formalise definition, detection and processing of global evidence


milosevic:
#4189 introduces two kind of evidences: individual and global. We need to formalise the notion of global evidence and explain how it is being detected and created from a fork (detected by the lite client). Furthermore, the specification of required data structures and global evidence processing on chain should be done.

melekes:
https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-056-proving-amnesia-attacks.md

Issue #4189, opened 2019-11-25
On chain execution of fork accountability protocol 


milosevic:
As described in https://github.com/tendermint/spec/blob/master/spec/consensus/fork-accountability.md there are several kind of attacks that can lead to a fork in case Tendermint failure model (https://github.com/tendermint/spec/blob/master/spec/consensus/light-client.md#tendermint-failure-model) assumptions do not hold. 

With respect to the way evidence misbehaviour is constructed and processed we make difference between: 
- individual evidence (equivocation, phantom and lunatic validators attack(https://github.com/tendermint/spec/blob/master/spec/consensus/fork-accountability.md#off-chain-attacks)) and 
- global evidence (flip flopping attacks (https://github.com/tendermint/spec/blob/master/spec/consensus/fork-accountability.md#flip-flopping)).  

In the former case, we can construct evidence of misbehaviour for an individual process and evidence is self-verifiable, i.e., it does not require access to the state of all (correct) validators to check its validity. In the latter case, detecting faulty validators requires access to the state of all (correct) validators and running fork accountability procedure (https://github.com/tendermint/spec/pull/55). 

In case of fork attack there is at some point in time +1/3 of voting power controlled by the faulty validators, therefore there is a risk that faulty validators will censor propagation and execution on the chain of the corresponding evidence (detected by some correct party). Therefore, one line of thinking would suggest halting the chain in case of fork detection and execution of the fork accountability procedure off-chain by relying on social consensus. On the other side, as validator set in Tendermint is dynamic, faulty validators might not be able to guarantee they would stay in power long enough to censor fork accountability protocol. For example, faulty validators at the moment of detection of attack do not control anymore 1/3+ of voting power of the current validator set as in the meantime there have been some movement of voting power due to delegation/redelegation or additional voting power being staked. Therefore, execution of the fork accountability protocol on the chain still seems like interesting system feature even in case of network forks. 

[...]

PR #4532, opened 2020-03-05
evidence: handling evidence from light client(s)

melekes:
Closes: https://github.com/tendermint/tendermint/issues/4530

## Description

This PR contains logic for both submitting an evidence by the light client (lite2 package) and receiving it on the Tendermint side (`/broadcast_evidence` RPC and/or `EvidenceReactor#Receive`). Upon receiving the `ConflictingHeadersEvidence` (introduced by this PR), the Tendermint validates it, then breaks it down into smaller pieces (`DuplicateVoteEvidence, LunaticValidatorEvidence, PhantomValidatorEvidence, PotentialAmnesiaEvidence`). Afterwards, each piece of evidence is verified against the state of the full node and added to the pool, from which it's reaped upon block creation.

______

For contributor use:

- [x] Wrote tests


lite2/client_test.go
```diff
@@ -940,3 +940,43 @@ func TestClientTrustedValidatorSet(t *testing.T) {
...
+			1: h1,
+			2: altH2,
+		},
+		map[int64]*types.ValidatorSet{
+			1: vals,
+			2: vals,
+		},
+	)
+
+	c, err := NewClient(
```
golangcibot: SA4006: this value of `err` is never used (from `staticcheck`)

rpc/client/httpclient.go
```diff
@@ -288,9 +288,13 @@ func (c *baseRPCClient) ConsensusState() (*ctypes.ResultConsensusState, error) {
 
 func (c *baseRPCClient) ConsensusParams(height *int64) (*ctypes.ResultConsensusParams, error) {
 	result := new(ctypes.ResultConsensusParams)
-	_, err := c.caller.Call("consensus_params", map[string]interface{}{"height": height}, result)
+	params := make(map[string]interface{}, 0)
```
golangcibot: S1019: should use make(map[string]interface{}) instead (from `gosimple`)

rpc/client/httpclient.go
```diff
@@ -319,34 +323,46 @@ func (c *baseRPCClient) Genesis() (*ctypes.ResultGenesis, error) {
...
-		return nil, errors.Wrap(err, "Genesis")
+		return nil, err
 	}
 	return result, nil
 }
 
 func (c *baseRPCClient) Block(height *int64) (*ctypes.ResultBlock, error) {
 	result := new(ctypes.ResultBlock)
-	_, err := c.caller.Call("block", map[string]interface{}{"height": height}, result)
+	params := make(map[string]interface{}, 0)
```
golangcibot: S1019: should use make(map[string]interface{}) instead (from `gosimple`)

rpc/client/httpclient.go
```diff
@@ -319,34 +323,46 @@ func (c *baseRPCClient) Genesis() (*ctypes.ResultGenesis, error) {
...
-		return nil, errors.Wrap(err, "Block")
+		return nil, err
 	}
 	return result, nil
 }
 
 func (c *baseRPCClient) BlockResults(height *int64) (*ctypes.ResultBlockResults, error) {
 	result := new(ctypes.ResultBlockResults)
-	_, err := c.caller.Call("block_results", map[string]interface{}{"height": height}, result)
+	params := make(map[string]interface{}, 0)
```
golangcibot: S1019: should use make(map[string]interface{}) instead (from `gosimple`)

melekes:
I am going to transform abci.Evidence from

```
message Evidence {
  string                    type      = 1;
  Validator                 validator = 2 [(gogoproto.nullable) = false];
  int64                     height    = 3;
  google.protobuf.Timestamp time = 4 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
  int64                     total_voting_power = 5;
}
```

to 

```
message Evidence {
  string                    type      = 1;
  repeated Validator                 validators = 2 [(gogoproto.nullable) = false];
  int64                     height    = 3;
  google.protobuf.Timestamp time = 4 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
  int64                     total_voting_power = 5;
}

```

making validators an array

melekes:
evidence handling update:

[x] RPC accepts ConflictingHeadersEvidence
[x] ConflictingHeadersEvidence gossiped to other nodes
[ ] Before adding it to a block, ConflictingHeadersEvidence is broken into parts (double vote, phantom, lunatic, potential flip flopping)
    [ ] if fork accountability needs to be run (~ we can't immediately understand who misbehaved), PotentialFlipFloppingEvidence (https://github.com/tendermint/tendermint/issues/4214) is created
    [ ] When PotentialFlipFloppingEvidence is committed, fork accountability reactor is started. As a result, AmnesiaEvidence is formed (https://github.com/tendermint/tendermint/issues/4189)

melekes:
TODO:

- [ ] for PhantomValidatorEvidence, check that validator is not part of the full node's validator set
- [ ] for LunaticValidatorEvidence, check that InvalidHeaderField indeed differs from what full node has

types/evidence.go
```diff
@@ -342,3 +355,581 @@ func (evl EvidenceList) Has(evidence Evidence) bool {
...
+	case !bytes.Equal(committedHeader.ValidatorsHash, alternativeHeader.ValidatorsHash):
+		invalidField = "ValidatorsHash"
+	case !bytes.Equal(committedHeader.NextValidatorsHash, alternativeHeader.NextValidatorsHash):
+		invalidField = "NextValidatorsHash"
+	case !bytes.Equal(committedHeader.ConsensusHash, alternativeHeader.ConsensusHash):
+		invalidField = "ConsensusHash"
+	case !bytes.Equal(committedHeader.AppHash, alternativeHeader.AppHash):
+		invalidField = "AppHash"
+	case !bytes.Equal(committedHeader.LastResultsHash, alternativeHeader.LastResultsHash):
+		invalidField = "LastResultsHash"
```
cmwaters, line 422: Is it possible to have multiple invalid fields - is it okay if we only submit one

melekes: It is possible

melekes: thought: we can make it an array

melekes:
```
=== RUN   TestGenesisAndValidators
    TestGenesisAndValidators: rpc_test.go:176: 
        	Error Trace:	rpc_test.go:176
        	Error:      	Not equal: 
        	            	expected: 10
        	            	actual  : 9
        	Test:       	TestGenesisAndValidators
    TestGenesisAndValidators: rpc_test.go:176: 
        	Error Trace:	rpc_test.go:176
        	Error:      	Not equal: 
        	            	expected: 10
        	            	actual  : 9
        	Test:       	TestGenesisAndValidators
```

is failing on CircleCI

types/evidence.go
```diff
@@ -342,3 +355,581 @@ func (evl EvidenceList) Has(evidence Evidence) bool {
...
+	return fmt.Sprintf("PhantomValidatorEvidence{%X voted for %d/%X}",
+		e.Vote.ValidatorAddress, e.Header.Height, e.Header.Hash())
+}
+
+//-------------------------------------------
+
+type LunaticValidatorEvidence struct {
+	Header             *Header `json:"header"`
+	Vote               *Vote   `json:"vote"`
+	InvalidHeaderField string  `json:"invalid_header_field"`
```
cmwaters, line 711: Do you want to add this field to the evidence handling adr

state/validation.go
```diff
@@ -176,23 +176,82 @@ func VerifyEvidence(stateDB dbm.DB, state State, evidence types.Evidence) error
...
-	ev := evidence
-	height, addr := ev.Height(), ev.Address()
-	_, val := valset.GetByAddress(addr)
-	if val == nil {
-		return fmt.Errorf("address %X was not a validator at height %d", addr, height)
+	addr := evidence.Address()
+	var val *types.Validator
+
+	if _, ok := evidence.(*types.PhantomValidatorEvidence); ok {
+		_, val = valset.GetByAddress(addr)
```
cmwaters, line 217: Do you want to add some comments for this. I'm guessing you are looping through all the validator sets in all the blocks committed within the MaxAgeNumBlocks in search of the validator to punish right?

state/validation.go
```diff
@@ -176,23 +176,82 @@ func VerifyEvidence(stateDB dbm.DB, state State, evidence types.Evidence) error
...
+		if val != nil {
+			return fmt.Errorf("address %X is a validator at height %d", addr, evidence.Height())
+		}
+
+		// TODO: something smarter here
```
cmwaters, line 222: Is there perhaps a list of validators that currently have something staked that you could loop through instead?

evidence/pool.go
```diff
@@ -96,34 +98,54 @@ func (evpool *Pool) Update(block *types.Block, state sm.State) {
...
 // AddEvidence checks the evidence is valid and adds it to the pool.
 func (evpool *Pool) AddEvidence(evidence types.Evidence) error {
-
-	// check if evidence is already stored
-	if evpool.store.Has(evidence) {
-		return ErrEvidenceAlreadyStored{}
+	state := evpool.State()
+	valSet, err := sm.LoadValidators(evpool.stateDB, evidence.Height())
+	if err != nil {
+		return errors.Wrapf(err, "can't load validators at height #%d", evidence.Height())
```
marbar3778, line 104:
```suggestion
		return fmt.Errorf("can't load validators at height #%d, %w", evidence.Height(), err)
```

to keep inline with the styleguide

evidence/pool.go
```diff
@@ -96,34 +98,54 @@ func (evpool *Pool) Update(block *types.Block, state sm.State) {
...
 	}
 
-	if err := sm.VerifyEvidence(evpool.stateDB, evpool.State(), evidence); err != nil {
-		return ErrInvalidEvidence{err}
+	evList := []types.Evidence{evidence}
+	blockMeta := evpool.blockStore.LoadBlockMeta(evidence.Height())
+	if blockMeta == nil {
+		return errors.Wrapf(err, "don't have block at height #%d", evidence.Height())
```
marbar3778, line 110:
```suggestion
		return fmt.Errorf("don't have block at height #%d, %w", evidence.Height(), err)
```
ditto

types/evidence.go
cmwaters: Actually I think one is sufficient

types/evidence.go
```diff
@@ -342,3 +355,581 @@ func (evl EvidenceList) Has(evidence Evidence) bool {
...
+	}
+
+	return nil
+}
+
+func (ev ConflictingHeadersEvidence) Equal(ev2 Evidence) bool {
+	switch ev2.(type) {
+	case ConflictingHeadersEvidence:
+		e2 := ev2.(ConflictingHeadersEvidence)
+		return bytes.Equal(ev.H1.Hash(), e2.H1.Hash())
```
cmwaters, line 573: Do you want to compare both header hashes or do you think one is sufficient? If a chain was forked into three, then you could have different conflicting headers being equal - not sure if this is a problem or not

types/evidence.go
```diff
@@ -342,3 +355,581 @@ func (evl EvidenceList) Has(evidence Evidence) bool {
...
+		return errors.New("empty header")
+	}
+
+	if e.Vote == nil {
+		return errors.New("empty vote")
+	}
+
+	// if err := e.Header.ValidateBasic(); err != nil {
+	// 	return fmt.Errorf("invalid header: %v", err)
+	// }
```
cmwaters, line 681: don't you want to validate the header as well -> if not then remove the comments

types/evidence.go
```diff
@@ -342,3 +355,581 @@ func (evl EvidenceList) Has(evidence Evidence) bool {
...
```
cmwaters, line 782: same here

evidence/pool.go
```diff
@@ -96,34 +98,54 @@ func (evpool *Pool) Update(block *types.Block, state sm.State) {
 
 // AddEvidence checks the evidence is valid and adds it to the pool.
 func (evpool *Pool) AddEvidence(evidence types.Evidence) error {
-
-	// check if evidence is already stored
-	if evpool.store.Has(evidence) {
-		return ErrEvidenceAlreadyStored{}
+	state := evpool.State()
```
cmwaters, line 101: I'm not too sure where exactly but we should do a quick `ValidateBasic()` of the evidence before we get into the heavier processing. Either here or perhaps when we loop through the evList right after checking whether the evidence is already in the database

types/evidence.go
```diff
@@ -233,10 +246,10 @@ func (dve *DuplicateVoteEvidence) ValidateBasic() error {
...
 	if err := dve.VoteA.ValidateBasic(); err != nil {
-		return fmt.Errorf("invalid VoteA: %v", err)
+		return fmt.Errorf("invalid VoteA: %w", err)
 	}
 	if err := dve.VoteB.ValidateBasic(); err != nil {
-		return fmt.Errorf("invalid VoteB: %v", err)
+		return fmt.Errorf("invalid VoteB: %w", err)
 	}
 	// Enforce Votes are lexicographically sorted on blockID
 	if strings.Compare(dve.VoteA.BlockID.Key(), dve.VoteB.BlockID.Key()) >= 0 {
```
cmwaters, line 255: Do we want to move all the mock evidence stuff to evidence_test.go as it's cluttering up the evidences file which is becoming more populated

types/evidence.go
melekes: I'll uncomment once https://github.com/tendermint/tendermint/pull/4638 is merged

evidence/pool.go
melekes: `ValidateBasic` is called in https://github.com/tendermint/tendermint/blob/4647810d8509686e2e3d38d18ac1c51a97d7de32/rpc/core/evidence.go#L15 when evidence is submitted through RPC and https://github.com/tendermint/tendermint/blob/4647810d8509686e2e3d38d18ac1c51a97d7de32/evidence/reactor.go#L73 when evidence is submitted through P2P.

types/evidence.go
melekes: No, you're right. We should compare both headers

evidence/pool.go
```diff
@@ -96,34 +97,66 @@ func (evpool *Pool) Update(block *types.Block, state sm.State) {
...
+	for _, ev := range evList {
+		// check if evidence is already stored
+		if evpool.store.Has(evidence) {
+			return ErrEvidenceAlreadyStored{}
+		}
+
+		// For lunatic validator evidence, a header needs to be fetched.
+		var header *types.Header
+		if _, ok := ev.(*types.LunaticValidatorEvidence); ok {
+			blockMeta := evpool.blockStore.LoadBlockMeta(ev.Height())
```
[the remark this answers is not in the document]
melekes: Not sure... Assuming goleveldb is holding some of it's state in memory, it should not be an issue. The correct approach is probably conduct a benchmark to confirm this.

state/validation.go
melekes: TM does not have a notion of stake. But we can track validators within `evidence.MaxAge`

rpc/core/evidence.go
```diff
@@ -10,6 +12,10 @@ import (
 // BroadcastEvidence broadcasts evidence of the misbehavior.
 // More: https://docs.tendermint.com/master/rpc/#/Info/broadcast_evidence
 func BroadcastEvidence(ctx *rpctypes.Context, ev types.Evidence) (*ctypes.ResultBroadcastEvidence, error) {
+	if err := ev.ValidateBasic(); err != nil {
+		return nil, errors.Wrap(err, "evidence.ValidateBasic failed")
```
melekes, line 16:
```suggestion
		return nil, fmt.Errorf("evidence.ValidateBasic failed: %w")
```

melekes:
After some consideration, I think `PhantomValidatorEvidence` should include `WasValidatorAtHeight` - last height at which validator was a part of validator set (note: it's possible that validator who's about to bond commits phantom validator attack at some previous height, so we should account for that too)

melekes:
After some consideration, I think PhantomValidatorEvidence should include WasValidatorAtHeight - last height at which validator was a part of validator set (note: it's possible that validator who's about to bond commits phantom validator attack at some previous height, so we should account for that too)

docs/architecture/adr-047-handling-evidence-from-light-client.md
```diff
@@ -26,6 +27,11 @@ type ConflictingHeadersEvidence struct {
 }
 ```
 
+_Remark_: Theoretically, only the header, which differs from what a full node
+has, needs to be sent. But sending two headers a) makes an evidence easily
+verifiable b) simplifies retransmission (since the full node does not have to
+query each receiver about which header it has: A or B)
```
melekes, line 33: let me know if you (the reader) disagree

state/validation.go
```diff
@@ -177,23 +177,48 @@ func VerifyEvidence(stateDB dbm.DB, state State, evidence types.Evidence) error
...
+	// For PhantomValidatorEvidence, check evidence.Address was not part of the
+	// validator set at height evidence.Height, but was a validator before OR
+	// after.
+	if phve, ok := evidence.(*types.PhantomValidatorEvidence); ok {
+		_, val = valset.GetByAddress(addr)
+		if val != nil {
+			return fmt.Errorf("address %X was a validator at height %d", addr, evidence.Height())
+		}
+
+		valset, err := LoadValidators(stateDB, phve.LastHeightValidatorWasInSet)
```
melekes, line 205: check if it's still bonded (`LastHeightValidatorWasInSet < Height - MaxAgeNumInBlocks`)

evidence/pool.go
```diff
@@ -171,6 +223,53 @@ func (evpool *Pool) removeEvidence(
...
+			}
+		}
+	}
+}
+
 func evMapKey(ev types.Evidence) string {
 	return string(ev.Hash())
 }
+
+func buildValToLastHeightMap(state sm.State, stateDB dbm.DB) valToLastHeightMap {
```
melekes, line 245: this needs to be tested

evidence/pool.go
```diff
@@ -171,6 +223,53 @@ func (evpool *Pool) removeEvidence(
 	}
 }
 
+func (evpool *Pool) cleanupValToLastHeight(blockHeight int64) {
```
melekes, line 226: same

docs/architecture/adr-047-handling-evidence-from-light-client.md
cmwaters: Is `ConflictingHeadersEvidence` sent between other full nodes though? My understanding was that this composite evidence was only sent from light clients to full nodes and then the evidence was immediately broken down and then sent to the other nodes. Regardless, two headers will need to be sent because the light client will send `ConflictingHeadersEvidence` to witnesses that it doesn't know which header the witness has.

evidence/pool.go
```diff
@@ -92,43 +105,81 @@ func (evpool *Pool) Update(block *types.Block, state sm.State) {
...
+		}
+
+		evList = ce.Split(&blockMeta.Header, valSet, evpool.valToLastHeight)
 	}
 
-	evpool.logger.Info("Verified new evidence of byzantine behaviour", "evidence", evidence)
+	for _, ev := range evList {
+		if evpool.store.Has(evidence) {
+			return ErrEvidenceAlreadyStored{}
+		}
```
cmwaters, line 145: AFAIK, from the moment of receiving evidence to storing at and even when we propose it, we never run a `ValidateBasic()` on the evidence. I would suggest running the check here just after we've checked whether we have the evidence or not.

types/evidence.go
```diff
@@ -245,6 +266,663 @@ func (dve *DuplicateVoteEvidence) ValidateBasic() error {
...
+						VoteA:  ev.H1.Commit.GetVote(i),
+						VoteB:  ev.H2.Commit.GetVote(j),
+					})
+				} else {
+					// if H1.Round != H2.Round we need to run full detection procedure => not
+					// immediately slashable.
+					evList = append(evList, &PotentialAmnesiaEvidence{
+						VoteA: ev.H1.Commit.GetVote(i),
+						VoteB: ev.H2.Commit.GetVote(j),
+					})
```
cmwaters, line 437: Are we doing Potential Amnesia as a batch of votes? In this case we may want to change this then.

melekes: Yes, but let's do it in a separate PR.

evidence/pool.go
melekes: https://github.com/tendermint/tendermint/pull/4532#discussion_r407422780

we validate evidence in all places we receive it.

melekes: but maybe I am missing something

evidence/pool.go
```diff
@@ -92,43 +105,81 @@ func (evpool *Pool) Update(block *types.Block, state sm.State) {
 
 	// remove evidence from pending and mark committed
 	evpool.MarkEvidenceAsCommitted(block.Height, block.Time, block.Evidence.Evidence)
+
+	evpool.cleanupValToLastHeight(block.Height)
```
cmwaters, line 109: `ValToLastHeight` also needs to be updated in the sense that the validators of the the new block should be added to this

evidence/pool.go
cmwaters, line 226: As we bound the expiration of evidence by both block height and time, this function should also remove the validator if the latest height is of a height which has expired (in relation to time).

cmwaters: If we implement this "is the validator still slashable" map properly then this can also become a proxy for checking for expiration and rather then have to check the params if the evidence is expired we just check whether the validators of the evidence are still punishable

cmwaters: Note that this highlights an incongruity in our thinking in the sense that it's possible that a validator who might be part of an evidence that is now expired could still be part of the validator set and therefore still be slashed.

evidence/pool.go
melekes: right! good catch

types/evidence.go
```diff
@@ -245,6 +266,663 @@ func (dve *DuplicateVoteEvidence) ValidateBasic() error {
...
+		alternativeHeader = ev.H2
+	} else {
+		alternativeHeader = ev.H1
+	}
+
+	// If there are signers(alternativeHeader) that are not part of
+	// validators(committedHeader), they misbehaved as they are signing protocol
+	// messages in heights they are not validators => immediately slashable
+	// (#F4).
+	for i, sig := range alternativeHeader.Commit.Signatures {
```
cmwaters, line 343: The problem I see with detecting phantom validators is if we are in a fork and we don't realise yet (I know this is a very rare case) then an honest validator could be fooled into thinking that it is part of the validator set for this alternativeHeader and feels that it is perfectly entitled to voting. Now when the network partition stops and honest nodes reconnect they will instantly start forming phantom validator evidence against one another and both forks will slash the honest validators of the other fork

cmwaters: In fact, unless a fork is detected at the root i.e. when the duplicate vote/amnesia attack happens it will almost certainly lead to a flurry of phantom validator evidence when it is discovered at a height that is greater than the root

evidence/pool.go
cmwaters: No you're right. I saw the `msg.ValidateBasic()` and thought that it validated the message itself somehow, not the individual evidences inside them

types/evidence.go
cmwaters: Okay, I've appended it to the refactor list for the modules package

types/evidence.go
melekes: > it will almost certainly lead to a flurry of phantom validator evidence

I'd have to think about it more, but first thought - no, because 1/3+ can not change validator set in incorrect way (without buy-in from either 1/3s)

state/validation.go
```diff
@@ -132,7 +132,7 @@ func validateBlock(evidencePool EvidencePool, stateDB dbm.DB, state State, block
 
 	// Validate all evidence.
 	for _, ev := range block.Evidence.Evidence {
-		if err := VerifyEvidence(stateDB, state, ev); err != nil {
+		if err := VerifyEvidence(stateDB, state, ev, &block.Header); err != nil {
```
cmwaters, line 135: If the evidence is of the type Lunatic Validator then we will need to retrieve the header from the nodes block store in order to validate it else this check will fail everytime.

cmwaters: As a side note, I think here we should also first check that have the evidence before trying to verify it again. That way we save time because we just need to check that it hasn't expired (or that the validator is still punishable). I can create an issue for it.

types/evidence.go
cmwaters: But the validator set can still change in a correct way. As in over time, with different transactions submitted, the validator set could slowly change.

docs/architecture/adr-047-handling-evidence-from-light-client.md
melekes: > Regardless, two headers will need to be sent because the light client will send ConflictingHeadersEvidence to witnesses that it doesn't know which header the witness has.

We could ask a witness, but not sure it's worth it. I'll remove the (b) sentence. Thanks!

types/evidence.go
melekes: let's maybe create a separate issue where we'll list all implications of full node following a fork, light client following a fork and what we can do about it.

state/validation.go
melekes: > If the evidence is of the type Lunatic Validator then we will need to retrieve the header from the nodes block store in order to validate it else this check will fail everytime.

we have the block here (we're validating it, so the header is always present).

> As a side note, I think here we should also first check that we have the evidence already stored before trying to verify it again

You're assuming that the node has all evidence, which might not be the case! But I agree, we should probably make this optimization! I'll create an issue.

melekes: https://github.com/tendermint/tendermint/issues/4696

cmwaters: > You're assuming that the node has all evidence, which might not be the case! But I agree, we should probably make this optimization! I'll create an issue.

Yeah I just meant it as an optimization thing

types/evidence.go
melekes: https://github.com/tendermint/tendermint/issues/4697

types/evidence.go
```diff
@@ -245,6 +266,663 @@ func (dve *DuplicateVoteEvidence) ValidateBasic() error {
...
+	}
+	return false
+}
+
+//-------------------------------------------
+
+// ConflictingHeadersEvidence is primarily used by the light client when it
+// observes two conflicting headers, both having 1/3+ of the voting power of
+// the currently trusted validator set.
+type ConflictingHeadersEvidence struct {
```
marbar3778, line 309: these types exceed the MaxEvidenceBytes variable. is this a problem?

melekes: no because we break it up into smaller pieces upon receiving

PR #4535, opened 2020-03-05
lite2: remove auto update

melekes:
## Description

We first introduced auto-update as a separate struct `AutoClient`, which
was wrapping `Client` and calling `Update` periodically. 

```go
// AutoClient can auto update itself by fetching headers every N seconds.
type AutoClient struct {
    base         *Client
    updatePeriod time.Duration
    quit         chan struct{}

    trustedHeaders chan *types.SignedHeader
    errs           chan error
}

// NewAutoClient creates a new client and starts a polling goroutine.
func NewAutoClient(base *Client, updatePeriod time.Duration) *AutoClient {
    c := &AutoClient{
        base:           base,
        updatePeriod:   updatePeriod,
        quit:           make(chan struct{}),
        trustedHeaders: make(chan *types.SignedHeader),
        errs:           make(chan error),
    }
    go c.autoUpdate()
    return c
}

// TrustedHeaders returns a channel onto which new trusted headers are posted.
func (c *AutoClient) TrustedHeaders() <-chan *types.SignedHeader {
    return c.trustedHeaders
}

// Err returns a channel onto which errors are posted.
func (c *AutoClient) Errs() <-chan error {
    return c.errs
}

// Stop stops the client.
func (c *AutoClient) Stop() {
    close(c.quit)
}

func (c *AutoClient) autoUpdate() {
    ticker := time.NewTicker(c.updatePeriod)
    defer ticker.Stop()

    for {
        select {
        case <-ticker.C:
            lastTrustedHeight, err := c.base.LastTrustedHeight()
            if err != nil {
                c.errs <- err
                continue
            }

            if lastTrustedHeight == -1 {
                // no headers yet => wait
                continue
            }

            newTrustedHeader, err := c.base.Update(time.Now())
            if err != nil {
                c.errs <- err
                continue
            }

            if newTrustedHeader != nil {
                 c.trustedHeaders <- newTrustedHeader
            }
        case <-c.quit:
            return
        }
    }
}
```

Later we merged it into the `Client` itself with the assumption that most clients will want it.

But now I am not sure. **Neither IBC nor cosmos/relayer are using it. It increases complexity (Start/Stop methods).**

That said, I think it makes sense to remove it until we see a need for it (until we better understand usage behavior). We can always introduce it later 😅. Maybe in the form of `AutoClient`.

______

For contributor use:

- [x] Wrote tests

### DIFF
--- a/docs/architecture/adr-046-light-client-implementation.md
+++ b/docs/architecture/adr-046-light-client-implementation.md
@@ -8,23 +8,13 @@
 ## Context
 
 A `Client` struct represents a light client, connected to a single blockchain.
-As soon as it's started (via `Start`), it tries to update to the latest header
-(using bisection algorithm by default).
 
-Cleaning routine is also started to remove headers outside of trusting period.
-NOTE: since it's periodic, we still need to check header is not expired in
-`TrustedHeader`, `TrustedValidatorSet` methods (and others which are using the
-latest trusted header).
-
-The user has an option to manually verify headers using `VerifyHeader` and
-`VerifyHeaderAtHeight` methods. To avoid races, `UpdatePeriod(0)` needs to be
-passed when initializing the light client (it turns off the auto update).
+The user has an option to verify headers using `VerifyHeader` or
+`VerifyHeaderAtHeight` or `Update` methods. The latter method downloads the
+latest header from primary and compares it with the currently trusted one.
 
 ```go
 type Client interface {
-	// start and stop updating & cleaning goroutines
-	Start() error
-	Stop()
 	Cleanup() error
 
 	// get trusted headers & validators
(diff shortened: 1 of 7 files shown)


docs/architecture/adr-046-light-client-implementation.md
```diff
@@ -8,23 +8,13 @@
...
+The user has an option to verify headers using `VerifyHeader` or
+`VerifyHeaderAtHeight` or `Update` methods. The latter method downloads the
+latest header from primary and compares it with the currently trusted one.
 
 ```go
 type Client interface {
-	// start and stop updating & cleaning goroutines
-	Start() error
-	Stop()
 	Cleanup() error
```
cmwaters: nit: I would swap this with the verification part of the interface. I like to have them in order of importance

PR #4542, opened 2020-03-09
libs/kv: remove unused type KI64Pair

tessr:
This type does not seem to be used. See also: https://github.com/tendermint/tendermint/pull/4466#discussion_r389352283


tessr:
(Proto-breakage is failing because the generated files changed, but that's the point of this change.)

melekes:
This needs a changelog entry.

tessr:
Good call, thanks.

Issue #1614, opened 2018-05-23
p2p: prevent bad peer from connecting to us for some time


melekes:
Suppose a peer sent us poorly formatted data or tried to DDOS us or did something else. What prevents it from connecting to us again after we stop it by calling `StopPeerForError`? Should we keep it in the address book as bad or refuse to connect for at least some period?

ebuchman:
Yes I think so. Maybe we can add a `badPeers` to the AddrBook that tracks peers that have been evicted and refuses to add them back to the address book for ~1 day or something.

ValarDragon:
I think we should only add IP's to this 'bad peer' section, if we've gotten a response from them that depends on input we gave (either a PoW or secret connection). My worry is that otherwise we will get people who will IP spoof to fill up our badPeers addressbook list. (Or even worse, add all the seed nodes / nodes in the network they know info our bad peers list)

ebuchman:
Note for reference bitcoin has a `banscore` and a `bantime` to determine when to ban a peer and for how long. The default is 1 day: https://bitcoin.org/en/developer-guide#misbehaving-nodes

I think adopting something similar is reasonable - but being banned should depend on erroring at one of the reactors, like sending a bad vote or block part. So it means you have to set up a full secret connection, exchange the node info, and send a bad reactor msg to get banned.

Issue #4415, opened 2020-02-17
[blockchain][v2] Should peer removal be permanent


brapse:
In the current  blockchain v2 scheduler can remove peer if they perform an error, timeout or are purged. Once a peer is removed, re-adding them will cause a fatal error. It isn't clear if this is the correct behaviour. Investigate wether removing a peer from the blockchain reactor scheduler should be permanent.

cmwaters:
I think there should be a distinction between whether a peer performs an error or whether they just timeout. At least with timeouts, I believe it would be better if they weren't removed but just put to the back of a list. I'm sure it could be integrated with the list of peers that are ordered on the amount of pending requests (which is written in the function `selectPeer()`). I think it's best to keep a relatively large list of peers so that you can evenly spread the load. What do you think?

brapse:
Yes the distinction you suggest makes sense. It isn't clear to me how to go about testing different strategies for `selectPeer()` which could substantial these kinds of arguments. What would you suggest here?

cmwaters:
Here is a stab at a few proposals:

**First Suggestion:**

As a rather natural and simple approach, if we were to remove the pruning process of peers on timeout: `rTryPrunePeer` then the peer would still have at least one pending block and would therefore be outcompeted by peers that had zero pending blocks putting it further back on the list of peers to request blocks from. We could even multiply the pending blocks by a factor (i.e by 3) so that is is further pushed back on this list. The main implementation would be handling when the peer eventually came back online and submitted a block at a height lower than the state height in which case `scBlockRecieved` would instead of ignoring the block, remove the pending block and therefore free up the peer again. The disadvantage would be if instead of the network the message itself dropped such that the peer had no idea it had a block requested and therefore had become blocked from the fast sync process. 

**Second Suggestion:**

Have a separate timeout pool that peers are dropped into after timeout and where iteratively based on a timer every peer in the timeout pool is pinged, if they respond they are promoted back into the main pool. This covers the prior suggestions flaw but would most likely be more implementation heavy.

What do you think?

brapse:
Both are perfectly valid suggestions! However my question was how would we compare different approaches. Even though we can do something here, both of your suggestions look good, it isn't clear what we would be improving and how we would measure such an improvement.

My intuition is that banning peers permanently results in some sending nodes being under utilized while some receiving nodes will somehow be starved of blocks. If that is the case then it should be observed first. A first step here might be to add some metrics around:
* Count of peers in which state
[...]

Issue #1444, opened 2018-04-09
p2p: certain peer errors should cause us to remove the address


ebuchman:
If we can't reach a peer, we should keep the address around.

However, if we successfully connect to the remote endpoint but fail to handshake or otherwise we should remove the peer from the address book so we don't keep trying.

This is really just an optimization (if we fail to connect enough times, we'll boot the peer), but it will make for a better UX

Also might be nice to remember peers we've removed from the addrbook for some time so we can avoid reconnecting

melekes:
> If we can't reach a peer, we should keep the address around.

what for?

So, currently, we remove the address if the peer is not available (16 attempts) or it fails authentication.

> Also might be nice to remember peers we've removed from the addrbook for some time so we can avoid reconnecting

https://github.com/tendermint/tendermint/pull/4548 does exactly that

tessr:
I think this is covered by #4548. I'm going to consider this issue closed!

PR #4548, opened 2020-03-10
p2p: ban bad peers 

cmwaters:
Closes: #1614

## Description

This is just a MVP implementation of having a blacklist that bans peers for a bantime (default is 1 day) and I do feel that this is potentially a bandage to be replaced when the entire p2p module gets redone. Below is a illustration of the implementation.

p2p

Reinstated peers are added to the "new" (unvetted) bucket. 

**Thoughts:**

The `ReinstateBadPeers()` function is run by the synchronous function `ensurePeers()`every 30 seconds if the `NeedMoreAddrs()` fails. The alternative would be to have `isBanned()` check within parts of the code that want to use the address.

It might be a good idea to also keep a count of the amount of bans such that repeat offenders can serve longer bans / be banned forever. 

______

For contributor use:

- [ ] Wrote tests


p2p/pex/known_address.go
```diff
@@ -54,6 +55,17 @@ func (ka *knownAddress) markGood() {
 	ka.LastSuccess = now
 }
 
+func (ka *knownAddress) ban() {
+	ka.LastBanTime = time.Now()
+}
+
+func (ka *knownAddress) isBanned(banTime time.Duration) bool {
+	if ka.LastBanTime.Add(banTime).Before(time.Now()) {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

p2p/pex/pex_reactor.go
```diff
@@ -8,7 +8,7 @@ import (
 
 	"github.com/pkg/errors"
 
-	amino "github.com/tendermint/go-amino"
```
tessr: why did this change?

p2p/pex/addrbook.go
```diff
@@ -324,10 +329,24 @@ func (a *addrBook) MarkAttempt(addr *p2p.NetAddress) {
...
-	a.RemoveAddress(addr)
+	if a.addBadPeer(addr) {
+		a.RemoveAddress(addr)
+	}
+}
+
+func (a *addrBook) ReinstateBadPeers() {
+	for _, ka := range a.badPeers {
+		if !ka.isBanned(defaultBanTime) {
+			a.mtx.Lock()
```
erikgrinaker: The mutex lock must be taken out at the start of the function call, otherwise a concurrent thread can modify `badPeers` while you are iterating over it and then bad things happen.

p2p/pex/addrbook.go
```diff
@@ -324,10 +329,24 @@ func (a *addrBook) MarkAttempt(addr *p2p.NetAddress) {
...
 }
 
-// MarkBad implements AddrBook. Currently it just ejects the address.
-// TODO: black list for some amount of time
+// MarkBad implements AddrBook. Kicks address out from book, places
+// the address in the badPeers pool.
 func (a *addrBook) MarkBad(addr *p2p.NetAddress) {
-	a.RemoveAddress(addr)
+	if a.addBadPeer(addr) {
+		a.RemoveAddress(addr)
```
erikgrinaker: `addBadPeer()` returns `false` if the peer is already considered bad, in which case it will not be removed from the address book. Shouldn't we always remove it regardless, since we do not want bad peers there under any circumstances? We may also want to refresh the ban timer in that case.

Also, since both of these calls take out mutex locks separately, the data can change between them, which may cause problems. Not sure if that will be an actual problem here, but in general it's a good idea to take out mutex locks in public methods, and have internal methods that assume locks are already held such that they can be combined and composed.

p2p/pex/known_address.go
```diff
@@ -54,6 +55,14 @@ func (ka *knownAddress) markGood() {
 	ka.LastSuccess = now
 }
 
+func (ka *knownAddress) ban() {
+	ka.LastBanTime = time.Now()
+}
+
+func (ka *knownAddress) isBanned(banTime time.Duration) bool {
```
erikgrinaker: I don't think it's necessary to pass banTime as a parameter here, maybe just use the constant directly.

cmwaters: I was considering whether we want to make it something that is either a variant - or at least configurable

erikgrinaker: Yeah, that makes sense. But I don't think it's something that all callers should have to deal with every time time they want to check if someone is banned, it's more convenient if `isBanned()` just takes all conditions into account automatically. But I'm not sure what the best way is to inject variables into the current architecture - I guess this is fine for a first iteration.

erikgrinaker: One option may be to have `ban()` take a parameter for how long to ban, and store `bannedUntil` as the time when the ban ends (the max of the current and given). Then `isBanned()` can simply compare `time.Now()` against `bannedUntil`.

cmwaters: yes that could be a good idea

p2p/pex/pex_reactor.go
cmwaters: I am not sure. My main guess would be that my IDE did it in the background

p2p/pex/known_address.go
cmwaters: Do we want to have increasing the ban time i.e. keep count of the amount of bans and therefore increase the duration based on the amount of bans? or is that unnecessary complexity

erikgrinaker: I don't think that's necessary in the first iteration. But it did occur to me that passing the ban time to `ban()` would allow us to vary the ban duration for different offences, if that's something we would need later.

p2p/pex/addrbook.go
cmwaters: I guess if it's bad but still in part of the address book then we should return true so that it still runs RemoveAddress. 

For your second paragraph about the mutexes, that makes sense

p2p/pex/known_address.go
cmwaters: Would we perhaps want to expose `Ban()` so other modules could use it?

erikgrinaker: Not sure. Let's consider than when there's an actual need for it.

cmwaters: Okay we can leave it for later. I was talking to Tess how each module sort of has their own way of categorising and dealing with peers that makes things at a whole inconsistent and potentially repetitive

erikgrinaker: For sure. That's something to consider for the p2p refactor.

p2p/pex/addrbook.go
cmwaters: actually I remember why I didn't add the mutexes in the public function because RemoveAddress has it's own mutex

erikgrinaker: Exactly. Which means that `addBadPeer` and `RemoveAddress` can't be called in a single critical section. If we had `removeAddress()` which did not take out a lock, and `RemoveAddress()` which does take out a lock and call `removeAddress()` internally, then we could use `removeAddress()` in a critical section together with `addBadPeer()`.

Again, not sure if it's worth spending time on in this case, since I suspect it's safe to call `RemoveAddress()` regardless of any state changes. But this is a general problem across the code base which we should keep in mind.

cmwaters: ah yes I understand. I think I will change it like that anyway

p2p/pex/addrbook_test.go
```diff
@@ -3,18 +3,17 @@ package pex
...
-	"testing"
-
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-
 	"github.com/tendermint/tendermint/libs/log"
 	tmmath "github.com/tendermint/tendermint/libs/math"
 	tmrand "github.com/tendermint/tendermint/libs/rand"
 	"github.com/tendermint/tendermint/p2p"
+	"io/ioutil"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

cmwaters:
Current usages of the ban list: 

- If a peer is supposed to be a persistent peer but doesn't respond after x amount of attempts
- when `ErrSwitchAuthenticationFailure` happens which I don't think ever happens

This makes me think do we differentiate based on whether a peer timed out or if they actually misbehaved  (https://github.com/tendermint/tendermint/issues/4415#issuecomment-595804887) and whether we want a banlist and a blacklist - one to not talk to a node for a while and the other to permanently never talk to that node - note that removing the address at the moment isn't enough because we don't keep track of it and the node can simply join again #1444

p2p/pex/addrbook.go
```diff
@@ -222,6 +222,14 @@ func (a *addrBook) IsGood(addr *p2p.NetAddress) bool {
 	return a.addrLookup[addr.ID].isOld()
 }
 
+func (a *addrBook) IsBanned(addr *p2p.NetAddress) bool {
+	a.mtx.Lock()
+	defer a.mtx.Unlock()
+
```
melekes: a.mtx.Lock()
_, ok := a.badPeers[addr.ID]
a.mtx.Unlock()

p2p/pex/addrbook.go
```diff
@@ -324,10 +332,28 @@ func (a *addrBook) MarkAttempt(addr *p2p.NetAddress) {
...
+func (a *addrBook) ReinstateBadPeers() {
+	a.mtx.Lock()
+	defer a.mtx.Unlock()
+	for _, ka := range a.badPeers {
+		if !ka.isBanned() {
+			bucket := a.calcNewBucket(ka.Addr, ka.Src)
+			a.addToNewBucket(ka, bucket)
+			delete(a.badPeers, ka.ID())
+			a.Logger.Info("Reinstated Address", "addr", ka.Addr)
```
melekes:
```suggestion
			a.Logger.Info("Reinstated address", "addr", ka.Addr)
```

p2p/pex/errors.go
```diff
@@ -63,3 +63,11 @@ type ErrAddrBookInvalidAddr struct {
 func (err ErrAddrBookInvalidAddr) Error() string {
 	return fmt.Sprintf("Cannot add invalid address %v: %v", err.Addr, err.AddrErr)
 }
+
+type ErrAddressBanned struct {
```
melekes: please add documentation

p2p/pex/pex_reactor.go
```diff
@@ -529,7 +538,7 @@ func (r *Reactor) dialPeer(addr *p2p.NetAddress) error {
 		// failed to connect to. Then we can clean up attemptsToDial, which acts as
```
melekes: can you remove this comment?

I think we need both attemptsToDial and blacklist

p2p/pex/addrbook.go
cmwaters: can you explain to me the difference?

p2p/pex/errors.go
cmwaters: What about for other errors - there isn't any documentation for them either

melekes: could you please write doc for them too?

p2p/pex/addrbook.go
melekes: no need in defer if you don't call other functions and code is not complex

p2p/pex/addrbook.go
```diff
@@ -725,6 +756,31 @@ func (a *addrBook) moveToOld(ka *knownAddress) {
...
+	}
+	a.Logger.Info("Remove address from book", "addr", addr)
+	a.removeFromAllBuckets(ka)
+}
+
+func (a *addrBook) addBadPeer(addr *p2p.NetAddress, banTime time.Duration) bool {
+	// check it exists in addrbook
+	ka := a.addrLookup[addr.ID]
+	// check address is not already there
+	if ka != nil {
```
erikgrinaker: Can simplify this with an early return, i.e. `if ka == nil { return false }`

p2p/pex/known_address.go
```diff
@@ -54,6 +55,14 @@ func (ka *knownAddress) markGood() {
 	ka.LastSuccess = now
 }
 
+func (ka *knownAddress) ban(banTime time.Duration) {
+	ka.LastBanTime = time.Now().Add(banTime)
```
erikgrinaker: I think this should only extend the ban, not shorten it. For example, if someone banned the address until 16:00, and then someone else wants to ban them until 13:00, they should remain banned until 16:00. This will matter when we start having different ban periods for various offences.

p2p/pex/pex_reactor.go
```diff
@@ -8,7 +8,7 @@ import (
 
 	"github.com/pkg/errors"
 
-	amino "github.com/tendermint/go-amino"
+	"github.com/tendermint/go-amino"
```
erikgrinaker: Should keep the old formatting. It's considered good form to use an explicit name if the package name differs from the URL name (`amino` vs `go-amino`), and to use a blank line between external and internal dependencies.

p2p/pex/errors.go
```diff
@@ -63,3 +63,12 @@ type ErrAddrBookInvalidAddr struct {
 func (err ErrAddrBookInvalidAddr) Error() string {
 	return fmt.Sprintf("Cannot add invalid address %v: %v", err.Addr, err.AddrErr)
 }
+
+// Err is thrown when the address is banned and therefore cannot be used
```
erikgrinaker: Go comments should start with name of what they're describing, i.e. `ErrAddressBanned is thrown ...`. I thought the linter would catch this, but we may have disabled it.

p2p/pex/pex_reactor.go
cmwaters: for some reason my ide doesn't see the difference between amino and go-amino but I will make the change regardless

melekes:
Where's the changelog pending entry?

melekes:
Also, tendermint/spec needs to be updated.

PR #4332, opened 2020-01-22
more linting and some types to proto

marbar3778:
- add more proto linting
- intorduce gogoslick as proto comiler 
- make blockIDFlag & signedmessageType proto enums
- make blockID & partsetHeader proto types

This PR is going into proto-breakage branch

* [ ] Referenced an issue explaining the need for the change
* [ ] Updated all relevant documentation in docs
* [ ] Updated all code comments where relevant
* [ ] Wrote tests
* [ ] Updated CHANGELOG_PENDING.md

PR #4557, opened 2020-03-11
proto: change int32 to uint32

marbar3778:
## Description

- address comments post merge from: https://github.com/tendermint/tendermint/pull/4332

proto goes down to int32/uint32 it does not provide a smaller size

For contributor use:

- [ ] Wrote tests


types/part_set.go
```diff
@@ -79,17 +73,18 @@ func (psh PartSetHeader) ValidateBasic() error {
...
 	parts         []*Part
 	partsBitArray *bits.BitArray
-	count         int32
+	count         uint32
 }
 
 // Returns an immutable, full PartSet from the data bytes.
 // The data bytes are split into "partSize" chunks, and merkle tree computed.
+// CONTRACT: partSize is greater than zero.
 func NewPartSetFromData(data []byte, partSize int) *PartSet {
```
erikgrinaker: Should we change partSize to be an uint64/32 too?

types/part_set.go
```diff
@@ -98,7 +93,7 @@ func NewPartSetFromData(data []byte, partSize int) *PartSet {
 	partsBitArray := bits.NewBitArray(total)
 	for i := 0; i < total; i++ {
 		part := &Part{
-			Index: int32(i),
+			Index: uint32(i),
```
erikgrinaker: May as well use uint32 in the for loop, i.e. `for i := uint32(0); ...`

types/part_set.go
```diff
@@ -111,11 +106,11 @@ func NewPartSetFromData(data []byte, partSize int) *PartSet {
 		parts[i].Proof = *proofs[i]
 	}
 	return &PartSet{
-		total:         int32(total),
+		total:         uint32(total),
```
erikgrinaker: Make `total` a uint32 in the first place.

types/part_set_test.go
```diff
@@ -21,17 +21,17 @@ func TestBasicPartSet(t *testing.T) {
 	partSet := NewPartSetFromData(data, testPartSize)
 
 	assert.NotEmpty(t, partSet.Hash())
-	assert.Equal(t, int32(100), partSet.Total())
+	assert.Equal(t, uint32(100), partSet.Total())
```
erikgrinaker: You can use `assert.EqualValues()` to ignore integer types.

types/part_set.go
marbar3778: is it better to cast to uint32 or to int?

erikgrinaker: If the number should never be negative, I'd say it should be a uint. And sized is better than usized, since `uint` will be either `uint32` or `uint64` depending on the hardware.

Issue #4546, opened 2020-03-10
Use caching of headers to improve bisection performance


cmwaters:
## Summary

I think the title is sufficient of a summary

## Problem Definition

As raised in #4504, non-recursive bisection suffers from the problem whenever it bisects the gap, that it will need to ask for the validator set from the primary again.

For a simple example. If in order to validate 256 blocks you must

grab block 256 -> fail -> grab block 128 -> fail -> grab block 64 -> success
test block 256 -> fail -> grab block 160 -> success
test block 256 -> success

It would be better however to cache block 128 and use this instead, making it most likely quicker than retrieving block 160

## Proposal

Have a map of retrieved blocks that the light clients caches and refers to instead of asking for a full node to send the light client the blocks

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

tessr:
> I think the title is sufficient of a summary

lol, you can just delete the section if you feel it's not relevant!

PR #4562, opened 2020-03-12
lite2: cache headers in bisection

cmwaters:
Closes: #4546

## Description

In short, we avoid asking primary providers for extra headers and validators

tendermint

The algorithm uses an array to store the headers and validators and populates it at every bisection (which is an unsuccessful verification). When a successful verification finally occurs it updates the new trusted header, trims that header from the cache (the array) and sets the depth pointer back to 0. Instead of retrieving new headers it will use the cached headers, incrementing in depth until it reaches the end of the cache which by then it will start to retrieve new headers from the provider. 

Mathematically, this method doesn't properly bisect after the first round but it will always choose a pivot header that is within 1/8th of the upper header's height. I.e. if we are trying to jump 128 headers, the maximum offset from bisection height (64) is 64 + 16(128/8) = 80, therefore a better heuristic would be to obtain the new pivot header height as the middle of these two numbers which would therefore mean to multiply it by 9/16ths instead of 1/2  (sorry this might be a bit more complicated in writing but I can try better explain if someone is interested). Therefore I would also, upon consensus, propose that we change the pivot height to 9/16th's of the previous height

______

For contributor use:

- [ ] Wrote tests


lite2/client.go
```diff
@@ -714,39 +714,52 @@ func (c *Client) bisection(
 	newVals *types.ValidatorSet,
 	now time.Time) error {
 
+	type HeaderSet struct {
+		sh  *types.SignedHeader
+		val *types.ValidatorSet
+	}
+
 	var (
+		headerCache = []HeaderSet{HeaderSet{newHeader, newVals}}
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
		headerCache = []HeaderSet{{newHeader, newVals}}
```

lite2/client.go
```diff
@@ -714,39 +714,52 @@ func (c *Client) bisection(
...
 		c.logger.Debug("Verify newHeader against trustedHeader",
 			"trustedHeight", trustedHeader.Height,
 			"trustedHash", hash2str(trustedHeader.Hash()),
-			"newHeight", interimHeader.Height,
-			"newHash", hash2str(interimHeader.Hash()))
+			"newHeight", headerCache[depth].sh.Height,
+			"newHash", hash2str(headerCache[depth].sh.Hash()))
 
-		err := Verify(c.chainID, trustedHeader, trustedVals, interimHeader, interimVals, c.trustingPeriod, now,
+		err := Verify(c.chainID, trustedHeader, trustedVals, headerCache[depth].sh, headerCache[depth].val, c.trustingPeriod, now,
```
golangcibot: line is 124 characters (from `lll`)

melekes:
> The algorithm uses an array

is it bounded? Do you think it make sense to make it a fixed size to ensure it doesn't grow indefinitely => Out of Memory exception?

lite2/client.go
```diff
@@ -714,39 +714,52 @@ func (c *Client) bisection(
 	newVals *types.ValidatorSet,
 	now time.Time) error {
 
+	type HeaderSet struct {
```
melekes: any particular reason this is exposed?

lite2/client.go
```diff
@@ -714,39 +714,52 @@ func (c *Client) bisection(
...
+		sh  *types.SignedHeader
+		val *types.ValidatorSet
```
```suggestion
		valSet *types.ValidatorSet
```

lite2/client.go
```diff
@@ -714,39 +714,52 @@ func (c *Client) bisection(
...
+	}
+
 	var (
+		headerCache = []HeaderSet{{newHeader, newVals}}
```
melekes: can't we just use `map[height]{Header, Vals}`? how's having a separate depth variable  better?

cmwaters: because we are jumping from header to header we just want a queue not a map, having height is not useful, we just increment the array

lite2/client.go
cmwaters: didn't realise that a struct declared in a function was exposed - I thought it was by default only available locally within that function

cmwaters:
> > The algorithm uses an array
> 
> is it bounded? Do you think it make sense to make it a fixed size to ensure it doesn't grow indefinitely => Out of Memory exception?

Sure I could set a capacity but even then I don't see the array getting too big. For example if we wanted to verify a million headers at once, this cache would, at maximum, reach a size of 20 => 2^20 = 1048576

melekes:
> Sure I could set a capacity but even then I don't see the array getting too big. For example if we wanted to verify a million headers at once, this cache would, at maximum, reach a size of 20 => 2^20 = 1048576

Fair. Let's go with unbounded then.

lite2/client.go
melekes: I guess what I was trying to say code is hard to understand without a visual. Can we at least add the above image to doc.go file or somewhere else. or create an asci version

cmwaters: I will try add some more clearer descriptions to doc.go and to the actual bisection method

lite2/client.go
```diff
@@ -504,11 +481,11 @@ func (c *Client) VerifyHeaderAtHeight(height int64, now time.Time) (*types.Signe
 
 // VerifyHeader verifies new header against the trusted state. It returns
 // immediately if newHeader exists in trustedStore (no verification is
-// needed).
+// needed). Else it performs one of the two types of verification:
 //
 // SequentialVerification: verifies that 2/3 of the trusted validator set has
 // signed the new header. If the headers are not adjacent, **all** intermediate
-// headers will be requested.
+// headers will be requested. Intermediate headers are not saved to database.
```
melekes: same applies for SkippingVerification below

lite2/client.go
```diff
@@ -707,46 +625,64 @@ func (c *Client) sequence(
...
+			// do add another header to the end of the cache
+			if depth == len(headerCache)-1 {
+				pivotHeight := (headerCache[depth].sh.Height + trustedHeader.Height) / 2
+				interimHeader, interimVals, err := c.fetchHeaderAndValsAtHeight(pivotHeight)
+				if err != nil {
+					return err
+				}
+				depth++
+				headerCache = append(headerCache, headerSet{interimHeader, interimVals})
+			} else { // or we shift to the next header in the cache
```
melekes: can't we execute depth++ after if and drop else branch? we have it in both branches.

cmwaters: Yup we can do that

lite2/client.go
```diff
@@ -675,7 +675,7 @@ func (c *Client) bisection(
 		case ErrNewValSetCantBeTrusted:
 			// do add another header to the end of the cache
 			if depth == len(headerCache)-1 {
-				pivotHeight := (headerCache[depth].sh.Height + trustedHeader.Height) / 2
+				pivotHeight := (headerCache[depth].sh.Height + trustedHeader.Height) * 9 / 16
```
melekes: magic const

cmwaters: there is some rationale behind the madness

melekes: could you please extract it into a const and add a comment?

cmwaters: sure :+1:

Issue #4537, opened 2020-03-06
Use bisection for parts of backwards verification


cmwaters:
## Summary

Given bisections speed, it would be faster when verifying a header that is not greater in height than the latest stored header, to find the trusted header in the store directly **before** the untrusted header and use `bisection` instead of `backwards` to verify the untrusted header.

## Problem Definition

Bisection is faster then backwards verification

## Proposal

Change the store interface by replacing `SignedHeaderAfter(height int) (*types.SignedHeader, error) `with `SignedHeaderBefore(height int) (*types.SignedHeader, error)`

When verifying a header prior to the latest block, check if the header is less in height than the `firstTrustedHeaderHeight`. If so, perform backwards verification like beforehand but use the header at the first trusted height as the initial trusted header. If not, then use the trusted header from `SignedHeaderBefore` and then use bisection to verify the untrusted header.  

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

milosevic:
This is fine. We just need to take care that with bisection trusted header need to be within trusted period.

Issue #4538, opened 2020-03-07
light: standardize error handling


melekes:
See https://github.com/tendermint/tendermint/pull/3989#issuecomment-556189223

Priority: low

PR #4575, opened 2020-03-16
lite2: use bisection for some of backward verification

cmwaters:
Closes: #4537 

## Description

Uses `SignedHeaderBefore` to find header before unverified header and then bisection to verify the header. Only when header is between first and last trusted header height else if before the first trusted header height then regular backwards verification is used. 

______

For contributor use:

- [x] Wrote tests

### DIFF
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -544,6 +544,7 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
 			panic(fmt.Sprintf("Unknown verification mode: %b", c.verificationMode))
 		}
 	} else {
+		// 2) If verifying before the first trusted header, perform backwards verification
 		var firstHeaderHeight int64
 		firstHeaderHeight, err = c.FirstTrustedHeight()
 		if err != nil {
@@ -559,6 +560,7 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
 				closestHeader = c.latestTrustedHeader
 			}
 			err = c.backwards(closestHeader, newHeader, now)
+			// 3) or if between trusted headers where the nearest has not expired, perform bisection verification, else backwards
 		} else {
 			closestHeader, err = c.trustedStore.SignedHeaderBefore(newHeader.Height)
 			if err != nil {
@@ -580,12 +582,13 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
 		c.logger.Error("Can't verify", "err", err)
 		return err
 	}
-
+	// 4) Compare header with other witnesses
 	if err := c.compareNewHeaderWithWitnesses(newHeader); err != nil {
 		c.logger.Error("Error when comparing new header with witnesses", "err", err)
 		return err
 	}
 
+	// 5) Once verified, save and return
 	return c.updateTrustedHeaderAndVals(newHeader, newVals)
 }
 
@@ -1031,7 +1034,7 @@ func (c *Client) signedHeaderFromPrimary(height int64) (*types.SignedHeader, err
 		if err == nil {
 			// sanity check
 			if height > 0 && h.Height != height {
-				return nil, fmt.Errorf("expected %d height, got %d: %w", height, h.Height, err)
+				return nil, fmt.Errorf("expected %d height, got %d", height, h.Height)
 			}
 			return h, nil
 		}
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -206,7 +206,7 @@ func NewClientFromTrustedStore(
 
 	// Validate the number of witnesses.
 	if len(c.witnesses) < 1 {
-		return nil, ErrNoWitnesses{}
+		return nil, errNoWitnesses{}
 	}
 
 	// Verify witnesses are all on the same chain.
@@ -913,7 +913,7 @@ func (c *Client) compareNewHeaderWithWitnesses(h *types.SignedHeader) error {
 	witnessesToRemove := make([]int, 0)
 	for attempt := uint16(1); attempt <= c.maxRetryAttempts; attempt++ {
 		if len(c.witnesses) == 0 {
-			return ErrNoWitnesses{}
+			return errNoWitnesses{}
 		}
 
 		for i, witness := range c.witnesses {
@@ -1014,7 +1014,7 @@ func (c *Client) replacePrimaryProvider() error {
 	defer c.providerMutex.Unlock()
 
 	if len(c.witnesses) <= 1 {
-		return ErrNoWitnesses{}
+		return errNoWitnesses{}
 	}
 	c.primary = c.witnesses[0]
 	c.witnesses = c.witnesses[1:]
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -499,6 +499,11 @@ func (c *Client) VerifyHeaderAtHeight(height int64, now time.Time) (*types.Signe
 // Intermediate headers are not saved to database.
 // https://github.com/tendermint/spec/blob/master/spec/consensus/light-client.md
 //
+// If the header, which is older than the currently trusted header, is
+// requested and the light client does not have it, VerifyHeader will perform:
+//		a) bisection verification if nearest trusted header is found & not expired
+//		b) backwards verification in all other cases
+//
 // It returns ErrOldHeaderExpired if the latest trusted header expired.
 //
 // If the primary provides an invalid header (ErrInvalidHeader), it is rejected
@@ -533,7 +538,7 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
 
 	var err error
 
-	// 1) If going forward, perform either bisection or sequential verification
+	// 1) If going forward, perform either bisection or sequential verification.
 	if newHeader.Height >= c.latestTrustedHeader.Height {
 		switch c.verificationMode {
 		case sequential:
@@ -544,13 +549,16 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
 			panic(fmt.Sprintf("Unknown verification mode: %b", c.verificationMode))
 		}
 	} else {
-		// 2) If verifying before the first trusted header, perform backwards verification
-		var firstHeaderHeight int64
+		// 2) If verifying before the first trusted header, perform backwards
+		// verification.
+		var (
+			closestHeader     *types.SignedHeader
+			firstHeaderHeight int64
+		)
 		firstHeaderHeight, err = c.FirstTrustedHeight()
 		if err != nil {
 			return fmt.Errorf("can't get first header height: %w", err)
 		}
-		var closestHeader *types.SignedHeader
 		if newHeader.Height < firstHeaderHeight {
 			closestHeader, err = c.TrustedHeader(firstHeaderHeight)
 			if err != nil {
@@ -560,8 +568,9 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
 				closestHeader = c.latestTrustedHeader
 			}
 			err = c.backwards(closestHeader, newHeader, now)
-			// 3) or if between trusted headers where the nearest has not expired, perform bisection verification, else backwards
 		} else {
+			// 3) OR if between trusted headers where the nearest has not expired,
+			// perform bisection verification, else backwards.
 			closestHeader, err = c.trustedStore.SignedHeaderBefore(newHeader.Height)
 			if err != nil {
 				return fmt.Errorf("can't get signed header before height %d: %w", newHeader.Height, err)
--- a/lite2/client.go
+++ b/lite2/client.go
@@ -149,7 +149,7 @@ func NewClient(
 	options ...Option) (*Client, error) {
 
 	if err := trustOptions.ValidateBasic(); err != nil {
-		return nil, errors.Wrap(err, "invalid TrustOptions")
+		return nil, fmt.Errorf("invalid TrustOptions: %w", err)
 	}
 
 	c, err := NewClientFromTrustedStore(chainID, trustOptions.Period, primary, witnesses, trustedStore, options...)
@@ -206,13 +206,13 @@ func NewClientFromTrustedStore(
 
 	// Validate the number of witnesses.
 	if len(c.witnesses) < 1 {
-		return nil, errors.New("expected at least one witness")
+		return nil, ErrNoWitnesses{}
 	}
 
 	// Verify witnesses are all on the same chain.
 	for i, w := range witnesses {
 		if w.ChainID() != chainID {
-			return nil, errors.Errorf("witness #%d: %v is on another chain %s, expected %s",
+			return nil, fmt.Errorf("witness #%d: %v is on another chain %s, expected %s",
 				i, w, w.ChainID(), chainID)
 		}
 	}
@@ -234,18 +234,18 @@ func NewClientFromTrustedStore(
 func (c *Client) restoreTrustedHeaderAndVals() error {
 	lastHeight, err := c.trustedStore.LastSignedHeaderHeight()
 	if err != nil {
-		return errors.Wrap(err, "can't get last trusted header height")
+		return fmt.Errorf("can't get last trusted header height: %w", err)
 	}
 
 	if lastHeight > 0 {
 		trustedHeader, err := c.trustedStore.SignedHeader(lastHeight)
 		if err != nil {
-			return errors.Wrap(err, "can't get last trusted header")
+			return fmt.Errorf("can't get last trusted header: %w", err)
 		}
 
 		trustedVals, err := c.trustedStore.ValidatorSet(lastHeight)
 		if err != nil {
-			return errors.Wrap(err, "can't get last trusted validators")
+			return fmt.Errorf("can't get last trusted validators: %w", err)
 		}
 
 		c.latestTrustedHeader = trustedHeader
@@ -300,7 +300,7 @@ func (c *Client) checkTrustedHeaderUsingOptions(options TrustOptions) error {
 			// remove all the headers (options.Height, trustedHeader.Height]
 			err := c.cleanupAfter(options.Height)
 			if err != nil {
-				return errors.Wrapf(err, "cleanupAfter(%d)", options.Height)
+				return fmt.Errorf("cleanupAfter(%d): %w", options.Height, err)
 			}
 
 			c.logger.Info("Rolled back to older header (newer headers were removed)",
@@ -322,7 +322,7 @@ func (c *Client) checkTrustedHeaderUsingOptions(options TrustOptions) error {
 		if c.confirmationFn(action) {
 			err := c.Cleanup()
 			if err != nil {
-				return errors.Wrap(err, "failed to cleanup")
+				return fmt.Errorf("failed to cleanup: %w", err)
 			}
 		} else {
 			return errors.New("refused to remove the stored headers despite hashes mismatch")
@@ -350,7 +350,7 @@ func (c *Client) initializeWithTrustOptions(options TrustOptions) error {
 	}
 
 	if !bytes.Equal(h.Hash(), options.Hash) {
-		return errors.Errorf("expected header's hash %X, but got %X", options.Hash, h.Hash())
+		return fmt.Errorf("expected header's hash %X, but got %X", options.Hash, h.Hash())
 	}
 
 	err = c.compareNewHeaderWithWitnesses(h)
@@ -365,7 +365,7 @@ func (c *Client) initializeWithTrustOptions(options TrustOptions) error {
 	}
 
 	if !bytes.Equal(h.ValidatorsHash, vals.Hash()) {
-		return errors.Errorf("expected header's validators (%X) to match those that were supplied (%X)",
+		return fmt.Errorf("expected header's validators (%X) to match those that were supplied (%X)",
 			h.ValidatorsHash,
 			vals.Hash(),
 		)
@@ -374,7 +374,7 @@ func (c *Client) initializeWithTrustOptions(options TrustOptions) error {
 	// Ensure that +2/3 of validators signed correctly.
 	err = vals.VerifyCommit(c.chainID, h.Commit.BlockID, h.Height, h.Commit)
 	if err != nil {
-		return errors.Wrap(err, "invalid commit")
+		return fmt.Errorf("invalid commit: %w", err)
 	}
 
 	// 3) Persist both of them and continue.
@@ -436,7 +436,7 @@ func (c *Client) TrustedValidatorSet(height int64) (valSet *types.ValidatorSet, 
 func (c *Client) compareWithLatestHeight(height int64) (int64, error) {
 	latestHeight, err := c.LastTrustedHeight()
 	if err != nil {
-		return 0, errors.Wrap(err, "can't get last trusted height")
+		return 0, fmt.Errorf("can't get last trusted height: %w", err)
 	}
 	if latestHeight == -1 {
 		return 0, errors.New("no headers exist")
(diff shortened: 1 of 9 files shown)


cmwaters:
Will I need to update the spec regarding the slight tweak in implementation

tessr:
test

tessr:
testing

lite2/store/store.go
```diff
@@ -44,7 +44,7 @@ type Store interface {
 	// SignedHeaderAfter returns the SignedHeader after the certain height.
```
melekes: please update

lite2/store/db/db_test.go
```diff
@@ -76,19 +76,19 @@ func Test_SaveSignedHeaderAndValidatorSet(t *testing.T) {
 	assert.Nil(t, valSet)
 }
 
-func Test_SignedHeaderAfter(t *testing.T) {
+func Test_SignedHeaderBefore(t *testing.T) {
 	dbStore := New(dbm.NewMemDB(), "Test_SignedHeaderAfter")
```
melekes:
```suggestion
	dbStore := New(dbm.NewMemDB(), "Test_SignedHeaderBefore")
```

lite2/store/db/db.go
```diff
@@ -207,14 +207,14 @@ func (s *dbs) FirstSignedHeaderHeight() (int64, error) {
 // height. It returns ErrSignedHeaderNotFound if no such header exists.
```
melekes: please don't forget to update documentation

lite2/store/db/db.go
```diff
@@ -207,14 +207,14 @@ func (s *dbs) FirstSignedHeaderHeight() (int64, error) {
...
 	if height <= 0 {
 		panic("negative or zero height")
 	}
 
-	itr, err := s.db.Iterator(
-		s.shKey(height+1),
-		append(s.shKey(1<<63-1), byte(0x00)),
+	itr, err := s.db.ReverseIterator(
+		nil,
+		s.shKey(height),
```
melekes: shouldn't we start with height - 1?

lite2/store/db/db.go
```diff
@@ -207,14 +207,14 @@ func (s *dbs) FirstSignedHeaderHeight() (int64, error) {
...
+func (s *dbs) SignedHeaderBefore(height int64) (*types.SignedHeader, error) {
 	if height <= 0 {
 		panic("negative or zero height")
 	}
 
-	itr, err := s.db.Iterator(
-		s.shKey(height+1),
-		append(s.shKey(1<<63-1), byte(0x00)),
+	itr, err := s.db.ReverseIterator(
+		nil,
```
melekes: don't we need to have a stop key?

lite2/client.go
```diff
@@ -814,6 +831,7 @@ func (c *Client) backwards(
 	now time.Time) error {
 
 	if HeaderExpired(initiallyTrustedHeader, c.trustingPeriod, now) {
+		c.logger.Info("Header Expired")
```
melekes: Shouldn't this be an error?

lite2/client.go
```diff
@@ -561,15 +561,32 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
...
 	} else {
-		// 2) Otherwise, perform backwards verification
-		// Find the closest trusted header after newHeader.Height
-		var closestHeader *types.SignedHeader
-		closestHeader, err = c.trustedStore.SignedHeaderAfter(newHeader.Height)
+		var firstHeaderHeight int64
+		firstHeaderHeight, err = c.FirstTrustedHeight()
 		if err != nil {
-			return errors.Wrapf(err, "can't get signed header after height %d", newHeader.Height)
+			return err
```
melekes: errors.Wrap?

lite2/client.go
```diff
@@ -561,15 +561,32 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
...
+		}
+		var closestHeader *types.SignedHeader
+		if newHeader.Height < firstHeaderHeight {
+			c.logger.Info("Using backwards")
+			closestHeader, err = c.TrustedHeader(firstHeaderHeight)
+			if err != nil {
+				return errors.Wrapf(err, "can't get signed header after height %d", newHeader.Height)
+			}
+			err = c.backwards(closestHeader, newHeader, now)
+			//if err != nil { return err }
```

lite2/client.go
```diff
@@ -561,15 +561,32 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
...
+		var firstHeaderHeight int64
+		firstHeaderHeight, err = c.FirstTrustedHeight()
 		if err != nil {
-			return errors.Wrapf(err, "can't get signed header after height %d", newHeader.Height)
+			return err
+		}
+		var closestHeader *types.SignedHeader
+		if newHeader.Height < firstHeaderHeight {
+			c.logger.Info("Using backwards")
+			closestHeader, err = c.TrustedHeader(firstHeaderHeight)
```
melekes: ??? why var name is closestHeader

lite2/client.go
```diff
@@ -561,15 +561,32 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
...
+			closestHeader, err = c.trustedStore.SignedHeaderBefore(newHeader.Height)
+			if err != nil {
+				return errors.Wrapf(err, "can't get signed header before height %d", newHeader.Height)
+			}
+			var closestValidatorSet *types.ValidatorSet
+			closestValidatorSet, _, err = c.TrustedValidatorSet(closestHeader.Height)
+			if err != nil {
+				return errors.Wrapf(err, "can't get validator set at height %d", closestHeader.Height)
+			}
+			err = c.bisection(closestHeader, closestValidatorSet, newHeader, newVals, now)
```
melekes: should we fallback to backwards verification if this fails?

lite2/client.go
```diff
@@ -561,15 +561,32 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
...
-		closestHeader, err = c.trustedStore.SignedHeaderAfter(newHeader.Height)
+		var firstHeaderHeight int64
+		firstHeaderHeight, err = c.FirstTrustedHeight()
 		if err != nil {
-			return errors.Wrapf(err, "can't get signed header after height %d", newHeader.Height)
+			return err
+		}
+		var closestHeader *types.SignedHeader
+		if newHeader.Height < firstHeaderHeight {
+			c.logger.Info("Using backwards")
```
melekes: debug?

lite2/client.go
melekes: should we actually check if closestHeader can be trusted before attempting `bisection`?

cmwaters: if it's in the trustedStore then we don't we assumed that is is trusted?

cmwaters: if bisection fails it goes to sequential verification automatically which is much the same algorithm as backwards (although backwards is considerably faster)

melekes: not if we're doing bisection? zarko: "That’s fine if trusted header with bisection is within trusted period."

lite2/store/db/db.go
cmwaters: I'm not too sure why because I am not that familiar with the db stuff but when I tried with height -1 it failed the test, so I guess the reverse iterator upper bound does not actually include it

lite2/store/db/db.go
cmwaters: Can you explain what a stop key is and why I need it, I thought using nil was sufficient when I read the comments regarding the reverse iterator

lite2/store/db/db.go
erikgrinaker: Correct. The bounds are `[start, end)`, i.e. start is inclusive and end is exclusive. This is regardless of the iterator direction. Perhaps not very intuitive, but that's how it is today.

lite2/client.go
cmwaters: Is the error itself returned from calling the FirstTrustedHeight function not sufficient enough

lite2/client.go
cmwaters: That's what the name was beforehand, I didn't bother to change it

lite2/client.go
cmwaters: true, I will change that

lite2/store/db/db.go
melekes: suppose you have keys

Y:100
H:100
H:200

...
H300

in your DB

in the absence of H:100, it will scan Y:100. It's ok since we won't include it. It's just extra work we don't really need to do IF WE SET END KEY

lite2/client.go
melekes: a) we panic there, which we should change
b) no, it's not enough since err will be smth like "can't create iterator"

lite2/client.go
```diff
@@ -564,26 +564,32 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
 		var firstHeaderHeight int64
 		firstHeaderHeight, err = c.FirstTrustedHeight()
 		if err != nil {
-			return err
+			return errors.Wrapf(err, "can't get sig")
```
melekes: what's sig stands for?

lite2/client.go
```diff
@@ -564,26 +564,32 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
...
 		}
 		var closestHeader *types.SignedHeader
 		if newHeader.Height < firstHeaderHeight {
-			c.logger.Info("Using backwards")
 			closestHeader, err = c.TrustedHeader(firstHeaderHeight)
 			if err != nil {
 				return errors.Wrapf(err, "can't get signed header after height %d", newHeader.Height)
 			}
+			if HeaderExpired(closestHeader, c.trustingPeriod, now) {
```
melekes: don't we check this inside backwards method?

lite2/client.go
```diff
@@ -561,15 +561,38 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
...
+			}
+			err = c.backwards(closestHeader, newHeader, now)
+		} else {
+			closestHeader, err = c.trustedStore.SignedHeaderBefore(newHeader.Height)
+			if err != nil {
+				return errors.Wrapf(err, "can't get signed header before height %d", newHeader.Height)
+			}
+			var closestValidatorSet *types.ValidatorSet
+			if HeaderExpired(closestHeader, c.trustingPeriod, now) {
+				closestHeader = c.latestTrustedHeader
```
melekes: not sure this will work... 

say firstHeaderHeight = 10
newHeaderHeight = 20
lastTrustedHeader = 30

if firstHeaderHeight expired, we'll try to bisect from 30 to 20.

melekes: complexity creates errors

cmwaters: right. I'm just going to use this:
```
if HeaderExpired(closestHeader, c.trustingPeriod, now) {
	err = c.backwards(c.latestTrustedHeader, newHeader, now)
}
```

lite2/client.go
```diff
@@ -561,15 +561,37 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
...
+			}
+			err = c.backwards(closestHeader, newHeader, now)
+		} else {
+			closestHeader, err = c.trustedStore.SignedHeaderBefore(newHeader.Height)
+			if err != nil {
+				return errors.Wrapf(err, "can't get signed header before height %d", newHeader.Height)
+			}
+			var closestValidatorSet *types.ValidatorSet
+			if HeaderExpired(closestHeader, c.trustingPeriod, now) {
+				err = c.backwards(c.latestTrustedHeader, newHeader, now)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

cmwaters:
I need to write a few more tests

lite2/client.go
```diff
@@ -561,15 +561,37 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
...
-		closestHeader, err = c.trustedStore.SignedHeaderAfter(newHeader.Height)
+		var firstHeaderHeight int64
+		firstHeaderHeight, err = c.FirstTrustedHeight()
 		if err != nil {
-			return errors.Wrapf(err, "can't get signed header after height %d", newHeader.Height)
+			return errors.Wrapf(err, "can't get first header height for verification")
+		}
+		var closestHeader *types.SignedHeader
+		if newHeader.Height < firstHeaderHeight {
+			closestHeader, err = c.TrustedHeader(firstHeaderHeight)
```
```suggestion
			firstHeader, err = c.TrustedHeader(firstHeaderHeight)
```

lite2/client.go
```diff
@@ -561,15 +561,37 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
...
+			if err != nil {
+				return errors.Wrapf(err, "can't get signed header after height %d", newHeader.Height)
```
```suggestion
				return errors.Wrapf(err, "can't get first header #%d", newHeader.Height)
```

lite2/client.go
```diff
@@ -561,15 +561,37 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
...
 	} else {
-		// 2) Otherwise, perform backwards verification
-		// Find the closest trusted header after newHeader.Height
-		var closestHeader *types.SignedHeader
-		closestHeader, err = c.trustedStore.SignedHeaderAfter(newHeader.Height)
+		var firstHeaderHeight int64
+		firstHeaderHeight, err = c.FirstTrustedHeight()
 		if err != nil {
-			return errors.Wrapf(err, "can't get signed header after height %d", newHeader.Height)
+			return errors.Wrapf(err, "can't get first header height for verification")
```
```suggestion
			return errors.Wrap(err, "can't get first header height")
```

lite2/client.go
```diff
@@ -561,15 +561,37 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
...
+		var closestHeader *types.SignedHeader
+		if newHeader.Height < firstHeaderHeight {
+			closestHeader, err = c.TrustedHeader(firstHeaderHeight)
+			if err != nil {
+				return errors.Wrapf(err, "can't get signed header after height %d", newHeader.Height)
+			}
+			if HeaderExpired(closestHeader, c.trustingPeriod, now) {
+				closestHeader = c.latestTrustedHeader
+			}
+			err = c.backwards(closestHeader, newHeader, now)
```
melekes: so we execute backwards from first header (height: 1 for example) to newHeader (height: 4 for example). how's that possible?

cmwaters: This only runs if new header height is less than first header height

lite2/client.go
cmwaters: shall I also start using fmt.Errorf

lite2/client.go
cmwaters: I just used closest header because it was already declared and it also is technically the closest header to the new header

lite2/client.go
cmwaters: Yes you're right I did see it before but forgot to change it

lite2/client.go
melekes: sure

lite2/client.go
melekes: Ah, okay

cmwaters:
I'm also going to change it such that if you choose sequential verification that you will use backwards instead of bisection

lite2/client.go
```diff
@@ -322,10 +320,10 @@ func (c *Client) checkTrustedHeaderUsingOptions(options TrustOptions) error {
 		if c.confirmationFn(action) {
 			err := c.Cleanup()
 			if err != nil {
-				return errors.Wrap(err, "failed to cleanup")
+				return fmt.Errorf("failed to cleanup: %w", err)
 			}
 		} else {
-			return errors.New("refused to remove the stored headers despite hashes mismatch")
+			return fmt.Errorf("refused to remove the stored headers despite hashes mismatch")
```
melekes: use errors.New when there are no args. It's written in STYLE_GUIDE I believe

lite2/client.go
```diff
@@ -436,19 +434,19 @@ func (c *Client) TrustedValidatorSet(height int64) (valSet *types.ValidatorSet,
 func (c *Client) compareWithLatestHeight(height int64) (int64, error) {
 	latestHeight, err := c.LastTrustedHeight()
 	if err != nil {
-		return 0, errors.Wrap(err, "can't get last trusted height")
+		return 0, fmt.Errorf("can't get last trusted height: %w", err)
 	}
 	if latestHeight == -1 {
-		return 0, errors.New("no headers exist")
+		return 0, fmt.Errorf("no headers exist")
```
melekes: same

lite2/client.go
```diff
@@ -436,19 +434,19 @@ func (c *Client) TrustedValidatorSet(height int64) (valSet *types.ValidatorSet,
...
 
 	switch {
 	case height > latestHeight:
-		return 0, errors.Errorf("unverified header/valset requested (latest: %d)", latestHeight)
+		return 0, fmt.Errorf("unverified header/valset requested (latest: %d)", latestHeight)
 	case height == 0:
 		return latestHeight, nil
 	case height < 0:
-		return 0, errors.New("negative height")
+		return 0, fmt.Errorf("negative height")
```
melekes: same

lite2/client.go
```diff
@@ -464,7 +462,7 @@ func (c *Client) compareWithLatestHeight(height int64) (int64, error) {
 // primary.
 func (c *Client) VerifyHeaderAtHeight(height int64, now time.Time) (*types.SignedHeader, error) {
 	if height <= 0 {
-		return nil, errors.New("negative or zero height")
+		return nil, fmt.Errorf("negative or zero height")
```
melekes: same

lite2/client.go
```diff
@@ -509,15 +507,15 @@ func (c *Client) VerifyHeaderAtHeight(height int64, now time.Time) (*types.Signe
 // provider.ErrValidatorSetNotFound error is returned.
 func (c *Client) VerifyHeader(newHeader *types.SignedHeader, newVals *types.ValidatorSet, now time.Time) error {
 	if newHeader.Height <= 0 {
-		return errors.New("negative or zero height")
+		return fmt.Errorf("negative or zero height")
```
melekes: ditto

lite2/client.go
```diff
@@ -604,10 +624,10 @@ func (c *Client) sequence(
 		err = VerifyAdjacent(c.chainID, trustedHeader, interimHeader, interimVals,
 			c.trustingPeriod, now)
 		if err != nil {
-			err = errors.Wrapf(err, "verify adjacent from #%d to #%d failed",
-				trustedHeader.Height, interimHeader.Height)
+			err = fmt.Errorf("verify adjacent from #%d to #%d failed: %w",
+				trustedHeader.Height, interimHeader.Height, err)
 
-			switch errors.Cause(err).(type) {
+			switch (err).(type) {
```
```suggestion
			switch errors.Unwrap(err).(type) {
```

lite2/client.go
melekes: So are you going to change this or not?

lite2/client.go
```diff
@@ -930,7 +955,7 @@ func (c *Client) compareNewHeaderWithWitnesses(h *types.SignedHeader) error {
 		time.Sleep(backoffTimeout(attempt))
 	}
 
-	return errors.New("awaiting response from all witnesses exceeded dropout time")
+	return fmt.Errorf("awaiting response from all witnesses exceeded dropout time")
```
melekes: ditto

lite2/client.go
cmwaters: I can change that but can you tell me what the difference between errors.Unwrap, errors.Cause and err.(type) which I have seen all used in the same way

lite2/client.go
```diff
@@ -3,12 +3,11 @@ package lite
 import (
 	"bytes"
 	"fmt"
+	"github.com/pkg/errors"
```
marbar3778: could you run`make format`

cmwaters:
```
callum@CallumsPC:~/go/src/github.com/tendermint/tendermint$ make format
find . -name '*.go' -type f -not -path "*.git*" -not -name '*.pb.go' -not -name '*pb_test.go' | xargs gofmt -w -s
find . -name '*.go' -type f -not -path "*.git*"  -not -name '*.pb.go' -not -name '*pb_test.go' | xargs goimports -w -local github.com/tendermint/tendermint
xargs: goimports: No such file or directory
Makefile:139: recipe for target 'format' failed
make: *** [format] Error 127
```

marbar3778: odd, do you have goimports installed locally?

lite2/client.go
melekes: Cause - https://github.com/pkg/errors#retrieving-the-cause-of-an-error
Unwrap is the same ^

melekes:
Please refrain from rebasing & force-pushing next time before the PR is approved. When/if you do a `git rebase` and force-push, we can't review only new changes and forced to review everything. 

Instead, use `git merge develop --no-ff` to merge upstream changes.

Thank you 🙇

melekes:
https://github.com/tendermint/tendermint/pull/4575/files#r394331618 still needs to be fixed I believe

lite2/store/db/db.go
cmwaters: Am I supposed to replace nil with append(s.shKey(1<<63-1), byte(0x00))

melekes: s.shKey(1)

cmwaters: okay thanks - sorry I'm just not so familiar with the db

lite2/client.go
```diff
@@ -544,15 +544,37 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
...
-		var closestHeader *types.SignedHeader
-		closestHeader, err = c.trustedStore.SignedHeaderAfter(newHeader.Height)
+		var firstHeaderHeight int64
+		firstHeaderHeight, err = c.FirstTrustedHeight()
 		if err != nil {
-			return errors.Wrapf(err, "can't get signed header after height %d", newHeader.Height)
+			return fmt.Errorf("can't get first header height: %w", err)
+		}
+		var closestHeader *types.SignedHeader
+		if newHeader.Height < firstHeaderHeight {
```
melekes: can we add a short comment here explaining the logic?

lite2/client.go
```diff
@@ -544,15 +544,37 @@ func (c *Client) verifyHeader(newHeader *types.SignedHeader, newVals *types.Vali
...
+			closestHeader, err = c.TrustedHeader(firstHeaderHeight)
+			if err != nil {
+				return fmt.Errorf("can't get first signed header: %w", err)
+			}
+			if HeaderExpired(closestHeader, c.trustingPeriod, now) {
+				closestHeader = c.latestTrustedHeader
+			}
+			err = c.backwards(closestHeader, newHeader, now)
+		} else {
```
melekes: can we add a short comment here explaining the logic?

lite2/errors.go
```diff
@@ -38,3 +38,9 @@ type ErrInvalidHeader struct {
 func (e ErrInvalidHeader) Error() string {
 	return fmt.Sprintf("invalid header: %v", e.Reason)
 }
+
+type ErrNoWitnesses struct{}
```
melekes: godoc please

melekes: also, what is the reason behind exposing this error?

lite2/client.go
```diff
@@ -1004,7 +1031,7 @@ func (c *Client) signedHeaderFromPrimary(height int64) (*types.SignedHeader, err
 		if err == nil {
 			// sanity check
 			if height > 0 && h.Height != height {
-				return nil, errors.Errorf("expected %d height, got %d", height, h.Height)
+				return nil, fmt.Errorf("expected %d height, got %d: %w", height, h.Height, err)
```
melekes: err equals nil here, no? why log it then?

lite2/client.go
cmwaters: sure

lite2/errors.go
cmwaters: My idea was that I wanted to standardize some of the errors. I only chose to have it exposed because the rest of the errors were also exposed

lite2/client.go
cmwaters: you're right - I can remove it

lite2/errors.go
melekes: > My idea was that I wanted to standardize some of the errors.

Can this be done in a separate PR? It's not a good idea to mix a lot of issues in one PR

melekes: https://github.com/tendermint/tendermint/issues/4538

Issue #4569, opened 2020-03-16
Add Docker image default configuration


greg-szabo:
## Summary
Add a default configuration to the tendermint docker image. Currently, the image does not have any configuration 'baked in'.

## Problem Definition
Currently, the tendermint docker image requires the end-user to run the `init` command with a mounted volume and then in a second command, it can run `node` if that mounted volume is attached. (Details here: https://hub.docker.com/r/tendermint/tendermint under the "How to use this image" section).

In some cases the user cannot mount a volume to the image and regardless, it is cumbersome to do it. The most recent use-case where it make the image unusable is GitHub Actions. GitHub Actions has a 'Services' component that can run docker images as backend services while running CI (for example integration testing). Examples include nginx, redis and postgres. The common point in all those examples is that the service starts with one command. The Services component does allow users to mount a volume, but the service is set up before the repository gets cloned, hence it becomes a chicken-and-egg problem.
I would like to use the Tendermint docker image as a service in some GitHub Actions.

Possible disadvantage of the feature is that end-users with more complex configuration need to be aware that they need to override the default configuration (by mounting a volume). This can be clarified in the documentation mentioned above.

## Proposal
During the docker image deployment, `tendermint init` should run and create a default configuration in the home folder. This can be achieved by modifying the `Dockerfile`.

Possibly the default ABCI application should be kvstore, not a TCP socket.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

tarcieri:
It'd also be nice to do this for the KMS, which has also moved over to GitHub actions:

https://github.com/iqlusioninc/tmkms/issues/15

PR #4577, opened 2020-03-16
DOCKER: update with defaults

greg-szabo:
Closes: #4569 

## Description
This change updates the `Dockerfile` to include defaults for a one-node `kvstore` testnet. It can be overwritten by adding a volume to mount point `/tendermint` as it was available before.
This change is a non-breaking change.
This change allows a developer to run `docker run tendermint/tendermint` and have a locally available one-node testnet at once (without running `init` beforehand).

Note: The `VOLUME` command was moved to the bottom of the file. The reason for this is that any files added to the mount point after this command ran would get lost during the build process. (The files have to be added before this command.)
______

For contributor use:

- [ ] Wrote tests - tested docker build on the local machine, no Go code change was introduced
- [X] Updated CHANGELOG_PENDING.md
- [X] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Updated relevant documentation (`docs/`) and code comments - no code comments or new Go code was introduced
- [X] Re-reviewed `Files changed` in the Github PR explorer

### DIFF
--- a/DOCKER/Dockerfile
+++ b/DOCKER/Dockerfile
@@ -22,17 +22,32 @@ RUN apk update && \
 # Run the container with tmuser by default. (UID=100, GID=1000)
 USER tmuser
 
-# Expose the data directory as a volume since there's mutable state in there
-VOLUME [ $TMHOME ]
-
 WORKDIR $TMHOME
 
 # p2p and rpc port
-EXPOSE 26656 26657
+EXPOSE 26656 26657 26660
 
 ENTRYPOINT ["/usr/bin/tendermint"]
-CMD ["node", "--moniker=`hostname`"]
+CMD ["node"]
 STOPSIGNAL SIGTERM
 
 ARG BINARY=tendermint
 COPY $BINARY /usr/bin/tendermint
+
+# Create default configuration for docker run.
+RUN /usr/bin/tendermint init && \
+    sed -i \
+      -e 's/^proxy_app\s*=.*/proxy_app = "kvstore"/' \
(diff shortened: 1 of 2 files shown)


DOCKER/Dockerfile
```diff
@@ -22,17 +22,32 @@ RUN apk update && \
...
 ARG BINARY=tendermint
 COPY $BINARY /usr/bin/tendermint
+
+# Create default configuration for docker run.
+RUN /usr/bin/tendermint init && \
+    sed -i \
+      -e 's/^proxy_app\s*=.*/proxy_app = "kvstore"/' \
+      -e 's/^moniker\s*=.*/moniker = "dockernode"/' \
+      -e 's/^addr_book_strict\s*=.*/addr_book_strict = false/' \
+      -e 's/^timeout_commit\s*=.*/timeout_commit = "500ms"/' \
```
melekes: why default got changed here?

greg-szabo: Good point. The defaults are good enough, but they are optimized for multi-node networks. In this small single-node example, it makes sense to make the network somewhat faster. I was contemplating turning on `skip_timeout_commit` since there's no one else to wait for, but it made the network so fast (tens of blocks per second) that it would've been harder for developers to see what's happening. Changing the `timeout_commit` makes it look cooler during development. :)

In a current use-case, we have a test in the tendermint-rs repository that checks if at least 10 blocks were created. Essentially, the test needs to sleep until 10 blocks were created. Normally, this would take 10 seconds, but with the modification, it's only 5.

Issue #2684, opened 2018-10-21
Integers in Tendermint


ebuchman:
Some time ago we resolved to use `int64` for Height, primarily because signed ints are better when they might be used for arithmetic (https://blog.cosmos.network/choosing-a-type-for-blockchain-height-beware-of-unsigned-integers-714804dddf1d).

That said, recent events brought to light some issues with integers in the Tendermint codebase:

## Amino

Ref https://github.com/tendermint/tendermint/issues/2682

Amino uses zig-zag encoding for `int64` in Go, which matches `sint64` in proto3, but not `int64`. Proto3 treats `int64` and `uint64` the same for positive values, but differs for `sint64`.

Even if we stick with signed values for fields that should never be negative, we probably don't want them zig-zag encoded (the whole point of zig-zag is more efficient negative numbers!)

If we fix this, we should be able to change the type to `uint64` later without breaking the encoding

## Input Validation

Ref https://github.com/tendermint/tendermint/issues/2683

Many messages contain vanilla `int` and don't perform any validation before handling, hence negative values can crash the program. Practically all the input validation that should happen is checking for negative numbers. All that code (yet to be written) would go away if we used `uint`.

Since moving to uint may take some time, we should meanwhile validate all our input. If we throw it away once switching to uint, so be it.

## Consistency

We make mixed use of `int` and `int64` throughout - eg. `Height.Height` and `Header.NumTxs` are `int64`, but `Header.BlockID.PartSetHeader.Total` is `int`. We also use `int` for `Round`.

We should probably be consistent, or at least specify the bit-size and have good reason why we go with 8/16/32/64 each time. 

## Summary

Arguably, the reason to stick with signed ints is because they are more natural to do arithmetic with.
First, you can't avoid being given `int` in Go, as it's the default value for integers (eg. `x := 5`, `len(values)`,  and `for i, v := range values {`). This means if you use anything else you have to cast to it everywhere (but note if we address the consistency issue, we'll have to do that anyways).

Second, even if some value should never be negative, you may still need to subtract two such values, and the difference could be negative. If you accidentally do this with `uint` you get a massive number, and it will be difficult to find out what went wrong. This can be worked around using [strict linters to avoid naive arithmetic](https://github.com/tendermint/tendermint/pull/2681#discussion_r226839084), but that makes the code a lot clunkier, though arguably much safer.

[...]

PR #4582, opened 2020-03-18
stricter ints for round and valIndex

marbar3778:
## Description

- this is to help reduce the overall size of some proto prs.
- round `int` types to `int32` (consensus)
- valIndex `int` to `int32`

ref #2684  

______

For contributor use:

- [ ] Wrote tests


privval/file_deprecated.go
```diff
@@ -58,7 +58,7 @@ func (oldFilePV *OldFilePV) Upgrade(keyFilePath, stateFilePath string) *FilePV {
 
 	pvState := FilePVLastSignState{
 		Height:    oldFilePV.LastHeight,
-		Round:     oldFilePV.LastRound,
+		Round:     int(oldFilePV.LastRound),
```
golangcibot: unnecessary conversion (from `unconvert`)

consensus/reactor.go
```diff
@@ -423,7 +423,7 @@ func (conR *Reactor) broadcastHasVoteMessage(vote *types.Vote) {
 		Height: vote.Height,
 		Round:  vote.Round,
 		Type:   vote.Type,
-		Index:  vote.ValidatorIndex,
+		Index:  int32(vote.ValidatorIndex),
```
golangcibot: unnecessary conversion (from `unconvert`)

consensus/types/round_state.go
```diff
@@ -135,11 +135,11 @@ func (rs *RoundState) NewRoundEvent() types.EventDataNewRound {
 
 	return types.EventDataNewRound{
 		Height: rs.Height,
-		Round:  rs.Round,
+		Round:  int(rs.Round),
```
marbar3778: this is needed since we in the past encoded and decoded ints as strings, if we were to change this to `1` instead of `"1"` then I would be able to remove these conversions

melekes:
round should be uint32 (or uint16 even)

melekes:
valIndex - uint16

marbar3778:
perfect will make the changes to `uint32` for now Id like to stay at 32 because in proto the smallest size is 32. In the future, I will test smaller sizes (8,16), but this will mean going back and forth from handwritten types and generated.

consensus/types/round_state.go
marbar3778: undid this in favor of having round represented not as a string in json

types/proposal.go
```diff
@@ -54,9 +55,15 @@ func (p *Proposal) ValidateBasic() error {
 	if p.Round < 0 {
 		return errors.New("negative Round")
 	}
+	if p.Round > math.MaxInt32 {
```
melekes: these checks do not make sense. p.Round won't ever be greater than MaxInt32.

What would be good is to add a check: when we increment Round/POLRound and Round == math.MaxInt32, we panic.

config/toml.go
```diff
@@ -473,8 +473,8 @@ var testPrivValidatorKey = `{
   }
 }`
 
-var testPrivValidatorState = `{
+var testPrivValidatorState = `{
```
tessr:
```suggestion
var testPrivValidatorState = `{
```

libs/math/safemath.go
```diff
@@ -0,0 +1,26 @@
+package math
+
+import (
+	"errors"
+	"math"
+)
+
+var ErrOverflowInt = errors.New("integer overflow")
+
+func SafeAddInt32(a, b int32) (int32, error) {
```
tessr: godoc, please!

(also: did these get copied from somewhere?)

privval/file_deprecated.go
```diff
@@ -58,7 +58,7 @@ func (oldFilePV *OldFilePV) Upgrade(keyFilePath, stateFilePath string) *FilePV {
 
 	pvState := FilePVLastSignState{
 		Height:    oldFilePV.LastHeight,
-		Round:     oldFilePV.LastRound,
+		Round:     int32(oldFilePV.LastRound),
```
tessr: What was this before? Do we know this is safe?

marbar3778: this was an `int` before. I didn't want to change the `oldFilePV` to abide by the new type.

libs/math/safemath.go
marbar3778: adjusted the ones we have for int64 in validator_set.go. will be moving those here in the future. but didn't want to this PR bigger than what it is now

types/validator_set_test.go
```diff
@@ -283,12 +286,13 @@ func TestProposerSelection3(t *testing.T) {
...
 		// times is usually 1
-		times := 1
+		times := int32(1)
 		mod := (tmrand.Int() % 5) + 1
 		if tmrand.Int()%mod > 0 {
 			// sometimes its up to 5
-			times = (tmrand.Int() % 4) + 1
+			times = (tmrand.Int31() % 4) + 1
 		}
+		fmt.Println(times)
```
melekes: ??

```suggestion
		t.Log(times)
```

privval/file_test.go
```diff
@@ -300,7 +300,7 @@ func TestDifferByTimestamp(t *testing.T) {
 	}
 }
 
-func newVote(addr types.Address, idx int, height int64, round int, typ byte, blockID types.BlockID) *types.Vote {
+func newVote(addr types.Address, idx, round int32, height int64, typ byte, blockID types.BlockID) *types.Vote {
```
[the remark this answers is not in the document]
marbar3778: to keep same types together. I can undo if its an issue

consensus/state.go
```diff
@@ -452,7 +454,10 @@ func (cs *State) updateHeight(height int64) {
 	cs.Height = height
 }
 
-func (cs *State) updateRoundStep(round int, step cstypes.RoundStepType) {
+func (cs *State) updateRoundStep(round int32, step cstypes.RoundStepType) {
+	if round > math.MaxInt32 {
```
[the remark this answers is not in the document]
marbar3778: line 854 takes care of the safeaddint32, removed this line

state/store.go
```diff
@@ -211,7 +211,7 @@ func LoadValidators(db dbm.DB, height int64) (*types.ValidatorSet, error) {
 				),
 			)
 		}
-		valInfo2.ValidatorSet.IncrementProposerPriority(int(height - lastStoredHeight)) // mutate
+		valInfo2.ValidatorSet.IncrementProposerPriority(int32(height - lastStoredHeight)) // mutate
```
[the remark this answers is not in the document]
marbar3778: added a func for safe conversion to int32

privval/file_test.go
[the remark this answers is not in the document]
melekes: please do

state/store.go
```diff
@@ -211,7 +211,7 @@ func LoadValidators(db dbm.DB, height int64) (*types.ValidatorSet, error) {
 				),
 			)
 		}
-		valInfo2.ValidatorSet.IncrementProposerPriority(tmmath.SafeConvertInt32(height - lastStoredHeight)) // mutate
+		valInfo2.ValidatorSet.IncrementProposerPriority(tmmath.SafeConvertInt32(int(height - lastStoredHeight))) // mutate
```
melekes: `int(height-lastStoredHeight)` is not safe on 32 bit arch

melekes: I guess it's fine since we don't support 32-bit anyway

marbar3778: since we are only using this here, it would be safer to have the fun accept a int64. will make the change then merge

PR #4585, opened 2020-03-20
blockchain/v1: remove unused version

tessr:
Blockchain reactor v1 was a big improvement from v0 and a critical stepping stone on the way to v2, but it seems like it's not been widely used. Per the conversation on yesterday's dev call, we'd like to delete it. 

______

For contributor use:

~- [ ] Wrote tests~
~- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.~
~- [ ] Re-reviewed `Files changed` in the Github PR explorer~


node/node.go
```diff
@@ -363,8 +362,7 @@ func createBlockchainReactor(config *cfg.Config,
 	switch config.FastSync.Version {
 	case "v0":
 		bcReactor = bcv0.NewBlockchainReactor(state.Copy(), blockExec, blockStore, fastSync)
-	case "v1":
-		bcReactor = bcv1.NewBlockchainReactor(state.Copy(), blockExec, blockStore, fastSync)
+    // TODO(tessr): allow configuring blockchain reactor v2 f
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
		// TODO(tessr): allow configuring blockchain reactor v2 f
```

marbar3778:
should this also be removed? https://github.com/tendermint/spec/blob/master/spec/reactors/block_sync/bcv1/impl-v1.md

erikgrinaker:
Given that `v1` appears to be in use, should we keep it for now and deprecate `v0` and `v1` when we make `v2` the default?

marbar3778:
we should add a deprecation warning to the doc.go if we are keeping it.

erikgrinaker:
Closing, since it's in use. Let's start moving towards `v2` as the default: https://github.com/tendermint/tendermint/issues/4595.

PR #4588, opened 2020-03-20
add support for block pruning via ABCI Commit response

### LINKED ISSUE #3652: Prune blockchain history
#OperationSaveHarddrives

At some point we'd like full nodes to be able to prune old blocks to free up hard drive space. Of course doing so would mean new nodes can't sync the chain from before the last pruning point, and users won't be able to query for txs from before this time. The former is solved by state sync (ie. #828), and the later by the good people who operate archive nodes. 

Aside from actually implementing state sync, we need to expose some way to responsibly prune the history. While this should work together with state sync, one could imagine other ways to sync state out-of-band, so that this feature could be used even independent of an in-protocol state-sync reactor.

Related #2543

erikgrinaker:
Fixes #3652, as discussed in [RFC-001](https://github.com/tendermint/spec/blob/master/rfc/001-block-retention.md).

## Description

Adds support for truncating the block store history (required for [state sync](https://github.com/tendermint/tendermint/issues/828)):

* Adds `BlockStore.PruneBlocks()` to prune the block store up to a given height, and `BlockStore.Base()` to return the first available height. All blocks between the base and height (inclusive) are guaranteed to be present.

* Adds `state.Store.PruneStates()` to prune the state database of validator sets, consensus parameters, and ABCI responses in a height range.

* Adds an ABCI field `ResponseCommit.retain_height` which allows the ABCI app to control block pruning. All heights below `retain_height` will be synchronously removed from the block store and state database after applying a block during consensus (but not during replay).

* Adds info about the earliest available block to the RPC endpoint `/status`.

* Blockchain reactor status messages now contain the block store base as well, to avoid requesting blocks from nodes that do not have them. This is a backwards-compatible change.

* If the blockchain or consensus reactors are not able to catch up because all nodes have pruned blockstores, the node halts.

I believe this requires updating the spec as well. I would appreciate a thorough review of the consensus reactor changes.

Once this is merged, and until we get state sync, it is important that some nodes in the network keep full block histories to allow new nodes to bootstrap.

______

For contributor use:

- [x] Wrote tests


config/config_test.go
```diff
@@ -134,27 +134,45 @@ func TestFastSyncConfigValidateBasic(t *testing.T) {
...
+		"TimeoutPrevoteDelta negative":         {func(c *ConsensusConfig) { c.TimeoutPrevoteDelta = -1 }, true},
+		"TimeoutPrecommit":                     {func(c *ConsensusConfig) { c.TimeoutPrecommit = time.Second }, false},
+		"TimeoutPrecommit negative":            {func(c *ConsensusConfig) { c.TimeoutPrecommit = -1 }, true},
+		"TimeoutPrecommitDelta":                {func(c *ConsensusConfig) { c.TimeoutPrecommitDelta = time.Second }, false},
+		"TimeoutPrecommitDelta negative":       {func(c *ConsensusConfig) { c.TimeoutPrecommitDelta = -1 }, true},
+		"TimeoutCommit":                        {func(c *ConsensusConfig) { c.TimeoutCommit = time.Second }, false},
+		"TimeoutCommit negative":               {func(c *ConsensusConfig) { c.TimeoutCommit = -1 }, true},
+		"PeerGossipSleepDuration":              {func(c *ConsensusConfig) { c.PeerGossipSleepDuration = time.Second }, false},
+		"PeerGossipSleepDuration negative":     {func(c *ConsensusConfig) { c.PeerGossipSleepDuration = -1 }, true},
+		"PeerQueryMaj23SleepDuration":          {func(c *ConsensusConfig) { c.PeerQueryMaj23SleepDuration = time.Second }, false},
```
golangcibot: line is 124 characters (from `lll`)

config/config_test.go
```diff
@@ -134,27 +134,45 @@ func TestFastSyncConfigValidateBasic(t *testing.T) {
...
 	}
-
-	for _, fieldName := range fieldsToTest {
-		reflect.ValueOf(cfg).Elem().FieldByName(fieldName).SetInt(-1)
-		assert.Error(t, cfg.ValidateBasic())
-		reflect.ValueOf(cfg).Elem().FieldByName(fieldName).SetInt(0)
+	for desc, tc := range testcases {
+		t.Run(desc, func(t *testing.T) {
+			cfg := DefaultConsensusConfig()
+			tc.modify(cfg)
```
golangcibot: Using the variable on range scope `tc` in function literal (from `scopelint`)

config/config_test.go
```diff
@@ -134,27 +134,45 @@ func TestFastSyncConfigValidateBasic(t *testing.T) {
...
+
+			err := cfg.ValidateBasic()
+			if tc.expectErr {
```
golangcibot: Using the variable on range scope `tc` in function literal (from `scopelint`)

store/store.go
```diff
@@ -171,6 +183,74 @@ func (bs *BlockStore) LoadSeenCommit(height int64) *types.Commit {
 	return commit
 }
 
+// PruneBlocks removes block up to (but not including) a height. It returns number of blocks pruned.
+func (bs *BlockStore) PruneBlocks(height int64) (uint64, error) {
+	bs.mtx.RLock()
+	if height > bs.height {
+		bs.mtx.RUnlock()
+		return 0, fmt.Errorf("Cannot prune the latest height %v", bs.height)
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

store/store.go
```diff
@@ -171,6 +183,74 @@ func (bs *BlockStore) LoadSeenCommit(height int64) *types.Commit {
...
+	}
+	base := bs.base
+	bs.mtx.RUnlock()
+	if base == 0 {
+		base = 1
+	}
+	if base > height {
+		return 0, fmt.Errorf("Cannot prune to height %v, it is lower than base height %v",
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

rpc/core/status.go
```diff
@@ -16,6 +16,23 @@ import (
...
 // More: https://docs.tendermint.com/master/rpc/#/Info/status
 func Status(ctx *rpctypes.Context) (*ctypes.ResultStatus, error) {
+	var (
+		earliestBlockMeta     *types.BlockMeta
+		earliestBlockHash     tmbytes.HexBytes
+		earliestAppHash       tmbytes.HexBytes
+		earliestBlockTimeNano int64
+	)
+	earliestBlockHeight := blockStore.Base()
+	if earliestBlockHeight == 0 {
```
melekes: can't we make this a default in blockStore, so we don't have to write such conditionals

state/errors.go
```diff
@@ -61,6 +66,11 @@ func (e ErrBlockHashMismatch) Error() string {
 func (e ErrAppBlockHeightTooHigh) Error() string {
 	return fmt.Sprintf("App block height (%d) is higher than core (%d)", e.AppHeight, e.CoreHeight)
 }
+
+func (e ErrAppBlockHeightBelowBase) Error() string {
+	return fmt.Sprintf("App block height (%d) is below block store base (%d)", e.AppHeight, e.StoreBase)
```
melekes:
```suggestion
	return fmt.Sprintf("app block height (%d) is below block store base (%d)", e.AppHeight, e.StoreBase)
```

state/errors.go
```diff
@@ -61,6 +66,11 @@ func (e ErrBlockHashMismatch) Error() string {
...
+}
+
 func (e ErrLastStateMismatch) Error() string {
 	return fmt.Sprintf(
 		"Latest tendermint block (%d) LastAppHash (%X) does not match app's AppHash (%X)",
```
melekes:
```suggestion
		"latest tendermint block (%d) LastAppHash (%X) does not match app's AppHash (%X)",
```

state/errors.go
```diff
@@ -61,6 +66,11 @@ func (e ErrBlockHashMismatch) Error() string {
 func (e ErrAppBlockHeightTooHigh) Error() string {
 	return fmt.Sprintf("App block height (%d) is higher than core (%d)", e.AppHeight, e.CoreHeight)
```
melekes:
```suggestion
	return fmt.Sprintf("app block height (%d) is higher than core (%d)", e.AppHeight, e.CoreHeight)
```

store/store.go
```diff
@@ -171,6 +183,74 @@ func (bs *BlockStore) LoadSeenCommit(height int64) *types.Commit {
...
+// PruneBlocks removes block up to (but not including) a height. It returns number of blocks pruned.
+func (bs *BlockStore) PruneBlocks(height int64) (uint64, error) {
+	bs.mtx.RLock()
+	if height > bs.height {
+		bs.mtx.RUnlock()
+		return 0, fmt.Errorf("cannot prune the latest height %v", bs.height)
+	}
+	base := bs.base
+	bs.mtx.RUnlock()
+	if base == 0 {
```
melekes: same question

store/store.go
```diff
@@ -171,6 +183,74 @@ func (bs *BlockStore) LoadSeenCommit(height int64) *types.Commit {
...
+		base = 1
+	}
+	if base > height {
```
melekes: that shouldn't even be possible, no? if so, maybe change it to assert (panic)

blockchain/v0/reactor.go
```diff
@@ -196,11 +197,15 @@ func (bcR *BlockchainReactor) Receive(chID byte, src p2p.Peer, msgBytes []byte)
...
+		src.TrySend(BlockchainChannel, cdc.MustMarshalBinaryBare(&bcStatusResponseMessage{
+			Height: bcR.store.Height(),
+			Base:   bcR.store.Base(),
+		}))
 	case *bcStatusResponseMessage:
 		// Got a peer status. Unverified.
-		bcR.pool.SetPeerHeight(src.ID(), msg.Height)
+		bcR.pool.SetPeerRange(src.ID(), msg.Base, msg.Height)
+	case *bcNoBlockResponseMessage:
+		bcR.Logger.Debug("Peer does not have requested block", "peer", src, "height", msg.Height)
```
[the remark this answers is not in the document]
erikgrinaker: Nah, the message only contains the height for the requested block, not the base and height of the peer's block store. What we're logging is the requested height.

blockchain/v2/scheduler.go
```diff
@@ -295,6 +297,11 @@ func (sc *scheduler) setPeerHeight(peerID p2p.ID, height int64) error {
 		return fmt.Errorf("cannot move peer height lower. from %d to %d", peer.height, height)
 	}
 
+	if base > height {
```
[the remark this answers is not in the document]
erikgrinaker: Agreed. Let's keep the assert anyway.

store/store.go
erikgrinaker: It's possible if the block store is empty, where in this case base=1 and height=0.

erikgrinaker: That's up to the caller. If the block store is at base 100 and the caller tries to prune to height 50, I think that should be an error. It generally shouldn't happen, but there could be e.g. race conditions. We should avoid panics in such situations, since it can turn an otherwise minor bug into a crash bug.

CHANGELOG_PENDING.md
```diff
@@ -8,17 +8,28 @@ Friendly reminder, we have a [bug bounty program](https://hackerone.com/tendermi
 
 ### BREAKING CHANGES:
 
+- Nodes are no longer guaranteed to contain all blocks up to the latest height. The block store can now have a truncated history (via `retain_blocks`) such that all blocks up to the block store base will be missing. All blocks between the base and latest height will be present.
```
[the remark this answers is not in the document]
erikgrinaker: Sure, done

blockchain/v0/reactor.go
```diff
@@ -446,34 +454,42 @@ func (m *bcBlockResponseMessage) String() string {
...
 
 // ValidateBasic performs basic validation.
 func (m *bcStatusRequestMessage) ValidateBasic() error {
+	if m.Base < 0 {
+		return errors.New("negative Base")
+	}
 	if m.Height < 0 {
 		return errors.New("negative Height")
 	}
 	return nil
```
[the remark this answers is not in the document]
erikgrinaker: Good call.

rpc/core/status.go
erikgrinaker: Yep, done

erikgrinaker:
I've been thinking that we should prune the state database when we prune the blockstore. It's more symmetric, and I'm not sure if there's any point in keeping it. We will have to handle this case with state sync anyway, since we won't backfill state metadata.

What do you think @melekes?

consensus/state.go
```diff
@@ -1427,7 +1427,7 @@ func (cs *State) finalizeCommit(height int64) {
 	}
 
 	// Prune the blockStore.
-	if cs.config.RetainBlocks > 0 && uint64(cs.blockStore.Height()) > cs.config.RetainBlocks {
+	if cs.config.RetainBlocks > 0 && cs.blockStore.Height()-cs.blockStore.Base()+1 > int64(cs.config.RetainBlocks) {
```
melekes: maybe it will be better to extract cs.blockStore.Height()-cs.blockStore.Base()+1 into a function
```go
(bs *BlockStore) NumRetainedBlocks() int { return bs.Height()-bs.Base()+1 }
```

melekes:
> What do you think @melekes?

We need to store validator sets & committed blocks at least for the duration of unbonding period (2-3 weeks). Network & light client security depends on us being able to punish validators if they misbehave within the unbonding period.

erikgrinaker:
> > What do you think @melekes?
> 
> We need to store validator sets & committed blocks at least for the duration of unbonding period (2-3 weeks). Network & light client security depends on us being able to punish validators if they misbehave within the unbonding period.

Yes, but I think that should be up to node operators to configure, and orthogonal to whether or not we should prune the state database. I think we should prune it, and we can discuss the network configuration issue separately.

consensus/state.go
erikgrinaker: Good idea

erikgrinaker:
We've been discussing how we should configure block retention and snapshotting, there's an RFC on this here: https://github.com/tendermint/spec/pull/84

I suggest we hold off on merging this until the RFC is resolved.

erikgrinaker:
Confirmed that appending struct fields is both forwards- and backwards-compatible with Amino, so this PR should not be protocol breaking in any way (apart from nodes not necessarily having old blocks anymore).

consensus/replay_test.go
```diff
@@ -1104,6 +1125,17 @@ func (bs *mockBlockStore) LoadSeenCommit(height int64) *types.Commit {
 	return bs.commits[height-1]
 }
 
+func (bs *mockBlockStore) PruneBlocks(height int64) (uint64, error) {
+	pruned := uint64(0)
+	for i := int64(0); i < height-1; i++ {
+		bs.chain[i] = nil
+		bs.commits[i] = nil
+		pruned++
```
melekes: aren't pruned always equals to `height-1`?

consensus/state.go
```diff
@@ -878,6 +882,9 @@ func (cs *State) needProofBlock(height int64) bool {
 	}
 
 	lastBlockMeta := cs.blockStore.LoadBlockMeta(height - 1)
+	if lastBlockMeta == nil {
+		panic("needProofBlock: last block meta not found")
```
```suggestion
		panic(fmt.Sprintf("needProofBlock: last block meta for height %d not found", height-1))
```

consensus/replay_test.go
erikgrinaker: I suppose. Normally it will be newBase - oldBase, but the mock generally always starts from the first.

tessr:
🎉 

Can you also make sure your commit message (and possibly also your PR title) conform to the [Go commit style](https://tip.golang.org/doc/contribute.html#commit_messages)?

consensus/reactor.go
```diff
@@ -803,15 +803,17 @@ OUTER_LOOP:
...
-				peer.TrySend(StateChannel, cdc.MustMarshalBinaryBare(&VoteSetMaj23Message{
-					Height:  prs.Height,
-					Round:   commit.Round,
-					Type:    types.PrecommitType,
-					BlockID: commit.BlockID,
-				}))
-				time.Sleep(conR.conS.config.PeerQueryMaj23SleepDuration)
+			if prs.CatchupCommitRound != -1 && prs.Height > 0 && prs.Height <= conR.conS.blockStore.Height() &&
+				prs.Height >= conR.conS.blockStore.Base() {
+				if commit := conR.conS.LoadCommit(prs.Height); commit != nil {
```
ebuchman, line 808: How could commit be nil at this point ?

consensus/state.go
```diff
@@ -1463,6 +1471,16 @@ func (cs *State) finalizeCommit(height int64) {
 
 	fail.Fail() // XXX
 
+	// Prune old heights, if requested by ABCI app.
+	if retainHeight > 0 {
```
ebuchman, line 1475: Why not do all this inside ApplyBlock?

rpc/core/status.go
```diff
@@ -16,6 +16,20 @@ import (
...
 func Status(ctx *rpctypes.Context) (*ctypes.ResultStatus, error) {
+	var (
+		earliestBlockMeta     *types.BlockMeta
+		earliestBlockHash     tmbytes.HexBytes
+		earliestAppHash       tmbytes.HexBytes
+		earliestBlockTimeNano int64
+	)
+	earliestBlockHeight := blockStore.Base()
+	earliestBlockMeta = blockStore.LoadBlockMeta(earliestBlockHeight)
+	if earliestBlockMeta != nil {
```
ebuchman, line 27: How could this be nil? Is it only if pruning races here? If so is it fine to just not print everything? Maybe we should consider having an atomic Base+Load command so we don't have to worry about pruning races between when we call .Base() and .LoadXXX() ?

state/store.go
```diff
@@ -125,6 +125,102 @@ type ABCIResponses struct {
...
+		return fmt.Errorf("validators at height %v not found", to)
+	}
+	paramsInfo := loadConsensusParamsInfo(db, to)
+	if paramsInfo == nil {
+		return fmt.Errorf("consensus params at height %v not found", to)
+	}
+
+	keepVals := make(map[int64]bool)
+	if valInfo.ValidatorSet == nil {
+		keepVals[valInfo.LastHeightChanged] = true
```
ebuchman, line 154: Instead of keeping older states, why not copy the relevant information into a state we're keeping? eg. load the validator set from the older height and persist it with the first not pruned height

state/store.go
```diff
@@ -125,6 +125,102 @@ type ABCIResponses struct {
...
 
+// PruneStates deletes states between the given heights (including from, excluding to). It is not
+// guaranteed to delete all states, since the last checkpointed state and states being pointed to by
+// e.g. `LastHeightChanged` must remain. The state at to must also exist.
+//
+// The from parameter is necessary since we can't do a key scan in a performant way due to the key
+// encoding not preserving ordering: https://github.com/tendermint/tendermint/issues/4567
+// This will cause some old states to be left behind when doing incremental partial prunes,
+// specifically older checkpoints and LastHeightChanged targets.
+func PruneStates(db dbm.DB, from int64, to int64) error {
```
ebuchman, line 136: Should this be deleting txindex storage from these heights too?

consensus/reactor.go
```diff
@@ -494,16 +494,16 @@ OUTER_LOOP:
 			}
 		}
 
-		// If the peer is on a previous height, help catch up.
-		if (0 < prs.Height) && (prs.Height < rs.Height) {
+		// If the peer is on a previous height that we have, help catch up.
+		if (0 < prs.Height) && (prs.Height < rs.Height) && (prs.Height >= conR.conS.blockStore.Base()) {
```
ebuchman, line 498: Is it possible for pruning to race this? So this condition might be valid but by the time we try to execute LoadBlockMeta below we've pruned it? I guess that's what all the nil checks help catch but maybe we want an atomic Base+Load

consensus/reactor.go
```diff
@@ -803,15 +803,17 @@ OUTER_LOOP:
...
+				prs.Height >= conR.conS.blockStore.Base() {
+				if commit := conR.conS.LoadCommit(prs.Height); commit != nil {
+					peer.TrySend(StateChannel, cdc.MustMarshalBinaryBare(&VoteSetMaj23Message{
+						Height:  prs.Height,
+						Round:   commit.Round,
+						Type:    types.PrecommitType,
+						BlockID: commit.BlockID,
+					}))
+					time.Sleep(conR.conS.config.PeerQueryMaj23SleepDuration)
+				}
```
ebuchman, line 816: Don't we need to also update gossipVotesRoutine? ie. `		// If peer is lagging by more than 1, send Commit.`

consensus/reactor.go
```diff
@@ -803,15 +803,17 @@ OUTER_LOOP:
...
-			if prs.CatchupCommitRound != -1 && 0 < prs.Height && prs.Height <= conR.conS.blockStore.Height() {
-				commit := conR.conS.LoadCommit(prs.Height)
-				peer.TrySend(StateChannel, cdc.MustMarshalBinaryBare(&VoteSetMaj23Message{
-					Height:  prs.Height,
-					Round:   commit.Round,
-					Type:    types.PrecommitType,
-					BlockID: commit.BlockID,
-				}))
-				time.Sleep(conR.conS.config.PeerQueryMaj23SleepDuration)
+			if prs.CatchupCommitRound != -1 && prs.Height > 0 && prs.Height <= conR.conS.blockStore.Height() &&
```
ebuchman, line 806: Might be nice to create a `heightInRange` convenience method for the check that `blockStore.Base() <= prs.height <= blockStore.Height()`

ebuchman (commented):
Great work. A few questions in the comments we might want to follow up on.

consensus/reactor.go
ebuchman: Funny timing: https://github.com/tendermint/tendermint/issues/5377

consensus/state.go
erikgrinaker: Block application and block pruning seem orthogonal? I.e. it should be possible to apply a block without pruning anything.

state/store.go
erikgrinaker: It's too expensive - this would require rewriting all of the states that point to the original height, which can be up to 100.000 entries (the checkpoint interval).

consensus/reactor.go
erikgrinaker: This is just defensive programming. `nil` is a valid return value for `LoadCommit`, so it makes sense to handle that condition. I find that it's generally simpler and safer to check rather than remembering which invariants apply in which scenarios.

state/store.go
erikgrinaker: Hm, good question! I sort of considered that to be user data, but I'm not sure if it makes much sense to keep it when the blocks are gone. Can we even delete indexed data? The `TxIndexer` interface doesn't have any deletion methods.

Anyone else have an opinion here?

consensus/reactor.go
erikgrinaker: Yeah, the `nil` check will handle that as well as other failure scenarios. An atomic base+load won't help here, since the peer may be requesting the block at height base+20 concurrently with base to base+100 being pruned - we'll have to check for `nil`. The base check is just to avoid having to query the database at all if we know it can't be satisfied.

rpc/core/status.go
erikgrinaker: Good point, will add an atomic `LoadBase()`.

consensus/reactor.go
erikgrinaker: Yes, added a check for it in #5382. It's not strictly necessary, since `PickSendVote()` handles the case of a `nil` commit internally, but good to be explicit.

state/store.go
melekes: 1) If we were to use external indexer (Postgresql), would we also request removing these heights? I'd say no
2) how useful txindex is without blocks? one can query txs (without ability to prove them)

consensus/state.go
erikgrinaker: A related point here is that we currently only prune blocks during consensus, not during block replay. We may want to do both?

state/store.go
erikgrinaker: Yeah, my take is that we consider indexed data to be "external data" that is simply exported as it was at the time. As a user, I think that's the behavior I'd be least surprised by, but I'm open to discussing this.

consensus/reactor.go
erikgrinaker: In this case, with a bunch of different chained predicates, I think it's better to keep the entire predicate explicit and inline to avoid having to jump around to fully understand the logic - but it's a judgement call.

consensus/state.go
ebuchman: They do seem orthogonal but given pruning is controlled by ABCI return values and all ABCI handling is supposed to be covered by ApplyBlock, I think it makes sense to consolidate.

erikgrinaker: Opened #5414 for this, will pick it up once I wrap up some other projects.

PR #4602, opened 2020-03-24
valIndex int32 to uint32

marbar3778:
## Description

This pr is a follow up to #4582 in which it moves the validator index to a uint32 instead of a int32

______

For contributor use:

- [ ] Wrote tests


consensus/types/round_state.go
```diff
@@ -131,7 +134,10 @@ func (rs *RoundState) RoundStateSimple() RoundStateSimple {
 // NewRoundEvent returns the RoundState with proposer information as an event.
 func (rs *RoundState) NewRoundEvent() types.EventDataNewRound {
 	addr := rs.Validators.GetProposer().Address
-	idx, _ := rs.Validators.GetByAddress(addr)
+	idx, _, err := rs.Validators.GetByAddress(addr)
+	if err != nil {
+		panic(err) // if proposer is not present in val set, panic
```
marbar3778: @melekes not sure if its okay to panic here? Tried looking through the code to see where -1 gets checked (-1 was idx when val was not present before this pr) but can't find anywhere.

melekes: It's okay to panic (at least for now). since we expect proposer to be in validator set

consensus/state.go
```diff
@@ -1887,7 +1887,10 @@ func (cs *State) signVote(
 		return nil, errors.Wrap(err, "can't get pubkey")
 	}
 	addr := pubKey.Address()
-	valIdx, _ := cs.Validators.GetByAddress(addr)
+	valIdx, _, err := cs.Validators.GetByAddress(addr)
+	if err != nil {
+		return nil, fmt.Errorf("address: %v is not associated with an active validator, err: %w", addr, err)
```
melekes, line 1892:
```suggestion
		return nil, fmt.Errorf("address %X is not associated with an active validator: %w", addr, err)
```

evidence/pool.go
```diff
@@ -107,7 +107,10 @@ func (evpool *Pool) AddEvidence(evidence types.Evidence) (err error) {
 	// fetch the validator and return its voting power as its priority
 	// TODO: something better ?
 	valset, _ := sm.LoadValidators(evpool.stateDB, evidence.Height())
-	_, val := valset.GetByAddress(evidence.Address())
+	_, val, err := valset.GetByAddress(evidence.Address())
+	if err != nil {
+		return fmt.Errorf("address: %v is not associated with a validator, err: %w", evidence.Address(), err)
```
melekes, line 112:
```suggestion
		return fmt.Errorf("address %X is not associated with a validator: %w", evidence.Address(), err)
```

lite2/test_helpers.go
```diff
@@ -109,14 +114,13 @@ func makeVote(header *types.Header, valset *types.ValidatorSet,
 	}
 	// Sign it
 	signBytes := vote.SignBytes(header.ChainID)
-	// TODO Consider reworking makeVote API to return an error
 	sig, err := key.Sign(signBytes)
 	if err != nil {
-		panic(err)
+		return nil, fmt.Errorf("could not sign bytes, err: %w", err)
```
melekes, line 119:
```suggestion
		return nil, fmt.Errorf("could not sign bytes: %w", err)
```

types/validator_set.go
```diff
@@ -232,23 +232,23 @@ func (vals *ValidatorSet) HasAddress(address []byte) bool {
...
+func (vals *ValidatorSet) GetByAddress(address []byte) (index uint32, val *Validator, err error) {
 	idx := sort.Search(len(vals.Validators), func(i int) bool {
 		return bytes.Compare(address, vals.Validators[i].Address) <= 0
 	})
 	if idx < len(vals.Validators) && bytes.Equal(vals.Validators[idx].Address, address) {
-		return int32(idx), vals.Validators[idx].Copy()
+		return uint32(idx), vals.Validators[idx].Copy(), nil
 	}
-	return -1, nil
+	return 0, nil, fmt.Errorf("could not find validator at index: %v, %w", idx, err)
```
melekes, line 244: what error are u wrapping?

types/validator_set.go
```diff
@@ -232,23 +232,23 @@ func (vals *ValidatorSet) HasAddress(address []byte) bool {
 	return idx < len(vals.Validators) && bytes.Equal(vals.Validators[idx].Address, address)
 }
 
-// GetByAddress returns an index of the validator with address and validator
-// itself if found. Otherwise, -1 and nil are returned.
-func (vals *ValidatorSet) GetByAddress(address []byte) (index int32, val *Validator) {
+// GetByAddress returns an index of the validator with address and validator if found.
+// Otherwise, 0, nil and a error are returned.
+func (vals *ValidatorSet) GetByAddress(address []byte) (index uint32, val *Validator, err error) {
```
melekes, line 237: What's the benefit of adding error? aren't -1 and nil enough? I am struggling to see how adding an error here helps.

marbar3778: can't return negative with uint32, I can check if its a -1 and convert otherwise. I am not the biggest fan of using -1 when something is nil.

melekes: how about `(index uint32, val *Validator, ok bool)`? just don't see why we need an error here. validator is either present or not

marbar3778: changed it to `ok bool`

go.sum
```diff
@@ -77,6 +77,7 @@ github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f/go.mod h1:E3G3o1h8I7cfc
 github.com/cosmos/go-bip39 v0.0.0-20180819234021-555e2067c45d h1:49RLWk1j44Xu4fjHb6JFYmeUnDORVwHNkDxaQ0ctCVU=
 github.com/cosmos/go-bip39 v0.0.0-20180819234021-555e2067c45d/go.mod h1:tSxLoYXyBmiFeKpvmq4dzayMdCjCnu8uqmCysIGBT2Y=
 github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d/go.mod h1:maD7wRr/U5Z6m/iR4s+kqSMx2CaBsrgA7czyZG/E6dU=
+github.com/cpuguy83/go-md2man/v2 v2.0.0 h1:EoUDS0afbrsXAZ9YQ9jdu/mZ2sXgT1/2yyNng4PGlyM=
```
marbar3778, line 80: ???

types/validator_set.go
```diff
@@ -774,12 +774,12 @@ func (vals *ValidatorSet) VerifyCommitTrusting(chainID string, blockID BlockID,
 
 		// We don't know the validators that committed this block, so we have to
 		// check for each vote if its validator is already known.
-		valIdx, val, err := vals.GetByAddress(commitSig.ValidatorAddress)
-		if err != nil {
+		valIdx, val, okVal := vals.GetByAddress(commitSig.ValidatorAddress)
+		if !okVal {
```
marbar3778, line 778: should have a better name

go.sum
marbar3778: changes due to running go mod tidy

types/validator_set.go
melekes: `ok` is a Go standard

marbar3778:
build artifacts is failing due to the build on arm:
```
CGO_ENABLED=0 go build -mod=readonly -ldflags "-X github.com/tendermint/tendermint/version.GitCommit=`git rev-parse --short=8 HEAD` -s -w" -tags 'tendermint' -o build/tendermint_linux_arm ./cmd/tendermint/ 

# github.com/tendermint/tendermint/libs/math
libs/math/safemath.go:46:7: constant 4294967295 overflows int
make: *** [Makefile:22: build] Error 2
```

not actually sure how to fix this. @tessr @melekes @erikgrinaker Any pointers?

tessr:
So, this means that we're trying to build on 32-bit architecture, I think. I also thought we'd made the explicit decision at some point to _not_ support 32 bit architecture, although I now can't find any evidence of that decision.

tessr:
But if we want to fix this, I think we can convert the int to be an int64 before doing the comparison.

PR #4610, opened 2020-03-29
ADR-053: strengthen and simplify the state sync ABCI interface

erikgrinaker:
## Description

Strengthen and simplify the ADR-053 state sync ABCI interface based on implementation findings. The main changes are:

* Snapshot discovery now includes chunk SHA-256 hashes. If there are any mismatches across nodes, the majority wins. This serves two purposes:

  1. In the case of accidental non-determinism in snapshot generation across nodes, we can make sure all fetched chunks are part of the same snapshot such that it can still be restored (subject to `app_hash` verification).

  2. An adversary now has to be able to provide the chosen snapshot manifest, and we can use majority-based heuristics to counteract this. Before, it was sufficient for a single adversary to be able to provide a single invalid chunk to cause the entire snapshot restoration to fail.

* Chunks are now exchanged simply as `bytes`, their maximum size is reduced to 16 MB, and they use 0-based indexing for parity with the SHA-256 hash array.

In addition, I am considering making `ApplySnapshotChunk` a streaming call, such that we can simply pass a stream of chunks when restoring a snapshot. This relieves the application from having to handle the asynchronicity itself. However, this is a bit more controversial (none of the other ABCI calls are streaming), so I'd like to discuss it first.

______

For contributor use:

- [x] ~Wrote tests~
- [x] ~Updated CHANGELOG_PENDING.md~
- [x] ~Linked to Github issue with discussion and accepted design OR link to spec that describes this work.~
- [x] ~Updated relevant documentation (`docs/`) and code comments~

### DIFF
--- a/docs/architecture/adr-053-state-sync-prototype.md
+++ b/docs/architecture/adr-053-state-sync-prototype.md
@@ -303,6 +303,8 @@ $ ./tools/stop.sh
 
 * **Tendermint:** node should go back to fast-syncing when lagging significantly [#129](https://github.com/tendermint/tendermint/issues/129)
 
+* **Tendermint:** backfill historical blocks [#4629](https://github.com/tendermint/tendermint/issues/4629)
+
 ## Status
 
 Accepted
--- a/docs/architecture/adr-053-state-sync-prototype.md
+++ b/docs/architecture/adr-053-state-sync-prototype.md
@@ -20,7 +20,6 @@ This ADR outlines the plan for an initial state sync prototype, and is subject t
     * ABCI: renamed `GetSnapshotChunk` to `LoadSnapshotChunk`.
     * ABCI: chunks are now exchanged simply as `bytes`.
     * ABCI: chunks are now 0-indexed, for parity with `chunk_hashes` array.
-    * ABCI: added `internal_error` reasons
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-053-state-sync-prototype.md
```diff
@@ -72,41 +71,45 @@ message ResponseListSnapshots {
...
-    enum Reason {
-        unknown = 0;        // Unknown or generic reason
-        verify_failed = 1;  // Chunk verification failed
+    bool   applied = 1;
+    Reason reason  = 2;      // Reason why chunk failed
+
+    enum Reason {            // Reason why chunk failed
+        unknown        = 0;  // Unknown or generic reason
+        verify_failed  = 1;  // Snapshot verification failed
+        internal_error = 2;  // Something unexpected went wrong
```
melekes: isn't this the same as 0 (unknown)?

erikgrinaker: The idea was that `unknown` would be "rejected for an unknown reason", implying that Tendermint should offer other snapshots instead, while `internal_error` would cause Tendermint to abort. However, I suppose we could continue trying other snapshots for internal errors too, in which case we can remove this.

PR #4615, opened 2020-03-30
proto: use docker to generate stubs

### LINKED ISSUE #4579: use docker to generate protobuf stubs
## Summary

Currently, we require everyone to locally download a few tools to work with proto. This causes headaches and could possibly produce different outputs.

## Problem Definition

Different protobuf stubs could be generated if used in a non-version-compatible environment.

## Proposal

Use docker to generate stubs.

This will remove the need to download tools.
generate the same stubs no matter the local environment 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

marbar3778:
## Description

Add command to generate proto stubs with docker. This PR should not be merged till https://github.com/tendermint/images/pull/48 is and it is tested again.

Kept non-docker to generate proto files as some people may not want to run docker non-stop.

closes #4579

______

For contributor use:

~~- [] Wrote tests~~

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -23,7 +23,7 @@ Friendly reminder, we have a [bug bounty program](https://hackerone.com/tendermi
 - [Docker] \#4569 Default configuration added to docker image (you can still mount your own config the same way) (@greg-szabo)
 - [lite2] [\#4562](https://github.com/tendermint/tendermint/pull/4562) Cache headers when using bisection (@cmwaters)
 - [all] [\#4608](https://github.com/tendermint/tendermint/pull/4608) Give reactors descriptive names when they're initialized
-- [tools] \#4615 Allow developers to use docker to generate proto stubs.
+- [tools] \#4615 Allow developers to use Docker to generate proto stubs, via `make proto-gen-docker`.
 
 ### BUG FIXES:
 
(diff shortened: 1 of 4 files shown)


CHANGELOG_PENDING.md
```diff
@@ -22,7 +22,8 @@ Friendly reminder, we have a [bug bounty program](https://hackerone.com/tendermi
 - [privval] \#4534 Add `error` as a return value on`GetPubKey()`
 - [Docker] \#4569 Default configuration added to docker image (you can still mount your own config the same way) (@greg-szabo)
 - [lite2] [\#4562](https://github.com/tendermint/tendermint/pull/4562) Cache headers when using bisection (@cmwaters)
-- [all] [\4608](https://github.com/tendermint/tendermint/pull/4608) Give reactors descriptive names when they're initialized
+- [all] [\#4608](https://github.com/tendermint/tendermint/pull/4608) Give reactors descriptive names when they're initialized
+- [tools] \#4615 Allow developers to use docker to generate proto stubs.
```
erikgrinaker:
```suggestion
- [tools] \#4615 Allow developers to use Docker to generate proto stubs, via `make proto-gen-docker`.
```

scripts/protocgen.sh
```diff
@@ -7,5 +7,5 @@ for dir in $proto_dirs; do
   protoc \
   -I. \
   --gogo_out=Mgoogle/protobuf/timestamp.proto=github.com/golang/protobuf/ptypes/timestamp,Mgoogle/protobuf/duration.proto=github.com/golang/protobuf/ptypes/duration,plugins=grpc,paths=source_relative:. \
-  $(find "${dir}" -name '*.proto')
+  "$(find "${dir}" -name '*.proto')"
```
erikgrinaker: Does this work? If I'm reading this right, it will output all files as a single argument e.g. `"a b c"` not `"a" "b" "c"`. I suppose the intention is to handle file names containing whitespace, but I think that needs to be done via e.g. `find -exec` or something, since I don't think Bash supports it natively.

marbar3778:
```suggestion
  $(find "${dir}" -name '*.proto')
```

Issue #4580, opened 2020-03-17
evidence: remove PubKey from DuplicateVoteEvidence


melekes:
It does not serve any purpose other than returning `PubKey.Address()` in `DuplicateVoteEvidence.Address()`. We can get the same address from either of the votes (A or B).

PR #4617, opened 2020-03-31
evidence: remove pubkey from DuplicateVoteEvidence

marbar3778:
## Description

- remove Pubkey from DuplicateVoteEvidence
- this pr is meant to unblock block to proto work

ref #4580

______

For contributor use:

- [x] Wrote tests

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -19,7 +19,9 @@ Friendly reminder, we have a [bug bounty program](https://hackerone.com/tendermi
 - [types] \#4582 Block: `Round` is now int32
 - [consensus] \#4582 RoundState: `Round`, `LockedRound` & `CommitRound` are now int32
 - [consensus] \#4582 HeightVoteSet: `round` is now int32
+- [evidence] \#4617 Remove `Pubkey` from DuplicateVoteEvidence
 - [privval] \#4582 `round` in private_validator_state.json is no longer a string in json.
+
 - Apps
 
 - Go API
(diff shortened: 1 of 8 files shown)


abci/example/kvstore/persistent_kvstore.go
```diff
@@ -99,7 +99,6 @@ func (app *PersistentKVStoreApplication) Query(reqQuery types.RequestQuery) (res
 		if err != nil {
 			panic(err)
 		}
```
marbar3778:
```suggestion
		}

```

CHANGELOG_PENDING.md
```diff
@@ -19,7 +19,9 @@ Friendly reminder, we have a [bug bounty program](https://hackerone.com/tendermi
 - [types] \#4582 Block: `Round` is now int32
 - [consensus] \#4582 RoundState: `Round`, `LockedRound` & `CommitRound` are now int32
 - [consensus] \#4582 HeightVoteSet: `round` is now int32
+- [evidence] \#4617 Remove `Pubkey` from DuplicateVoteEvidence
```
melekes: why is it breaking again?

marbar3778: amino encoding would change. some nodes in a network would have the full type and some wouldn't I believe this would cause issues. If not then I can make the changes to merge it to master

melekes: That may be true. I guess the only way is to check 😛

marbar3778: it is breaking. Removing a field in amino would change the field number. If we try to decode on a 0.33 deployed net it will fail.

Issue #4618, opened 2020-03-31
lite2: VerifyCommitTrusting will report double vote error


melekes:
if `ValidatorAddress` is not found in `valSet` more than once

Fix: do not check `seenVals` if validator with `ValidatorAddress` is not found (idx=-1, val=nil)

marbar3778:
We should remove `-1` as a placeholder for nil, `GetByAddress` should return an error and then we can remove the dependence on `-1`. This is a bit of a refactor but its mainly just checking for `err`

melekes:
> We should remove -1 as a placeholder for nil, GetByAddress should return an error and then we can remove the dependence on -1. This is a bit of a refactor but its mainly just checking for err

I disagree.

PR #4620, opened 2020-03-31
lite2: Prevent falsely returned double voting error

cmwaters:
Closes: #4618

______

For contributor use:

- [ ] Wrote tests


cmwaters:
Q: How come we check for double voting for `VerifyCommitTrusting()` but not for `VerifyCommit()` ?

melekes:
we need to write a test here

melekes:
> Q: How come we check for double voting for VerifyCommitTrusting() but not for VerifyCommit() ?

because in VerifyCommit we know all validators (1 - 1), if vote from earlier validator is used - signature verification will fail.

in VerifyCommitTrusting, we don't know all validators

cmwaters:
I created tests but I'm still not convinced about needing to check for double votes.

> in VerifyCommitTrusting, we don't know all validators

It's not about the validators though, we are checking to see if a commit has two votes and if we're assuming the commit is not faulty then this double vote can only occur when the validator set has two of the exact validators which shouldn't occur if someone is using a validator set properly.

melekes:
> assuming the commit is not faulty

what makes you think you can make that assumption? malicious attackers can construct any data

Issue #4573, opened 2020-03-16
Pex reactor punishment policy should match spec


cmwaters:
## Summary

The pex reactor code, with specific reference to the punishment policy, must be aligned with the spec. Furthermore, there are cases of peer behavior in which the suitable punishment has not yet been determined. 

## Problem Definition

In regards to the [pex spec](https://github.com/tendermint/spec/blob/master/spec/reactors/pex/pex.md), in order to prevent spam, there is a 'Disconnect and Mark' procedure which involves disconnecting with the peer and banning them for a period of time. Such listed examples are "if a peer sends us an unsolicited list of peers, or if the peer sends a request too soon after another one, we Disconnect and MarkBad". 

Currently however:
- If a peer sends an unsolicited list of peers and error is returned which run's `StopPeerForError`: We disconnect but don't ban
- If a peer sends a request too soon after another one we run `StopPeerForError`: We disconnect peer but don't ban 

**Do we want to change the code to match the spec?**

Otherwise:
- If a peer doesn't respond within the max attempts (last attempt being 15 hours), we ban for 24 hours before the peer is added again and can be connected. 
- If in an attempt to dial a peer, there is an `ErrSwitchAuthenticationFailure` the peer is banned for 24 hours
- If a peer gives us a banned address we ignore it and log an error

**Are there any other cases we want to punish**

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #4621, opened 2020-03-31
p2p: PEX message abuse should ban as well as disconnect

cmwaters:
Closes: #4573

## Outstanding Issue

If a peer sends a request too soon after another one we should disconnect and mark bad. 
However, when I run a test, if a peer sends a request we don't necessarily save them in the address book hence if they send us too many requests we can't ban them because they aren't in the address book in the first place. What do you think I should do about it?
a) make sure when a request from a peer comes in we always add it to the address book
b) we allow addresses to be banned that weren't in the address book to begin with
**c) we ignore this part of the spec and we only disconnect from them (CURRENTLY)**

______

For contributor use:

- [x] Wrote tests


p2p/pex/errors.go
```diff
@@ -72,3 +72,12 @@ type ErrAddressBanned struct {
 func (err ErrAddressBanned) Error() string {
 	return fmt.Sprintf("Address: %v is currently banned", err.Addr)
 }
+
+// ErrUnsolicitedList is thrown when a peer provides a list of addresses that have not been asked for -> Disconnect and MarkBad
```
alexanderbez, line 76:
```suggestion
// ErrUnsolicitedList is thrown when a peer provides a list of addresses that have not been asked for.
// Handling the error should disconnect the peer and mark it as bad.
```

p2p/pex/errors.go
```diff
@@ -72,3 +72,12 @@ type ErrAddressBanned struct {
 func (err ErrAddressBanned) Error() string {
 	return fmt.Sprintf("Address: %v is currently banned", err.Addr)
 }
+
+// ErrUnsolicitedList is thrown when a peer provides a list of addresses that have not been asked for.
+type ErrUnsolicitedList struct {
```
melekes, line 77: again, I don't understand a need for exposing this error

why not

```go
var errUnsolicitedList = errors.New("unsolicited pexAddrsMessage")
```

cmwaters: I'm easy either way, but bez was saying in the other PR that it's good to expose typed errors so that external logic can use them to check the errors. With the check evidence PR, I changed an error back to public again so the RPC could use them. Perhaps it would be a good idea to make a clear ruling of when to use private or public error types

cmwaters: Okay, so from what I understand, `ErrUnsolicitedList` is a potential error from the public function `RecieveAddrs` and so should therefore be public

cmwaters: I could drop the peer as a variable of the error as this is part of the input of the function and so we can assume the external logic already has it.

melekes: ErrUnsolicitedList{src} - caller should add src if needed

PR #4623, opened 2020-03-31
docs/architecture: add adr for proto repo design

marbar3778:
## Description

- Defining the proto types in a proto directory `/proto`
- define util functions for going back and forth from proto and handwritten types where it is needed.

______

For contributor use:

- [ ] Wrote tests


docs/architecture/adr-055-protobuf-design.md
```diff
@@ -0,0 +1,45 @@
...
+
+- `FlatBuffers`: FlatBuffers is an efficient cross platform serialization library. Flatbuffers are more efficient than Protobuf due to the fast that there is no parsing/unpacking to a second representation. FlatBuffers has been tested and used in production but is not widely adopted.
+
+- `CapnProto`: Cap’n Proto is an insanely fast data interchange format and capability-based RPC system. Cap'n Proto does not have a encoding/decoding step. It has not seen wide adoption throughout the industry.
+
+## Decision
+
+Transition Tendermint to Protobuf because of its performance and tooling. The Ecosystem behind Protobuf is vast and has outstanding [support for many languages](https://developers.google.com/protocol-buffers/docs/tutorials).
+
+To make this move possible we will not be removing handwritten types but instead creating a `/proto` directory in which all the `.proto` files and types will live. Then where encoding is needed, on disk and over the wire, we will call util functions that will transition the types from handwritten go types to protobuf generated types.
```
erikgrinaker: Any reason why we'll be putting all files in a single directory, instead of keeping them in the modules where they are used?

docs/architecture/adr-055-protobuf-design.md
```diff
@@ -0,0 +1,45 @@
...
+
+Currently we use [go-amino](https://github.com/tendermint/go-amino) throughout Tendermint. Amino enables quick prototyping and development of features. While this is nice amino does not provide the performance and developer convenience that is expected. For this reason moving to a different encoding format that is widely adopted and supports multiple languages is needed.
+
+There are a few options to pick from:
+
+- `Protobuf`: Protocol buffers are Google's language-neutral, platform-neutral, extensible mechanism for serializing structured data – think XML, but smaller, faster, and simpler. It is supported in countless languages and has been proven in production for many years.
+
+- `FlatBuffers`: FlatBuffers is an efficient cross platform serialization library. Flatbuffers are more efficient than Protobuf due to the fast that there is no parsing/unpacking to a second representation. FlatBuffers has been tested and used in production but is not widely adopted.
+
+- `CapnProto`: Cap’n Proto is an insanely fast data interchange format and capability-based RPC system. Cap'n Proto does not have a encoding/decoding step. It has not seen wide adoption throughout the industry.
```
erikgrinaker: Cap'n'Proto is awesome. It was written by one of the original Protobuf developers to fix some of its issues, and supports e.g. random access to process huge messages without loading them into memory and an (opt-in) canonical form which would be very useful when determinism is needed (e.g. in the state machine). That said, I suspect Protobuf is the better choice due to wider adoption, although it makes me kind of sad since Cap'n'Proto is technically better.

docs/architecture/adr-055-protobuf-design.md
```diff
@@ -0,0 +1,45 @@
+# ADR 055: Protobuf Design
+
+## Changelog
+
+- 20202-3-31: Created
```
erikgrinaker:
```suggestion
- 2020-03-31: Created
```

docs/architecture/adr-055-protobuf-design.md
marbar3778: It will be easier in the future when we decide to pull the proto files out of the repo and into the spec repo for versioned specs. It can go either way, a proto dir in the module its being used or one proto dir with directories for modules. I believe not tying the type to the module may be beneficial, but these benefits will not be seen for a while.

```
├── proto
|   ├── consensus
|   |   ├── types.proto
|   └── types
|       ├── types.proto
```

erikgrinaker: Ah, yeah, that makes sense.

docs/architecture/adr-055-protobuf-design.md
marbar3778: yes agreed. I would love to have it configurable in the future, similar to how the sdk currently/will have.

erikgrinaker: I'm not sure if that would make sense, since we'd have to assume a lowest common denominator of features. E.g. we can't use GB-sized messages for anything in Tendermint, because if you configure it with Protobuf then it's going to blow up.

docs/architecture/adr-055-protobuf-design.md
```diff
@@ -0,0 +1,45 @@
...
+
+- `FlatBuffers`: FlatBuffers is an efficient cross platform serialization library. Flatbuffers are more efficient than Protobuf due to the fast that there is no parsing/unpacking to a second representation. FlatBuffers has been tested and used in production but is not widely adopted.
+
+- `CapnProto`: Cap’n Proto is an insanely fast data interchange format and capability-based RPC system. Cap'n Proto does not have a encoding/decoding step. It has not seen wide adoption throughout the industry.
+
+## Decision
+
+Transition Tendermint to Protobuf because of its performance and tooling. The Ecosystem behind Protobuf is vast and has outstanding [support for many languages](https://developers.google.com/protocol-buffers/docs/tutorials).
+
+To make this move possible we will not be removing handwritten types but instead creating a `/proto` directory in which all the `.proto` files and types will live. Then where encoding is needed, on disk and over the wire, we will call util functions that will transition the types from handwritten go types to protobuf generated types.
```
melekes, line 23: Is there a generator for util functions? Is it possible to write one?

I sort of hate bringing a pile of new code that looks like https://github.com/tendermint/tendermint/blob/master/types/protobuf.go. Not a lot of overhead, but still, we'll have to support it.

marbar3778: there wasn't anything I was able to find. We could write a tool to do this, but then it may add more overhead to write and maintain the tool.

tessr: > it may add more overhead to write and maintain the tool 

That sounds likely to me.

docs/architecture/adr-055-protobuf-design.md
```diff
@@ -0,0 +1,45 @@
+# ADR 055: Protobuf Design
+
+## Changelog
+
+- 2020-03-31: Created
```
tessr, line 5: do you want to put your name on this? that seems to be the convention. 

(otoh, not sure why every ADR needs a changelog when everything gets checked into git anyways.)

docs/architecture/adr-055-protobuf-design.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Context
+
+Currently we use [go-amino](https://github.com/tendermint/go-amino) throughout Tendermint. Amino enables quick prototyping and development of features. While this is nice amino does not provide the performance and developer convenience that is expected. For this reason moving to a different encoding format that is widely adopted and supports multiple languages is needed.
```
tessr, line 9: Are the various grievances documented anywhere? I know you've kept at least a mental tally of projects that have run into trouble with Amino, but it could be useful to have the issues thoroughly documented.

docs/architecture/adr-055-protobuf-design.md
tessr, line 23: Just to make sure I understand your use of the phrase "handwritten types" - right now we have all "handwritten types" and whenever we need to serialize them we use amino; in the future, we will continue to use these handwritten types but whenever we need to serialize them we will instead turn them into proto-generated types, and _then_ serialize the proto-generated types. Right?

docs/architecture/adr-055-protobuf-design.md
```diff
@@ -0,0 +1,45 @@
...
+
+Proposed
+
+## Consequences
+
+### Positive
+
+- Allows for modular types in the future
+- Less refactoring
+- Allows the proto files to be pulled into the spec repo in the future.
```
tessr, line 37: Performance?

docs/architecture/adr-055-protobuf-design.md
marbar3778: yes that is correct. Will better define what handwritten types in the adr

docs/architecture/adr-055-protobuf-design.md
tessr: I actually think Erik's top comment here should go into this ADR; it's exactly the kind of discussion that belongs in an ADR for future reference!

